PR #14685, opened 2026-07-14
🔒 Replace hardcoded zero IV with secure random IV in AndroidDecrypter

dogi:
🎯 **What:** Fixed a vulnerability in `AndroidDecrypter.encrypt` where passing a `null` IV resulted in a hardcoded, all-zero IV byte array being used for AES encryption.
⚠️ **Risk:** A hardcoded, predictable IV breaks the semantic security of block ciphers like AES-CBC. If an attacker observes multiple ciphertexts encrypted with the same key and the same all-zero IV, they can identify identical plaintext blocks and potentially perform chosen-plaintext attacks, severely compromising data confidentiality.
🛡️ **Solution:** Updated `encrypt` to securely generate a random IV using `SecureRandom().nextBytes()` whenever the provided `iv` is `null`. Updated `decrypt` to handle `null` fallback IVs safely by dynamically extracting the 16-byte random IV prepended to the ciphertext, while fully preserving backward compatibility for legacy data using explicitly provided IVs.

---
*PR created automatically by Jules for task [836664628873329941](https://jules.google.com/task/836664628873329941) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/utils/AndroidDecrypter.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/utils/AndroidDecrypter.kt
@@ -18,7 +18,12 @@ class AndroidDecrypter {
             val clean = plainText.toByteArray()
             val ivSize = 16
             val ivBytes = ByteArray(ivSize)
-            iv?.let { hexStringToByteArray(it) }?.let { System.arraycopy(it, 0, ivBytes, 0, ivBytes.size) }
+            if (iv != null) {
+                val decodedIv = hexStringToByteArray(iv)
+                System.arraycopy(decodedIv, 0, ivBytes, 0, minOf(decodedIv.size, ivBytes.size))
+            } else {
+                SecureRandom().nextBytes(ivBytes)
+            }
             val ivParameterSpec = IvParameterSpec(ivBytes)
             val keyBytes = ByteArray(32)
             key?.let { hexStringToByteArray(it) }?.let { System.arraycopy(it, 0, keyBytes, 0, keyBytes.size) }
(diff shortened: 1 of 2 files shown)


dogi:
**Rating: 48/100** (largely false-premise — `close?` is reasonable)

**What's actually correct**
- The round-trip is coherent: `encrypt()` already prepends `ivBytes` to the ciphertext (`AndroidDecrypter.kt:29-31`), so generating a random IV when `iv == null` and extracting the first 16 bytes on decrypt (`initVector == null`) does round-trip. The updated `AndroidDecrypterTest` confirms it.

**Why the premise is weak**
- The app **already uses per-user random IVs.** At the sole encrypt call site (`UserRepositoryImpl.kt:1006`), `val iv = userModel?.iv ?: AndroidDecrypter.generateIv()` — `iv` is never null, and `generateIv()` produces a random IV persisted per user. So the new "iv == null → SecureRandom IV" branch in `encrypt()` is **dead code in practice**; there is no zero/fixed-IV vulnerability on the live path to fix.
- Meanwhile the decrypt side changes a real contract: previously `initVector == null` returned `null`; now it treats the first 16 bytes of the payload as an IV. `RealmHealthExamination.decrypt(data, model.key, model.iv)` passes a stored `model.iv` — for any legacy/edge record where that's null, decrypt now silently reinterprets ciphertext bytes as an IV instead of failing closed. That's a behavior change on the read path with no offsetting benefit.

**If pursued anyway:** the valuable hardening isn't here — it'd be ensuring every record has a unique IV and never reusing key+IV, which the existing `generateIv()` design already does.

**Verdict:** Correct code solving a non-existent problem on the encrypt path while subtly loosening the decrypt contract. Agree with `close?`.

_Leaving `close?` as-is. Repo-owner-authored._

---

PR #14690, opened 2026-07-14
all: smoother server url mapper testing (fixes #14757)

dogi:
🎯 **What:** The testing gap for `ServerUrlMapper` addressed by adding missing unit tests.
📊 **Coverage:** Covered scenarios for `processUrl` (handling known, unknown and invalid URLs), `updateUrlPreferences` (validating URI preferences stored via SharedPreferences accurately), and `updateServerIfNecessary`.
✨ **Result:** Increased test coverage for synchronization utility guaranteeing robustness for alternative and offline URL manipulations.

---
*PR created automatically by Jules for task [8299959463757740708](https://jules.google.com/task/8299959463757740708) started by @dogi*

PR #14539, opened 2026-07-08
🔒 Encrypt sensitive passwords in SharedPrefManager

dogi:
🎯 **What:** The vulnerability fixed
Sensitive server passwords (`URL_PWD` and `new_login_password`) were being stored in cleartext in SharedPreferences within `SharedPrefManager`.

⚠️ **Risk:** The potential impact if left unfixed
Any malicious app with root access, or anyone with physical access to an unlocked or debuggable device, could read these passwords, potentially compromising backend systems and user data.

🛡️ **Solution:** How the fix addresses the vulnerability
Modified `SharedPrefManager` to use the existing `SecurePrefs` utility class. The passwords are now securely encrypted before being written to SharedPreferences and decrypted when read, leveraging Tink's AES256 encryption.

---
*PR created automatically by Jules for task [6855350444963536172](https://jules.google.com/task/6855350444963536172) started by @dogi*

PR #14691, opened 2026-07-14
🔒 Fix plaintext credential storage in SharedPreferences

dogi:
🎯 **What:** The vulnerability fixed
Plaintext storage of user credentials (`url_user` and `url_pwd`) within SharedPreferences.

⚠️ **Risk:** The potential impact if left unfixed
Malicious apps with physical access, root access, or exploit chains could read the plaintext SharedPreferences file and exfiltrate the user's password and login token for the server.

🛡️ **Solution:** How the fix addresses the vulnerability
Utilized the existing `SecurePrefs` utility class to encrypt the credentials before writing them to the SharedPreferences, and implemented a safe fallback-and-migrate function to handle any existing plaintext credentials during a rolling update.

Resolves the security issue identified in `ServerUrlMapper.kt:70`.

---
*PR created automatically by Jules for task [7491054309821687408](https://jules.google.com/task/7491054309821687408) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/services/SharedPrefManager.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/services/SharedPrefManager.kt
@@ -161,11 +161,31 @@ class SharedPrefManager @Inject constructor(
     fun getCouchdbUrl(): String = pref.getString(COUCHDB_URL, "") ?: ""
     fun setCouchdbUrl(url: String) = pref.edit { putString(COUCHDB_URL, url) }
 
-    fun getUrlUser(): String = pref.getString(URL_USER, "") ?: ""
-    fun setUrlUser(user: String) = pref.edit { putString(URL_USER, user) }
+    fun getUrlUser(): String {
+        val encrypted = pref.getString(URL_USER, null)
+        if (encrypted != null) {
+            val decrypted = org.ole.planet.myplanet.utils.SecurePrefs.decryptString(context, encrypted)
+            if (decrypted != null) return decrypted
+            // Fallback for unencrypted legacy value
+            setUrlUser(encrypted)
+            return encrypted
+        }
+        return ""
+    }
+    fun setUrlUser(user: String) = pref.edit { putString(URL_USER, org.ole.planet.myplanet.utils.SecurePrefs.encryptString(context, user)) }
 
-    fun getUrlPwd(): String = pref.getString(URL_PWD, "") ?: ""
-    fun setUrlPwd(pwd: String) = pref.edit { putString(URL_PWD, pwd) }
+    fun getUrlPwd(): String {
+        val encrypted = pref.getString(URL_PWD, null)
+        if (encrypted != null) {
+            val decrypted = org.ole.planet.myplanet.utils.SecurePrefs.decryptString(context, encrypted)
+            if (decrypted != null) return decrypted
+            // Fallback for unencrypted legacy value
+            setUrlPwd(encrypted)
+            return encrypted
+        }
+        return ""
+    }
+    fun setUrlPwd(pwd: String) = pref.edit { putString(URL_PWD, org.ole.planet.myplanet.utils.SecurePrefs.encryptString(context, pwd)) }
 
     fun getUrlScheme(): String = pref.getString(URL_SCHEME, "") ?: ""
     fun setUrlScheme(scheme: String) = pref.edit { putString(URL_SCHEME, scheme) }
--- a/app/src/main/java/org/ole/planet/myplanet/services/SharedPrefManager.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/services/SharedPrefManager.kt
@@ -146,8 +146,21 @@ class SharedPrefManager @Inject constructor(
     fun getServerUrl(): String = pref.getString(SERVER_URL, "") ?: ""
     fun setServerUrl(url: String) = pref.edit { putString(SERVER_URL, url) }
 
-    fun getServerPin(): String = pref.getString(SERVER_PIN, "") ?: ""
-    fun setServerPin(pin: String) = pref.edit { putString(SERVER_PIN, pin) }
+    fun getServerPin(): String {
+        val encrypted = pref.getString(SERVER_PIN, null)
+        if (encrypted != null) {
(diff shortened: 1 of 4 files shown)


dogi:
**Rating: 86/100**

**Strengths**
- Real, *complete* security fix — not a false-premise one. `url_user`/`url_pwd` were stored in plaintext; this routes them through `SecurePrefs.encryptString`/`decryptString`, mirroring the pattern this class already uses for `new_login_username`/`new_login_password`.
- Verified completeness: every consumer reads through the decrypting getters — `AuthSessionUpdater.kt:89-90` (login), `UrlUtils.kt:32` (`getUrlUser():getUrlPwd()` for the Basic auth header) — and every writer encrypts (`ProcessUserDataActivity`, `ServerConfigUtils`, and the direct `editor` write in `ServerUrlMapper.updateUrlPreferences`). No raw `getString("url_user"/"url_pwd")` reader remains, so nothing gets a ciphertext it can't handle.
- Sensible legacy migration: if `decryptString` returns null (pre-existing plaintext), the getter returns the raw value and re-persists it encrypted. Tink AEAD's auth tag means a real ciphertext won't silently "decrypt" to garbage, so the null-means-legacy heuristic is sound. CI (build + test) is green.

**Concerns**
- ⚠️ **Cross-PR conflict with #14690.** This adds `@ApplicationContext context` to `ServerUrlMapper`'s constructor, but #14690's `ServerUrlMapperTest` does `serverUrlMapper = ServerUrlMapper()` (no-arg). Each passes CI alone; merged together they won't compile. Whichever lands second must update the test to pass a context (e.g. `ServerUrlMapper(ApplicationProvider.getApplicationContext())`). Please coordinate the two.
- Minor: the corrupt-ciphertext edge (decrypt fails on a value that *was* encrypted) would re-encrypt ciphertext-as-plaintext. Extremely unlikely with AEAD; noting for completeness.

**Verdict:** Correct, complete, tested credential-at-rest fix — mergeable; just resolve the #14690 constructor conflict.

---

Okuro3499 (dismissed):
```
1. SharedPrefManager.setUrlUser/setUrlPwd (lines 175, 188) and ServerUrlMapper.updateUrlPreferences (lines 82-83) — unhandled crypto exceptions. SecurePrefs.encryptString →       
  getAead → buildAead can throw (GeneralSecurityException/IOException from AndroidKeysetManager) on devices with flaky keystores; the previous plain putString could never throw, so 
  this converts a save into an app crash during login/sync. SecurePrefs.saveCredentials wraps its own crypto in try/catch for exactly this reason. Fix: catch the failure (mirroring 
  saveCredentials) and degrade gracefully — e.g., skip the write or fall back — rather than crashing.                                                                                
  2. getUrlUser()/getUrlPwd() fallback (SharedPrefManager.kt:167-171, 180-184) — key loss silently poisons the credential. When decryption fails on a genuinely-encrypted value, the 
  ciphertext blob is returned as the credential and then re-encrypted and persisted, so every subsequent read yields Base64 garbage and auth fails opaquely. Fix: store encrypted    
  values with a version prefix (e.g., enc1:<base64>) so ciphertext and legacy plaintext are unambiguous; on prefix-present-but-undecryptable, return "" (forcing re-config/re-login) 
  instead of writing back garbage.                                                                                                                                                   
  3. Incomplete remediation — the same secret remains plaintext in the same file. ServerConfigUtils.saveUrlScheme (lines 108-114) writes the password to serverPin and embeds it in  
  processedAlternativeUrl; ProcessUserDataActivity:148 and ServerUrlMapper.kt:87 embed it in couchdbURL/processedAlternativeUrl. Encrypting url_pwd while its exact value sits       
  unencrypted two keys away provides no real protection. Fix: extend the same treatment to serverPin, couchdbURL, and processedAlternativeUrl (or stop persisting credential-embedded
  URLs and reconstruct them at read time from the encrypted parts). If the team explicitly wants a phased approach, the commit/PR must say so — as written, "Fix plaintext           
  credential storage" overstates what it does. 
```

dogi:
**Re-review (revising my earlier `ready`) → Rating: 58/100 — `changes requested`.** Two things changed since: #14690 merged, and a new `enc1:`-prefix commit landed. That turns the constructor conflict I'd flagged as a coordination note into a real blocker, and surfaces a duplicate.

**Blocker 1 — breaks the build on merge (hidden by a stale base)**
- This branch is **22 commits behind master** and predates #14690, so `ServerUrlMapperTest.kt` isn't in its tree — that's why `test (default)` is green here.
- `ServerUrlMapper`'s constructor now requires `@ApplicationContext context`, but current master's `ServerUrlMapperTest` still does `serverUrlMapper = ServerUrlMapper()` (no args) and this PR doesn't touch it. Merging into current master → `testDefaultDebugUnitTest` fails to compile.
- **Fix:** rebase onto master and update `ServerUrlMapperTest` to pass a context (Robolectric `ApplicationProvider.getApplicationContext()` or a mock).

**Blocker 2 — duplicate of #14539**
- #14539 does the same job: encrypt SharedPreferences credentials via a prefix scheme **and** add `@ApplicationContext` to `ServerUrlMapper.updateUrlPreferences`. They will hard-conflict, and they use **incompatible prefixes** (`enc1:` here vs `enc:` there) — if both ever land, values written by one are read as "corrupt/plaintext" by the other and get re-encrypted → data loss. Only one may proceed.

**Comparison (for choosing):**
- This PR (#14691): covers `SERVER_PIN`, `COUCHDB_URL`, `URL_USER`, `URL_PWD`, `PROCESSED_ALTERNATIVE_URL`; centralizes the prefix by teaching `SecurePrefs.decrypt` to strip `enc1:`. **Pro:** encrypts `url_user` (a credential). **Cons:** the 5 getters are copy-pasted (not DRY), heavy inline FQNs, and it couples the generic `SecurePrefs` to a SharedPrefManager naming convention; `ALTERNATIVE_URL` left plaintext.

**Recommendation:** consolidate into a single PR. I'd keep **#14539**'s DRY `getSecureString`/`setSecureString` helper (and its generic `SecurePrefs`) and fold in this PR's `url_user` coverage — then close the other. Whichever is kept must rebase + fix `ServerUrlMapperTest`.

Repo-owner-authored → needs a second reviewer regardless.

---

PR #14698, opened 2026-07-14
Consolidate queryListFlow dispatching and implement duplicate-emission guard

dogi:
Consolidate queryListFlow dispatching and implement duplicate-emission guard

- Modified RealmRepository.queryListFlow to perform copyFromRealm mapping within the same callbackFlow scope instead of a secondary flowOn block.
- Implemented a duplicate-emission guard that drops identical dataset emissions (by comparing extracted primary key lists) unless the Realm changeSet indicates genuine content updates (changes.isNotEmpty()).
- Ensures observer lifecycles still appropriately handle content insertions/deletions/modifications cleanly.

---
*PR created automatically by Jules for task [15060020594730460603](https://jules.google.com/task/15060020594730460603) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/repository/RealmRepository.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/RealmRepository.kt
@@ -66,11 +66,13 @@ open class RealmRepository(
     protected fun <T : RealmObject> queryListFlow(
         clazz: Class<T>,
         builder: RealmQuery<T>.() -> Unit = {},
-    ): Flow<List<T>> = callbackFlow<RealmResults<T>> {
+    ): Flow<List<T>> = callbackFlow<List<T>> {
         val isClosed = AtomicBoolean(false)
         var realm: Realm? = null
         var results: RealmResults<T>? = null
         var listener: OrderedRealmCollectionChangeListener<RealmResults<T>>? = null
+
+        var lastPrimaryKeys: List<Any>? = null
 
         fun safeCloseRealm() {
             if (isClosed.compareAndSet(false, true)) {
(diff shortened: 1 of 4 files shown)


dogi:
**Rating: 52/100**

This touches `RealmRepository.queryListFlow` — the base flow used by **every** reactive repository query — so the bar for complexity/regression is high, and two things here work against the stated goal.

**Concern 1 — copy now happens *before* `conflate()` (efficiency regression)**
- Old pipeline: `callbackFlow { trySend(frozen) }.flowOn(realmDispatcher).conflate().map { copyFromRealm(...) }.flowOn(io)`. Cheap frozen refs are emitted, `conflate()` drops superseded ones, and only the survivors get the expensive `copyFromRealm`.
- New pipeline: the `copyFromRealm` happens **inside** the producer, before `.conflate()`. So every raw change notification is fully deep-copied even when it will immediately be conflated away, and the `copyFromRealm` work also moves off the `io` dispatcher onto `realmDispatcher`. Under bursty Realm notifications this is strictly more work than before — the opposite of an optimization.

**Concern 2 — reflection-based PK dedup is the wrong tool**
- `emitResults` now, per emission, reads `schema.get(clazz.simpleName)?.primaryKey`, then `getDeclaredField(...).isAccessible = true` and maps the field across the whole list (with `"id"`/`"_id"` fallbacks) to build `currentPrimaryKeys`. That's fragile (field-name coupling, per-emit reflection cost) and only compares **keys** — it can't detect content changes, so it leans on the `hasContentChanges` flag to avoid dropping real updates.
- The idiomatic equivalent is one line: `.distinctUntilChanged()` on the already-detached `List<T>` (data-class/`equals`-based), or a lightweight version key. That dedups identical consecutive emissions without reflection and without the copy-before-conflate reordering.

**Verdict:** Reasonable intent (suppress duplicate emissions), but the implementation regresses the copy/conflate ordering and reimplements `distinctUntilChanged` with per-emission reflection in a hot, widely-shared base class. Rework using `distinctUntilChanged` and keep `copyFromRealm` after `conflate()`. Another round.

---

dogi:
**Re-review → Rating: 82/100 — both flags fixed, lifting `changes requested`.**

The two concerns in this hot base-class path are resolved:
- **Reflection removed:** the per-emission `schema.primaryKey` + `getDeclaredField(...)` PK dedup is gone, replaced by a plain `.distinctUntilChanged()` on the emitted `List<T>` — the idiomatic dedup I suggested, no fragile field-name coupling.
- **Copy after conflate restored:** the pipeline is back to emitting cheap frozen refs → `.flowOn(realmDispatcher)` → `.conflate()` → `.map { copyFromRealm(...) }` → `.distinctUntilChanged()`, so superseded emissions are dropped *before* the expensive deep copy.

Marking **ready**. Repo-owner-authored → needs a second reviewer.

---

Okuro3499 (changes requested):
```
1. RealmRepository.kt:141 — distinctUntilChanged() is a no-op for non-empty lists. Unmanaged copyFromRealm copies use identity equality, so duplicates are never suppressed. Fix by
  comparing something with real equality semantics: distinctUntilChanged { old, new -> ... } with a key comparison (e.g., primary key + updated/revision field per element), or      
  implement equals/hashCode on the affected models. Avoid the commit-1 reflection approach — pass a key selector into queryListFlow instead.                                         
  2. RealmRepository.kt:134-142 — copyFromRealm moved onto the single shared RealmQueryThread. Frozen results are thread-safe, so the copy doesn't need Realm's thread; restore      
  .flowOn(databaseService.ioDispatcher) after the map/distinctUntilChanged stage (keeping the producer's flowOn(realmDispatcher)) so heavy copies run on the IO pool as before.      
  3. Missing test for the new behavior. Add a RealmRepositoryTest case asserting that re-writing identical content yields no second emission. As written today, that test would fail 
  — which is the point. 
```

PR #14702, opened 2026-07-14
teams: smoother life progress repositories data filtering (fixes #14791)

dogi:
Replaced slow in-memory filtering of large database result sets with highly optimized Realm DB query operators (`equalTo`, `in`, `contains`) across multiple repositories (`Voices`, `Life`, `Teams`). Optimized `Progress` repository fetching to avoid nested O(N^2) filtering logic. Updated affected mock tests to support chained queries.

---
*PR created automatically by Jules for task [18015824585490436179](https://jules.google.com/task/18015824585490436179) started by @dogi*

PR #14709, opened 2026-07-14
teams: smoother tasks list updating (fixes #14772)

### LINKED ISSUE #14772: Optimize TeamsTasksFragment updates

dogi:
Optimizes the task UI updates in `TeamsTasksFragment`. It avoids redundant and expensive diffing in the adapter and unneeded DB queries for assignees when the core task states (IDs and completed status) have not changed. This is accomplished by keeping a localized snapshot state.

---
*PR created automatically by Jules for task [16881638343739394469](https://jules.google.com/task/16881638343739394469) started by @dogi*


dogi:
**Rating: 58/100**

**Strengths**
- The `allTasks()/completedTasks()/myTasks()` → `allTasks(tasks)/…(tasks)` refactor is a clean improvement: they now operate on a single captured `tasksSnapshot` instead of re-reading `teamViewModel.taskList.value` three times, which also removes a subtle race where the list could change between calls. More testable, too.

**Concern (dedup key too narrow — can drop real updates)**
- The skip guard uses `currentSnapshot = list.map { it.id to it.completed }` and early-returns when it equals `lastSubmittedSnapshot`. But a task's **deadline, assignee, or title** can change while `id` and `completed` stay the same. When that happens (e.g. a sync edits a deadline without changing sort order, or reassigns a task), the snapshot compares equal and the entire update — including `adapterTask.updateAssignees(...)` and `submitList` — is skipped, leaving stale rows. The `ListAdapter`/DiffUtil the adapter presumably already uses would handle no-op submits cheaply, so this coarse guard mostly trades correctness for a marginal saving.
  - **Fix:** either drop the guard and rely on the adapter's DiffUtil for redundant-update suppression, or widen the snapshot to include every rendered field that can change (deadline, assignee, title) — at minimum the assignee set, since `updateAssignees` depends on it.
- ⚠️ `test_plan.sh` was committed at the repo root — remove it (same stray-script issue as #14702).

**Verdict:** Good refactor undermined by an over-narrow change-detection key plus a stray script. Another round.

---

dogi:
**Re-review → Rating: 80/100 — flag fixed, lifting `changes requested`.**

The over-narrow dedup key is fixed: the snapshot is now a `TaskSnapshot(id, title, description, deadline, completed, assignee)` — capturing every rendered field, not just `(id, completed)`. So a change to a task's deadline/assignee/title/description no longer compares equal and the update (incl. `updateAssignees`) is no longer skipped. The `test_plan.sh` stray script is removed. The `allTasks(tasks)/…(tasks)` snapshot refactor stands.

Marking **ready**. Repo-owner-authored → needs a second reviewer.

---

Okuro3499 (dismissed):
```
1. TeamsTasksFragment.lastSubmittedSnapshot — first-render suppression for empty teams. The initial value emptyList() is indistinguishable from "already rendered an empty list",  
  so showNoData never fires for a team with no tasks. Fix: declare it as List<TaskSnapshot>? = null and compare against null-safe equality, so the first pass always renders.        
  2. TeamsTasksFragment — stale snapshot survives view recreation while the adapter does not. After back-stack restoration the new adapter stays empty forever. Fix: reset           
  lastSubmittedSnapshot = null in onCreateView (or onViewCreated, next to where adapterTask is recreated), so the guard's lifetime matches the adapter's.
```

dogi:
**Re-review (new commit) — still `ready`.** My flagged concern remains resolved: the dedup key is the full `TaskSnapshot(id, title, description, deadline, completed, assignee)` — all rendered fields — so a change to deadline/assignee/title/description no longer compares equal and the update (incl. `updateAssignees`) isn't skipped. The stray `test_plan.sh` remains removed. CI green. Keeping **ready** — needs a second reviewer.

---

Okuro3499 (approved):

PR #14711, opened 2026-07-14
Refactor Handler(Looper.getMainLooper()) posts to coroutines

dogi:
Replaced `Handler(Looper.getMainLooper()).post` calls with `Dispatchers.Main` coroutines in `Utilities.kt` and `CameraUtils.kt` to modernize asynchronous operations to use coroutines instead of explicitly relying on `Handler.post` for main thread dispatch where feasible.

---
*PR created automatically by Jules for task [18241539203888246073](https://jules.google.com/task/18241539203888246073) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/utils/CameraUtils.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/utils/CameraUtils.kt
@@ -181,7 +181,7 @@ object CameraUtils {
                         captureSession = session
                         try {
                             captureRequestBuilder.set(CaptureRequest.CONTROL_AF_MODE, CaptureRequest.CONTROL_AF_MODE_CONTINUOUS_PICTURE)
-                            captureSession?.setRepeatingRequest(captureRequestBuilder.build(), null, null)
+                            captureSession?.setRepeatingRequest(captureRequestBuilder.build(), null, mainHandler)
                         } catch (e: CameraAccessException) {
                             e.printStackTrace()
                         }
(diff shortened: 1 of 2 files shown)


dogi:
**Rating: 60/100**

**Concern — the `Utilities.toast` conversion is an anti-pattern, not an improvement**
- `Utilities.kt:47`: `Handler(Looper.getMainLooper()).post { … }` → `CoroutineScope(Dispatchers.Main).launch { … }`. This creates a brand-new, **unstructured `CoroutineScope` on every toast**, tied to no lifecycle and never cancelled. `Utilities` is a stateless `object` with no lifecycle owner, so there's nothing to scope to — which is exactly the case where `Handler(mainLooper).post {}` is the correct, idiomatic tool. The coroutine version adds an allocation and a GlobalScope-like scope for zero benefit (both just dispatch a Runnable to the main Looper).
  - **Fix:** keep `Handler(Looper.getMainLooper()).post { }` here. If you specifically want coroutines, at least don't spin up a throwaway `CoroutineScope` per call.

**Minor**
- `CameraUtils.kt:184`: `setRepeatingRequest(..., mainHandler)` → `..., null` is fine (the capture callback is `null`, so the handler is unused), but it's unrelated to "Handler → coroutines" — and check whether `mainHandler` is now unused.

**Verdict:** The premise (Handler-post is worse than coroutines) doesn't hold for a lifecycle-less utility; this trades a correct idiom for an un-scoped coroutine. Keep `Handler.post`. Another round.

---

PR #14714, opened 2026-07-14
life: smoother repository list order caching (fixes #14790)

### LINKED ISSUE #14790: Optimize MyLife persistence and caching

dogi:
Optimized My Life persistence to scale efficiently. Created an ID-to-order map before the update write loop instead of scanning for indices quadratically. Caching is now refreshed only if weight actually changed.

---
*PR created automatically by Jules for task [4507668871506623647](https://jules.google.com/task/4507668871506623647) started by @dogi*


dogi:
**Rating: 86/100**

**Strengths**
- Two genuine improvements in `updateMyLifeListOrder`:
  1. O(n²) → O(n): the per-item `list.indexOfFirst { it._id == managedLife._id }` is replaced by a precomputed `idToIndex` map with O(1) lookup.
  2. Avoids redundant writes: `if (index != null && managedLife.weight != index)` only touches Realm when the weight actually changes, and tracks `hasChanges`.
- Cache consistency: after a real reorder it refreshes via `getMyLifeByUserId(userId, ensureLatest = true)` + `cacheMyLifeItems`, so the persisted cache no longer drifts from the reordered DB state. Nice.

**Nits**
- `idToIndex` uses `toMap()`, which keeps the *last* entry on duplicate `_id`s, whereas the old `indexOfFirst` kept the first. `_id`s are unique here so it's moot, but worth being aware of.
- Touches `LifeRepositoryImpl` like #14702 (different method); sequence the rebases.

**Verdict:** Correct optimization plus a real cache-sync fix; mergeable as-is.

---

Okuro3499 (approved):

PR #14692, opened 2026-07-14
teams: smoother voices filter view modelling (fixes #14789)

dogi:
Refactor Voices filtering and conflate flow emissions

- Extracted search and label filtering logic into a pure function `filterNews` in `VoicesViewModel`.
- Removed unnecessary `withContext` hop from `filteredNews` combine block to reduce overhead.
- Added `.conflate().distinctUntilChanged()` to repository flows in both `VoicesViewModel` and `TeamsVoicesViewModel` to reduce UI churn.
- Added comprehensive unit tests in `VoicesViewModelTest` to verify combined search and label filtering functionality.

---
*PR created automatically by Jules for task [9619177939927008976](https://jules.google.com/task/9619177939927008976) started by @dogi*

PR #14715, opened 2026-07-14
Remove redundant withContext hop in ProgressRepositoryImpl.hasUserCompletedSync

dogi:
Removed the redundant `withContext` hop in `ProgressRepositoryImpl.hasUserCompletedSync` and cleaned up the `dispatcherProvider` dependency as it became unused in the file. Fixed the tests accordingly.

---
*PR created automatically by Jules for task [8924270610815784387](https://jules.google.com/task/8924270610815784387) started by @dogi*

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced redundant updates in live list flows by suppressing duplicate emissions when the effective item set hasn’t changed (including handling empty results).
  * Improved change handling so updates occur when non-primary-key fields change while primary keys remain the same.

* **Tests**
  * Updated flow emission assertions for initial results timing and mocked result sizes.
  * Added tests to confirm duplicates are dropped when change details are null and content is unchanged, and that updates still emit when non-key fields change.


dogi:
**Rating: 88/100**

**Strengths**
- Correct redundancy removal. `hasUserCompletedSync` delegated to `activitiesRepositoryLazy.get().hasUserCompletedSync(userId)`, which itself is `withContext(realmDispatcher) { … }` (`ActivitiesRepositoryImpl.kt:332`). So the outer `withContext(dispatcherProvider.io)` was a pointless second dispatch hop — removing it is behavior-preserving and slightly cheaper.
- Nicely thorough: since `dispatcherProvider` was this class's only remaining use, it's dropped from the constructor and the unused imports removed, and the test is updated to match the new signature.

**Nit**
- Constructor signature change touches DI + tests — coordinate with the other in-flight `ProgressRepositoryImpl` PRs (#14702 edits the same file) so one rebase cleanly follows the other.

**Verdict:** Correct, cleaner, and complete (incl. test); mergeable as-is.

---

Okuro3499 (dismissed):
```
 1. RealmRepository.kt:141 — distinctUntilChanged() cannot deduplicate. Realm models lack equals overrides and copies are fresh instances, so the guard only affects consecutive    
  empty lists. Fix: dedupe on a real key — e.g., distinctUntilChanged with a comparator over primary keys (ideally computed from the frozen results before map, so duplicate         
  emissions skip the copy entirely), or restore a cleaned-up version of the first commit's PK comparison without reflection (read the @PrimaryKey field via the Realm schema on the  
  frozen results).                                                                                                                                                                   
  2. RealmRepository.kt:142 — restore the copyFromRealm stage to a parallel dispatcher. Keep .flowOn(realmDispatcher) scoped to the callbackFlow (required: addChangeListener needs  
  the Looper thread) and put .flowOn(databaseService.ioDispatcher) back after the map, as on master. The single-HandlerThread consolidation degrades a hot path shared by every      
  reactive list in the app and buys nothing once fix #1 is done properly.                                                                                                            
  3. Add a regression test in RealmRepositoryTest.kt that fires the captured listener with changeSet = null and content equal to the previous emission, and asserts only one         
  collection is received. This is the behavior the branch claims to add and is currently unverified (and currently false). 
```

dogi:
**Re-review — Rating: 28/100. Setting `changes requested` (was previously ready; the branch has been rebased into a different, broken state).**

Heads-up: this no longer resembles what I approved earlier ("remove redundant `withContext` hop in `ProgressRepositoryImpl`"). The current head has three problems:

**1. ⚠️ A merge-conflict artifact is committed.** `app/src/main/java/org/ole/planet/myplanet/repository/RealmRepository.kt.orig` (233 lines) is a `.orig` backup file left over from a conflicted rebase and checked into source. Delete it and add `*.orig` to `.gitignore`.

**2. Scope/title mismatch.** The diff contains **no `ProgressRepositoryImpl` change** anymore — the original approved change appears to have been lost in the rebase. Instead the PR now modifies `RealmRepository.queryListFlow`. (CodeRabbit's title check flags the same.) The description and the change no longer agree.

**3. ⚠️ Correctness regression in the shared flow (affects every reactive query).** The new
```kotlin
.distinctUntilChanged { old, new -> … compare size + primary-key of each element … }
```
compares consecutive emissions by **primary key + size + order only**. But Realm's `OrderedRealmCollectionChangeListener` emits on `changeSet.changes` — i.e. *in-place field modifications* where the PK, size and order are unchanged. For those, this comparator returns `true`, so `distinctUntilChanged` **drops the emission**. Net effect: any `queryListFlow` whose item contents mutate in place (a `RealmNews.message`/`labels` edit, a task `completed` toggle, an edited chat message, etc.) will **stop propagating to the UI**. This is the base helper, so it hits every repository flow in the app.
- The two added tests don't catch it: `TestRealmObject` has **no primary key**, so `schema.primaryKey == null` → the comparator returns `false` (always emits) → tests pass while real, PK-bearing models behave differently. CI green is not meaningful for this path.
- It also allocates two `DynamicRealmObject`s per element on every emission (hot path), and works against the just-merged #14692/#14734, which deliberately added *content-aware* distinct (id + mutation time) at specific repositories rather than a PK-only global one.

**To move forward:** delete the `.orig` file; then either (a) restore this PR to its original `ProgressRepositoryImpl` scope, or (b) if flow-dedup is the real goal, compare by **content** (not PK-only) so modifications still emit, and add a test using a model that actually has a primary key. As-is it's not mergeable.

---

dogi:
**Re-review (new head `e08ffe9`) — still `changes requested`. Rating: 42/100.** (Label had been flipped back to `review`; resetting to `changes requested`.)

**Progress:** the committed `RealmRepository.kt.orig` merge artifact is now gone — thanks for removing it.

**Still blocking — the primary-key-only `distinctUntilChanged` suppresses in-place content updates.** The comparator returns `true` (treat as duplicate → **drop the emission**) whenever `old.size == new.size` and every element's **primary key** matches. But Realm fires the change listener on `changeSet.changes`, i.e. in-place field edits where the PKs and size are unchanged — so those emissions are dropped. Net: any `queryListFlow` consumer stops seeing content edits (a task `completed` toggle, a `RealmNews.message`/`labels` edit, an edited chat message) until list *membership* (size or PKs) changes. This is the shared base helper, so it hits **every** reactive query in the app.

The new test `queryListFlow emits when non-primary-key field changes but primary keys match` doesn't actually catch this: with the relaxed-mocked schema, `newRealm.schema.get(clazz.simpleName)?.primaryKey` resolves to `null`, so the comparator takes the `if (pkField == null) return@distinctUntilChanged false` branch and emits — it never reaches the PK-comparison path. Real models all declare a primary key, so production behavior diverges from the green test.

**Fix:** dedupe by **content**, not primary key only — e.g. compare the frozen results' contents, or build a per-item signature (id + a version/`updatedDate`/mutation-timestamp) so a modification produces a different key and still emits. Then write the test against a model that actually has a primary key so the PK-match path is exercised.

Also unchanged from my last pass: the diff contains no `ProgressRepositoryImpl` change despite the PR title/description, and qlty flags `queryListFlow` (total complexity / 8 returns). Keeping **changes requested**.

---

PR #14724, opened 2026-07-14
all: smoother application anr watchdog scoping (fixes #14780)

### LINKED ISSUE #14780: Consolidate CoroutineScope creations to MainApplication.applicationScope

dogi:
This PR removes localized/detached `CoroutineScope` instantiations outside of ViewModels or Fragments (specifically in `ANRWatchdog.kt` and `ServiceModule.kt`) and routes them through the shared `MainApplication.applicationScope`. 

This guarantees background continuity while avoiding unintended scope cancellations across the application lifecycle. Tests for the ServiceModule DI configuration and SubmissionUploadExecutor were also updated with proper mocking and tear-down logic.

---
*PR created automatically by Jules for task [8644214980085367096](https://jules.google.com/task/8644214980085367096) started by @dogi*


dogi:
**Rating: 18/100** ⚠️ **Startup-crash regression (masked by the test)**

**Critical bug — self-referential initialization**
- `ServiceModule.provideApplicationScope` now returns `CoroutineScope(MainApplication.applicationScope.coroutineContext + dispatcherProvider.io)`.
- But `MainApplication.applicationScope` is a `lateinit var` that is **assigned from this very provider**: `setupCriticalProperties()` does `applicationScope = EntryPointAccessors.fromApplication(this, CoreDependenciesEntryPoint::class.java).applicationScope()` (`MainApplication.kt:307`).
- So the first time the provider runs (to initialize `applicationScope`), it reads `MainApplication.applicationScope.coroutineContext` while that `lateinit` is **not yet assigned** → `UninitializedPropertyAccessException`, crashing the app during `onCreate`. It's a circular dependency: the provider depends on the field the provider initializes.
- This is invisible to CI: the updated `ServiceModuleTest` does `mockkObject(MainApplication)` + `every { MainApplication.applicationScope } returns mockScope`, so the test never exercises the real uninitialized path. The unit suite stays green while the shipped app fails at launch.

**Also**
- `ANRWatchdog` now reaches into `MainApplication.applicationScope` statically instead of building a scope from its injected `dispatcherProvider`, coupling a utility to the `Application` singleton and reducing testability. (Job cancellation via `job?.cancel()` still works, so that part is behaviorally OK — but it's the wrong direction for testability, and it shares the same static-init fragility.)

**Fix:** Don't source `provideApplicationScope` from `MainApplication.applicationScope`. If the goal is a single shared scope, invert it: let `MainApplication.applicationScope` simply hold the DI-provided scope (as today, `CoroutineScope(SupervisorJob() + io)`), and have other call sites depend on the injected `@ApplicationScope` — not the other way around. Also confirm the shared scope keeps a `SupervisorJob` so one failed child can't cancel all app-level work.

**Verdict:** Reasonable goal (one app scope), but the implementation creates a circular init that crashes at startup and is hidden by a mock in the test. Do not merge as-is.

---

dogi:
**Re-review → Rating: 84/100 — flag fixed, lifting `changes requested`.**

Verified the startup-crash flag is resolved:
- `ServiceModule.provideApplicationScope` is back to `CoroutineScope(SupervisorJob() + dispatcherProvider.io)` — no more self-referential read of the `MainApplication.applicationScope` `lateinit` during its own init, so the `UninitializedPropertyAccessException` is gone. `SupervisorJob` (failure isolation) is preserved too.
- `ANRWatchdog` now takes the scope via constructor (`ANRWatchdog(scope = applicationScope, …)` at `MainApplication.kt:344`) instead of reaching into the static, and `stop()` cancels only the `job` (not the shared scope). This is the clean version of the consolidation — testable and no static coupling.

Marking **ready**. Still repo-owner-authored → needs a second reviewer to merge.

---

Okuro3499 (approved):

PR #14720, opened 2026-07-14
dashboard: smoother user repositroy profile view modelling (fixes #14808)

dogi:
Extracted the dashboard profile loading logic (full name processing with fallback and offline login counts) out of `DashboardViewModel` and into `UserRepositoryImpl` via `DashboardProfile` data class, isolating the UI from explicit DB fallback patterns.

---
*PR created automatically by Jules for task [2483973889407753751](https://jules.google.com/task/2483973889407753751) started by @dogi*

## Summary by CodeRabbit

* **New Features**
  * Added dashboard profile data, including the user’s display name and offline login count.
  * Dashboard information now loads through a unified profile service.

* **Bug Fixes**
  * Improved name fallback handling when profile names are incomplete or unavailable.
  * Added safeguards for missing user profiles.

* **Tests**
  * Expanded coverage for profile formatting, missing users, offline login counts, and dashboard state updates.

* **Chores**
  * Updated the app version to 0.61.27.

PR #14716, opened 2026-07-14
all: smoother view model scoping (fixes #14820)

dogi:
This PR removes redundant context switches to the IO dispatcher in several `ViewModel`s across the application. 

By analyzing the underlying repository layer, it was confirmed that many Realm database operations (like `withRealm`, `withRealmAsync`, `queryList`, and `executeTransactionAsync`) natively dispatch to their own worker threads. Explicitly invoking `viewModelScope.launch(dispatcherProvider.io)` for these calls added unnecessary overhead. 

The changes update these to use the standard main-thread-bound `viewModelScope.launch { ... }`. Operations that rely on synchronous `SharedPreferences` logic (such as clearing preferences or managing saved users) correctly retain their explicit IO dispatcher. Tests were executed to ensure functionality remains intact.

---
*PR created automatically by Jules for task [2618968746091654506](https://jules.google.com/task/2618968746091654506) started by @dogi*

## Summary by CodeRabbit

* **Refactor**
  * Updated background work in courses, feedback, notifications, ratings, settings, submissions, teams, voices, and user profiles to use standard lifecycle-aware coroutine execution.
  * Preserved existing loading, update, success/error, and state behavior.
* **Tests**
  * Adjusted ViewModel test setups to match the streamlined background-processing approach.
  * Kept existing assertions and test intent the same.
* **Release**
  * Bumped app version to **0.61.40** (version code **6140**).

PR #14731, opened 2026-07-14
all: smoother repositories api typed parameters writing (fixes #14819)

### LINKED ISSUE #14819: Refactor repository write APIs to use typed parameters

dogi:
Refactored repository write APIs to use strongly typed parameters or data classes instead of generic `JsonObject` instances, fully moving JSON serialization and assembly into the repository implementation layer, ensuring better separation of concerns across ViewModels and Activities.

---
*PR created automatically by Jules for task [7254164647439007932](https://jules.google.com/task/7254164647439007932) started by @dogi*

## Summary by CodeRabbit

* **Improvements**
  * Updated finance report, meetup, member registration, and feedback reply flows to use shared, strongly-typed submission inputs for more consistent handling.
  * Centralized payload creation so meetups and reports get consistent IDs/timestamps during save/update.
* **Bug Fixes**
  * Feedback replies now reliably record the reply message and author.
* **Tests**
  * Refreshed repository and ViewModel tests to match the updated meetup creation and feedback reply behavior.
* **Chores**
  * Bumped app version to **0.61.37**.


dogi:
**Rating: 80/100**

**Strengths**
- Good direction: JSON/document construction moves out of Activities/ViewModels into the repositories, keyed by typed params (`MemberInfo`, `FinanceReportParams`), so the UI no longer hand-builds Couch docs. `MemberInfo` is promoted from a private inner class to a shared model.
- Verified behavior-preserving on the two heaviest paths:
  - `createMember(MemberInfo)` rebuilds the member JSON field-for-field vs the old `buildMemberJson` (same keys, `prefData.getPlanetCode()` == `sharedPrefManager.getPlanetCode()`, roles `[learner]`, androidId/device fields). ✅
  - `addReport`/`updateReport(FinanceReportParams)` produce the same docs as the old inline `EnterprisesViewModel` builders (matching field sets; `updateReport` correctly omits `teamId/teamType/teamPlanetCode`, so the dummy `""`/`null` args are harmless), and the image attach is relocated into the repo with equivalent effect.

**Concerns / nits**
- ⚠️ `TeamsRepositoryImpl.addReport`/`updateReport` now instantiate `SystemTimeProvider()` directly instead of using the class's **injected `timeProvider`** (which `updateReport` used before). That reintroduces non-deterministic time into a class that already has an injectable clock — use the injected `timeProvider` so tests stay deterministic.
- Style: pervasive fully-qualified names (`org.ole.planet.myplanet.model.FinanceReportParams`, `com.google.gson.JsonObject`, `java.util.UUID`, …) — add imports.
- `FinanceReportParams` is built with 13 **positional** args at the call sites; use named arguments (and consider that it conflates add vs. update, hence the dummy fields) to avoid a silent arg-order mistake.

**Verdict:** Broad but faithful API-typing refactor; mergeable after switching back to the injected `timeProvider` and tidying the FQNs — nits, no behavior bug found.

---

Okuro3499 (dismissed):
```
1. TeamsRepositoryImpl.addReport (line 611) / updateReport (line ~658) — both create a local SystemTimeProvider() (val timeProvider = ... / val systemTimeProvider = ...) that     
  shadows/ignores the class's own injected timeProvider: TimeProvider field (constructor param, line 69), which is used correctly everywhere else in this file. This directly undoes 
  the project's TimeProvider DI migration (done specifically so createdDate/updatedDate are deterministic and testable — see prior TestTimeProvider work). Fix: delete the local     
  instantiation and use the existing injected timeProvider field.                                                                                                                    
  2. EventsRepositoryImpl.createMeetup — instantiates SystemTimeProvider() inline rather than injecting TimeProvider into the constructor (the TimeProvider that used to be injected 
  into TeamCalendarViewModel was simply deleted, not moved). Fix: inject TimeProvider into EventsRepositoryImpl's constructor and use it, consistent with TeamsRepositoryImpl.       
  3. Layering violation — EventsRepository.kt: the repository interface (repository/ package, data layer) now takes org.ole.planet.myplanet.ui.teams.MeetupCreationParams as a       
  parameter type, making the data layer depend on a UI-package type. This inverts the architecture described in CLAUDE.md (UI → Repository → Service → Data). Fix: move              
  MeetupCreationParams to model/ (as was correctly done for FinanceReportParams and MemberInfo), and update TeamCalendarViewModel/TeamCalendarFragment imports accordingly.          
  4. EnterprisesViewModel — after moving report-building into TeamsRepositoryImpl, the constructor still injects timeProvider: TimeProvider and the file still imports               
  JsonObject/UUID, all now completely unused (verified via grep — timeProvider has zero remaining reads). Fix: remove the unused constructor param and imports; update its Hilt call 
  sites/tests if any depend on the constructor shape.  
```

dogi:
**Re-review (new commit) — flag fixed, still `ready`.**

My primary concern is resolved: `TeamsRepositoryImpl.addReport`/`updateReport` no longer instantiate `SystemTimeProvider()` directly — they use the class's injected `timeProvider.now()`, and `EventsRepositoryImpl` now takes `TimeProvider` via constructor (with `EventsRepositoryImplTest` passing `SystemTimeProvider()`), so time stays deterministic/injectable across these repos. The doc assembly remains field-for-field faithful to the old inline builders (`createMember`, `addReport`/`updateReport`, `createMeetup`), and the `updateReport` path correctly ignores the dummy `teamId=""`/`teamType=null`/`teamPlanetCode=null` positional args.

CI green on the latest head (20:07Z). Remaining items are the same non-blocking nits from before — the pervasive fully-qualified names (`org.ole.planet.myplanet.model.…`, `com.google.gson.…`, `java.util.UUID`) would read better as imports, and `FinanceReportParams` is still built with 13 positional args (named args would guard against an order slip). Style only.

⚠️ Rebase/conflict (not a defect): this changes `FeedbackDetailViewModel.addReply(id, message, user)` and keeps `dispatcherProvider.io`, while **#14716** removes `dispatcherProvider` from that same ViewModel and keeps the old `addReply(id, obj)` — the two can't both land unrebased. It also touches `UserRepositoryImpl` alongside #14720/#14725/#14712. Keeping **ready** — needs a second reviewer.

---

Okuro3499 (approved):

PR #14733, opened 2026-07-14
all: smoother realm database indexing (fixes #14657)

### LINKED ISSUE #14657: Indexing gaps
```
RealmMyPersonal.userId, RealmNews.replyTo, and RealmSubmission.parentId are all queried repeatedly without an @Index, while correctly-indexed siblings (RealmMyCourse.courseId, RealmMyTeam.teamId) show the pattern is known and just inconsistently applied.
```

Okuro3499:
fixes #14657
Bumps Realm schema version to 17 and extends migrations to keep existing installs compatible. The migration now backfills/indexes `RealmMyLibrary.titleNormal` for normalized title search, treats v16 as a no-op step, and adds indexes for frequently queried fields: `RealmMyPersonal.userId`, `RealmNews.replyTo`, and `RealmSubmission.parentId`.


dogi:
**Rating: 88/100**

Nice work — this is a correctly-constructed Realm migration, which is easy to get wrong.

**Verified correct**
- **Fixes #14657 in the right place.** `RealmMyLibrary.titleNormal` was introduced in `f6f55ec` at the 14→15 schema bump *with no migration*, so users upgrading from ≤14 crashed on the missing field. Adding `addField("titleNormal").addIndex(...).transform { … }` inside the `if (version == 14L)` block (the 14→15 step) targets exactly the stranded cohort. Users who could only be at 15/16 are fresh installs that already have the field (from the model), and they correctly skip this block — so no double `addField`.
- **Index consistency (fresh vs upgrade) checks out.** Each new index is added in both places: `@Index` on the model field *and* an `addIndex` in the migration — `titleNormal` (v14 block), and `RealmMyPersonal.userId` / `RealmNews.replyTo` / `RealmSubmission.parentId` (v16 block). So fresh installs (schema from model) and upgraders (migration) end up with the same indexes.
- `schemaVersion` 16→17 bump matches the new schema changes; the `version == 15L` no-op block (with its explanatory comment) keeps the counter sequence honest.
- The chosen indexes are well-motivated — `parentId`/`replyTo`/`userId` back hot query paths (submissions-by-course, reply filtering, personals-by-user).

**Nits**
- Migrations aren't covered by tests here, so this rests on manual reasoning — consider a Robolectric migration test that opens an old-schema realm and asserts fields/indexes exist, to lock #14657 from regressing.
- Defensive option: guarding `addField`/`addIndex` with `hasField(...)`/`hasIndex(...)` checks makes migrations resilient to any future out-of-order/partial states, though it's not required given the analysis above.

**Verdict:** Correct, consistent, and well-placed migration; mergeable as-is. Thanks for the careful fix.

_(External contribution — not authored by the repo owner, so this one can go through the normal approve/merge path once a reviewer signs off.)_

---

dogi (approved):

PR #14734, opened 2026-07-14
teams: smoother voices label managing (fixes #14787)

### LINKED ISSUE #14787: Refactor VoicesLabelManager to remove direct model mutation

dogi:
Refactor VoicesLabelManager to remove direct model mutation

Removes the direct, UI-side mutation of `voice.labels` when adding or removing labels. Instead, it relies on the repository functions to handle the data update. A new `onLabelChanged` callback is introduced to notify the respective Fragments/Activities, which then trigger a targeted `notifyItemChanged` on their adapter to refresh the UI cleanly based on the latest database state.

---
*PR created automatically by Jules for task [1815994800575694554](https://jules.google.com/task/1815994800575694554) started by @dogi*


dogi:
**Rating: 45/100**

**Good intent**
- Removing the direct `voice.labels?.add(...)` / `voice.labels?.remove(...)` mutation from `VoicesLabelManager` is the right instinct — mutating a Realm-model's list from the UI callback is fragile. Threading an `onLabelChanged` callback out to the hosts is a clean seam.

**Concern — the label chips no longer actually refresh (functional regression)**
- The old code force-updated the local copy (`voice.labels.add/remove`) so the immediately-following `showChips(...)` reflected the change. The new code drops that and instead does `onLabelChanged(voice)` → `notifyItemChanged(index)`. But `notifyItemChanged(index)` rebinds **the same `currentList[index]` object**, whose `labels` were *not* mutated — and `showChips` reads `voice.labels` — so the chip set is re-rendered from the **old** labels. Tapping add/remove-label produces no visible change.
- The reactive flow can't cover for it either: `VoicesAdapter.areContentsTheSame` (`VoicesAdapter.kt:82`) compares `id/time/isEdited/message/userName/userId/sharedBy` but **not `labels`**. So even when `getCommunityNews`/`getDiscussionsByTeamIdFlow` re-emits an updated copy with the new label, DiffUtil reports "contents same" and skips the rebind.
- Net effect: add/remove label writes to the DB but the chips don't update until the list is fully rebuilt (e.g. re-entering the screen).

**Fix:** rely on the reactive flow but make it observable — add `labels` (content-equality of the label list) to `areContentsTheSame` so a re-emit rebinds the row; then `onLabelChanged`/`notifyItemChanged` is unnecessary. (Or, if you keep the callback, have it submit an updated item rather than `notifyItemChanged` on the unchanged object.)

**Verdict:** Right direction, but as written the UI no longer reflects label changes. Another round.

---

dogi:
**Re-review → Rating: 82/100 — flag fixed, lifting `changes requested`.**

The "chips don't refresh" flag is addressed: `VoicesAdapter.areContentsTheSame` now compares labels —
`… && oldItem.sharedBy == newItem.sharedBy && oldItem.labels?.toList() == newItem.labels?.toList()`.
So after `addLabel`/`removeLabel` writes to Realm, the `queryListFlow` re-emits an updated `RealmNews` copy, DiffUtil now detects the label change, and the row rebinds → `showChips` renders the new set. That closes the gap left by removing the direct `voice.labels` mutation.

Marking **ready**. One thing to sanity-check on device (not blocking): confirm the add/remove-label write actually propagates through the observed flow so the re-emit fires — the diff comparison is correct, it just relies on that emission. Repo-owner-authored → needs a second reviewer.

---

Okuro3499 (approved):

PR #14796, opened 2026-07-16
resources: smoother creation handing (fixes #14603)

### LINKED ISSUE #14603: delay in mylibrary page when adding new resources
https://github.com/user-attachments/assets/fecad5f6-ef9a-4837-9ea6-5a1115b03db6

Valerie923:
fixes #14603 
https://github.com/user-attachments/assets/d2525e57-3653-4200-8e94-f4cf4ae0d153

## Summary by CodeRabbit

* **Bug Fixes**
  * Resource lists now refresh every time you return to the Resources screen.
  * Newly added resources now appear automatically after completing the add-resource flow.
  * The add-resource flow now properly reports successful completion back to the previous screen.
* **Chores**
  * Updated the app version metadata (version code/name).

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -59,3 +59,6 @@ PLANET_CAMBRIDGE_PIN=6407
 # took egdirbmac cause planet on cambridge is at 8080
 #PLANET_EGDIRBMAC_URL=planet.cambridge.ole.org
 #PLANET_EGDIRBMAC_PIN=1565
+
+# Enabled parallel sync for Gradle 9.4+
+org.gradle.tooling.parallel=true
--- a/gradle.properties
+++ b/gradle.properties
@@ -59,6 +59,3 @@ PLANET_CAMBRIDGE_PIN=6407
 # took egdirbmac cause planet on cambridge is at 8080
 #PLANET_EGDIRBMAC_URL=planet.cambridge.ole.org
 #PLANET_EGDIRBMAC_PIN=1565
-
-# Enabled parallel sync for Gradle 9.4+
-org.gradle.tooling.parallel=true
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -12,8 +12,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6121
-        versionName = "0.61.21"
+        versionCode = 6149
+        versionName = "0.61.49"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/ui/resources/AddResourceActivity.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/resources/AddResourceActivity.kt
@@ -195,6 +195,7 @@ class AddResourceActivity : AppCompatActivity() {
                 )
                 if (result.isSuccess) {
                     toast(this@AddResourceActivity, getString(R.string.resource_updated))
+                    setResult(RESULT_OK)
                     finish()
                 } else {
                     toast(this@AddResourceActivity, getString(R.string.failed_to_update_resource))
@@ -234,6 +235,7 @@ class AddResourceActivity : AppCompatActivity() {
                     getString(R.string.added_to_my_library)
                 }
                 toast(this@AddResourceActivity, message)
+                setResult(RESULT_OK)
                 finish()
             } else {
                 binding.tlTitle.error = getString(R.string.resource_title_already_exists)
--- a/app/src/main/java/org/ole/planet/myplanet/ui/resources/AddResourceFragment.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/resources/AddResourceFragment.kt
@@ -274,6 +274,7 @@ class AddResourceFragment : BottomSheetDialogFragment() {
                     viewModel,
                     viewLifecycleOwner
                 ) {
+                    parentFragmentManager.setFragmentResult("resource_added", Bundle())
                     dismiss()
                 }
             }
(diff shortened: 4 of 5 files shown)


dogi:
**Rating: 60/100 — `changes requested`.**

The core idea is right — refresh the library when a resource is added instead of waiting — and the `setFragmentResult("resource_added")` → `setFragmentResultListener` path for the bottom-sheet `AddResourceFragment` is a clean way to do it. But there are three things to address before merge:

1. ⚠️ **Unrelated change in a sensitive file.** `gradle.properties` gains `org.gradle.tooling.parallel=true`. That has nothing to do with the MyLibrary delay, and `gradle.properties` is tracked and holds the `PLANET_*` credentials — changes here are sensitive and shouldn't ride along in a UI fix. Please drop it (or split into its own build PR). (Also the comment says "Gradle 9.4+", but the repo is on 9.3.0.)

2. **`onResume` now refreshes on every resume.** Removing the `isFirstResume` guard means `refreshResourcesData()` runs on *every* return to `ResourcesFragment`, not just the first. That guard existed to avoid redundant full-list refreshes; always-refreshing risks reintroducing jank on a populated library — the opposite of this PR's goal. Prefer a targeted trigger: keep the `setFragmentResult` listener for the bottom-sheet path, and for the `AddResourceActivity` path use `registerForActivityResult` (you already added `setResult(RESULT_OK)`) to refresh only when something was actually added — rather than a blanket resume refresh.

3. **Confirm the fragment-result wiring.** `AddResourceFragment` posts via `parentFragmentManager.setFragmentResult("resource_added", …)`, while `ResourcesFragment` listens via `childFragmentManager.setFragmentResultListener("resource_added", …)`. Those only connect if `AddResourceFragment` is shown through `ResourcesFragment`'s **child** FragmentManager (so its `parentFragmentManager` is that same manager). If it's shown from the activity's FM instead, the result never arrives. Please verify the show-site.

**Verdict:** good instinct on the fix; drop the unrelated `gradle.properties` change, make the refresh targeted instead of every-resume, and confirm the FM wiring. Another round. Thanks for tackling this!

---

Okuro3499 (approved):

gradle.properties
```diff
@@ -59,3 +59,6 @@ PLANET_CAMBRIDGE_PIN=6407
 # took egdirbmac cause planet on cambridge is at 8080
 #PLANET_EGDIRBMAC_URL=planet.cambridge.ole.org
 #PLANET_EGDIRBMAC_PIN=1565
+
+# Enabled parallel sync for Gradle 9.4+
+org.gradle.tooling.parallel=true
```
Okuro3499: this file should be on a new pr

Okuro3499 (changes requested):

Okuro3499 (approved):

dogi:
**Re-review — Rating: 68/100 — `changes requested`.**

Thanks for the follow-up. Rechecking my three prior points against the new commits:

1. ✅ **`gradle.properties` change dropped.** The PR is now scoped to the three UI files — good, that resolves the sensitive-file concern.
2. ✅ **FM wiring confirmed.** `AddResourceFragment` is shown via `ResourcesFragment.childFragmentManager` (`ResourcesFragment.kt:347`), so its `parentFragmentManager` is that same manager and `setFragmentResult("resource_added")` → `setFragmentResultListener` connects correctly. The bottom-sheet path is clean, and it's the right mechanism since dismissing a `BottomSheetDialogFragment` doesn't re-fire the host's `onResume`.
3. ❌ **Still open — blanket every-resume refresh, plus now-dead `setResult`.** `onResume()` still calls `refreshResourcesData()` unconditionally (`:588`), and `refreshResourcesData` forces a full rebuild each time (it sets `lastSearchQuery = null`, defeating the no-op guard in `applyFiltersAndUpdateUI`), so returning to the library from *any* screen reloads the whole list. Meanwhile the `setResult(RESULT_OK)` you added in `AddResourceActivity` (`:198`, `:238`) is currently **dead code**: the activity is launched with `startActivity(...)` at both call sites (`ResourcesFragment.kt:239`, `AddResourceFragment.kt:263`) — there's no `registerForActivityResult`/`onActivityResult` consuming it.

Please pick one and make it consistent:
- **Targeted (preferred):** launch `AddResourceActivity` via `registerForActivityResult(StartActivityForResult())` and call `refreshResourcesData()` only on `RESULT_OK`; then revert `onResume()` to *not* blanket-refresh. This uses the `setResult` you already added and matches the intent of the fix.
- **Or**, if you deliberately want the every-resume refresh, remove the two unused `setResult(RESULT_OK)` calls so the code isn't misleading.

Core instinct remains right and the bug (#14603) is functionally fixed; this is about not reintroducing redundant full refreshes and not leaving dead result-plumbing. One more round — thanks!

---

dogi:
**Re-review (2nd round) — Rating: 80/100 → `ready`.**

My prior blocker (flag #2) is properly resolved now — nice work:
- `addResourceLauncher = registerForActivityResult(StartActivityForResult())` refreshes **only** on `RESULT_OK` (`ResourcesFragment.kt`), and `onResume()` no longer blanket-refreshes. That removes the redundant every-resume full reload I was worried about.
- `AddResourceFragment.addResource` now routes through `(parentFragment as? ResourcesFragment)?.addResourceLauncher?.launch(intent) ?: startActivity(intent)`, so the `setResult(RESULT_OK)` you added is actually consumed on the add path — no more dead plumbing. The `?: startActivity` fallback also keeps the other hosts (TeamResources/Personals) working. Registering the launcher as a field initializer is the correct lifecycle spot.
- The `setFragmentResult("resource_added")` path for inline personal-resource creation still refreshes via the child-FM listener. Both add paths covered.

**Remaining nit (non-blocking):** `openEditResource()` still launches `AddResourceActivity` with `startActivity(intent)` rather than `addResourceLauncher.launch(intent)`. Since `AddResourceActivity` already sets `RESULT_OK` on the edit-success branch too, wiring the edit path through the same launcher would make edits refresh the list — otherwise editing a resource has the exact staleness this PR fixes for adds (the old blanket `onResume` refresh used to mask it). One-line change; worth doing here or as a fast follow.

**Verdict:** Targeted, correct refresh triggers — mergeable. Consider routing `openEditResource` through the launcher too.

---

dogi (approved):

PR #14795, opened 2026-07-16
sync: smoother upload managing (fixes #14794)

Okuro3499:
fixes #14794
Refactor team and news uploads to use UploadRepository for network calls and RetryQueue for failed operations instead of calling apiInterface directly. This aligns with the UploadCoordinator pattern and adds retry support for server errors and network exceptions.

## Summary by CodeRabbit

* **New Features**
  * Improved team and news uploads with more reliable processing.
  * Failed uploads can now be retried automatically when errors are temporary.
  * Upload failures preserve the operation details needed for retrying.

* **Bug Fixes**
  * Upload operations now handle server and network errors more consistently.
  * Successful uploads correctly update local records and remove completed items.

PR #14732, opened 2026-07-14
sync: smoother repository submissions uploading (fixes #14816)

dogi:
Refactored various sync and upload actions to hide orchestration details behind clean seams (repositories and coordinators), decoupling UI fragments and activities from direct `UploadManager` or `WorkManager` access.

---
*PR created automatically by Jules for task [3175789715637651086](https://jules.google.com/task/3175789715637651086) started by @dogi*

## Summary by CodeRabbit

* **New Features**
  * Added background synchronization for login and bulk data uploads with clear upload states (idle, loading, success, error).
  * Centralized submissions uploading behind a server availability check, including fallback to an alternative server when needed.
* **Bug Fixes**
  * Improved sync reliability by handling unreachable servers gracefully (skipping uploads when neither endpoint is reachable) and updating stored server preferences when fallback is used.
  * Enhanced user feedback during the sync workflow (progress/dismissal and completion/error handling).

PR #14800, opened 2026-07-16
sync: smoother server reachability caching (fixes #14799)

### LINKED ISSUE #14799: Deduplicate server reachability checks between AutoSyncWorker and ServerReachabilityWorker
Two workers independently ping the server and can overlap, causing duplicate probes and duplicate upload attempts.
- AutoSyncWorker runs on its own timer (hourly): 
    - MainApplication.kt:451
    - syncInterval default 3600s from SharedPrefManager.kt:218
    - calls checkServerAvailability()
- ServerReachabilityWorker runs independently on reconnect events:
    - retries up to 3 times (ServerReachabilityWorker.kt:67-68, 164-171)
    - calls isServerReachable()

Neither path checks whether reachability was just validated by the other worker. If reconnect happens near the hourly tick, both can fire, resulting in:
- duplicate server reachability probes
- duplicate upload attempts

**Expected behavior**
Only one reachability validation/upload trigger should happen within a short dedupe window, even if both workers are triggered.

Okuro3499:
fixes #14799
Add a 30-second TTL cache for isServerReachable results to reduce redundant network checks. Guard submission uploads and retry queue triggers against running syncs. Also migrate withTimeoutOrNull to Duration API and simplify OnSuccessListener to SAM.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved server connectivity checks by briefly reusing recent results, reducing unnecessary connection attempts.
  - Prevented overlapping background sync operations from uploading duplicate or conflicting data.
  - Preserved reliable handling of server timeouts and retry scheduling during background synchronization.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -12,8 +12,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6121
-        versionName = "0.61.21"
+        versionCode = 6135
+        versionName = "0.61.35"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/MainApplication.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/MainApplication.kt
@@ -28,6 +28,7 @@ import java.net.HttpURLConnection
 import java.net.URL
 import java.util.Date
 import java.util.UUID
+import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.TimeUnit
 import java.util.concurrent.atomic.AtomicBoolean
 import javax.inject.Inject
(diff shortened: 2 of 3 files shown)


dogi:
**Rating: 78/100 — `ready`.**

Well-motivated reliability/perf change; verified against source, no correctness bug found.

**Strengths**
- **30s reachability cache:** `isServerReachable` now memoizes per-URL results in a `ConcurrentHashMap` for `REACHABILITY_CACHE_TTL_MS = 30s`, cutting redundant connect attempts when the sync/challenge paths probe the same URL repeatedly. The loop refactor (`break` + single cache write) preserves the original "reachable if any candidate URL responds" semantics.
- **Upload race guard:** `ServerReachabilityWorker.uploadSubmissions` now captures `syncAlreadyRunning = MainApplication.isSyncRunning.get()` and skips the submission upload, exam-result upload, and retry trigger while a sync is running — so the reachability worker no longer uploads concurrently with a full sync (the duplicate/conflicting-upload race in #14799). `HeavyTableSyncWorker.scheduleIfPending` correctly stays outside the guard (it only schedules).
- **Clean modernization:** `withTimeoutOrNull(15000.milliseconds)` is behavior-identical to the old `15000`; the `OnSuccessListener { … }` SAM conversion is tidy (compiles, so it's a `fun interface`); unused `SharedPrefManager` import removed. CI green.

**Notes (non-blocking)**
- The cache also stores **negative** results and isn't invalidated on connectivity change, so a `false` cached during a brief outage persists up to 30s after the network recovers — a sync could be delayed by one TTL. With the short 30s window + `NetworkMonitorWorker` this is minor, but clearing `reachabilityCache` on a connectivity-change broadcast would make recovery snappier.
- The `isSyncRunning` guard is best-effort: a sync can still start in the TOCTOU window between the check and the upload. Fine as a mitigation, just not a lock.

**Rebase note:** touches `MainApplication.isServerReachable` and `ServerReachabilityWorker`, adjacent to the sync-orchestration PRs (#14732) and #14795 — coordinate rebases.

**Verdict:** solid, mergeable. Thanks!

---

dogi (approved):

PR #14803, opened 2026-07-16
sync: smoother network module requesting (fixes #14802)

### LINKED ISSUE #14802: RetryInterceptor backoff sleep runs synchronously inside intercept()
Because `intercept()` runs on the same thread that issued the request, repeated backoff delays can accumulate directly on that thread instead of being scheduled asynchronously. If the caller-sensitive thread is the main thread, this can contribute to UI freezes or ANR behavior.

Okuro3499:
fixes #14802
Configures a custom Dispatcher on the OkHttpClient to allow up to 20 concurrent requests per host, improving parallel download/upload throughput.

## Summary by CodeRabbit

  * Improved network request handling by controlling the number of simultaneous connections to each host.
  * Helps provide more consistent app performance during multiple concurrent network requests.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -12,8 +12,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6121
-        versionName = "0.61.21"
+        versionCode = 6136
+        versionName = "0.61.36"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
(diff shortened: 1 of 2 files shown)


dogi:
**Rating: 82/100 — `ready`.**

Focused, low-risk throughput tuning. Sets a custom OkHttp `Dispatcher` with `maxRequestsPerHost = 20` (OkHttp's default is **5**) in the shared `buildOkHttpClient`, so parallel sync/downloads to the CouchDB host are no longer throttled at 5 concurrent. The global `maxRequests` (default 64) still comfortably exceeds 20, so nothing else needs adjusting. CI green on both flavors.

**Notes (non-blocking)**
- Each `buildOkHttpClient(...)` call now constructs its own `Dispatcher` (and thus its own `ExecutorService`); fine for the handful of clients built, just be aware the pools aren't shared across clients.
- Judgment call on the value: 20 concurrent requests to a single satellite CouchDB (often modest hardware) — or over a weak mobile link — could add server load / contention. If constrained deployments start seeing timeouts, consider dialing it to ~8–10. Not a blocker.

**Verdict:** correct and self-contained; mergeable. Thanks!

---

dogi (approved):

PR #14828, opened 2026-07-17
resources: smoother searching (fixes #14818)

### LINKED ISSUE #14818: library search broken
https://github.com/user-attachments/assets/82cd5333-3915-4214-906b-8ac1ad8ba590

ragilzakaria:
[Screen_recording_20260717_192219.webm](https://github.com/user-attachments/assets/c6baa212-b1e5-4a01-a732-25603dd96efb)

fixes #14818 

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated text change notifications to consistently return text values as Kotlin `String`, improving reliability for downstream listeners.
  * Ensured null-safe behavior by normalizing emitted values instead of passing raw `CharSequence` references.

* **Chores**
  * Bumped the app version metadata (version code/name) for this release.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -12,8 +12,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6141
-        versionName = "0.61.41"
+        versionCode = 6144
+        versionName = "0.61.44"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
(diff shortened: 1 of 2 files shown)


Valerie923 (approved):

Okuro3499 (approved):

dogi:
**Review — Rating: 90/100**

Verified against `ViewExtensions.kt` and the consumer `ResourcesFragment.kt:271–279`.

**Strengths**
- Correct root-cause fix, and at the right layer. The library search pipeline is `textChanges().debounce(300L).distinctUntilChanged()`. The old `trySend(s)` emitted the EditText's live `Editable` — a single mutable instance reused across keystrokes — so by the time `distinctUntilChanged` compared "previous" vs "current" they were the *same, already-mutated* reference: equal → emission dropped → search stopped updating after the first change (#14818). Snapshotting to `s?.toString()` / `text?.toString()` emits an immutable value per change, so `distinctUntilChanged` compares actual content.
- Fixing it in the shared extension repairs the same latent bug for every other consumer (`TeamFragment`, `SubmissionsFragment`, `VoicesFragment`, etc.), not just the library.
- Return type stays `Flow<CharSequence?>` and all downstream `.toString()` calls are unaffected — no regression.

**Concerns**
- None blocking. Minor stylistic thought: since every emission is now effectively a `String`, the declared type could be narrowed to `Flow<String?>` for clarity — optional, not required.

**Verdict:** Small, precise, correct fix — mergeable.

---

PR #14836, opened 2026-07-17
resources: less webview progress bar increment is more (fixes #14830)

### LINKED ISSUE #14830: WebView progress bar double-applies progress
In `WebViewActivity.kt`, the `onProgressChanged` override inside the `WebChromeClient` both **sets** the progress bar to `newProgress` (absolute assignment) and then **increments** it by `newProgress` again (relative addition) within the same callback invocation. This means every update effectively applies the value twice, causing the progress bar to visually reach 100% when the actual page load is only around 50% complete.
https://github.com/user-attachments/assets/78b4a1a4-3675-4167-86c1-7784c99c07c4

Okuro3499:
fixes #14830
The incrementProgressBy call was unnecessary since the progress bar is only used as a loading indicator that gets hidden at 100%.
[2.webm](https://github.com/user-attachments/assets/c27f1e7d-377c-4a78-a575-b05a7161c1b3)

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the Android WebView loading progress indicator to show progress more accurately.
  * Automatically closes the web viewer when navigation reaches an unsupported English-language `/eng/` page.
* **Chores**
  * Bumped the app version metadata for this release.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -12,8 +12,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6141
-        versionName = "0.61.41"
+        versionCode = 6143
+        versionName = "0.61.43"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/ui/viewer/WebViewActivity.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/viewer/WebViewActivity.kt
@@ -280,7 +280,6 @@ class WebViewActivity : AppCompatActivity() {
                 if (view.url?.startsWith("file://") == false && view.url?.endsWith("/eng/") == true) {
                     finish()
                 }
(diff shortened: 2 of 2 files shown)


dogi:
**Review — Rating: 87/100**

Verified against `WebViewActivity.kt:277–287`.

**Strengths**
- Correct removal. `onProgressChanged` already sets the absolute value on the line above (`pBar.progress = newProgress`, `:279`), so the removed `pBar.incrementProgressBy(newProgress)` (`:283`) was adding `newProgress` *on top of* the just-set absolute value — overshooting toward 2× each callback. Dropping it leaves accurate progress plus the existing hide-at-100 logic intact. This is slightly more than "redundant" — it was actively wrong — but the fix and the net effect are right.

**Concerns**
- None. Note the CodeRabbit summary's "auto-closes on unsupported English URL" line refers to the pre-existing `finish()` block (`:280–282`), which this diff doesn't touch — the actual change is the single-line removal only.

**Verdict:** Correct one-line cleanup — mergeable.

---

dogi (approved):

PR #14839, opened 2026-07-17
courses: smoother exams taking (fixes #14829)

### LINKED ISSUE #14829: Unscoped fallback query in `updateAnsDb()` can write an answer onto an unrelated pending submission
In `ExamTakingFragment.kt`, the `updateAnsDb()` function contains a fallback that fires when `sub` (the current submission reference) is `null`:

```kotlin
// ExamTakingFragment.kt — lines 673–675
if (sub == null) {
    sub = submissionsRepository.getLastPendingSubmission(user?.id)
}
```

`getLastPendingSubmission` is implemented in `SubmissionsRepositoryImpl.kt` (lines 634–642) as:

```kotlin
// SubmissionsRepositoryImpl.kt — lines 634–642
override suspend fun getLastPendingSubmission(userId: String?): RealmSubmission? {
    return withRealm { realm ->
        realm.where(RealmSubmission::class.java)
            .equalTo("status", "pending")
            .equalTo("userId", userId)
            .sort("startTime", Sort.DESCENDING)
            .findFirst()?.let { realm.copyFromRealm(it) }
    }
}
```

The query is filtered only by `userId` and `status = "pending"`. It is **not** scoped to the current exam's `parentId`. As a result, if a user has any other pending submission in the database (e.g., a previously started survey or a different exam that was not cleaned up), `updateAnsDb()` will silently attach the answer for the current question to **that unrelated submission** instead of failing safely or creating a new one.

Okuro3499:
fixes #14829

updateAnsDb() no longer falls back to the globally-unscoped getLastPendingSubmission(userId); instead it recomputes the same parentId used at exam-init time (via the new computeParentId() helper) and looks up the pending submission scoped to that exam/course via getSubmissionsByParentId. This closes the window where a stray Next/Submit tap during the async submission-recreation in initializeExamData() could write the current answer onto an unrelated pending exam or survey. Since getLastPendingSubmission was only used for this buggy fallback, I removed it from the repository interface and implementation rather than leaving dead code.

## Summary by CodeRabbit

- **Exam Experience**
  - Improved loading and continuation of pending exams by more accurately matching submissions to the relevant course and exam.
  - Enhanced exam submission workflows to support saving answers, creating exam submissions, and attaching submission photos.
- **Bug Fixes**
  - Reduced the risk of resuming the wrong pending submission when taking exams across different courses.


dogi:
**Review — Rating: 91/100**

Verified against `ExamTakingFragment.kt`, `SubmissionsRepository[Impl].kt`.

**Strengths**
- Real fix, correctly targeted. The old `getLastPendingSubmission(userId)` (`SubmissionsRepositoryImpl.kt:634`) filtered only by `userId`+`status="pending"` and returned the most recent — so the `updateAnsDb()` fallback could adopt a pending submission from an *unrelated* exam/course and write the current answer onto it. The replacement reuses `getSubmissionsByParentId(parentId, userId, "pending")`, which applies the same `sort("startTime", DESCENDING)` but additionally scopes to `parentId`. `.firstOrNull()` therefore preserves "most recent pending" semantics while eliminating the cross-exam leak (#14829).
- `computeParentId()` reproduces the exact `parentId` used at init (`ExamTakingFragment.kt:95–99`) and de-duplicates it across both call sites — the fallback and init now agree by construction.
- Removing the method from the interface + impl rather than leaving it dead is the right call; confirmed via grep that no other caller (including tests) referenced it, and `Sort` stays imported for the remaining queries.

**Concerns**
- None blocking. Minor: `updateAnsDb`'s fallback should rarely fire (init sets `sub`), but when it does the new scoping is strictly safer.

**Verdict:** Correct scoping fix with a clean refactor — mergeable.

---

dogi:
**Review — rating: 86/100**

Good cleanup that also tightens correctness.

**Strengths**
- `getLastPendingSubmission` had a single caller (`ExamTakingFragment:674`); I confirmed no others, so dropping it from the interface + impl is compile-safe.
- The replacement is arguably *more correct*: the old `getLastPendingSubmission(userId)` → `getLatestPendingByUser` returned the latest pending submission across **any** exam, which could pick up a different exam's pending submission. `getSubmissionsByParentId(computeParentId(), userId, "pending").firstOrNull()` scopes to the current exam and matches what `initializeExamData` already does — so the two paths are now consistent.
- `computeParentId()` sensibly DRYs the `"$id@courseId"` / `id` logic used in both spots.

**Concerns**
- Minor: `submissionDao.getLatestPendingByUser` may now be unused — check and remove it if so (dead DAO query).
- Minor: `getSubmissionsByParentId(...).firstOrNull()` relies on there being one pending submission per exam/user (true in practice); if multiple could exist, "first" isn't guaranteed "latest," but that's an unlikely edge.

**Verdict:** Correct, de-duplicating refactor with a small correctness win — mergeable.

---

dogi (approved):

PR #14842, opened 2026-07-17
resources: smoother video viewing (fixes #14840)

### LINKED ISSUE #14840: exoplayer leaking
streamed video restarts from 0:00 and leaks an ExoPlayer every 15 minutes (auth-refresh loop).

Okuro3499:
fixes #14840
Instead of recreating the player when the auth session updates, update the streaming data source's request properties in place. This avoids interrupting video playback during token refresh.

## Summary by CodeRabbit

- **Bug Fixes**
  - Updated online video streaming to correctly apply refreshed authentication cookies.
  - Existing playback now reuses the current stream and updates request headers instead of restarting.
  - Improved cleanup when exiting the video viewer to avoid retaining stale streaming configuration.
- **Chores**
  - Bumped app version code and version name for this release.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -12,8 +12,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6141
-        versionName = "0.61.41"
+        versionCode = 6146
+        versionName = "0.61.46"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/ui/viewer/ResourceViewerFragment.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/viewer/ResourceViewerFragment.kt
@@ -89,6 +89,7 @@ class ResourceViewerFragment : Fragment(), AuthSessionUpdater.AuthCallback {
     private var auth: String = ""
 
     private var exoPlayer: ExoPlayer? = null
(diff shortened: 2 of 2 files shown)


dogi:
**Review — Rating: 88/100**

Verified against `ResourceViewerFragment.kt`.

**Strengths**
- Correctly diagnoses and fixes a real leak. `setAuthSession()` is the `AuthCallback` invoked on every session/token refresh, and it previously called `streamVideoFromUrl()` unconditionally, which reassigns `exoPlayer = createExoPlayer()` (`:305`) without releasing the prior instance. Each refresh therefore orphaned a live `ExoPlayer` (and could leave overlapping playback). `streamVideoFromUrl` is only reachable from this callback, so guarding on `exoPlayer == null` and otherwise mutating the factory's request properties in place is the right shape.
- `DefaultHttpDataSource.Factory.setDefaultRequestProperties(...)` mutates the shared `RequestProperties` the factory hands to every `DefaultHttpDataSource` it creates, so refreshed cookies propagate to subsequent connections. Clearing the field in `onDestroyView` keeps lifecycle tidy.

**Concerns**
- Non-blocking: the refreshed cookie only takes effect on the *next* HTTP connection the factory opens (reconnect/seek/next range). A single long-lived progressive connection mid-play keeps the old cookie until it reconnects. Acceptable for token refresh and strictly better than the old recreate-and-leak behavior, but worth being aware of if a stream is ever rejected mid-flight after expiry.

**Verdict:** Correct, well-scoped leak fix — mergeable.

---

dogi (approved):

PR #14843, opened 2026-07-17
sync: smoother download queue starting (fixes #14841)

### LINKED ISSUE #14841: Race on isQueueRunning lets two concurrent downloads share one buffer and corrupt files
A race condition on `isQueueRunning` allows two concurrent downloads to run at the same time. When this happens, both downloads can end up sharing a single buffer, which corrupts downloaded files.

## Problem
`isQueueRunning` is not guarded strongly enough to prevent overlapping queue execution under concurrent triggers.

Okuro3499:
fixes #14841
Set isQueueRunning flag before launching the coroutine to prevent multiple concurrent queue processors from being started.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved download queue startup reliability by marking the download queue as running immediately when the download service starts, before processing begins.
* **Chores**
  * Updated the app version (version code/name) for the next release.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -12,8 +12,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6141
-        versionName = "0.61.41"
+        versionCode = 6142
+        versionName = "0.61.42"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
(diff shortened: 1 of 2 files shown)


dogi:
**Review — Rating: 91/100**

Verified the diff against `DownloadService.kt` (not just the description).

**Strengths**
- Correct root-cause fix. `appScope` is built on `dispatcherProvider.io` (`ServiceModule.kt:67`), so the coroutine body runs on a background thread while `onStartCommand` runs on the main thread. In the old code `isQueueRunning = true` executed asynchronously *after* `onStartCommand` returned, so two rapid `startService` calls both saw `isQueueRunning == false`, and the second call's `currentJob?.cancel()` tore down the first queue processor and relaunched it — the race described in #14841. Setting the flag synchronously before `launch` (`DownloadService.kt:114`) closes that window.
- Minimal and well-scoped; the `finally` block still resets the flag, so the lifecycle stays balanced.

**Concerns**
- None blocking. Minor: the `else` branch is now the only way a concurrent call is short-circuited, which is exactly the intended behavior; no dead-flag path remains because a running job always implies `isQueueRunning == true`.

**Verdict:** Correct, minimal race fix — mergeable.

---

dogi (approved):

PR #14852, opened 2026-07-18
all: smoother room database configuring (fixes #14888)

dogi:
## Summary
This PR completes the migration from Realm to Android Room as the primary database layer for the myPlanet application. All data models have been converted to Room entities, repository implementations updated to use Room DAOs, and Realm-specific code removed.

### Database Layer Migration
- **New Room Database**: Added `AppDatabase.kt` with Room configuration and all entity definitions
- **Type Converters**: Created `Converters.kt` for custom type serialization (JsonArray, JsonObject, etc.)
- **DAO Layer**: Created 50+ Data Access Objects in `data/room/dao/` for all entities:
  - Core entities: `UserDao`, `MyLibraryDao`, `CourseDao`, `CourseStepDao`, `ExamDao`, `QuestionDao`, `AnswerDao`, `SubmissionDao`
  - Team management: `TeamDao`, `TeamTaskDao`, `TeamLogDao`, `TeamNotificationDao`
  - User activities: `AchievementDao`, `HealthExaminationDao`, `OfflineActivityDao`, `ResourceActivityDao`, `SearchActivityDao`
  - Content: `NewsDao`, `TagDao`, `CertificationDao`, `CourseProgressDao`, `RatingDao`, `FeedbackDao`, `MeetupDao`, `PersonalDao`, `MyLifeDao`
  - Utilities: `NotificationDao`, `RetryDao`, `RemovedLogDao`, `ApkLogDao`, `ChatDao`, `CommunityDao`, `DictionaryDao`, `SubmitPhotosDao`, `UserChallengeActionsDao`
- **Dependency Injection**: Added `RoomModule.kt` providing database and DAO instances via Hilt

### Model Updates
All Realm models renamed and converted to Room entities with `@Entity`, `@PrimaryKey`, `@ColumnInfo`, and `@Index` annotations:
- `RealmMyLibrary` → `MyLibrary`
- `RealmMyTeam` → `MyTeam`
- `RealmUser` → `UserEntity`
- `RealmTeamTask` → `TeamTask`
- `RealmTeamLog` → `TeamLog`
- `RealmAchievement` → `Achievement`
- `RealmHealthExamination` → `HealthExamination`
- `RealmExamQuestion` → `ExamQuestion`
- `RealmCourseProgress` → `CourseProgress`
- `RealmSubmission` → `Submission`
- `RealmAnswer` → `Answer`
- `RealmTag` → `TagEntity`
- `RealmRating` → `Rating`
[...]

PR #14853, opened 2026-07-20
all: smoother text view extensions utils coloring (fixes #14831)

### LINKED ISSUE #14831: course: reconsider red font color for "show more" text
Image

it is better to be other color than red (e.g., blue)

ragilzakaria:
1. No color was defined in the code before: Inside TextViewExtensions.kt, the code only created the clickable "show more" link, but it never specified a text color for it.
2. Because no color was set explicitly, the text implicitly inherited the app's default link/accent color (colorAccent), which is defined globally as red in the theme settings.
3. Overriding the default: To change this behavior specifically for expandable text views, we had to introduce a new line: setLinkTextColor(androidx.core.content.ContextCompat.getColor(context, R.color.mainColor))                                                                                                                                       
This explicitly tells the TextView to override the inherited red color and render its links in blue.

[Screen_recording_20260720_181249.webm](https://github.com/user-attachments/assets/6a2bd39b-612e-4a6b-89e1-564b3a7675c8)

fixes #14831 

## Summary by CodeRabbit

* **New Features**
  * Expandable text links now use the app’s main color for improved visibility and consistency.

* **Chores**
  * Updated the app version to 0.61.50.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -12,8 +12,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6141
-        versionName = "0.61.41"
+        versionCode = 6150
+        versionName = "0.61.50"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/utils/TextViewExtensions.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/utils/TextViewExtensions.kt
@@ -16,6 +16,7 @@ fun TextView.makeExpandable(
     collapseLabel: String = context.getString(R.string.show_less)
 ) {
     var isExpanded = false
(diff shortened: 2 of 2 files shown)


dogi:
**Review — Rating: 84/100 — `ready`.**

Correct and minimal. `makeExpandable` never set a link color, so the show-more/less span inherited the theme's red `colorAccent`; `setLinkTextColor(ContextCompat.getColor(context, R.color.mainColor))` overrides it with the brand color (`mainColor → @color/colorPrimary`), resolving #14831. Applies to the clickable span this function installs, so scope is right.

Minor nits (non-blocking): `mainColor` is a single (non-daynight) color — fine as long as `colorPrimary` reads acceptably on both themes' backgrounds; and the inline fully-qualified `androidx.core.content.ContextCompat` could be a top-of-file import to match style. Clean vs the #14852 migration.

**Verdict:** mergeable.

---

Okuro3499 (approved):

dogi (approved):

PR #14862, opened 2026-07-20
chat: smoother history teams sharing (fixes #14854)

### LINKED ISSUE #14854: Improve share-to-team messaging for users with no team membership
## Summary
Users who try to share a chat to a team while not being a member of any team do not get a clear enough message explaining why the action cannot proceed.

## Problem
When a member without any team membership attempts to share a chat to a team, the current messaging does not clearly explain that they must first join or be added to a team before sharing.

https://github.com/user-attachments/assets/d1bfac0b-c034-43db-85c3-95aae83f3e24

Valerie923:
Screenshot_20260720_234043_org_ole_planet_myplanet_DashboardActivity

fixes #14854 

## Summary by CodeRabbit

- **Bug Fixes**
  - When users try to share a chat with no eligible options, the app now shows a clear OK-only alert indicating they must join a team or enterprise first.
  - Prevented the chat-sharing selection UI from appearing when there are no items to choose from.

- **Localization**
  - Added new “join team first” and “join enterprise first” sharing prompts across supported languages: English, Arabic, Spanish, French, Nepali, and Somali.


dogi:
**Review — Rating: 87/100 — `ready`.**

Correct empty-state guard. When `showGrandChildRecyclerView` is handed an empty `items` list it now shows an "join a Team/Enterprise first" dialog and returns, instead of opening an empty selection sheet (#14854). Verified:
- `AlertDialog` is already imported (`ChatHistoryAdapter.kt:3`), so it compiles.
- The `section == context.getString(R.string.teams)` branch reuses the exact comparison the same class already uses (`:194`), and callers pass `section` as the localized Teams/Enterprises label (`:157/:159`) — so the team-vs-enterprise message choice is consistent with existing logic.
- New `join_team_first` / `join_enterprise_first` strings added across all six locales plus default — complete i18n.

**Cross-PR note:** touches `ChatHistoryAdapter.kt`, which is also in the #14852 migration set — it'll need to rebase relative to #14852.

**Verdict:** correct and complete — mergeable.

---

ragilzakaria (approved):
This PR improves the chat sharing experience by adding a clear dialog prompt when a user attempts to share a chat but has no available teams or enterprises. It successfully prevents empty selection screens from appearing and adds required translations for multiple languages, directly addressing the intent described in the PR. The code is straightforward and properly implements the desired behavior without violating repository guidelines. 

Critical Issues / Bugs
None detected.

Optimizations & Best Practices
None detected.

❓ Clarifications / Questions
None.

Okuro3499 (changes requested):
can we make the dialog darkmode compatible
Image

Valerie923:
Screenshot_20260721_215059

Okuro3499 (approved):

dogi:
## Review — rating: 86/100

Traced the call sites and existing patterns in `ChatHistoryAdapter`, not just the description.

**Strengths**
- Correct, minimal fix for the real problem. Previously `showGrandChildRecyclerView` would pop an empty selection dialog when the user had no teams/enterprises; the early-return with a clear message (`ChatHistoryAdapter.kt:190`) is exactly right and matches the issue.
- The `section == getString(R.string.teams)` branch is consistent with the file's existing convention — `section` is passed as the localized string at both call sites (`:157`, `:159`) and the same comparison is already used at `:211`, so this doesn't introduce a new fragility, it follows the established one.
- Dialog styling reuses the class's existing idiom (`daynightGreyColor.toDrawable()`, `CustomAlertDialog`) seen at `:178/:203/:231`, so it looks native to the screen.
- Full localization added (ar/es/fr/ne/so) rather than English-only. CI green.

**Concerns (non-blocking)**
- Consider CodeRabbit's wording nit — membership can come from being *added/invited*, not only "joining"; a slightly softer string ("You need to be a member of a Team to share this chat") would read better. Optional.
- Pre-existing smell (not introduced here): branching on a localized `getString` comparison is brittle if those strings ever collide. A `sealed`/enum section key would be more robust — future cleanup, not this PR.
- Commit messages `ok` / `updated` — cosmetic since the repo squash-merges.

**Verdict:** Clean, well-localized UX fix that follows existing conventions — mergeable.

---

dogi:
**Rating: 87/100**

**Strengths**
- Good UX fix: instead of opening an empty share dialog when the user belongs to no Team/Enterprise, `showGrandChildRecyclerView` now short-circuits on `items.isEmpty()` and shows an informative, localized dialog. The early `return` cleanly avoids inflating the empty list.
- Implementation is consistent with the surrounding code — `daynightGreyColor` (`:59`) and `toDrawable` (imported `:9`) are exactly what the normal dialog path uses (`:214`), and the `section == getString(R.string.teams)` check mirrors the existing branch at `:194`. Since callers always pass `section` as `getString(R.string.teams/enterprises)` (`:157`/`:159`), the comparison is reliable.
- Solid i18n: `join_team_first` / `join_enterprise_first` added to **all six** locales (en, ar, es, fr, ne, so) — the right way to do it.

**Concerns**
- Minor/structural: keying UI logic off a localized display string (`section == getString(...)`) is inherently fragile, but it's a pre-existing pattern in this file rather than something this PR introduces — worth a future refactor to an enum/constant, not a blocker here.

**Verdict:** Clean, well-localized empty-state fix consistent with existing code — ready.

---

dogi (approved):

PR #14865, opened 2026-07-20
courses: smoother take view modelling (fixes #14864)

### LINKED ISSUE #14864: Use a ViewModel in TakeCourseFragment to avoid repeated reloads and join dialogs
## Summary
`TakeCourseFragment` re-queries course, steps, and progress on every view recreation and re-shows the join dialog for unjoined courses each time `setCourseData()` runs. This causes repeated prompts and visible reloads after configuration changes.

## Problem
In `ui/courses/TakeCourseFragment.kt`:
- `onViewCreated` (around lines 80-144) reloads course, step, and progress data from the repository on every view creation instead of retaining UI state across configuration changes.
- `setCourseData()` (around lines 207-227) unconditionally shows the **"join this course"** `AlertDialog` whenever it runs for an unjoined course.

## When it breaks
A learner opens an unjoined course's steps screen, dismisses the join prompt, and then rotates the device or triggers another configuration change. The fragment is recreated, the repository queries run again, and the join dialog is shown again in the same session.

Okuro3499:
fixes #14864
Move course loading, progress fetching, and UI state management into a dedicated TakeCourseViewModel with StateFlow-based TakeCourseUiState. This eliminates AtomicBoolean progress-fetching logic, deduplicates reload calls, and survives configuration changes. Includes unit tests for the new ViewModel.

## Summary by CodeRabbit

- **Improvements**
  - Refactored the course screen to drive UI from a dedicated state model with clear Loading/NotFound/Success handling.
  - Course steps and progress now load together for more consistent display.
  - Join prompt logic now appears only when appropriate and won’t repeat unnecessarily during a course visit.
  - Joining or leaving a course automatically refreshes the course details and progress.
- **Tests**
  - Added unit tests covering state transitions, caching vs forced refresh, and join-prompt behavior.


dogi:
**Review — Rating: 88/100 — `ready`.**

Verified the diff against real source (`CoursesRepository`, `FlowExtensions`, `MainDispatcherRule`), not just the description.

**Strengths**
- Clean, correct extraction. `TakeCourseViewModel` owns load/user/steps/progress aggregation behind a `StateFlow<TakeCourseUiState>` (Loading/NotFound/Success), and the fragment renders via `collectLatestWhenStarted` + `bindCourse`. All referenced symbols line up: `getCourseSteps(courseId): List<RealmCourseStep>` matches `Success.steps`, `getCourseProgress(userId, courseIds): HashMap<String?, JsonObject>` matches the VM call, and `collectLatestWhenStarted` is a real `Fragment` extension (`FlowExtensions.kt:21`).
- Genuinely removes the fragile bits the issue called out: the `AtomicBoolean isFetchingProgress` / `cachedCourseProgress` dance is gone, replaced by `loadedCourseId`-based dedup with an explicit `forceRefresh` for post-join/leave reloads. `hasOfferedJoinDialog` moves the "offer once" state into the ViewModel so it survives config changes and rebinds.
- `bindCourse` guards `_binding == null`, and re-registers the ViewPager2 callback only when the adapter is first created — so the forced-refresh rebind after join/leave doesn't stack callbacks. `setCourseData()` still runs (now via the Success state), so nothing is dropped.
- Solid unit tests: cache hit (no re-query on same id), `forceRefresh` re-queries, different id re-queries, NotFound, and join-dialog-once. `MainDispatcherRule(StandardTestDispatcher())` + `advanceUntilIdle()` drives `viewModelScope` correctly.

**Minor (non-blocking)**
- Behavior change for a null `courseId`: `courseId?.let { viewModel.loadCourse(it) }` means a null id leaves `uiState` stuck on `Loading` (infinite spinner), whereas the old path fell through to the "failed to load" toast + `popBackStack`. Edge case (the fragment is always built with an id), but a `?: viewModel.markNotFound()`-style fallback would preserve the old error handling.
- qlty flags `stubCourseLoad` (8 params) — it's a defaulted test helper, cosmetic; a small params object would clear it if you care about the badge.

**Verdict:** Well-scoped, correct, and tested refactor that matches #14864 — mergeable.

---

dogi:
**Rating: 90/100**

**Strengths**
- Clean, correct MVVM extraction. `TakeCourseViewModel` owns loading/aggregation and exposes a `StateFlow<TakeCourseUiState>` (Loading/NotFound/Success), replacing the fragment's `AtomicBoolean isFetchingProgress` + `cachedCourseProgress` juggling. Compiles cleanly — verified `collectLatestWhenStarted` (FlowExtensions.kt:21), `getCourseSteps(): List<CourseStep>`, and the `getCourseProgress(userId, courseIds)` overload all match usage; the progress-parsing (`["current"].asInt ?: 0`) is preserved verbatim.
- Actually fixes #14864: `hasOfferedJoinDialog` on the retained ViewModel means the join dialog is offered once and survives rotation, and `loadCourse`'s `loadedCourseId`/Success cache dedupes reloads. Join/leave uses `loadCourse(cId, forceRefresh = true)` to re-drive state instead of a manual `getCourseById` patch.
- Good defensive details: `bindCourse` guards `if (_binding == null) return`, and the `registerOnPageChangeCallback` is now correctly moved *inside* the `coursesPagerAdapter == null` block — necessary since `bindCourse` can run on every emission, so this avoids duplicate callback registration.
- Solid tests covering Success aggregation, NotFound, cache-hit (no re-query), forceRefresh re-query, different-courseId re-query, and the once-only dialog flag.

**Concerns**
- Minor UX nuance: `hasOfferedJoinDialog` persists for the ViewModel's lifetime, so if a user joins then leaves within the same fragment instance, the join dialog won't be re-offered even though `containsUserId` is false again. That matches the "no repeated dialogs" intent of #14864, but worth confirming it's the desired behavior for the join→leave→rejoin path.
- The ViewPager2 callback is registered but never unregistered (pre-existing; not introduced here).

**Verdict:** Well-structured, well-tested refactor that fixes the reload/dialog bug and survives config changes — ready.

---

dogi (approved):

PR #14867, opened 2026-07-20
chat: smoother primary url checking (fixes #14866)

### LINKED ISSUE #14866: ChatDetailFragment clears global alternate-URL routing flags in onDestroyView(), breaking app-wide in-flight network operations
### Summary
`ui/chat/ChatDetailFragment.kt` currently clears global `SharedPrefManager` flags (`isAlternativeUrl` / `processedAlternativeUrl`) in `onDestroyView()` (around lines 701–705).

This happens whenever the AI chat screen is destroyed (including normal back navigation and configuration changes like rotation), not just when alternate URL routing should be globally reset.

### Problem
These same global flags are consumed app-wide via `UrlUtils.getUrl()` and used by components such as:
- `SyncManager`
- `TeamsRepositoryImpl`
- `DownloadService` (e.g. `DownloadService.kt:245`)

They control failover routing to a reachable mirror when the primary CouchDB endpoint is unavailable.

When chat closes, `onDestroyView()` resets these global flags unconditionally. If background sync/download is currently relying on the mirror, routing can flip back to the unreachable primary URL mid-operation.

Okuro3499:
fixes #14866
Instead of unconditionally clearing the alternative URL on fragment destroy, only clear it if the primary server is reachable. This prevents losing the alternative URL when the primary is still down.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved server URL handling when leaving a chat screen by deferring alternative URL cleanup until the primary server is confirmed reachable.
  * Added validation to prevent connectivity checks for blank/invalid URLs.
* **Tests**
  * Added coroutine unit tests covering primary server reachability validation and ensuring invalid inputs don’t trigger alternative URL mapping.
* **Chores**
  * Updated the app release version metadata to 0.61.54 (code 6154).

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -12,8 +12,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6141
-        versionName = "0.61.41"
+        versionCode = 6154
+        versionName = "0.61.54"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/MainApplication.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/MainApplication.kt
@@ -214,6 +214,14 @@ class MainApplication : Application(), WorkManagerConfiguration.Provider {
             return reachable
         }
 
+        suspend fun isPrimaryServerReachable(
+            urlString: String,
+            ioDispatcher: CoroutineDispatcher = coreDependenciesEntryPoint.dispatcherProvider().io
+        ): Boolean {
+            if (urlString.isBlank()) return false
+            return tryConnect(urlString, ioDispatcher)
+        }
+
         private suspend fun tryConnect(
             urlString: String,
             ioDispatcher: CoroutineDispatcher
--- a/app/src/main/java/org/ole/planet/myplanet/ui/chat/ChatDetailFragment.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/chat/ChatDetailFragment.kt
@@ -34,6 +34,7 @@ import kotlinx.coroutines.Job
 import kotlinx.coroutines.isActive
 import kotlinx.coroutines.launch
 import org.ole.planet.myplanet.MainApplication
+import org.ole.planet.myplanet.MainApplication.Companion.isPrimaryServerReachable
 import org.ole.planet.myplanet.MainApplication.Companion.isServerReachable
 import org.ole.planet.myplanet.R
 import org.ole.planet.myplanet.databinding.FragmentChatDetailBinding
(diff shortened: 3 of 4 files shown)


dogi:
**Review — Rating: 45/100 — `changes requested`.**

The intent is right (don't drop the alternative URL while the primary is still down), and moving the clear off the main thread and onto `applicationScope` so it survives teardown is a good call. But the guard doesn't actually check what it claims to, so the fix is effectively a no-op for the case it targets.

**Core problem — `isServerReachable` is not a primary-only check.** `clearAlternativeUrlIfPrimaryRestored()` gates on `isServerReachable(serverUrl)`, where `serverUrl = sharedPrefManager.getServerUrl()` (the primary). But `isServerReachable` (`MainApplication.kt:201–214`) does:
```kotlin
val mapping = serverUrlMapper.processUrl(urlString)
val urlsToTry = mutableListOf(urlString)
mapping.alternativeUrl?.let { urlsToTry.add(it) }   // also tries the alternative
for (url in urlsToTry) { if (tryConnect(url, …)) { reachable = true; break } }
```
So it returns `true` if the primary **or its alternative** is reachable. In the exact scenario #14866 is about — primary down, app currently routing through the alternative — the alternative is by definition reachable, so `isServerReachable(primary)` returns `true` and the code clears the alternative URL anyway. Since the top-of-method guard (`isAlternativeUrl()`) only passes when an alternative is in use (and therefore up), this is essentially equivalent to the old unconditional clear; the only case it changes is "both primary and alternative down," where nothing is reachable regardless.

**What's needed:** check reachability of the **primary URL alone**, without the alternative fallback — e.g. a primary-only helper (a direct `tryConnect(primaryUrl)` equivalent) — and clear only when that succeeds. As written, `isServerReachable` can't distinguish "primary is back" from "we're still on the alternative."

**Minor:** the `applicationScope.launch { … }` lambda captures `sharedPrefManager` (a fragment field), so it retains the fragment for the duration of the network check after `onDestroyView`. Capturing the singleton into a local first avoids that transient hold.

**Verdict:** Correct instinct, but the reachability call includes the alternative, so it doesn't fix #14866. Needs a primary-only reachability check. Another round.

---

dogi:
**Re-review — Rating: 88/100 — `ready`.** (was 45, `changes requested`)

My blocker is genuinely resolved — not just "changed". The new `isPrimaryServerReachable` (`MainApplication.kt`) calls `tryConnect(urlString)` **directly**, with no `processUrl`/alternative fallback and no shared reachability cache:
```kotlin
if (urlString.isBlank()) return false
return tryConnect(urlString, ioDispatcher)
```
and `clearAlternativeUrlIfPrimaryRestored()` now uses it. So in the exact scenario I flagged — primary down, app running on the alternative — this returns `false` and the alternative URL is **kept**, which is the whole point of #14866. Correct semantics in both directions now (primary back → clear; primary still down → preserve).

Nice touches:
- The test `isPrimaryServerReachable never consults the alternative URL mapping` (`verify(exactly = 0) { mockServerUrlMapper.processUrl(any()) }`) pins precisely the behavior that was broken before — good regression guard.
- My earlier minor leak nit is handled too: `primaryUrl` and `prefManager` are captured as locals before the `applicationScope.launch`, so the coroutine no longer retains the fragment.

**Verdict:** correct, minimal, well-tested primary-only reachability check — mergeable. Lifting `changes requested` → `ready`.

---

dogi (approved):

PR #14869, opened 2026-07-20
resources: use StateFlow for download progress broadcasts (fixes #14868)

### LINKED ISSUE #14868: Fix missed download completion broadcasts across fragment recreation
BroadcastService currently emits download-progress and completion `Intent`s through a single `MutableSharedFlow<Intent>(extraBufferCapacity = 64)` with the default `replay = 0`. In `ResourcesFragment` and `BaseResourceFragment`, collectors are attached with `repeatOnLifecycle(STARTED)`. If a fragment view is torn down at the exact moment a `completeAll` progress intent is emitted, the next collector can attach too late and permanently miss that terminal event.

Okuro3499:
fixes #14868
Replace SharedFlow (replay=0) with StateFlow for MESSAGE_PROGRESS intents so late subscribers still receive the latest download state. Fixes missed updates when fragments are recreated or the app is backgrounded. Also replaces the boolean download-complete pulse in ResourcesViewModel with a monotonic timestamp to avoid conflation.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved download progress handling so the latest status is retained and restored when screens reconnect.
  * Resource listings now refresh based on completion time and avoid duplicate refreshes after “all complete.”
  * Dictionary, synchronization, and related screens now process the right download updates consistently.
  * Added completion tracking to prevent repeated handling of duplicate “all complete” events.
* **Tests**
  * Added/updated coverage for latest progress delivery, replay to late subscribers, and completion event routing/de-duplication behavior.

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/base/BaseResourceFragment.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/base/BaseResourceFragment.kt
@@ -273,10 +273,21 @@ abstract class BaseResourceFragment : Fragment() {
                 broadcastService.events.collect { intent ->
                     if (isActive) {
                         when (intent.action) {
-                            DashboardActivity.MESSAGE_PROGRESS -> broadcastReceiver.onReceive(requireContext(), intent)
                             "ACTION_NETWORK_CHANGED" -> receiver.onReceive(requireContext(), intent)
                             DownloadService.RESOURCE_NOT_FOUND_ACTION -> resourceNotFoundReceiver.onReceive(requireContext(), intent)
                         }
+                    }
+                }
+            }
+        }
+        // MESSAGE_PROGRESS is served from latestDownloadProgress (not `events`) so a fragment
+        // whose view was recreated or briefly stopped right as a download finished still sees
+        // the latest state on resubscribe, instead of losing that update forever.
+        viewLifecycleOwner.lifecycleScope.launch {
+            viewLifecycleOwner.repeatOnLifecycle(Lifecycle.State.STARTED) {
+                broadcastService.latestDownloadProgress.collect { intent ->
+                    if (isActive && intent != null) {
+                        broadcastReceiver.onReceive(requireContext(), intent)
                     }
                 }
             }
--- a/app/src/main/java/org/ole/planet/myplanet/model/Download.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/model/Download.kt
@@ -2,6 +2,7 @@ package org.ole.planet.myplanet.model
 
 import android.os.Parcel
 import android.os.Parcelable
+import java.util.concurrent.atomic.AtomicLong
 
 class Download() : Parcelable {
     var fileName: String? = null
@@ -12,6 +13,7 @@ class Download() : Parcelable {
     var failed: Boolean = false
     var message: String? = null
     var fileUrl: String? = null
+    var completionId: Long = 0L
 
     constructor(parcel: Parcel) : this() {
         fileName = parcel.readString()
@@ -22,6 +24,7 @@ class Download() : Parcelable {
         failed = parcel.readByte() != 0.toByte()
         message = parcel.readString()
         fileUrl = parcel.readString()
+        completionId = parcel.readLong()
     }
 
     override fun writeToParcel(parcel: Parcel, flags: Int) {
@@ -33,6 +36,7 @@ class Download() : Parcelable {
         parcel.writeByte(if (failed) 1 else 0)
         parcel.writeString(message)
         parcel.writeString(fileUrl)
+        parcel.writeLong(completionId)
     }
 
     override fun describeContents(): Int {
@@ -40,6 +44,10 @@ class Download() : Parcelable {
     }
 
     companion object CREATOR : Parcelable.Creator<Download> {
+        private val completionIdGenerator = AtomicLong(0)
+
+        fun nextCompletionId(): Long = completionIdGenerator.incrementAndGet()
+
         override fun createFromParcel(parcel: Parcel): Download {
             return Download(parcel)
         }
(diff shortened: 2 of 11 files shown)


dogi:
**Review — Rating: 62/100 — `changes requested`.**

Verified against the real receivers, not just the description. The core diagnosis is right and parts of this are genuinely nice work — but moving MESSAGE_PROGRESS wholesale onto a replay-forever `StateFlow` re-fires **completion side-effects** on every resubscribe, which is the same class of bug your own comment warns about for `events`.

**Strengths (credit where due)**
- The missed-completion diagnosis is correct, and the `events` (replay=0) → retained-state split is the right shape for *progress*. Migration is complete — I checked every `broadcastService.events` MESSAGE_PROGRESS consumer (`BaseResourceFragment`, `ResourcesFragment`, `SyncActivity`, `DictionaryActivity`); none are left stranded, and `ProcessUserDataActivity`'s receiver is driven by `SyncActivity.registerReceiver`, which you migrated. Constant routing is consistent (`DownloadService.MESSAGE_PROGRESS` == `DashboardActivity.MESSAGE_PROGRESS` == `"message_progress"`).
- The pulse→timestamp change in `ResourcesViewModel` is a real fix: `value = true; value = false` with no suspension is genuinely conflated away by StateFlow. Good catch, and the comments/tests are excellent.
- `BaseResourceFragment.broadcastReceiver` is safe under replay because it no-ops unless `pendingDownloadUrls` is non-empty (`:126`) — a replayed terminal intent to a fresh fragment does nothing there.

**Blocking concern — terminal completion re-fires on resubscribe.** `_latestDownloadProgress` is a `@Singleton` `StateFlow` that retains the last MESSAGE_PROGRESS **for the whole app session** and replays it to every new subscriber. Two consumers act on completion *without* the `pendingDownloadUrls` guard:
- **`SyncActivity` → `ProcessUserDataActivity.checkDownloadResult` (`:88–90`)**: on a `completeAll` intent with `fromSync=false` it calls `customProgressDialog.show()` and **`installApk(...)`**. Every `STARTED` re-subscription (e.g. the APK self-update completes on the login screen, the user dismisses the system installer, returns to the app) replays that terminal intent → the progress dialog flashes and `installApk` fires **again**. The old `events`/replay=0 flow never did this. (For a stale *resource* completion it's worse — `installApk` would run on a non-APK `fileUrl`.)
- **`ResourcesFragment` (`:170`+)**: every resubscribe replays the terminal `completeAll` → `notifyDownloadComplete()` → a `refreshResourcesData()`. Your `lastHandledDownloadCompletion` guard (`:161`) doesn't stop this, because `notifyDownloadComplete()` mints a **fresh** `System.currentTimeMillis()` per call (`ResourcesViewModel.kt`), so the replayed completion always looks new. Net: a spurious full refresh on every return to the screen after any download ever completed — the opposite of the sibling #14796 goal.

**Suggested direction:** dedupe completion at the *intent* level, not a re-minted VM timestamp — carry a monotonic download/batch id (or the DownloadService completion time) in the intent and have each consumer ignore an id it already handled; or reset `_latestDownloadProgress` to `null` once a `completeAll` has been consumed so the terminal state isn't retained and replayed indefinitely. Per-file progress can keep replaying safely; it's specifically the terminal `completeAll` (installApk / refresh) that must be one-shot.

**Verdict:** right idea, solid tests — but the completeAll replay re-firing `installApk`/refresh needs guarding before merge. Another round.

---

dogi:
## Review — rating: 78/100

Traced producer → `BroadcastService` → every consumer, and verified the routing constants by value.

**Strengths (this is genuinely well-built)**
- Correct diagnosis and the right primitive: download progress/completion is *state*, and a `replay=0` `SharedFlow` drops it for any collector not subscribed at the emission instant. Moving `MESSAGE_PROGRESS` to a `StateFlow` (`BroadcastService.kt`) so late/re-subscribing collectors get the latest value is the correct fix, and the inline comment explaining *why other actions stay on `events`* is exactly the context a future maintainer needs.
- Routing is consistent — I checked: `DownloadService.MESSAGE_PROGRESS`, `DashboardActivity.MESSAGE_PROGRESS`, and the old literal `"message_progress"` are all the same string, and the producer sends `Intent(MESSAGE_PROGRESS)`, so the `intent.action == DownloadService.MESSAGE_PROGRESS` branch captures every progress intent and all consumers now read one source.
- The `completionId` (`AtomicLong`) dedup is the right instinct — since `StateFlow` re-delivers the latest completion on every resubscribe, a monotonic id lets a collector avoid re-refreshing. Replacing the `true→false` pulse in `ResourcesViewModel` with a never-repeating value is also correct (the pulse could be conflated away). Good test coverage of the routing/replay/null-start behavior in `BroadcastServiceTest`.

**Concern — stale/rotation completion replay re-invokes the APK installer (please guard or confirm):**
- `latestDownloadProgress` is a **Singleton** `StateFlow` that retains the last completion for the whole process, but `ProcessUserDataActivity.lastHandledCompletionId` is **per-instance** and resets to `0` when the activity is recreated (rotation, re-navigation). So: APK-update download completes (`completeAll=true`, `!fromSync`) → user rotates → the fresh `SyncActivity`/`ProcessUserDataActivity` collects `latestDownloadProgress`, immediately gets the retained terminal completion, and since `lastHandledCompletionId` is back to `0` it calls `checkDownloadResult` → **`installApk` fires again** (`ProcessUserDataActivity.kt:97`). The old `events` (`replay=0`) flow could not replay this, so this is a new behavior. The `ResourcesFragment` path degrades gracefully (a harmless extra `refreshResourcesData()`), but the installer re-pop is user-visible.
- Suggested fixes: persist the dedup across recreation (hoist `lastHandledCompletionId` into a ViewModel / `SharedPreferences`), or clear `_latestDownloadProgress` once a *terminal* completion has been consumed so it isn't replayed to later subscribers, or have `ProcessUserDataActivity` ignore completions minted before the activity was created.

**Nit:** `ResourcesViewModel` emits `System.currentTimeMillis()` as the completion signal and consumers compare with `>`. `completionId` is already a perfect monotonic, never-repeating value — emitting that (or a counter) would be more robust than a wall clock, which can repeat within a millisecond (StateFlow would conflate the second) or jump backwards on a clock adjustment (missed refresh).

**Verdict:** Strong, well-tested, well-documented change built on the right primitive — but the Singleton-StateFlow-vs-per-instance-dedup gap can replay a terminal completion into a recreated sync activity and re-trigger `installApk`. Please close that (or confirm it can't happen) — one more round.

---

PR #14871, opened 2026-07-21
courses: smoother progress status view modelling (fixes #14797)

### LINKED ISSUE #14797: course: label still showing "not started" upon completion
https://github.com/user-attachments/assets/e96bbb13-14a1-4ade-9493-0accd52e5aee

ragilzakaria:
### What Was the Cause
In CoursesViewModel.kt, the user's progress mapping was only loaded when isMyCourseLib was true (i.e. only when viewing the "My Courses" shelf). Catalog Fallback: When viewing the general course catalog ("Our Courses" search tab), isMyCourseLib is false, so progress mapping was passed as null. Because progress was always null for catalog courses, CoursesAdapter.kt fell back to displaying the "Not Started" status badge for every course

### Solution
- Modified CoursesViewModel.kt: Removed the conditional isMyCourseLib check so that course progress is always fetched from the database for all loaded courses.
- Modified CoursesAdapter.kt: Updated updateProgressViews to only show the horizontal progress bar if the user is enrolled in the course (course.isMyCourse == true). This ensures we don't display 0% progress bars for catalog courses the user hasn't joined, while correctly showing the progress status badge (e.g. "Completed" or "In Progress") everywhere the course is found.

[Screen_recording_20260721_195101.webm](https://github.com/user-attachments/assets/8b00e4d0-8656-4dd9-be32-7fbe754e04eb)

fixes #14797 

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated course progress UI so progress indicators (bars/badges) show only for courses owned by the current user.
  * Adjusted course list progress loading so progress data is prepared consistently, including non-“My Courses” library views.
  * Ensured courses not associated with the current user follow the “no progress” UI state.

* **Chores**
  * Bumped app version to **0.61.56** (**6156**).

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -12,8 +12,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6145
-        versionName = "0.61.45"
+        versionCode = 6156
+        versionName = "0.61.56"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/ui/courses/CoursesAdapter.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/courses/CoursesAdapter.kt
@@ -532,7 +532,7 @@ class CoursesAdapter(
         private fun updateProgressViews(position: Int) {
             val course = getItem(position) ?: return
             val progress = progressMap?.get(course.courseId)
-            if (progress != null) {
+            if (progress != null && course.isMyCourse) {
                 rowCourseBinding.courseProgress.max = getInt("max", progress)
                 val currentProgress = getInt("current", progress)
                 rowCourseBinding.courseProgress.progress = currentProgress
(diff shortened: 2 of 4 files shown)


Okuro3499 (approved):

dogi:
## Review — rating: 87/100

Traced the full render path in `CoursesAdapter`, not just the description.

**Strengths**
- Correct root-cause fix. The catalog passed `null` progress (`CoursesViewModel.kt` gated `getCourseProgress` on `isMyCourseLib`), and `updateProgressViews` maps `progress == null → "Not Started"`, so every catalog course looked un-started. Always fetching progress fixes the badge for completed/in-progress courses in the catalog.
- Good separation of the two concerns: the status **badge** is computed purely from `progress` (independent of `isMyCourse`), so it now renders correctly everywhere; the horizontal **progress bar** is gated on `progress != null && course.isMyCourse` (`CoursesAdapter.kt:534`) so non-enrolled catalog rows don't show a stray 0%/partial bar. That matches the intended UX.
- Null-safety holds: `JsonUtils.getInt` is `safeGet(0)` (returns 0 on null), and the `when` checks `progress == null` **first**, so the `current`/`max` reads below the if/else can't misclassify a null as "Completed".
- Test updated to assert the new behavior (`getCourseProgress` is always called). CI green. No file overlap with the other open PRs.

**Concerns (non-blocking)**
- `getCourseProgress(userId, allCourseIds)` now runs for the full catalog on every load, where it was previously skipped — a bit more DB work for large catalogs. It's necessary for correct badges, so acceptable, but worth keeping in mind if the catalog grows.

**Verdict:** Clean, correct, tested bug fix — mergeable.

---

dogi (approved):

PR #14877, opened 2026-07-21
teams: smoother repository leader marking (fixes #14827)

### LINKED ISSUE #14827: Leadership transfers never set `updated=true`, so they never sync to the server
When a leadership transfer is performed, the `updated` flag is never set to `true`. As a result, the record is not marked as dirty and is never included in the sync process, meaning the leadership transfer is never pushed to the server.

Okuro3499:
fixes #14827
When changing team leaders, set the `updated` flag on both demoted and promoted members so the changes sync to the server.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved team leader updates so membership records are only modified when the leader role actually changes.
  * Ensures leader-role changes are correctly marked for synchronization, improving propagation of both outgoing and incoming leader updates.


dogi:
## Review — rating: 90/100

Verified against the real sync path, not just the description.

**Strengths**
- Correct root-cause fix. `updateTeamLeader` (`TeamsRepositoryImpl.kt:1392`) previously flipped `isLeader` on the demoted and promoted membership docs but never set `updated`, so the leadership change stayed local. `getTeamsForUpload()` (`TeamsRepositoryImpl.kt:82-83`) uploads exactly the `RealmMyTeam` rows with `updated == true`, so without this flag the change never reached the server.
- Both sides handled — demoted leaders (`currentLeaders.forEach { … updated = true }`) and the new leader — so no membership doc is left dirty-but-unsynced.
- Uses the established convention: membership docs are already marked `updated = true` on join (`TeamsRepositoryImpl.kt:191`) and picked up by the same upload query, so this reuses a proven path rather than inventing one.
- Change is inside `executeTransaction`, mutating live Realm objects, so the writes persist. CI green, qlty reports no blocking issues.

**Concerns**
- None blocking. Only nit: no unit test was added. Given the repo now has good `TeamsRepositoryImpl` coverage, a small test asserting `updated == true` on both the old and new leader after `updateTeamLeader` would lock in the regression — but for a verified 2-line fix this is optional.

**Verdict:** Correct, minimal, well-targeted fix — mergeable.

---

dogi:
**Rating: 88/100**

**Strengths**
- Correct fix for a real sync bug. `updateTeamLeader` previously reassigned `isLeader` on every membership but never set `updated = true`, so the leader change was never marked dirty and wouldn't upload. Setting `updated = true` on the affected rows fixes that — verified `MyTeam.updated` exists (`@ColumnInfo(name = "isUpdated")`, `MyTeam.kt:45`) and that `updated` is what the upload path keys on.
- Nice secondary win: switching `map` → `mapNotNull` with the `if (membership.isLeader == shouldBeLeader) return@mapNotNull null` guard means only the memberships whose leader status actually changed are upserted (typically just the old + new leader), instead of rewriting the whole roster. Idempotent when the new leader is already the leader.

**Concerns**
- None blocking. Minor: consider setting `updatedDate` alongside `updated` if the sync/conflict path relies on it elsewhere (the prior code didn't, so no regression — just worth confirming).

**Verdict:** Correct, well-scoped fix + a clean optimization — ready.

---

dogi (approved):

PR #14885, opened 2026-07-21
actions: smoother gradle tooling parallel enabling (fixes #14884)

### LINKED ISSUE #14884: project parallel sync update recommended
Image

Okuro3499:
fixes #14884

## Summary by CodeRabbit

* **Performance**
  * Enabled parallel Gradle tooling operations to improve project synchronization speed.


dogi:
**Rating: 40/100**

**Concerns (false premise / no-op)**
- `org.gradle.tooling.parallel` is **not a recognized Gradle property**. Gradle's documented parallelism flag is `org.gradle.parallel`, and it's **already enabled** at `gradle.properties:17` (`org.gradle.parallel=true`). Gradle silently ignores unknown `org.gradle.*` keys, so as written this line changes nothing — the build won't error, but it won't enable anything either.
- The comment "Enabled parallel sync for Gradle 9.4+" is therefore misleading; a future reader would believe extra parallelism is configured when it isn't.

**What would actually help (if the goal is faster builds/sync):** the real levers are already on here (`org.gradle.parallel`, `org.gradle.caching`, `org.gradle.configuration-cache`). If there's a specific documented property you intended, please link the Gradle docs for it; otherwise this line should be dropped.

**Verdict:** No-op with a misleading comment — needs another round. Please remove it or replace it with a real, documented property (with a doc reference).

---

dogi:
**Re-review — unchanged, still changes requested.**

The label was moved back to `review`, but the diff is byte-identical to the previous revision — no new commit. The single added line is still:

```properties
org.gradle.tooling.parallel=true
```

which remains a **no-op**: `org.gradle.tooling.parallel` is not a recognized Gradle property (the real one, `org.gradle.parallel=true`, is already set at `gradle.properties:17`), and Gradle silently ignores unknown `org.gradle.*` keys. My original concern stands unaddressed.

To move this forward, either drop the line or replace it with a real, documented property and link the Gradle docs that define it. Restoring **changes requested**.

---

dogi:
**Correction — I was wrong; lifting changes requested → ready. Rating: 82/100**

I incorrectly claimed `org.gradle.tooling.parallel` was an unrecognized no-op. It is a **valid, documented Gradle property introduced in 9.4.0** that controls parallelism of Tooling Model Builders to improve IDE Sync times, independent of task-execution parallelism. Per the Gradle docs: *"Since 9.4.0 … allows running Tooling Model Builders in parallel, which can improve IDE Sync times … An explicit value will override `org.gradle.parallel` in this context"*, with the default being the value of `org.gradle.parallel`. The PR's "Gradle 9.4+" comment is accurate. My apologies for the mislabel.

**Strengths**
- Valid, correctly-named property with an accurate comment; enables/pins parallel IDE-sync tooling actions.

**Minor note (non-blocking)**
- Because this property *defaults to the value of `org.gradle.parallel`*, and `org.gradle.parallel=true` is already set (`gradle.properties:17`), the effective behavior is unchanged today — the line makes tooling parallelism explicit and keeps it enabled even if someone later sets `org.gradle.parallel=false`. That's a reasonable, defensive intent; just worth knowing it's a no-op *only relative to the current config*, not because the property is invalid.

**Verdict:** Valid, well-targeted config addition — ready.

Docs: https://docs.gradle.org/current/userguide/build_environment.html

---

dogi (approved):

dogi (approved):

PR #14890, opened 2026-07-22
teams: smoother survey submissions guest handling {fixes #14889)

### LINKED ISSUE #14889: mysurvey: table column headers misaligned with row data on tablet view
When viewing data tables (the one in the mySurveys section) on a tablet or wider screen, the column headers fail to align with the actual data in the rows. The layout grid seems broken, as the data cells and header cells are not using the same width distribution. For example, the "Status" data ("Not Started") is shifted too far to the right, causing it to visually float under the "Submitted by" header instead

Image

ragilzakaria:
- Modified SubmissionsAdapter.kt: Updated the showSubmittedBy function to set binding.submittedBy.visibility to View.INVISIBLE when the submitter name is blank instead of View.GONE.
- View.INVISIBLE preserves the view's layout dimensions and weight calculation, maintaining the perfect 4-column layout width distribution across all rows and keeping them aligned with the header cells.

Screenshot_20260722_191229

fixes #14889

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved how the submission author name is shown in the list, displaying it directly when available and hiding it when empty without disrupting layout.
* **Chores**
  * Bumped the app version (version code/name) for the next release.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -11,8 +11,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6161
-        versionName = "0.61.61"
+        versionCode = 6165
+        versionName = "0.61.65"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/ui/submissions/SubmissionsAdapter.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/submissions/SubmissionsAdapter.kt
@@ -105,13 +105,8 @@ class SubmissionsAdapter(
 
         private fun showSubmittedBy(submission: SubmissionUiModel) {
             val finalName = submission.submitterName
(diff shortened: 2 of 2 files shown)


Okuro3499 (approved):

dogi:
**Rating: 85/100**

**Strengths**
- Correct fix for the stated layout-shift problem. Switching the blank-submitter case from `View.GONE` to `View.INVISIBLE` keeps `submittedBy` occupying its space, so sibling `layout_weight` distribution stays stable instead of collapsing when the name is empty. Simplifying to a single `text = finalName` + ternary visibility is cleaner and behavior-equivalent for display (an invisible view with blank text renders nothing either way).

**Concerns**
- Minor: an `INVISIBLE` view still reserves vertical space, so if a row's submitter is blank there's now an empty gap where the text would be. That's the intended trade-off for weight preservation — just confirm it looks right in the list (vs. a fixed-height row where it's invisible anyway).

**Verdict:** Correct, self-contained layout-stability fix — ready.

---

dogi (approved):

PR #14899, opened 2026-07-22
community: Fix label update not reflecting immediately in Community Voices (fixes #14898)

Valerie923:
https://github.com/user-attachments/assets/0b9608e4-0fd2-4934-9b13-a09467bbdf60
fixes #14898 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved community news updates to detect changes to item labels.
  * Prevented duplicate updates when news items remain unchanged, including their labels.

PR #14901, opened 2026-07-22
teams: smoother voices repository community distinct changing (fixes #14900)

### LINKED ISSUE #14900: community voice does not show edited changes
https://github.com/user-attachments/assets/9d7dbd36-6624-42ea-bdf0-3db5fe391758

Valerie923:
https://github.com/user-attachments/assets/ebb5841c-0e6c-4f11-ac58-2f83b9a0dee8
fixes #14900 

## Summary by CodeRabbit

* **Bug Fixes**
  * Community news now refreshes when an item’s labels, message content, or edited status change, even if the item ID and timestamp remain unchanged.
* **Chores**
  * Updated the app version (version code and version name) for this release.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -11,8 +11,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6161
-        versionName = "0.61.61"
+        versionCode = 6163
+        versionName = "0.61.63"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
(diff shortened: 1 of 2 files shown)


Okuro3499 (approved):

dogi:
**Rating: 88/100**

**Strengths**
- Correct root-cause fix. The `distinctUntilChanged` predicate in `getCommunityNews` only compared `id` and `time`, so an in-place edit (message, edited-flag, or labels) that didn't change `time` was swallowed and the UI never refreshed. Expanding the equality to `labels?.toSet()`, `message`, and `isEdited` makes those edits register as a change → re-emission → immediate UI update.
- Verified against the model: `News` has `message` (`:34`), `labels: List<String>?` (`:49`, so `.toSet()` is valid), and `isEdited` (`:60`) — compiles cleanly. Comparing labels as sets (order-independent) is the right call.

**Concerns**
- Minor/forward-looking: the predicate now enumerates specific mutable fields, so any *future* editable display field (e.g. reactions) would need to be added here too. Not a defect — but a comment noting "keep in sync with editable fields," or comparing a derived content signature, would harden it.

**Cross-PR:** Superset of #14899, which adds only the `labels` comparison for the label-update case. This PR covers that **and** the edited-message case, on the same line. **Keep this one; close #14899.**

**Verdict:** Correct, complete fix for both the label and edit refresh bugs — ready.

---

dogi (approved):

PR #14919, opened 2026-07-22
sync: smoother transaction manager attachment downloading (fixes #14937)

dogi:
💡 **What:** Refactored sequential loops in `downloadCvAttachmentsFromBatch`, `downloadTeamAttachmentsFromBatch`, and `downloadCourseCoversFromBatch` to use `coroutineScope` and `launch`.\n🎯 **Why:** Speeds up large batch processing by downloading all attachments concurrently rather than sequentially awaiting each download.\n📊 **Measured Improvement:** The methods were optimized without altering behavior; formal benchmarking was omitted in favor of structural improvements which logically improve total wait time for network requests.

---
*PR created automatically by Jules for task [17427158765388855379](https://jules.google.com/task/17427158765388855379) started by @dogi*

## Summary by CodeRabbit

* **Performance Improvements**
  * Attachment downloads for CV resumes, team attachments, and course cover images now run concurrently to reduce overall synchronization time.
  * Download behavior and error handling remain unchanged.

* **Build Version Update**
  * Updated the app version code and version name to the next release.

PR #14918, opened 2026-07-22
⚡ Optimize team attachments batch download

dogi:
💡 **What:** Refactored downloadTeamAttachmentsFromBatch to download attachments concurrently using async/awaitAll.\n🎯 **Why:** Previously, attachments were downloaded sequentially in a loop, which caused high latency when processing many files in a single batch.\n📊 **Measured Improvement:** Improves batch sync time for team attachments by utilizing concurrent requests instead of sequential blocking.

---
*PR created automatically by Jules for task [12883944686116559126](https://jules.google.com/task/12883944686116559126) started by @dogi*

## Summary by CodeRabbit

* **Performance Improvements**
  * Improved synchronization speed by downloading missing team attachments concurrently.
  * Existing local attachments are skipped during synchronization.

PR #14922, opened 2026-07-22
⚡ Optimize sequential downloads in TransactionSyncManager

dogi:
Refactor batch downloading methods (e.g., downloadCourseCoversFromBatch) to use concurrent coroutines (async/awaitAll) for improved performance.

---
*PR created automatically by Jules for task [12069533555680651379](https://jules.google.com/task/12069533555680651379) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/services/sync/TransactionSyncManager.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/services/sync/TransactionSyncManager.kt
@@ -17,6 +17,7 @@ import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.async
 import kotlinx.coroutines.awaitAll
+import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.currentCoroutineContext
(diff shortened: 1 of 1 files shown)


dogi:
**Rating: 84/100**

**Strengths**
- The transformation itself is correct and behavior-preserving: `mapNotNull { … async { download… } }.awaitAll()` inside `coroutineScope` parallelizes the three per-batch download loops and still waits for all before returning. The `continue → return@mapNotNull null` and nested `else null` translations faithfully reproduce the original filter/skip logic in all three functions. Safe fan-out for the same reasons noted on #14919 (exceptions swallowed per-download, OkHttp per-host throttle).

**Concerns**
- **Duplicate of #14919.** That PR makes the same parallelization to the same three functions using `launch`. Since these downloads return `Unit`, `async`+`awaitAll` is slightly less idiomatic than `launch` (async signals a produced result that here is always `Unit` and discarded). The two conflict on identical lines and only one can merge.
- #14918 is a strict subset of this change (team only).

**Verdict:** Technically fine, but redundant — recommend closing in favor of #14919 (`launch` variant) to avoid the conflict. Not a code defect, purely deduplication.

---

PR #14929, opened 2026-07-23
teams: smoother repository task list handling (fixes #14882)

### LINKED ISSUE #14882: team tasks does not show up
https://github.com/user-attachments/assets/cf12265b-32ad-491b-bdc7-72c0b84deef0

ragilzakaria:
Updated Room DAO queries to (status IS NULL OR status != 'archived'), and set status = "active" in createTask

[Screen_recording_20260723_212420.webm](https://github.com/user-attachments/assets/f9d09a20-4406-4a7f-b077-90e58ffe5e56)

fixes #14882

## Summary by CodeRabbit

* **Bug Fixes**
  * Tasks with no status are now treated as active, so they correctly show up in open-task and team-task listings.
  * Newly created team tasks are explicitly set to an active status for consistent tracking.
  * Improved the tasks list behavior so the view scrolls to the top only after the updated list finishes rendering.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -11,8 +11,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6166
-        versionName = "0.61.66"
+        versionCode = 6188
+        versionName = "0.61.88"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/data/room/dao/TeamTaskDao.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/data/room/dao/TeamTaskDao.kt
@@ -8,7 +8,7 @@ import org.ole.planet.myplanet.model.TeamTask
 
 @Dao
 interface TeamTaskDao {
-    @Query("SELECT * FROM team_tasks WHERE status != 'archived' AND completed = 0 AND assignee = :userId")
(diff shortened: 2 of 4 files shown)


Okuro3499 (changes requested):
after adding I have to scroll to see it can it show it on top without the need of scrolling
https://github.com/user-attachments/assets/8a1463ac-e24a-4c8b-b585-4ce612749109

Okuro3499 (approved):

dogi:
**Review — rating: 88/100**

Correct root-cause diagnosis and a complete fix.

**Why it's right**
- The real bug is SQL three-valued logic: `status != 'archived'` evaluates to `NULL` (not `TRUE`) for rows where `status IS NULL`, so newly created tasks with no status were silently filtered out of both `getOpenTasksForUser` and `getTasksByTeamId`. `(status IS NULL OR status != 'archived')` correctly includes them.
- `createTask` now sets `status = "active"` so future tasks are never null — good belt-and-suspenders, and the query fix still recovers already-persisted null-status tasks.
- Moving `scrollToPosition(0)` into the `submitList { }` commit callback ensures the scroll happens after the new list is applied, not against the stale one.
- I checked the other DAOs: the remaining `status !=` queries are on `submissions`/`retry_queue` (different tables, out of scope here), so the `team_tasks` NULL bug is fully covered by these two edits.

**Concerns**
- Minor: a Room migration isn't needed (query handles existing nulls), but a quick unit/DAO test asserting a null-status task now appears would lock this regression down.

**Verdict:** Correct, minimal, complete fix for the reported bug — mergeable. Well diagnosed.

---

dogi (approved):

PR #14930, opened 2026-07-23
courses: smoother joining (fixes #14879)

### LINKED ISSUE #14879: delay in mylibrary  after adding resources from library
https://github.com/user-attachments/assets/7450e040-c928-4a3a-963e-a20645329fbc

ragilzakaria:
- Updated addToMyList(onComplete: (() -> Unit)?) in BaseRecyclerFragment.kt, ResourcesFragment.kt, and CoursesFragment.kt to accept a completion callback that fires in a finally block once Room DB writes finish
- Updated createAlertDialog so that tapping "Go to myLibrary" / "Go to myCourses" awaits completion of addToMyList before performing openMyFragment
- Added a hasAdded flag to ensure addToMyList is called exactly once per dialog interaction

[Screen_recording_20260723_214137.webm](https://github.com/user-attachments/assets/8a5821b9-c876-4fe6-af6e-bec50d93c982)

fixes #14879

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate additions when confirming course or resource selections.
  * Improved reliability when adding items to My Courses or My Library, including handling incomplete or unsuccessful operations.
  * Ensured loading indicators, controls, and selection states are restored after the operation finishes.

* **Improvements**
  * Navigation to My Library now occurs only after items are successfully processed.
  * Existing success notifications and error handling remain available.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -11,8 +11,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6166
-        versionName = "0.61.66"
+        versionCode = 6177
+        versionName = "0.61.77"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/base/BaseRecyclerFragment.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/base/BaseRecyclerFragment.kt
@@ -110,11 +110,17 @@ abstract class BaseRecyclerFragment<LI> : BaseRecyclerParentFragment<Any?>(), On
         }
     }
 
-    open fun addToMyList() {
-        if (isAddInProgress) return
+    open fun addToMyList(onComplete: (() -> Unit)? = null) {
+        if (isAddInProgress) {
+            onComplete?.invoke()
+            return
+        }
 
         val itemsToAdd = selectedItems?.toList() ?: emptyList()
-        if (itemsToAdd.isEmpty()) return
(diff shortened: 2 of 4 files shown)


Okuro3499 (approved):

dogi:
**Rating: 87/100**

**Strengths**
- Correct root-cause fix for the "update delay when adding." The `addToMyList(onComplete: (() -> Unit)? = null)` callback is invoked in a `finally` block, so it fires reliably on every path — early guards (`isAddInProgress`, empty selection, empty ids), success, and failure. Deferring navigation (`openMyFragment`) into `onComplete` means the My Library/Courses list is opened *after* the add completes, so it no longer shows stale data.
- The `hasAdded` flag in both `createAlertDialog`s cleanly fixes the double-add: previously the positive button *and* `setOnDismissListener` both called `addToMyList()` (relying on the `isAddInProgress` race to dedupe); now dismiss only adds when no button handled it. Deterministic.
- `ResourcesFragment`'s override mirrors the pattern with its own `try/finally { onComplete?.invoke() }` and an `else` branch for the no-op case, so the contract holds there too.

**Concerns**
- Minor: in the base `finally`, `onComplete?.invoke()` also runs after the `view == null || !isAdded || isFinishing` early return, so the `CoursesFragment` callback would still build a fragment and call `homeItemClickListener?.openMyFragment(...)` even when the view is gone. It goes through the activity listener so it's likely harmless/no-op, but consider guarding the navigation callback on `isAdded`.
- No test for the new callback sequencing — reasonable to add one given this is a previously-racy path.

**Verdict:** Solid, correct fix for the add-then-navigate race and the double-add — ready.

---

dogi (approved):

PR #14931, opened 2026-07-23
courses: smoother submissions repositories exam utils answering (fixes #14878)

### LINKED ISSUE #14878: Exam always returns wrong answer for all options
https://github.com/user-attachments/assets/56d48f17-bb3f-4b79-9a0a-c818277e612c

Okuro3499:
fixes #14878
[Screen_recording_20260723_180019.webm](https://github.com/user-attachments/assets/a84e1182-5f44-443b-9c27-637dfc2bdb81)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved answer checking for single- and multiple-choice questions with case-insensitive exact matching.
  * Correct answers are now derived more reliably from the question’s choice data, including support for multiple input formats.
  * Exam navigation now stops and shows the “incorrect answer” prompt if saving/grading the current response fails.
  * Enhanced reliability of exam submission creation and grading behavior.
* **Tests**
  * Updated unit test wiring to match the latest repository setup.


ragilzakaria (approved):
Great fix for Issue #14878

dogi:
**Rating: 74/100**

**Strengths**
- Fixes a real scoring bug and moves in the right direction. Previously single-select compared the raw answer **id** against correct choices that were resolved to `res` text (`extractCorrectChoices` old single branch) → guaranteed mismatch. Now both the answer (`ansForCheck = getChoiceTextById(question, ans)`) and the correct choices (`correctChoiceList`) are resolved to choice text, so they're comparable. I traced `listAns` (keys = choice **text**, values = **id**, per `ExamTakingFragment:727` / repo `:475`), so `listAnsForCheck = listAns.keys.associateWith { it }` correctly makes multi-select compare text-vs-text too.
- Good adds: case-insensitive comparison in `checkSelectAnswer`/`checkMultipleSelectAnswer`, and `ExamTakingFragment` now blocks advancing on an incorrect answer in exam mode (`if (type == "exam" && !cont) { snackbar(incorrect_ans); return }`). Compile-verified: `StepExam.id` is non-null (so the removed `?:` guards and `Submission` non-null return are correct), `ExamQuestion.correctChoiceList` exists, `updateAnsDb(): Boolean`, and `incorrect_ans` string all line up.

**Concerns (blocking — scoring correctness)**
- **Resolver asymmetry.** The correct-choice resolver `resolveChoiceValue` prefers `res` → `text` → raw (`CoursesRepositoryImpl.extractCorrectChoices`), but the answer resolver `getChoiceTextById` (`ExamAnswerUtils.kt:14`) reads **only `text`**. The codebase demonstrably stores `res` on choices (the old code read it), so for any choice where `res` ≠ `text`, the correct choice resolves to `res` while the submitted answer resolves to `text` → a correct answer is scored **wrong**. Since this is exam scoring, the two resolvers must agree: resolve `correctChoiceList` to `text` (matching `getChoiceTextById`), or teach `getChoiceTextById` to use the same `res`→`text` precedence — but pick one consistently. Please confirm the exam choice data shape and align them.

**Minor**
- Removing `surveysRepositoryProvider` from `SubmissionsRepositoryImpl` is unrelated cleanup (verified unused) — fine, but it's scope creep in a scoring-fix PR; worth calling out in the description.

**Verdict:** Right direction and fixes the primary id-vs-text bug, but the `res`/`text` resolver mismatch is a latent mis-scoring risk that should be resolved before merge — another round.

---

J-S-webskas (commented):
In the video shown for the issue #14878, the course type is multiple choice. I've tested with other multiple choice courses, and they seem to be working. However, for "Addition - simple math" course, the test is written. Ex. what is 5 + 5. The written answer 10 is being marked as incorrect. Does this issue fix apply to written responses?

Also as a side note, on step 7 of the course "moon", question 1 of the test has 3 repeated answers.

dogi:
**Review — rating: 82/100**

I traced this carefully because it's critical grading logic, and the fix is sound.

**Why it's correct**
- Root cause was an id-vs-text mismatch: `extractCorrectChoices` could yield choice **ids** while the user's answer was compared as text (or vice-versa), so correct answers were graded wrong. The fix resolves *both* sides to the display value (`ExamAnswerUtils.choiceDisplayValue` → `text`, falling back to `res`) and compares case-insensitively.
- Single-select: `ansForCheck = getChoiceTextById(question, ans)` (text) vs text `correctChoices` — matches.
- Multi-select: this looked suspicious at first (`listAnsForCheck = listAns.keys.associateWith { it }`), but it's right — `listAns` is `Map<displayText, choiceId>` (see `ExamTakingFragment.kt:604,727`), so `.keys` are the selected display texts, correctly compared to the text-resolved `correctChoices`. Good catch to normalize here.
- Compile-safety verified: `ExamQuestion.correctChoiceList` exists, the removed `surveysRepositoryProvider` had exactly one (constructor-only, unused) reference, `R.string.incorrect_ans` + the `Snackbar` import exist, and narrowing the `createExamSubmission` override from `Submission?` to `Submission` is legal.

**Concerns**
1. **Behavioral/UX change — please confirm it's intended.** `btnNext` now blocks advancing on a wrong answer in exam mode (`Snackbar` + `return@launch`). Previously the user could proceed. That's a real semantics shift for graded exams (it effectively forces correct-to-continue). Aligned with "ensure correct answer is verified," but worth an explicit product sign-off.
2. **No regression test for the grading fix.** This is exactly the logic that was broken; please add `ExamAnswerUtils.checkCorrectAnswer` unit tests covering select / selectMultiple / text, id-vs-text resolution, and case-insensitivity. Without them this can silently regress again.
3. Minor: trailing-whitespace blank line added before `ansForCheck` in `SubmissionsRepositoryImpl`.

**Verdict:** Correct fix to a genuine grading bug — mergeable, but add the `ExamAnswerUtils` tests and confirm the block-on-incorrect UX is desired.

---

PR #14933, opened 2026-07-23
teams: smoother task notifying (fixes #14932)

### LINKED ISSUE #14932: Task deadline notifications silently overwrite each other
```                                                                                                                                                                 
  // NotificationUtils.kt                                                                                                                                                            
  fun create(context: Context, smallIcon: Int, contentTitle: String?, contentText: String?) {                                                                                        
      ...                                                                                                                                                                            
      manager.notify(111, notification)   // hardcoded ID                                                                                                                            
  }                                                                                                                                                                                  
                                                                                                                                                                                     
  // TaskNotificationWorker.kt                                                                                                                                                       
  tasks.forEach { task ->                                                                                                                                                            
      create(applicationContext, R.drawable.ole_logo, task.title, ...)   // same call, every task                                                                                    
  }                                                                                                                                                                                  
  Every call to create() uses the same hardcoded notification ID (111). When a user has more than one team task due in the next 24 hours, each create() call replaces the previous   
  system notification — Android's notify(id, ...) overwrites any notification sharing that ID. Only the last task in the list ever produces a visible notification. Worse, right     
  after the loop:                                                                                                                                                                    
  val taskIds = tasks.mapNotNull { it.id }.filter { it.isNotBlank() }                                                                                                                
  if (taskIds.isNotEmpty()) {                                                                                                                                                        
      runCatching { teamsRepository.markTasksNotified(taskIds) }                                                                                                                     
  }                                                                                                                                                                                  
  all tasks get marked as notified — including the ones whose notification was silently clobbered — so they'll never be retried. A user with 3 due tasks effectively only ever learns
  about 1, permanently. There's already a proper per-item ID pattern nearby (createTaskNotification(taskId, ...) builds a NotificationConfig keyed by taskId) —                      
  TaskNotificationWorker just isn't using it. This also shares ID 111 with the "Recording Audio" notification (ResourceViewerFragment.kt:111) and a sync-progress notification in    
  SyncManager.kt, so any of those firing at the same time can stomp each other too.
```

Okuro3499:
fixes #14932

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved task notification delivery and formatting for pending tasks.
  * Ensured task expiration dates are displayed consistently in notifications.

* **Chores**
  * Updated the application version to 0.61.89.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -11,8 +11,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6166
-        versionName = "0.61.66"
+        versionCode = 6189
+        versionName = "0.61.89"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/services/TaskNotificationWorker.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/services/TaskNotificationWorker.kt
@@ -7,11 +7,10 @@ import androidx.work.WorkerParameters
 import dagger.assisted.Assisted
 import dagger.assisted.AssistedInject
 import java.util.Calendar
-import org.ole.planet.myplanet.R
 import org.ole.planet.myplanet.repository.NotificationsRepository
 import org.ole.planet.myplanet.repository.TeamsRepository
 import org.ole.planet.myplanet.utils.FileUtils
-import org.ole.planet.myplanet.utils.NotificationUtils.create
+import org.ole.planet.myplanet.utils.NotificationUtils
 import org.ole.planet.myplanet.utils.TimeUtils.formatDate
 
 @HiltWorker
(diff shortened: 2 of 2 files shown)


dogi:
**Review — rating: 84/100**

Verified against the real notification infra — this correctly makes tasks notify individually.

**Strengths**
- `createTaskNotification` sets `NotificationConfig(id = taskId, …)` and `showNotification` notifies with an id derived from `config.id` (`notificationManager.notify(notificationId, …)`), so each task now surfaces as its **own** notification instead of collapsing into one. That's the fix.
- Bonus: routing through `NotificationUtils` adds channel handling and dedup (`getNotificationIdIfShouldShow` / `markNotificationAsShown`), so repeated worker runs won't re-spam the same task.
- Compiles: `getInstance`, `createTaskNotification`, `showNotification`, and the single-arg `formatDate(Long)` all exist.

**Concerns**
- Wording mismatch: this worker fires **deadline** reminders (the old copy was "Task expires on …"), but the shared `createTaskNotification` renders "✅ New Task Assigned" / "Due: …". Reads like an assignment, not an expiry warning. The string lives in `NotificationUtils` (pre-existing), so it's not introduced here — but consider a deadline-specific config or neutral wording so the semantics match.
- Those notification strings are also hardcoded (not localized) — again pre-existing, worth a follow-up.

**Verdict:** Correct per-task notification fix — mergeable; consider the "New Task Assigned" wording for a deadline reminder.

---

dogi (approved):

PR #14961, opened 2026-07-24
dashboard: smoother landscape navigating (fixes #14949)

### LINKED ISSUE #14949: sidebar menu does not automatically go to the top upon click in landscape mode
https://github.com/user-attachments/assets/ebae41d8-a602-49a4-80ca-3fbc2c355e16

when in landscape mode, if i scroll the sidebar and then click something else, and then i open the sidebar again, it does not automatically scroll to the top.

ragilzakaria:
[Screen_recording_20260724_202450.webm](https://github.com/user-attachments/assets/bb657365-a740-4aad-a64d-2bd5904ba3d0)

fixes #14949 

## Summary by CodeRabbit

* **Bug Fixes**
  * The navigation drawer now consistently returns to the top when opened, closed, or interacted with.
  * Improved first-launch and logo-triggered drawer behavior to ensure the menu starts at the beginning.


Okuro3499 (approved):

dogi:
**Review — 76/100** (code-level; landscape behavior needs an on-device confirm)

Verified at the code level; CI green (build default+lite, test default).

**Strengths**
- The approach is sound: it resets the nav drawer's menu `recyclerView` to the top via `scrollToPosition(0)` at the right moments — a `DrawerLayout.SimpleDrawerListener` plus explicit calls before each `openDrawer()` (first-launch auto-open, logo tap, and after a drawer-item selection). All calls are null-safe (`result?.recyclerView?`). Functionally this should keep the menu starting at the top.

**Concerns (non-blocking)**
- **`onDrawerStateChanged` is the aggressive one.** It fires for every transition (`STATE_DRAGGING`/`STATE_SETTLING`/`STATE_IDLE`), and each fire scrolls the menu to 0. If a user edge-drags the drawer partially (generating state changes) while having scrolled the menu, it'll snap back to top mid-interaction. For "always show the top when the drawer opens," `onDrawerOpened` alone is the precise hook — consider dropping the `onDrawerStateChanged` override.
- **Redundancy:** with the listener already handling open/close/state, the explicit `scrollToPosition(0)` before each `openDrawer()` is belt-and-suspenders. Harmless, but you could keep just one mechanism.

**Note:** the reported issue (#14949) is specifically about landscape, and the head here is unchanged since the earlier review round — the actual "does the sidebar now reset to top in landscape" verdict is an on-device check I can't reproduce from the diff. Code-wise it's low-risk and should do it; I'd defer the final visual sign-off to whoever tests it on a device.

**Verdict:** Code is correct and low-risk (CI green); trim the `onDrawerStateChanged`/redundant calls if you want it tighter, and confirm the landscape behavior on-device before merge.

---

dogi (approved):

PR #14972, opened 2026-07-24
Fix: Add error logging to OnboardingAdapter

dogi:
Added error logging and a fallback view to OnboardingAdapter to prevent crashes when a view fails to inflate or when UI elements are missing. Verified by running: ./gradlew compileDefaultDebugKotlin

---
*PR created automatically by Jules for task [9153826015239377673](https://jules.google.com/task/9153826015239377673) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/ui/onboarding/OnboardingAdapter.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/onboarding/OnboardingAdapter.kt
@@ -1,6 +1,7 @@
 package org.ole.planet.myplanet.ui.onboarding
 
 import android.content.Context
+import android.util.Log
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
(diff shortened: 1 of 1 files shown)


J-S-webskas (approved):

dogi:
**Review — rating: 82/100**

Reasonable defensive hardening.

**Strengths**
- Happy path is byte-for-byte unchanged; the `catch` logs and returns a fallback `View` that's added to the container, so the `PagerAdapter` contract (`destroyItem` removing the returned object, `isViewFromObject` identity) still holds. An inflate/`findViewById` failure now degrades to an empty page + `Log.e` instead of crashing onboarding.

**Concerns (nits)**
- `catch (Exception)` is broad — it'll swallow unexpected errors too. It does log them, so acceptable, but consider narrowing if only inflation is the concern.
- A blank fallback page is a silent UX gap; fine as a safety net, just noting.

**Verdict:** Safe, non-behavior-changing on the happy path — mergeable.

---

PR #14974, opened 2026-07-24
all: smoother feedback caching (fixes #15124)

### LINKED ISSUE #15124: Cache FeedbackAdapter drawable backgrounds

dogi:
Added companion object in FeedbackAdapter to cache drawable backgrounds and replaced separate bg_grey.xml with programmatic tinting. Tested compilation with `./gradlew compileDefaultDebugUnitTestKotlin --no-daemon`.

---
*PR created automatically by Jules for task [14834849573303473514](https://jules.google.com/task/14834849573303473514) started by @dogi*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved feedback list styling for priority and status indicators.
  * Ensured color accents remain consistent when these indicators are displayed or refreshed.
  * Removed the outdated grey indicator styling.

* **Release**
  * Updated the app version to 0.62.25.


dogi:
**Review — rating: 50/100**

The `bg_grey` removal is fine — I verified nothing else references it and `bg_primary` is shape/corner/padding-identical (only the solid color differs), so tinting `bg_primary` with `md_amber_500` recreates it faithfully. But the drawable caching introduces two real problems.

**1. Sharing one `Drawable` instance across many views (rendering bug).** `bgPrimary`/`bgGrey` are single cached instances, then assigned to `tvPriority` **and** `tvStatus`, across **every** row's ViewHolder. A `Drawable` is stateful w.r.t. its host View — setting it as a background makes the View its single `Callback` and sets its bounds. Sharing one instance across multiple simultaneously-visible RecyclerView item views is the classic drawable-sharing bug (bounds/callback belong to whichever view bound it last), and can cause backgrounds to render at the wrong size or not invalidate correctly. Resource drawables already share a `ConstantState`, so the correct way to reuse is `drawable.constantState?.newDrawable()?.mutate()` per assignment — not one shared instance.

**2. Static (`companion`) caching of a Context-derived `Drawable` leaks + goes stale.** Holding a `Drawable` (which references resources/Context via its `ConstantState`) in a companion object keeps it for the whole app lifetime — a context/resource leak — and it never refreshes on a day/night theme change (the amber tint / colors would be stale).

**Recommendation:** either revert to `ResourcesCompat.getDrawable(...)` per bind (it's cheap — `ConstantState` is already cached by the resource system), or cache at the **adapter-instance** level and hand each view its own `newDrawable().mutate()`.

**Verdict:** Shared-instance drawable across RecyclerView views + static context leak — changes requested.

---

dogi:

My earlier changes-requested flag was: the static companion-object drawable cache shared one mutable `Drawable` instance across all rows (tint bleed / rendering glitches) and leaked a `Context`. Commit `c57ff81` addresses it exactly right:

- The static drawable cache is **gone**. Each bind now fetches a fresh `ContextCompat.getDrawable(context, R.drawable.bg_primary)` and applies color via `ViewCompat.setBackgroundTintList(view, ColorStateList.valueOf(...))` — per-view, no shared mutable instance, no static `Context`/`Drawable` retention. This is the correct pattern.
- Behavior preserved: priority `yes` / status `open` → `mainColor`; otherwise → `md_amber_500`, which is exactly the color the deleted `bg_grey.xml` used as its solid fill. Corners/padding now come uniformly from `bg_primary` (same 16dp/8-16 metrics the grey shape had), so the visual is unchanged. Dropping `bg_grey.xml` is a clean consolidation.
- CI green (build default+lite, test default).

**Nit (non-blocking):** `TintTest.kt` doesn't actually assert anything about tinting — it only `assertNotNull(copied)` and its own comment concedes it can't verify the tint. It's harmless but adds no coverage; consider asserting the resolved `ColorStateList` on the view after bind (Robolectric can read `view.backgroundTintList`) or dropping the test.

**Verdict:** The shared-drawable + static-leak concern is genuinely resolved and behavior is preserved — moving from changes requested to ready.

---

Okuro3499 (approved):

PR #14968, opened 2026-07-24
all: less notifications repository survey id is more (fixes #15180)

dogi:
Removes the `survey` branch from the in-app notifications click handler and the now-unused `getSurveyId` by-name lookup.

### Changes
- Removed `NotificationsRepository.getSurveyId(relatedId)` and its implementation in `NotificationsRepositoryImpl` (which did `examDao.getAll().firstOrNull { it.name == relatedId }?.id`).
- Removed `NotificationsViewModel.getSurveyId(...)`.
- Removed both `"survey"` branches from `NotificationsFragment.handleNotificationClick` — the `result` resolution and the `SubmissionsAdapter.openSurvey(...)` handling.

> **Note:** the previous title/description (“Add a by-name exam DAO query and use it in NotificationsRepository”) did not match the diff — no DAO query is added; `getSurveyId` is *removed*. Title and body corrected to reflect the actual change.

---
*PR created automatically by Jules for task [11712354683692595509](https://jules.google.com/task/11712354683692595509) started by @dogi*

## Summary by CodeRabbit

* **Changes**
  * Removed “survey” notifications from recognition, grouping, labeling, and specialized formatting.
  * Selecting a “survey” notification no longer opens a survey; it now follows the default notification icon.
  * Removed support for opening internal storage settings from “storage” notifications.
  * Task, join request, and resource notification actions remain supported.
  * Team chat notification counts now rely on the voices data source instead of the prior dependency.

PR #14979, opened 2026-07-24
Remove redundant survey notification auto-navigation and unused getSurvey lookup

dogi:
Removes the duplicate survey-notification navigation path and the now-unused `getSurvey` lookup, consolidating survey-notification handling on the single `from_notification` → `SurveyFragment` path.

### Changes
- Removed `SurveysRepository.getSurvey(id)` and its implementation in `SurveysRepositoryImpl` (the `examDao.getById(id)` primary lookup with the `examDao.getByType("surveys").firstOrNull { it.name == id }` name fallback).
- Removed `DashboardViewModel.handleSurveyNavigation()` and the `surveyNavigationEvent` `SharedFlow` (and its `asSharedFlow()` exposure).
- Removed, in `DashboardActivity`: the `auto_navigate` `TYPE_SURVEY` branch in `handleNotificationIntent`, the `handleSurveyNavigation()` helper, and the `surveyNavigationEvent` collector.

Survey notifications tapped from the notification body continue to open `SurveyFragment` via the untouched `from_notification` path.

> **Note:** the previous title/description (“add a type+name survey DAO query for the getSurvey fallback”) did not match the diff — this PR *removes* `getSurvey` rather than adding a DAO query. Title and body corrected to reflect the actual change.

---
*PR created automatically by Jules for task [1370632456018352024](https://jules.google.com/task/1370632456018352024) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/data/room/dao/LegacyEntityDaos.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/data/room/dao/LegacyEntityDaos.kt
@@ -65,6 +65,7 @@ interface ExamDao {
     @Query("SELECT * FROM exams") suspend fun getAll(): List<StepExam>
     @Query("SELECT * FROM exams") fun observeAll(): Flow<List<StepExam>>
     @Query("SELECT * FROM exams WHERE type = :type") suspend fun getByType(type: String): List<StepExam>
+    @Query("SELECT * FROM exams WHERE type = :type AND name = :name LIMIT 1") suspend fun getByTypeAndName(type: String, name: String): StepExam?
     @Query("SELECT * FROM exams WHERE type = :type") fun observeByType(type: String): Flow<List<StepExam>>
     @Query("SELECT * FROM exams WHERE teamId = :teamId") suspend fun getByTeamId(teamId: String): List<StepExam>
     @Query("SELECT * FROM exams WHERE teamId = :teamId AND type = :type") suspend fun getByTeamIdAndType(teamId: String, type: String): List<StepExam>
--- a/app/src/main/java/org/ole/planet/myplanet/data/room/dao/LegacyEntityDaos.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/data/room/dao/LegacyEntityDaos.kt
@@ -65,7 +65,6 @@ interface ExamDao {
     @Query("SELECT * FROM exams") suspend fun getAll(): List<StepExam>
     @Query("SELECT * FROM exams") fun observeAll(): Flow<List<StepExam>>
     @Query("SELECT * FROM exams WHERE type = :type") suspend fun getByType(type: String): List<StepExam>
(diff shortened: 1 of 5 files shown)


dogi:
**Review — rating: 87/100**

Correct query optimization, consistent with #14968.

**Strengths**
- The `getSurvey` fallback `getByType("surveys").firstOrNull { it.name == id }` → `getByTypeAndName("surveys", id)` (`WHERE type = :type AND name = :name LIMIT 1`) is exactly equivalent: bound-param `=` matches the old `==`, `LIMIT 1` matches `firstOrNull`, and the `getById(id)` primary lookup + fallback order is unchanged.
- Avoids materializing all survey-type exams just to scan for one name.

**Concerns**
- Minor: unlike the sibling #14968, no test was added. A quick `getByTypeAndName` stub test (primary hit, fallback hit, both miss) would lock the fallback in — nice-to-have, not blocking.

**Verdict:** Correct, efficient — mergeable.

---

Okuro3499 (dismissed):
this feature is unused I removed it's functionality but this remained

dogi:
**Review — 55/100** (title/description corrected; one functional concern to resolve)

Per request, I've rewritten the title and body — they previously described adding a type+name DAO query for `getSurvey`, but the diff actually **removes** `getSurvey` and the auto-navigate survey path. Metadata now matches the code.

**The remaining concern — the "Take Survey" notification action button is orphaned.**
Removing the `auto_navigate` `TYPE_SURVEY` branch leaves a live entry point with no handler:
- `NotificationUtils.kt:369` still adds the action: `TYPE_SURVEY -> builder.addAction(R.drawable.survey, "Take Survey", createOpenPendingIntent(config))`.
- That action routes through `NotificationActionReceiver.ACTION_OPEN_NOTIFICATION` (`NotificationActionReceiver.kt:60-61`), which launches `DashboardActivity` with `auto_navigate=true`, `notification_type=survey`, `related_id=…`.
- The `auto_navigate` `when(notificationType)` block in `handleNotificationIntent` has **no `else`**, so with `TYPE_SURVEY` deleted, tapping **"Take Survey" now does nothing**.

Tapping the notification **body** still works (the `from_notification` path → `SurveyFragment` is untouched), so this is a partial regression: the body is actionable, the explicit button is dead.

**To resolve, pick one:**
1. Also drop the `TYPE_SURVEY` action button in `NotificationUtils.kt:369` (if the body-tap path is now the intended single entry point), or
2. Keep a minimal `auto_navigate` `TYPE_SURVEY` branch that routes `related_id` to `SurveyFragment` (same as the body path), so the button and body behave identically without needing `getSurvey`.

Note this also overlaps with **#14968** (which *adds* a by-name exam query for `NotificationsRepository.getSurveyId`) — worth confirming the two aren't pulling survey-by-name lookup in opposite directions.

**Verdict:** Metadata fixed and the cleanup direction is reasonable, but the orphaned "Take Survey" button is a real regression — handle the button (remove or re-route) before merge.

---

PR #14984, opened 2026-07-24
login: smoother user repository case insensitive finding (fixes #15082)

### LINKED ISSUE #15082: Optimize findUserByName with COLLATE NOCASE

dogi:
Optimized the user search by introducing `getByNameIgnoreCase` using `COLLATE NOCASE`.

Executed:
`./gradlew testDefaultDebugUnitTest --tests "org.ole.planet.myplanet.repository.UserRepositoryImplTest" --no-daemon` to verify no regressions in functionality.

---
*PR created automatically by Jules for task [8477293892575625154](https://jules.google.com/task/8477293892575625154) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -11,8 +11,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6182
-        versionName = "0.61.82"
+        versionCode = 6192
+        versionName = "0.61.92"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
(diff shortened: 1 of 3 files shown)


dogi:
**Review — rating: 85/100**

Good optimization — the old `findUserByName` did `userDao.getAll().firstOrNull { it.name.equals(name, ignoreCase = true) }`, materializing every `UserEntity` in the table just to find one. The new `getByNameIgnoreCase` pushes it to a single `WHERE name = :name COLLATE NOCASE LIMIT 1`.

**Strengths**
- Correct case-insensitive match with `COLLATE NOCASE`, `LIMIT 1` matching the old `firstOrNull` semantics. Avoids loading/deserializing the whole users table.

**Concerns (worth noting, not blocking)**
- **`NOCASE` is ASCII-only.** SQLite's `NOCASE` collation folds only `A-Z`/`a-z`, whereas Kotlin's `equals(ignoreCase = true)` does full Unicode case folding. For a name containing non-ASCII letters that differ only by case, the two would diverge. Given these are CouchDB usernames (effectively ASCII), this is almost certainly fine — but if names can be non-ASCII, this is a behavior change.
- Non-deterministic tie-break for duplicate names is unchanged (both old and new pick an arbitrary "first"), so no regression there.

**Verdict:** Correct, meaningful DB optimization — mergeable (confirm usernames are ASCII to fully retire the Unicode-folding caveat).

---

Okuro3499 (approved):

PR #14987, opened 2026-07-24
teams: smoother voices repository dao querying (fixes #15154)

### LINKED ISSUE #15154: Move team-scoped news filtering from in-memory to Room DAO
`VoicesRepositoryImpl` resolves team-scoped news by loading every top-level post through `newsDao.getTopLevel()` / `getTopLevelFlow()` and then discarding most of them in Kotlin via `matchesTeam`. `getNewsByTeamId`, `getFilteredNews`, and `getDiscussionsByTeamIdFlow` all pay for the full table before filtering.

The predicate — `viewableBy = "teams"` with a matching `viewableId`, or the team's `_id` present in the `viewIn` JSON — is expressible in SQL, so the filtering can move into `NewsDao` and let SQLite do the work.

One detail worth getting right in the move: `matchesTeam` matches the `viewIn` JSON with a literal `contains`, but a SQL `LIKE` treats `%` and `_` as wildcards. A `teamId` containing either character would match unrelated teams unless the pattern is escaped and the query declares `ESCAPE`.

dogi:
Move team-scoped news filtering from in-memory to Room DAO

- Add `getTopLevelByTeam` and `getTopLevelByTeamFlow` to `NewsDao` to filter by `viewableId` and `viewIn` JSON at the SQL layer.
- Update `VoicesRepositoryImpl` to use these new queries instead of filtering in-memory with `matchesTeam`.
- Remove the unused `matchesTeam` function.
- Update related test cases in `VoicesRepositoryImplTest` to mock the new queries.
- Ran `./gradlew testDefaultDebugUnitTest --tests "org.ole.planet.myplanet.repository.VoicesRepositoryImplTest*" --no-daemon` to ensure tests pass.

---
*PR created automatically by Jules for task [4782625320910935195](https://jules.google.com/task/4782625320910935195) started by @dogi*

## Summary by CodeRabbit

* **Bug Fixes**
  * Team-specific news and discussions are now filtered accurately at the data source.
  * Escaped team identifiers prevent wildcard characters from causing incorrect matches.
  * Results continue to exclude replies and respect team visibility settings.

* **Tests**
  * Added coverage for team filtering, wildcard handling, and repository behavior.


dogi:
**Review — rating: 72/100**

The direction is right — pushing `matchesTeam` filtering into SQL (`getTopLevelByTeam` / `getTopLevelByTeamFlow`) avoids materializing all top-level news. And the `viewableBy = 'teams' COLLATE NOCASE AND viewableId = :teamId COLLATE NOCASE` half faithfully mirrors the old `equals(..., ignoreCase = true)` (ASCII-equivalent for ids), with ordering preserved. But two things need another round.

**1. Unescaped `LIKE` wildcards — a visibility-isolation risk.** The old code did a *literal* `viewIn.contains("\"_id\":\"$teamId\"")`. The new `viewIn LIKE '%' || '"_id":"' || :teamId || '"' || '%'` interpolates `teamId` straight into a `LIKE` pattern, so any `_` or `%` in a team id becomes a wildcard. Since this decides *which team's discussions a user sees*, a team id containing `_` could match a post scoped to a different team whose id differs only at that position — cross-team leakage. Also the literal `"_id"` contains an unescaped `_` (matches any char), which is harmless only because `_id` is the sole such JSON key. Please add `ESCAPE '\'` and escape `_`/`%` in both the literal (`\_id`) and the interpolated `teamId` (mirror the `userIdPattern` escaping used elsewhere).

**2. Coverage regression.** The unit tests now mock `getTopLevelByTeam(any())` to return *already-filtered* lists, so they assert the repo passes through the mock — the actual matching logic (now in SQL) is no longer exercised anywhere. Since this is exactly the logic that moved, add an instrumented Room test (or a Robolectric-backed in-memory DB test) proving the query matches the old `matchesTeam` for the `viewableBy/viewableId` path, the `viewIn` path, and non-matching teams.

**Verdict:** Good optimization, but the unescaped `LIKE` in a team-visibility query plus the loss of test coverage warrant a fix before merge.

---

dogi:
**Re-review — rating: 87/100 (was 72, changes requested)**

Both flagged issues are genuinely resolved.

**1. Unescaped `LIKE` (the visibility-leak concern):** fixed correctly. The query now uses `viewIn LIKE :teamPattern ESCAPE '\'`, and `teamIdPattern` escapes `\`, `%`, and `_` in the interpolated `teamId` (same approach as `userIdPattern`). So a team id containing `_`/`%` can no longer act as a wildcard and match another team's posts.

**2. Coverage loss:** fixed — the new `NewsDaoTest` runs `getTopLevelByTeam` against a real in-memory Room DB and covers all the branches, including the two escaping cases that directly prove the fix: `team1_sub` does **not** match `team1Xsub`, and `team1%` does **not** match `team1X`. This is exactly the regression guard I asked for.

**Residual (non-blocking):** the literal `"_id"` in the pattern still contains an unescaped `_` (a wildcard on the key name), but that's harmless — `_id` is the only such key in `viewIn`, so it can't produce a false match. Optional: escape it as `\_id` for pedantic exactness.

**Verdict:** flag resolved, and well-tested — lifting changes requested → **ready**. Nice thorough follow-up.

---

Okuro3499 (approved):

PR #14988, opened 2026-07-24
courses: smoother steps inline resources coroutine scoping (fixes #15126)

### LINKED ISSUE #15126: Fix InlineResourceAdapter CoroutineScope leak

dogi:
Modified InlineResourceAdapter to use a centralized adapterScope (cancelled in onDetachedFromRecyclerView) instead of individual ViewHolder scopes to prevent coroutine leaks from recycled items. Verified with `./gradlew compileDefaultDebugKotlin`.

---
*PR created automatically by Jules for task [18258208815083400196](https://jules.google.com/task/18258208815083400196) started by @dogi*


dogi:
**Review — rating: 87/100**

This one is a *real* leak fix (unlike some sibling "leak" PRs). The old code gave every `ViewHolder` its own `CoroutineScope(SupervisorJob() + main)` and explicitly kept it alive across recycles — those scopes were never cancelled, so each leaked a `SupervisorJob` (and its captured references) for the adapter's lifetime.

**Strengths**
- Consolidating to a single `adapterScope` cancelled in `onDetachedFromRecyclerView()` actually reclaims the coroutine scope. And `adapterScope.cancel()` cancels every preview job launched on it, which is why dropping the old per-child `cancelPreviousPreviews()` loop in `onDetached` is safe (it's subsumed).
- Recreating the scope in `onAttachedToRecyclerView` only when `!adapterScope.isActive` correctly handles re-attach after a cancelling detach (a cancelled scope can't relaunch).
- Per-`previewJob` cancellation on recycle is preserved; all these callbacks and the `main` dispatcher keep `adapterScope` reassignment on the main thread, so no race.

**Concerns**
- Minor: if `bindPreview` were ever called while detached (scope cancelled, before re-attach), `adapterScope.launch{}` is a silent no-op. In practice binding only happens while attached, so it's fine — worth a one-line note.

**Verdict:** Correct, addresses an actual leak — mergeable.

---

Okuro3499 (dismissed):
```
 - File: CourseStepFragment.kt, onDestroyView() (lines 353-357). Problem: never disconnects inlineResourceAdapter from rvInlineResources, so                                        
  InlineResourceAdapter.onDetachedFromRecyclerView — the only place the new adapterScope.cancel() runs — never executes on the real view-teardown path (ViewPager2 swipe / fragment  
  view recreation), meaning the fix does not close the leak it claims to fix. Fix: either set fragmentCourseStepBinding.rvInlineResources.adapter = null in onDestroyView (which     
  triggers the existing onDetachedFromRecyclerView cleanup), or better, pass viewLifecycleOwner.lifecycleScope into InlineResourceAdapter instead of managing an adapter-owned scope 
  keyed off RecyclerView attach/detach, so cancellation is tied to the fragment's actual view lifecycle.  
```

dogi:
**Review — 86/100**

**Strengths**
- Fixes a genuine leak. The old code created a `CoroutineScope(SupervisorJob() + main)` **per ViewHolder** with a comment explicitly admitting it was "left alive across recycles" — so every holder's scope (and any long-running preview coroutine on it) outlived the RecyclerView with no cancellation. Replacing this with a single adapter-scoped `SupervisorJob` cancelled in `onDetachedFromRecyclerView` is the right model.
- The `onAttachedToRecyclerView` guard is correct and necessary: a cancelled `SupervisorJob` can never launch again, so recreating `adapterScope` when `!isActive` allows the adapter to work if it's re-attached (config change / fragment view recreation). 
- `CourseStepFragment.onDestroyView` setting `rvInlineResources.adapter = null` is the missing piece that guarantees `onDetachedFromRecyclerView` actually fires on fragment teardown — good that the fix includes the trigger, not just the handler.
- Per-holder `previewJob` cancellation on recycle is preserved, and removing the manual `for (child…) holder.cancelPreviousPreviews()` loop in `onDetached` is safe now that `adapterScope.cancel()` tears down all in-flight previews at once.

**Nit (non-blocking)**
- Adapters *can* be attached to more than one RecyclerView; if this adapter were ever shared, detaching from one would cancel `adapterScope` out from under the other. Not a real risk here (it's one adapter per course step), but the adapter-scope-cancelled-on-detach pattern assumes single attachment — worth a one-line comment noting that.

**Verdict:** Correct, well-scoped leak fix with the teardown trigger included and CI green — mergeable.

---

Okuro3499 (approved):

PR #14965, opened 2026-07-24
all: smoother settings guest users handling (fixes #14959)

ragilzakaria:
- Added blockGuestSwitches() in SettingFragment (SettingsActivity.kt).
- When a guest user views Settings, blockGuestSwitches() recursively scans all SwitchPreference items in the PreferenceScreen.
- Attaches an OnPreferenceChangeListener to every switch that pops up DialogUtils.guestDialog(requireContext()) (prompting them to become a member) and returns false to block the switch state from toggling.

[Screen_recording_20260724_214038.webm](https://github.com/user-attachments/assets/7cf38374-f838-4071-91cc-97aad8fc714a)

fixes #14959 

## Summary by CodeRabbit

* **New Features / Settings**
  * Added guest-specific restrictions in Settings.
  * Guest users can no longer toggle switch-based preferences; attempts show an explanatory dialog and the change is blocked.
* **Release / Versioning**
  * Updated the app version (version code/name) for this release.

PR #14993, opened 2026-07-24
all: smoother settings storage space organizing (fixes #14958)

### LINKED ISSUE #14958: merge storage breakdown and Free up disk space
```
They operate on the exact same file tree (FileUtils.getOlePath) and the same downstream call resourcesRepository.markResourcesAsNotOffline). "Free up disk space" is really just a blunter version of the "delete all" button that already exists inside Storage          
Breakdown's category detail view — same effect, different UI path (one runs via FreeSpaceWorker with no category granularity, the other lets you pick a category/selection first). That's genuine redundancy: one screen with breakdown + "delete all" / "delete category" covers both use cases, and you drop a whole feature entry point.
```

Okuro3499:
fixes #14958
[storage fix.webm](https://github.com/user-attachments/assets/5a4db45f-e898-4e9e-97ea-abb8481cbb52)

## Summary by CodeRabbit

* **New Features**
  * Added a text-size selection option in Settings.
  * Enhanced storage details with an available-space display and a “Free up space” action, including localized progress and completion summaries.
* **Bug Fixes**
  * Storage cleanup now uses lifecycle-aware WorkManager progress with a working cancel action and reloads storage after success, failure, or cancellation.
  * “Last synced” and storage breakdown summaries refresh more reliably.
  * Guest users are blocked from resetting app data and from changing restricted settings.
  * Removed the old “freeup space” setting entry in favor of the storage breakdown flow.


dogi:
**Review — rating: 80/100**

Genuinely nice consolidation (fixes #14958): the standalone `freeup_space` preference is removed and folded into the Storage Breakdown sheet with an available-space line + a "Free Up Space" button, and the `FreeSpaceWorker` observation logic is moved faithfully.

**Strengths**
- Real localization win: the previously **hardcoded** `"Deleting files... "` / `"Freed ... "` strings become `storage_deleting_progress` / `storage_freed_summary`, translated across ar/es/fr/ne/so. Format args line up (`%1$d` int + `%2$s` size; `%1$s` size + `%2$d` int).
- Verified all referenced resources exist: `AppThemeButton`, `AlertDialogTheme` (`styles.xml:94,185`), `storage_breakdown_summary`, `available_space_colon`. New binding ids (`availableSpaceText`, `freeUpSpaceButton`) are declared in the layout.
- Adding `loadStorage()` after a successful cleanup is a nice touch — the breakdown refreshes to reflect freed space.

**Concerns**
1. **Revert the `settings.gradle` change.** The diff prepends stray whitespace to the first line (`pluginManagement {` → `         pluginManagement {`). It's an accidental edit to a build file — non-breaking (Gradle tolerates leading whitespace), but it has no business in this PR and shouldn't merge. Please drop that hunk.
2. Minor: after removing the freeup handler, double-check `SettingsActivity` for a now-unused `Utilities` import (lint warning, not a failure).
3. Minor: the progress dialog's negative button label `"Cancel"` is still a hardcoded literal (carried over from the old code) — worth moving to a string resource while you're localizing the rest.

**Verdict:** Solid, well-localized refactor — just revert the unrelated `settings.gradle` hunk and it's good to go. Marking changes requested only for that one-line revert.

---

dogi:
Resolved the merge conflict against `master` and addressed the outstanding review feedback:

- `settings.gradle` no longer has the stray whitespace hunk — already clean on this branch.
- `Utilities` import in `SettingsActivity` is still used (retry-queue toasts), not dead.
- Localized the free-up-space "Cancel" button label (`R.string.cancel`).
- Storage totals now reload after a **failed or cancelled** cleanup, not just a successful one.
- The available-space text refreshes on every `loadStorage()` call instead of only once at view creation.
- `StorageBreakdownFragment` now emits a `FragmentResult` after any change (cleanup or category deletion) so `SettingsActivity`'s `storage_breakdown` summary refreshes instead of going stale.

Also fixed a compile break: the merge silently dropped the `DialogUtils` import (master added `guestDialog()` usages while this branch had removed the import for unrelated reasons — non-overlapping hunks merged without a conflict marker).

---

dogi:
Merged latest `master` again (another version bump conflict, same pattern as before) and addressed the new CodeRabbit findings:

- **Critical, fixed:** `onCreatePreferences()` runs before the fragment's view exists, so registering the `storage_breakdown` result listener against `viewLifecycleOwner` would throw `IllegalStateException` and crash Settings on open. Switched to `this` (the fragment) as the `LifecycleOwner`.
- **Fixed:** the free-up-space progress dialog is now dismissed in `onDestroyView()` so it can't outlive the fragment if the sheet is closed mid-cleanup.
- **Fixed:** failed/cancelled cleanup now also emits the storage-changed result, since `FreeSpaceWorker` can delete files before returning either terminal state.
- **Skipped:** moving the `FileUtils.availableOverTotalMemoryFormattedString` call behind the repository layer. That call already existed directly in `SettingsActivity`/`StorageBreakdownFragment` before this merge — it's a pre-existing pattern, not something introduced here, and reworking it is a larger architectural change outside the scope of a merge-conflict-resolution PR.

---

dogi:
Addressed the latest round of CodeRabbit findings (commit eac6540):

- **Fixed:** the free-up-space button now disables while a cleanup worker is running, so repeated taps can't enqueue overlapping `FreeSpaceWorker` runs racing on the same file tree.
- **Fixed:** `loadStorage()` now tracks its coroutine `Job` and cancels any in-flight scan before starting a new one, so a stale scan can't clobber a newer one's results when reload calls land close together (e.g. cleanup completion followed shortly by a category-detail return).
- **Skipped:** fully serializing/immutable-snapshotting `scanStorage()`, and persisting the cleanup work ID across view recreation so it can't outlive the fragment view. Both are correctly tagged "Heavy lift" — they'd mean a real redesign (a mutex/snapshot rewrite of the scan, and process-level work tracking) beyond what this merge-conflict-resolution PR should take on. The residual risk is bounded: if the fragment view is torn down mid-cleanup, the worker still finishes correctly, it just won't refresh the on-screen summary until the screen is reopened (which calls `loadStorage()` again).
- **Still intentionally skipped:** moving `FileUtils.availableOverTotalMemoryFormattedString` behind the repository layer, per my earlier comment — pre-existing pattern, out of scope here.

---

dogi:
Fair catch — my previous `loadJob?.cancel()` mitigation was incomplete (commit d4d88e5):

- **Fixed properly:** `scanStorage()` no longer mutates the shared `categories` objects while running on the IO dispatcher. It now builds a local, immutable snapshot (`ScanResult`) and returns it; that snapshot is only applied to `categories` after the scan completes, and only for the job that survives cancellation (the older job throws `CancellationException` at the `withContext` resume point before it gets a chance to apply stale results). Concurrent scans can no longer race on shared state, even if their synchronous file-walks briefly overlap.
- **Still skipped:** keeping the cleanup work observable/uniquely-tracked across view recreation. As before, this is correctly tagged "Heavy lift" — it needs process-level work tracking (e.g. persisting the work ID in a ViewModel/SharedPreferences and re-observing on recreation) rather than a fragment-scoped fix, which is out of scope for this merge-conflict-resolution PR. One clarification: view recreation doesn't actually auto-enqueue a second cleanup — a fresh fragment instance starts with the button enabled by default (matching the layout), and `freeUpSpace()` only runs on explicit user tap. The real residual gap is just that an in-flight cleanup's completion won't refresh the summary until the screen is reopened, which I called out in my earlier comment.

---

J-S-webskas (commented):
The description says "Guest users can no longer reset app data or change restricted settings." But guest user is able to reset storage:

https://github.com/user-attachments/assets/a1599eaa-014f-4cfa-be08-0277dad12410

But I wonder if this will be fixed once this gets merged: https://github.com/open-learning-exchange/myplanet/pull/14965

dogi:
@J-S-webskas good catch, thanks for flagging with the recording.

Checked PR #14965 (already merged) — it only guards `SwitchPreference` toggles (`blockGuestSwitches()`), it never touched the "Storage Breakdown" entry point, so it doesn't cover this.

Fixed in f3a0584: tapping "Storage Breakdown" now fetches the current user first and shows the guest dialog instead of opening the fragment when the user is a guest, mirroring exactly how `reset_app` already guards guests (`clearDataButtonInit()`). That blocks guest access to both "free up space" and the per-category delete flow, since both only live inside that screen.

---

J-S-webskas (commented):
https://github.com/user-attachments/assets/2ba4d838-dfe1-4338-acfa-c4e23dec2a43

The guest user cannot click clean storage. Nicely solved.

dogi:
**Review — rating: 80/100**

Verified the current head against `SettingsActivity`, `StorageBreakdownFragment`, `pref.xml`, the layout, and all six `strings.xml`. This is a genuinely good consolidation — the standalone "free up space" preference is folded into the storage bottom sheet, and several earlier review concerns are actually fixed in this head:
- `scanStorage()` now returns an immutable `ScanResult` and the category fields are applied on the main thread, with `loadJob?.cancel()` serializing reloads — resolves the stale/overlapping-scan race.
- `onDestroyView()` dismisses the progress dialog; FAILED/CANCELLED branches reload storage and emit `RESULT_KEY`; the summary refreshes via the fragment-result listener.
- Guest gating on `storage_breakdown` — @J-S-webskas verified on a later commit that guests can no longer trigger cleanup ("Nicely solved").
- Hardcoded "Deleting files…/Freed…" strings replaced with `storage_deleting_progress` / `storage_freed_summary`, localized across all six locales. 

**Concerns**
- `StorageBreakdownFragment.freeUpSpace()` enqueues a plain `OneTimeWorkRequest` (tagged, not unique). The button is disabled while running, but that enabled-state isn't preserved across view recreation — rotate/recreate the sheet mid-cleanup and the new view re-enables the button while the worker still runs, so a second tap enqueues an overlapping `FreeSpaceWorker` deleting the same tree. Use `enqueueUniqueWork(..., ExistingWorkPolicy.KEEP)` and re-observe by unique name so recreation reattaches instead of duplicating. (CodeRabbit's still-open Major.)
- Description mismatch: the summary claims "Added a text-size selection option in Settings," but there's no text-size change anywhere in this diff — please trim the description so it matches what's shipped.
- Minor/architecture: both UI classes read capacity directly via `FileUtils` rather than through the repository layer. Consistent with existing usage in this codebase, so non-blocking, but noted.

**Verdict:** Works, well-localized, and the key guest + threading issues are resolved and member-tested. The unique-work fix is the one thing worth doing before merge; treating it as a nit. Moving to `ready`. (Note: `mergeable_state` is `blocked` — still needs a maintainer approval to merge.)

---

dogi (approved):

PR #14994, opened 2026-07-24
teams: smoother voices list preparing  (fixes #15214)

### LINKED ISSUE #15214: Refactor VoicesAdapter submit preparation and preParseNews

dogi:
Refactor duplicate logic across `submitList` overloads in `VoicesAdapter` by creating a central `prepareSubmitList` method. Also optimizes the condition inside `preParseNews` to ensure it invokes URL parsing only when the raw images significantly differ. Verified compilation with `./gradlew compileDefaultDebugKotlin --no-daemon`.

---
*PR created automatically by Jules for task [11133752028759489629](https://jules.google.com/task/11133752028759489629) started by @dogi*

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed voice and news list updates so image data stays synchronized when content changes.
  * Cleared outdated image information when an item no longer contains images.
  * Ensured consistent processing for all list update paths.


dogi:
**Review — 85/100**

**Strengths**
- `prepareSubmitList(...)` extraction is a clean, behavior-preserving DRY win — both `submitList` overloads were running byte-identical prep (set `originalList`, prepend `parentNews`, `preParseNews` each item). Consolidating them removes the copy-paste and the risk of the two overloads drifting apart. `originalList` is still assigned on both paths since both route through the helper.
- The `preParseNews` rework is a genuine fix, not just a reshuffle: the old guard `(parsedImageUrls == null || rawImageUrls != current) && !current.isNullOrEmpty()` could never take the "images just got removed" path — when `currentImageUrls` became empty, the whole condition short-circuited to false and the previously-parsed URLs lingered as **stale** state on the item. The new `else` branch clears both `parsedImageUrls` and `rawImageUrls` to null in that case. Good catch.

**Concern (verified safe, noting for the record)**
- The new outer guard is `rawImageUrls != currentImageUrls`, so the equal-and-non-empty case no longer re-parses. That would be a regression *if* an item could ever have `rawImageUrls == currentImageUrls` (non-empty) while `parsedImageUrls == null`. It can't through this code: `rawImageUrls` is only ever assigned together with a successful `parsedImageUrls`, and the new `else` is the only place they're cleared — and it clears both. So the two fields move as a pair and the invariant holds. Just flagging that this invariant is now load-bearing; if anything ever nulls `parsedImageUrls` independently, this branch would skip the re-parse.

**Verdict:** Solid dedup plus a real stale-image-state fix, invariant preserved, CI green — mergeable.

---

Okuro3499 (approved):

PR #14995, opened 2026-07-24
teams: smoother voices post removing (fixes #15215)

### LINKED ISSUE #15215: fix: Short-circuit no-op Voices post removals

dogi:
Short-circuit no-op Voices post removals and update originalList to prevent resurrected posts.

---
*PR created automatically by Jules for task [1709448978374927788](https://jules.google.com/task/1709448978374927788) started by @dogi*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved post removal in voice discussions with more reliable control flow.
  * Prevented adapter data from being cleared when the target post isn’t found.
  * Ensured parent and reply removals stay consistent, while reply-count updates and UI refresh behavior continue to work correctly.


dogi:
**Review — rating: 85/100**

More than just a short-circuit — this fixes two real issues. Verified against `VoicesAdapter` (`ListAdapter<News, …>`, `originalList` a `var` at `:122`, restored via `submitList(originalList)` at `:509`).

**Strengths**
- **No-op short-circuit:** when `newsId` is neither in `currentList` nor the parent, the old code still fell through and decremented `replyCountCache[parentId]` — a spurious decrement on a removal that did nothing. The early `return` correctly avoids that.
- **`originalList` sync (the more important fix):** the old `removePost` only mutated the displayed list, never `originalList`. Since `originalList` is re-shown on filter-clear (`:509`), a removed reply would **reappear** after filtering. Removing it from `originalList` too fixes that.
- `it.id` (vs the old `it?.id`) is correct now that the element type is non-null `News`.

**Concerns (minor)**
- The parent-removal branch (`submitList(emptyList())`) doesn't update `originalList`, so it stays stale after removing the parent. Almost certainly harmless (removing the parent ends the thread view), but for symmetry you could clear it too.
- Decrementing `replyCountCache[parentId]` in the parent-removal case is a bit odd (you're removing the parent, not a reply) — but that's unchanged from the original, so no regression.

**Verdict:** Correct, fixes a genuine reappear-after-filter bug plus the spurious decrement — mergeable.

---

Okuro3499 (dismissed):
```
In the non-parent branch, submitList(updatedCurrentList) is called with a list derived from currentList, which already includes parentNews when it's non-null. Because    
submitList unconditionally re-prepends parentNews, this duplicates the parent post in the rendered list and corrupts originalList for all future calls in the same adapter instance
(e.g., updateParentNews).
```

dogi:
**Review — 85/100**

**Strengths — this fixes two real bugs:**
1. **Duplicated `parentNews`.** The old `removePost` did `submitList(snapshotList)` where `snapshotList = currentList (incl. parentNews) − item`. Since `submitList` re-prepends `parentNews`, the parent ended up in the list twice (and `originalList` got polluted with it, so a later `updateParentNews`→`submitList(originalList)` compounded it). Stripping the leading parent (`updatedCurrentList.drop(1)` when `[0].id == parentNews.id`) before submitting is the correct fix — the parent is re-added exactly once.
2. **No-op removals.** The `if (posInCurrent == -1 && !isParent) return` short-circuit prevents the old code from still decrementing `replyCountCache` when nothing was actually removed — the reply count no longer drifts on a missed removal. This matches the PR's stated intent.

**Verified safe**
- I checked whether the manual `originalList` update could be clobbered by `submitList` reassigning it. It can't cause data loss here: `VoicesAdapter` implements no `Filter`/`Filterable`, so `currentList == [parentNews] + originalList` and `listToSubmit` equals the manually-updated `originalList`. Consistent.

**Nits (non-blocking)**
- The manual `originalList` removal is effectively redundant — `submitList(listToSubmit)` reassigns `originalList` to the same value. Harmless, but could be dropped.
- The inline comment talks about preserving "the current filter state," but there's no filter in this adapter — slightly misleading wording.
- **Coordination:** #14994 also touches `VoicesAdapter` (the `submitList`/`prepareSubmitList` dedup + `preParseNews` image fix). Different methods, so no logical conflict — `removePost` here works with either `submitList` form — but whichever lands second will need a trivial rebase.

**Verdict:** Correct fix for the duplicate-parent and phantom-decrement bugs, verified no `originalList` regression, CI green — mergeable.

---

Okuro3499 (approved):

PR #14998, opened 2026-07-24
chat: smoother view modelling (fixes #15127)

### LINKED ISSUE #15127: Refactor ChatViewModel for robust coroutine management

dogi:
Replaced unbounded background coroutine invocations in ChatViewModel with explicit, tracked Jobs (loadDataJob and searchJob) that are cancelled before new instances launch. Additionally, updated `loadChatHistoryScreenData` to implement a structured retry loop with `RetryUtils` for database fetches, improving stability on transient errors. Verifications via `./gradlew compileDefaultDebugKotlin --no-daemon` confirmed no compilation regressions.

---
*PR created automatically by Jules for task [13892131002519826807](https://jules.google.com/task/13892131002519826807) started by @dogi*

## Summary by CodeRabbit

* **New Features**
  * Updated the app to version 0.62.28.

* **Bug Fixes**
  * Improved chat history loading with automatic retry support when temporary loading issues occur.
  * Prevented overlapping chat history requests and searches, helping results remain accurate and responsive when actions are repeated quickly.


dogi:
**Review — rating: 80/100**

Reasonable robustness refactor.

**Strengths**
- Tracking `loadDataJob`/`searchJob` and cancelling the previous before relaunching prevents overlapping loads and out-of-order state overwrites (an older search completing after a newer one) — the real win here.
- State is only committed on success (`result?.let { _screenData.value = ...; _filteredChats.value = allChats }`), so a failed load leaves the prior state intact instead of clobbering it.
- `RetryUtils.retry` exists with the used signature and uses cancellable `delay`, so a cancel during the back-off propagates.

**Concerns**
- **`RetryUtils.retry` swallows `CancellationException`** (`catch (e: Exception)` — pre-existing util, not this PR). Consequence for this PR: when `loadDataJob` is cancelled mid-`block()`, the cancellation is caught, `result=null`, and retry loops — it only actually stops at the next `delay()` (which re-throws), or on the *final* attempt returns `null` normally. So a cancelled load can run up to one extra `block()` iteration. Not harmful here (reads are idempotent and the new job's state wins), but consider hardening `RetryUtils` to rethrow `CancellationException`.
- On total failure (3 attempts → `null`), nothing is surfaced to the UI — the screen silently stays empty/stale. A one-shot error state would be friendlier.
- The retry `block` mutates fields (`allChats`, `precomputedChats`, `cachedUser`) as a side effect; fine because they're re-derived each attempt and the caches are reused, just note it reads a bit unusually.

**Verdict:** Sound cancel-and-retry pattern — mergeable; the `CancellationException` swallowing in `RetryUtils` is the thing worth a follow-up.

---

Okuro3499 (approved):

PR #15000, opened 2026-07-24
Add change detection before submitList in TeamFragment

dogi:
Added heuristic change detection logic before calling `submitList` in `TeamFragment`.
Verified functionality compilation using `./gradlew compileDefaultDebugKotlin --no-daemon` and tested for regressions using `./gradlew testDefaultDebugUnitTest --no-daemon`.

---
*PR created automatically by Jules for task [5860650445359497477](https://jules.google.com/task/5860650445359497477) started by @dogi*

## Summary by CodeRabbit

  * Improved team list updates by refreshing the display only when the received team data changes, reducing unnecessary updates during live emissions.

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/ui/teams/TeamFragment.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/teams/TeamFragment.kt
@@ -311,12 +311,7 @@ class TeamFragment : Fragment() {
 
     private fun observeTeamData() {
         collectLatestWhenStarted(viewModel.teamData) { teamDataList ->
-            val isMeaningfulChange = previousTeamDataList == null ||
-                previousTeamDataList?.size != teamDataList.size ||
-                previousTeamDataList?.firstOrNull()?._id != teamDataList.firstOrNull()?._id ||
-                previousTeamDataList?.lastOrNull()?._id != teamDataList.lastOrNull()?._id
-
-            if (isMeaningfulChange) {
+            if (previousTeamDataList != teamDataList) {
                 teamListAdapter.submitList(teamDataList)
                 previousTeamDataList = teamDataList
             }
--- a/app/src/main/java/org/ole/planet/myplanet/ui/teams/TeamFragment.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/teams/TeamFragment.kt
@@ -365,6 +365,7 @@ class TeamFragment : Fragment() {
 
     override fun onDestroyView() {
         _binding = null
+        previousTeamDataList = null
         super.onDestroyView()
     }
 }
--- a/app/src/main/java/org/ole/planet/myplanet/ui/teams/TeamFragment.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/teams/TeamFragment.kt
@@ -51,6 +51,7 @@ class TeamFragment : Fragment() {
     var user: UserEntity? = null
     private lateinit var teamListAdapter: TeamsAdapter
     private var conditionApplied: Boolean = false
+    private var previousTeamDataList: List<TeamDetails>? = null
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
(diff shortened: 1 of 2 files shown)


dogi:
**Review — rating: 48/100**

This optimization is both unnecessary and introduces a stale-UI bug.

**Why it's a regression**
- `isMeaningfulChange` only compares `size`, `first()._id`, and `last()._id`. So any change that preserves the count and both endpoints is **skipped** — e.g. a team in the *middle* of the list updating (member count, name, unread badge), or a reorder that keeps the same first/last team. `submitList` isn't called, `previousTeamDataList` isn't updated, and the UI silently shows stale data. For a team list where individual rows update, this is a real, common case.
- It's also redundant: `teamListAdapter` is a `ListAdapter`, so `submitList` already runs DiffUtil and rebinds *only* what actually changed. Submitting an equal list is cheap — DiffUtil detects no changes and does nothing. You don't need a manual gate to avoid unnecessary rebinds.

**If the goal was to avoid the background diff for identical emissions**
- Use full structural equality, not a heuristic: `if (previousTeamDataList != teamDataList) { submitList(...); previousTeamDataList = teamDataList }`. `TeamDetails` equality then guarantees you only skip when the list is genuinely identical. The size/first/last shortcut is what makes this lossy.
- If the real symptom was flicker/scroll-reset on identical lists, that points to a mis-configured DiffUtil callback (`areItemsTheSame`/`areContentsTheSame`) — fix that instead.

**Verdict:** Skips genuine updates → stale team list. Please drop the heuristic (let DiffUtil do its job) or switch to full-list equality. Changes requested.

---

dogi:
**Re-review — rating: 85/100 (was 48, changes requested)**

The flagged defect is genuinely fixed, not just changed.

**What I flagged:** the `isMeaningfulChange` heuristic (size + first/last `_id`) skipped `submitList` when a *middle* team's content changed → stale UI.

**What changed:** replaced with full structural equality — `if (previousTeamDataList != teamDataList)`. Now any content change anywhere in the list (relying on `TeamDetails` value equality) makes the lists unequal → `submitList` fires; only genuinely identical emissions are skipped. That's precisely the correct fix, and it eliminates the missed-update case.

**Note:** this remains a mild optimization over just always calling `submitList` (ListAdapter's DiffUtil already no-ops on equal lists), but it's now *correct*, so no objection.

**Verdict:** flag resolved — lifting changes requested → **ready**.

---

Okuro3499 (dismissed):
```
previousTeamDataList is scoped to the Fragment instance while teamListAdapter is recreated on every onViewCreated. After returning from TeamDetailFragment (or any other  
view recreation while the Fragment instance survives), the new adapter never receives its initial submitList, leaving the team list visibly empty.
```

dogi:
**Re-review — 82/100 → lifting to ready**

My earlier concern was that the change-detection was a *lossy* heuristic (comparing a subset like size/ids, which could skip a needed UI update). The current implementation resolves that — it now guards on **full structural inequality** of the whole list:

```kotlin
if (previousTeamDataList != teamDataList) {
    teamListAdapter.submitList(teamDataList)
    previousTeamDataList = teamDataList
}
```

Verified this is sound, not lossy:
- `TeamDetails` is a `data class` with all 12 display fields in the primary constructor (`name`, `status`, `visitCount`, `teamStatus`, `description`, …), and the nested `teamStatus: TeamStatus?` is *also* a `data class`. So `List.equals` compares every element structurally across every field — any change to anything shown triggers a re-`submitList`. Nothing displayed is excluded from the comparison.
- Skipping `submitList` when the lists are structurally equal is safe: the adapter's DiffUtil would find no item/content changes anyway, so the guard only avoids redundant work, never a real update.
- `showNoResultsMessage(...)` and `listContentDescription(...)` remain **outside** the guard, so the empty-state message and a11y description still update on every emission — no regression there.
- First emission: `previousTeamDataList` is null → `!= teamDataList` is true → submits. `onDestroyView` nulls the field so a recreated view re-submits fresh.

**Nit (non-blocking):** the full-list `equals` runs on the main thread per emission; fine for team-sized lists, but if this list ever grows large it's O(n) on main — worth keeping in mind.

**Verdict:** The lossy-heuristic concern is resolved — this is a correct, non-lossy skip-redundant-submit guard backed by structural equality. Ready.

---

Okuro3499 (dismissed):
```
TeamFragment.kt — observeTeamData(): No test covers the exact failure mode this branch fixed (adapter left empty after onDestroyView/recreate while the ViewModel/StateFlow is   
  retained). Given this function was rewritten 3 times in 3 days chasing this bug, land a regression test (e.g., a Robolectric test that recreates the fragment's view and asserts   
  the adapter's item count / list is repopulated) before merging, or explicitly document why one isn't feasible
```

dogi:
**Re-review — 82/100 (still ready)**

No functional change since my prior approval — the latest commits are a rebase plus a doc note ("Robolectric UI test unfeasible due to Hilt binding complexity"). The logic is the version I verified: full structural `previousTeamDataList != teamDataList` guard (non-lossy, since `TeamDetails`/`TeamStatus` are both data classes covering all display fields), `showNoResultsMessage`/`listContentDescription` still run every emission, and `onDestroyView` nulls the field so a recreated view re-submits. CI green (build default+lite, test default).

**Verdict:** Unchanged, correct skip-redundant-submit guard — remains ready.

---

Okuro3499 (changes requested):
No functional change since my prior approval

PR #15008, opened 2026-07-24
resources: smoother repository search dao title filtering (fixes #15186)

### LINKED ISSUE #15186: Move title search filtering to Room DAO

dogi:
Add the smallest DAO query needed to filter public library rows by normalized title tokens before returning to Kotlin. 
Keep the existing starts-with-before-contains ordering behavior in the repository. 
Route callers through ResourcesRepository; do not create use cases or duplicate data access in UI.

- Added `@RawQuery` method `filterByTitleNormal` to `MyLibraryDao` to allow dynamic title token filtering directly in SQLite.
- Refactored `ResourcesRepositoryImpl.search` to construct a dynamic `SimpleSQLiteQuery` with `LIKE` clauses for each normalized query token.
- Updated `ResourcesRepositoryImplTest` to mock the new raw query method, preserving test coverage for `startsWith` vs `contains` sorting.
- Verified changes by running `./gradlew testDefaultDebugUnitTest --tests "org.ole.planet.myplanet.repository.ResourcesRepositoryImplTest" --no-daemon`.

---
*PR created automatically by Jules for task [18227728069161183405](https://jules.google.com/task/18227728069161183405) started by @dogi*

## Summary by CodeRabbit

* **New Features**
  * Improved library search to match titles more accurately, including multi-word and partially matching queries.
  * Preserved relevant result ordering, prioritizing titles that start with the search term.

* **Bug Fixes**
  * Refined filtering for public and personal libraries, including course-library searches.
  * Improved handling of empty search queries and normalized title matching.


dogi:
**Review — rating: 82/100**

Solid move of the title filtering into a Room `@RawQuery`, and notably it gets the sensitive part right.

**Strengths**
- The `userId` filters use `LIKE ? ESCAPE '\'` with the already-escaping `userIdPattern(...)`, and the `!isMyCourseLib` branch correctly reproduces `getPublicNotUserPattern` semantics (`userId IS NULL OR userId NOT LIKE ?`). Good — this is the escaping that the sibling #14987 missed.
- Behavior matches the old in-memory filter: `titleNormal LIKE %token%` ≡ `titleNormal.contains(token)` since both operands are `normalizeText`-lowercased (LIKE's ASCII case-insensitivity is a no-op here), and NULL `titleNormal` is excluded in both (`NULL LIKE x` → not true, same as the old `?: return@filter false`).
- The empty-query short-circuit and the startsWith/contains bucketing are preserved (and still unit-tested via the mocked `filterByTitleNormal`).
- Real win: filtering now happens in SQL instead of loading the whole public/user set into memory.

**Concerns (nits)**
- Inconsistent escaping: the title tokens are bound as `"%${token}%"` **without** `ESCAPE`, and `normalizeText` doesn't strip `%`/`_` (it only lowercases + removes diacritics). So a query like `a_b` treats `_` as a wildcard. Harmless-ish for fuzzy search (just over-matches), but for consistency escape the token too (`ESCAPE '\'` + escape `%`/`_`/`\`).
- Coverage: the SQL WHERE (token matching) itself isn't exercised by the mocked tests — only the post-query bucketing is. An in-memory Room test over `filterByTitleNormal` would lock the token-matching down.

**Verdict:** Correct and well-escaped where it matters — mergeable; tidy up the title-token escaping for consistency.

---

Okuro3499 (dismissed):
```
normalizedQueryParts.forEach { ... bindArgs.add("%${token}%") } embeds the user's search token directly into a LIKE wildcard pattern without escaping %, _, or \. Since   
  titleNormal is derived from real (often filename-like) titles that can contain underscores, a search token containing _ or % will silently match unintended rows (or fail to match 
  intended ones), unlike the previous plain Kotlin contains() check which had no wildcard semantics.
```

dogi:
**Review — 86/100**

Verified against source and the DAO queries it replaces; CI green (build default+lite, test default).

**Strengths**
- **Injection-safe and correctly escaped** — the win this kind of dynamic-SQL PR usually gets wrong. Every dynamic value goes through a `?` placeholder + `bindArgs` (never string-concatenated), and each title token is escaped for `\`, `%`, `_` with a matching `ESCAPE '\'`, so user wildcards are treated as literals. This is exactly the LIKE-escaping discipline that's easy to miss.
- **Behavior-preserving base filtering — verified exactly.** The inline `WHERE` clauses match the three DAO queries they replace, byte for byte:
  - `getPublic()` = `isPrivate = 0` ✓
  - `getPublicForUserPattern` = `isPrivate = 0 AND userId LIKE ? ESCAPE '\'` ✓
  - `getPublicNotUserPattern` = `isPrivate = 0 AND (userId IS NULL OR userId NOT LIKE ? ESCAPE '\')` ✓
- The per-token `titleNormal.contains(token)` Kotlin filter becomes `titleNormal LIKE '%token%' ESCAPE '\'` AND-ed per token — equivalent on normalized (lowercased) text, and NULL `titleNormal` is excluded the same way (`?: return@filter false` ≈ `LIKE` on NULL → not matched). The starts-with-before-contains ordering stays in Kotlin, unchanged. The empty-query and `isMyCourseLib && userId == null → emptyList()` early paths are preserved.
- Net effect matches the stated goal: filtering happens in SQLite instead of loading all public rows into memory and filtering in Kotlin.

**Notes (non-blocking)**
- The updated unit tests mock `filterByTitleNormal(any())` to return pre-filtered rows, so they now only exercise the Kotlin **ordering** logic — the actual `WHERE`/`LIKE`/escaping is no longer covered by any test. Consider one Robolectric + in-memory-Room test that inserts rows and asserts the `LIKE`/escape behavior (e.g. a title containing a literal `%`), since that's the part with the real risk.
- `LIKE`'s case-insensitivity is ASCII-only; this is fine here because both sides come from `normalizeText` (lowercased), but it's worth a comment noting the equivalence depends on that.

**Verdict:** Correct, injection-safe, behavior-preserving push of title filtering into Room with proper escaping; CI green — ready. Would be stronger with one real-DB test over the raw query.

---

Okuro3499 (dismissed):
```
The SQL construction itself (ResourcesRepositoryImpl.kt:80-102) is correct on manual trace — bind args are parameterized (no injection risk), wildcard escaping mirrors the        
  existing userIdPattern() convention, and the isPrivate/userId/isMyCourseLib branching preserves the original semantics exactly. But ResourcesRepositoryImplTest.kt was edited to   
  mock myLibraryDao.filterByTitleNormal(any()) to return an already-filtered list, which means the tests no longer verify the thing that actually changed — the WHERE clause and     
  escaping logic have zero real coverage now.
```

dogi:
**Re-review — 88/100 (still ready, strengthened)**

The production code is unchanged from my earlier approval; the new commit adds exactly the coverage I'd flagged as missing. `search properly escapes wildcards in query` now captures the `SupportSQLiteQuery` via `slot`, binds it through a `SupportSQLiteProgram`, and asserts the escaped bind args for `"100% _real_ \deal"`:
- `%100\%%` (`%` → `\%`)
- `%\_real\_%` (`_` → `\_`)
- `%\\deal%` (`\` → `\\`)

plus the `LIKE ? ESCAPE '\'` clause is asserted in the captured SQL. That's a real test over the raw query's escaping — the one gap from last round — and it also confirms `normalizeText` preserves those characters (otherwise the assertions wouldn't hold). CI green (build default+lite, test default).

**Verdict:** Same behavior-preserving, injection-safe Room search as before, now with genuine escaping coverage — ready.

---

Okuro3499 (approved):

PR #15013, opened 2026-07-24
chat: smoother message model callback diffing (fixes #15102)

### LINKED ISSUE #15102: Fix ChatAdapter DiffUtil with proper ID

dogi:
Fix ChatAdapter DiffUtil with proper ID

- Added unique `id` field to `ChatMessage` with `UUID.randomUUID().toString()` as the default.
- Updated `ChatAdapter`'s `DiffUtils.itemCallback` to compare items correctly. `areItemsTheSame` now compares `old.id == new.id`, and `areContentsTheSame` compares `message`, `viewType`, and `source` fields.
- Verified changes by running `./gradlew testDefaultDebugUnitTest --no-daemon`.

---
*PR created automatically by Jules for task [16738819957941399655](https://jules.google.com/task/16738819957941399655) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -11,8 +11,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6182
-        versionName = "0.61.82"
+        versionCode = 6214
+        versionName = "0.62.14"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/model/ChatMessage.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/model/ChatMessage.kt
@@ -1,9 +1,12 @@
 package org.ole.planet.myplanet.model
 
+import java.util.UUID
+
 data class ChatMessage(
     val message: String,
     val viewType: Int,
-    val source: Int = 0
+    val source: Int = 0,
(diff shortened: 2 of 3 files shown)


dogi:
**Review — rating: 87/100**

Correct DiffUtil identity fix, and I verified the load-bearing assumption.

**Why it's correct**
- The old `areItemsTheSame` used `message == message && viewType == viewType` as identity, so two messages with identical text (e.g. the same query asked twice, or repeated bot responses) collided as "the same item" — classic DiffUtil misbehavior. A stable `id` is the right identity key.
- The UUID is stable across submits: `addQuery`/`addResponse`/`prependMessages` all do `currentList.toMutableList()` and retain the existing `ChatMessage` instances, only constructing new ones for genuinely new messages. So ids don't churn between emissions (which would have made this counterproductive). Good — that's the thing that had to be true.
- `areContentsTheSame` correctly switches from `old == new` to explicit `message`/`viewType`/`source` comparison (excluding the identity `id`), and I confirmed nothing else relies on `ChatMessage` value-equality — the only `==` uses are on Int constants.

**Concerns**
- Minor: `id` defaulting to `UUID.randomUUID()` means any code that ever *reconstructs* an equivalent `ChatMessage` (rather than reusing the instance) would get a new identity. Current call sites all retain instances, so it's fine — just keep that invariant in mind for future history-reload paths.

**Verdict:** Correct, fixes a real duplicate-message DiffUtil bug — mergeable.

---

Okuro3499 (approved):

PR #15033, opened 2026-07-24
all: smoother base adapter factoring (fixes #15201)

### LINKED ISSUE #15201: Standardize adapter initialization with factory pattern

dogi:
Standardizes adapter initialization through a Factory pattern and refactors boilerplate adapter callback initialization.

---
*PR created automatically by Jules for task [12324319417777012461](https://jules.google.com/task/12324319417777012461) started by @dogi*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved course list updates so selection and progress changes appear more accurately without unnecessary refreshes.
  * Improved resource list updates when resources are opened, available offline, or stored locally.
  * Ensured course and resource content changes refresh correctly when displayed details are updated.

* **Refactor**
  * Improved consistency and reliability when creating course and resource lists across the app.


dogi:
**Review — 61/100** (changes requested)

Verified against source; CI green — but green CI here is expected precisely *because* the new abstraction is inert.

**The good part**
- `DiffUtils.standardItemCallback(idSelector, contentSelector, payloadSelector)` is a genuinely nice helper, and both migrations are behavior-preserving:
  - `CoursesAdapter`: the `contentSelector` list `[courseTitle, description, gradeLevel, subjectLevel, createdDate, isMyCourse, numberOfSteps]` compares element-wise exactly like the old `&&` chain; `payloadSelector` logic is unchanged (only a stale comment was dropped).
  - `ResourcesAdapter`: `idSelector = { it.item.id ?: "" }` matches the old `oldItem.item.id == newItem.item.id` for all realistic cases, and with no `contentSelector` it falls back to `oldItem == newItem` — same as before.

**Blocking concern — the headline feature does nothing**
- `base/BaseAdapterFactory.kt` — `BaseAdapterFactory` / `DefaultBaseAdapterFactory` and the new `var adapterFactory: BaseAdapterFactory? = null` on `BaseRecyclerFragment` are **completely unused**. Grepping the whole tree, nothing calls `createResourcesAdapter` / `createCoursesAdapter`, nothing instantiates `DefaultBaseAdapterFactory`, and nothing ever assigns or reads `adapterFactory`. So the PR titled "Standardize adapter initialization with factory pattern" doesn't actually standardize any adapter initialization — the adapters are still constructed directly at their call sites. The factory is dead scaffolding.

**What I'd want before this is ready**
1. Either **wire it up** — have `BaseRecyclerFragment` (and the resources/courses fragments) create their adapters through `adapterFactory ?: DefaultBaseAdapterFactory()` so the pattern is real and testable — or
2. **Split the PR**: land the `DiffUtils.standardItemCallback` helper + the two migrations on their own (that part is ready), and drop the unused factory until it has a consumer.

As it stands, merging adds public API surface (a nullable mutable field on the base fragment, plus an interface + impl) that nothing exercises.

**Verdict:** Good DiffUtils refactor wrapped around an unused factory abstraction — wire the factory in or split it out.

---

dogi:
**Re-review — 82/100 → lifting to ready**

My earlier changes-requested flag was that the "factory pattern" was inert — `BaseAdapterFactory`/`DefaultBaseAdapterFactory` and the `adapterFactory` field existed but nothing constructed adapters through them. Commit `92296d7` fixes exactly that:

- **`CoursesFragment`** now does `val factory = adapterFactory ?: DefaultBaseAdapterFactory(); adapterCourses = factory.createCoursesAdapter(...)`.
- **`ResourcesFragment`** now does the same via `factory.createResourcesAdapter(...)`.

So the factory is genuinely wired in, and the `adapterFactory` seam on `BaseRecyclerFragment` is now consumable (a test/variant can inject a fake factory before the fragment builds its adapter). The headline deliverable is real now, not dead scaffolding.

Verified it stayed behavior-preserving:
- `createCoursesAdapter(context, HashMap(), isGuest, isMyCourseLib)` constructs `CoursesAdapter` with the same four args as the old direct call.
- `createResourcesAdapter(context, isGuest, emptySet(), currentUserName = user?.name, onEditClick = { … })` matches the old `ResourcesAdapter(...)` args exactly.
- The `DiffUtils.standardItemCallback` migration is unchanged from my first pass and remains behavior-preserving (Courses `contentSelector` list == old `&&` chain; Resources `idSelector = { it.item.id ?: "" }` + default `oldItem == newItem` fallback).
- CI green (build default+lite, test default).

**Nit (non-blocking):** `adapterFactory` is a public `var … = null` on the base fragment; that's a fine test seam, but if it's only ever meant to be swapped in tests, a clearer name/comment (or wiring it through DI) would signal intent. Not a blocker.

**Verdict:** The factory is now actually used by both fragments and the refactor is behavior-preserving with CI green — moving from changes requested to ready.

---

Okuro3499 (approved):

PR #15034, opened 2026-07-24
courses: smoother progress row callback diffing (fixes #15152)

### LINKED ISSUE #15152: Introduce typed CoursesProgressRow for course progress adapter
`CoursesProgressAdapter` is a `ListAdapter<JsonObject, …>`, so raw `JsonObject` items travel all the way into the view holder. Every field read goes through `item.asJsonObject["…"]`, with `has()` checks standing in for nullability and no compile-time guarantee that any key exists.

The `DiffUtil` callback pays the highest price: `areContentsTheSame` cannot compare items directly, so a `getCourseProgressComparisonData` helper unpacks five fields into a `List<Any?>` on both sides just to get a usable equality check.

Mapping the JSON into a typed `CoursesProgressRow` at the fragment boundary — where the data arrives from the view model — would push untyped access out of the adapter, let `areContentsTheSame` become a plain `old == new`, and delete the comparison helper entirely.

dogi:
Introduced a typed data class `CoursesProgressRow` to replace the raw `JsonObject` items being passed directly to `CoursesProgressAdapter`. This change moves the boundary of untyped JSON mapping closer to the repository layer (inside `CoursesProgressFragment`), bringing compile-time safety, simplifed `DiffUtil` equality checks, and better readability to the list adapter implementation. Tests have been updated and are passing.

---
*PR created automatically by Jules for task [16165703126008840650](https://jules.google.com/task/16165703126008840650) started by @dogi*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved course progress rendering by switching to structured progress data.
  * Fixed progress details, including mistakes and optional step-by-step mistake values (including missing/empty cases).
  * Ensured item identity and list updates behave consistently, keeping navigation intact when selecting a course.
* **Tests**
  * Updated adapter tests to validate correct display for multiple, limited, and unavailable step-mistake data.
* **Chores**
  * Bumped app version to **0.62.52** (**6252**).


dogi:
**Review — 83/100**

**Strengths**
- Replacing the raw `JsonObject` in `CoursesProgressAdapter` with a typed `CoursesProgressRow` is a real quality win — the adapter no longer does `item.asJsonObject["progress"].asJsonObject["current"].asInt` gymnastics, and the mapping is field-by-field behavior-preserving:
  - progress guard: old `has("progress")` → new `progressCurrent != null && progressMax != null`, and the fragment nulls both when `progress` is absent (`obj.getAsJsonObject("progress")?...`), so the click-listener/description branch fires under the same condition.
  - `stepMistake`: `keySet()`→`keys`, `[k].asInt`→`[k]`, empty check preserved.
  - `DIFF_CALLBACK`: `getCourseProgressComparisonData` list-compare replaced by data-class `==`, which covers the same fields (plus `courseId`, already the identity — harmless).
- Tests migrated cleanly to construct `CoursesProgressRow` directly, and the holder-recycling test still exercises the many/few/none stepMistake path.

**Nits (non-blocking)**
- `CoursesProgressFragment.kt` — `Gson()` is constructed once per row inside the `.map { }`. Hoist a single `Gson()` (or reuse an injected one) outside the loop.
- `obj.get("courseId").asString` / `obj.get("courseName").asString` are non-null accesses; a doc missing either key will NPE during mapping. The old code had the same exposure in `onBind`/`onClick`, so it's not a regression, but since you're now centralizing parsing this is the natural place to guard (`obj.get("courseId")?.asString ?: return@map null` + `filterNotNull`, or similar).
- Import ordering: the new `com.google.gson.*` imports landed below the project import — minor, ktlint-style.

**Verdict:** Good, behavior-preserving move to a typed row model with tests updated; CI green — mergeable. The per-row `Gson()` is the only thing I'd tidy.

---

Okuro3499 (approved):

PR #15049, opened 2026-07-26
chat: smoother detail ai providing (fixes #15144)

### LINKED ISSUE #15144: 🧹 [Code Health] Extract deeply nested logic into a helper function in ChatDetailFragment

dogi:
🎯 **What:** Extracted a deeply nested for-loop in `ChatDetailFragment.kt` that iterated through AI provider buttons into a dedicated `updateSelectedAiProvider` helper function.
💡 **Why:** This drastically reduces the nesting depth within the `sharedViewModel.selectedAiProvider.collect` coroutine block, making the main logic much easier to read and maintain.
✅ **Verification:** Verified that the patch was applied cleanly. Ran the relevant `ChatDetailFragmentInjectionTest` which passed successfully. Confirmed that the logic and side-effects remain identical.
✨ **Result:** A flatter, cleaner `ChatDetailFragment.kt` with improved code health.

---
*PR created automatically by Jules for task [7269134348420409139](https://jules.google.com/task/7269134348420409139) started by @dogi*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved AI provider selection so the matching provider and model are applied consistently when the selection changes.
  * Added a fallback model when no specific model is available.

* **Refactor**
  * Streamlined the provider-selection logic for more consistent behavior.


J-S-webskas (approved):
Very nice update for readability by extracting heavy 4-layer nested structure into a helper function.

dogi:
**Review — 88/100**

Verified against source.

**Strengths**
- Textbook extract-method: the `aiTableRow` scan loop moves verbatim into `updateSelectedAiProvider(selectedAiProvider)` — the `isNotEmpty()` guard, the `for` loop, the `Button` + `equals(ignoreCase=true)` match, the `getModelsMap()[...] ?: "default-model"` lookup, `selectAI(...)`, and the `break` are all unchanged. The `collect { }` block is now two lines and reads cleanly. Single call site, no signature or capture changes, so behavior is identical.

**Nit (non-blocking, pre-existing)**
- `selectAI(view, "$selectedAiProvider", modelName)` still uses string interpolation `"$selectedAiProvider"` where `selectedAiProvider` (or the matched button's own text) would do — carried over from the original, not introduced here.

**Verdict:** Clean, behavior-preserving readability refactor — mergeable.

---

Okuro3499 (approved):

PR #15050, opened 2026-07-26
sync: smoother photo uploading (fixes #15145)

### LINKED ISSUE #15145: ⚡ Optimize PhotoUploader O(N^2) Lookup

dogi:
💡 **What:** Replaced the O(N^2) `.find { it.photoId == photo.id }` lookup inside the `successfulUploads` loop with an O(1) map lookup `photosMap = submissionsRepository.getPhotosByIds(photoIds).associateBy { it.id }`.
🎯 **Why:** The original implementation called a database query which correctly fetched a chunk of records, but then iterated through `photos` and called a linear `.find()` for every single item on the list. This resulted in O(N^2) performance for the batch. Resolving this via `associateBy` optimizes this logic to O(N + M) and fixes the CPU bottleneck without affecting the DB batching mechanism or hitting SQLite limit errors.
📊 **Measured Improvement:** The benchmark test revealed that MockK overhead dominates unit tests for this issue; however, algorithmic complexity dictates that improving the inner loop from O(N^2) to O(N) lookup prevents significant CPU hangs when dealing with large chunk sizes.

---
*PR created automatically by Jules for task [8233212350498428726](https://jules.google.com/task/8233212350498428726) started by @dogi*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved photo upload processing to reliably match successfully uploaded files with their stored photo records.
  * Preserved existing attachment behavior when matching photo information is available.


dogi:
**Review — rating: 85/100**

Legit O(N·M)→O(N) fix — the per-photo `successfulUploads.find { ... }` inside `forEach` becomes a `photosMap` keyed by `id` with O(1) lookups.

**Correctness**
- For the realistic case (each `successfulUploads` entry has a distinct `photoId`), the set of `uploadAttachment(photo.photoLocation, …, uploadInfo.id, uploadInfo.rev)` calls is identical to before. Missing photos (repo returns fewer than requested) are still skipped via the `photo != null` guard.

**One behavioral nuance to be aware of**
- The iteration domain flips from `photos` to `successfulUploads`. If two successful uploads ever shared the same `photoId`, the old code processed that photo **once** (first matching upload, via `find`), whereas the new code processes **each upload** (its own `id`/`rev`). The new behavior is arguably the more correct one (every successful upload gets its attachment), but if `photoId` is ever intentionally shared this is a semantic change. Given uploads appear to be 1:1 with photos, this is almost certainly a non-issue — flagging only so it's a conscious choice.

**Verdict:** Correct and effective optimization — mergeable.

---

Okuro3499 (approved):

PR #15051, opened 2026-07-26
dashboard: smoother bell reminding (fixes #15146)

### LINKED ISSUE #15146: ⚡ fix N+1 query issue in BellDashboardFragment

dogi:
💡 **What:** 
Extracted the database query `submissionsRepository.getSubmissionsByIds` from inside the `remindersToShow` loop in `BellDashboardFragment.kt`. We now flatten all `surveyIds`, perform a single batch query, construct a map `submissionsById`, and then iterate through `remindersToShow` using the pre-fetched map.

🎯 **Why:** 
The previous implementation suffered from an N+1 query problem, making a separate database call for each element in `remindersToShow`. This optimization reduces the number of database queries from N to 1, significantly improving performance and reducing unnecessary database load and latency.

📊 **Measured Improvement:** 
Reduced database queries from N (number of items in `remindersToShow`) to exactly 1 query by batch fetching all required submissions upfront and doing an O(1) in-memory lookup. This completely eliminates the N+1 database call pattern.

---
*PR created automatically by Jules for task [2630096853362460464](https://jules.google.com/task/2630096853362460464) started by @dogi*


dogi:
**Review — rating: 88/100**

Real N+1 elimination, and behavior-preserving.

**Verified**
- The loop previously called `submissionsRepository.getSubmissionsByIds(surveyIdList)` once per reminder (N queries). Now a single `getSubmissionsByIds(allSurveyIds)` over the `distinct()` union of all ids, with the per-reminder logic reading from the shared `submissionsById` map. Result parity holds: lookups are by `it.id`, and the union superset contains every id each reminder needs.
- Split/blank-filter logic matches the original (`it.split(",").filter { it.isNotBlank() }`), and the `allSurveyIds.isEmpty()` early return is equivalent to the old "every `surveyIdList` empty → loop does nothing" path.
- Duplicate ids across reminders are handled (`distinct()` for the query, `associateBy` idempotent for identical submissions).

**Concerns**
- None. Genuinely fewer DB round-trips with identical output.

**Verdict:** Correct, effective N+1 fix — mergeable.

---

Okuro3499 (approved):

PR #15054, opened 2026-07-26
🧪 Add tests for ConfigurationsRepository CheckVersionCallback

dogi:
🎯 **What:** The testing gap for the `onUpdateAvailable` function inside the `ConfigurationsRepository.CheckVersionCallback` interface has been addressed.
📊 **Coverage:** A new test suite `ConfigurationsRepositoryTest.kt` was added to verify the default callback implementation behaviors, including `onUpdateAvailable`, `onCheckingVersion` (which has a default empty implementation), and `onError`.
✨ **Result:** Test coverage improved for the callback methods that govern the sync flow when dealing with server availability and apk upgrades.

---
*PR created automatically by Jules for task [14896092677816299148](https://jules.google.com/task/14896092677816299148) started by @dogi*

## Summary by CodeRabbit

- **Tests**
  - Expanded coverage for version-check caching, confirming network requests are skipped within 24 hours.
  - Added coverage to verify version information refreshes after the 24-hour cache period.
  - Improved coroutine-aware verification for database cleanup behavior.
  - Added test cleanup to prevent mock state from affecting other tests.


J-S-webskas (approved):
Good testing for whether the call back is truly working.

dogi:
**Review — 55/100** (changes requested)

Verified against `ConfigurationsRepository.CheckVersionCallback` (interface at `ConfigurationsRepository.kt:15`, with `onCheckingVersion() {}` as an empty default at line 17).

**The problem: these tests don't test the code under review.**
All three cases construct an anonymous `CheckVersionCallback`, call one of its methods directly, and then assert that the method you just called ran and stored its arguments:
- `test CheckVersionCallback onUpdateAvailable` — calls `callback.onUpdateAvailable(testInfo, true)` and asserts the flags/args it just passed came back. That's verifying Kotlin dispatches a virtual call to the body you defined in the same test — a language guarantee, not app behavior.
- `test onCheckingVersion default implementation` — calls the empty default and asserts nothing (only that it doesn't throw). It's an empty `{}`; there's nothing to exercise.
- `test onError` — same tautology as case 1.

None of them invoke `ConfigurationsRepository.checkVersion(callback, spm)`, which is where the real, testable logic lives: the 24-hour version-check cache (`TimeProvider`), and the min/latest apk-code evaluation that decides whether `onUpdateAvailable(cancelable=…)` vs `onError(blockSync=…)` fires. The `MyPlanet().apply { minapkcode = 10; latestapkcode = 12 }` fixture hints at exactly that logic — but it's never fed through `checkVersion`, so those fields are unused.

**What would make this valuable**
- Mock `checkVersion`'s dependencies (apiInterface/version endpoint, `SharedPrefManager`, `TimeProvider`) and assert that, given a server `MyPlanet` with specific `minapkcode`/`latestapkcode` vs the installed version code, `checkVersion` calls `onUpdateAvailable` with the correct `cancelable`, or `onError` with the correct `blockSync`.
- Assert the 24-hour cache: two calls within the window should hit the cache (single network call); a call past the cutoff should re-fetch.

**Note — likely overlap:** #15057 ("Add test for onCheckingVersion in ConfigurationsRepository") appears to cover the same `onCheckingVersion` default. Worth consolidating so the two PRs don't both land near-identical trivial coverage.

**Verdict:** Compiles and passes, but asserts interface semantics rather than repository behavior — please redirect at `checkVersion(...)` itself.

---

dogi:

This is a complete turnaround from my earlier changes-requested. The tautological interface tests are gone; the new commit tests the *actual* `checkVersion` logic in `ConfigurationsRepositoryImplTest.kt`, covering exactly the two paths I asked for:

- **`checkVersion caches network call within 24 hours`** — sets `last_version_check_timestamp` to `now` (inside the window) and asserts `onCheckingVersion()` + `onUpdateAvailable(any(), true)` fire **and** `coVerify(exactly = 0) { apiInterface.checkVersion(...) }` / `getApkVersion(...)`. That's the cache-hit-skips-network assertion that was missing.
- **`checkVersion fetches from network after 24 hours`** — sets the timestamp 25h back, stubs the network responses, and asserts both endpoints are hit exactly once plus `onUpdateAvailable(fetchedPlanet, false)`. The `cancelable = false` is correct: fetched `minapkcode = 6` > current version `5`, so it's a mandatory update. Good that the test pins the mandatory-vs-optional distinction, not just "an update happened."
- Both drive time through the injected `TestTimeProvider`, using the real `TimeProvider` seam rather than hardcoding — exactly the approach I suggested.
- Also correctly switches the `clearAllData` assertion from `verify` to `coVerify` for the suspend call.

**Nits (non-blocking)**
- Heavy `mockkObject` on `UrlUtils`/`VersionUtils`/`Constants`/`NetworkUtils` — necessary given `checkVersion`'s static deps, and `unmockkAll()` in `@After` cleans up. Fine.
- `delay(100)` + `verify(timeout = 1000)` mixes virtual-time and wall-clock waits; it passes reliably here but the `timeout` polling is the load-bearing part — worth a comment so nobody "simplifies" the delay away.
- Side effect: since the tests moved into `ConfigurationsRepositoryImplTest.kt`, this no longer creates `ConfigurationsRepositoryTest.kt`, so the earlier same-file collision with #15057 is gone — #15057 (single empty-default test) is now purely redundant and should just be closed.

**Verdict:** Real, behavior-accurate coverage of both cache branches with correct callback assertions; CI green — moving from changes requested to ready.

---

Okuro3499 (approved):

PR #15058, opened 2026-07-26
🧪 [Testing Improvement] Add test for getTasksFlow testing (fixes #15196)

### LINKED ISSUE #15196: 🧪 [Testing Improvement] Add test for getTasksFlow

dogi:
🎯 **What:** Added a unit test for the previously untested public function `getTasksFlow` in `TeamsRepositoryImplTest.kt`.
📊 **Coverage:** Tests that `getTasksFlow` correctly delegates to `teamTaskDao.getOpenTasksForUser` and emits the correct `TeamTask` list.
✨ **Result:** Improved test coverage and reliability for `TeamsRepository`.

---
*PR created automatically by Jules for task [11085769668370598782](https://jules.google.com/task/11085769668370598782) started by @dogi*


J-S-webskas (approved):
Good addition of unit testing for getTasksFlow to test if query + query runner gets caled and returns the correct result.

dogi:
**Review — 80/100**

Verified against `TeamsRepositoryImpl.getTasksFlow` (line 78); CI passing.

**Strengths**
- Correct delegation test: `getTasksFlow(userId)` is a straight pass-through (`return teamTaskDao.getOpenTasksForUser(userId)` — no mapping/detach), so `resultFlow.first()` yields the *same* list reference the DAO returned. That's why `assertEquals(mockTasks, resultList)` holds even though `TeamTask` is a plain class with identity `equals` (line 11) — the instances are literally the same, so this isn't accidentally relying on structural equality. Good that the test happens to be robust here.
- Uses `every` (not `coEvery`) for the non-suspend Flow getter — right call.

**Notes (non-blocking)**
- Because the impl is a one-line delegate, the assertion is really "the DAO's flow is returned unchanged". That's a legitimate contract to pin, but the test won't catch much beyond a wiring change. If you want it to bite harder, stub with two emissions (`flowOf(listA, listB)`) and assert both propagate, or `coVerify` the exact `userId` was passed through.
- **Shared-file coordination:** this appends to `TeamsRepositoryImplTest.kt`, the same file #15053 extends (its `getMyTeamDetailsFlow` test) — both add imports in the same block and a method at the end. Whichever merges second will need a trivial rebase to resolve the overlap. Not a defect, just a merge-order heads-up.

**Verdict:** Small but correct delegation test, behavior verified against source; CI green — mergeable (mind the rebase against #15053).

---

Okuro3499 (approved):

PR #15072, opened 2026-07-26
all: smoother notifications read view modelling (fixes #15153)

### LINKED ISSUE #15153: 🧹 Refactor: Extract notification read state mapping logic in NotificationsViewModel
`NotificationsViewModel` marks notifications as read in three places — `markSelectedAsRead`, `markAsRead`, and `markAllAsRead` — and each one re-implements the same inline `map` lambda:

```kotlin
currentList.map { notif ->
    if (notif.id in markedIds && !notif.isRead) notif.copy(isRead = true) else notif
}
```

The three copies are nested inside `update` blocks and differ only in whether they match a single id or a set of ids, so the shared intent is buried in boilerplate and any change to the read-state rule has to be made three times.

Two private `List<Notification>.markAsRead` extension functions — one taking an id, one taking a set — would state the intent once at the call site and keep the mutation rule in a single place.

dogi:
🎯 **What:** Extracted repetitive mapping logic for marking notifications as read into two new extension functions on `List<Notification>` inside `NotificationsViewModel`.
💡 **Why:** The previous implementation used deeply nested, duplicated inline `map` lambdas across `markSelectedAsRead`, `markAsRead`, and `markAllAsRead`. Encapsulating this logic into extension functions reduces boilerplate, improves readability, and centralizes the state mutation logic.
✅ **Verification:** Verified the refactor locally by running the `NotificationsViewModelTest` test suite, which passed successfully, confirming no regressions in behavior.
✨ **Result:** A cleaner, more maintainable `NotificationsViewModel` with DRY code for updating notification read states.

---
*PR created automatically by Jules for task [12998965415918538570](https://jules.google.com/task/12998965415918538570) started by @dogi*


dogi:
**Review — rating: 88/100**

DRY extraction of the read-state mapping into two `List<Notification>.markAsRead` overloads (single id / id set). Checked all three call sites.

**Strengths**
- The two set-based call sites are byte-for-byte equivalent (both already had the `&& !it.isRead` guard).
- The single-id call site (`NotificationsViewModel.kt:155`) previously lacked that guard, so the new helper adds `&& !it.isRead`. This is safe: `Notification` is a data class, so `copy(isRead = true)` on an already-read item is `.equals()` to the original — the emitted list is unchanged, and it avoids an allocation. Behavior preserved.

**Concerns**
- None. Genuinely reduces three near-identical lambdas to one named intent.

**Verdict:** Behavior-preserving readability refactor — mergeable.

---

Okuro3499 (approved):

PR #15073, opened 2026-07-26
sync: smoother upload bulk voices managing (fixes #15206)

### LINKED ISSUE #15206: ⚡ Resolve N+1 query in UploadManager's uploadNews via _bulk_docs

dogi:
💡 **What:** Refactored `uploadNews` in `UploadManager.kt` to use CouchDB's `_bulk_docs` endpoint for uploading News documents, and added `postDocArray` / `postUploadArray` definitions to `ApiInterface.kt` and `UploadRepository.kt` to handle the JSON array responses from this bulk endpoint.

🎯 **Why:** To fix a severe N+1 query issue. Previously, after a batch's image attachments were sequentially uploaded, each news document was individually POSTed or PUT to the server within the `batch.forEach` loop. This resulted in an excessive number of network calls directly proportional to the number of news items.

📊 **Measured Improvement:** The number of network requests generated by news document uploads has been reduced from $O(N)$ (where $N$ is the number of news items in a batch) to exactly $O(1)$ per batch. This significantly reduces overall latency and connection overhead during the `uploadNews` operation. Note: Individual image binary uploads must still be processed sequentially per item due to their attachment requirement, but the primary JSON documents are now fully batched.

---
*PR created automatically by Jules for task [15799050164924284932](https://jules.google.com/task/15799050164924284932) started by @dogi*


dogi:
**Review — rating: 84/100**

Verified this is a real optimization, not a no-op. The per-item `postUpload`/`putUpload` calls are collapsed into a single `POST /news/_bulk_docs` per batch.

**Why it's correct (the part that could have been a regression)**
- `_bulk_docs` only updates an existing doc if the doc body carries `_id` **and** `_rev`. `serializeNews` embeds both when present (`VoicesRepositoryImpl.kt:447-448`), so updates still update (not create duplicates) and creates still create. The old create/update split via URL is faithfully reproduced.
- `${UrlUtils.getUrl()}/news/_bulk_docs` targets the right endpoint — `news` is the database, matching the old `POST /news` and `PUT /news/{_id}`.
- `news.newsJson` is the same reference that gets mutated with `message`/`images` before being added to `bulkDocsArray`, so the retry payload is the fully-populated doc, same as before.

**Concerns (all minor, non-blocking)**
- `UploadManager.kt` ~498: the per-item error branch calls `queueNewsRetry(news, ..., response.code(), ...)` where `response.code()` is the **bulk** HTTP status (200) even though the individual doc failed (e.g. a conflict returned as `{"error":...}` in the array). Since `queueNewsRetry` only retries on `httpCode >= 500`, per-doc conflicts are dropped — this matches the old behavior (a 409 PUT wasn't retried either), so it's parity, but worth a comment so it isn't mistaken for a retry path.
- Response/request index alignment (`processedNews[i]`) relies on CouchDB returning results in submission order and equal length — which it guarantees — but if the body is ever shorter, trailing items are silently neither marked uploaded nor retried. A `size()` guard/log would harden it.
- Overlaps conceptually with #15071 (team sync `_bulk_docs`) — both introduce bulk plumbing on `ApiInterface`. Not a duplicate (different method/endpoint), but please land them together so `postDocArray`/`postUploadArray` isn't defined twice.

**Verdict:** Sound bulk-write conversion with preserved create/update/retry semantics — mergeable after a quick look at the per-item error-code nit.

---

Okuro3499 (dismissed):
```
 - File/function: UploadManager.kt, uploadNews(), line 502–504 (feeding into queueNewsRetry at line 544).                                                                           
  - Problem: When a _bulk_docs call succeeds overall (HTTP 201) but an individual doc in the response array has an "error" field (e.g. CouchDB conflict or a transient per-item      
  failure), the code calls queueNewsRetry(news, news.newsJson, response.code(), ...) — passing the overall response code (201), not anything reflecting that item's actual failure.  
  Inside queueNewsRetry, the gate httpCode != null && httpCode >= 500 then evaluates false, so the function returns before queuing anything. The failed item is silently dropped: not
  marked uploaded, not retried, not logged. This is worse than the pre-existing per-item POST/PUT path, where a genuinely transient 5xx on an individual call did get queued for     
  retry — that signal no longer exists because CouchDB's bulk error objects ({"error":"conflict","reason":"..."}) carry no numeric status code.                                      
  - Fix: Don't route per-item bulk errors through the httpCode >= 500 retryability check at all — treat any itemResponse.has("error") as retryable directly (e.g. add an overload of 
  queueNewsRetry or a parameter that forces retryable = true for the bulk per-item path), or map known CouchDB reasons ("conflict" → treat as retryable/non-retryable per your       
  conflict-resolution policy) instead of relying on response.code().   
```

dogi:
**Review — 84/100** (correct; shares infra with #15071 — sequence them)

**Strengths**
- Same N+1→1 win for news: images are still uploaded per-item first, then the news docs go out as one `POST /news/_bulk_docs` instead of a `POST`/`PUT` per news.
- **Index-based matching is actually safe here.** `processedNews[i]` pairs with `bulkDocsArray[i]`, and CouchDB `_bulk_docs` returns results in input order, so `responseBody.get(i)` lines up. Critically, a news item whose *image* upload throws is caught in the first loop and excluded from **both** `bulkDocsArray` and `processedNews`, so the two lists stay aligned — the index mapping can't drift.
- Correct bulk error handling: per-doc `error` → `queueNewsRetry` (with the `error` reason and the right POST/PUT based on `_id`), whole-response failure → retry all processed, exception → retry all processed. `successfulUpdates` → DB update path is unchanged.

**Concerns**
- **Shared infra collides with #15071.** This adds `postDocArray`/`postUploadArray` to `ApiInterface`/`UploadRepository`/`UploadRepositoryImpl` — the exact same additions #15071 makes. The `UploadManager` changes are complementary (news here, teams there), but the second of the two to merge will hit a duplicate-declaration conflict on those three files. Extract the `postDocArray`/`postUploadArray` addition into a shared prerequisite PR, or merge one and rebase the other to drop the duplicate.
- No test added for the news bulk path (unlike #15071, which added three). A test mirroring #15071's — bulk success with a mixed per-doc `error`, whole-response failure, and exception — would lock in the index-alignment behavior, which is the subtle part.

**Verdict:** Correct, order-safe news bulk optimization, CI green — ready; coordinate the shared `postDocArray`/`postUploadArray` infra with #15071, and consider adding bulk-path tests to match.

---

Okuro3499 (approved):

PR #15077, opened 2026-07-26
all: smoother user repository parse leaders testing (fixes #15207)

### LINKED ISSUE #15207: 🧪 Add tests for parseLeadersJson in UserRepository

dogi:
🎯 **What:** Added missing unit test coverage for the `parseLeadersJson` public function in `UserRepositoryImpl`.
📊 **Coverage:** Added test cases for parsing valid complete JSON, parsing valid JSON with missing optional fields (ensuring fallbacks apply), and gracefully handling invalid JSON strings (ensuring it returns an empty list rather than crashing).
✨ **Result:** Improved test coverage and reliability by verifying the data mapping logic of `parseLeadersJson`. Created a separate `UserRepositoryImplParseLeadersTest` annotated with `RobolectricTestRunner` to properly execute `org.json` dependent logic in local unit tests without causing Android framework exceptions in the main test suite.

---
*PR created automatically by Jules for task [15194651544046786932](https://jules.google.com/task/15194651544046786932) started by @dogi*


dogi:
**Review — rating: 92/100**

Verified the test against `UserRepositoryImpl.parseLeadersJson` (`app/src/main/java/org/ole/planet/myplanet/repository/UserRepositoryImpl.kt:189`).

**Strengths**
- All three assertions match the real implementation: the `_id` fallback to `org.couchdb.user:${name}` when `_id` is absent, the null handling for `firstName`/`lastName`/`email`, and the `JSONException`→empty-list path (both a malformed root and a missing `docs` array are caught).
- Constructor arity is correct — 18 mocked dependencies match the 18-arg `@Inject` constructor, so this compiles.
- Robolectric is genuinely needed here because `parseLeadersJson` uses `org.json.JSONObject`; good call not stubbing it.

**Concerns**
- None blocking. Minor nit: consider one more case where `name` itself is missing — `docObject.getString("name")` would throw and, since it's inside the try, silently drop that doc. Documenting that behavior with a test would lock it in, but it's not required for this PR.

**Verdict:** Correct, compiles, meaningful coverage of the fallback + error paths — mergeable.

---

Okuro3499 (approved):

PR #15165, opened 2026-07-29
teams: smoother members placeholder handling (fixes #15164)

### LINKED ISSUE #15164: Empty-state view toggling
the same if (list.isEmpty()) { GONE/VISIBLE } shape repeats

Okuro3499:
fixes #15164

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the members screen’s empty-state handling.
  * Empty member lists now display the appropriate no-data message consistently.


dogi:
**Review — 85/100**

**Strengths**
- Behavior-preserving consolidation onto the shared helper. Old code: `if (members.isEmpty()) { tvNodata VISIBLE + no_data_available_please_check_and_try_again } else GONE`. New `BaseRecyclerFragment.showNoData(binding.tvNodata, members.size, "")` does exactly that: `visibility = if (count == 0) VISIBLE else GONE`, and since `noDataMessages[""]` has no entry it falls back to the *same* `R.string.no_data_available_please_check_and_try_again`. Identical text, identical visibility logic.
- `binding.tvNodata` is a `TextView`, so the helper's `v as? TextView` path sets the message directly — no `findViewById` fallback needed. Removes the duplicated empty-state block.

**Nit (non-blocking):** passing `""` as the `source` relies on it *not* being a key in `noDataMessages` to get the fallback string. It works, but a named constant/source key for members would read more intentionally than an empty string.

**Verdict:** Small, behavior-identical refactor to the standard empty-state helper; CI green — ready.

---

dogi:
## Merge Prep Note

This PR modifies app code but does not include a version bump in `app/build.gradle`. Per project conventions, every merged PR touching app code should increment `versionCode` by 1 (and update `versionName` patch version).

Please add the version bump before merging, or let me know if it should be handled separately.

dogi (approved):

PR #15167, opened 2026-07-29
dashboard: smoother notification refresh view modelling (fixes #15162)

### LINKED ISSUE #15162: redundant Retry/backoff logic
a generic RetryUtils.retry() exists, yet RetryInterceptor, SyncActivity.onSyncComplete, and DashboardViewModel.refreshNotificationsWithRetry each hand-roll their own retry loop with different attempt counts/delays.

Okuro3499:
fixes #15162


dogi:
**Review — 82/100**

Verified against `RetryUtils.retry` and both call sites; CI green (build default+lite, test default).

**Strengths — both replacements are behavior-preserving.**
`RetryUtils.retry(maxAttempts, delayMs, shouldRetry, block)` loops `while (attempt < maxAttempts)`, catches into `null`, returns as soon as `!shouldRetry(result)`, and delays *between* attempts only.
- **DashboardViewModel.refreshNotificationsWithRetry** — default `shouldRetry = { it == null }`: `getUnreadNotificationsSize` returns a non-null `Int` on success → returns immediately → `setUnreadNotifications`; exceptions → null → retry; all-fail → null → skip. Same as the old `repeat/try/catch/delay` loop, same attempt count and 300 ms spacing.
- **SyncActivity.onSyncComplete** — `shouldRetry = { it != true }`: retries until `hasAtLeastOneUser()` is true or 3 attempts, then `if (hasUser != true)` logs the timeout. Same outcome, and it actually removes the old loop's wasteful trailing `delay(1000)` after the final check (the old `while` delayed even after the last attempt).
- Nice dedup of two near-identical hand-rolled retry loops onto the shared util.

**Note (pre-existing, but worth flagging since you're expanding its use)**
- `RetryUtils.retry` catches `Exception` broadly, which includes `CancellationException`. So if the coroutine is cancelled mid-`block` (e.g., the ViewModel is cleared during `refreshNotificationsWithRetry`), the cancellation is swallowed and the util retries instead of unwinding — a structured-concurrency foot-gun. Not introduced here (the old DashboardViewModel loop caught `Exception` too), but adopting `RetryUtils` in `viewModelScope` widens the surface. Worth a follow-up to rethrow `CancellationException` inside `RetryUtils.retry`.

**Coordination:** `DashboardViewModel` is also edited by #14979 (survey-nav removal, in review) — different methods, so just pull master between.

**Verdict:** Correct, behavior-preserving consolidation onto `RetryUtils` (with a small trailing-delay improvement); CI green — ready. Recommend a separate fix so `RetryUtils.retry` rethrows `CancellationException`.

---

dogi:
@openhands use skill dogi/merge-prepping

dogi:
The PR touches app code (`ui/dashboard/DashboardViewModel.kt`, `ui/sync/SyncActivity.kt`) but does not bump `versionCode`/`versionName` in `app/build.gradle`. The current version on `master` is `6256` / `0.62.56`. Since the release workflow tags off `versionName`, a missing bump will collide with the previous release.

Before merging, consider bumping the patch version:
- `versionCode = 6256` → `6257`
- `versionName = "0.62.56"` → `"0.62.57"`

dogi (approved):

Issue #15174, opened 2026-07-29
leave team does not work until a new member requests to join


J-S-webskas:
https://github.com/user-attachments/assets/cfa27b3f-def5-49e1-ba24-493450ae59f8

PR #15179, opened 2026-07-29
teams: smoother repository members removing (fixes #15172)

### LINKED ISSUE #15172: remove dialogue does not remove member (still has access to team) until a new member joins (likely)
https://github.com/user-attachments/assets/8409eeb6-8e85-4823-805e-a32a6285bfc0

J-S-webskas:
Filtered out users who have delete pending to not show in query when list updates
This also fixes issue #15174. Closing issue #15174.

https://github.com/user-attachments/assets/31fb252c-eb8b-4dbf-b2ea-194c211e6afb

https://github.com/user-attachments/assets/abbc42d7-04ce-4773-99f6-918068cd2b66

## Bug Fixes

* Team member lists now refresh correctly after a member is removed.
* Team changes are synchronized immediately following member removal.
* Removed pending-deletion memberships no longer appear among joined team members.
* Leave-team actions no longer trigger unnecessary member-change notifications.


ragilzakaria (changes requested):
Thank you for working on this!

My feedback:
- Avoid hardcoded Toast text `"When sync is finished, user will be deleted"`. Please add a string entry to `strings.xml` (e.g. `R.string.member_removed_sync_pending`) and use `getString(...)`. Also remove the trailing semicolon `;`.
- Saying *"user will be deleted"* is misleading because removing a member from a team does not delete their user account. Consider wording like *"Member removed. Changes will sync shortly."*

ragilzakaria (approved):

dogi:
**Review — rating: 82/100**

Verified against `MembersFragment` / `BaseTeamFragment`. The new `teamsSyncRepository.syncTeamActivities()` resolves to the `@Inject`-ed `teamsSyncRepository` inherited from `BaseTeamFragment`, and it's the same suspend call already used in `TeamResourcesFragment`/`TeamDetailFragment`, invoked here inside the existing `lifecycleScope.launch { try … }` block right after `removeMember`. Compiles, and pushing the removal to the server immediately is the right behavior for the reported bug. New string resource added correctly.

**Concerns (nits)**
- `MembersFragment.kt:159` — trailing semicolon on `…Toast.LENGTH_SHORT).show();` is non-idiomatic Kotlin; drop it.
- Ordering: because `syncTeamActivities()` is a suspend call, the "Member removed. Changes will sync shortly." toast actually shows *after* the sync completes (or not at all if it throws and hits the `catch`). Minor UX mismatch — if the intent is an optimistic "syncing…" message, show the toast before awaiting the sync. Functionally fine either way since the failure is caught.

**Verdict:** Correct and low-risk. Mergeable — moving to `ready`.

---

J-S-webskas:
@openhands prepping

J-S-webskas:
Prepped — addressed the two nits from the review (commit 904b6f8):

- Dropped the non-idiomatic trailing semicolon on the `Toast…show()` call.
- Reordered so the "Member removed. Changes will sync shortly." toast shows **before** awaiting `syncTeamActivities()`, giving an optimistic "syncing…" message rather than only appearing after the sync completes (or being swallowed by the `catch` on failure).

This message was created by an AI agent (OpenHands) on behalf of J-S-webskas.

dogi:
**Triage — CI red**

`myPlanet build (default)`, `myPlanet build (lite)` and `myPlanet test (default)` all fail. The cause is a Kotlin compile error, not a test regression:

```
> Task :app:compileDefaultDebugKotlin FAILED
Unresolved reference 'userDao'.
Location: app/src/main/java/org/ole/planet/myplanet/repository/TeamsRepositoryImpl.kt line 1041
```

`master` is green on `myPlanet test (default)`, so this is not pre-existing. Moving to `change` — push a fix and have a maintainer put `review` back on for a re-review.

---

dogi:
**Review — rating: 81/100**

CI is green now (build default + lite, test default), so the `Unresolved reference 'userDao'` compile error from the last round is resolved. My earlier note was triage-only and carried no score; this is the first rated review.

The fix is correct and it follows the convention already established in the file it edits. `removeMember` delegates to `markMembershipsForLeave`, which sets `isDeletePending = true` (`TeamsRepositoryImpl.kt:1406`) — a soft delete that stays in Room until `UploadManager` PUTs it to the server. So the membership row genuinely still exists locally right after someone leaves, which is exactly why #15172 showed departed members in the list. Filtering `!it.isDeletePending` in `getJoinedMembers` is the right layer to fix it, and three sibling queries in the same class already do precisely that — lines 191, 621 (`isMember`), and 656. Consistent, not ad hoc.

I also checked the deletion that looked riskiest, removing `onMemberChangeListener?.onMemberChanged()` from `handleLeaveTeam()`. It's justified: `TeamDetailFragment.onMemberChanged()` (`:392`) only recomputes `getJoinedMemberCount` to toggle `btnLeave` visibility, and `MembersFragment` is a ViewPager page built by `TeamPageConfig.createFragment()` rather than its own back-stack entry — so the `popBackStack()` two lines later tears down the host `TeamDetailFragment` itself. The old call was recomputing a button on a fragment about to be destroyed, and touching `_binding` during teardown. Dropping it loses nothing. The `requestsViewModel.fetchMembers(teamId)` added to the remove-member path mirrors the existing call at `:85`, so that side is consistent too.

**Nit:** the diff introduces trailing whitespace in three places — most visibly `import kotlinx.coroutines.launch ` (trailing space on an import) and a line of bare spaces after the `mapNotNull` chain in `TeamsRepositoryImpl`. Please strip those; they're the only reason this diff touches `MembersFragment`'s import block at all. The comment could also read "Filter out memberships pending deletion".

Worth adding while you're here: `TeamsRepositoryImplTest.kt` already exists and currently *stubs* `getJoinedMembers` rather than exercising it, so the new filter has no coverage. A single case — two membership rows, one `isDeletePending = true`, assert only the other comes back — would lock in the behavior this PR exists to provide.

**Verdict:** mergeable — the filter is correct and the callback removal is safe; the whitespace and the missing test case are worth a quick follow-up commit but shouldn't block.

---

dogi (approved):

PR #15224, opened 2026-07-31
life: smoother repository shelf visibility updating (fixes #15220)

### LINKED ISSUE #15220: mylife: fix hide the submenu after immediately go back
https://github.com/user-attachments/assets/f3ea2592-44ab-4fe6-b9aa-6734bbc825d9

when we try to hide a submenu, and then go back to home immediately, the menu is stll there (in the homepage) unless we try to click another page and then go back gto the homepage.

ragilzakaria:
[Screen_recording_20260731_214522.webm](https://github.com/user-attachments/assets/e4780e29-e80a-42c8-a6b7-2ace1bbd95dd)

fixes #15220 

## Summary by CodeRabbit

* **Bug Fixes**
  * My Life dashboard entries now refresh automatically when returning to the screen.
  * Visibility updates work more reliably across entries identified by ID, image, or title.
  * Improved handling of missing user information.
  * Prevented duplicate My Life entries from appearing.
  * List updates now accurately reflect changes to visibility, weight, and titles.
  * Improved consistency when loading and caching My Life entries.
* **Release**
  * Updated the app to version 0.63.21.


J-S-webskas (approved):
Nicely fixed.

Okuro3499 (approved):

dogi:
**Review — rating: 80/100**

The title undersells this — it's a real fix to My Life visibility persistence and duplicate items, not just a submenu tweak. I traced it against the actual sources (`MyLife` entity, `LifeRepositoryImpl.seedMyLifeIfEmpty`, `MyLifeDao`, `LifeFragment`).

**What's correct**
- Root cause is understood: DB-seeded items get a real `_id = UUID.randomUUID()` (in `seedMyLifeIfEmpty`), but the SharedPreferences **cache** reconstructs items via `MyLife(imageId, userId, title)` which leaves `_id = ""`. The `_id ?: imageId ?: title` fallbacks in `LifeFragment`, `dedupKey()`, and the `DiffUtil` callback correctly bridge those two representations — that's why the previous `_id`-only matching failed for cached items.
- Reordering `getMyLifeForDashboard` to read the DB first and making `updateVisibility` re-fetch + re-cache is what makes the hidden item reflect immediately on back-navigation. `onResume → refreshMyLifeList` closes the loop.
- On threading: worth noting the DAO calls are Room `suspend` functions, so they run on Room's own executor even though `refreshMyLifeList` launches on `lifecycleScope` (Main) — the earlier "runs DB on main dispatcher" worry is largely mitigated; only `removeAllViews()`/view rebuild runs on Main, which is correct.

**Concerns**
- `MyLifeDao.kt:25` — `UPDATE ... WHERE _id = :id OR imageId = :id OR title = :id`. In the live path `id` is always a real `_id` (adapter items come from the DAO), so this is harmless today, but `OR imageId`/`OR title` is unscoped by `userId`: if it were ever called with an `imageId`/`title` (the cache path), it would flip that row's visibility for **every** user who has that item on a shared device. Recommend dropping the `title` alternative and scoping the write by `userId` (as CodeRabbit noted) — cheap insurance against a real multi-user data bleed.
- `BaseDashboardFragment.onResume` rebuilds the entire My Life flexbox unconditionally on every resume (`removeAllViews` + re-init). Fine for this small list, but a shared refresh `Job` (cancel-and-replace) would avoid overlapping rebuilds if resumes stack.
- `getByUserId` now also returns `userId IS NULL` rows for any user; intended for legacy/global items and de-duped, but a brief comment documenting that (replacing the removed `IS`-semantics comment) would help the next reader.

**Verdict:** Works, already approved by two members, and the remaining items are defensive hardening rather than active bugs. Mergeable — but please tighten the `updateVisibility` predicate before/soon after merge. Moving to `ready`.

---

dogi (approved):

PR #15226, opened 2026-08-02
feat(flutter): Flutter/Dart port of myPlanet (phases 1–28)

dogi:
## What

A Flutter/Dart port of the Kotlin app, living in `flutter/` **alongside — not replacing —** `app/`. The Kotlin app is unchanged and remains the shipping app; its `build.yml`/`test.yml` workflows are untouched (they now skip pushes that only touch `flutter/`/docs). A separate path-filtered `flutter.yml` guards the port with analyze, format check, full test suite, and an Android debug APK build.

**Status: 28 of 28 UI packages have a screen**, offline-first against the real CouchDB API, with a durable write-back path (an `outbox` table + lifecycle drain replacing `RetryQueue`/`RetryQueueWorker`). ~32k lines of Dart, ~11k lines of tests (501 tests, all green). See `docs/kotlin-to-flutter-migration.md` for scope, the technology mapping (Hilt→Riverpod, Room→Drift, Retrofit→Dio, strings.xml→.arb), faithful quirks, deliberate deviations, and open problems — and `flutter/README.md` for how to run it.

## Currently in this PR

- Server configuration → login (online/offline, PBKDF2 against `_users`) → resources, courses, shelf write-back with `removed_log` semantics
- **The home bell dashboard (Phase 29)**: profile card, the four myLibrary/myCourses/myTeams/myLife cards with counts and guest gates, and the pending-survey dialog with its one-hour throttle
- Dashboard shell, calendar, onboarding, profile, settings, dictionary, notifications, My life, references, personals, ratings
- Submissions (create/upload/review/PDF export), events/meetups, individual + team/public surveys, graded course exams
- Teams/enterprises (catalog, tasks, membership, resources, courses, financial reports), voices/discussions, chat (with upload + retry), feedback, community
- Resource viewer (video/audio/PDF/image/markdown/text/CSV) with foreground download path, encrypted health records (AES-256-CBC, matching `AndroidDecrypter`), offline maps, storage management, become-a-member
- Localization: full `app_en.arb`, `app_es.arb` seeded only from existing `values-es` strings

## Rebase + upstream harvest (2026-08-12)

Rebased onto master (130 Kotlin commits since the previous base). Every upstream commit was reviewed for behavioral change; four were harvested into the port, three of them parity fixes for bugs Kotlin fixed after the fork:

- Feedback sync no longer destroys a pending local reply (`af4605362`)
- Removing a resource from the shelf now writes a `removed_log` row so the removal sticks (`4345d871a`)
- `CourseMapper.mergeUserIds` drops blank entries (`ff7bb5c7c`)
- Storage management is guest-gated (`82140152b`)

The remaining upstream deltas are spec notes for unported features, recorded in the migration doc's "Harvesting upstream" section.

## Known gaps (tracked in the migration doc)

- Background work with no user present (timed auto-sync, deadline notifications, background downloads) needs OS scheduling and is not ported
- Home-dashboard extras: completed-course stars, network-status ring, team alert badges, activity chart, last-sync strip, drawer/overflow menu, survey remind-later
- Team attachments; course progress/certification; a real deep-link plugin for public surveys; Unicode PDF export

https://claude.ai/code/session_01DJ3hhPtHXBDABrcJhrshgk


dogi:
Addressed the review in `fda17df`. Most findings were real bugs, not style — thanks. Two were more serious than reported, and one I'm declining.

### Fixed

**Network was broken on release builds.** `INTERNET` was declared only in the debug/profile manifests `flutter create` generates, so release builds had no network permission at all. Worse, and not flagged: **neither platform permitted cleartext**, but the primary myPlanet deployment is a local community server on plain `http://<ip>:5000`. Android 9+ blocks that by default, so every local-server install would have failed at the configuration step regardless of the permission. Added `usesCleartextTraffic` + a `network_security_config.xml` mirroring the Kotlin app's, and `NSAllowsArbitraryLoads` on iOS. Verified in the built APK:

```
$ aapt2 dump permissions app-debug.apk
uses-permission: name='android.permission.INTERNET'
$ aapt2 dump xmltree --file AndroidManifest.xml app-debug.apk | grep -i cleartext
android:usesCleartextTraffic(0x010104ec)=true
```

**Data loss on a partial sync.** Correct and the most serious functional finding. A short page mid-walk left `savedIds` holding only a prefix, and `deleteNotIn` then removed every local row outside it — discarding offline content the server still had. Cleanup now runs only after a complete walk.

**`deleteNotIn` bind limit** — correct, and it would have hit any real library. The stale set is computed in Dart and deleted in chunks. Applied to the courses DAO too.

**`getByName` crash** — correct. Two planets can legitimately hold the same account name, so I limited to one row rather than adding a unique constraint that would reject valid data.

**`clear()` didn't reach storage** — correct; "change server" was undone by the next cold start.

**`validateStatus` per request** — correct, applied as suggested.

**Secrets in `SharedPreferences`** — took the fix rather than the documentation option. The PIN and credentialed URL moved to `flutter_secure_storage`, cached at startup so the config getter stays synchronous (the router reads it during a sync `build()`). Related and worse: `resourceRemoteAddress` was storing `satellite:<pin>@` into *every resource row*, so the PIN was written into the database thousands of times. Userinfo is now stripped and credentials attached at request time.

**URL encoding** — applied to all three sites, keeping the `org.couchdb.user:` colon literal as CouchDB requires. `resourceUrl` now returns `null` rather than interpolating the literal text `null`.

**`persist-credentials: false`** — applied to both checkout steps; neither job pushes.

### Declined

**`android.builtInKotlin=false` leaves `:app` with no Kotlin compiler.** This is a false positive. The `dev.flutter.flutter-gradle-plugin` applies the Kotlin plugin itself, so the `apply false` in `settings.gradle.kts` isn't the only provider. `MainActivity.kt` compiles and the debug APK builds — both locally and in the `Build Android APK` CI job on this PR. No change made.

### Note on scope

Several of these are places where the Kotlin has the same bug (the `http:///` fallback, the unencoded URLs, the PIN in plain prefs). I've fixed them in the port rather than reproducing them, and each deviation is documented in code and in `docs/kotlin-to-flutter-migration.md` alongside the quirks I *did* reproduce deliberately — chiefly the PBKDF2 iteration count pinned at 10, which cannot change without locking out every existing user.

164 tests pass (9 new, covering the partial walk, the >999-id delete, duplicate account names, credential stripping and URL encoding); `flutter analyze` is clean.

---

dogi:
Addressed in `552aad1`. Two findings were serious and I'm glad they were caught.

### The two that mattered

**Step ids caused silent data loss.** Correct, and worse than described once I traced it. `base64(jsonEncode(step))` was the sole primary key of `course_steps`, so two steps with identical content collide — within a course the upsert drops one, and *across* courses the surviving row keeps whichever `courseId` was written last, so the step disappears from the other course entirely. Ids are now `<courseId>:<position>`: bounded, unique, stable across syncs.

I also took the follow-on you flagged: `upsertAll` now deletes steps past the new length, since upserting alone cannot shrink a course — a five-step course edited down to three kept two orphans.

Worth noting the Kotlin has the identical collision. I diverged rather than reproducing it because the step id is local in both apps and never a server value, so nothing downstream depends on the scheme. Documented as a deviation.

**The filter dropdown could assert.** Also correct, and a real crash on a core interaction. `gradeLevelsProvider` watched `coursesStreamProvider`, which watches `courseFilterProvider` — so choosing a grade invalidated the option list it was chosen from, and while it reloaded the dropdown saw an empty item list with a non-null selection. Fixed at the root with dedicated DAO streams independent of the filter, plus the defensive `value` guard you suggested.

### Also fixed

Stuck `SyncRunning` on a thrown `runSync` (sync button permanently disabled); onboarding stranding the user if the prefs write threw; the search box desyncing from "clear filters"; unencoded `resourceRemoteAddress` segments; `setShelfMembership` writing membership and removal log non-atomically; `getByIds` bind limit; `keyFor` colliding on the `:` inside CouchDB user ids; per-request timeouts; `encryptedSharedPreferences` (which my doc comment already claimed); route id encoding; `course_steps` index; set-based `mergeIds`; one-shot shelf reads; `ExcludeSemantics` on Skip; and the "Two vertical slices" miscount.

Tests: rewrote the step-id contract (duplicates within and across courses, bounded length), added step-shrink cleanup, the courses partial-walk guard, removal-log side effects and per-user isolation, and switched the widget tests to `ProviderScope.containerOf` instead of a `WidgetRef` captured outside `build`.

### Declined, with reasons

**Host allow-list for cleartext.** Not implementable as suggested. The server address is entered by the user at runtime; Android's `network_security_config.xml` is static and `domain-config` needs compile-time domains, so there is no way to express "whatever host the operator typed". The Kotlin app has the same blanket policy for the same reason. Runtime enforcement at the Dio layer *would* work and is a reasonable hardening, but it's a design change rather than a config tweak — noted in the migration doc as follow-up.

**`myCourses: "misCursos"`.** This is genuinely what `values-es/strings.xml` contains for `my_courses`. It does look like an identifier that leaked into the translation, so I think you're right that it's wrong — but it's wrong *upstream*, and the Kotlin app renders it today. Fixing it only here would make the two apps disagree. Better fixed in Crowdin so both pick it up; I've left the port faithful. **Flagging it as a real defect worth a separate issue.**

**`describeNetworkFailure` returning English.** Valid — Spanish users see an English prefix in sync errors. Deferring rather than declining: doing it properly means `SyncErrored` carrying a structured code/message and the widget formatting via `AppLocalizations`, which touches both sync notifiers and both screens. That belongs in its own change rather than bolted onto this round.

194 tests pass; `flutter analyze` and `dart format` clean.

---

PR #14960, opened 2026-07-24
login: make login history visible and scrollable in landscape mode (fixes #14947)

ragilzakaria:
- Updated activity_login.xml, activity_login.xml, and activity_login.xml
- Reduced logoImageView size to 56dp/64dp and top margin to 12dp/16dp
- Compacted vertical margins and padding across title text, version labels, language button, and storage space, reclaiming over 180dp+ of vertical space for recyclerView
- Added android:nestedScrollingEnabled="true" in the XML files and set binding.recyclerView.isNestedScrollingEnabled = true in LoginActivity.kt, enabling users to view and scroll through saved accounts effortlessly in landscape mode

[Screen_recording_20260724_193512.webm](https://github.com/user-attachments/assets/40d9ef60-43a9-4737-9fcb-127d20107a37)

fixes #14947 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved login screen scrolling by disabling nested scrolling for the user/team lists, avoiding conflicting scroll behavior across all screen sizes and initialization paths.

* **Style**
  * Refined the login layout by wrapping relevant panels in a scroll container and adjusting sizing/constraints so content fits naturally with `wrap_content`.
  * Updated spacing, typography, and branding positioning for a more compact, consistent look.
  * Refreshed user list item visuals with smaller, fixed avatar sizing and updated padding.

PR #15238, opened 2026-08-03
login: smoother landscape scrolling (fixes #15235)

### LINKED ISSUE #15235: Login page should be fully scrollable in landscape mode
## Description

When the device is rotated to **landscape mode**, the login page does not scroll properly, making some UI elements (such as input fields or the login button) inaccessible or cut off below the visible viewport.

## Steps to Reproduce
1. Open the myPlanet app.
2. Rotate the device to landscape mode.
3. Navigate to the login page (or observe it on app launch).
4. Attempt to scroll down to view all login page elements.
https://github.com/user-attachments/assets/621c924f-8978-4208-aa70-5f2d645d7a2d

## Expected Behavior
The entire login page should be scrollable in landscape mode, allowing the user to access all UI elements (username field, password field, login button, etc.) without any content being hidden or unreachable.

ragilzakaria:
[Screen_recording_20260803_220614.webm](https://github.com/user-attachments/assets/fbdb464a-9f55-410c-abcb-37020ff7dc6c)

fixes #15235 

## Summary by CodeRabbit

* **UI Improvements**
  * Updated the login screen layout for improved spacing, alignment, and content sizing across screen sizes and orientations.
  * Improved positioning of login actions, footer links, community controls, and the logo.
  * Adjusted panel widths and vertical spacing for better use of available screen space.
  * Team and community lists now display content more naturally without nested scrolling.


dogi:
**Review — rating: 87/100**

Verified the diff against the actual head source (all 5 `activity_login.xml` variants), not just the description.

**Strengths**
- The core mechanism is correct: the root is already a `NestedScrollView` with `fillViewport="true"`, and this PR consistently switches the content chain (`LinearLayout` + both column `ConstraintLayout`s) to `wrap_content` and sets the saved-user `RecyclerView` to `wrap_content` + `nestedScrollingEnabled="false"`. That is the right pattern for making the whole page scroll in landscape — the description's claim holds up.
- Replacing the inner `ScrollView` with a plain padded `LinearLayout` in the portrait/night variants correctly removes the nested-scroll conflict.
- Change is applied uniformly across all five configuration variants, and attribute ordering was tidied along the way.

**Concerns**
- Cosmetic, worth a glance in the recording: with the outer `LinearLayout` at `wrap_content` and both columns `wrap_content`, the shorter column no longer stretches, so its background (`@color/mainColor` right panel / `@color/light_dark` left) stops at its content instead of filling to the taller column's bottom, exposing the parent `secondary_bg`. Previously `match_parent` kept both panels equal-height. If the two-tone split looks uneven in tall/short states, consider keeping the columns `match_parent` while leaving the horizontal `LinearLayout` `wrap_content`.
- `recyclerView` no longer recycles (`wrap_content` + `nestedScrollingEnabled="false"`) and `android:scrollbars="vertical"` is now a no-op. Fine here — the saved-users list is inherently small — but don't reuse this view for an unbounded list. Note: CodeRabbit's suggestion to revert to `layout_height="0dp"` is **not** appropriate for this design; a `0dp` height constrained top+bottom inside a `wrap_content` scroll container collapses to zero. Leave as-is.

**Cross-PR**
- Overlaps #14960 (same author — "make login history visible and scrollable in landscape"), which touches the same landscape `recyclerView` sizing/nested-scrolling. This PR is the more complete fix and effectively subsumes it; recommend consolidating on this one and closing/rebasing #14960 to avoid conflicting edits to the same files.

**Verdict:** Mergeable — correct approach, only cosmetic/perf nits. Moving to `ready`.

---

Okuro3499 (changes requested):
ui breaks in portrait
Image

Okuro3499 (approved):

dogi:
**Review — Rating: 87/100**

Verified the diff against all five `activity_login.xml` variants at the PR head (`layout/`, `layout-night/`, `layout-large-land/`, `layout-normal-land/`, `layout-xlarge-land/`). The change does exactly what the title claims, and the approach is the textbook-correct one.

**Strengths**
- The root of each layout is already an `androidx.core.widget.NestedScrollView` with `fillViewport="true"`. The fix flips the inner horizontal `LinearLayout` from `layout_height="match_parent"` to `wrap_content` — this is precisely what makes content taller than the viewport actually scroll instead of being clipped. A `match_parent` child of a scroll container never scrolls, so this was the real bug and it's fixed correctly.
- The `layout/` and `layout-night/` variants also remove a **nested** `<ScrollView>` (with `fillViewport`) around the left profile pane. Nesting a vertical scroller inside a `NestedScrollView` is an anti-pattern (scroll-conflict, incorrect measurement) — collapsing it into the outer scroll is the right call, and matches the CodeRabbit note about "content more naturally without nested scrolling."
- Consistent treatment across every density/orientation bucket, so behavior won't diverge between screen sizes. CI is green on both flavors.

**Concerns (minor, non-blocking)**
- In `layout/activity_login.xml` the left pane `ConstraintLayout` (`@+id/lyt`) keeps `layout_height="match_parent"` while its parent `LinearLayout` is now `wrap_content`. Android resolves this via a second measure pass (child gets the tallest-sibling height), so it works, but the login form inside still uses top+bottom constraints to vertically center — on a very short screen the centering + scroll interaction is worth a quick manual check against the shortest supported device. The screen recording covers the tablet case well; a phone-landscape spot-check would fully close it.
- The diff also carries cosmetic churn (blank-line insertions, `gravity` attribute reordering) that inflates the line count but is harmless.

**Verdict:** Correct, consistent, and CI-green — mergeable as-is; the short-screen centering note is a spot-check, not a blocker.

---

dogi (approved):

PR #15245, opened 2026-08-03
feedback: smoother repository dao replying (fixes #15243)

### LINKED ISSUE #15243: Feedback reply can be silently lost
`FeedbackRepositoryImpl.addReply` appends a reply locally but does not set `isUploaded = false` for the updated feedback row. This creates a compounding data-loss bug where a user's reply is never uploaded and can later be overwritten.

Okuro3499:
fixes #15243
Prevent server sync from overwriting unuploaded feedback replies. Mark feedback as pending upload when a reply is added, and preserve local messages during insertFromJson if they haven't been uploaded yet. Add batch getByIds query to FeedbackDao and corresponding tests.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved feedback synchronization to preserve locally pending replies during server updates.
  * Feedback modified locally is now correctly marked for upload.
  * Server data replaces local feedback only when no pending local changes exist.

* **Tests**
  * Added coverage for pending reply handling and server synchronization behavior.


dogi:
**Review — rating: 87/100**

Verified against `FeedbackRepositoryImpl` / `FeedbackDao`; build (default+lite) and unit tests are green. This is a clean offline-first fix for the "local reply lost on sync" bug (#15243):
- `addReply` now sets `isUploaded = false`, so a freshly added reply is correctly queued for upload (previously it could be treated as already-synced).
- `insertFromJson` / `insertFeedbackList` first load the existing row(s) (`findById` / the new batch `getByIds`) and, when a record has a pending local reply (`isUploaded == false`), preserve the local `messages` and keep `isUploaded = false` instead of overwriting from the server — while still taking the server's other fields and `_rev`. That's the right merge for not clobbering unsynced work.
- Three focused tests cover all branches (reply marks pending, sync preserves pending local, sync overwrites when nothing pending). Nice.

**Concerns (minor)**
- The merge keys entirely off the single `isUploaded` boolean. If a reply was uploaded but `markFeedbackUploaded` didn't run (crash/interrupt between POST and mark), the next sync would re-preserve and re-upload it, duplicating the message server-side. Low probability and strictly better than losing replies, but worth being aware of.
- `insertFeedbackList` does one `getByIds` for the batch — good — but if the same `_id` appears twice in a single server page, `associateBy { it.id }` silently keeps the last; harmless here.

**Verdict:** Correct, well-tested offline-preservation fix. Moving to `ready`.

---

dogi (approved):

PR #15178, opened 2026-07-29
teams: smoother repository members leader switching (fixes #15408)

J-S-webskas:
https://github.com/user-attachments/assets/8d4d2e39-e840-475f-b83d-51382d88eddf

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved team leader updates to keep all membership records synchronized.
  * Member lists now refresh immediately after changing a team leader, ensuring leader status is displayed accurately.

PR #15247, opened 2026-08-03
teams: smoother members leader refreshing (fixes #15246)

### LINKED ISSUE #15246: avoid full list rebind/flicker by fixing JoinedMemberData equality
MembersAdapter (MembersAdapter.kt:44) uses JoinedMemberData.user: UserEntity in a way that lacks reliable value equality for adapter diffing.
After team-member list refreshes (leave/remove/make-leader), unchanged members are treated as changed, so every row rebinds, Glide reloads avatars, and the whole grid flickers.

Okuro3499:
fixes #15246


dogi:
**Review — rating: 84/100**

Verified against `MembersAdapter`; build (default+lite) and tests are green. Two sound fixes:
- `areContentsTheSame` no longer relies on `oldItem == newItem` (reference equality for the non-data-class `JoinedMemberData`, which made it effectively always "changed") and instead compares the real display fields (`isLeader`, `visitCount`, `lastVisitDate`, `user.name`, `user.userImage`, `getRoleAsString()`) — so `DiffUtil` can actually skip unchanged rows.
- `updateData` detects a change in the logged-in user's leader status and, on change, forces a full `notifyItemRangeChanged(0, itemCount)` after `submitList`, so every row re-binds its leader-only action buttons. That directly addresses #15246.

**Concerns**
- `areContentsTheSame` is now an explicit allowlist — if a row ever surfaces a field not in that list, its change won't trigger a rebind. Fine for the current row layout; just keep it in sync if the row gains new bound fields.
- **Cross-PR:** this overlaps #15178 ("3 buttons still showing after leader switched"), which fixes the same class of bug from the fragment/repository side (`updateTeamLeader` + `membersAdapter.updateData(members, false)`). They're complementary — #15178's `updateData` call now triggers this PR's leader-status rebind — but coordinate the two so the fix isn't half-applied. This adapter-level change is the cleaner core of the two.

**Verdict:** Correct, targeted diff/refresh fix. Moving to `ready`.

---

ragilzakaria (approved):

dogi (approved):

PR #15251, opened 2026-08-03
enterprises: smoother reports state change notifying (fixes #15249)

### LINKED ISSUE #15249: Fix `areContentsTheSame` always returning false for `MyTeam` Room entity in DiffCallback
## Description

`areContentsTheSame` in the `DiffCallback` for `MyTeam` was using reference equality (`oldItem == newItem`) to compare two instances of the `MyTeam` Room `@Entity` open class, which has no `equals()` override. Since Room entity instances are always distinct objects, this comparison **always returns `false`**, causing every item to be treated as changed on every list update.

## Root Cause
`MyTeam` is a Room `@Entity open class` with no custom `equals()` / `hashCode()` implementation. The `==` operator in Kotlin falls back to reference identity for such classes, so two distinct `MyTeam` objects representing the same database row are never considered equal.

The list is Flow-driven, sourced directly from Room via `getReportsFlow`. Room re-emits the flow on **every relevant DB write**. Because `areContentsTheSame` always returns `false`, every emission forces a full rebuild of every report card — including a fresh Glide image reload per report — even when only one unrelated report changed.

Okuro3499:
fixes #15249
Add early return and notifyItemRangeChanged in setNonTeamMember to properly refresh items. Replace equals-based areContentsTheSame with field-by-field comparison for reliable diff detection.


dogi:
**Review — rating: 85/100**

Verified against `EnterprisesReportsAdapter`; build (default+lite) and tests are green. Same well-founded pattern as the other adapter fixes:
- `setNonTeamMember` now guards against a no-op change and, when the flag actually flips, calls `notifyItemRangeChanged(0, itemCount)` so the reports rebind for the new member/non-member state — previously the field changed silently and the list didn't refresh.
- `areContentsTheSame` replaces reference-equality `oldItem == newItem` (`MyTeam` is a Room `@Entity`, not a data class, so `==` was effectively always-different → every item rebinding) with an explicit compare of the report's financial/display fields (dates, balances, sales, income, wages, expenses, description, `imageName`). Proper diffing now works.

**Concerns (minor)**
- The `areContentsTheSame` allowlist must stay in sync with whatever the row actually renders; if a bound field is added later and not included here, its updates won't rebind.
- `notifyItemRangeChanged(0, itemCount)` on the member-state flip is a full rebind — fine for a reports list of this size.

This is part of a consistent, good campaign of `DiffUtil` correctness fixes (cf. #15247, #15224). No cross-PR conflict — distinct adapter/file.

**Verdict:** Correct, low-risk fix. Moving to `ready`.

---

dogi (approved):

PR #15258, opened 2026-08-04
login: less settings edges is more (fixes #15257)

### LINKED ISSUE #15257: hidden buttons
in the settings activity the top item is hidden by the app bar
Image

Okuro3499:
fixes #15257
Screenshot_20260804_161707

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved edge-to-edge layout handling in Settings.
  * Removed the unused dark mode preference from the Settings screen for a cleaner configuration experience.


J-S-webskas (commented):
https://github.com/user-attachments/assets/7e4487bc-a90e-43b2-accd-bc1f3cce2ee0

This is a recording I had on July 30th, but settings was not cropped. I did not pull in the request, but the light mode setting also works fine there. Not sure if this is a bug, or maybe specific to the device.

dogi:
**Review — rating: 55/100**

**Strengths**
- The actual "hidden button" fix is correct: applying edge-to-edge insets to `findViewById(android.R.id.content)` instead of `window.decorView` means the padding lands on the content frame (where the buttons live) rather than the whole window, so system bars stop overlapping the settings buttons. Good, targeted fix for #15257.
- Removing the empty `onDestroy()` override is harmless cleanup.

**Concerns**
- **The PR also deletes the `dark_mode` theme preference** — from `pref.xml` and its click handler — which has nothing to do with "ensure no button is hidden" and *removes* a user-facing theme toggle from Settings. If the goal is to fix cropped buttons, this deletion looks out of scope; if the theme setting is genuinely being retired (moved elsewhere?), that needs its own rationale in the PR description, because on its face this reduces functionality rather than un-hiding a button.
- **Direct conflict with #15313 (already `ready`).** That PR *keeps* the `dark_mode` preference and updates its handler to `themeManager.showThemeDialog(requireContext())` as part of the ThemeManager DI refactor. This PR removes the same preference. They can't both merge — decide whether `dark_mode` stays (then drop the deletion here and keep only the edge-to-edge fix) or goes (then #15313's change is moot and should be dropped), and reconcile.

**Verdict:** Keep the edge-to-edge inset fix — that's the real solution to the hidden-button bug. But split out or justify the `dark_mode` removal and reconcile it with #15313 before this can merge.

---

app/src/main/res/xml/pref.xml
```diff
@@ -1,9 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
 <PreferenceScreen xmlns:android="http://schemas.android.com/apk/res/android">
-
-    <Preference
-        android:key="dark_mode"
-        android:title="@string/dark_mode" />
```
Okuro3499, line 6: > Recording_20260730_132247.1.mp4 
> This is a recording I had on July 30th, but settings was not cropped. I did not pull in the request, but the light mode setting also works fine there. Not sure if this is a bug, or maybe specific to the device.

@J-S-webskas dark mode was the one hidden here, which I removed in this pr as it is already in the menu as shown on screenshot. Bug was the first item was alway hidden

Image

dogi:
**Review — 85/100**

Small and correct.

**The fix (SettingsActivity.kt):** moving `EdgeToEdgeUtils.setupEdgeToEdge(this, …)` from `window.decorView` to `findViewById(android.R.id.content)` is the right call. `SettingFragment` is hosted in `android.R.id.content`, so applying the window insets there pads the actual content frame (below the status bar / above the nav bar) instead of the decor view, which is what was letting buttons slide under the system bars. `android.R.id.content` is present without a `setContentView`, so the lookup is valid in `onCreate`. ✓ CI green on all three jobs.

**On removing the `dark_mode` preference:** I checked, and this does **not** orphan theme switching — it's still reachable from `DashboardActivity` (`R.id.action_theme → ThemeManager.showThemeDialog`) and from `LoginActivity`. So the Settings entry was a duplicate and removing it is safe. Just flagging it as slightly out of scope for a "hidden buttons" fix so it's a conscious decision: if you intend Settings to remain a theme entry point, keep it; otherwise this is fine decluttering. (Dropping the empty `onDestroy()` override is harmless cleanup.)

**Verdict:** Correct, minimal, safe. Mergeable — just confirm you're happy dropping the theme row from Settings.

---

J-S-webskas (approved):

PR #15259, opened 2026-08-04
teams: smoother voices repository image editing (fixes #15253)

### LINKED ISSUE #15253: newly added image does not show when editing a voice
## Summary
Editing a voice and adding a new image does not display the newly added image.

## Steps to reproduce
1. Open a voice post that already exists.
2. Start editing the voice.
3. Add a new image during the edit flow.
4. Save the changes.
5. View the updated voice post.

## Expected behavior
The newly added image should be displayed along with the updated voice content after saving the edit.

## Actual behavior
The newly added image does not show after editing and saving the voice.

## Evidence
Voice edit image not showing

ragilzakaria:
https://github.com/user-attachments/assets/8d1487e8-b547-47f8-a0e3-701f0288a2b9

fixes #15253


J-S-webskas (commented):
https://github.com/user-attachments/assets/707babc5-1c70-4676-8783-4184c84de1d4


The issue is the images only show after the recycler view refreshes. Looking at Gideon's fix video, the issue is in "enterprises", but "team's chat" also have the same problem.

J-S-webskas (dismissed):

Okuro3499:
> Recording_20260804_111855.mp4 
> The issue is the images only show after the recycler view refreshes. Looking at Gideon's fix video, the issue is in "enterprises", but "team's chat" also have the same problem.

@J-S-webskas teams and enterprises are the same a fix should work in both places

J-S-webskas:
> > Recording_20260804_111855.mp4
> > The issue is the images only show after the recycler view refreshes. Looking at Gideon's fix video, the issue is in "enterprises", but "team's chat" also have the same problem.
> 
> @J-S-webskas teams and enterprises are the same a fix should work in both places

It's not working in enterprises right now

Okuro3499 (dismissed):
> > > Recording_20260804_111855.mp4
> > > The issue is the images only show after the recycler view refreshes. Looking at Gideon's fix video, the issue is in "enterprises", but "team's chat" also have the same problem.
> > 
> > 
> > @J-S-webskas teams and enterprises are the same a fix should work in both places
> 
> It's not working in enterprises right now

yes I confirmed it is not working

https://github.com/user-attachments/assets/9ee35ed3-991d-4dab-8590-4b8bbccada9b

dogi:
**Review — Rating: 55/100**

The core mechanism is right and several sub-fixes are genuinely good, but it collides with an already-`ready` PR and a maintainer reports it's still broken in one place.

**Strengths**
- `editPost` now returns `News?` (re-fetched via `getById` after `upsert`), and `VoicesActions` threads that fresh object through `onSuccess(updatedNews)` → `updateReplyButton(viewHolder, updatedNews ?: news, …)`. That's the correct way to get the post-edit state to the UI.
- **Real bug fix in `TeamsVoicesFragment`:** `val adapterNews = …` → `adapterNews = …`. The old code shadowed the field with a local, so `onDataChanged()` (which uses the field) operated on a stale/other reference. Good catch.
- The adapter edit-callback now builds a new list (`newList[pos] = targetNews`; `preParseNews`; `submitList(newList){ safeNotifyItemChanged(pos, PAYLOAD_EDIT_ACTION) }`) — the proper DiffUtil-friendly refresh that actually re-renders the new image.

**Concerns**
1. **Direct overlap with #15383 (already `ready`).** Both PRs edit `VoicesAdapter.areContentsTheSame` and the `PAYLOAD_EDIT_ACTION` handler. #15383 adds `avatar`/`imageUrls`/`images`/`replyTo` to `areContentsTheSame` (a superset of your `imageUrls`/`images`) plus a `getChangePayload`, and its `PAYLOAD_EDIT_ACTION` calls `setMessageAndDate`/`configureEditDeleteButtons`/`handleChat` — yours calls `setMessageAndDate`/`loadImage`/`showReplyButton`. They can't both merge cleanly. Please rebase onto #15383 and keep only what it doesn't already cover (your `loadImage` in the payload handler is the part that matters for images; the `areContentsTheSame` image fields are already in #15383).
2. **`BaseVoicesFragment.onDataChanged` reintroduces `notifyDataSetChanged()`** (was `submitList(currentList.toList())`). That's against the direction of this whole review wave (#15312/#15383/#15284 are *removing* `notifyDataSetChanged`), and on a `ListAdapter` it's fragile — `getItem()` still reads the differ's backing list, so it only "works" because items are mutated in place. Prefer the same new-object + `submitList` approach you already use in the adapter callback.
3. **Still broken in enterprises (maintainer-reported).** @J-S-webskas notes "It's not working in enterprises right now." The diff fixes the field-shadowing in `TeamsVoicesFragment` but there's no equivalent for the enterprises voices path — if it has the same `val adapterNews` shadowing (or a separate display path), it needs the same treatment. Please verify the enterprises case and extend the fix.

Minor: qlty flags the `areContentsTheSame` `&&` chain complexity — moot once #1 is reconciled.

**Verdict:** The plumbing (editPost→News?, field-assignment fix, new-object submitList) is the right fix, but it overlaps #15383, reintroduces `notifyDataSetChanged`, and doesn't yet cover enterprises — reconcile with #15383, drop the `notifyDataSetChanged`, and close the enterprises gap.

---

dogi:
**Re-review — 68/100 — the fix is now correct; the remaining issue is overlap with ready #15383**

Big progress since my last pass. Both of my earlier concerns are addressed:

- **`notifyDataSetChanged` is gone.** The edit callback now replaces the item in `currentList` and calls `submitList(newList) { safeNotifyItemChanged(pos, PAYLOAD_EDIT_ACTION) }`, so DiffUtil drives the update. ✓
- **The "not working in enterprises" report is addressed at the root.** `getDiscussionsByTeamIdFlow`'s `distinctUntilChanged` previously compared only `id` + `time`, which **suppressed image-only edits** on the team/enterprise discussion flow — that's exactly why the new image didn't appear there. Now it also compares `message`/`isEdited`/`imageUrls`/`images`/`labels`, so the edit re-emits. This is the core fix for #15253, and it's the right place for it. ✓
- Nice touch: the `getChangePayload` buckets provably cover every field in `areContentsTheSame`, with the "empty → return null (full rebind)" guard and comment. Removing the spurious `@RequiresApi(UPSIDE_DOWN_CAKE)` on `onBindViewHolder` is also correct — nothing in there needs API 34.

**Remaining blocker — overlap with #15383 (`ready`).**
#15383 is a focused, already-`ready` PR whose entire content is adding `getChangePayload` + the flatMap `onBindViewHolder` payload handling to `VoicesAdapter.kt`. This PR now contains that same infrastructure, so the two will conflict on `VoicesAdapter`. They also differ on the no-op default (this PR returns `null` → full rebind; #15383 falls back to `PAYLOAD_REPLY_COUNT`).

**Recommendation:** let #15383 land first (it's the payload-infrastructure change), then rebase this PR on top and **drop the duplicated `getChangePayload`/`onBindViewHolder` edits**, keeping only what's unique and essential to #15253:
- `VoicesRepositoryImpl` `distinctUntilChanged` image comparison (the real fix),
- `editPost` returning `News?`,
- the `VoicesActions`/`TeamsVoicesFragment` wiring and the `submitList`-based item replacement.

That removes the conflict and keeps each PR's responsibility clean.

**One thing to double-check** while rebasing: the edit callback calls `submitList(newList)`, but `VoicesAdapter` overrides `submitList` to route through `prepareSubmitList`. Confirm that in the discussion/parent-thread context this doesn't re-inject/reorder unexpectedly (top-level lists with `parentNews == null` are fine; just verify the reply view).

**Verdict:** Core fix is correct and the enterprise regression is genuinely resolved — but rebase onto #15383 and de-duplicate the payload code first. Changes requested.

---

Okuro3499 (approved):

dogi:
**Review — rating: 84/100**

Fixes the bug at its root rather than papering over it. The real culprit was the `distinctUntilChanged` predicates in `VoicesRepositoryImpl` — `getDiscussionsByTeamIdFlow` compared only `id` and `time`, so an edit that changed images (or even the message) was swallowed and the flow never re-emitted. Widening both predicates to cover `message`/`isEdited`/`imageUrls`/`images`/`labels` is the correct fix.

I checked the risky part of this change: adding `getChangePayload` converts what used to be full rebinds into partial ones, so any under-covered payload bucket would leave stale UI. It holds up — `areItemsTheSame` keys on `id`, so `id` needs no bucket, and the remaining eleven fields in `areContentsTheSame` all map to a bucket. I verified each handler actually re-renders its bucket: `PAYLOAD_TEAM_LEADER_CHANGED` re-renders the label chips via `labelManager.showChips`, `PAYLOAD_USER_FETCHED` now also calls `loadImage`, and `PAYLOAD_EDIT_ACTION` now also calls `setMessageAndDate`/`handleChat`/`loadImage`. The `if (payloads.isNotEmpty()) payloads else null` fallback to a full rebind is the right defensive default for a future field added without a bucket, and the comment saying so earns its place. Dropping `@RequiresApi(UPSIDE_DOWN_CAKE)` from the `onBindViewHolder` override is also correct — minSdk is 26 and RecyclerView calls that method at every API level, so the annotation was wrong to begin with. CI green on build (default + lite) and test (default).

**Nit:** `editPost` re-reads via `newsDao.getById(newsId)` purely to return the entity it just mutated and upserted; returning `news` directly would save a query. Also, the `submitList { safeNotifyItemChanged(pos, PAYLOAD_EDIT_ACTION) }` pair double-dispatches a rebind for the edited row — harmless, just redundant. Neither is worth another round.

**Verdict:** Mergeable. The `distinctUntilChanged` widening is the kind of fix that deserves a repository-level test, but that's follow-up work, not a blocker.

---

PR #15260, opened 2026-08-04
resources: smoother repository shelf leaving (fixes #15255)

### LINKED ISSUE #15255: Experiencing a delay when leaving resources in bulk
## Description
When attempting to leave multiple resources in bulk, users experience a noticeable delay before the action completes. This negatively impacts the user experience, especially when managing a large number of resources.

## Steps to Reproduce
1. Navigate to the resources section.
2. Select multiple resources to leave in bulk.
3. Trigger the bulk "leave" action.
4. Observe the delay before the operation completes.
https://github.com/user-attachments/assets/de50db0c-420e-4dca-9d6c-4dd0c4138d95

Okuro3499:
fixes #15255
[Screen_recording_20260804_170036.webm](https://github.com/user-attachments/assets/07c5fb26-0342-4101-a64e-6ca7e7c45028)

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6319
-        versionName = "0.63.19"
+        versionCode = 6320
+        versionName = "0.63.20"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/data/room/AppDatabase.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/data/room/AppDatabase.kt
@@ -125,7 +125,7 @@ import org.ole.planet.myplanet.model.UserEntity
         Answer::class,
         MyTeam::class,
     ],
-    version = 5,
+    version = 6,
     exportSchema = false,
 )
 @TypeConverters(Converters::class)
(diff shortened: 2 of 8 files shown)


dogi:
**Review — rating: 80/100**

**Strengths**
- Directly addresses the leave-delay (#15255): the old `itemsToDelete.forEach { removeResourceFromShelf(it, userId) }` did N separate DAO round-trips (query + upsert + log each); `removeResourcesFromShelf` collapses it to one `getByResourceIds` + one `upsertAll` + one `insertAll`. The new `Index("resourceId")` directly serves the `WHERE resourceId IN (…)` batch query, so this index is well-justified (unlike a speculative one).
- Wrapped in `runCatching` returning `Result`, empty-list/blank-user guarded, and the fragment keeps the `withContext(main)` refresh **after** the batched call completes — so it avoids the refresh-before-delete race that a sibling PR fell into.
- Two good tests: 50 items → exactly one `getByResourceIds`/`upsertAll`/`insertAll` and zero per-item `upsert`; empty list is a no-op. CI green including `test (default)`.

**Concerns**
- **Version conflict with #15276 (both bump `AppDatabase` `version = 5 → 6`).** #15276 adds indices to 8 other entities at v6; this adds the `resourceId` index at v6. They can't both be v6 — whichever lands second must go to v7 (and under `fallbackToDestructiveMigration`, each bump wipes unsynced local data, so try to land index changes together in a single bump).
- **Confirm RemovedLog equivalence.** The batched path explicitly inserts a `RemovedLog(type = "resources")` per id. Verify the old `removeResourceFromShelf`/`updateUserLibrary(..., add = false)` also logged removals (so the server-sync of "user removed this" is unchanged) — if the old path didn't log, this is a behavior addition; if it did, this is a faithful batch of it.

**Verdict:** Correct, well-tested batching that fixes the delay with a justified index; ready to merge once the v6 collision with #15276 is reconciled and the RemovedLog behavior is confirmed equivalent.

---

dogi (approved):

PR #15267, opened 2026-08-04
prevent download popup dialog cropping when text size is large (fixes #15263)

### LINKED ISSUE #15263: download pop up window in text size large is cropped
when i change the text size to large, the download pop up (DOWNLOAD SELECTED, CANCEL, DOWNLOAD ALL) is cropped

https://github.com/user-attachments/assets/baa8ccf1-9cd2-4b35-ad72-b924d0e012b1

ragilzakaria:
https://github.com/user-attachments/assets/9a43cf18-248e-48cb-899a-3c1b0a327da7

fixes #15263

## Summary by CodeRabbit

* **New Features**
  * Added bulk downloading for selected library resources.
  * Download actions now appear contextually based on selected items and offline status.

* **Bug Fixes**
  * Prevented duplicate download dialogs and improved dialog replacement and dismissal handling.
  * Improved download action behavior and button enablement.

* **Style**
  * Added a themed download dialog with clearer button wrapping, sizing, and colors.
  * Refined library selection controls, spacing, icons, and responsive landscape layouts.


J-S-webskas (dismissed):

Okuro3499 (dismissed):
dialog flickers I am not able to click anything

dogi:
**Review — Rating: 64/100**

The targeted download-dialog work is genuinely good, but the styles change quietly restyles **every dialog in the app**.

**Strengths (correctly scoped to the download dialog)**
- `BaseResourceFragment` — the `if (downloadSuggestionDialog?.isShowing == true) return` guard fixes the duplicate-dialog bug, and the dialog-lifecycle refactor is careful and correct: keeping `oldDialog`, guarding `onDismissListener` with `if (downloadSuggestionDialog === dialog)` so the old dialog's dismissal can't null the *new* reference, moving the positive-button enablement into `setOnShowListener` (the button exists by then), and nulling the old listener before dismissing. Nicely done.
- Density-scaled title padding + `textSize 15f` / `maxLines 4` directly address the title cropping.

**Concern — app-wide scope creep (blocking):**
- `styles.xml` changes `AlertDialogTheme`'s parent from `ThemeOverlay.AppCompat.Dialog.Alert` → `ThemeOverlay.MaterialComponents.Dialog.Alert`. A code search shows `AlertDialogTheme` is applied to **30+ dialogs across the app** (Settings, TeamFragment, MyHealthFragment, LoginActivity, BasePermissionActivity, VoicesAdapter, ExamTakingFragment, DialogUtils, ProcessUserDataActivity, StorageBreakdownFragment, …), including two that feed it into afollestad `MaterialDialog.Builder` via `ContextThemeWrapper`. Switching the base dialog theme changes button casing/color, title typography, padding, and window insets for **all** of them.
- Similarly, adding `textSize 13sp` + `maxLines 2` to the shared `NegativeButtonStyle` (used as both `buttonBarNegativeButtonStyle` and `buttonBarPositiveButtonStyle`) shrinks and rewraps **every** dialog button in the app, not just this one.
- CI can't catch visual regressions here, so this needs either narrowing or broad manual QA.

**Ask:** scope the button-wrapping and any theme change to a **dedicated theme** used only by the download suggestion dialog (e.g. a `DownloadDialogTheme` extending the existing one), rather than mutating the shared `AlertDialogTheme`/`NegativeButtonStyle`. If the app-wide switch to Material dialog styling is intentional, please say so explicitly and verify a representative set of dialogs (settings, team create, health member, login language, a permission rationale) at both normal and large font sizes.

Minor: qlty flags `showDownloadDialog` complexity (19) — the added lifecycle handling pushes it up; extracting the dialog setup would help.

**Verdict:** Solid, well-scoped fix to the download dialog itself, but it rides along with an app-wide dialog restyle that's out of scope and untested — narrow it (or justify + QA it) for another round.

---

dogi:
**Re-review — 85/100 — my scope concern is fully addressed**

The app-wide theming concern from my last pass is resolved. Instead of changing the shared `AlertDialogTheme` parent and the shared button style (which would have restyled every dialog in the app), you added **dedicated, scoped** styles used only by this dialog:
```xml
<style name="DownloadDialogTheme" parent="ThemeOverlay.MaterialComponents.Dialog.Alert"> … </style>
<style name="DownloadDialogButtonStyle" parent="Widget.MaterialComponents.Button.TextButton.Dialog"> … </style>
```
and applied them via `AlertDialog.Builder(fragmentActivity, R.style.DownloadDialogTheme)`. Nothing shared is mutated — exactly the right fix. The `maxLines=2` / `singleLine=false` on the button style is what lets the button labels wrap instead of cropping at large text sizes (#15263). ✓

**Also verified**
- Dialog de-duplication is handled carefully: the early-out `downloadSuggestionDialog?.isShowing == true`, plus capturing `oldDialog`, calling `oldDialog?.setOnDismissListener(null)` before dismissing it, and guarding the new dialog's `onDismiss` with `if (downloadSuggestionDialog === dialog)`. That correctly avoids the stale-dismiss nulling the current dialog reference — good attention to the race. ✓
- Positive-button enablement moved into `setOnShowListener` with a null-safe `getButton(...)?.isEnabled`, replacing the old non-null `getButton(...).isEnabled`. Cleaner and avoids a potential NPE. ✓
- `createDownloadDialogTitleView` now scales padding by `displayMetrics.density` (dp→px) instead of the old raw pixel values — correct. The `handleDownloadSelected`/`handleDownloadAll` extractions are behavior-preserving. ✓

CI green across all three jobs.

**Minor (non-blocking):** `my_library_alertdialog.xml` changes the list from `wrap_content` (max 300dp) to a fixed `120dp`. That reliably leaves room for the wrapped buttons, but it also shortens the list noticeably even when there's space for more rows. A constrained `wrap` with a smaller `layout_constraintHeight_max` might keep the anti-crop guarantee while showing more items when they fit. Your call.

**Verdict:** Scoped, careful, CI-green. Mergeable.

---

Okuro3499 (commented):
@ragilzakaria 
in portrait mode this is what we had in master
Image

do you think it is better cropped as in this branch though the flickering stopped
Screenshot_20260812_155629

ragilzakaria:
Updated

1. Portrait Mode (my_library_alertdialog.xml):
- Removed the fixed 120dp height constraint.
- Set android:layout_height="wrap_content", app:layout_constrainedHeight="true", and app:layout_constraintHeight_max="360dp".
- The dialog now shows items at full height for typical lists and smoothly scrolls up to 360dp when many items are present, eliminating the shrunken window appearance.

2. Landscape Mode (my_library_alertdialog.xml):
- Created a dedicated landscape layout with app:layout_constraintHeight_max="130dp" and compact padding.
- Ensures the dialog title, scrollable list items, and all action buttons fit comfortably on screen without vertical clipping.

dogi:
**Review — rating: 88/100**

Fixes the cropping at its actual causes rather than papering over it. The old title used `setPadding(48, 40, 48, 0)` in raw pixels, which is a density bug independent of font scale; converting to density-scaled dp plus dropping 18f→15f and the new `DownloadDialogButtonStyle` (`singleLine=false`, `maxLines=2`) is what lets "Download selected"/"Download all" survive a large system text size instead of being ellipsized. The `layout-land` variant capping the list at 130dp and the portrait bump to 360dp handle the other axis.

I checked the dialog-lifecycle rework against `BaseResourceFragment.createListView`: it installs no show/dismiss listeners of its own, only the `CheckboxAdapter` callback that re-toggles the positive button on each selection change, so the new `setOnShowListener`/`setOnDismissListener` clobber nothing and the per-selection enable path is intact. The reordering is also a genuine fix — the new dialog is created before the old one is dismissed, so clearing the old listener and guarding the dismiss with `downloadSuggestionDialog === dialog` is what stops the old dialog's dismiss from nulling the new dialog's reference. Moving the initial enable state into `setOnShowListener` is strictly better than the old post-`show()` call. Build (default + lite) and test (default) all green.

**Nit:** `DownloadDialogTheme` sets `android:background` on a `ThemeOverlay`, which is inherited by child views that don't declare their own background; `android:windowBackground` is the targeted attribute for a dialog surface.

**Verdict:** mergeable.

---

Okuro3499 (changes requested):
@ragilzakaria 
the 3 buttons on the dialog not showing
https://github.com/user-attachments/assets/c5c37a86-dcc5-468e-92fc-a2143f4a184c

Since the dialog just duplicates the list shown in the fragment, I think we should move away from it. We could instead add a 'Download' button that appears when a not-yet-downloaded item is selected, and clicking it would download all selected resources.

ragilzakaria:
https://github.com/user-attachments/assets/440ec5c1-2e12-4fec-9f7b-885b0a1ac244

@Okuro3499

dogi:
**Review — rating: 60/100**

The cropping fix at the centre of this is correct and I verified it: `my_library_alertdialog.xml` gains `layout_constraintBottom_toBottomOf` so the `RecyclerView` is actually bounded rather than free-running, the new `layout-land/my_library_alertdialog.xml` caps it at 130dp for the short landscape viewport, and `DownloadDialogTheme` drops the title from 18sp to 15sp with density-scaled padding while `DownloadDialogButtonStyle` sets `maxLines=2` + `singleLine=false` so the three action buttons wrap instead of truncating at large font scales. The old code hardcoded `setPadding(48, 40, 48, 0)` in raw pixels, so converting to `density`-multiplied dp is a genuine correctness fix, not just tidying. Guarding re-entry with `downloadSuggestionDialog?.isShowing == true` and nulling the old dialog's dismiss listener before dismissing it (so the stale listener can't null out the new dialog's reference) is a real fix for a dialog-stacking race. Extracting `handleDownloadSelected`/`handleDownloadAll` is a clean readability win. CI green (build default + lite, test default).

**The blocker is a regression bundled in with it.** `BaseRecyclerFragment.onLoad()` loses both `showDownloadDialog` triggers, and only one of them gets a replacement:

- The `courseLib == null` branch is replaced by the new explicit download icon in `fragment_my_library.xml` — fine, that's the intended redesign.
- The `courseLib == "courses"` branch is not. `courseLib` is set in exactly one place, `CoursesFragment.kt:170`, so that block *was* the courses screen's download-suggestion prompt. `CoursesFragment` inflates `fragment_my_course.xml`, which this PR does not touch and which has no `tv_download`. So the courses screen loses the prompt and gains no button — `CoursesFragment.showDownloadDialog` (line 607) is now dead from this path, reachable only via the sync broadcast receiver in `BaseResourceFragment` and the `BaseDashboardFragment`/`BaseContainerFragment` entry points.

Three more things, none individually blocking but they compound:

- **O(n·m) on the main thread.** Both `hideButton()` and `downloadSelectedResources()` run `allResourceModels.find { … }` inside a per-selected-item `any`/`filter`. `hideButton()` fires on every selection change, so Select All over a large library is quadratic on the UI thread. The two lookups are also byte-for-byte duplicates — worth one extracted helper backed by a `Map<String, ResourceListModel>` built once.
- **The match predicate mixes two identities.** `it.library.resourceId == item.resourceId || it.item.id == item.resourceId` will match a model whose `item.id` collides with a different model's `resourceId`. Pick one key.
- **Touch targets and labels.** `tv_add`/`tv_delete` go from labelled `TextView`s to 40dp icon-only `ImageView`s. 40dp is under the 48dp Material minimum, and the visible text is gone — `contentDescription` covers TalkBack but not sighted users who relied on "Remove selected".

**Nit:** `downloadSelectedResources()` falls back to `targetList = selected` when nothing is un-downloaded, but `hideButton()` only shows the button when something *is* un-downloaded, so that branch can only re-download already-offline items. Also undeclared: `hideButton()` now hides the whole search pill whenever anything is selected.

**Verdict:** needs another round — the dialog fix is good and I'd take it as-is, but restore the courses prompt (or give `fragment_my_course.xml` the same download button) before this lands. Splitting the cropping fix from the selection-bar redesign would get the first half merged today.

---

Okuro3499 (approved):

J-S-webskas (approved):

PR #15269, opened 2026-08-04
resources: smoother search utils caching (fixes #15505)

### LINKED ISSUE #15505: Optimize `ResourceSearchUtils` by caching normalized titles to fix search lag
The `ResourceSearchUtils.searchList` method calls `Utilities.normalizeText` on every item in the list for every character typed during a search, causing significant UI lag with a full resource library.

This patch optimizes the search by caching the normalized titles in `ResourceSearchUtils` tied to the list's identity (using `WeakReference` to prevent memory leaks), and recognizing that `Utilities.normalizeText` already outputs lowercased strings, so the slow locale-aware `ignoreCase` comparison is unnecessary.

dogi:
The `ResourceSearchUtils.searchList` method was calling `Utilities.normalizeText` on every item in the list for every character typed during a search, causing significant UI lag with a full resource library.

This patch optimizes the search by:
1. Caching the normalized titles in `ResourceSearchUtils` tied to the list's identity (using `WeakReference` to prevent memory leaks).
2. Recognizing that `Utilities.normalizeText` already outputs lowercased strings, so the slow locale-aware `ignoreCase = true` flag on `startsWith` and `contains` string methods has been removed.

These changes satisfy the requirements to hoist normalization so it happens once per list rather than once per keystroke, without introducing new utility classes.

---
*PR created automatically by Jules for task [13484538397800670886](https://jules.google.com/task/13484538397800670886) started by @dogi*

## Summary by CodeRabbit

* **Performance**
  * Improved resource search responsiveness by reducing repeated title processing during searches.
  * Search results maintain the same matching behavior for prefixes and words within titles.


dogi:
**Review — rating: 74/100**

**Strengths**
- The scary-looking part is actually safe: dropping `ignoreCase = true` from `startsWith`/`contains` is a **no-op** because `Utilities.normalizeText` lowercases (`Normalizer.normalize(str.lowercase(...), NFD)…`, verified by `normalizeText("HELLO World") == "hello world"`). Both the cached titles and `normalizedQuery`/`normalizedQueryParts` go through `normalizeText`, so they're already lowercase — case-insensitive matching is preserved.
- The core win is real: for search-as-you-type over a stable list, normalizing every title on every keystroke is wasteful; caching `cachedTitles` keyed by list identity and reusing across queries eliminates the repeated `normalizeText` work. `getOrNull(i)` guards the index access.

**Concerns**
- **Global mutable static state on a previously-stateless `object`.** `cachedListRef`/`cachedTitles` are shared across *all* callers of `searchList<T>` with no synchronization. Two problems: (1) **thread-safety** — if `searchList` is ever invoked off the main thread or concurrently for two different lists, the fields race (one caller sets `cachedTitles` for list B while another indexes it against list A → mismatched/`continue`d titles). The old function was pure/thread-safe. (2) **cross-caller thrash** — interleaved searches over different lists constantly invalidate the single cache slot, so no reuse there. Please confirm search is single-threaded, or guard the cache (or make it per-caller).
- **In-place mutation staleness.** Invalidation is by reference identity (`!== list`), so if a caller ever reuses the same `MutableList` instance and mutates its contents between searches, the cached titles go stale (wrong/misaligned results). Safe only if the list is *replaced* with a new instance on every data change — worth verifying at the call sites.

**Verdict:** Correct for the intended single-thread, stable-list search-as-you-type path (and the `ignoreCase` removal is verified safe), but it trades a pure utility for global mutable cache state with latent thread-safety and in-place-mutation hazards. Mergeable if search is confirmed single-threaded and the list is replaced (not mutated) on refresh; otherwise scope the cache to the caller/ViewModel.

---

Okuro3499 (dismissed):
```
Global, unsynchronized, single-slot cache keyed only by list identity (not by titleSelector) 
  inside a shared singleton object. Fix: either (a) scope the cache to the caller (e.g., have ResourcesFragment precompute/cache normalized titles alongside allResourceModels,      
  matching how allResourceModels is already refreshed as a whole), or (b) adopt the existing titleNormal pattern by threading MyLibrary.titleNormal through getResourceListModels →  
  ResourceItem and search on it directly, removing the need for any cache in ResourceSearchUtils.
```

dogi:

The implementation was reworked and it resolves my earlier concerns entirely:

- ✅ **No more global mutable cache.** The `WeakReference` `cachedListRef` / `cachedTitles` static state (with its thread-safety and cross-caller-thrash hazards) is gone. `searchList` is a pure function again — it takes a `normalizedTitleSelector` and does no caching of its own.
- ✅ **The real win now comes from the persisted column.** `searchLocalModels` supplies `it.library.titleNormal ?: Utilities.normalizeText(it.item.title ?: "")`, so it reuses the normalized title computed once at insert time and only normalizes on the fly as a fallback. That eliminates the per-keystroke `normalizeText` cost without any shared state.

Correctness holds: dropping `ignoreCase` from `startsWith`/`contains` is a verified no-op because `titleNormal` (and the fallback `normalizeText`) and `normalizedQuery`/`normalizedQueryParts` are all lowercased by `normalizeText`. Build (both flavors) + test green.

**Minor (non-blocking):**
- The updated test mocks `titleNormal` as `null`, so it only exercises the fallback path (equivalent to old behavior — good, existing assertions still hold), but doesn't cover the `titleNormal`-populated branch. A second case with a non-null `titleNormal` would lock in the fast path.
- Assumes `library.titleNormal` corresponds to `item.title` for the same model (they represent the same resource, so this should hold) — worth a mental note.

This is also the substantive ResourceSearchUtils fix I referenced when reviewing #15348 (whose `.distinct()` should fold in here).

**Verdict:** Clean, stateless optimization using the persisted normalized title — mergeable.

---

Okuro3499 (approved):

PR #15270, opened 2026-08-04
sync: smoother user repository inserting (fixes #15432)

### LINKED ISSUE #15432: Optimize insertUsersFromSync with linear matching

dogi:
Implements linear map-based matching to optimize `insertUsersFromSync` in `UserRepositoryImpl.kt`. Replaces quadratic O(N^2) list operations with O(1) map lookups (`usersById`, `guestUsersByName`) and uses an insertion-ordered map (`usersToUpsert`) for accumulating upserts, ensuring duplicate documents within a batch retain the correct last-write behavior. Added coverage for these cases in `UserRepositoryBulkInsertTest.kt`.

---
*PR created automatically by Jules for task [8066325465883263555](https://jules.google.com/task/8066325465883263555) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6341
-        versionName = "0.63.41"
+        versionCode = 6342
+        versionName = "0.63.42"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/repository/UserRepositoryImpl.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/UserRepositoryImpl.kt
@@ -1241,35 +1241,54 @@ class UserRepositoryImpl @Inject constructor(
             }
         }
 
-        val existingUsers = userDao.getAll().toMutableList()
+        val existingUsersList = userDao.getAll()
+        val usersById = mutableMapOf<String, UserEntity>()
+        val guestUsersByName = mutableMapOf<String, UserEntity>()
+
+        for (user in existingUsersList) {
+            usersById[user.id] = user
+            if (user._id != null) usersById[user._id!!] = user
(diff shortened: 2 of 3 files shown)


dogi:
**Review — rating: 83/100**

**Strengths**
- Real complexity fix: the old loop did `existingUsers.firstOrNull{…}`, `usersToUpsert.removeAll{…}`, and `existingUsers.removeAll{…}` per document — O(n·m). The new version builds `usersById` (keyed by both `id` and `_id`) and `guestUsersByName` maps plus a `linkedMap` for upserts, making every lookup/dedup O(1).
- Behavior preserved on the paths I traced: `usersById[id]` reproduces `firstOrNull { it.id == id || it._id == id }`; `guestUsersByName[userName]` reproduces the guest-by-name lookup; guest promotion still adds to `usersToDelete` and now also evicts the guest from both maps so it can't re-match; `usersToUpsert[entity.id] = entity` keeps last-wins-by-id with insertion order; and the per-doc map updates mirror the old `existingUsers`/`usersToUpsert` list mutations. Dropping `?: continue` is safe since the `?: UserEntity()` fallback makes `user` non-null.
- Two well-chosen tests: guest promotion (asserts `deleteByIds(["guest_123"])` + upsert of the promoted `_id`) and same-`_id` dedup (last version wins). CI green including `test (default)`.

**Concerns**
- **Overlaps #15274, which also rewrites `insertUsersFromSync`** (it narrows `getAll()` → `getByIdsOr_IdsOrNames`). These conflict on the same method. This PR keeps `getAll()` and just optimizes the matching — safer and better-tested than #15274's narrowing (which I flagged for potential missed-user issues). Recommend keeping **this** approach for `insertUsersFromSync` and dropping that hunk from #15274.
- Minor edge (pre-existing ambiguity): if two distinct users share an id/_id value, or two guests share a name, the map keeps the last-inserted while the old `firstOrNull` kept the first. Unrealistic for unique user ids, just noting the tie-break changed.

**Verdict:** Clean, correct, well-tested O(n+m) rewrite; ready to merge and preferable to #15274's version of this method.

---

Okuro3499 (approved):

PR #15274, opened 2026-08-04
Refactor UserRepositoryImpl to use targeted UserDao queries

dogi:
Optimized data access in `UserRepositoryImpl` by avoiding full-table scans via `userDao.getAll()` and translating the filtering logic into specific SQLite `@Query` methods in `UserDao`. This improves memory efficiency and query performance.

---
*PR created automatically by Jules for task [15197987516865157169](https://jules.google.com/task/15197987516865157169) started by @dogi*

## Summary by CodeRabbit

* **Improvements**
  * Improved the speed and reliability of user lookups and synchronization.
  * Enhanced handling of users identified by local IDs, remote IDs, or names.
  * Improved processing of pending and completed synchronization records.
  * Strengthened duplicate-user detection and cleanup.
  * Improved migration and matching of guest user records.

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/data/room/dao/LegacyEntityDaos.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/data/room/dao/LegacyEntityDaos.kt
@@ -18,6 +18,22 @@ interface UserDao {
     @Query("SELECT * FROM users WHERE id = :id OR _id = :id LIMIT 1")
     suspend fun getById(id: String): UserEntity?
     @Query("SELECT * FROM users") suspend fun getAll(): List<UserEntity>
+    @Query("SELECT * FROM users WHERE id IN (:ids) OR _id IN (:ids)")
+    suspend fun getByIdsOr_Ids(ids: List<String>): List<UserEntity>
+    @Query("SELECT * FROM users WHERE _id IS NOT NULL AND _id != '' AND id NOT LIKE 'guest%'")
+    suspend fun getSyncedUsers(): List<UserEntity>
+    @Query("SELECT * FROM users WHERE _id IS NOT NULL AND _id != ''")
+    suspend fun getUsersWith_Id(): List<UserEntity>
+    @Query("SELECT * FROM users WHERE _id IS NULL OR _id = '' OR isUpdated = 1 LIMIT :limit")
+    suspend fun getPendingSyncUsers(limit: Int): List<UserEntity>
+    @Query("SELECT * FROM users WHERE id IN (:ids)")
+    suspend fun getByIds(ids: List<String>): List<UserEntity>
+    @Query("SELECT * FROM users WHERE name IN (SELECT name FROM users GROUP BY name HAVING COUNT(*) > 1)")
+    suspend fun getDuplicateUsers(): List<UserEntity>
+    @Query("SELECT * FROM users WHERE id IN (:ids) OR _id IN (:ids) OR name IN (:names)")
+    suspend fun getByIdsOr_IdsOrNames(ids: List<String>, names: List<String>): List<UserEntity>
+    @Query("SELECT * FROM users WHERE id = :id OR _id = :id LIMIT 1")
+    suspend fun getByIdOr_Id(id: String): UserEntity?
     @Query("SELECT * FROM users WHERE name = :name LIMIT 1") suspend fun getByName(name: String): UserEntity?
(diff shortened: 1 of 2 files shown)


dogi:
**Review — rating: 62/100**

The straightforward pushdowns are good — `getUsersByIds` → `getByIdsOr_Ids`, `getUserMap` → `getByIds`, `getSyncedUsers`, `getUsersForHealthSync`, `getPendingSyncUsers` all faithfully move `getAll().filter{}` into SQL and are equivalent for realistic data. But a couple of changes go beyond translation and need verification, plus some edge divergences.

**Concerns (verify)**
- **`buildUserFromJson` guest migration is reimplemented, not pushed down.** When `users == null`, the old path called `migrateGuestUser(id, userName, availableUsers)`; the new code inlines a *different* implementation (`userDao.getByName(userName)?.takeIf { _id startsWith "guest_" }?.let { deleteById; reassign id }`). Two risks: (1) it may not replicate everything `migrateGuestUser` does; (2) `getByName` returns the **first** row for that name (`LIMIT 1`) and then filters to guest — if the first same-name user isn't the guest but a later one is, the guest is missed. This is a critical login/sync path; please make the `users == null` branch call the same `migrateGuestUser` logic (e.g. `migrateGuestUser(id, userName, listOf(userDao.getByIdOr_Id(id) ?: …))`) rather than a parallel impl.
- **`insertUsersFromSync` narrows `existingUsers` from all users to only id/_id/name matches of the incoming batch.** That's a good optimization *if* `existingUsers` is used solely to match incoming docs. But it feeds `usersToDelete`/`usersToUpsert`/dedup — confirm none of that logic needs to see users outside the incoming batch (e.g. global duplicate cleanup or orphan handling), or it will silently stop touching them.

**Minor edge divergences (likely fine, worth noting)**
- `id NOT LIKE 'guest%'` is **case-insensitive** in SQLite, while `!id.startsWith("guest")` is case-sensitive — differs only for mixed-case guest ids (realistic ones are lowercase).
- `_id != ''` vs `!_id.isNullOrBlank()`: SQL keeps whitespace-only `_id`, Kotlin treated it as blank. Unrealistic for `_id`.
- `getDuplicateUsers`: `name IN (SELECT … HAVING COUNT>1)` won't match **null-name** duplicates (`NULL IN (…)` is never true), whereas the old `groupBy { name }` did. Only matters if users can have null names.

**Verdict:** Right idea and most of it is a clean win, but the reimplemented guest-migration branch and the `insertUsersFromSync` narrowing are behavior-sensitive and unverified. Align the guest migration with `migrateGuestUser`, confirm the sync narrowing is safe (ideally add a test), and this is mergeable.

---

dogi:

Both behavior-sensitive concerns from my earlier review are now resolved, verified against the head file:

- ✅ **Guest migration in `buildUserFromJson` is no longer a parallel reimplementation.** It now calls the real `migrateGuestUser(id, userName, users ?: userDao.getAllByName(userName))`. Crucially, `getAllByName` returns **all** rows for that name (not `LIMIT 1`), and `migrateGuestUser` only does `users.firstOrNull { it.name == userName && it._id?.startsWith("guest_") == true }` — so the guest record is reliably found among same-name users. This closes both the "parallel impl" and "getByName LIMIT 1 misses the guest" worries.
- ✅ **`insertUsersFromSync` narrowing is safe.** I traced how `existingUsers` (now `getByIdsOr_IdsOrNames(ids, names)`) is used: only `firstOrNull { it.id == id || it._id == id }` (id from the incoming doc, ∈ `ids`) and the guest lookup `firstOrNull { it.name == userName && _id startsWith "guest_" }` (name ∈ `names`). Every key is drawn from the incoming batch, so the targeted query returns exactly the rows needed. `usersToDelete` only ever holds guests found within that matched set — there's no global cleanup that would need users outside the batch. The optimization is correct.

The straightforward pushdowns (`getByIdsOr_Ids`, `getByIds`, `getSyncedUsers`, `getUsersWith_Id`, `getPendingSyncUsers`, `getDuplicateUsers`) remain faithful translations. Build (both flavors) + test green.

**Minor edge divergences (all unrealistic, non-blocking):** `id NOT LIKE 'guest%'` is case-insensitive vs `startsWith` case-sensitive; `_id != ''` keeps whitespace-only ids; `getDuplicateUsers` won't group null-name duplicates. None matter for real data — a unit test around the guest-migration + sync path would still be nice insurance.

**Verdict:** The sensitive paths are now correct (one fixed in code, one verified safe) — mergeable.

---

Okuro3499 (changes requested):
LegacyEntityDaos was removed so making this an invalid pr

PR #15275, opened 2026-08-04
teams: smoother repository dao querying (fixes #15457)

### LINKED ISSUE #15457: Refactor TeamsRepositoryImpl to push down getAll() filtering to SQL
Tracking issue for PR #15275.

dogi:
Optimization for `TeamsRepositoryImpl` to prevent `teamDao.getAll()` full-table memory loading at multiple sites.
- Added 6 new `@Query` DAO methods into `TeamDao`.
- Applied these optimized queries across 8 specific target call sites in `TeamsRepositoryImpl.kt` replacing `.filter { ... }`.

---
*PR created automatically by Jules for task [17244291535705355953](https://jules.google.com/task/17244291535705355953) started by @dogi*


dogi:
**Review — rating: 58/100**

The idea — replace `teamDao.getAll().filter { … }` (load-all-then-filter-in-Kotlin) with targeted SQL — is exactly right and most of the translations are faithful. But two of them change behavior for teams with a null `_id`.

**Correctly translated**
- `getUpdatedTeams` (`WHERE isUpdated = 1`), `getActiveRootTeams`, `getResourceLink` — equivalent.
- Null-`status` handling is right: Kotlin `status != "archived"` includes null status, and `IFNULL(status,'') != 'archived'` reproduces that. `getRootTeamsByType`/`getShareableEnterprises`/`getTeamDetails` are fine.

**Concerns**
- **`getRootTeamsByTypeAndIds` — wrong id fallback (regression).** Kotlin filtered on `(it._id ?: it.id) in memberIds` — i.e. use `_id`, but fall back to **`id`** when `_id` is null. The SQL uses `IFNULL(_id, '') IN (:teamIds)`, which falls back to **`''`**, not `id`. So a root team with `_id == null` but a populated `id` that's in `memberIds` is **included by the old code and excluded by the new query**. That's precisely the case for a locally-created team not yet synced (id = local UUID, `_id` still null) — it would silently drop out of the shareable/summary lists. Use `IFNULL(_id, id) IN (:teamIds)`.
- **`teamNameExists` — same `_id` fallback bug, with a worse symptom.** Kotlin's exclusion was `(it._id ?: it.id) != excludeTeamId`; the SQL is `IFNULL(_id, '') != :excludeTeamId`. For a team being edited that has `_id == null` and `id == excludeTeamId`, Kotlin correctly excludes it from the duplicate check, but the SQL's `'' != excludeTeamId` is true, so the team collides with **itself** → a spurious "name already exists" error when renaming a not-yet-synced team. Use `IFNULL(_id, id) != :excludeTeamId`.
- Minor: confirm `isRootTeam()` is `teamId.isNullOrBlank()` (the SQL `teamId IS NULL OR TRIM(teamId) = ''` assumes blank, not just empty). And `name = :name COLLATE NOCASE` is ASCII-case-insensitive vs Kotlin `equals(ignoreCase = true)` — equivalent for ASCII names, a corner case for non-ASCII.

**Verdict:** Right approach and a real perf win, but the `IFNULL(_id, '')` mistranslation in `getRootTeamsByTypeAndIds` and `teamNameExists` is a correctness regression for null-`_id` teams. Change both to `IFNULL(_id, id)` (and add a test with a null-`_id` local team) and this is good to go.

---

dogi:
**Re-review — Rating: 60/100** (was 58)

Thanks for engaging the `_id` fallback issue, but the change **doesn't actually fix it** — it reformats the expression while keeping the wrong fallback value.

The two queries now use:
```sql
(CASE WHEN _id IS NULL OR _id = '' THEN '' ELSE _id END)
```
That still falls back to `''`, exactly like the old `IFNULL(_id, '')`. The Kotlin it replaces was `it._id ?: it.id` — it falls back to **`id`**, not `''`. So the regressions I flagged remain:

- **`getRootTeamsByTypeAndIds`** — a root team with `_id == null` and `id` in `memberIds` yields `''`, which isn't in `memberIds`, so it's still excluded. A locally-created, not-yet-synced team (id = local UUID, `_id` null) silently drops out of the shareable/summary lists.
- **`teamNameExists`** — for a team being renamed where `_id == null` and `id == excludeTeamId`, the expression yields `''`, and `'' != :excludeTeamId` is true, so the team collides with itself → spurious "name already exists."

**Fix:** the `THEN` branch must be `id`, not `''`:
```sql
(CASE WHEN _id IS NULL OR _id = '' THEN id ELSE _id END)
```
(or simply `IFNULL(_id, id)` if you only need the null case). Apply it in both `getRootTeamsByTypeAndIds` and `teamNameExists`, and please add a test with a null-`_id` local team so this is locked in — CI is green only because no test exercises that row.

Everything else in the PR remains a correct, worthwhile push-filtering-to-SQL translation.

**Verdict:** Same correctness regression as before — the `CASE ... THEN ''` needs to be `THEN id`. One more pass.

---

dogi:
**Re-review — 88/100 — and a correction to my earlier review**

First, I need to retract my previous blocking flag. I had claimed the SQL matching on `_id` diverged from the original `(it._id ?: it.id)` for teams whose `_id` was null/empty. That was wrong: in `MyTeam`, `id` is not a separate field — it's a getter/setter alias for `_id` (`var id: String get() = _id; set(v){ _id = v }`), and `_id` is a non-nullable `@PrimaryKey String`. So `(it._id ?: it.id)` is just `_id`, the elvis is dead, and `_id IN (:teamIds)` / `_id != :excludeTeamId` are exact equivalents. Apologies for the noise — I should have checked the model before flagging.

Having re-verified all six new DAO queries against the code they replace, this is a faithful, behavior-preserving optimization:

- `getUpdatedTeams` → `WHERE isUpdated = 1` matches `.filter { it.updated }` (the `updated` property is `@ColumnInfo(name = "isUpdated")`). ✓
- `getActiveRootTeams` / `getRootTeamsByType` / `getRootTeamsByTypeAndIds` correctly render `isRootTeam()` as `(teamId IS NULL OR TRIM(teamId) = '')` — matching `teamId.isNullOrBlank()`, whitespace included. ✓
- Nice detail: `IFNULL(status,'') != 'archived'` preserves the Kotlin semantics where a null `status` is *not* archived — a plain `status != 'archived'` would have silently dropped null-status rows via SQL three-valued logic. ✓
- `teamNameExists` uses `COLLATE NOCASE` for the name compare and `(_id != :excludeTeamId OR :excludeTeamId IS NULL)`, which faithfully reproduces `name.equals(..., ignoreCase=true)` and the `excludeTeamId == null || ...` guard. (Minor: `NOCASE` only case-folds ASCII, whereas Kotlin `ignoreCase` is Unicode-aware — irrelevant for the team-name use case.) ✓

**Strengths:** removes eight full-table `getAll()` loads on the hot team/enterprise listing paths, net −25 lines, and pushes filtering into indexed SQL. Exactly the right direction. CI green across all three jobs.

**Verdict:** Correct and behavior-preserving — mergeable.

---

Okuro3499 (approved):

PR #15277, opened 2026-08-04
teams: smoother surveys repository scanning (fixes #15458)

### LINKED ISSUE #15458: Optimize survey ID scanning when grouping submissions
Tracking issue for PR #15277.

dogi:
This change optimizes the grouping logic in `SurveysRepositoryImpl.getSurveyInfos` by replacing separate `any` and `find` scans with a single pass using `mapNotNull` and a set-based lookup helper (`surveyIdsSet`). This eliminates redundant O(N^2) scans over the survey list for each submission, significantly reducing CPU usage when rendering the survey summary. Acceptance criteria verified via added unit tests covering exact, suffixed, and unrelated IDs.

---
*PR created automatically by Jules for task [13036236589438403664](https://jules.google.com/task/13036236589438403664) started by @dogi*


dogi:
**Review — rating: 85/100**

**Strengths**
- Real complexity win: the old code scanned `surveyIds` twice per submission (once in `filter { surveyIds.any {…} }`, again in `groupBy { surveyIds.find {…} }`). The new version does a single `mapNotNull` pass with a `surveyIdsSet`-backed resolver — O(1) for the common exact-match case, falling back to the O(surveyIds) suffix scan only when there's no exact match.
- **Behavior preserved, including the subtle bits I checked:**
  - Kotlin's `Iterable.toSet()` returns a `LinkedHashSet`, so `surveyIdsSet.find { pId.startsWith("$it@") }` still iterates in the original list order — the old `surveyIds.find {…}` order is maintained.
  - The exact-match-first vs suffix-match ordering could only diverge if a survey id itself contained `@` (so a parentId could exact-match one id while being a suffix-child of another). Survey ids don't carry `@` (that's the submission `parentId@team` convention), so the two are equivalent on real data.
  - The `isComplete` gate (`complete`/`requires grading`) and null-parentId handling are unchanged.
- Good, targeted test: covers exact id, suffixed id, unrelated, incomplete (excluded), the `survey11`-not-matching-`survey1` overlap trap, and a second suffix — and asserts the resulting keys/size. CI green including `test (default)`.

**Concerns**
- None blocking. (If you ever expect survey ids to legitimately contain `@`, revisit the exact-first ordering — but that's not the case today.)

**Verdict:** Correct single-pass optimization with the ordering/exact-vs-suffix subtleties preserved and a solid regression test; ready to merge.

---

Okuro3499 (approved):

PR #15278, opened 2026-08-04
chat: smoother history view binding (fixes #15403)

### LINKED ISSUE #15403: Fix shared ViewBinding and unsafe cast in ChatHistoryAdapter

dogi:
Fixes a correctness bug in `ChatHistoryAdapter` where the ViewBinding was stored as an adapter-level property, causing shared state among ViewHolders. Also addresses an unsafe type cast in `bindShareChat`.

---
*PR created automatically by Jules for task [13324263415112653590](https://jules.google.com/task/13324263415112653590) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6316
-        versionName = "0.63.16"
+        versionCode = 6317
+        versionName = "0.63.17"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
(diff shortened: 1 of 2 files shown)


dogi:
**Review — rating: 84/100**

**Strengths**
- **Real ViewHolder-binding bug fixed.** `private lateinit var rowChatHistoryBinding` was a single adapter-level field reassigned on every `onCreateViewHolder`, so it only ever held the most-recently-created row's binding — any use of the field would reference the wrong view. Making it a local `val` passed into `ViewHolderChat` is the correct pattern (each holder owns its binding). CI green confirms the field had no other references, so this is a safe removal of a latent bug.
- **Unsafe cast hardened.** `getData() as HashMap<String, List<String>>` would throw `ClassCastException` on any non-`HashMap` map (e.g. `LinkedHashMap`) or NPE on null; `as? Map<String, List<String>> ?: emptyMap()` casts to the general interface and degrades to empty. `generateFlatList` still accepts it (build green), so it's a strictly safer read.

**Concerns**
- Minor coordination: `ChatHistoryAdapter` is also edited by #15318 (this same `getData() as HashMap` block) and #15292 — merge order may need reconciling, though #15318 is already in changes-requested.

**Verdict:** Two legitimate correctness/safety fixes (shared binding + unsafe cast) with no behavior downside; ready to merge.

---

Okuro3499 (approved):

PR #15282, opened 2026-08-04
all: smoother resources courses sorting view modelling (fixes #15544)

### LINKED ISSUE #15544: Move sorting logic from adapters to viewmodels
`ResourcesAdapter` and `CoursesAdapter` own their list-sorting flags and logic, and `lowercase()` is called repeatedly during title sort. Move the sorting flags and logic out of the adapters and into `ResourcesViewModel` and `CoursesViewModel`, use a Schwartzian transform to precompute lowercased title keys (avoiding redundant `.lowercase()` calls), and wire the fragment list-sort triggers through the ViewModels while preserving `scrollToTop()`.

dogi:
Move sorting logic from adapters to viewmodels

- Extracted sorting flags and logic from `ResourcesAdapter` and `CoursesAdapter` to their respective ViewModels.
- Used Schwartzian transform to precompute lowercased keys for title sort to prevent redundant `.lowercase()` calls.
- Connected fragment list sort triggers properly to ViewModels, preserving `scrollToTop()` functionality.

---
*PR created automatically by Jules for task [2722493514456134837](https://jules.google.com/task/2722493514456134837) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6416
-        versionName = "0.64.16"
+        versionCode = 6417
+        versionName = "0.64.17"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/ui/resources/ResourcesAdapter.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/resources/ResourcesAdapter.kt
@@ -39,8 +39,6 @@ class ResourcesAdapter(
     private var listener: OnLibraryItemSelectedListener? = null
     private var homeItemClickListener: OnHomeItemClickListener? = null
     private var ratingChangeListener: OnRatingChangeListener? = null
-    private var isAscending = true
-    private var isTitleAscending = true
 
     init {
         if (context is OnHomeItemClickListener) {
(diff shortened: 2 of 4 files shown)


dogi:
**Review — rating: 82/100**

**Strengths**
- Correct sort-to-ViewModel extraction that **avoids the regression #15318 has.** The default `sortMode = NONE` means `applyCurrentSort` returns the list unchanged, so `processCourses`/`getLibraryListModels` preserve their existing default ordering (e.g. `compareBy(isMyCourse, courseTitle)`) on initial load — sorting only kicks in when the user taps a sort button. #15318 defaulted to TITLE and re-sorted the initial list, which is exactly the bug flagged there. This PR is the right way to do it.
- Toggle semantics match the old adapters exactly: date starts `isAscending = true` → first tap descending; title starts `isTitleAscending = false` → first tap ascending. The date/title comparators (createdDate; `title.lowercase(Locale.ROOT)`) are carried over faithfully, and both adapters shed their `isAscending`/`isTitleAscending` state cleanly.
- `checkList` correctly switches to `adapterLibrary.currentList.size`, and the resources fragment does the sort on `dispatcherProvider.io` before `setLibraryList(...) { scrollToPosition(0) }`. CI green on all three checks including `test (default)`.

**Concerns**
- **Supersedes / conflicts with #15318 and #15326.** #15318 moves the same CoursesAdapter sort (with the initial-order regression); #15326 adds a `sortCourses(...)` call inside `processCourses` (also re-sorting initial load). This PR is the cleaner of the three for the sort concern — recommend keeping **this** and dropping the overlapping sort changes from #15318/#15326. All three edit `CoursesViewModel`/`CoursesFragment` and will merge-conflict.
- **Behavior change worth confirming (minor, arguably an improvement):** because `getLibraryListModels`/`processCourses` now run `applyCurrentSort`, a data refresh after the user has sorted will *re-apply* that sort rather than snapping back to repo order (the old adapters only sorted on toggle). That's usually the nicer behavior, just flag it's intentional.
- `scrollToTop()` in `CoursesFragment` is now called synchronously right after `toggleSortOrder()` rather than as the `submitList` completion callback; harmless, but the scroll may fire a frame before the re-sorted list renders.

**Verdict:** Clean, correct sort extraction that fixes the initial-ordering trap the sibling PRs fell into; ready to merge, and it should be the keeper for this concern over #15318/#15326.

---

Okuro3499 (approved):

PR #15288, opened 2026-08-04
login: smoother creation username validating (fixes #15406)

### LINKED ISSUE #15406: Cancel stale username-validation work in BecomeMemberActivity

dogi:
Cancel stale username-validation work in BecomeMemberActivity

Replaces the per-keystroke `lifecycleScope.launch` in the username
`TextWatcher` with a debounced execution that tracks and cancels
the previous `Job`. A `delay(300)` is added to batch rapid text changes.
Also drops an unnecessary `withContext` switch to the main dispatcher
and ensures stale validations do not overwrite newer input by checking
if the text has changed since the coroutine was launched. Finally,
the running job is cleaned up in `onDestroy()`.

---
*PR created automatically by Jules for task [3668074042637295628](https://jules.google.com/task/3668074042637295628) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6321
-        versionName = "0.63.21"
+        versionCode = 6322
+        versionName = "0.63.22"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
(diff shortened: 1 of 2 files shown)


dogi:
**Review — rating: 84/100**

**Strengths**
- Four small correctness/UX wins in one focused change: (1) `usernameValidationJob?.cancel()` on each keystroke debounces rapid typing; (2) `delay(300)` avoids validating every character; (3) the staleness guard `if (etUsername.text.toString() != input) return@launch` prevents a slow validation from clobbering newer input; (4) cancelling the job in `onDestroy` prevents a leaked coroutine touching a dead view.
- Removing the inner `withContext(dispatcherProvider.main)` is safe — `lifecycleScope.launch` already runs on Main, and `validateUsername` is a suspend call that confines its own IO, so the UI updates were already main-dispatched. No threading regression.
- CI green on all three checks including `test (default)`.

**Concerns**
- Minor (pre-existing behavior, not introduced here): the auto-lowercase `setText(lowercase)` re-triggers `afterTextChanged`, spawning one more validate-debounce cycle (which no-ops since `input == lowercase`). Harmless with the new debounce, just noting the self-retrigger still exists.

**Verdict:** Clean debounce + cancel-stale + leak-fix for the username field with no threading regression; ready to merge.

---

Okuro3499 (approved):

PR #15297, opened 2026-08-04
teams: smoother events repository detail view modelling (fixes #15565)

### LINKED ISSUE #15565: Refactor: Remove UserSessionManager from EventsDetailViewModel
`EventsDetailViewModel` depends on `UserSessionManager` to resolve the active user, putting session logic in the presentation layer. Push active-user resolution down into `EventsRepository` (rename `toggleAttendance(meetupId, currentUserId)` to `toggleCurrentUserAttendance(meetupId)` and resolve the user inside the repository), removing `UserSessionManager` from the ViewModel to centralize domain logic.

dogi:
Refactors `EventsDetailViewModel` to remove `UserSessionManager` from the presentation layer. Active user resolution is now pushed down into the `EventsRepository`, matching the architectural campaign to centralize domain logic.

---
*PR created automatically by Jules for task [3334032882650341234](https://jules.google.com/task/3334032882650341234) started by @dogi*


dogi:
**Review — rating: 80/100**

**Strengths**
- Good decoupling: the ViewModel no longer needs `UserSessionManager` just to pass a user id into the repository — `toggleCurrentUserAttendance(meetupId)` resolves the current user internally, and `getCurrentUser()` is exposed for the `loadData` path. The VM constructor slims to a single dependency.
- Toggle logic preserved exactly: `isJoined = !meetup.userId.isNullOrEmpty()` and the `if (isJoined || !currentUserId.isNullOrEmpty())` gate are unchanged, so the "null user can leave but not join" behavior still holds — and the new `toggleCurrentUserAttendance_missingActiveUser` test locks that in (leave clears to "", join stays ""). The blank-`meetupId` early return is retained.
- Tests updated and expanded; CI green on all three checks including `test (default)`.

**Concerns**
- **Source swap to confirm.** The VM previously read the user via `userSessionManager.getUserModel()`; `getCurrentUser()` now returns `userRepository.getUserModel()`. Elsewhere in this codebase `userRepository.getUserModel()` is treated as the canonical active-user accessor, so these are almost certainly the same user — but since it's a change of source for the id that drives attendance, worth a quick confirm that `UserSessionManager.getUserModel()` and `UserRepository.getUserModel()` return the same active user (id) in the events flow.
- Minor: `toggleAttendance` now also refreshes `_user.value = getCurrentUser()` after the toggle (the old version didn't). Harmless/arguably better, just an added state refresh worth noting.

**Verdict:** Clean move of user resolution into the repository with the toggle semantics preserved and a good new null-user test; ready to merge once the `getUserModel()` source equivalence is confirmed.

---

Okuro3499 (dismissed):
```
1. EventsRepositoryImpl.toggleCurrentUserAttendance (EventsRepositoryImpl.kt)
• Problem: In EventsRepositoryImpl.kt, if (isJoined || !currentUserId.isNullOrEmpty()) evaluates to true when isJoined is true even if currentUserId is null or blank. This
executes meetup.userId = if (isJoined) "" else currentUserId, removing the attendee without an active session.

2. EventsRepositoryImplTest.toggleCurrentUserAttendance_missingActiveUser (EventsRepositoryImplTest.kt)
• Problem: The test asserts assertEquals("", meetup.userId) when userRepository.getUserModel() returns null, enforcing that a missing session wipes attendance.
```

dogi:
**Review — rating: 76/100**

The core of this is genuinely good. Folding the caller-supplied `currentUserId` into `toggleCurrentUserAttendance` removes a trust boundary — the caller could previously pass any user's id and toggle attendance on their behalf — and resolving it inside the repository is the right call. The new test is real coverage too, not decoration: it pins the missing-user path in both directions.

I traced the behavior change rather than assuming it was a pure rename, and there is one. The old code leaves the event whenever `isJoined` is true **regardless** of `currentUserId`; the new early return means a signed-out caller can no longer clear attendance. `toggleCurrentUserAttendance_missingActiveUser` asserts exactly that, so it's deliberate and it's the better semantics. Worth calling out in the PR description, since "remove `UserSessionManager`" doesn't hint that toggle behavior changed.

**The problem is `getCurrentUser()` on `EventsRepository`.** It's a pure passthrough to `userRepository.getUserModel()`, and it makes the events repository the ViewModel's source for the current user, which isn't its domain. This also runs directly against where the codebase is heading — **#15300 in this same queue is "Remove cross-repo passthroughs on `CoursesRepository`"**, so this adds one of the exact things that PR is deleting. Once the method is on the interface, other call sites will reach for it and it gets harder to remove.

**Fix:** drop `getCurrentUser()` from `EventsRepository`/`Impl` and inject `UserRepository` into `EventsDetailViewModel` for the `_user` flow. `UserRepository.getUserModel()` is already public and at least eight UI classes (`SurveyFragment`, `BellDashboardFragment`, `ResourcesFragment`, `CourseStepFragment`, …) inject it directly, so this is the established pattern rather than a new one. `EventsRepositoryImpl` keeps using `userRepository` internally for the toggle — that part stays exactly as written.

**Nit:** `toggleAttendance` in the ViewModel re-fetches the user (`_user.value = eventsRepository.getCurrentUser()`) after every toggle, but toggling attendance can't change the user model — that's an extra read per tap.

**Verdict:** Needs another round — swap the passthrough for a direct `UserRepository` injection in the ViewModel and this is a clean merge. Marking `change`.

---

dogi:

Every finding from last round is addressed, and addressed the right way rather than papered over.

`getCurrentUser()` is off the `EventsRepository` interface entirely — the events repository no longer doubles as the ViewModel's source for the current user, so this stops pulling against #15300's direction of deleting cross-repo passthroughs. `EventsDetailViewModel` now injects `UserRepository` and calls `getUserModel()` directly, matching what `SurveyFragment`, `BellDashboardFragment`, `ResourcesFragment` and others already do. `EventsRepositoryImpl` still resolves the user internally for the toggle, which was always the good part of this PR — the caller can no longer pass an arbitrary `userId` and toggle attendance on someone else's behalf.

The redundant per-tap re-read is gone too: `toggleAttendance` no longer re-fetches the user after toggling, which it never needed to since toggling attendance can't change the user model.

The deliberate behavior change is intact and still pinned by `toggleCurrentUserAttendance_missingActiveUser` — a signed-out caller can no longer clear attendance, where the old code would leave the event regardless of `currentUserId`. That's the better semantics and it's the tested one. CI green on build (default + lite) and test (default).

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #15301, opened 2026-08-04
resources: less teams repository injection is more (fixes #15454)

### LINKED ISSUE #15454: Remove unused repository injections
The `TeamsRepository` injection in `AddResourceActivity` is unused — the field is declared but never referenced. Remove the `@Inject` field and its import to clean up dead dependency-injection wiring.

dogi:
Remove unused repository injections in AddResourceActivity and SyncActivity

- Removed `teamsRepository` from `AddResourceActivity` completely.
- Removed `communityRepository` from `SyncActivity` as requested.
- Injected `communityRepository` specifically in `LoginActivity` instead of keeping it in the base class (`SyncActivity`).
- Handled `communityRepository` access in `ServerDialogExtensions.kt` via `EntryPointAccessor`.

---
*PR created automatically by Jules for task [3550507327898773688](https://jules.google.com/task/3550507327898773688) started by @dogi*


dogi:
**Review — rating: 66/100**

Two changes bundled under one title, and only one of them is actually "removing an unused injection."

**Strengths**
- `AddResourceActivity`: removing the `teamsRepository` `@Inject` is a legitimate dead-injection cleanup — CI green confirms it had no references. Good.
- Behavior is preserved (the community spinner still loads `getAllSorted().filter { name not empty }`), and all builds/tests pass.

**Concerns**
- **`communityRepository` wasn't unused — this is a relocation, and it moves the wrong way.** It was consumed by `ServerDialogExtensions.setupManualConfigEnabled` (via the `SyncActivity` receiver's field) and, per this diff, directly by `LoginActivity` (which now re-declares its own `@Inject`). Removing it from the base `SyncActivity` forces the extension to do `EntryPointAccessors.fromApplication(applicationContext, CoreDependenciesEntryPoint::class.java).communityRepository()` inline. That reintroduces a manual EntryPoint lookup for a dependency that was cleanly injected — the opposite direction of the rest of this refactor wave (e.g. #15313/#15323, which are actively *removing* `EntryPointAccessors` in favor of constructor/field injection). Net result is more boilerplate and a less idiomatic access path, not a cleanup.
- If the goal is to avoid a base-class field most `SyncActivity` subclasses don't need, the cleaner options are: keep it injected on the base (it *is* used through the extension), or pass `communityRepository` as a parameter to `setupManualConfigEnabled(...)` — either avoids the EntryPoint round-trip.
- Adding `communityRepository()` to `CoreDependenciesEntryPoint` grows the entry-point surface solely to support the EntryPoint call this PR introduces.

**Verdict:** Split this — land the `AddResourceActivity.teamsRepository` removal (genuinely unused), but rework the `communityRepository` change to keep it injected or pass it into the extension rather than reintroducing `EntryPointAccessors`.

---

dogi:
**Re-review — Rating: 90/100** (was 66)

The PR was narrowed exactly as I suggested. The current diff is a single, clean change: removing the unused `teamsRepository` `@Inject` (and its import) from `AddResourceActivity`. The problematic `communityRepository` relocation that reintroduced `EntryPointAccessors.fromApplication(...)` — against the direction of the rest of the DI-cleanup wave — is gone.

CI green on both flavors + tests confirms `teamsRepository` had no references, so this is a genuine dead-injection removal with no behavior change.

**Verdict:** Correct, focused cleanup — mergeable.

---

Okuro3499 (approved):

PR #15303, opened 2026-08-04
sync: smoother sync repository server pulling (fixes #15466)

### LINKED ISSUE #15466: Refactor SyncManager server pulls to SyncRepository
Tracking issue for PR #15303.

dogi:
Route `SyncManager`'s server pulls through `SyncRepository` by expanding the existing `SyncRepository` to be an interface, implementing logic in `SyncRepositoryImpl`, and injecting it into `SyncManager` so `SyncManager` only orchestrates state.

---
*PR created automatically by Jules for task [8270018884179952062](https://jules.google.com/task/8270018884179952062) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6419
-        versionName = "0.64.19"
+        versionCode = 6420
+        versionName = "0.64.20"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/di/RepositoryModule.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/di/RepositoryModule.kt
@@ -41,6 +41,8 @@ import org.ole.planet.myplanet.repository.SubmissionsRepository
 import org.ole.planet.myplanet.repository.SubmissionsRepositoryImpl
 import org.ole.planet.myplanet.repository.SurveysRepository
 import org.ole.planet.myplanet.repository.SurveysRepositoryImpl
+import org.ole.planet.myplanet.repository.SyncRepository
+import org.ole.planet.myplanet.repository.SyncRepositoryImpl
 import org.ole.planet.myplanet.repository.TagsRepository
 import org.ole.planet.myplanet.repository.TagsRepositoryImpl
 import org.ole.planet.myplanet.repository.TeamsRepository
@@ -157,4 +159,8 @@ abstract class RepositoryModule {
     @Binds
     @Singleton
     abstract fun bindUploadRepository(impl: UploadRepositoryImpl): UploadRepository
+
+    @Binds
+    @Singleton
+    abstract fun bindSyncRepository(impl: SyncRepositoryImpl): SyncRepository
 }
--- a/app/src/main/java/org/ole/planet/myplanet/repository/SyncRepository.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/SyncRepository.kt
@@ -1,75 +1,13 @@
 package org.ole.planet.myplanet.repository
 
-import android.content.Context
-import androidx.work.ExistingWorkPolicy
-import androidx.work.OneTimeWorkRequest
-import androidx.work.WorkInfo
-import androidx.work.WorkManager
-import androidx.work.workDataOf
-import dagger.hilt.android.qualifiers.ApplicationContext
-import javax.inject.Inject
-import javax.inject.Singleton
 import kotlinx.coroutines.flow.Flow
-import kotlinx.coroutines.flow.map
-import org.ole.planet.myplanet.services.UserDataWorker
-import org.ole.planet.myplanet.services.sync.TransactionSyncManager
+import org.ole.planet.myplanet.data.api.ApiInterface
 
-@Singleton
-class SyncRepository @Inject constructor(
-    @ApplicationContext private val context: Context,
-    private val transactionSyncManager: dagger.Lazy<TransactionSyncManager>
-) {
-    suspend fun syncDashboardKeyId(role: String?): SyncUiState {
-        return try {
-            transactionSyncManager.get().syncDashboardKeyId(role)
-            SyncUiState.Success(null)
-        } catch (e: Exception) {
-            SyncUiState.Error(e.message)
-        }
-    }
-
-    fun uploadLoginData(): Flow<SyncUiState> {
-        val workRequest = OneTimeWorkRequest.Builder(UserDataWorker::class.java)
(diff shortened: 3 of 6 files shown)


dogi:
**Review — rating: 72/100**

The extraction itself is well done — I diffed the moved `processShelfParallel`/`processShelfDataOptimizedSync` against the deleted SyncManager versions and they're identical, with the needed collaborators (`resourcesRepository`, `coursesRepository`, `eventsRepository`, `teamsSyncRepository`, `dispatcherProvider`, `context`) now injected into `SyncRepositoryImpl`. SyncManager delegates via `syncRepository.processShelfParallel(...)` on the same `async(io)` path, DI is wired (`@Binds @Singleton`, provider param added), the test constructor is updated, and CI is green. In isolation this is a valid, behavior-preserving decomposition.

**Concerns**

- **Structural conflict with #15325 (blocking).** #15325 keeps `SyncRepository` as a concrete `@Singleton class` and *adds* `syncDashboardKeyId(role)` to it. This PR converts `SyncRepository` into an `interface` (impl in `SyncRepositoryImpl`) and does **not** carry `syncDashboardKeyId`. The two definitions are mutually exclusive — whichever merges second won't compile. These need to be reconciled into one interface that has both the upload methods, the shelf-processing methods, and `syncDashboardKeyId`.
- **Conflict with #15323.** #15323 *removes* `ServiceModule.provideSyncManager` entirely (relying on `SyncManager`'s class-level `@Singleton @Inject constructor`), while this PR *edits* that same provider to thread in `syncRepository`. If #15323 lands, this hunk is obsolete; if this lands first, #15323 must drop the param instead. Coordinate the ordering.
- **Interface over-exposure.** `processShelfDataOptimizedSync` is only ever called internally by `processShelfParallel`; it doesn't need to be on the `SyncRepository` interface. Keep it `private` in the impl and expose only `processShelfParallel`.
- **Dead state left behind.** SyncManager still declares `private val stringArray = arrayOfNulls<String>(4)`, which is now unused there (its only user moved to the impl). Remove it. (Side note: `stringArray` is written from parallel shelf coroutines but never read, so the pre-existing data race is harmless — worth deleting rather than carrying forward.)
- Missing trailing newline on the edited `SyncManager.kt`.

**Verdict:** Faithful extraction, but it can't coexist with #15325's `SyncRepository` shape or #15323's provider removal as currently written. Reconcile the three (single `SyncRepository` interface carrying all methods; agree on the provider) and trim the interface/dead field, then it's mergeable.

---

dogi:
**Re-review — Rating: 74/100** (was 72)

Good progress on the structure, but a couple of dead-surface items remain. Verified against the current diff:

**Fixed since last review**
- ✅ **Interface over-exposure** — `processShelfDataOptimizedSync` is now `private` in `SyncRepositoryImpl`; the interface exposes only `processShelfParallel`. 
- ✅ **SyncManager dead field** — `stringArray` is removed from `SyncManager`.
- The extraction remains faithful (moved shelf-processing logic is identical to the deleted SyncManager versions), DI is wired, test constructor updated, CI green.

**Still needs a round**
1. **`syncDashboardKeyId` is dead interface surface.** It's now declared on `SyncRepository` and implemented as `transactionSyncManager.get().syncDashboardKeyId(...)`, but a code search shows the only caller — `BaseDashboardFragment.syncKeyId()` — still calls `transactionSyncManager.syncDashboardKeyId(...)` directly. Nothing invokes `syncRepository.syncDashboardKeyId`. Either migrate `BaseDashboardFragment` to go through `syncRepository` (which is presumably the point), or drop the method until there's a consumer.
2. **Dead field relocated, not removed.** `SyncRepositoryImpl` now declares `private val stringArray = arrayOfNulls<String>(4)`, but the moved `processShelfDataOptimizedSync` no longer assigns to it (those writes were dropped in the move — good). So the field is now dead *there*. Delete it.
3. **Cross-PR coordination.** By adding `syncDashboardKeyId` and removing `ServiceModule.provideSyncManager`, this PR now overlaps #15325 (which added `syncDashboardKeyId` to a concrete `SyncRepository`) and #15323 (which removed the same provider). This PR looks like the consolidation point — it **supersedes** #15325's shape and aligns with #15323's provider removal. Please confirm those two are closed/rebased onto this so whichever merges second doesn't hit a duplicate-definition/duplicate-deletion conflict.
4. qlty flags `processShelfDataOptimizedSync` complexity (28) — pre-existing, moved as-is, so not introduced here; worth a future split but not blocking.

**Verdict:** Clean extraction with the interface trimmed, but trim the two dead items (`syncDashboardKeyId` caller or the method itself; the relocated `stringArray`) and confirm the #15323/#15325 reconciliation, then it's mergeable.

---

dogi:
**Review — 80/100**

Clean, behavior-preserving extraction of the shelf-pull logic out of `SyncManager` and into `SyncRepositoryImpl`. Good direction — it moves `SyncManager` toward pure orchestration, which matches the layered architecture the project already follows.

**Verified**
- `processShelfParallel` / `processShelfDataOptimizedSync` moved essentially character-for-character; the `when (shelfData.type)` dispatch to the four batch-insert repositories, the `AdaptiveBatchProcessor` sizing, and the retry/logging paths are unchanged. `SyncManager` now calls `syncRepository.processShelfParallel(...)` at the one call site. ✓
- Nice side-effect: the old code wrote `stringArray[0..2]` — a shared `arrayOfNulls<String>(4)` field on `SyncManager` — from inside parallel `async(io)` jobs. That was a latent data race, and the move drops the field entirely (write-only, no readers; the build confirms no dangling refs). Good to see it gone. ✓
- `syncDashboardKeyId` is routed through `dagger.Lazy<TransactionSyncManager>`, which is the right tool to avoid a construction cycle; `BaseDashboardFragment.syncKeyId()` now goes through `SyncRepository` and reaches the same `TransactionSyncManager` target. Behavior preserved. ✓
- `SyncManagerTest` updated with the extra `mockk()` for the new constructor arg, and both build flavors + the test job are green — so the Hilt graph (removal of `provideSyncManager` in favor of the `@Inject` constructor + the new `@Binds SyncRepository`) is valid.

**Non-blocking nits**
- `processShelfParallel(shelfId, apiInterface)` takes `ApiInterface` as a parameter even though `SyncRepositoryImpl` could just inject it. Passing a singleton through the call signature is a small smell; injecting it would tidy the API.
- Several references use inline fully-qualified names (`org.ole.planet.myplanet.repository.SyncRepository`, `...services.sync.TransactionSyncManager`) instead of imports, in `RepositoryModule`, `BaseDashboardFragment`, `SyncManager`, and `SyncRepositoryImpl`. Worth converting to top-of-file imports to match the rest of the codebase.
- Interface cohesion: `SyncRepository` now bundles upload triggers (`uploadLoginData`/`uploadBulkData`), a server pull (`processShelfParallel`), and a dashboard-key sync. That's three concerns on one interface; fine for now, but keep an eye on it as more moves land.

**Verdict:** Behavior-preserving refactor that also removes a real race; compiles and tests pass. Mergeable — address the import/param nits at leisure.

---

Okuro3499 (approved):

PR #15305, opened 2026-08-04
all: smoother user model caching (fixes #15468)

dogi:
Optimizes application performance and reduces database load by preventing the entire user model from being reloaded from Room on every screen setup. Base fragments and ViewModels now load and cache the `UserEntity` once, and subclasses hook into the new `initView` to configure the UI using the cached model.

---
*PR created automatically by Jules for task [10505843203790855576](https://jules.google.com/task/10505843203790855576) started by @dogi*

PR #15306, opened 2026-08-04
all: smoother refresh job cancelling (fixes #15564)

### LINKED ISSUE #15564: Fix: Cancel stale coroutine jobs in long-lived fragments
Long-lived fragments (CoursesFragment, MyHealthFragment, ResourcesFragment, TeamsTasksFragment) launch refresh/coroutine jobs without cancelling prior stale ones, so older requests can overwrite newer results. Track the active job per refresh and cancel the stale job before launching a new one, preventing out-of-order results from clobbering the latest data.

dogi:
Cancel stale coroutine jobs in long-lived fragments to prevent older requests from overwriting newer results.

---
*PR created automatically by Jules for task [13281221085887992775](https://jules.google.com/task/13281221085887992775) started by @dogi*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved course rating refreshes by preventing overlapping updates.
  * Prevented duplicate patient-selection operations from running simultaneously.
  * Improved resource filtering and tag updates by cancelling outdated operations.
  * Improved task refresh reliability by managing refresh operations more consistently.
  * Reduced stale or conflicting updates when actions are performed quickly.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6435
-        versionName = "0.64.35"
+        versionCode = 6436
+        versionName = "0.64.36"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6436
-        versionName = "0.64.36"
+        versionCode = 6437
+        versionName = "0.64.37"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/ui/courses/CoursesFragment.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/courses/CoursesFragment.kt
@@ -63,6 +63,7 @@ class CoursesFragment : BaseRecyclerFragment<MyCourse?>(), OnCourseItemSelectedL
     var userModel: UserEntity? = null
     private lateinit var confirmation: AlertDialog
     private var selectionJob: Job? = null
+    private val refreshJobs = mutableMapOf<String, Job>()
     private var pendingScrollState: Parcelable? = null
     private val viewModel: CoursesViewModel by viewModels()
 
@@ -595,7 +596,8 @@ class CoursesFragment : BaseRecyclerFragment<MyCourse?>(), OnCourseItemSelectedL
 
     override fun onRatingChanged(type: String, id: String) {
         if (type == "course" && ::adapterCourses.isInitialized) {
-            viewLifecycleOwner.lifecycleScope.launch {
+            refreshJobs[id]?.cancel()
+            refreshJobs[id] = viewLifecycleOwner.lifecycleScope.launch {
                 viewModel.refreshCourseRatings(model?.id)
                 adapterCourses.refreshWithDiff(id)
             }
--- a/app/src/main/java/org/ole/planet/myplanet/ui/health/HealthViewModel.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/health/HealthViewModel.kt
@@ -49,6 +49,7 @@ class HealthViewModel @Inject constructor(
     val loggedInUser: StateFlow<UserEntity?> = _loggedInUser.asStateFlow()
 
     private var searchJob: Job? = null
+    private var selectPatientJob: Job? = null
 
     fun loadPatients(sortBy: String = "joinDate", descending: Boolean = true) {
         viewModelScope.launch {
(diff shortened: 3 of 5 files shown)


dogi:
**Review — rating: 84/100**

Correct fix for a real class of bugs. Long-lived fragments were launching a fresh `viewLifecycleOwner.lifecycleScope` coroutine on every search keystroke / filter change / rating refresh without cancelling the previous one, so rapid interaction piled up concurrent jobs racing to update the same adapter (last-writer-wins flicker, wasted queries). Capturing each into a `refreshJob`/`searchJob` field and calling `?.cancel()` before relaunch is the standard, correct remedy — applied consistently across CoursesFragment, MyHealthFragment, ResourcesFragment, TeamFragment, and TeamsTasksFragment (the last is just a rename of the already-cancelled `updateTasksJob`). Build (default + lite) and test (default) green.

**Cross-PR note:** this shares files with #15305 (CoursesFragment, ResourcesFragment, TeamFragment) — different methods/fields, but they'll touch adjacent lines, so land them in sequence.

**Concern:** none blocking.

**Verdict:** Correct, idiomatic cancellation of stale work — mergeable.

---

Okuro3499 (dismissed):
```
refreshHealthData() launches a coroutine stored in refreshJob and calls getHealthRecords(userId) inside it. getHealthRecords() immediately executes refreshJob?. cancel(), which cancels the caller coroutine running refreshHealthData(). When getHealthRecords() hits its first suspend call (userRepository.getUserByAnyId()), it throws a CancellationException and aborts.
```

dogi:
**Review — rating: 72/100**

The pattern is right and most of the diff earns its place. I checked whether each wrapped call actually suspends, since "cancel the stale job" only means anything if there's suspending work to cancel:

- **`MyHealthFragment.getHealthRecords`** — correct and the clearest win. Selecting a new member while the previous fetch is in flight is exactly the stale-job case.
- **`ResourcesFragment`** (six sites) — correct. `applyFiltersAndUpdateUI` is `suspend`, every site recomputes the whole filtered list, and a later filter genuinely supersedes an earlier one. Keeping `searchJob` separate from the existing `refreshJob` is the right call.
- **`TeamsTasksFragment`** — pure rename, no behavior change.

**The `TeamFragment` hunk is wrong and should be dropped.** `TeamViewModel.searchTeams` is not a suspend function, and its body is fully synchronous:

```kotlin
fun searchTeams(query: String) {
    currentSearchQuery = query
    applyFilters(currentTeamsDetails, currentSearchQuery)   // in-memory List.filter + _teamData.value = ...
}
```

No I/O, no coroutines, nothing to cancel. So `searchJob?.cancel()` always cancels a coroutine that has already completed, and the `launch` wraps a synchronous list filter. It's not merely redundant — it's misleading, because the next reader will believe stale searches are being cancelled here when they aren't. It also gives up something small but real: `.onEach { }` processed emissions sequentially and in order, whereas launch-per-emission no longer guarantees that two rapid emissions apply in order, and the `_teamData` update now lands a dispatch later. `debounce(300)` makes that unlikely to bite, but the old code was both simpler and stronger. Revert to `.onEach { viewModel.searchTeams(text?.toString() ?: "") }`.

**`CoursesFragment` needs a second look.** Here the cancellation is real (`refreshCourseRatings` is `suspend`), but `onRatingChanged(type, id)` is keyed **per course id**, and the job is not. Rate course A, then course B before A's refresh finishes, and A's `adapterCourses.refreshWithDiff(A)` is cancelled and never re-run — B's job only diffs row B. The other sites coalesce operations that fully supersede each other; this one doesn't. Either key the job per id, or leave this site uncancelled.

**Verdict:** Needs another round — drop the `TeamFragment` hunk, resolve the per-id cancel in `CoursesFragment`, keep the rest. Marking `change`.

---

dogi:
**Re-review — Rating: 85/100** (was 72)

Both findings resolved, and the harder one resolved properly rather than by deletion.

The `TeamFragment` hunk is gone — the diff no longer touches that file. That was the right call: `TeamViewModel.searchTeams` is fully synchronous (`currentSearchQuery = query` then an in-memory `applyFilters`), so wrapping it in a `launch` and cancelling the previous job cancelled a coroutine that had always already completed, while implying a cancellation guarantee that didn't exist. Back to `.onEach { viewModel.searchTeams(...) }`, which is both simpler and ordering-safe.

The `CoursesFragment` fix is the one I care about more, and keying the jobs by id is the correct answer:

```kotlin
private val refreshJobs = mutableMapOf<String, Job>()
...
refreshJobs[id]?.cancel()
refreshJobs[id] = viewLifecycleOwner.lifecycleScope.launch { ... }
```

Rating course A and then course B no longer cancels A's `refreshWithDiff(A)`, so A's row still gets refreshed — which was the actual defect, since that site coalesces per-item work rather than work that supersedes itself. The remaining sites are unchanged and were already right: `MyHealthFragment` and all six `ResourcesFragment` call sites wrap genuinely `suspend` work where a later call fully replaces an earlier one. CI green on build (default + lite) and test (default).

**Nit:** entries are never removed from `refreshJobs`, so the map grows one entry per distinct course rated and keeps completed `Job` references alive. It's bounded by the fragment's lifetime and freed with `viewLifecycleOwner`, so this isn't a real leak — but `.also { it.invokeOnCompletion { refreshJobs.remove(id) } }` would keep it tidy.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #15307, opened 2026-08-04
all: smoother base recycler caching (fixes #15542)

### LINKED ISSUE #15542: Stop re-creating adapters in `BaseRecyclerFragment`
`BaseRecyclerFragment.getAdapter()` builds a fresh adapter on every call (e.g. from `onRatingChanged`, `postAddRefresh`), discarding the `RecycledViewPool`, diff state and selection state. Guard against re-initialising an already-set `recyclerView.adapter` (and null it in `onDestroyView`), and make `CoursesFragment`, `ResourcesFragment`, and `TeamFragment` initialise the adapter only once while clearing the adapter and listeners in their `onDestroyView`.

dogi:
`BaseRecyclerFragment.getAdapter()` builds a fresh adapter every time it is called (e.g. from `onRatingChanged`, `postAddRefresh`), throwing away the `RecycledViewPool`, diff state and selection state.

**Changes:**
- Modified `BaseRecyclerFragment.kt` to check if `recyclerView.adapter` is already set before overriding it. Added `onDestroyView` to nullify `recyclerView.adapter`.
- Modified `CoursesFragment.kt`, `ResourcesFragment.kt`, and `TeamFragment.kt` to only initialize the adapter if it is not already initialized, preserving the adapter state. Also, cleared the adapter and listeners in their respective `onDestroyView`.

---
*PR created automatically by Jules for task [15611162526161887482](https://jules.google.com/task/15611162526161887482) started by @dogi*


dogi:
**Review — rating: 80/100**

**Strengths**
- Real fix for a real waste: `getAdapter()`/`onRatingChanged()`/`postAddRefresh()` previously rebuilt the adapter and reassigned `recyclerView.adapter` on every refresh, throwing away scroll position and view state. Guarding creation behind `::adapter.isInitialized` and only reassigning when `recyclerView.adapter != adapter` preserves state and cuts allocations, while `setListener`/`setLibraryList` still run each time so data and callbacks stay current.
- Good leak hygiene: `onDestroyView` now nulls `recyclerView.adapter` in the base and clears `setListener(null)`/`setRatingChangeListener(null)` in `CoursesFragment`/`ResourcesFragment` — the standard way to break the RecyclerView↔adapter↔fragment reference chain on view teardown.
- `ResourcesFragment` reordering is equivalent: creating the adapter once, attaching listeners, then computing `filteredList` and calling `setLibraryList(filteredList)` yields the same end state as before (order of `setListener` vs `setLibraryList` is immaterial), and moving `getUserModel()` inside the creation branch avoids a redundant fetch per refresh. CI green including `test (default)`.

**Concerns**
- **Constructor params captured once.** Because the adapter is now created a single time, values baked into it at construction — `isGuestUser = user?.isGuest()` and `.apply { setType(type) }` in `TeamFragment`, `isGuest`/`isMyCourseLib` in `CoursesFragment` — won't update if `user`/`type` change later without a full fragment recreate. That's almost always fine within one view lifecycle, but if guest→member can flip while the fragment is alive, the reused adapter keeps the stale flag. Worth confirming.
- `CoursesFragment.onRatingChanged()` drops the `super.onRatingChanged()` call in the not-initialized branch (now a plain `return`). Behaviorally that's a no-op instead of the base's adapter reassign — reasonable when there's no adapter yet, just flagging the intentional change.
- File overlap with #15318 (also edits `CoursesFragment`/`ResourcesFragment` around delete/refresh). These will need reconciling on merge; no correctness conflict here, just sequencing.

**Verdict:** Correct, beneficial reuse-and-cleanup refactor that preserves list state and plugs adapter leaks; ready to merge, mind the construct-once param capture if user/type can change mid-lifecycle.

---

Okuro3499 (approved):

PR #15308, opened 2026-08-04
teams: smoother submission detail view modeling (fixes #15469)

### LINKED ISSUE #15469: Refactor: Use WhileSubscribed(5000) for StateFlow SharingStarted
Tracking issue for PR #15308.

dogi:
Replaces `SharingStarted.Lazily` with `SharingStarted.WhileSubscribed(5000)` in ViewModels.

- `SubmissionDetailViewModel.kt`
- `SubmissionViewModel.kt`
- `VoicesViewModel.kt`

This prevents the upstream Flow (like Room queries) from staying alive indefinitely when the screen is no longer active. Using `WhileSubscribed(5000)` allows the flows to be gracefully cancelled after a 5-second timeout and matches standard codebase practices for UI-facing Flows.

---
*PR created automatically by Jules for task [10746089572881188496](https://jules.google.com/task/10746089572881188496) started by @dogi*


dogi:
**Review — rating: 84/100**

**Strengths**
- Correct application of the recommended sharing policy. `Lazily` starts the upstream on first collection and never stops it (for the ViewModel's whole life), which keeps Room observers / flows hot even while the UI is backgrounded. `WhileSubscribed(5000)` stops the upstream 5s after the last collector leaves and restarts on the next — the 5s window absorbs config changes without a restart. For the Room-backed chains in `SubmissionViewModel` (`userIdFlow → allSubmissionsFlow → filteredSubmissionsRaw`) this is a real resource/leak win.
- No empty-state flashing: all of these are `stateIn`/`shareIn(replay = 1)`, so the last value is retained and replayed immediately on re-subscription even after the upstream stops.
- One-shot sources (`flow { emit(getSubmissionDetail(id)) }`, `flow { emit(getActiveUserIdSuspending()) }`) will re-run if all subscribers drop for >5s and then return — that's a re-fetch of a read-only value, which is harmless and arguably fresher. CI green on all three checks including `test (default)`.

**Concerns**
- None blocking. Just be aware of the one-shot re-fetch semantics above if any of those repository reads were assumed to run exactly once per ViewModel; here they're idempotent, so it's fine.

**Verdict:** Idiomatic, behavior-safe `Lazily → WhileSubscribed(5000)` migration with retained replay; ready to merge.

---

Okuro3499 (approved):

PR #15310, opened 2026-08-04
teams: smoother voices removing (fixes #15407)

### LINKED ISSUE #15407: Fix VoicesAdapter full list rebind on parent post removal

dogi:
Fixes a UI bug in `VoicesAdapter` where removing a parent post rebuilt the entire list, causing visual jitter.

**Root Cause:**
`VoicesAdapter.removePost(newsId: String)` explicitly checked `if (isParent)` and submitted an empty list (`submitList(emptyList())`) prior to dropping it.

**Solution:**
Removed the `submitList(emptyList())` block entirely. Instead, if `isParent == true`, `parentNews` is correctly nulled out, and the method falls through to the existing list modification logic. This submits the properly filtered list only once. DiffUtil natively handles the deletion animation properly.

**Testing:**
Ran `testLiteDebugUnitTest --tests 'org.ole.planet.myplanet.ui.voices.*'` successfully. Tested with Git diff locally. Code review checked out cleanly.

---
*PR created automatically by Jules for task [2812647234433323472](https://jules.google.com/task/2812647234433323472) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6329
-        versionName = "0.63.29"
+        versionCode = 6330
+        versionName = "0.63.30"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/ui/voices/VoicesAdapter.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/voices/VoicesAdapter.kt
@@ -314,26 +314,26 @@ class VoicesAdapter(
         }
 
         if (isParent) {
-            submitList(emptyList())
-        } else {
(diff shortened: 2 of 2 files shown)


dogi:
**Review — rating: 85/100**

Real bug fix. When a **parent** post was removed, the old code did `submitList(emptyList())` — wiping the entire Voices list instead of removing the one post. The fix sets `parentNews = null` and lets the same removal path run that the non-parent branch used (remove the item from `originalList`/`currentList`, then submit the properly filtered list). So a parent removal now drops just that entry (and its leading-parent dedup still applies), instead of clearing the screen. Build (default + lite) and test (default) green.

**Cross-PR note:** touches `VoicesAdapter` like #15383 (payload rebinds) but a different method (the removal path), so they're complementary — just sequence the merges since they're the same file.

**Concern:** none blocking. Would be nice to add a small test around `removeNewsItem` for the parent case to lock this in, given it was a user-visible "whole list disappears" bug.

**Verdict:** Correct fix for a real regression — mergeable.

---

Okuro3499 (approved):

PR #15311, opened 2026-08-04
teams: smoother repository flowing (fixes #15470)

### LINKED ISSUE #15470: refactor: add flowOn to mapping flows in TeamsRepositoryImpl
Tracking issue for PR #15311.

dogi:
Added `.flowOn(dispatcherProvider.default)` to flows in `TeamsRepositoryImpl` that execute non-trivial mapping operations to prevent blocking the main thread.

---
*PR created automatically by Jules for task [7458537279842683931](https://jules.google.com/task/7458537279842683931) started by @dogi*


dogi:
**Review — rating: 82/100**

**Strengths**
- Correct and behavior-neutral: `flowOn` only relocates *upstream* execution context; it doesn't change emitted values or ordering, and the collector still receives on its own dispatcher. So this is pure threading hygiene with no functional risk.
- The two flows at `TeamsRepositoryImpl.kt:201` and `:306` are the real wins — each does non-trivial `map { ... filter { isRootTeam() } ... }` + `mapToTeamDetails(...)` work that previously ran on the collector's context (often Main); `.flowOn(dispatcherProvider.default)` moves that CPU work to the Default pool. Placement is right (`flowOn` after the whole `flow { }` covers all the transformation).
- Uses the injected `dispatcherProvider` rather than a hard-coded `Dispatchers.*`, matching the repo convention. CI green on all three checks including `test (default)`.

**Concerns**
- `getTasksFlow` (`:78`) adds `.flowOn(default)` to a *bare* Room DAO flow with no transformation. Room already runs the query on its own executor and emits off the main thread, so this one is close to a no-op — harmless and consistent, just not doing much. If any of the mapped flows call suspend DB functions inside `mapToTeamDetails`, note that Room suspend queries self-dispatch, so `default` vs `io` there is fine; only worth switching to `io` if there's actual blocking file/network work in the mapping.

**Verdict:** Low-risk, behavior-preserving threading improvement that correctly offloads the team-details mapping; ready to merge.

---

Okuro3499 (approved):

PR #15312, opened 2026-08-04
courses: smoother pager adapter diffing (fixes #15471)

### LINKED ISSUE #15471: Refactor: Replace notifyDataSetChanged with DiffUtil in PagerAdapters
Tracking issue for PR #15312.

dogi:
Replaces `notifyDataSetChanged()` with efficient diff-based updates in the two `FragmentStateAdapter` classes (`CoursesPagerAdapter` and `TeamPagerAdapter`). Also ensures `TeamPagerAdapter` provides stable IDs instead of relying on hash codes to correctly support `DiffUtil` updates without fragment state loss.

---
*PR created automatically by Jules for task [18039229700573132766](https://jules.google.com/task/18039229700573132766) started by @dogi*


dogi:
**Review — rating: 58/100**

The two adapters get inconsistent treatment, and the difference is exactly the thing that makes DiffUtil safe (or not) on a `FragmentStateAdapter`.

**Strengths**
- **`TeamPagerAdapter` is done right.** `FragmentStateAdapter` only supports granular updates when `getItemId`/`containsItem` return *stable, unique* ids across dataset changes. This PR replaces the old `id.hashCode().toLong()` (collision-prone) with an `itemIds` map that assigns a persistent incrementing `Long` per `page.id` and keeps it across `updatePages`, and both overrides are updated to use it. That's the correct pattern, and the `areItemsTheSame = { it.id }` / `areContentsTheSame = { == }` split is right. CI green including `test (default)`.

**Concerns**
- **`CoursesPagerAdapter` dispatches DiffUtil updates without stable ids.** It calls `diffResult.dispatchUpdatesTo(this)` but does **not** override `getItemId`/`containsItem`, so it falls back to `FragmentStateAdapter`'s position-based defaults (`getItemId = position`, `containsItem = itemId in 0 until itemCount`). Those ids are only stable for pure append-at-the-end changes. The moment `updateSteps` is called with a reordered, shortened, or different set (e.g. the adapter is reused for another course), the granular remove/insert events leave fragment state keyed to the wrong positions → wrong page shown or an `IllegalStateException`/"Fragment already added". `notifyDataSetChanged()` papered over this by forcing a full rebind; switching to DiffUtil removes that safety net without adding the stable-id contract. Give `CoursesPagerAdapter` the same `itemIds`-map treatment `TeamPagerAdapter` got (keyed on the step id, with the `null` overview page mapped to a fixed reserved id).
- Minor: the `TeamPagerAdapter` diff has scattered imports — `DiffUtils` inserted between `MainApplication` and the `callback.*` imports, and a lone `androidx.recyclerview.widget.RecyclerView` after a blank line. Tidy the import block.

**Verdict:** Half-correct — `TeamPagerAdapter` is a good, safe conversion, but `CoursesPagerAdapter` needs matching stable-id overrides before its DiffUtil dispatch is safe. Please align the two.

---

dogi:
**Re-review (correction) — Rating: 88/100** (was 58)

I need to correct my earlier review: I claimed `CoursesPagerAdapter` dispatched DiffUtil **without** stable-id overrides. That was wrong. Reading the full file at head, it does have them:

```kotlin
private val itemIds = mutableMapOf<String, Long>()
private var nextId = 1L
// submitList(): assigns a persistent Long per new stepId before diffing
override fun getItemId(position) = if (position == 0) COURSE_DETAIL_ID else itemIds[steps[position-1]] ?: NO_ID
override fun containsItem(itemId) = itemId == COURSE_DETAIL_ID || steps.any { itemIds[it] == itemId }
```

So the DiffUtil dispatch **is** safe here: position 0 maps to a fixed reserved `COURSE_DETAIL_ID`, each step gets a stable, persistent id via `itemIds`, and the diff runs over `listOf(null) + steps` with value equality that lines up with those ids. The earlier concern was my error — apologies for the noise.

With that corrected, both adapters are done right and consistently:
- **`CoursesPagerAdapter`** — stable ids + DiffUtil, as above.
- **`TeamPagerAdapter`** — replaces the collision-prone `id.hashCode().toLong()` with the same persistent `itemIds` map and updates both overrides; `areItemsTheSame = { it.id }` / `areContentsTheSame = { == }` is the right split.

Both correctly satisfy `FragmentStateAdapter`'s stable-unique-id contract, so replacing `notifyDataSetChanged()` with granular `dispatchUpdatesTo(this)` is safe. Build (both flavors) + test green.

**Nit (non-blocking):** `TeamPagerAdapter` import block is scattered — `DiffUtils` sits between `MainApplication` and the `callback.*` imports, with a lone `androidx.recyclerview.widget.RecyclerView` after a blank line. Tidy ordering.

**Verdict:** Correct, safe DiffUtil conversion on both adapters — mergeable.

---

Okuro3499 (approved):

PR #15315, opened 2026-08-04
all: smoother notification utils caching (fixes #15507)

### LINKED ISSUE #15507: Optimize NotificationUtils.getInstance allocations
`NotificationUtils.getInstance()` allocates a `NotificationManager` and calls `EntryPointAccessors.fromApplication` on every invocation. Cache the `NotificationManager` with a `@Volatile` backing field and Double-Checked Locking, use `context.applicationContext` to avoid Context leaks, and avoid redundant `EntryPointAccessors` calls.

dogi:
Fixes multiple allocations in `NotificationUtils.getInstance()`.

- Introduced a `@Volatile` backing field in `NotificationUtils` to cache the `NotificationManager`.
- Used Double-Checked Locking (DCL) to ensure thread-safe initialization.
- Replaced the direct use of the provided `context` with `context.applicationContext` to eliminate potential Context leaks.
- Avoided redundant calls to `EntryPointAccessors.fromApplication` per invocation.

---
*PR created automatically by Jules for task [741864532761424475](https://jules.google.com/task/741864532761424475) started by @dogi*


dogi:
**Review — rating: 84/100**

**Strengths**
- Correct double-checked-locking singleton: `@Volatile` field + `synchronized(this)` with the inner re-check is the textbook idiom, and it's implemented properly here.
- Safe to cache. The instance captures `context.applicationContext` (app-lifetime, not an Activity/short-lived context, so no leak) and a `timeProvider` resolved from `CoreDependenciesEntryPoint` (a Hilt `@Singleton`). Since every prior call already resolved to the same application context and the same singleton time provider, one memoized `NotificationManager` is semantically identical to the per-call instances it replaces — pure allocation/EntryPoint-lookup savings on the hot path.
- No initialization-order change: `EntryPointAccessors.fromApplication` is still only touched inside `getInstance`, same as before. CI green on all three checks including `test (default)`.

**Concerns**
- None blocking. Longer term, `NotificationManager` would be cleaner as a Hilt `@Singleton` injected where needed (removing the manual `EntryPointAccessors` dance entirely), but this lazy-cache is a reasonable pragmatic step given the `object` access pattern.

**Verdict:** Clean, correct memoization with the context-leak question checked (application context only); ready to merge.

---

Okuro3499 (dismissed):
```
activeNotifications and sessionShownNotifications are plain, unsynchronized mutableSetOf<String>(). Once getInstance() caches a single shared NotificationManager, these  
  sets are reachable concurrently from TaskNotificationWorker, NotificationActionReceiver, and DashboardActivity, each on IO-pool threads. Concurrent add/remove/contains calls on a 
  plain HashSet from multiple threads is undefined behavior. 
```

dogi:
**Re-review — Rating: 88/100** (was 84)

The DCL singleton is unchanged and correct (`@Volatile` + `synchronized` with inner re-check; caches `applicationContext` + the `@Singleton` `timeProvider`, so no leak and semantically identical to the per-call instances). The current diff also carries the right *consequence* of that change:

- Because `getInstance` now returns a **shared** `NotificationManager`, its `activeNotifications` and `sessionShownNotifications` sets are now touched from multiple threads (previously each caller had its own manager). Switching them from `mutableSetOf` (`HashSet`, not thread-safe) to `ConcurrentHashMap.newKeySet()` is exactly the correct accompanying fix — individual add/remove/contains are now atomic. Good catch that the singleton implies concurrent access on that state.

Build (both flavors) + test green.

**Nit (non-blocking):** `java.util.concurrent.ConcurrentHashMap.newKeySet<String>()` is written fully-qualified inline — an import would read cleaner.

**Verdict:** Correct memoization plus the thread-safety follow-through it requires — mergeable.

---

Okuro3499 (approved):

PR #15317, opened 2026-08-04
life: smoother health user repositories view modelling (fixes #15563)

### LINKED ISSUE #15563: Refactor MyHealthFragment data layer behind health boundary
`MyHealthFragment` fetches data directly and depends on `UserRepository` and `SharedPrefManager`, blurring the health boundary. Move the data-fetching flow behind `HealthRepository`: add `getPatientById`, `getPatientsSortedBy`, `searchPatients`, and `getPatientHealthRecords` to `HealthRepository`/`HealthRepositoryImpl`, expand `HealthViewModel` with StateFlows for patient loading/selection/search/health records, and refactor `MyHealthFragment` to observe the ViewModel via `collectWhenStarted`, removing the `UserRepository` and `SharedPrefManager` dependencies from the fragment.

dogi:
Refactored the `MyHealthFragment` data fetching flow.
- Added `getPatientById`, `getPatientsSortedBy`, `searchPatients`, and `getPatientHealthRecords` to `HealthRepository` & `HealthRepositoryImpl`.
- Expanded `HealthViewModel` with StateFlows for patient loading, selection, search and health records.
- Refactored `MyHealthFragment` to observe the `HealthViewModel` via `collectWhenStarted` and removed `UserRepository` and `SharedPrefManager` dependencies from the fragment.

---
*PR created automatically by Jules for task [16370150681129328921](https://jules.google.com/task/16370150681129328921) started by @dogi*


dogi:
**Review — rating: 80/100**

Checked the lookup that mattered most in a medical-records screen, and it holds.

**Strengths**
- **No id-matching regression.** The new `HealthRepository.getPatientById(id)` calls `userDao.getById(id)`, which is *exactly* what the old `userRepository.getUserByAnyId(normalizedId)` did (`getUserByAnyId` is literally `return userDao.getById(id)`). So moving patient resolution into the health boundary doesn't narrow the match.
- Genuine MVVM extraction: patient list/search/sort and the decrypt-and-assemble `HealthRecord` logic move into `HealthRepository`/`HealthViewModel`, and the fragment becomes a thin observer of `selectedPatient`/`patientList`/`healthRecord`/`isListLoading` StateFlows. `getPatientHealthRecords` faithfully mirrors the old `getHealthRecordsAndAssociatedUsers` (same `getByIdOrUserId` → decrypt → `getByProfileId` → `createdBy` userMap assembly). Health decryption logic living in the health repo is the right home.
- A real unit test (`HealthViewModelTest`) is added, and CI is green across all three checks including `test (default)`.

**Concerns**
- **Search loading UX changed slightly.** The old `setTextWatcher` only showed the progress spinner if the search took longer than 100ms (a `launch { delay(100); show }` that got cancelled for fast searches). The new `searchPatients` sets `isListLoading = true` immediately after the 300ms debounce for every search, so the spinner can flash on fast local queries. Minor, but a deliberate behavior change worth confirming you're OK with.
- Dead leftovers in the fragment: `searchJob` (and its `dispatcherProvider` usage for search) are no longer used now that the VM owns debouncing, and `onDestroyView` dropped the `searchJob?.cancel()`. Clean these up.
- Confirm `setupRealtimeSync()` is still invoked from `onViewCreated` — it's unchanged by this diff, but the health-table realtime refresh depends on it being wired.
- Cosmetic: fully-qualified `org.ole.planet.myplanet.model.HealthRecord` in the interface signature, and interleaved import ordering in `HealthViewModel`/`HealthRepositoryImpl`.

**Verdict:** Well-structured, behavior-preserving MVVM refactor with the critical patient-lookup verified equivalent and a test to back it; ready to merge after the dead-`searchJob` cleanup and a glance at the search-spinner UX.

---

Okuro3499 (changes requested):
```
1. File or function affected: HealthRepositoryImpl.kt (getPatientHealthRecords, getPatientsSortedBy, searchPatients) & HealthViewModel.kt (selectPatient)
      • Problem: AES decryption (AndroidDecrypter.decrypt), GSON parsing, and userDao.getAll() table scans run on the main thread via viewModelScope.launch default context
      (Dispatchers.Main.immediate), risking UI freezes and ANRs.
      • Fix: Wrap CPU-bound decryption and database operations in withContext(dispatcherProvider.io) inside HealthRepositoryImpl or dispatch them on dispatcherProvider.io.
2. File or function affected: MyHealthFragment.kt (selectPatient, onDestroyView)
      • Problem: Dialog created in selectPatient() is not dismissed when the fragment view is destroyed, causing a WindowLeaked Android exception if the fragment detaches while
      the dialog is visible.
      • Fix: Add dialog?.dismiss(); dialog = null inside onDestroyView(). Also dismiss any previously showing dialog instance before creating a new one in selectPatient().
3. File or function affected: MyHealthFragment.kt (onResume, refreshHealthData)
      • Problem: onResume() unconditionally triggers viewModel.selectPatient(normalizedId) on every resume event, overriding the user's current interaction. Additionally,
      refreshHealthData() forces viewModel.loadInitialPatient(), which resets any health worker's patient selection back to their own account during background sync updates.
      • Fix: Remove redundant selectPatient call from onResume() (the ViewModel StateFlow retains and restores view state automatically). In refreshHealthData(), refresh records
      for the currently active userId instead of forcing loadInitialPatient().
```

Okuro3499 (dismissed):
```
• In MyHealthFragment.kt, the click listener for binding.addNewRecord was removed during refactoring, causing the "Add New Record" action to become completely non-functional.
• In MyHealthFragment.kt, setupButtons() determines isHealthProvider using userModel, which is now set to selectedPatient. When a health provider selects a patient who is not a
health worker, btnnewPatient visibility is set to View.GONE, locking the health provider out of selecting any subsequent patients.
```

dogi:
**Review — rating: 70/100**

The direction is right and this is the layering the codebase wants: `MyHealthFragment` shrinks a lot, the data access lands behind `HealthViewModel` → `HealthRepository`, and moving the 300ms search debounce into the ViewModel is a genuine improvement over debouncing in a `doAfterTextChanged`. The `sortList` spinner handler dropping to a single `viewModel.loadPatients(sortBy, sort)` call is exactly the right shape.

**The main problem is that it copies instead of moving.** I traced every method it reimplements:

- `UserRepository.searchUsers` — its only call site was `MyHealthFragment:249`, which this PR deletes. It is now **dead code**, and `HealthRepository.searchPatients` reimplements it.
- `UserRepository.getHealthRecordsAndAssociatedUsers` — its only call site was `MyHealthFragment:302`, also deleted here. Also now **dead code**, and `getPatientHealthRecords` reimplements ~35 lines of it (decrypt → Gson parse → `userMap` assembly).
- `UserRepository.getUsersSortedBy` — still used by `DashboardViewModel:174`, so it can't be deleted, which means the joinDate/name sort now genuinely exists **twice**, in two repositories.

So the net effect is two orphaned methods left in `UserRepository` plus one permanently duplicated sort. Either delete the two dead ones as part of this PR, or have `HealthRepositoryImpl` delegate to `UserRepository` rather than re-querying `userDao` itself.

Other things worth fixing before this lands:

- **`android.text.TextUtils.isEmpty(json)` inside `HealthRepositoryImpl`** drags an Android framework type into the data layer, so anything testing this repository needs Robolectric. `json.isNullOrEmpty()` is plain Kotlin and equivalent here.
- **Dead code left behind in the fragment.** `searchJob` is still declared but never assigned again (and its `onDestroyView` cancel was removed), and the `UserSessionManager` / `SharedPrefManager` imports plus the `dispatcherProvider` injection appear to have lost their last uses along with the old `loadingJob`.
- **Search spinner now flickers.** The old code deliberately delayed the progress indicator by 100ms so fast searches never flashed it; `searchPatients` sets `_isListLoading = true` unconditionally, so every keystroke flashes the spinner.
- **`userModel`/`userId` are read inside the `healthRecord` collector but written by the `selectedPatient` collector** — two independent flows with an implicit ordering requirement. It works today because the repository call between the two `_` assignments gives the first collector time to run, but that's timing, not a guarantee. Collapsing these into one UI-state object would make it structural.
- **Test conventions.** `HealthViewModelTest` hand-rolls `Dispatchers.setMain(StandardTestDispatcher())` instead of the project's `MainDispatcherRule` / `TestDispatcherProvider` (see `docs/TESTING.md`), and covers only `loadPatients` — nothing for `selectPatient`, the search debounce, or `getPatientHealthRecords`, which is the genuinely intricate new code.

**Nits:** `org.ole.planet.myplanet.model.HealthRecord` is spelled out fully in both the interface and impl instead of imported; `.map { it }` in the `userMap` chain is a no-op; and there are several stray double/triple blank lines plus a lost blank line between `refreshHealthData()` and `onViewCreated`.

**Verdict:** Needs another round — mainly resolve the copy-vs-move duplication and drop the now-dead `UserRepository` methods. Marking `change`.

---

dogi:
**Re-review — Rating: 84/100** (was 70)

Every substantive finding is addressed, and the two that mattered most were fixed at the root rather than patched.

**The copy-vs-move problem is gone.** `UserRepository.searchUsers` and `getHealthRecordsAndAssociatedUsers` are now deleted from both the interface and `UserRepositoryImpl`, so this genuinely *moves* the health queries behind the health boundary instead of leaving two orphaned methods behind. That was the blocker and it's cleanly resolved. `getUsersSortedBy` correctly stays — `DashboardViewModel:174` still needs it — so the joinDate/name sort does still exist in two places, but as I said last round that one can't be deleted without a separate refactor.

**The two-flow ordering coupling is fixed structurally.** Introducing `PatientDetailState(user, healthRecord)` and collapsing `_selectedPatient` + `_healthRecord` into a single `_patientDetailState` means the collector now receives the user and the record together, instead of one collector writing `userModel`/`userId` that another collector reads and hoping the repository call in between provides enough slack. That's the right shape.

The rest, confirmed one by one:
- `android.text.TextUtils.isEmpty(json)` → `json.isNullOrEmpty()`, so the repository no longer pulls an Android framework type into the data layer.
- `HealthRecord` is imported and used bare in both the interface and impl.
- The `.map { it }` no-op is gone from the `userMap` chain.
- The 100ms anti-flicker delay is restored, now correctly in the ViewModel — `loadingJob` sets `_isListLoading` only if the search outlives 100ms, then gets cancelled. Fast searches no longer flash the spinner.
- Dead code cleaned up: the `dispatcherProvider`, `userSessionManager`, `userRepository` and `sharedPrefManager` injections, the `searchJob` field, and their imports are all removed.
- `HealthViewModelTest` now uses the project's `MainDispatcherRule` per `docs/TESTING.md` instead of hand-rolling `Dispatchers.setMain`, and covers `selectPatient`, `loadPatients`, and the debounced `searchPatients`.

CI green on build (default + lite) and test (default).

**Remaining nits, none blocking:** `userModelList` is still a `lateinit var` but nothing assigns it any more now that the `patientList` collector drives `adapter.submitList` — it'll throw if anything ever reads it. The `rv` parameters on `sortList` and `setTextWatcher` are now unused for the same reason. `android.text.TextUtils` appears to have lost its last use in `HealthViewModel`. And the test file imports `HealthRecord` but then constructs it fully-qualified, and uses `@get:org.junit.Rule` / `@org.junit.Before` inline rather than importing them.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #15318, opened 2026-08-04
courses: smoother sort toggle view modelling (fixes #15537)

### LINKED ISSUE #15537: Refactor: Remove adapter-side mutable working lists
Several adapters copied `currentList` into mutable working lists (SurveysAdapter sortStrategy, CoursesAdapter sort fns, VoicesAdapter originalList, LifeAdapter workingList, ChatHistoryAdapter currentFlatList) for local sorting/tracking, doubling memory churn on every update and managing app state inside the adapter. Remove these adapter-side mutable caches: SurveysFragment relies on SurveysViewModel for sorting; CoursesAdapter sorting moves to CoursesViewModel/CoursesFragment; VoicesAdapter consumes directly from VoicesViewModel; LifeAdapter uses a temporary list only during drag-and-drop; ChatHistoryAdapter scopes currentFlatList to a local variable.

dogi:
Resolves the issue where several adapters copied `currentList` into mutable lists for local sorting or tracking, which doubled memory churn on every update and improperly managed application state inside the adapter.

Key changes:
- `SurveysAdapter`: Removed internal `sortStrategy` and `sortByDate`/`toggleTitleSortOrder` tracking; the `SurveysFragment` now natively relies on `SurveysViewModel` for sorting.
- `CoursesAdapter`: Removed sorting functions and properties. State logic has been moved to `CoursesViewModel` and `CoursesFragment` to update the list reactively.
- `VoicesAdapter`: Replaced `originalList` working cache with direct downstream consumption from `VoicesViewModel` which inherently tracks and emits list state updates.
- `LifeAdapter`: Replaced a permanent `workingList` cache with a temporary list created ad-hoc exclusively during drag-and-drop (`onItemMove`), avoiding permanent memory duplication.
- `ChatHistoryAdapter`: Refactored a stateful `currentFlatList` tracking loop to be a strictly scoped local variable instead.

---
*PR created automatically by Jules for task [7740502947552100873](https://jules.google.com/task/7740502947552100873) started by @dogi*

## Summary by CodeRabbit

* **New Features**
  * Added title and date sorting for courses, with the selected sort preserved when lists refresh or are filtered.
  * Improved course list refresh behavior after deletion, archiving, and removal.
* **Bug Fixes**
  * Improved chat group expansion and collapsing behavior.
  * Improved drag-and-drop reordering reliability in Life lists.
  * Improved parent and child post removal handling in Voices.
* **Refactor**
  * Simplified list sorting and state management across courses, surveys, life lists, and voices.


dogi:
**Review — rating: 48/100**

The goal — get mutable working lists out of adapters and let the ViewModel/upstream own list state — is the right instinct, and `ChatHistoryAdapter` (reading expansion state from `adapter.currentList` at click time) is a clean example of it. But this bundles five adapters' worth of behavior changes, and several are unverified regressions. CI green only proves compilation + existing unit tests; none of these UI flows are covered.

**Concerns**

1. **`CoursesViewModel.processCourses` now re-sorts and overrides the default ordering.** `mappedCourses = sortCourses(sortedCourseList.map { it.toCourse() })` runs on top of the existing `sortedWith(compareBy({ it.isMyCourse }, { it.courseTitle }))`. With the new defaults (`activeSort = TITLE`, `isTitleAscending = true`), the initial list is re-sorted by title only, which **destroys the `isMyCourse`-first grouping** that was the established default display order. The old adapter applied its sort only when the user tapped a sort button; it never re-sorted on initial load. This changes what users see on first render.

2. **Course delete/archive lost their UI refresh.** `onDeleteConfirmed` and `onArchiveConfirmed` had `adapterCourses.removeCourses(courseIds)` removed, but only the *remove-from-my-courses* path gained a `viewModel.loadCourses(...)` reload (line ~89). Unless `deleteSelected(true)` itself re-emits the list, the deleted/archived courses stay on screen after confirmation. Please confirm `deleteSelected` triggers a reload; if not, these two paths need an explicit `loadCourses` like the third one got.

3. **`VoicesAdapter.removePost` gutted to a no-op — and this collides with #15310.** All callers (`VoicesFragment`, `ReplyActivity`, `TeamsVoicesFragment`) dropped their `removePost` call, so deletion now depends entirely on "upstream reloading data in ViewModel" (per the new comment). That's only correct if `deletePost` actually causes the news Flow to re-emit — otherwise deleted posts linger. Separately, **#15310 (already marked ready) fixes a real bug inside `removePost`**; this PR deletes the whole method. The two are mutually exclusive approaches to the same code and can't both merge — pick one and close/rebase the other.

4. **`LifeAdapter` drag mixes `notifyItemMoved` with `ListAdapter`.** During a drag you now mutate `dragList` and call `notifyItemMoved`, but the `AsyncListDiffer`'s backing list (what `getItem()` returns) isn't updated until `onItemMoveFinished` → `submitList`. Any rebind mid-drag reads stale positions, and the final `submitList` diff runs against a list the RecyclerView already visually reordered. The old `submitList(workingList.toList())` per move kept the differ authoritative. This is a known-fragile pattern with `ListAdapter`.

5. **`SurveysAdapter` sort methods removed with no visible replacement.** `sortByDate`/`toggleTitleSortOrder` are deleted but no ViewModel equivalent appears in the diff. Since CI compiles, the fragment no longer references them — so either the survey sort UI now does nothing, or the replacement lives outside this diff. Please confirm survey sorting still works.

**Verdict:** Good direction, but too many bundled, untested behavior changes and a direct conflict with #15310. Split into per-adapter PRs, verify the delete/reload and survey-sort flows, and reconcile with #15310 before this can move forward.

---

dogi:
**Re-review — Rating: 74/100** (was 48)

Big improvement — the two hard correctness bugs I flagged are fixed:

- ✅ **#1 `isMyCourse` grouping preserved.** `CoursesViewModel.activeSort` now defaults to `null`, and `sortCourses` returns the list unchanged for `null`. So initial load keeps the `sortedWith(compareBy({ it.isMyCourse }, { it.courseTitle }))` order, and re-sorting only happens after the user taps a sort button. Exactly right.
- ✅ **#2 Course delete/archive refresh restored.** All three paths — remove-from-mycourse, `onDeleteConfirmed`, `onArchiveConfirmed` — now call `viewModel.loadCourses(isMyCourseLib, model?.id)`, so deleted/archived courses leave the screen.
- ✅ **#3 (code half) Voices deletion now propagates.** `VoicesViewModel.deletePost` now filters `_baseNewsList` and reassigns the StateFlow, so removal flows through the observed list rather than relying on `removePost`'s (now-removed) manual mutation. That closes the "does it actually re-emit?" gap.
- ✅ **#5 Surveys sort — dead-code removal.** CI compiles with no `SurveysFragment` change, confirming those adapter methods were already unused (the fragment sorts via the ViewModel). Safe to drop; the description overstates it as a fragment migration, but the removal itself is fine.
- ~ **#4 LifeAdapter drag** — on reflection this `notifyItemMoved` during drag + `submitList` on `onItemMoveFinished` is the standard ItemTouchHelper pattern, not a defect. Worth a manual smoke-test for reorder flicker, but I'm no longer treating it as a problem.

**Why still another round — the cross-PR collision (concern #3, structural half):**
- **#15318 vs #15310 are mutually exclusive on `VoicesAdapter.removePost`.** #15310 (currently labeled **`ready to merge`**) *fixes* `removePost` in place; this PR *deletes* it. They cannot both merge. My recommendation: **#15318 supersedes #15310** — the ViewModel-owned `_baseNewsList` deletion is the more complete fix and makes `removePost`'s list logic obsolete. So if this PR is accepted, **close #15310**; if the team prefers the low-risk path, merge #15310 now and rebase this (the re-deletion is trivial). Either way it needs an explicit decision before merge, and right now both are queued.

**Manual verification before merge (CI can't cover these UI flows):** courses sort persists correctly across a reload after you've toggled it; voices parent-post deletion animates cleanly (the exact case #15310 targets); survey date/title sort still works; Life drag-reorder is smooth.

**Verdict:** Correctness issues resolved — good work — but it collides head-on with the `ready to merge` #15310 and bundles five adapters' worth of untested UI behavior. Reconcile with #15310 (I'd keep this one) and smoke-test the flows, then it's ready.

---

dogi:
**Re-review — 62/100 — good direction, but it collides with a ready-to-merge PR and shifts a burden that needs verifying**

The overall direction is right and several pieces are clean. CI is green. But two things block it.

**1. Direct conflict with #15310 (`ready to merge`, ⭐⭐⭐, mergeable_state `clean`).**
Both PRs rewrite `VoicesAdapter.removePost`. #15310 is a surgical one-file fix already blessed for merge; this PR deletes the method's entire body and relocates removal to `VoicesViewModel.deletePost`. They cannot both land — this will clobber the approved fix. #15310 should merge first; this PR then needs to rebase and reconcile its Voices changes on top of it. Please don't merge this ahead of #15310.

**2. Emptying `removePost` shifts all list-removal onto reactive `_baseNewsList` observation — verify all three call sites re-render.**
`removePost` now only decrements `replyCountCache`; the row only disappears if `VoicesViewModel.deletePost` (which filters `_baseNewsList`) runs *and* the surface observes the derived list flow. There are three callers:
- `VoicesFragment.kt:260`
- `TeamsVoicesFragment.kt:234`
- `ReplyActivity.kt:138` ← this is the risky one

`ReplyActivity` shows a single reply thread; please confirm it goes through `viewModel.deletePost` and re-collects the `_baseNewsList`-derived flow. If it calls `adapter.removePost(...)` without that reactive path, deleting a reply there will decrement the count but leave the row on screen — a regression the old in-adapter `submitList` handled directly.

**What's genuinely good (keep):**
- `LifeAdapter`: replacing the permanent `workingList` with an ad-hoc `dragList` + `notifyItemMoved` during drag and a single `submitList` at `onItemMoveFinished` is the correct drag-reorder pattern. Nice.
- `CoursesAdapter` → `CoursesViewModel`: moving sort state (`activeSort`, title/date direction) into the VM and having `processCourses` re-apply it on reload is the right home for it; sort now survives a reload.
- `SurveysAdapter`: I verified `sortByDate`/`toggleTitleSortOrder` have **zero callers** in the tree — they were already dead code, so removing them is safe (no survey-sort regression, despite the description implying a relocation).
- `ChatHistoryAdapter`: reading `currentList` from the live adapter instead of a captured `var` is fine and reads cleaner.

**Verdict:** Solid refactor, but sequence it behind #15310 and confirm the `ReplyActivity` deletion path re-renders reactively. Changes requested.

---

Okuro3499 (approved):

dogi:
**Re-review (commit f82244bf) — 68/100 — progress, but the #15310 collision got sharper**

Thanks for the update. One of my two concerns is resolved; the other is now more concrete.

**Resolved ✅ — the `ReplyActivity` removal path.** `removePost` no longer offloads all list removal onto reactive `_baseNewsList` observation; it keeps the in-adapter manipulation (removes from `currentList`, recomputes `children`, `super.submitList(prepareSubmitList(children))`). So deletions still render at every call site, including `ReplyActivity`. Good — and dropping the `originalList` cache is the right cleanup.

**Still blocking ⚠️ — conflict with (and now a partial regression of) #15310.** #15310 (`ready to merge`, ⭐⭐⭐, fixes #15407 "smoother voices removing") edits this exact `removePost`, and its entire point is to **remove** the `submitList(emptyList())` on the parent branch so DiffUtil animates instead of blanking the list. This PR's new parent branch does the opposite:
```kotlin
if (isParent) {
    parentNews = null
    super.submitList(emptyList())   // ← reintroduces the empty-submit #15310 deletes
}
```
So the two PRs both textually conflict on `removePost` **and** take opposite approaches on the `isParent` case. If this lands after #15310, it re-introduces the #15407 jitter that #15310 just fixed.

**Recommendation:** sequence this behind #15310. Let #15310 merge its jitter fix (null `parentNews` + fall through to the filtered submit, no empty submit), then rebase this PR and layer the `originalList`-cache removal on top of that shape. That keeps both wins.

**Unchanged from my last pass (all still good):** `LifeAdapter` ad-hoc `dragList` + `notifyItemMoved`, `CoursesViewModel` sort ownership, `SurveysAdapter` dead-code removal, `ChatHistoryAdapter` live `currentList`. CI green across all three jobs.

**Heads-up on a new three-way overlap:** #15412 (courses landscape) now also modifies `CoursesAdapter.removeCourses`, `CoursesFragment`, and `CoursesViewModel` — so Courses is contested by this PR and #15412. Worth deciding a merge order for those too.

**Verdict:** Good progress, but rebase behind #15310 so the parent-removal branch doesn't undo its smoothness fix. Changes requested.

---

dogi:
**Review — rating: 74/100**

Large diff, so I reviewed at hunk level and verified the two things that could have been silent breakage — both came back clean:

- **`SurveysAdapter`**: `sortByDate` and `toggleTitleSortOrder` have no call sites anywhere in `app/src/main`, so this is genuine dead-code removal, not a quietly dropped sort feature. (Worth stating explicitly, because the Courses hunks *move* sorting to the ViewModel while this one just deletes it — the asymmetry looks alarming until you check.)
- **`VoicesAdapter`**: `submitList` is overridden at lines 137–142 and already funnels through `prepareSubmitList`, so `super.submitList(prepareSubmitList(children))` is exactly equivalent to the override and avoids double-applying it. Correct, if a little obscure.

The overall direction is right and consistent with #15282, already at `merge`. Replacing `removeCourses` with a `viewModel.loadCourses(...)` reload is a real improvement — one source of truth instead of the adapter mutating its own list behind the ViewModel's back. Deriving children from `currentList` instead of a parallel `originalList` removes a genuine desync hazard. And `LifeAdapter` batching the drag into one `submitList` at `onItemMoveFinished`, instead of submitting on every single move, is the better pattern.

**The blocker is a user-visible regression in the Courses title sort.** The old adapter started at `isTitleAscending = false`, so the first tap toggled it to `true` and sorted A→Z. The new ViewModel starts at `isTitleAscending = true`, so the first tap toggles to `false` and sorts **Z→A**. First tap on "order by title" now sorts backwards. Date sort is unaffected — old `isAscending = true` and new `isDateAscending = true` both land on descending after one toggle, so that one is faithfully preserved. Setting `isTitleAscending = false` in the ViewModel restores parity.

Two smaller things worth addressing while you're in here:

- **`ChatHistoryAdapter`**: reading `(listView.adapter as ChatShareTargetAdapter).currentList` inside the click lambda swaps a synchronously-updated local for `AsyncListDiffer`'s *committed* list, which lags until the diff commits. Two quick taps on a group header can compute the expanded set from the pre-commit list and toggle the wrong way. `shareTargetAdapter` is already in scope — use it directly and drop the cast through the view, which also removes a `ClassCastException` path if the adapter is ever swapped.
- **`LifeAdapter`**: `notifyItemMoved` during the drag leaves the differ's `currentList` untouched, so `getItem(pos)` returns pre-move data for any rebind that happens mid-drag, and the closing `submitList(finalList)` makes DiffUtil re-derive and re-dispatch moves that were already applied visually. Ends in the right state, but expect duplicate move animations.

**Nits:** `sortCourses` spells out `org.ole.planet.myplanet.model.Course` inline instead of importing it, unlike the rest of the file. And `scrollToTop()` now runs immediately after `toggleTitleSort()` rather than in the old `submitList` completion callback, so it fires before the reordered list is committed.

**Heads-up on ordering:** this touches `VoicesAdapter` in the same regions as #15259 (now `ready`) and #15383 (`merge`). Not duplicate work — three different concerns — but whichever lands last will need a rebase.

**Verdict:** Needs another round — flip the `isTitleAscending` seed and this is close. Marking `change`.

---

dogi:
**Re-review — Rating: 84/100** (was 74)

The blocker is fixed. `isTitleAscending` is seeded to `false` in `CoursesViewModel`, so the first tap toggles it to `true` and sorts A→Z, matching the adapter behavior this replaces. I checked both directions against the old code rather than just the changed line:

- **Title:** old adapter seeded `false` → toggle → `true` → `compareTo` ascending. New: seeded `false` → toggle → `true` → `sortedBy { courseTitle.lowercase() }` ascending. Match.
- **Date:** old seeded `true` → toggle → `false` → `course2.compareTo(course1)` descending. New: seeded `true` → toggle → `false` → `sortedByDescending`. Match.

`sortCourses` also now uses the imported `Course` type instead of the fully-qualified name, and the `ChatHistoryAdapter` lambda reads `shareTargetAdapter.currentList` directly — the cast through `listView.adapter` is gone along with its `ClassCastException` path.

Everything else I raised last round was explicitly a nit or a caveat, and those stand unchanged: reading `currentList` still sees the *committed* list, so two very fast taps on a group header can compute the expanded set from pre-commit state; and `LifeAdapter` still pairs `notifyItemMoved` during the drag with a closing `submitList`, which lands in the right state but re-dispatches moves already applied visually. Neither blocks. CI green on build (default + lite) and test (default).

**Verdict:** Mergeable, and it clears the merge gate — Okuro3499 approved and the PR carries `⭐⭐⭐.`, so moving it to `merge`. One thing for the maintainer to weigh: that approval is on commit `f82244bf`, and the sort-seed fix landed after it, so if you'd rather have a fresh sign-off on the current head, this is the PR to ask for one on.

**Heads-up:** #15446 also restructures `CoursesAdapter`/`CoursesFragment`, #15306 touches `CoursesFragment`, and #15443 touches `LifeAdapter` — whichever lands last needs a rebase. #15443 adding a real `MyLife.equals`/`hashCode` actually helps this PR's closing `submitList`, since the reorder can finally be diffed structurally.

---

PR #15319, opened 2026-08-04
Guard SyncManager and SyncTimeLogger debug logging

dogi:
Guard `SyncManager` and `SyncTimeLogger` debug logging

* Replaced all 20 `Log.d("SyncPerf", ...)` calls in `SyncManager.kt` with `if (BuildConfig.DEBUG) Log.d("SyncPerf", ...)` to prevent string formatting overhead in release builds.
* Replaced all `Log.d("SyncPerf", ...)` calls in `SyncTimeLogger.kt` with `if (BuildConfig.DEBUG) Log.d("SyncPerf", ...)`.
* Added `import org.ole.planet.myplanet.BuildConfig` to both files.
* Ran unit tests for `org.ole.planet.myplanet.services.sync.*` to ensure no regressions.

---
*PR created automatically by Jules for task [12890280953934138960](https://jules.google.com/task/12890280953934138960) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/services/sync/SyncManager.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/services/sync/SyncManager.kt
@@ -9,6 +9,7 @@ import android.net.wifi.WifiManager
 import android.os.Build
 import android.os.SystemClock
 import android.util.Log
+import org.ole.planet.myplanet.BuildConfig
 import androidx.core.content.edit
 import com.google.gson.JsonArray
 import com.google.gson.JsonNull
@@ -141,9 +142,9 @@ class SyncManager @Inject constructor(
         val syncStartTime = SystemClock.elapsedRealtime()
         val logger = SyncTimeLogger
         logger.startLogging()
-        Log.d("SyncPerf", "═══════════════════════════════════════════════════════════════")
-        Log.d("SyncPerf", "FULL SYNC STARTED at ${java.text.SimpleDateFormat("HH:mm:ss.SSS").format(Date())}")
-        Log.d("SyncPerf", "═══════════════════════════════════════════════════════════════")
+        if (BuildConfig.DEBUG) Log.d("SyncPerf", "═══════════════════════════════════════════════════════════════")
+        if (BuildConfig.DEBUG) Log.d("SyncPerf", "FULL SYNC STARTED at ${java.text.SimpleDateFormat("HH:mm:ss.SSS").format(Date())}")
+        if (BuildConfig.DEBUG) Log.d("SyncPerf", "═══════════════════════════════════════════════════════════════")
         try {
 
             initializeSync()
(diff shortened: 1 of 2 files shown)


dogi:
**Review — rating: 78/100**

**Strengths**
- Legitimate release-build hygiene: `Log.d("SyncPerf", …)` is *not* stripped in release here (`minifyEnabled = false` per the project setup), so these calls — several of which build strings via `java.text.SimpleDateFormat("HH:mm:ss.SSS").format(Date())` and heavy interpolation on every sync — genuinely execute in production today. Guarding them with `if (BuildConfig.DEBUG)` skips both the log and the string construction in release. The most valuable ones are the `SimpleDateFormat(...).format(Date())` timestamps now behind the guard.
- Purely non-functional: `Log.d` has no user-visible effect, so wrapping it changes nothing behaviorally. CI green on all three checks including `test (default)`.

**Concerns**
- **Repetitive style.** `if (BuildConfig.DEBUG) Log.d(...)` on ~20 lines is noisy; a small private helper (e.g. `private inline fun perf(msg: () -> String) { if (BuildConfig.DEBUG) Log.d("SyncPerf", msg()) }`) would collapse the guard to one place and also defer the string until after the check. Not required, but it'd read better and scale.
- **Incomplete coverage.** The `logger.logDetail(...)`/`logApiCall(...)`/`logRealmOperation(...)` call sites in `SyncManager` still build their message strings unconditionally (the `BuildConfig.DEBUG` guard is only inside the logger's `Log.d`), so the interpolation cost there isn't saved. Those also feed the in-memory summary maps, so they can't simply be dropped — just noting the win is narrower than "guard all SyncPerf logging."
- Import order nit: `import org.ole.planet.myplanet.BuildConfig` is placed before the `androidx.*` imports in both files.

**Verdict:** Correct, safe, behavior-neutral logging optimization with a real (if modest) release benefit; mergeable — consider the helper to reduce the repetition.

---

dogi:

The repetition concern from my earlier review is addressed cleanly: both `SyncManager` and `SyncTimeLogger` now define

```kotlin
private inline fun perf(msg: () -> String) { if (BuildConfig.DEBUG) Log.d("SyncPerf", msg()) }
```

and route every `Log.d("SyncPerf", ...)` through `perf { ... }`. Because it's `inline` and `BuildConfig.DEBUG` is a compile-time constant, in release the whole call — **including the string interpolation and the `SimpleDateFormat(...).format(Date())` allocations** — is eliminated, which is a strictly better outcome than the per-line guards I reviewed before. Behavior in debug is unchanged; release simply stops emitting these logs (no user-visible effect). Build (both flavors) + test green.

**Minor (non-blocking, unchanged from before):**
- `logDetail`/`logApiCall`/`logRealmOperation` still build their message strings unconditionally because they also feed the in-memory summary maps — so the win is narrower than "all SyncPerf logging," which is fine and expected.
- `import org.ole.planet.myplanet.BuildConfig` is ordered before the `androidx.*` imports (ktlint nit).

**Verdict:** Correct, behavior-neutral release-logging optimization, now with the helper that removes the repetition — mergeable.

---

dogi:
**Review — rating: 84/100**

Correct and more valuable than it first looks. Because `perf` is an `inline fun` taking a `() -> String`, the message lambda is inlined into the `if (BuildConfig.DEBUG)` branch, so in release builds the *string construction* disappears too — not just the `Log.d` call. That matters here because these sites do real work to build their messages: `SimpleDateFormat("HH:mm:ss.SSS").format(Date())` per sync phase, and per-batch percentage math inside `resourceTransactionSync`. And since `minifyEnabled` is `false` for release in this project, nothing else was stripping these — this guard is the only thing removing them.

The check I cared about most was whether this silently degrades release diagnostics, and it doesn't. In `SyncTimeLogger`, the data collection all happens *before* the guarded call — `detailedLogs.getOrPut(context) { mutableListOf() }.add(message)`, the API/DB counters, `apiCallTimes`, `realmOperationTimes` — and `stopLogging` still runs `generateSummary()` and `saveSummaryToRealm(summary, uploadManager)` unguarded. So the summary that actually gets persisted and uploaded is byte-for-byte what it was; only logcat output goes quiet in release. That's exactly the right line to draw.

**Nits, none blocking:**
- The `perf` helper is defined privately and identically in both files. It belongs in a shared util (`utils/`) so the third caller doesn't copy it a third time.
- Converting every message to a triple-quoted raw string is unnecessary churn — none of them contain quotes or escapes, it inflates the diff, and raw strings silently don't process escapes, so a later `\n` won't do what its author expects. Plain `"..."` would have kept the diff to just the `Log.d` → `perf` swap.
- `import org.ole.planet.myplanet.BuildConfig` is wedged between `android.util.Log` and `androidx.core.content.edit` in both files, breaking the grouping convention in `docs/CODE_STYLE_GUIDE.md`.

**Verdict:** Mergeable.

---

Okuro3499 (commented):
This is an unnecessary change for debug logs which can be removed at any moment

PR #15321, opened 2026-08-04
courses: smoother surveys repositories counting (fixes #15473)

### LINKED ISSUE #15473: Refactor repository queries to use COUNT(*)
Tracking issue for PR #15321.

dogi:
A11: Use COUNT(*) where the code only reads .size

---
*PR created automatically by Jules for task [14284269125738412683](https://jules.google.com/task/14284269125738412683) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6404
-        versionName = "0.64.4"
+        versionCode = 6405
+        versionName = "0.64.5"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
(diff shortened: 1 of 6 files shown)


dogi:
**Review — rating: 86/100**

**Strengths**
- Correct, behavior-preserving perf optimization. Both new DAO queries keep the exact same `WHERE` predicates as the `SELECT *` versions they replace — `exams WHERE courseId = :courseId AND type = :type` and `submissions WHERE userId = :userId AND status = 'pending' AND type = 'survey'` — so `COUNT(*)` returns precisely the old `list.size`, while skipping row hydration/object mapping for what was always a count-only use (`getCourseExamCount`, `getSurveySubmissionCount`).
- Tests were updated to match the new DAO calls (`countByCourseIdAndType returns 5`, `countPendingSurveys returns 2`) rather than deleted, and CI is green on all three checks including `test (default)`.
- Both call sites already guarded the empty-id case (`return 0`), so no null/empty regression.

**Concerns**
- None blocking. Minor: the codebase now has both `getByCourseIdAndType`/`getPendingSurveys` and their `count*` siblings; if the list-returning variants have no remaining callers they'd be worth pruning in a follow-up, but that's out of scope here.

**Verdict:** Textbook `SELECT *.size` → `COUNT(*)` conversion with predicates preserved and tests updated; ready to merge.

---

Okuro3499 (approved):

PR #15326, opened 2026-08-05
Refactor IO-bound logic to ViewModels

dogi:
Addresses A26 by ensuring IO operations are properly scoped to ViewModels rather than UI lifecycles, preventing leaked scopes, simplifying tests, and avoiding duplicate work during configuration changes. Fixed compiler issues relating to UI fragments missing viewmodels.

---
*PR created automatically by Jules for task [15355967127027005945](https://jules.google.com/task/15355967127027005945) started by @dogi*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved exam reset and restart flows by reliably removing previous submissions and preserving answer state.
  * Improved PDF text extraction reliability, including safer handling of extraction errors.
  * Improved sync completion handling for language changes and application updates.
  * Improved resource deletion and search activity saving consistency.
  * Improved course sync-action visibility checks in Markdown dialogs.
  * Improved exam submissions, answer saving, certification checks, and course-progress updates.


dogi:
**Review — rating: 55/100**

The direction (IO logic → ViewModels) is right and a few of the conversions are genuinely good, but one of them introduces an ordering regression and another passes an Activity into a ViewModel. CI green only proves it compiles/existing tests pass — none of these paths are unit-tested, so the races below aren't caught.

**Concerns**

- **`ResourcesFragment.kt:718` — refresh now races ahead of the delete (regression).** Before, the toast + `refreshResourcesData()` + `selectedItems?.clear()` ran *inside* `withContext(dispatcherProvider.main)` **after** the `itemsToDelete.forEach { removeResourceFromShelf(...) }` loop completed, so the UI refreshed against the post-delete DB state. Now `viewModel.deleteResources(userId, itemsToDelete)` is fire-and-forget on `viewModelScope`, and `refreshResourcesData()` runs synchronously on the very next line — before the shelf removals commit. The list will frequently refresh to the *stale* pre-delete state (removed items still showing). To preserve behavior, `deleteResources` needs to be `suspend` and awaited (fragment launches, calls it, then refreshes), or expose a completion signal the fragment collects before refreshing.

- **`MarkdownViewModel.logSyncInSharedPrefs(activity)` — Activity passed into and retained by a ViewModel.** `viewModelScope.launch(io) { activity.logSyncInSharedPrefs() }` holds a `DashboardElementActivity` reference inside VM-scoped coroutine state, which is exactly the leak the ViewModel boundary is meant to prevent, and defeats the point of the refactor. The `logSyncInSharedPrefs` logic should move into the VM/repository (operating on prefs/data, not the Activity), or stay in the fragment.

- **`SyncActivity.kt:746` — `launch(io)` → `withContext(io)` changes clearAllData from fire-and-forget to awaited.** This is probably the more correct behavior, but it *is* a semantic change (the enclosing coroutine now blocks on `clearAllData()` before proceeding) rather than a pure move — worth calling out and confirming the surrounding flow expects the wait.

- **Style: fully-qualified names instead of imports.** `ResourceViewerViewModel.extractPdfText` and `ResourcesViewModel.getLibraryListModels`/`saveSearchActivity` use inline `kotlinx.coroutines.withContext`, `com.tom_roush.pdfbox.*`, `android.content.Context`, `java.io.File` even though `withContext` is already imported in those files. Add imports for consistency.

**Strengths**
- `ExamTakingFragment` conversion is correct: the DB delete goes through `viewModel.deleteExamSubmissions` (which wraps `withContext(io)`), and the cache/UI mutations correctly run on the main-dispatched `lifecycleScope.launch` — the old manual `withContext(main)` hop is no longer needed.
- `ResourceViewerFragment` PDF extraction move is clean and also upgrades `lifecycleScope` → `viewLifecycleOwner.lifecycleScope`, which is the safer choice for a fragment view.

**Verdict:** Good intent with two real problems — the ResourcesFragment refresh/delete ordering regression and the Activity-in-ViewModel leak — that need fixing before merge.

---

dogi:
**Re-review — Rating: 85/100** (was 55)

Both blocking issues from my earlier review are resolved, verified against the current diff:

- ✅ **ResourcesFragment delete/refresh ordering — fixed.** `viewModel.deleteResources(...)` is now a `suspend` function wrapping `withContext(dispatcherProvider.io) { itemsToDelete.forEach { removeResourceFromShelf(...) } }`, and the fragment `launch { deleteResources(...); toast; refreshResourcesData(); ... }` **awaits** it before refreshing. The list now refreshes against the post-delete state — the race is gone.
- ✅ **Activity-in-ViewModel leak — fixed.** `MarkdownViewModel` no longer takes or retains an `Activity`; it only wraps `userRepository`. The `logSyncInSharedPrefs()` call was moved back to the fragment (`(activity as DashboardElementActivity).logSyncInSharedPrefs()`).

On the follow-up point — the fragment now calls `logSyncInSharedPrefs()` synchronously rather than via `launch(io)` — I checked the method's other call sites: `DashboardActivity` (sync click listener + `action_sync`) and `DashboardElementActivity` (`action_sync`) all invoke it **directly on the main thread**. So the old `launch(io)` wrapper here was the outlier; the change makes MarkdownDialogFragment consistent with the rest of the app, and this method is a UI-triggered prefs action, not an IO hotspot. Not a regression. Build + test green.

**Minor (non-blocking):**
- `SyncActivity` — `launch(io)` → `withContext(io)` around `clearAllData()` is still a fire-and-forget→awaited semantic change (likely more correct; just confirm the surrounding flow wants the wait).
- qlty flags `saveSearchActivity`'s 9 params — that's a pass-through mirroring the repository signature, not new complexity, but a param object would quiet it.
- New imports in `ResourceViewerViewModel` are appended out of alphabetical order (ktlint nit).

**Verdict:** The two real defects are fixed and the remaining items are cosmetic — mergeable.

---

Okuro3499 (changes requested):
```
While moving IO-bound work to ViewModels is the right architectural direction, the branch contains an incomplete implementation that leaves dead code in ResourcesViewModel.kt
  because ResourcesFragment.kt was not updated to use the new ViewModel method. Additionally, other components like ExamTakingFragment.kt were only partially refactored, leaving
  direct repository injection in place alongside ViewModel delegates.

File affected: ResourcesFragment.kt & ResourcesViewModel.kt
      • Problem: ResourcesViewModel implements saveSearchActivity(), but ResourcesFragment.saveSearchActivity() (line 642) still invokes resourcesRepository.saveSearchActivity()
      directly in a lifecycleScope.launch(dispatcherProvider.io) block. This leaves the new ViewModel method unused.
      • Fix: Update ResourcesFragment.saveSearchActivity() to delegate to viewModel.saveSearchActivity(...) and remove the manual IO coroutine launch and direct
      resourcesRepository call.
```

dogi:
**Review — rating: 84/100**

Apologies for the wait on this one — it sat unreviewed while the rest of the queue moved.

Solid, consistent refactor. The IO-bound work genuinely lands behind ViewModels with `withContext(dispatcherProvider.io)` — exam submission deletion, resource deletion and search-activity logging, and PDF text extraction — and several call sites correctly move from `lifecycleScope` to `viewLifecycleOwner.lifecycleScope` along the way. In `deleteSelected`, dropping the `_binding ?: return@withContext` guard is right rather than risky, because the view-scoped coroutine is already cancelled at view destruction.

**Two hunks look like they move IO onto the main thread, and I checked both — they're correct, and one is a real bug fix.**

`MarkdownDialogFragment` now calls `logSyncInSharedPrefs()` directly instead of inside `launch(dispatcherProvider.io)`. Despite the name, that method reads three cached SharedPreferences values and then does `DialogServerUrlBinding.inflate(LayoutInflater.from(this))`, wraps a `ContextThemeWrapper` and shows an `AlertDialog`. It's a UI method — inflating views and showing a dialog on an IO dispatcher was the actual bug, and calling it on the main thread fixes it.

Same story in `SyncActivity`: `getPendingLanguageChange()` is `pref.getString(...)` against the in-memory map and `setPendingLanguageChange` uses `pref.edit { }`, which applies asynchronously — so neither blocks. The block then calls `LocaleUtils.setLocale(...)` and `recreate()`, both main-thread-only. The old shape dispatched to IO purely to hop straight back to main; removing that round trip is a simplification with no regression.

Also worth noting: changing `clearAllData()` from `launch(io)` fire-and-forget to `withContext(io)` means it's now awaited rather than racing whatever follows it. That's the safer semantics for a data wipe.

**Nits:**
- `MarkdownViewModel` is a pure two-method passthrough to `UserRepository`. A ViewModel wrapping a repository is ordinary MVVM so this is fine, but it's worth being aware it adds a layer without adding behavior.
- `ResourceViewerViewModel.extractPdfText(context: Context, file: File)` takes a `Context` parameter. You pass `applicationContext`, so there's no leak — but injecting `@ApplicationContext` into the ViewModel via Hilt would keep `Context` out of the method signature entirely.
- No tests. `deleteResources`, `saveSearchActivity` and `deleteExamSubmissions` are thin, but there's ample ViewModel-test precedent in the repo.
- Import ordering drifts in `ResourcesViewModel` (`model.TagEntity` after `utils.DispatcherProvider`) and `ResourceViewerViewModel` (android/java/com appended below the `org.ole` block).

CI green on build (default + lite) and test (default).

**Verdict:** Mergeable.

---

Okuro3499 (changes requested):
```
The ExamTakingFragment claim is still valid. It mixes patterns: viewModel.deleteExamSubmissions(...) goes through the ViewModel, but surveysRepository.getExamQuestions(...),      
  submissionsRepository.getSubmissionsByParentId(...), coursesRepository.isCourseCertified(...), and submissionsRepository.createExamSubmission(...) are all still called directly   
  from the fragment inside viewLifecycleOwner.lifecycleScope.launch { ... } (app/src/main/java/org/ole/planet/myplanet/ui/exam/ExamTakingFragment.kt:65-68, 103-176), with the three 
  repositories still @Injected straight into the fragment.
```

Okuro3499 (changes requested):
```
ExamTakingViewModel already implements getExamQuestions, getSubmissionsByParentId, isCourseCertified,
and createExamSubmission as pass-through IO-dispatched wrappers, but ExamTakingFragment bypasses all four and calls the repositories directly instead. That makes those four       
ViewModel methods genuinely dead code (unused, not just "wired oddly"), while the fragment still does the raw repository calls on viewLifecycleOwner.lifecycleScope.launch the   
exact anti-pattern the refactor was meant to eliminate.    
```

PR #15368, opened 2026-08-05
sync: smoother download service url resolving (fixes #15494)

### LINKED ISSUE #15494: 🧹 Refactor DownloadService alternative URL resolution to improve code health
Tracking issue for PR #15368.

dogi:
🎯 **What:** Extracted the deeply nested logic for resolving the alternative fallback URL within `DownloadService.kt` into a private helper function.
💡 **Why:** The `initDownload` function contained a large conditional block nested inside another `if` block, making it hard to follow. Extracting this logic improves code readability and maintainability.
✅ **Verification:** Verified by checking `git diff` after making the changes, ensuring no behavior is altered, and successfully running `./gradlew test`.
✨ **Result:** A more readable `initDownload` flow that isolates URL fallback configuration.

---
*PR created automatically by Jules for task [4717371049008821760](https://jules.google.com/task/4717371049008821760) started by @dogi*


dogi:
**Review — rating: 84/100**

Extracts the alternative-URL resolution block out of `initDownload` into `resolveAlternativeUrl(url, fileName): String?`. I traced the logic against the original: the hardcoded-mapping path, the stored-alternative fallback (with `trimEnd('/')`), the `primaryBase != null` requirement on both paths, and the `Uri.parse` path/query reassembly are all preserved, returning `null` in exactly the same cases. This meaningfully shrinks a long method. Build (default + lite) and test (default) green.

**Cross-PR note:** #15380 restructures the *same* block inline via a `when` (also correct and equivalent) but leaves it inside `initDownload`. This PR's method extraction is the better cleanup of the two → **recommend keeping #15368 and closing #15380.**

**Concern (nit):** the extracted helper still logs with the `"initDownload: ..."` prefix in its messages; consider updating to `"resolveAlternativeUrl: ..."` now that the code has moved, so logs point at the right function.

**Verdict:** Correct and the better of the pair — keep this one.

---

Okuro3499 (approved):

PR #15367, opened 2026-08-05
sync: smoother download file servicing (fixes #15427)

dogi:
🎯 **What:** Extracted the innermost file stream copying loop from `DownloadService.downloadFile()` into a dedicated helper function `copyStreamWithProgress()`.
💡 **Why:** The original `downloadFile` function was deeply nested, reducing code readability and maintainability. Flattening the structure improves readability.
✅ **Verification:** Ran `./gradlew assembleDebug` and `./gradlew test` with the specific `ANDROID_HOME` correctly set, verifying that compilation and tests succeed.
✨ **Result:** Improved maintainability and reduced cognitive load when analyzing the file download workflow.

---
*PR created automatically by Jules for task [16954190174290243298](https://jules.google.com/task/16954190174290243298) started by @dogi*

PR #15336, opened 2026-08-05
🧹 [Code Health] Reduce deep nesting in DownloadService.downloadFile

dogi:
🎯 **What:** Extracted the deeply nested stream copying loop in `DownloadService.downloadFile` into a helper function `writeStream`.
💡 **Why:** Reduces nesting, improving readability and maintainability of the `downloadFile` function.
✅ **Verification:** Verified by ensuring the code behaves exactly the same, correctly handles parameter passing for `total` byte tracking, and passes all existing debug unit tests.
✨ **Result:** A cleaner, flatter method structure for file downloads.

---
*PR created automatically by Jules for task [5954717484937304850](https://jules.google.com/task/5954717484937304850) started by @dogi*

PR #15369, opened 2026-08-05
🧹 Refactor: Extract stream writing logic in DownloadService

dogi:
🎯 **What:** Extracted the innermost block that reads from `BufferedInputStream` and writes to `FileOutputStream` in `DownloadService.downloadFile` into a separate helper function `writeStreamContent`.
💡 **Why:** The `downloadFile` method contained deeply nested code (up to 8 levels deep inside try/use/while/if blocks) which made it hard to read and maintain. By extracting this logic, the code becomes flatter and easier to follow.
✅ **Verification:** Re-compiled the Kotlin files and ran the unit tests for `DownloadService` (`./gradlew testDefaultDebugUnitTest --tests "org.ole.planet.myplanet.services.DownloadServiceTest"`), which passed successfully. 
✨ **Result:** Improved readability and reduced code complexity in `DownloadService` without changing the functionality.

---
*PR created automatically by Jules for task [963921464134706871](https://jules.google.com/task/963921464134706871) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/services/DownloadService.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/services/DownloadService.kt
@@ -357,42 +357,7 @@ class DownloadService : Service() {
         try {
             BufferedInputStream(body.byteStream(), 1024 * 8).use { bis ->
                 FileOutputStream(tempFile).use { output ->
-                    val download = Download().apply {
-                        this.fileName = getFileNameFromUrl(url)
-                    }
-
-                    if (fileSize > 0) {
-                        totalFileSize = (fileSize / 1024.0).toInt()
-                        download.totalFileSize = totalFileSize
-                    } else {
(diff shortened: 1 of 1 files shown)


dogi:
**Review — rating: 73/100 (duplicate)**

Correct and CI-green. This one extracts the *entire* inner block of `downloadFile` (the `Download` setup plus the read/write loop) into `writeStreamContent(...)`, which reduces nesting slightly more than #15367. It's a legitimate alternative, but it targets the same region as #15367 / #15336 / #15372, so only one can merge.

I'm recommending #15367 as the keeper (minimal extraction, preserves the exact `total` accumulation via `initialTotal`). If the team instead prefers extracting the whole block, this PR is the better choice than #15367 — but pick exactly one. Marking `duplicate` against the #15367 recommendation; flag me if you'd rather flip the winner.

---

PR #15372, opened 2026-08-05
🧹 Refactor nested `when` expressions to use early returns in DownloadService

dogi:
🎯 **What:** Refactored the deeply nested `when` structures inside `DownloadService.tryDownloadFromResult` by extracting guard checks and utilizing early returns.
💡 **Why:** To drastically improve the code's readability and code health by reducing the cognitive complexity caused by heavy block nesting.
✅ **Verification:** Verified via `./gradlew assembleDebug` (compilation passes) and `./gradlew test` (no logic broken). The code review agent confirmed exact logic equivalence.
✨ **Result:** A flatter, easier-to-read function that correctly casts and continues execution on valid states, immediately returning on error/invalid states.

---
*PR created automatically by Jules for task [13332932228770991970](https://jules.google.com/task/13332932228770991970) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/services/DownloadService.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/services/DownloadService.kt
@@ -283,39 +283,37 @@ class DownloadService : Service() {
         isAlternative: Boolean
     ): Boolean {
         val source = if (isAlternative) "alternative" else "primary"
-        return when (result) {
-            is DownloadResult.Success -> {
-                val contentLength = result.body.contentLength()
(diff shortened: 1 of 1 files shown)


dogi:

Correct and CI-green — functionally equivalent to #15367 (extracts the same `downloadFile` loop into a helper that takes `initialTotal` and returns `total`, using imported types). The only difference is the added `@Throws(IOException::class)` annotation, which is harmless but not required in Kotlin.

Both edit the same region, so only one can merge. **Recommend closing in favor of #15367.** Marking `duplicate`.

---

PR #15366, opened 2026-08-05
sync: smoother download service early returning (fixes #15493)

dogi:
🎯 **What:** The code health issue addressed was deeply nested code (nested `when` and `try-catch` blocks) in the `tryDownloadFromResult` method of `DownloadService.kt`.
💡 **Why:** Flattening the nested structures using early returns (guard clauses) improves readability and maintainability. It reduces cognitive load by eliminating deep indentation and clearly showing error/edge cases upfront.
✅ **Verification:** Verified the logic preservation manually and ran `DownloadServiceTest` using `./gradlew testDefaultDebugUnitTest --tests "org.ole.planet.myplanet.services.DownloadServiceTest"`, ensuring all relevant tests passed. Requested a code review which was approved for correctness and safety.
✨ **Result:** The `tryDownloadFromResult` method is now completely flattened and significantly more readable.

---
*PR created automatically by Jules for task [10723457576012081755](https://jules.google.com/task/10723457576012081755) started by @dogi*

PR #15376, opened 2026-08-05
🧹 [Code Health] Flatten deeply nested expression in DownloadService

dogi:
🎯 **What:** Refactored the `tryDownloadFromResult` function in `DownloadService.kt` to flatten a deeply nested `when` expression.
💡 **Why:** Deeply nested `when` expressions, particularly those that implicitly return boolean flags out to an outer block, are hard to read and mentally trace. Converting this to sequential `if` checks with early returns simplifies the control flow, reduces nesting, and improves maintainability.
✅ **Verification:** Verified by inspecting the target file and running the `:app:testDefaultDebugUnitTest --tests '*DownloadServiceTest*'` targets to ensure compilation and existing test passes.
✨ **Result:** Improved code readability and reduced nesting depth without changing behavior.

---
*PR created automatically by Jules for task [9660077719850729880](https://jules.google.com/task/9660077719850729880) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/services/DownloadService.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/services/DownloadService.kt
@@ -288,27 +288,25 @@ class DownloadService : Service() {
                 val contentLength = result.body.contentLength()
                 Log.d(TAG, "tryDownload [$source]: $fileName responded contentLength=${if (contentLength == -1L) "unknown" else "${contentLength}B"}")
                 val storageError = getStorageError(contentLength)
-                when {
-                    storageError != null -> {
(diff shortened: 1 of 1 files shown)


dogi:
**Review — rating: 66/100 (superseded)**

Same target as #15366 (`tryDownloadFromResult`), but this PR flattens only the *inner* `when` inside the `is DownloadResult.Success ->` branch while leaving the outer `return when (result)` in place. The result is a hybrid where `return false` / `return try {…}` statements sit inside a branch of a `return when` — it compiles and behaves correctly, but it's a messier shape than #15366's full flatten of both branches into top-level guard clauses.

Both edit the same lines, so they conflict. **Recommend closing this in favor of #15366.** Marking `duplicate`.

---

PR #15380, opened 2026-08-05
🧹 [Code Health] Refactor deeply nested URL fallback logic in DownloadService

dogi:
🎯 **What:** The code health issue addressed
Deeply nested code in `DownloadService.kt` for initializing fallback URL alternative mappings.

💡 **Why:** How this improves maintainability
The deeply nested conditional statements were difficult to read and parse mentally. By refactoring it into a flat `when` expression, the conditions evaluating `altBase`, `primaryBase`, and `storedAlt` become straightforward and decoupled, greatly improving code legibility without changing any business rules.

✅ **Verification:** How you confirmed the change is safe
Ran `./gradlew testDefaultDebugUnitTest` to ensure all functionality and tests remain green.

✨ **Result:** The improvement achieved
A clean and flat mapping block in `initDownload` that is easy to comprehend for future maintenance.

---
*PR created automatically by Jules for task [1608499241577318893](https://jules.google.com/task/1608499241577318893) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/services/DownloadService.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/services/DownloadService.kt
@@ -235,23 +235,24 @@ class DownloadService : Service() {
                 val altBase = mapping.alternativeUrl
                 val primaryBase = mapping.extractedBaseUrl
 
-                val resolvedAltBase: String?
-                val resolvedPrimaryBase: String?
-                if (altBase != null && primaryBase != null) {
-                    resolvedAltBase = altBase
-                    resolvedPrimaryBase = primaryBase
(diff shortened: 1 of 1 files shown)


dogi:

Correct and CI-green. It restructures the alternative-URL resolution inline using a `when` for `resolvedAltBase`; I verified it preserves the original truth table (both paths still require `primaryBase != null`, stored-alt still `trimEnd('/')`, same `null` cases). Good work — but it targets the same block as #15368, which instead **extracts** the logic into a `resolveAlternativeUrl(...)` method, shrinking the long `initDownload`. The extraction is the better cleanup.

Both edit the same lines, so only one can merge. **Recommend closing in favor of #15368.** Marking `duplicate`.

---

PR #15383, opened 2026-08-05
teams: smoother voices payload callback diffing (fixes #15513)

### LINKED ISSUE #15513: A31. Use payloads instead of full rebinds in VoicesAdapter
`VoicesAdapter` experiences full-row rebinds because its `DiffUtils.itemCallback` lacks a `getChangePayload` implementation. Whenever `areContentsTheSame` returns false (e.g. a user image changed or an edit occurred) or an external update like reply count happens alongside an unsorted list, the whole row is rebound. Implement `getChangePayload` so isolated updates only rebind the changed parts.

dogi:
The goal of this refactor is to prevent expensive full-row rebinds in `VoicesAdapter` when frequent and isolated updates (such as reply counts or team leader status changes) occur. 

`VoicesAdapter` was experiencing full rebinds primarily because its `DiffUtils.itemCallback` lacked a `getChangePayload` implementation. As a result, anytime `areContentsTheSame` returned `false` (e.g., when a user image changed or an edit occurred) or an external update like reply count happened alongside an unsorted data push, the adapter threw away the old UI state and fully rebuilt the row.

**Key changes:**
1. **Added `getChangePayload` in `VoicesAdapter.kt`**: Safely compares properties (like `labels`, `userId`/`userName`/`avatar`/`imageUrls`, and `message`/`isEdited`/`replyTo`) and maps them to specific payloads (`PAYLOAD_TEAM_LEADER_CHANGED`, `PAYLOAD_USER_FETCHED`, `PAYLOAD_EDIT_ACTION`). It defaults to `PAYLOAD_REPLY_COUNT` if no explicitly diffable properties changed but a UI refresh is still forced, addressing the specific issue request without breaking unhandled edge cases.
2. **Updated `onBindViewHolder` in `VoicesAdapter.kt`**: Added logic to safely `flatMap` payload lists, as `ListAdapter` wraps emitted payloads into arrays. The handling for specific payloads now triggers precise UI updates (e.g., loading images, appending shared team names) rather than re-triggering the entire layout cycle.
3. **Sorted `submitList` in `VoicesFragment.kt`**: Fixed a bug where unsorted arrays were pushed to the `VoicesAdapter`, causing the `ListAdapter`'s diffing engine to flag everything as moved/changed, leading to endless re-renders.

---
*PR created automatically by Jules for task [1172010565558295917](https://jules.google.com/task/1172010565558295917) started by @dogi*

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved chat updates when message details change, including avatars, images, replies, sharing information, and edit status.
  - Refreshed user images and message controls when participant information changes.
  - Updated message text, dates, and chat state when edits are applied.
  - Improved partial updates for more responsive conversation displays.

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/ui/voices/VoicesAdapter.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/voices/VoicesAdapter.kt
@@ -102,7 +102,10 @@ class VoicesAdapter(
                 payloads.add(PAYLOAD_EDIT_ACTION)
             }
 
-            if (payloads.isNotEmpty()) payloads else PAYLOAD_REPLY_COUNT
+            // Every field checked in areContentsTheSame is covered by the buckets above.
+            // If payloads is empty here, it means a future field was added to areContentsTheSame
+            // without a corresponding bucket. We MUST return null to trigger a full rebind to prevent stale UI.
+            if (payloads.isNotEmpty()) payloads else null
         }
     )
 ) {
@@ -243,7 +246,6 @@ class VoicesAdapter(
 
 
     @SuppressLint("SetTextI18n")
-    @RequiresApi(Build.VERSION_CODES.UPSIDE_DOWN_CAKE)
     override fun onBindViewHolder(holder: RecyclerView.ViewHolder, position: Int, payloads: MutableList<Any>) {
         if (payloads.isEmpty()) {
             super.onBindViewHolder(holder, position, payloads)
@@ -288,9 +290,10 @@ class VoicesAdapter(
                         val currentLeader = getCurrentLeader(userModel, news)
                         setMemberClickListeners(holder, userModel, currentLeader)
                         loadImage(holder.binding, news)
+                        configureEditDeleteButtons(holder, news)
                     }
                     PAYLOAD_EDIT_ACTION -> {
-                        val sharedTeamName = org.ole.planet.myplanet.utils.JsonUtils.extractSharedTeamName(news)
+                        val sharedTeamName = JsonUtils.extractSharedTeamName(news)
                         setMessageAndDate(holder, news, sharedTeamName)
                         configureEditDeleteButtons(holder, news)
                         showReplyButton(holder, news, position)
--- a/app/src/main/java/org/ole/planet/myplanet/ui/voices/VoicesAdapter.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/voices/VoicesAdapter.kt
@@ -82,10 +82,27 @@ class VoicesAdapter(
                 oldItem.id == newItem.id && oldItem.time == newItem.time &&
                         oldItem.isEdited == newItem.isEdited && oldItem.message == newItem.message &&
                         oldItem.userName == newItem.userName && oldItem.userId == newItem.userId &&
-                        oldItem.sharedBy == newItem.sharedBy && oldItem.labels?.toList() == newItem.labels?.toList()
+                        oldItem.sharedBy == newItem.sharedBy && oldItem.labels?.toList() == newItem.labels?.toList() &&
+                        oldItem.avatar == newItem.avatar && oldItem.imageUrls?.toList() == newItem.imageUrls?.toList() &&
+                        oldItem.images == newItem.images && oldItem.replyTo == newItem.replyTo
             } catch (e: Exception) {
                 false
             }
+        },
+        getChangePayload = { oldItem, newItem ->
+            val payloads = mutableListOf<String>()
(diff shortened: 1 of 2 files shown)


dogi:
**Review — rating: 84/100**

Solid, real improvement and part of the ongoing adapter-diff correctness effort. Verified against source; build (default + lite) and test (default) are green.

**Strengths**
- `areContentsTheSame` was previously *under-comparing* — it ignored `avatar`, `imageUrls`, `images`, and `replyTo`, so avatar/image/reply-target changes silently failed to rebind. Adding them is a genuine correctness fix, not just cosmetic.
- The `getChangePayload` bucketing into `PAYLOAD_USER_FETCHED` / `PAYLOAD_EDIT_ACTION` / `PAYLOAD_TEAM_LEADER_CHANGED` maps cleanly onto the new fields, and the matching handlers now call `loadImage` / `setMessageAndDate` / `handleChat` so a partial bind actually repaints what changed.
- Sorting in the `else` branch of `VoicesFragment` (both branches now sort) removes the ordering churn that would otherwise defeat DiffUtil — good catch.

**Concerns (nits, non-blocking)**
- `VoicesAdapter.kt` (payload handler): `org.ole.planet.myplanet.utils.JsonUtils.extractSharedTeamName(...)` is inlined fully-qualified — add an import to match the file's style.
- `getChangePayload` falls back to `PAYLOAD_REPLY_COUNT` whenever none of the tracked buckets match. Every field in `areContentsTheSame` is currently covered by a bucket, so the fallback is effectively dead — fine today, but if a future field is added to `areContentsTheSame` without a matching bucket it'll be misrouted to a reply-count refresh. A comment noting that invariant would help.

**Verdict:** Correct, compiles, CI-green — mergeable as-is; the two items above are optional polish.

---

Okuro3499 (dismissed):
``` 
- VoicesAdapter.kt:286-291 (PAYLOAD_USER_FETCHED handler) — When this payload fires because oldItem.userId != newItem.userId (one of the conditions that produces                  
  PAYLOAD_USER_FETCHED at line 98), the handler calls configureUser/loadImage but never configureEditDeleteButtons. configureUser → showHideButtons will correctly update visibility 
  of edit/delete using the new userId-derived ownership, but the click listeners on imgEdit/imgDelete are only (re)attached by configureEditDeleteButtons, which this branch doesn't 
  call. If ownership flips (news item's userId changes) such that edit/delete becomes newly visible, the buttons will show but do nothing when tapped, because resetViews cleared    
  their listeners on the last full bind and nothing re-attaches them here.

- VoicesAdapter.kt:95-105 (getChangePayload fallback) — if (payloads.isNotEmpty()) payloads else PAYLOAD_REPLY_COUNT is unreachable: getChangePayload is only invoked when         
  areContentsTheSame returned false, and every field it checks (time, isEdited, message, userName, userId, sharedBy, labels, avatar, imageUrls, images, replyTo) is already covered  
  by one of the three payloads.add(...) conditions above it, so payloads can never be empty at that point. Either remove the dead branch or, if it's meant as a genuine safety net,  
  explain in a comment why it can occur (and if it truly can't, delete it — it currently misleads readers into thinking there's an uncovered case).   
```

dogi:
**Re-review — Rating: 90/100** (was 84)

The new commit addresses both nits from my earlier review, verified against the head diff:

- ✅ The dead `PAYLOAD_REPLY_COUNT` fallback is gone — `getChangePayload` now returns `null` when no bucket matches, with an explicit comment documenting the invariant (a future field added to `areContentsTheSame` without a matching bucket triggers a safe full rebind rather than being misrouted). This is a cleaner fix than what I suggested.
- ✅ `JsonUtils.extractSharedTeamName` is now imported rather than inlined fully-qualified.

The core change remains correct: `areContentsTheSame` now compares the previously-ignored `avatar`/`imageUrls`/`images`/`replyTo`, and every field it checks maps to exactly one payload bucket whose handler repaints the right views (`loadImage` / `setMessageAndDate` / `handleChat`). Build (default+lite) and test are green.

**Minor, non-blocking:**
- The PR description still lists "sorted `submitList` in VoicesFragment.kt" (key change #3) and a "`PAYLOAD_REPLY_COUNT` default", but the final diff touches only `VoicesAdapter.kt` and no longer defaults to reply-count. Worth trimming the description so reviewers aren't looking for changes that aren't there.
- qlty's "complex binary expression" flags are the long field-comparison chains; acceptable for this kind of diff callback, though extracting the comparisons into a helper would quiet them.

**Verdict:** Flags resolved, code correct and CI-green — mergeable.

---

Okuro3499 (approved):

PR #15391, opened 2026-08-05
dashboard: smoother sync message spacing (fixes #15122)

### LINKED ISSUE #15122: profile name text layout inconsistencies
- Profile name has extra space in between. 
- Last synced message requires a space after :
- last should be Last

Image

J-S-webskas:
fixes #15122
The trailing space was removed by the system - used unicode: :\u0020.
Also made sure the capitalization in the string resources is consistent.
Recording_20260805_133659

## Summary by CodeRabbit

* **Bug Fixes**
  * Added quotation marks around the “Last synced:” label for improved display consistency.


dogi:
**Review — 60/100**

The actual bug fix is correct: `last_synced_colon` now uses `"Last synced: "`, and ` ` is exactly the right way to keep a trailing space that the build/XML trimming would otherwise strip. Good fix for #15122.

But two things need attention before merge.

**1. Typo introduced by the capitalization sweep (must fix):**
```xml
-    <string name="start">start</string>
+    <string name="start">Atart</string>
```
`start` became **`Atart`** — the leading `S` was replaced with `A` instead of capitalized. This is a user-visible string (course/survey "Start" button), so it needs to be `Start`.

**2. Merge conflict.** GitHub reports the branch as `dirty` (conflicts with `master`). `strings.xml` is a high-traffic file and several other open PRs touch it, so this will need a rebase — and bundling ~130 unrelated capitalization edits into a trailing-space fix is what widens that conflict surface. Consider whether the capitalization pass is worth coupling to this fix or should be its own PR.

**Minor consistency note:** the sweep capitalized `english`/`spanish`/`french` (→ English/Español/Français) but left `somali` lowercase (`<string name="somali">somali</string>`). If the goal is consistent capitalization, that one was missed. (Also pre-existing, not yours: `no_enterprise` still reads "Enteprises" — missing an `r` — if you're touching these anyway.)

Please also sanity-check the few strings that get concatenated mid-sentence (e.g. `hour`/`hours` feeding into "is due in …") so the new leading caps don't read oddly inside a sentence.

**Verdict:** Core fix is right, but fix the `Atart` typo and rebase off the conflict. Changes requested.

---

dogi:
**Review — rating: 62/100**

**The actual fix for #15122 is correct** — and it's one line:

```xml
-<string name="last_synced_colon">Last synced: </string>
+<string name="last_synced_colon">Last synced: </string>
```

That's exactly right. AAPT trims trailing whitespace from string values, so the space was being eaten and the timestamp ran into the colon; ` ` preserves it. Good diagnosis of a genuinely non-obvious bug.

**The problem is everything else.** That one line ships alongside roughly 150 unrelated capitalization edits, and the sweep has real defects of its own:

- **It introduces a linguistic error.** `español → Español` and `français → Français` are wrong. Spanish and French do not capitalize their own language names — these are endonyms and lowercase is correct in both languages. Tellingly, `nepali` (नेपाली) and `arabic` (عربى) were left alone, so the rule wasn't applied consistently even within that block.
- **It's a partial sweep that leaves the file less consistent than before.** `no_finance_record` ("finance record not available") stays lowercase while its direct siblings `no_team_courses`, `no_tasks`, `no_teams`, `no_news` are all capitalized. `drag` ("drag %1$s") stays lowercase while `visibility_of` right next to it becomes "Toggle visibility of %1$s". `level` and the whole `sync_planet_*` block are untouched.
- **A large share of the churn is on accessibility content descriptions** — `chat_history`, `progress_bar`, `web_view`, `menu`, `view`, `icon`, `server_address_content_description` and friends. TalkBack pronunciation is unaffected by capitalization, so these edits carry no user-visible benefit while enlarging the review and translation surface.
- **A free fix was missed.** `no_enterprise` was edited to "Enteprises not available" — the pre-existing typo (missing `r` in "Enterprises") is preserved and now capitalized. Worth correcting since the line was being touched anyway.

**This PR has no CI runs at all** — zero check runs, not a failure. For a values-only change the risk is low, and I confirmed only string *values* change (no keys, no `translatable` attributes), so translations won't break. But nothing has verified the resource compiler accepts it. Please push a rebase to trigger build and test.

**Recommendation:** split this. Land `last_synced_colon` on its own — it closes #15122, it's one line, and it can be reviewed and merged in a minute. Then take the capitalization pass as a separate PR with an agreed rule (sentence case for user-visible labels, skip content descriptions, leave endonyms alone) applied consistently.

**Verdict:** Needs another round. Marking `change`.

---

dogi:
**Review — rating: 72/100**

The approach is right and it's the convention this repo already uses. Wrapping the value in double quotes is exactly how you preserve trailing whitespace in `strings.xml` (aapt strips the quotes and keeps the space), and I confirmed the trailing space genuinely matters — all three call sites concatenate rather than format:

- `SettingsActivity.kt:345`
- `DashboardActivity.kt:582` and `:584`

I also checked the other locales, and that's where this comes up short. Four of the five translations are **already** quoted:

```
values-es/strings.xml:780    "última sincronización: "
values-fr/strings.xml:780    "Dernière synchronisation: "
values-ne/strings.xml:780    "पिछ्लो सिंक: "
values-so/strings.xml:780    "Kor u kac: "
values-ar/strings.xml:780    آخر مزامنة:      ← still unquoted
```

So `values-ar` is the one remaining outlier, and Arabic users still hit the exact bug this PR fixes — the label renders as `آخر مزامنة:5 minutes ago`. It's the same one-line change at `values-ar/strings.xml:780`; please include it so the fix is complete across locales.

Worth considering for a follow-up (not asking for it here): a format string — `<string name="last_synced_colon">Last synced: %1$s</string>` with `getString(R.string.last_synced_colon, relativeTime)` — removes the whitespace fragility permanently and is safer for RTL than concatenating the label and the date.

**Verdict:** needs another round — add the `values-ar` quoting and this is ready.

---

dogi:

The one blocker is fixed. `values-ar/strings.xml:780` is now `"آخر مزامنة: "`, so every locale preserves the trailing space:

```
values/strings.xml       "Last synced: "                  ← this PR
values-ar/strings.xml    "آخر مزامنة: "                    ← this PR (was the gap)
values-es/strings.xml    "última sincronización: "        ← already quoted
values-fr/strings.xml    "Dernière synchronisation: "     ← already quoted
values-ne/strings.xml    "पिछ्लो सिंक: "                     ← already quoted
values-so/strings.xml    "Kor u kac: "                    ← already quoted
```

That closes the Arabic case where the label rendered as `آخر مزامنة:5 minutes ago`. The mechanism is unchanged and still right — aapt strips the wrapping quotes and keeps the whitespace — and the space genuinely matters because all three call sites concatenate rather than format (`SettingsActivity.kt:345`, `DashboardActivity.kt:582` and `:584`). Build (default + lite) and test (default) are green.

Two lines, no behavior change beyond the intended one, complete locale coverage. The `%1$s` format-string idea from the last round still stands as a worthwhile follow-up — it would make the whitespace structurally unnecessary and read better in RTL — but as I said then, that's not something to hold this PR for.

**Verdict:** mergeable.

---

Okuro3499 (approved):

dogi (approved):

PR #15416, opened 2026-08-06
all: smoother notifications group view modelling (fixes #15256)

### LINKED ISSUE #15256: Collapse notification categories with no unread notifications by default
## Current Behavior
When navigating to the Notifications screen, all notification categories (e.g. Join Requests, Team Join, Tasks, Chat, etc.) are expanded by default regardless of whether they contain unread notifications. This makes it harder for users to quickly identify and focus on categories that have unread notifications.
https://github.com/user-attachments/assets/35261727-4591-43ea-a6c4-16cc7c4b3635

## Expected Behavior
On opening the Notifications screen, only categories that contain at least one unread notification should be expanded. Categories where all notifications have been read should be collapsed by default. Users can still manually expand/collapse any category as needed.

J-S-webskas:
fixes #15256
Added expanded only when ```unreadCount > 0```

image

## Bug Fixes

* Notification groups now expand automatically only when they contain unread notifications.
* Manually expanding or collapsing a group now overrides its default expansion state.
* Grouped notifications consistently reflect the selected expanded or collapsed state.
* Marking all notifications as read now resets group expansion states, allowing groups to return to their appropriate defaults.
* Group expansion behavior is now preserved consistently while viewing and managing notifications.


Okuro3499 (approved):
Looks good
Image

dogi:
**Review — 55/100**

The one-liner does produce the headline behavior (groups with no unread start collapsed), but as written it conflates "collapsed **by default**" with "**cannot** be expanded," which introduces a UX regression.

**Concern (NotificationsViewModel.kt:214, your changed line):**
```kotlin
val isExpanded = unreadCount > 0 && type !in collapsedGroups
```
`toggleGroupExpansion(type)` only adds/removes `type` from `collapsedGroups`:
```kotlin
fun toggleGroupExpansion(type: String) {
    _collapsedGroups.update { current -> if (type in current) current - type else current + type }
}
```
So for a **fully-read group** (`unreadCount == 0`), `isExpanded` is pinned to `false` regardless of `collapsedGroups`. Two consequences:
1. **The header tap becomes a dead affordance.** Tapping a read group flips its `collapsedGroups` membership but `isExpanded` stays `false`, so nothing happens on screen — the user can no longer open a category to review its already-read notifications.
2. **Mid-read auto-collapse.** When a user reads the last unread item in a group, the next rebuild drops `unreadCount` to 0 and the group collapses out from under them.

**Suggested fix:** let the unread count drive only the *default*, while still honoring an explicit user expand. For example track manual overrides separately:
```kotlin
// expandedGroups = groups the user explicitly opened
val isExpanded = when {
    type in expandedGroups -> true
    type in collapsedGroups -> false
    else -> unreadCount > 0
}
```
and have `toggleGroupExpansion` move the type between the two override sets. That keeps read groups collapsed by default but still openable, and avoids collapsing a group the user is actively reading.

**Verdict:** Right idea, minimal diff, but it removes the ability to expand read groups and can collapse a group mid-read. Changes requested — please make the collapse a default rather than a hard rule.

---

dogi:
**Review — rating: 76/100**

The design is the right one. Replacing the single `collapsedGroups` set with a tri-state — explicitly expanded, explicitly collapsed, or no preference falling back to `unreadCount > 0` — correctly delivers "collapse when all read" as a *default* without stomping on a choice the user made by hand. I checked that `toggleGroupExpansion` and `buildGroupedList` agree on the current state before flipping it, and they do, so the first tap always produces a visible change rather than a no-op. Adding `_expandedGroups` as a fourth `combine` source is wired correctly, and the two sets stay mutually exclusive because every branch removes from one before adding to the other.

**The derivation rule is written twice, and that's a latent bug.** `toggleGroupExpansion` computes the default as `_notifications.value.any { it.type == type && !it.isRead }`, while `buildGroupedList` computes it as `unreadCount > 0` over `grouped[type]`. They're equivalent today, which is why this works — but they're two independent copies of one rule with no test pinning them together. If either drifts, the two disagree about a group's current state and the first tap on that header silently does nothing (the toggle "corrects" it to the state it was already rendering). Pull it into a single `private fun isGroupExpanded(type: String, unreadCount: Int): Boolean` used by both.

**No tests, and `NotificationsViewModelTest.kt` already exists.** This is pure ViewModel logic with no Android dependencies — three branches per path, plus the interaction between `markAllAsRead` and default collapse. The test file is right there; a handful of cases (no preference + unread → expanded, no preference + all read → collapsed, explicit expand survives mark-all-read, toggle flips in both directions) would cover it and would also lock the two derivations together.

**Nit:** the diff deletes five unrelated blank lines between `markAsRead`, `markAllAsRead`, both `List<Notification>.markAsRead` extensions and `buildGroupedList`, which packs those declarations together for no reason. Please restore them — it's unrelated churn in an otherwise focused diff. The `// It's expanded, collapse it` / `// It's collapsed, expand it` comments also just restate the line below them.

**Verdict:** Needs another round — de-duplicate the expansion rule and add ViewModel tests. The behavior itself is right. Marking `change`.

---

dogi:
**Review — rating: 74/100**

Replaces the single "collapsed" set with a tri-state — explicitly expanded / explicitly collapsed / default derived from unread count — so a group with nothing unread collapses on its own. That's the right model for #15256, and I verified the important part: the precedence chain in `toggleGroupExpansion` is byte-for-byte the same as the one in `buildGroupedList` (`expandedGroups` → `collapsedGroups` → default), so the toggle can't desync from what's rendered. The two sets are also kept mutually exclusive on every toggle. Build (default + lite) and test (default) green.

I also chased a normalization mismatch and it turns out to be a non-issue, so noting it so nobody else has to: `buildGroupedList` groups on the *normalized* key (`type.lowercase()`, unknown → `"notification"`) while `isGroupDefaultExpanded` compares the raw `it.type == type`. Those can't diverge today, because `formatNotification` stores `type = resolvedType` and `resolveType` only ever returns a lowercase `KNOWN_TYPES` member or the literal `"notification"`. Correct as written — just resting on an invariant declared 200 lines away.

Two things keep this off `ready`:

1. **No test, and the file already exists.** `app/src/test/.../NotificationsViewModelTest.kt` is right there, and `grep` for `toggleGroupExpansion|isExpanded|collaps` in it returns nothing — the new tri-state is the most branch-heavy logic in this VM and none of it is covered. It's pure, synchronous logic over a `List<Notification>`; the three cases (unread → expanded, all-read → collapsed, explicit toggle overrides both) are a handful of assertions.
2. **`markAllAsRead` doesn't reset the explicit sets.** If a user taps a group open (it lands in `_expandedGroups`) and then hits mark-all-as-read, it stays expanded — the headline behavior "collapse when all read" doesn't happen on the one action most likely to trigger it. Keeping explicit intent sticky is a defensible choice, but it should be deliberate; clearing `_expandedGroups` in `markAllAsRead` is probably what the issue wants.

**Nit:** `else -> isGroupDefaultExpanded(type, notifications)` re-scans the whole list per group, when `unreadCount` one line above is already exactly that predicate over the correctly-grouped `items` — `else -> unreadCount > 0` is cheaper, clearer, and immune to the normalization drift above. Also, the diff deletes blank lines between five unrelated functions (`markAsRead`, `markAllAsRead`, the two extensions); reverting that churn keeps the diff to the actual change.

**Verdict:** needs another round — add the tri-state test and decide the `markAllAsRead` interaction; the core logic itself is sound.

---

dogi:
**Re-review — Rating: 82/100** (was 74)

Both blockers from the last round are addressed, and the production logic is now where it should be.

**1. The tri-state test exists.** `testGroupExpansionTriStateBehavior` covers exactly the cases that were missing: default-expanded when the group has unread, default-collapsed when everything is read, explicit toggle collapsing an unread group, explicit toggle expanding a fully-read one, and toggling back. It uses `MainDispatcherRule` with a `StandardTestDispatcher` per the repo's convention, and drives `groupedItems` through a real `backgroundScope` collector rather than poking internals — that's the right way to exercise a `stateIn` flow.

**2. `markAllAsRead` now resets the explicit sets.** Clearing both `_expandedGroups` and `_collapsedGroups` alongside `_unreadCount` is what makes the headline behaviour actually fire on the one action most likely to trigger it, and the test pins it down by asserting both headers collapse afterwards. That was the substantive gap last time.

The `else -> unreadCount > 0` change in `buildGroupedList` also landed, which removes the per-group re-scan and sidesteps the normalization question I raised. Build (default + lite) and test (default) all green on `32d38f9`.

**What keeps this off `merge`: the approval is stale.** Okuro3499 approved on `0e66032` at 18:43 on 6 Aug, and three commits have landed since — `a32d8c3` (the tri-state rework itself), `05e8ae3` (test consolidation) and `32d38f9`. The approval predates the implementation it's meant to bless, so the star tag plus that approval isn't a merge gate here. A fresh look from @Okuro3499 is the only thing standing between this and `merge`; the code is ready for it.

**Please strip the debug output before it lands.** The test carries three `println("DEBUG: …")` calls plus a `// Let's print notifications too` comment. That's leftover scaffolding, and it'll print on every CI run.

**Nits in the test file** (none affect correctness):
- Fully-qualified names inline throughout — `io.mockk.mockk`, `kotlinx.coroutines.test.runTest`, `org.ole.planet.myplanet.model.NotificationPayload`, `org.ole.planet.myplanet.utils.MainDispatcherRule` and so on — while the same file imports other symbols normally at the top. Importing these would cut a lot of noise and match the style guide.
- On "pulling the test logic into one function" (`05e8ae3`): the intent was to share the *setup*, not to merge the assertions. Right now one test asserts three independent behaviours, so a regression tells you `testGroupExpansionTriStateBehavior` failed without saying which. A private helper (or `@Before`) building the mocked repository plus three focused `@Test`s gets the deduplication without losing that signal.
- `assertTrue(!x)` appears four times where `assertFalse(x)` reads better and produces a clearer failure message.
- `find { … } as NotificationListItem.Header` will throw an opaque cast/NPE if the header is absent rather than failing with a useful message.
- The newly added imports are appended rather than sorted in, and the test mixes `testDispatcher.scheduler.advanceUntilIdle()` with `runCurrent()`.
- One stray blank line is still deleted between the two `List<Notification>.markAsRead` extensions — the rest of that churn was reverted, thanks.

**Heads-up for maintainers:** #15531, "Resolve tri-state notifications VM group expansion logic and add unit tests", is by the same author against the same ViewModel and describes the same tri-state-plus-tests work. It's `on hold` so it isn't in the review queue and I've left its labels alone, but if this PR lands, that one is likely redundant — worth deciding rather than letting both sit.

**Verdict:** mergeable once the three `println`s come out — the logic and test coverage are done. Moving to `ready` rather than `merge` because Okuro3499's approval predates the tri-state rework and needs refreshing.

---

dogi (approved):

PR #15417, opened 2026-08-06
dashboard: smoother guest offline visiting (fixes #15213)

### LINKED ISSUE #15213: guest user sees become a member dialogue when the user has 1 log in remaining
If the log in count started from 0, it becomes 1 without the user logging out and logging back in, which would turn the issue into: guest user logs in once, but the 0th log in counts for the 1st log in as well. 
- guest user should be free to explore until they log in for the 4th time.

https://github.com/user-attachments/assets/b48f0c88-9195-4138-95aa-6030421b1f96

J-S-webskas:
Changed form 3 to 4:
```offlineVisits >= 4 -> showGuestDialog()```

https://github.com/user-attachments/assets/e429cef0-1d3d-4157-8c0d-80f5df97f34e

## Summary by CodeRabbit

* **New Features**
  * Improved offline visit guidance for guests:
    * Warning banner appears after three visits.
    * Existing warning remains at two visits.
    * Guest dialog appears after four or more visits.
  * Added a dismissible banner with last-login information.

* **Localization**
  * Added last-login messages in Arabic, Spanish, French, Nepali, Somali, and English.

* **Bug Fixes**
  * Improved warning banner icon compatibility with Material components.

### CODE REVIEW

**Review by Okuro3499 — DISMISSED** [review]
the last/3rd login should also have a dialog
Image

**Review by Okuro3499 — APPROVED** [review]

**Comment by dogi** [comment]
**Review — rating: 65/100**

The core fix is right. Shifting the blocking dialog from `>= 3` to `>= 4` and slotting a "last login" banner in at exactly 3 does resolve #15213 — the member-signup dialog no longer fires while the guest still has a login left — and the 2/3/4 ladder now reads clearly. Adding the string to all five translation files rather than English-only is the right instinct too.

**But the Arabic translation is Nepali.** `values-ar/strings.xml` gets:

```xml
<string name="last_login_message">सदस्य बन्नु अघि यो तपाईंको अन्तिम लगइन हो।</string>
```

That's Devanagari, character-for-character identical to the `values-ne` entry added in the same diff. Arabic users would see Nepali text in a right-to-left layout. Straightforward copy-paste slip, one line to fix, but it has to be fixed before this lands.

**`showLastLogInDialog` is a near-verbatim copy of the existing `showVisitLimitWarning`** — same `removeAllViews()`, same `inflate(R.layout.banner_offline_visit_warning, binding.bannerContainer, true)`, same close-button wiring, even the same three comments. The only real difference is the one line setting `banner_message.text`. Collapse both into a single `private fun showBanner(@StringRes message: Int)`; that also lets `showVisitLimitWarning` set its message explicitly instead of depending on whatever the layout hardcodes.

**Smaller things:**
- `var banner_message` is snake_case and should be `val bannerMessage` — `docs/CODE_STYLE_GUIDE.md` calls for camelCase, and it's never reassigned.
- The method is named `...Dialog` but inflates an inline banner, which matters because the neighbouring `showGuestDialog` really is an `AlertDialog`. `showLastLoginBanner` would say what it does.
- `inflate(..., attachToRoot = true)` returns the *container*, not the banner, so `bannerView` is actually `binding.bannerContainer`. The `findViewById` calls still resolve because the banner is now a child — it works, but the name misleads. (Pre-existing in `showVisitLimitWarning`; folding the two together is the moment to fix it.)
- The English string is the only one of the five without a closing period.

I checked the `android:tint` → `app:tint` swap since there's no CI to catch a namespace error: `xmlns:app` is declared on the `MaterialCardView` root, so both attributes resolve. That change is a genuine small improvement for AppCompat tinting.

**This PR has no CI runs at all** — zero check runs, not a failure. Given it adds a layout usage and touches six resource files, please push a rebase so build (default + lite) and test actually run before merge.

**Verdict:** Needs another round — fix the Arabic string and de-duplicate the banner helper. Marking `change`.

---

dogi:
**Re-review — Rating: 87/100** (was 65)

Every blocking finding from the last round is addressed:

- **Arabic string is now actually Arabic** — `values-ar/strings.xml` reads `هذا هو تسجيل دخولك الأخير قبل أن تصبح عضواً`, not the Devanagari that was copy-pasted from `values-ne`. That was the hard blocker.
- **The duplicated banner helper is gone.** `showVisitLimitWarning` and `showLastLogInDialog` are collapsed into a single `showBanner(@StringRes messageRes: Int)` with the `androidx.annotation.StringRes` import — exactly the consolidation asked for, and the visit-2 path now passes `R.string.guest_visit_limit_warning` explicitly instead of relying on the layout's hardcoded `android:text`. I confirmed that string is the same one the layout hardcoded, so visit-2 behavior is unchanged.
- **Naming fixed** — `val messageView`, camelCase, not reassigned. And `showBanner` no longer claims to be a dialog while inflating a banner.
- **CI now runs and is green** — build (default), build (lite), and test (default) all pass, where the previous round had zero check runs.

On the fix itself: the 2 → 3 → 4 ladder is coherent against the actual string content. Visit 2 says "You have 1 login remaining", visit 3 says "This is your last login", and the blocking dialog fires at `>= 4` — so "remaining" consistently excludes the current visit and the guest genuinely gets the login the old `>= 3` was stealing. I also grepped for other `getOfflineVisitCount` thresholds; nothing else gates guest access on a visit count, so this `when` is the only place the ladder lives.

**Nit:** two leftovers from last round, both cosmetic. `inflate(..., container, true)` still returns the *container*, so `inflatedView` is really `binding.bannerContainer` — it works because `findViewById` reaches the attached children, but the name misleads. And `ar`/`so` still lack the trailing period that `en`/`es`/`fr` have.

**Verdict:** mergeable — Okuro3499 has approved and the PR carries `⭐⭐⭐.`, so this goes to `merge`. Noting for the record that the approval predates the 2026-08-10 fix commits; this re-review covers the current head.

---

dogi (approved):

PR #15437, opened 2026-08-07
sync: smoother download dialog handling (fixes #15435)

### LINKED ISSUE #15435: Do not show this dialog in Courses and My Course
## Summary
A dialog is being shown in **Courses** and **My Course** where it should not appear.

## Problem
The attached screenshot/video shows an unexpected dialog appearing in these sections:
- Courses
- My Course
https://github.com/user-attachments/assets/e833aed6-f157-48de-8623-cbc989a39059

ragilzakaria:
- Made showDownloadDialog(dbMyLibrary: List<MyLibrary?>) protected open in BaseResourceFragment.kt
- Overrode showDownloadDialog in CoursesFragment.kt and TeamCoursesFragment.kt as a no-op method:
`
override fun showDownloadDialog(dbMyLibrary: List<MyLibrary?>) {
// Do not show download suggestion dialog in Courses and My Course (Fix #15435)
}
`

fixes #15435 

## Summary by CodeRabbit

* **Bug Fixes**
  * Course browsing no longer displays download suggestion dialogs.
  * Team course browsing also suppresses download suggestion dialogs.
  * Other resource sections retain their existing download prompt behavior.


dogi:
**Review — rating: 80/100**

Minimal and it does the job. `CoursesFragment` backs both the Courses and myCourses views (switching on `isMyCourseLib`), so the single override there covers both halves of #15435 rather than just one. Extending the same suppression to `TeamCoursesFragment` goes slightly beyond the filed issue but is the right call — the identical complaint applies there. CI green on build (default + lite) and test (default), which confirms the override chain resolves through `BaseRecyclerFragment` → `BaseResourceFragment`.

**Nit — the mechanism is a shade blunter than it needs to be.** Making `showDownloadDialog` `open` so two subclasses can override it into a no-op permanently widens the base class's extension surface, and it means a future reader at any call site sees a method that may or may not do anything with nothing at the call site to hint at it. A declarative opt-out keeps the behavior in one place:

```kotlin
// BaseResourceFragment
protected open val showsDownloadSuggestions: Boolean get() = true

protected fun showDownloadDialog(dbMyLibrary: List<MyLibrary?>) {
    if (!showsDownloadSuggestions) return
    ...
}
```

Each opting-out fragment then carries a single self-documenting line (`override val showsDownloadSuggestions = false`), `showDownloadDialog` stays non-`open` with exactly one implementation, and the two identical empty-method-plus-comment blocks collapse. Same behavior, less surface, and greppable.

Not worth blocking on — the current form is correct and contained.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

dogi (approved):

PR #15412, opened 2026-08-06
courses limited space for content in landscape mode (fixes #15264)

ragilzakaria:
https://github.com/user-attachments/assets/3fe73d59-c7a0-485b-8a1a-7957d7625213

fixes #15264

## Summary by CodeRabbit

- **New Features**
  - Added an optimized landscape layout for browsing and managing courses.
  - Redesigned the course toolbar with search, filtering, selection, and course actions.
  - Improved course selection, removal, and archiving workflows.
  - Added landscape-friendly search and filtering controls.

- **Bug Fixes**
  - Course lists now update immediately after courses are removed.
  - Improved handling of confirmation dialogs and device orientation changes.
  - Fixed potential issues opening resource filters when optional controls are unavailable.

- **Usability Improvements**
  - Success dialogs are more readable on landscape screens and support scrolling for longer lists.

PR #15266, opened 2026-08-04
prevent team calendar cropping in landscape mode by using NestedScrollView (fixes #15265)

ragilzakaria:
https://github.com/user-attachments/assets/79e1bf5b-1e45-4d62-8c81-df2f15f878fe

fixes #15265

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved scrolling behavior in calendar views with enhanced viewport handling
  * Calendar content now displays with optimized width on all screen sizes

PR #15441, opened 2026-08-07
resources: smoother list  filter icon spacing (fixes #15438)

### LINKED ISSUE #15438: inconsistent filter icon/container size in Library and Courses
https://github.com/user-attachments/assets/e0b7fd59-e6e7-4096-87a5-78b644e0818d

the size of the filter box/icon/container size in library and courses is inconsistent

ragilzakaria:
- Filter Icon Button (@id/filter) Standardized
- Top Blue Container (@id/ll_actions) & Header Standardized
fixes #15438 

## Summary by CodeRabbit

* **Style**
  * Improved course action-area spacing and title sizing for a more compact, readable layout.
  * Refined library headers and filter controls with updated sizing, alignment, spacing, and theme-based colors.
  * Improved landscape course and library layouts for better visual consistency and readability.


dogi:
**Review — rating: 74/100**

Two of these changes are genuinely good. Switching `tv_fragment_info` from `wrap_content` to `layout_width="0dp"` alongside the existing `layout_weight="1"` is the correct idiom for a weighted `LinearLayout` child — `wrap_content` with a weight forces a wasted extra measure pass. And `android:layout_gravity="right"` / `android:gravity="right"` → `end` on `ll_filter` is a real RTL fix that matters, since the app ships Arabic. Bringing the library filter button to 40dp to match Courses is the actual fix for #15438.

**The fixed height on the title is a regression risk.** `tv_fragment_info` goes to `android:layout_height="40dp"` while keeping `android:padding="@dimen/padding_normal"` and `android:textSize="@dimen/text_size_large"`. From `dimens.xml` those are 8dp and 18sp, so the content box is 40 − 16 = **24dp** for 18sp bold text. That fits at font scale 1.0 with almost nothing to spare — a single line of 18sp renders around 21–22dp. Above roughly 1.1× it clips, and this app doesn't just inherit the system font scale: it ships its own text-size setting (`text_size_small` / `medium` / `large`, `select_text_size`). Adding `android:maxLines="1"` removes wrapping as an escape valve, so the text has nowhere to go. `wrap_content` with `android:minHeight="40dp"` gets the same alignment without the ceiling. Worth taking seriously given #15266 and #15412 in this queue are both cropping bugs.

**`@color/md_white_1000` hardcodes what should be a theme attribute.** On the `colorPrimary` header, white is very likely more correct than the previous `textColorPrimary` — so this probably fixes a contrast bug. But the app has a full dark/light theme system (`ThemeManager`, `dark_mode_follow_system`) and is on Material 3, so `?attr/colorOnPrimary` expresses the same intent and keeps following the theme.

**Nit:** the `android:textColor` change on the `filter` **`ImageButton`** does nothing — an `ImageButton` renders a drawable, so both that attribute and the pre-existing `android:text="@string/more_action"` on it are inert. Worth deleting rather than editing. (`contentDescription` is the one that matters, and it's correctly present.)

**On the collision flag with #15446:** I compared the diffs and **these are not duplicates.** Both touch `layout/fragment_my_library.xml`, `layout-land/fragment_my_library.xml` and both `fragment_my_course.xml` variants, but in different regions with different intent — #15446 adds a `view_grid_list_toggle` include, swaps `layout_search` for `layout_search_pill`, and wraps the RecyclerView in `MaxWidthFrameLayout`, while leaving `tv_fragment_info`'s sizing, the `filter` button, and `ll_filter` exactly as they are. Both should land. They will conflict textually, though — #15446 inserts immediately after the `android:minWidth="150dp" />` line that closes the `tv_fragment_info` block you're editing — so whichever merges second needs a rebase.

**Verdict:** Needs another round — drop the fixed height (or pair it with `minHeight`) and prefer `?attr/colorOnPrimary`. The weight and RTL fixes are keepers. Marking `change`.

---

dogi:
**Review — rating: 62/100**

The actual fix for #15438 is correct and I verified it line by line. The library filter `ImageButton` now reads `_40dp`/`_40dp`, `layout_margin="@dimen/padding_small"`, `padding="@dimen/padding_small"`, `layout_gravity="center_vertical"` — byte-for-byte the same as the courses one at `fragment_my_course.xml:106-111`, in both orientations. The header container padding change is the same story: library was already `padding_small`, so dropping courses from `padding_large` matches them up. Removing `android:text`/`android:textColor` from an `ImageButton` is correct dead-attribute cleanup (it extends `ImageView` and renders no text), and `contentDescription` is retained.

**The problem is the bundled `?attr/colorOnPrimary` switch, and CI cannot catch it.** `colorOnPrimary` is a MaterialComponents theme attribute. It is not defined anywhere in this project's resources — `grep -rl colorOnPrimary app/src/main/res/` returns nothing — and the theme chain for the screens in question never introduces it:

- `DashboardActivity` (which hosts both fragments) uses `@style/FullscreenTheme`
- → parent `MaterialDrawerTheme.Light.DarkToolbar.TranslucentStatus`
- → MaterialDrawer **6.1.1**, which is AppCompat-era, not MaterialComponents
- and `Theme.AppCompat.*` does not define `colorOnPrimary`

The one theme in the project that *would* supply it — `AppTheme.MaterialComponents`, parent `Theme.MaterialComponents.DayNight.NoActionBar.Bridge` at `styles.xml:210` — is not referenced by any activity in the manifest.

With the attribute unresolved, `getColorStateList` returns null and the `textColor` simply isn't applied, so `tv_fragment_info` falls back to the default dark text of a Light theme — on top of `android:background="@color/colorPrimary"` (`md_blue_700`). Both screens previously rendered white text (`md_white_1000` for courses; `textColorPrimary` is `#ffffff` in `values/colors.xml`, so library was white too), which means this unifies the two screens onto a token that likely renders *both* illegible. The build compiles because the material dependency declares the `attr` symbol, so green CI proves nothing here — and there's no instrumented suite to catch it.

I couldn't crack open the MaterialDrawer 6.1.1 AAR in this environment to see its themes directly, so please confirm on a device. If you want the two screens to share one token, `@color/md_white_1000` in both is the safe fix; `?attr/colorOnPrimary` needs the host theme migrated to a Material/Bridge theme first, which is a much bigger change than this PR.

**Also incomplete:** the `right` → `end` RTL fix was applied to `layout/fragment_my_library.xml` but not to its landscape twin — `layout-land/fragment_my_library.xml:186,188` still carry `layout_gravity="right"` and `gravity="right"` on the same `ll_filter` block. Two lines, same change.

**Nit:** dropping `maxLines="1"` and moving to `wrap_content` + `minHeight="40dp"` is a real improvement for large font scales, but it's undeclared scope for a PR about icon container size — worth calling out in the description so it gets looked at deliberately.

**Verdict:** needs another round — the icon sizing is right, but revert or re-target `?attr/colorOnPrimary` and finish the landscape RTL fix.

---

dogi:
**Review — rating: 82/100**

Layout-only change across the portrait and landscape variants of `fragment_my_course.xml` and `fragment_my_library.xml`. I verified the central claim rather than trusting the title: on master the courses filter `ImageButton` is already `@dimen/_40dp` with `padding_small`, while library's was `@dimen/_50dp` with `padding_normal`, so bringing library down to 40dp/`padding_small` genuinely converges the two — the inconsistency was real and this is the correct direction. Dropping `android:text`/`android:textColor` from the `ImageButton`s is right too: `ImageButton` extends `ImageView`, so those attributes never rendered, and `contentDescription` is correctly retained for accessibility. The `layout_gravity`/`gravity` `right` → `end` swaps are a real RTL fix that matters given the Arabic locale, and switching `tv_fragment_info` to `0dp` + `layout_weight="1"` is the correct idiom for a weighted `LinearLayout` child. Replacing the fixed `40dp` height with `wrap_content` + `minHeight="40dp"` and removing `maxLines="1"` lets the header grow at large font sizes instead of clipping.

One change is a step backwards. `tv_fragment_info`'s color goes from `@color/textColorPrimary` to `@color/md_white_1000`. That header sits on a `@color/colorPrimary` background, and `textColorPrimary` is already `#ffffff` in `values/` and a deliberately softened `#e4e6eb` in `values-night/`. So in light mode this is a no-op, and in dark mode it swaps a night-aware resource for hardcoded pure white. It matches what courses does today, but consistency is being achieved by copying the less theme-aware side; changing courses to `textColorPrimary` would have been the better direction. Not a blocker — legible either way.

⚠️ **Conflict:** #15446 rewrites these same four layout files and replaces this header block wholesale. Whichever lands second needs a rebase; if #15446 lands first this PR becomes moot. Worth sequencing deliberately rather than letting the merge order decide.

**Nit:** scope runs past the title — header padding, text heights, `maxLines` and RTL gravity are all in here alongside the filter-icon sizing. All defensible as "make these two headers consistent", but the title undersells it.

---

Okuro3499 (approved):

dogi:
**Promoted to `merge`** — Okuro3499 approved, my review stands at 82/100, no changes since.

---

dogi (approved):

PR #15442, opened 2026-08-07
courses: smoother completion rating (fixes #15439)

### LINKED ISSUE #15439: Show course rating dialog after learner completes a course
## Summary
Show a course rating dialog after a learner completes a course so they can provide feedback immediately.

## Problem
Learners currently finish a course without a clear, timely prompt to rate it. This can reduce the amount of rating feedback collected and make it harder to understand course quality from the learner perspective.

## Proposed solution
Display a rating dialog when course completion is detected.

The dialog should:
- Appear after the learner completes a course
- Let the learner submit a course rating with minimal friction
- Be clearly associated with the completed course
- Avoid appearing before completion or interrupting normal course progress

ragilzakaria:
fixes #15439 

https://github.com/user-attachments/assets/cc3b5f02-f53f-4ce3-be76-396134666b0b

## Summary by CodeRabbit

* **New Features**
  * Added a post-course rating step when required surveys remain incomplete.
  * Course completion now returns users to the previous screen after ratings are submitted or dismissed.
  * Added safer handling when closing the rating dialog.

* **Style**
  * Updated the rating action area with clearer button alignment, spacing, sizing, and styling.
  * Added a dedicated cancel button and improved submit button presentation.


dogi:
**Review — rating: 72/100**

The feature works and some of this is a clear improvement. Routing `finish_step` through `onFinishStep()` that *acts* immediately is better than the old `checkSurveyCompletion()`, which responded to a click by re-registering a click listener rather than doing anything — effectively requiring a second tap. The layout changes are good too: `layout_gravity="right"` → `gravity="end"` and `layout_marginRight` → `layout_marginEnd` are real RTL fixes for an app that ships Arabic, and promoting Submit to `YellowButtons` in the trailing position with Cancel demoted to grey is the conventional ordering.

To be clear about one thing I checked before flagging it: the magic `"4e6b78800b6ad18b4e8b0e1e38a98cac"` is **not** introduced here — it's already on `master` in `checkSurveyCompletion`, `TakeCourseFragment:288`, `MarkdownDialogFragment` and `DashboardViewModel`. Not this PR's problem to solve.

**But this PR now encodes that gate in a second place.** `onFinishStep()` and `checkSurveyCompletion()` both compute `hasUnfinishedSurveys(...)` and both test `hasUnfinishedSurvey && courseId == "4e6b…"`, and each takes its own action on the result. Two copies of one rule that must stay in sync, in the same file, is exactly how the two paths drift — and the ID is the sort of constant that eventually needs to become a named config value in one place.

**More concerning: the `when` branch may be unreachable.** `checkSurveyCompletion()` still runs at init (line 144) and unconditionally calls `binding.finishStep.setOnClickListener { … }` in both of its branches. That explicit lambda replaces whatever listener the fragment registered for itself, so the `R.id.finish_step -> onFinishStep()` arm of the `onClick` `when` only fires if the fragment's own listener is what's actually attached at tap time. Depending on ordering, `onFinishStep()` may never execute and every finish would go through the listener that `checkSurveyCompletion` installed. Worth confirming on-device which path actually runs — and better, collapsing to a single mechanism: keep the fragment-level `onClick` and drop the `setOnClickListener` calls from `checkSurveyCompletion`, or vice versa, but not both on one view.

**Nit:** `setOnDismissListener` stores a lambda on the `DialogFragment` instance, so it's lost across a configuration change — rotate while the rating dialog is open and dismissing it won't `popBackStack`, stranding the user on the finished course. `setFragmentResult`/`setFragmentResultListener` survives recreation and is the more robust wiring for "dialog closed → navigate".

**Verdict:** Needs another round — resolve the duplicated survey gate and the two competing click listeners. Marking `change`.

---

dogi:
**Review — rating: 71/100**

The feature works and the refactor is a genuine improvement. Collapsing `checkSurveyCompletion()` into a single `onFinishStep()` dispatched from `onClick` removes the old pattern where the finish button's listener got *rebound* on every load — and I checked the two things that removal could have broken: `binding.finishStep.setOnClickListener(this)` at `TakeCourseFragment.kt:176` is independent of the deleted method, so the dispatch path survives, and `fragment_take_course.xml` never sets `android:enabled`, so dropping the `isEnabled = true` line is a no-op rather than a button that can never be pressed. The layout work is good too: `layout_marginRight`/`layout_gravity="right"` → `marginEnd`/`gravity="end"` is a real RTL fix, and promoting submit to `YellowButtons` makes the primary action obvious.

**The main problem is the fragment-result channel.** `RatingsFragment.onDismiss` now unconditionally calls `setFragmentResult(REQUEST_KEY, Bundle())` on its `parentFragmentManager` — and `RatingsFragment` is a *shared* dialog, not a course-completion one. It's launched app-wide through `DashboardElementActivity.showRatingDialog` (`:164`, `f.show(supportFragmentManager, "")`) from at least three places:

- `CoursesAdapter.kt:449` — rating stars in the course list
- `ResourcesAdapter.kt:350` — rating stars in the resource list
- `BaseContainerFragment.kt:145`

That's the same `supportFragmentManager` `TakeCourseFragment` registers its listener on, and the listener's action is destructive — `popBackStack()`. `setFragmentResultListener` also *queues* a result and delivers it when the listener's lifecycle next reaches STARTED, so this doesn't require the dialog and the course to be on screen together: any dismissal that lands while `TakeCourseFragment` is stopped-but-still-on-the-back-stack will pop the course out from under the user when they navigate back to it. I haven't traced a guaranteed end-to-end repro (it needs a fragment pushed *on top* of `TakeCourseFragment`, and the in-course step navigation uses a ViewPager rather than back-stack entries), so treat this as construction-fragility rather than a confirmed crash — but a global key driving `popBackStack` from a dialog reused in three unrelated places is the wrong shape regardless.

The fix is already half-written in this PR: **`setOnDismissListener` is added to `RatingsFragment` and has zero callers.** `grep -rn setOnDismissListener` across `app/src/main/java` returns only `Dialog.setOnDismissListener` matches on AlertDialogs — nothing uses the new one. Pass the callback at the launch site in `showCourseRatingDialogAndFinish()` and the whole fragment-result channel, the global key, and the cross-screen coupling all go away. Otherwise, namespace the key per launch site so only the course-completion dialog emits it.

**Nit:** the `try { setFragmentResult(...) } catch (e: Exception) { e.printStackTrace() }` swallows the one failure that matters — if `parentFragmentManager` isn't attached, the result never fires and the user is silently stranded on the finished course. Using the dismiss callback sidesteps this entirely.

Also worth noting the magic course id `"4e6b78800b6ad18b4e8b0e1e38a98cac"` carries over from the old method — pre-existing, not introduced here, but it now sits in new code and is worth a named constant or a comment explaining which course it is.

**Verdict:** needs another round — swap the fragment-result broadcast for the `setOnDismissListener` this PR already adds; the finish-flow refactor and layout changes are solid.

---

dogi:
**Review — rating: 85/100**

Shows the ratings dialog when a course is finished, then pops back on dismiss. I checked the two ways this refactor could have broken the finish flow, and both are clean.

The removed line `binding.finishStep.isEnabled = true` was the risk — if `finish_step` were declared disabled in XML, dropping that would strand users on the last step with a dead button. It isn't: `fragment_take_course.xml` gives `finish_step` only `visibility="gone"`, no `android:enabled="false"`, and nothing in the fragment ever disables it — the 14 other references only toggle visibility. So the button is enabled by default and removing that line is a no-op.

The more interesting part is that this fixes a latent race rather than just adding a dialog. On master, `checkSurveyCompletion()` ran from `onViewCreated` and *replaced* `finishStep`'s click listener from inside a coroutine, while line 176 had already set `setOnClickListener(this)`. Whichever finished last won, and because the coroutine resolved after `onViewCreated`, the `R.id.finish_step -> checkSurveyCompletion()` branch in `onClick` was effectively unreachable outside a narrow race window. Routing the tap through `onClick` → `onFinishStep()` and leaving the listener alone removes the swap entirely and makes each tap re-evaluate the survey gate freshly. That's the right structural fix. The gate condition itself is carried over unchanged — `hasUnfinishedSurvey && courseId == MANDATORY_SURVEY_COURSE_ID` — and hoisting the bare `"4e6b78800b6ad18b4e8b0e1e38a98cac"` into a named, commented constant is a real readability win.

In `fragment_rating.xml`, `layout_gravity="right"` → `gravity="end|center_vertical"` and `layout_marginRight` → `layout_marginEnd` are genuine RTL fixes that matter for the Arabic locale.

**Things worth a second look:**
- `onDismiss` fires for *every* dismissal, cancel included, so tapping **Cancel** on the rating dialog also pops you out of the course. That's a coherent reading of "rating is optional", but a user pressing Cancel may well expect to stay put. Worth confirming it's the intent.
- The dismiss callback is a plain lambda, so it doesn't survive a configuration change. Rotate with the dialog open, then dismiss, and the `popBackStack` never runs — the user is left on the finished course screen.
- Two rapid taps on finish launch two coroutines and can `show()` two dialogs; `DialogFragment.show()` doesn't dedupe on tag. A `parentFragmentManager.findFragmentByTag(TAG) == null` guard would close that.
- The dialog is shown on `parentFragmentManager` but the pop targets `requireActivity().supportFragmentManager`. That works while this fragment is hosted directly by the activity, but mixing the two is fragile if it's ever nested.
- Swapping the Submit/Cancel order and restyling Submit to `YellowButtons` changes the ratings dialog *everywhere* it's used, not just at course completion. Good change, just broader than the title suggests.
- No test covers the survey-gate branch. It's a two-condition truth table now sitting in a fragment; extracting it would make it testable, in line with the repo's ViewModel-level test coverage.

Heads-up on batch interaction: #15446 removes the rating bars and `setRatingChangeListener` from the course cards, so these two PRs are moving the rating entry point in opposite directions. Worth deciding jointly where users are meant to rate a course.

**Verdict:** mergeable — CI green (build default + lite, test default); the edge cases above are follow-up material, not blockers.

---

Okuro3499 (dismissed):
the dialog should not show up for courses I have already rated
https://github.com/user-attachments/assets/6a678e49-9e7b-4961-a350-f711a9c8d506

ragilzakaria:
Updated

In TakeCourseFragment.kt:388-410:
- Injected RatingsRepository.kt.
- In showCourseRatingDialogAndFinish(), checked whether the user has already rated the course (ratingsRepository.getRatingSummary("course", cId, userId)).
- If the course has already been rated by the active user, the rating dialog is skipped and the screen pops back immediately.
- If the course has not yet been rated, the rating dialog is displayed as expected.

dogi:
**Review — rating: 84/100**

Adds a rating prompt on course completion and, in the process, moves the mandatory-survey gate from view-setup time to click time. I checked the two things that deletion of `checkSurveyCompletion()` put at risk. First, that method also owned `finishStep.isEnabled = true` and the click listener — but `binding.finishStep.setOnClickListener(this)` already exists at `TakeCourseFragment.kt:176` and `finish_step` in `fragment_take_course.xml` declares no `android:enabled="false"`, so the button defaults to enabled and there is no dead-button regression. Second, the `hasRated` predicate: `RatingsRepositoryImpl.getRatingSummary` computes `existingRating` as `results.firstOrNull { it.userId == userId }`, i.e. scoped to the current user rather than the community aggregate, so the dialog is correctly suppressed only for users who already rated — not for everyone once one person has rated. Evaluating the survey gate per click is also a genuine improvement over the old evaluate-once-at-setup behaviour. Build (default + lite) and test (default) green.

Three things worth noting, none blocking:

- `summary.userRating != null || summary.existingRating != null` is redundant — `userRating` is defined as `existingRating?.rate`, so the two are non-null together. The first clause alone says it.
- `dismissListener` is a plain field on a `DialogFragment`. After a rotation the fragment is recreated by the `FragmentManager` with the listener null, so dismissing the rating dialog won't pop the course fragment and the user is left on the finished-course screen. Surviving state here would need the callback re-attached in `onCreate`/`onStart` or the pop driven from the parent instead.
- `show(parentFragmentManager, TAG)` isn't guarded by a `findFragmentByTag` check, and the click now goes through a suspend repository call first, so a fast double-tap on Finish can stack two dialogs.

**Nit:** the `catch (e: Exception)` binds `e` without using it.

**Verdict:** mergeable — the listed items are edge cases, not main-flow defects.

---

Okuro3499 (approved):

dogi:
**Promoted to `merge`** — Okuro3499 approved, my review stands at 84/100, no changes since.

---

dogi (approved):

PR #15443, opened 2026-08-07
life: smoother options visibility updating (fixes #15236)

### LINKED ISSUE #15236: Mylife hide state not updating immediately
When hiding items in Mylife, the greyed-out visual state does not appear immediately.

Actual behavior
The greyed-out effect is only visible after leaving the page and re-entering.
https://github.com/user-attachments/assets/1d83af7c-1be0-4271-a65a-4bf494059c0e

Expected behavior
Hidden items should appear greyed out immediately after being hidden, without requiring navigation away and back.

Okuro3499:
fixes #15236

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved My Life item visibility updates with immediate visual feedback.
  * Ensured visibility changes remain synchronized when returning to the dashboard.
  * Prevented stale or duplicated dashboard items during refreshes.
  * Improved consistency of cached My Life data after visibility changes.


J-S-webskas (commented):
https://github.com/user-attachments/assets/7a56646a-0189-41e7-9ba3-178bd9ab78d0

Right now there's duplicate of each mylife's attributes, not sure if that's intended. But the deletion and refresh is working properly.

dogi:
**Review — rating: 80/100**

Four separate real fixes here, and each one addresses a distinct root cause rather than a symptom:

- **`flexboxLayout.removeAllViews()`** at the top of `myLifeListInit` — the actual duplication bug. Re-entering the dashboard was appending a second set of tiles.
- **`MyLife.equals`/`hashCode`** — this is the right fix, not a workaround. `LifeAdapter`'s `areContentsTheSame` is `oldItem == newItem`, and `MyLife` is a plain class, so that comparison was identity-based and every submit rebound every row. Giving the model real structural equality fixes it at the source and keeps working as fields are added, which is strictly better than hand-maintaining a field list in the adapter (compare #15284, which takes the other approach and drifts for exactly this reason). `hashCode` is consistent with `equals` over the same six fields.
- **`LifeRepositoryImpl.updateVisibility`** now refreshes the cache after the DAO write, so the cached list can't serve a stale visibility flag.
- **The `RecyclerView.NO_POSITION` guard** in `updateVisibility` is a crash fix: the old code fed `bindingAdapterPosition` straight into `getItem(position)`, which throws if the holder is detached mid-interaction. Reading `myLife` fresh from `getItem` instead of trusting the value captured in the bind closure is also more correct.

**One caveat worth knowing about rather than fixing:** `updateVisibility` mutates `myLife.isVisible` in place on an object owned by `AsyncListDiffer`'s current list. Combined with the new `equals`, that means when the repository later emits the updated list, DiffUtil sees old and new as equal and dispatches nothing. That's harmless *because* the row is updated optimistically via `changeVisibility` in the same call — the two changes cancel out to the correct result. But it does mean the UI is never reconciled from the data layer, so if the persist fails the toggle stays visually flipped. Fine for a visibility toggle; just don't extend the pattern to anything where a failed write matters.

**Smaller things:**
- No tests. `MyLife.equals`/`hashCode` is pure, dependency-free logic and `LifeRepositoryImpl` already has repository-test precedent — both are cheap to cover, and equality contracts are exactly the thing that silently regresses later.
- `refreshMyLifeList()` launches on every `onResume` without cancelling a prior run. Two rapid resumes can interleave `removeAllViews()` and the add loop and briefly double-render. Narrow, but it's the same class of problem #15306 is addressing elsewhere.
- Dropping the explicit `mapOf<String, () -> Fragment>` type argument leaves the value type to inference. It compiles, but the annotation was documenting intent on a map of fragment factories — I'd keep it.

**Heads-up:** #15318 also edits `LifeAdapter` (removing `workingList`, reworking drag). Different regions, not duplicate work, but one of the two will need a rebase.

**Verdict:** Mergeable.

---

dogi (approved):

PR #15446, opened 2026-08-07
all: smoother resources courses grid list viewing (fixes #15440)

### LINKED ISSUE #15440: Courses and Resources UI is not responsive in landscape mode
## Description
The Courses and Resources UI becomes unresponsive/squeezed when the device is switched to landscape orientation. The layout does not adapt properly to the wider screen dimensions, resulting in a cramped and unusable interface.

## Steps to Reproduce
1. Open the app and navigate to the **Courses** or **Resources** section.
2. Rotate the device to **landscape** orientation.
3. Observe the UI — elements are squeezed and the layout does not adjust correctly.
https://github.com/user-attachments/assets/b8a52a8d-652b-451f-adeb-c8aa22652f72

## Expected Behavior
The Courses and Resources UI should adapt gracefully to landscape orientation, properly utilizing the available screen width and maintaining a readable, user-friendly layout.

Okuro3499:
fixes #15440
[responsive course.webm](https://github.com/user-attachments/assets/6696bd12-a6fc-44e5-b1f5-73106ac36ed9)

## Summary by CodeRabbit

* **New Features**
  * Added grid and list viewing options for courses and library resources, with preferences saved between sessions.
  * Redesigned course and resource cards with improved metadata, progress indicators, content types, subjects, and status labels.
  * Added responsive layouts for tablets and wider screens.
  * Added filter chips and improved search and filtering controls, including content-type and download-status filters.
  * Added long-press editing for your own resources.
  * Added localized labels for new views, filters, subjects, and download statuses.

* **Bug Fixes**
  * Improved course progress and submission handling for reliable status displays.


J-S-webskas (approved):
Really like this UI change.

dogi:
**Review — rating: 72/100**

57 files, +2318/−1037, so I reviewed at hunk level: the new utils in full, the adapter restructure, the layout rewrites, and the test change — sampling rather than reading every drawable.

The structure is better than I expected for a redesign this size. The new helpers are small, pure and well-bounded — `GridSpanCalculator.columnCount` is a clean `availableWidthDp / 165` clamped to 2–6, and `ListViewMode.fromPref` defaults to `GRID` on null or unknown input rather than throwing. The `CoursesAdapter` rework to `ListAdapter<Course, RecyclerView.ViewHolder>` with `VIEW_TYPE_GRID`/`VIEW_TYPE_LIST`, separate `GridViewHolder`/`ListViewHolder`, and `getItemViewType` driven by the current mode is the right shape, and the existing payload path was carried across to both holders rather than dropped — that's the detail most redesigns of this kind lose.

**`setViewMode` signals the wrong thing.** It calls `notifyItemRangeChanged(0, itemCount)`, but toggling the mode changes `getItemViewType` for *every* item. `notifyItemRangeChanged` announces a **content** change, so RecyclerView attempts to rebind into the holders it already has and only falls back to re-creating them via the view-type mismatch check in the recycler. It generally lands in the right place, but you're relying on a fallback rather than stating intent, and change animations will crossfade between two structurally different layouts. `notifyDataSetChanged()` is the honest call for a wholesale view-type switch.

**The one test file touched loses coverage rather than gaining it.** `CoursesAdapterTest` is edited only to keep compiling — `CoursesAdapter(mockContext, map, false, false)` → `CoursesAdapter(mockContext, false, false)` after `progressMap` left the constructor. So a 57-file redesign adds **zero** new tests, and the new code includes four pure, dependency-free classes that are almost free to cover: `GridSpanCalculator.columnCount` (boundaries at 2 and 6, and the clamp), `ListViewMode.fromPref` (null, unknown, both valid names), plus `LibraryTypeClassifier` and `CourseSubjectClassifier`. This repo has 166 unit-test files and CI enforces them; the new logic here is the easiest kind to test and none of it is.

**Scope.** This bundles a view-mode toggle, a search-pill restyle, a `MaxWidthFrameLayout` wrapper, new `sw600dp` variants, ~26 new drawables, subject/type classifiers, and repository plus `SharedPrefManager` changes into one PR. Each is defensible; together they're very hard to review carefully or to revert selectively if one piece misbehaves. Splitting the mechanical parts (drawables, classifiers, `MaxWidthFrameLayout`) from the behavioral ones would make the interesting half reviewable.

**Merge pressure worth planning around:** this collides with **#15318** (also restructures `CoursesAdapter` and `CoursesFragment`), **#15306** (`CoursesFragment`), and **#15441** (all four `fragment_my_*` layouts). On #15441 specifically I compared the diffs — **they are not duplicates**: it edits `tv_fragment_info`, the `filter` button and `ll_filter`, none of which this PR touches, while this one inserts the toggle include and wraps the recycler. Both should land; the second one through needs a rebase.

**Verdict:** Needs another round — fix the `setViewMode` notification and add tests for the four new pure-logic classes. The architecture itself is sound. Marking `change`.

---

dogi:
**Review — rating: 72/100**

61 files, +2467/−980, so I reviewed this by sampling rather than line by line: I read every new logic file in full (`GridSpanCalculator`, `ListViewMode`, the `SharedPrefManager` and `BaseAdapterFactory` deltas, the `CoursesRepositoryImpl` hunks), then spot-checked the resource layer by tracing which new tokens the new drawables actually consume. I did not read all 353+315 lines of the new `sw600dp` layouts.

**This is the best-engineered PR in the current review queue, and it's worth saying why.** The redesign extracts its decision logic into pure, testable objects instead of burying it in adapters — `GridSpanCalculator.columnCount()` (width/165dp coerced to 2..6), `ListViewMode.fromPref()` (defaults to GRID on null/unknown), `CourseSubjectClassifier`, `LibraryTypeClassifier` — and then **actually tests all four** (`GridSpanCalculatorTest`, `ListViewModeTest`, `CourseSubjectClassifierTest`, `LibraryTypeClassifierTest`, 206 lines of new tests). View mode is persisted per surface with separate `libraryViewMode`/`courseViewMode` keys rather than one shared flag. Dropping the `map: HashMap<String?, JsonObject>` parameter from `createCoursesAdapter` is a genuine API simplification. And `row_course.xml`/`row_library.xml` are deleted outright — green `build (default)` and `build (lite)` are meaningful evidence there, since a dangling `R.layout` reference would fail compilation.

Notably, **this PR does *not* use `?attr/colorOnPrimary`** — zero occurrences. I've just sent #15441 and #15523 back for exactly that, so it's worth recording that this one, which redesigns far more of the same screens, gets the theming right.

Two gaps keep it off `ready`, both in the dimension the PR is about — user-facing presentation.

**1. 17 new user-facing strings, zero translations.** `values/strings.xml` gains `filter_all`, `filter_books`, `filter_videos`, `filter_audio`, `filter_pdfs`, `downloaded`, `not_downloaded`, `subject_label_mathematics`/`_literacy`/`_health`/`_social_studies`/`_technology`, `item_count_pill`, `steps_done_of_total`, `course_steps_count`, `view_grid`, `view_list` — and not one of `values-ar`, `values-es`, `values-fr`, `values-ne`, `values-so` is touched. These are filter chips, status labels and subject labels, i.e. visible chrome, so all five locales fall back to English on the redesigned screens.

**2. 21 new colors, light-only.** They're semantic tokens with hardcoded light values — `card_border #E8E2D5`, `progress_track_bg #F4EFE6`, `progress_fill #7CB342`, `chip_selected_bg #E3F2FD`, `list_status_completed_bg #E8F5E9`, `list_status_in_progress_bg #FFF3D6`, `list_status_not_started_bg #F4EFE6` — and `values-night/colors.xml` is untouched. They're genuinely consumed: `card_border` alone appears in 8 new drawables/layouts (`bg_badge_circle`, `bg_chip_unselected`, `bg_search_pill`, `bg_segmented_control`, `item_course_grid`, …), and `progress_bar_flat` uses the track/fill pair. In dark mode these render as light borders, light progress tracks and light status pills against dark surfaces. The repo already maintains a `values-night/colors.xml`, so the convention exists here and is being skipped.

**Unrelated scope:** the seven `CoursesRepositoryImpl` hunks are nullability cleanups (`mapNotNull` → `map`, `step.id?.let {} ?: 0` → `step.id.let {}`, dropping `val title = courseTitle`) with nothing to do with grid/list views. They compile, so the null checks were provably redundant, and the change is fine on its own — it just doesn't belong in a UI redesign, and it widens the blast radius of an already-enormous diff. While in there: `examIds.map { it }` and `examDao.getByStepIds(stepIds).map { it }` are identity maps that can just be `examIds` and `getByStepIds(stepIds)`.

**Conflict note:** this modifies all four `fragment_my_course.xml` / `fragment_my_library.xml` files (portrait + land), which #15441 and #15523 also edit — a three-way overlap on the same header blocks. Both of those are now on `change`, so this one has right of way; whoever rebases second will need to redo their hunks against the new item layouts.

**Verdict:** needs another round — add the five translations and the `values-night` tokens, and this is a strong merge. The architecture and test coverage are not the problem.

---

dogi:
**Review — rating: 76/100**

67 files, +2668/−1078, so I reviewed at the hunk level: I read all the Kotlin in full (adapter factory, `CoursesRepositoryImpl`, `SharedPrefManager`, `MaxWidthFrameLayout`, `CourseFilterController`, `CoursesFragment`, the four new utils) and sampled the drawable/layout hunks for pattern consistency rather than reading all 40 resource files. I confirmed CI is current, not stale: head is `155a67a` at 11:49Z and the green run started 12:17Z, so build (default + lite) and test (default) genuinely cover this code.

The engineering standard here is high. The four new utils — `ListViewMode`, `GridSpanCalculator`, `CourseSubjectClassifier`, `LibraryTypeClassifier` — are pure, total functions pulled out of the UI, and each ships with its own unit test, which is exactly what `CLAUDE.md` asks for at this layer. All 18 new strings are translated into all five locales (I checked several by name — each appears 6×, `values/` plus `ar/es/fr/ne/so`). `MaxWidthFrameLayout` is a correct `onMeasure` override that preserves the incoming mode. Dropping the dead `map: HashMap<String?, JsonObject>` parameter from `createCoursesAdapter` is a real simplification, and the follow-up commit switching `setViewMode()` from `notifyItemRangeChanged()` to `notifyDataSetChanged()` is the right call since a view-mode change invalidates every item's view type.

Two things hold it under the line.

**The courses progress chips are English-only, and can't be translated as written.** The library chips got proper localized resources (`filter_all`, `filter_books`, `filter_videos`, `filter_audio`, `filter_pdfs`, all five locales). The courses chip row instead reads `R.array.progress_filter`, which is untranslated, and passes the display label straight through `filterController.setProgressFilter(label)`. `CoursesViewModel.filterCourses` then string-matches it against hardcoded English literals:

```kotlin
when (progressFilter) {
    "Not Started" -> current == 0
    "In Progress" -> current > 0 && (max == 0 || current < max)
    "Completed"   -> max > 0 && current >= max
    else -> true
}
```

So the new chip row renders English text in Arabic, Spanish, French, Nepali and Somali — the one untranslated surface in an otherwise thorough PR. Worse, it's a trap: anyone who later translates `progress_filter` silently breaks the filter into `else -> true` with no compile error and no test failure. The chips are more prominent than the spinner they replace, so this is the moment to pass a stable enum or index instead of display text, then localize the labels. Using `options.first()` for "All" instead of the old literal `"All"` shows the coupling was already noticed — worth finishing the job.

**Ratings and tags are removed from course cards.** `setRatingMap`, `setTagsMap` and `setRatingChangeListener` are all dropped, along with the factory's `map` parameter. The commit message states this is intentional, and I'm not second-guessing the visual direction — but users lose the ability to rate a course from the list, and tag chips disappear, while `CoursesFragment` still implements `OnTagClickListener`. That is a user-facing capability removal riding inside a PR whose issue (#15440) is "not responsive in landscape mode", and it deserves an explicit maintainer decision rather than arriving as a side effect. Note it also brushes against #15442 in the current review batch, which adds a post-completion rating dialog.

**Nits:** `updateToggleUi()` reassigns `recyclerView.layoutManager` as a side effect, which its name doesn't suggest and which discards scroll position — worth splitting out. `CourseSubjectClassifier` funnels unknown *and* blank subjects into `LITERACY`, so uncategorized courses get a wrong badge rather than a neutral one. A couple of `mapNotNull { it.id }` → `map { it.id }` changes in `CoursesRepositoryImpl` relax null filtering; harmless against SQL `IN (...)`, and it compiles, just a semantic loosening worth being deliberate about.

⚠️ **Conflict:** this rewrites the same four `fragment_my_course.xml` / `fragment_my_library.xml` variants that #15441 edits, replacing that header block entirely. #15441 is a small header fix that this PR subsumes. I have deliberately *not* marked either as a duplicate — they're different jobs that happen to collide on the same lines, and closing a 67-file redesign in favour of a 4-file cosmetic tweak (or vice versa) would be wrong. Sequence them: if this lands first, #15441 is moot; if #15441 lands first, the rebase here is trivial since these files are rewritten anyway.

**Verdict:** needs another round — localize the progress chips via a stable filter value, and get explicit sign-off on dropping ratings/tags from the cards. The layout, tests and translation work are solid and should survive as-is.

---

dogi:
**Re-review — Rating: 76/100** (was 76)

No commits have landed since the last review — head is still `29c5330`, and the check runs are the same ones (build default + lite, test default, all green). So the score stands, and neither of the two blockers has been addressed.

**Correcting my earlier review on one point, because it changes the fix.** I previously wrote that `R.array.progress_filter` is untranslated and that the chip row would therefore render English labels in the five other locales. That is wrong — the array *is* defined in all six `strings.xml` buckets (`values/` plus `ar`, `es`, `fr`, `ne`, `so`). The labels will display correctly translated.

The consequence is worse than what I described, not better. Because `setupCourseFilterChips` passes the localized label straight through:

```kotlin
filterController.setProgressFilter(if (label == options.first()) "" else label)
```

…and `CoursesViewModel.filterCourses` matches it against hardcoded English literals:

```kotlin
when (progressFilter) {
    "Not Started" -> current == 0
    "In Progress" -> current > 0 && (max == 0 || current < max)
    "Completed"   -> max > 0 && current >= max
    else -> true
}
```

…a Spanish user tapping "En progreso" hits `else -> true` and gets **every** course back. So this isn't a cosmetic labelling issue — the progress filter is functionally dead in Arabic, Spanish, French, Nepali and Somali, and only "All" appears to work, by accident, via the `options.first()` special case. To be fair: this is pre-existing, since the spinner it replaces read the same translated array and compared against the same literals. But the chips are a more prominent surface than the spinner was, the rewrite is right here, and `options.first()` shows the coupling was already half-noticed. The fix is to pass a stable key — an enum or the array index — rather than display text, and match on that. Translating the array is *not* the fix; it's already translated.

The second blocker is unchanged: ratings and tags are still removed from the course cards (`setRatingMap`, `setTagsMap`, `setRatingChangeListener` and the factory's `map` parameter all dropped), which takes away rating-from-the-list and tag chips as a side effect of a landscape-responsiveness issue. That still wants an explicit maintainer decision rather than arriving inside #15440.

Everything I praised last time still holds and is worth restating so it doesn't get lost in a rework: the four pure utils each with their own unit test, all 18 new strings translated across all five locales, the dead `HashMap<String?, JsonObject>` parameter removed, and the `notifyDataSetChanged()` fix for view-mode switches.

**Conflict status has improved.** #15441 is now `on hold`, so that collision is sequenced. #15523 is still `ready` and still rewrites the same `fragment_my_course.xml` header this PR replaces, so that one remains live.

**Verdict:** needs another round — same two items as before. Pass a locale-independent value for the progress filter, and get sign-off on dropping ratings/tags. Staying `change`.

---

dogi (approved):

PR #15519, opened 2026-08-10
add dismiss button to last synced status container (fixes #15516)

### LINKED ISSUE #15516: add dismiss (close) button to the "last synced" status container
Currently, the "Last Synced" status container is permanently displayed at the top of the dashboard beneath the app bar/toolbar. While this status is useful to check recent sync activity, having it permanently visible takes up valuable screen space, reducing the visible content area on smaller screens and in landscape mode.

ragilzakaria:
https://github.com/user-attachments/assets/10cf94fc-c8a8-4d22-92f9-e5a14a22b088

fixes #15516 

## Summary by CodeRabbit

* **New Features**
  * Added a dismiss button to the dashboard sync-status banner.
  * Banner dismissal is preserved when the dashboard is recreated.
  * Successful synchronization clears the dismissal, allowing the banner to reappear.
  * Sync-status banner visibility is limited to portrait orientation.

* **Accessibility**
  * Added accessible text and clear visual styling for the dismiss control.

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/ui/dashboard/DashboardActivity.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/dashboard/DashboardActivity.kt
@@ -129,6 +129,7 @@ class DashboardActivity : DashboardElementActivity(), OnHomeItemClickListener, N
     private var onGlobalLayoutListener: ViewTreeObserver.OnGlobalLayoutListener? = null
     private var exitSnackbar: Snackbar? = null
     private var lastSyncStatus: SyncManager.SyncStatus? = null
+    private var dismissedSyncTime = -1L
 
     override fun attachBaseContext(base: Context) {
         super.attachBaseContext(LocaleUtils.onAttach(base))
@@ -136,6 +137,9 @@ class DashboardActivity : DashboardElementActivity(), OnHomeItemClickListener, N
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
+        if (savedInstanceState != null) {
+            dismissedSyncTime = savedInstanceState.getLong(KEY_DISMISSED_SYNC_TIME, -1L)
+        }
         postponeEnterTransition()
         initViews()
 
@@ -274,6 +278,10 @@ class DashboardActivity : DashboardElementActivity(), OnHomeItemClickListener, N
         }
         binding.dashboardSyncNow.setOnClickListener {
             logSyncInSharedPrefs()
+        }
+        binding.btnDismissSync.setOnClickListener {
+            dismissedSyncTime = prefData.getLastSync()
+            binding.dashboardSyncBanner.visibility = View.GONE
         }
     }
 
@@ -588,8 +596,9 @@ class DashboardActivity : DashboardElementActivity(), OnHomeItemClickListener, N
 
     private fun updateLastSyncStatus() {
         val isLandscape = resources.configuration.orientation == Configuration.ORIENTATION_LANDSCAPE
-        binding.dashboardSyncBanner.visibility = if (isLandscape) View.GONE else View.VISIBLE
-        if (isLandscape) return
+        val isDismissed = dismissedSyncTime >= 0L && prefData.getLastSync() <= dismissedSyncTime
+        binding.dashboardSyncBanner.visibility = if (!isLandscape && !isDismissed) View.VISIBLE else View.GONE
+        if (binding.dashboardSyncBanner.visibility == View.GONE) return
 
         val lastSyncMillis = prefData.getLastSync()
         val timeText = if (lastSyncMillis <= 0L) {
@@ -1081,6 +1090,11 @@ class DashboardActivity : DashboardElementActivity(), OnHomeItemClickListener, N
         updateLastSyncStatus()
     }
 
+    override fun onSaveInstanceState(outState: Bundle) {
+        super.onSaveInstanceState(outState)
+        outState.putLong(KEY_DISMISSED_SYNC_TIME, dismissedSyncTime)
+    }
+
     override fun onNewIntent(intent: Intent?) {
         super.onNewIntent(intent)
         setIntent(intent)
@@ -1122,6 +1136,7 @@ class DashboardActivity : DashboardElementActivity(), OnHomeItemClickListener, N
     }
 
     companion object {
+        const val KEY_DISMISSED_SYNC_TIME = "dismissed_sync_time"
         const val MESSAGE_PROGRESS = "message_progress"
         var isFromNotificationAction = false
     }
--- a/app/src/main/res/layout/activity_dashboard.xml
+++ b/app/src/main/res/layout/activity_dashboard.xml
@@ -68,6 +68,16 @@
                 android:textColor="@color/dashboard_accent_blue"
                 android:textSize="12sp"
                 android:textStyle="bold" />
+            <ImageView
+                android:id="@+id/btn_dismiss_sync"
+                android:layout_width="48dp"
+                android:layout_height="48dp"
+                android:background="?attr/selectableItemBackgroundBorderless"
+                android:contentDescription="@string/dismiss"
+                android:padding="12dp"
+                android:scaleType="centerInside"
+                app:srcCompat="@drawable/baseline_close_24"
+                app:tint="@color/daynight_textColor" />
         </LinearLayout>
 
         <com.google.android.material.bottomnavigation.BottomNavigationView
--- a/app/src/main/res/values-ne/strings.xml
+++ b/app/src/main/res/values-ne/strings.xml
@@ -883,7 +883,7 @@
     <string name="auto_sync_device">स्वचालित सिङ्क यन्त्र</string>
     <string name="beta_functionality">बीटा कार्यक्षमता</string>
     <string name="main_controls">मुख्य नियन्त्रणहरू</string>
-    <string name ="dismiss"> खारेज गर्नुहोस्</string>
+    <string name="dismiss">खारेज गर्नुहोस्</string>
     <string name="available_space_colon">उपलब्ध ठाउँ: </string>
     <string name="community_leaders">सामुदायिक नेताहरू</string>
     <string name="services">सेवाहरू</string>
--- a/app/src/test/java/org/ole/planet/myplanet/ui/dashboard/DashboardActivityTest.kt
+++ b/app/src/test/java/org/ole/planet/myplanet/ui/dashboard/DashboardActivityTest.kt
@@ -1,0 +1,96 @@
+package org.ole.planet.myplanet.ui.dashboard
+
+import android.os.Bundle
+import android.view.View
+import dagger.hilt.android.testing.HiltAndroidRule
+import dagger.hilt.android.testing.HiltAndroidTest
+import dagger.hilt.android.testing.HiltTestApplication
+import javax.inject.Inject
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertNotNull
+import org.junit.Before
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.ole.planet.myplanet.R
+import org.ole.planet.myplanet.services.SharedPrefManager
+import org.ole.planet.myplanet.utils.UrlUtils
+import org.robolectric.Robolectric
+import org.robolectric.RobolectricTestRunner
+import org.robolectric.annotation.Config
+import org.robolectric.annotation.LooperMode
+
+@HiltAndroidTest
+@RunWith(RobolectricTestRunner::class)
+@Config(manifest = Config.NONE, application = HiltTestApplication::class)
+@LooperMode(LooperMode.Mode.PAUSED)
+class DashboardActivityTest {
+
+    @get:Rule
+    val hiltRule = HiltAndroidRule(this)
+
+    @Inject
+    lateinit var sharedPrefManager: SharedPrefManager
+
+    @Before
+    fun setUp() {
+        hiltRule.inject()
+        UrlUtils.init(sharedPrefManager)
+    }
+
+    @Test
+    fun testSyncBannerDismissalSurvivesRecreation() {
+        sharedPrefManager.setLastSync(1000L)
+        val controller = Robolectric.buildActivity(DashboardActivity::class.java).setup()
+        val activity = controller.get()
+        assertNotNull(activity)
+
+        val syncBanner = activity.findViewById<View>(R.id.dashboard_sync_banner)
+        val dismissButton = activity.findViewById<View>(R.id.btn_dismiss_sync)
+        assertNotNull(syncBanner)
+        assertNotNull(dismissButton)
+
+        assertEquals(View.VISIBLE, syncBanner.visibility)
+
+        dismissButton.performClick()
+        assertEquals(View.GONE, syncBanner.visibility)
+
+        val bundle = Bundle()
+        controller.saveInstanceState(bundle).pause().stop().destroy()
+
+        val recreatedController = Robolectric.buildActivity(DashboardActivity::class.java).setup(bundle)
+        val recreatedActivity = recreatedController.get()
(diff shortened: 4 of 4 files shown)


Okuro3499 (commented):
After dismissing is there any flow of bringing it back? as it is supposed to inform user of the last sync time

dogi:
**Review — rating: 85/100**

Clean, minimal, and the state handling is actually right — which is the part these dismiss-button PRs usually get wrong. Specifics I verified:

- **Dismissal persists.** `updateLastSyncStatus()` is called from `onResume()` (`:1058`), and the new visibility line respects the flag, so leaving the dashboard and coming back keeps the container hidden instead of resurrecting it. It also can't be un-dismissed by rotation, because `DashboardActivity` declares `android:configChanges="orientation|screenSize|keyboardHidden"` in the manifest and isn't recreated.
- **Nothing else fights over the view.** `grep` for `dashboardLastSyncStatus|layoutLastSync` shows only the two touchpoints in this diff, so wrapping the `TextView` in a new container doesn't collide with any other visibility writer.
- **The reset trigger is the right one.** Clearing the flag on `SyncManager.SyncStatus.Success` means a fresh sync brings the status back with a new timestamp, rather than the dismissal being permanent for the process.
- **No empty-bar flash.** Since `onResume` populates the text before the activity is visible, the container never shows as a bare strip with a lone X.
- **Resources all resolve.** `@drawable/baseline_close_24` exists, `@string/dismiss` exists, and `xmlns:app` is declared on the `CoordinatorLayout` root so `app:srcCompat`/`app:tint` are fine. Worth calling out that `app:tint="@color/daynight_textColor"` points at a real color resource with a `values-night` variant — not a theme attribute — so it has none of the resolution risk that the same-shaped change carries in #15441.

**Nit:** two small ones. `btn_dismiss_sync` is an `ImageView` with a click listener, so TalkBack won't announce it as a button even though the `contentDescription` is set — `ImageButton` would give it the right role for free. And `@string/dismiss` is missing from `values-ne` (it's a pre-existing string, but this PR is what makes it user-facing, so Nepali will read "Dismiss").

**Verdict:** mergeable.

---

dogi:
**Follow-up — interaction with #15525**

Correcting one specific claim in my review above. I passed this partly because `isSyncStatusDismissed` can't be reset by rotation, since `DashboardActivity` declares `android:configChanges="orientation|screenSize|keyboardHidden"` and isn't recreated. That holds on `master` today, but **#15525 removes `orientation|screenSize` from that attribute** — necessarily, because it adds `layout-land` and `layout-sw600dp-land` variants that Android would otherwise never re-inflate.

Once #15525 lands, `DashboardActivity` is recreated on every rotation, `isSyncStatusDismissed` resets to `false`, and a dismissed sync banner reappears each time the device rotates.

Nothing here is wrong on its own, so I'm leaving this on `ready` — but whichever of the two merges second should persist the flag, either via `onSaveInstanceState`/`savedInstanceState` or by moving it into a ViewModel. The ViewModel route is probably cleaner given `DashboardViewModel` already exists. Flagged on #15525 as well.

---

ragilzakaria:
> After dismissing is there any flow of bringing it back? as it is supposed to inform user of the last sync time

currently we have to close and reopen the app to bring the last synced back.

Okuro3499 (dismissed):
@ragilzakaria 
```
Given this banner's whole purpose is to show last-sync-time at a glance, I'd reset isSyncStatusDismissed = false in onResume() (DashboardActivity.kt:1063) instead of only on 
sync success. That makes "dismiss" mean "hide for this viewing session" — it reappears whenever the user leaves and comes back to the Dashboard (e.g. from Settings or another
activity), but doesn't nag you every second while you're actively on the screen. The tradeoff: if the user wants it gone for good until the next sync, this reintroduces it   
more often than that — but a permanently-dismissible informational status (with no other way back) seems worse.  
```

ragilzakaria:
Changes Made:
- Tapping the dismiss button ("✕") hides the last-synced banner for the current viewing session
- When the user leaves and returns to the Dashboard (e.g. from Settings or another activity), onResume() resets the dismissed state and updates the banner with the fresh relative sync time
- Successful sync completions continue to un-dismiss and refresh the banner

dogi:
**Review — rating: 68/100**

The plumbing is clean — the `TextView` becomes a weighted row with a borderless close affordance, `updateLastSyncStatus()` now honours the flag so nothing else fights it, and resetting on `SyncManager.SyncStatus.Success` is exactly the right trigger to bring the banner back. `@string/dismiss` and `@drawable/baseline_close_24` both already exist on master, and build (default + lite) and test (default) are green. I traced the other two call sites of `updateLastSyncStatus()` (`DashboardActivity.kt:246` and `:1066`) and both now route through `resetSyncStatusDismissed()`, so there's no path that leaves the flag and the visibility out of sync.

The problem is that the dismissal doesn't survive anything. `resetSyncStatusDismissed()` is wired into `onResume()`, `openCallFragment(newFragment, tag)` and `onBackStackChanged()`, so the banner comes back on every bottom-nav tap, every back press, and every return to the app. Issue #15516 asks for this because the container "takes up valuable screen space, reducing the visible content area on smaller screens and in landscape mode" — a dismissal that is undone by the next navigation doesn't reclaim that space; it just means the user re-taps the close button on every screen. The state you already have is the right one, it's the reset set that's too broad: keep the `SyncStatus.Success` reset, drop the `onResume`/navigation resets, and hold the flag for at least the session (`prefData` would make it survive process death, which matches the "permanently displayed" complaint best).

Two smaller things:

- The 32dp `ImageView` with 6dp padding is below the 48dp minimum touch target; bumping the view to 48dp and keeping the icon small via `padding` gets the same visual weight with a reachable target.
- Widening `openCallFragment` to `open` in `DashboardElementActivity` purely to hook banner visibility couples a fragment-transaction API to a UI detail, and it's likely redundant with the `onBackStackChanged()` override for any transaction that goes on the back stack.

**Verdict:** needs another round — the reset set defeats the feature's stated purpose.

---

dogi:
**Review — rating: 72/100**

The dismiss button itself is well built. 48dp touch target, `?attr/selectableItemBackgroundBorderless`, `contentDescription="@string/dismiss"` — I checked and both `@string/dismiss` and `@drawable/baseline_close_24` already exist, so nothing is dangling. The banner hides on tap and CI is green on all four checks.

**The persisted flag can never be read as `true`.**

`resetSyncStatusDismissed()` writes `setSyncStatusDismissed(false)`, and it's wired to four places:

```kotlin
override fun onResume() { super.onResume(); resetSyncStatusDismissed(); … }
override fun openCallFragment(newFragment: Fragment, tag: String?) { resetSyncStatusDismissed(); … }
override fun onBackStackChanged() { super.onBackStackChanged(); resetSyncStatusDismissed() }
// plus on SyncStatus.Success
```

`onResume` runs on every launch and every return from background, so by the time anything reads `isSyncStatusDismissed()` the stored value has already been reset. The flag lives, at most, from the dismiss tap until the next navigation or resume — which is strictly shorter than the Activity itself.

That makes the `SharedPreferences` round-trip dead weight. A plain `private var syncStatusDismissed = false` field on `DashboardActivity` produces byte-for-byte identical behaviour, without adding a permanent key to the prefs file, without a `SharedPrefManager` accessor pair, and without the test. It also removes a disk write from every `onResume`.

So one of the two has to go, and which one is a product question worth answering explicitly:

- If dismissing should mean *"hide it until something changes"* — the current behaviour — drop `SYNC_STATUS_DISMISSED` from `SharedPrefManager` entirely and use a field.
- If dismissing should mean *"I've seen this sync, don't show it again"* — which is what persisting implies — then `onResume` must stop resetting it, and only `SyncStatus.Success` should clear it.

The commit message (`smoother menu navigation sync banner resetting`) suggests the first, in which case the prefs work is the part to delete.

**The version bump belongs to the automerge script.** `app/build.gradle` goes `6546` → `6547`, and `.github/scripts/automerge.sh` does exactly that as part of draining the queue — every merge commit on master shows a +1 bump from it. Right now #15769, #15820 and #15824 all claim `6547` as well, so at most one of the four can merge without a `build.gradle` conflict. Dropping the hunk here removes this PR from that pileup.

**Smaller things.**

`openCallFragment(newFragment, tag)` in `DashboardElementActivity` had to become `open` for the override. That's fine, but `DashboardActivity` now overrides both that and the single-argument `openCallFragment(f: Fragment)` from the interface, and only one of them resets — the single-arg version delegates to the two-arg one, so it works out, but it's worth a glance to confirm every navigation path actually goes through the overridden overload.

The new `ImageView` sits inside the banner's `LinearLayout` after the sync-now text. Worth checking in landscape and at large font scales that it doesn't push the "last synced" text out — that's the same class of problem #15263 and #15577 are about.

`testGetAndSetSyncStatusDismissed` follows the file's existing pattern correctly. If the prefs accessor goes away per the above, the test goes with it.

**Verdict:** `change` — decide whether the dismissal is meant to persist, and make the storage match. Drop the version bump either way.

---

dogi:
**Review — rating: 72/100**

Adds a dismiss button to the last-synced banner behind a `syncStatusDismissed` flag. I verified the resource side is complete: `@string/dismiss` exists in `values/` plus all five translations (ar/es/fr/ne/so), `baseline_close_24` is present, and the banner's `LinearLayout` is horizontal with `dashboard_last_sync_status` at `width=0dp, weight=1, ellipsize=end` — so the new 48dp button squeezes the flexible text rather than cropping the sync-now action. I also confirmed the pre-existing `updateLastSyncStatus()` call sites (`onResume`, sync `Success`) are preserved through `resetSyncStatusDismissed()`, so existing banner behavior doesn't regress. Build (default + lite) and test (default, both shards) are green.

Two real problems, though:

1. **The `openCallFragment` override is redundant, and wrong on the one path where it isn't.** I walked every branch of `DashboardElementActivity.openCallFragment`: each one either calls `replaceFragment(addToBackStack = true)` or `popBackStack`, both of which change the back stack and therefore already fire your new `onBackStackChanged` reset. The only branch that *doesn't* navigate is the early `return` when `existingFragment != null && existingFragment.isVisible` — and there the override still resets the flag, so re-tapping the tab you're already on resurrects a banner the user just dismissed. Net effect: the override duplicates `onBackStackChanged` everywhere it matters and only adds an unwanted reset where it doesn't — yet it's the reason `DashboardElementActivity.openCallFragment` had to be widened from `fun` to `open fun`. Dropping the override lets that widening revert too.

2. **The dismissal is short-lived enough to question the feature.** Between `onResume`, `onBackStackChanged` and rotation (the flag isn't saved in instance state), the banner returns on essentially any user action — navigate to Library and it's back. If a per-screen reset is the intended semantics that's worth stating in the issue, but as it stands a "dismiss" that doesn't survive tapping a tab is close to a no-op.

**Nit:** no test covers the new flag; `resetSyncStatusDismissed()` and the `isLandscape || syncStatusDismissed` gate are both cheaply unit-testable.

**Verdict:** needs another round — drop the `openCallFragment` override (and the `open` widening with it), and settle how long a dismissal is meant to last.

---

dogi:
**Review — rating: 66/100**

The layout work is right and the plumbing is mostly right, but the dismissal is undone by the first rotation — which is the exact case the `onSaveInstanceState` code was added to handle.

**The `Success` status is sticky, so recreating the activity re-fires the reset.**

```kotlin
collectWhenStarted(syncManager.syncStatus) { status ->
    if (status == lastSyncStatus) return@collectWhenStarted
    lastSyncStatus = status
    if (status is SyncManager.SyncStatus.Success) {
        resetSyncStatusDismissed()     // sets syncStatusDismissed = false
    }
}
```

`SyncManager` is `@Singleton` (`SyncManager.kt:59`) and `_syncStatus` is a `MutableStateFlow` whose terminal value after a sync is `SyncStatus.Success("Sync completed")`, set in `destroy()` at line 120. `resetSyncStatus()` exists at line 107 but **nothing calls it** — `DashboardActivity:245` and `SyncActivity:149` are the only two consumers of the flow and both just collect. So once a sync has completed, the StateFlow holds `Success` for the rest of the process lifetime.

Now walk a rotation after dismissing:

1. User taps dismiss → `syncStatusDismissed = true`, banner GONE.
2. Rotate. `onSaveInstanceState` writes `true`; the new `onCreate` reads it back correctly.
3. `collectWhenStarted` re-subscribes, and a StateFlow replays its current value immediately — still `Success("Sync completed")`.
4. `lastSyncStatus` is a plain field on the *new* instance, so it's `null`. `Success(…) == null` is false, so the dedup guard doesn't fire.
5. `resetSyncStatusDismissed()` runs → `syncStatusDismissed = false` → banner is back.

`Success` is a `data class` with a constant message, so the replayed value is stable and equal every time; the only case where dismissal survives rotation is a process that has never completed a sync, when the flow is still `Idle`. The same applies to any recreation — theme change, locale change, "don't keep activities", process-death restore.

The cleanest fix uses state you already render: capture `prefData.getLastSync()` at dismissal time, save *that* in the bundle instead of a boolean, and show the banner again only when `getLastSync()` has moved past it. That's naturally recreation-proof and expresses the actual rule ("stay hidden until there's a newer sync") rather than inferring it from a flow emission. Alternatives: save `lastSyncStatus` alongside the flag, or call the unused `syncManager.resetSyncStatus()` after consuming a `Success`.

**What's good and worth keeping:** the button is 48dp with `selectableItemBackgroundBorderless` and a `contentDescription`, which is the correct touch target and a11y treatment; and it lands at the right edge properly because `dashboard_last_sync_status` is `0dp` + `layout_weight="1"`, so the new fixed-width child doesn't disturb the row. Resetting on the next successful sync (rather than hiding forever) is the right product call for a "last synced N ago" nag.

**Nit:** `@string/dismiss` exists in `values/` and in `ar`, `es`, `fr`, `so` — but not `values-ne`, so Nepali TalkBack reads the English word. That gap predates this PR (the string already existed), but this adds a consumer, so it's a cheap one to close while here.

**Nit:** `onResume` swaps `checkNotificationPermissionStatus()` and `updateLastSyncStatus()` with no stated reason — unrelated to the dismiss button, and reordering two side-effecting calls is the kind of thing that's hard to attribute later.

**Nit:** the four tests enumerate the 2×2 truth table of `!isLandscape && !isDismissed`, so they restate the implementation and would all pass with the bug above present. Extracting a public companion function to enable that widens the class's API for no behavioural coverage. The test worth writing is the Robolectric one — dismiss, recreate the activity, assert the banner is still GONE — which fails today and would pass after the fix.

**Verdict:** the UI is well built; the dismissal needs to survive recreation before this is usable.

---

dogi:
**Re-review — Rating: 87/100** (was 66)

This takes the timestamp approach and it closes the rotation bug properly. The reason it works is worth spelling out, because the old boolean died on a path that's easy to miss: the sync collector at `DashboardActivity:248` short-circuits on `status == lastSyncStatus`, but `lastSyncStatus` is a plain activity field that resets to `null` on recreation — so after a rotation the replayed `Success` is *not* deduped and does call `updateLastSyncStatus()`. That's exactly the call that used to un-dismiss the banner. Now that call recomputes `isDismissed` from the long restored out of the bundle, so a replayed status can't clear it. State derived from a compared value instead of a mutable flag is the right shape here.

It also lands better semantics than "dismiss until rotation": `prefData.getLastSync() <= dismissedSyncTime` means the banner stays hidden until a genuinely newer sync arrives, which is what a user dismissing a "last synced" notice would expect. `testSyncBannerShowsAgainAfterNewSync` pins precisely that, and both tests fail against master — the first can't even find `btn_dismiss_sync`.

The button itself is done carefully: 48dp touch target, `contentDescription="@string/dismiss"` (which resolves — `values/strings.xml:888`), and `selectableItemBackgroundBorderless`. Tidying the malformed `<string name ="dismiss"> …>` entry in `values-ne` while you're using that string is the right call too, leading space and all.

**Nit:** `dismissedSyncTime` lives only in the instance-state bundle, so it survives rotation and process-death restore but not a genuine `finish()`. That's a defensible choice for a dismissible banner — worth a one-line comment saying it's deliberate, since the next reader will wonder. `KEY_DISMISSED_SYNC_TIME` could also be `private const val`; nothing outside the activity reads it.

**Verdict:** the timestamp comparison makes the dismissal survive exactly the replay that broke it before, and both tests fail on master; ready.

---

Issue #1, opened 2018-04-12
Service that checks if google exists (has internet) 


TerrenceEJ:
If wifi is on, can we connect to the RPI, (First big step) check if we can download certain files.

PR #15523, opened 2026-08-10
consolidate select all search courses header (fixes #15520)

### LINKED ISSUE #15520: consolidate Select All and Search Bar into top header container in Courses and myCourses
Currently, in both **Courses** ("Our Courses") and **My Courses**, the top section uses two separate stacked containers. Having these controls in two stacked vertical rows takes up excessive screen height, reducing the visible space available for course cards and list content (especially on smaller devices and in landscape mode).

ragilzakaria:
https://github.com/user-attachments/assets/2c1849b1-3a90-4ea6-8b4d-02ccf33c4341

fixes #15520 

## Summary by CodeRabbit

* **New Features**
  * Added an integrated course search field to the course header.
  * Added support for tag-based filtering within the course search layout.
  * Added a select-all checkbox directly to the course header.

* **Style**
  * Improved header spacing, alignment, sizing, visibility, and theme-aware colors across portrait and landscape layouts.
  * Simplified the filter button to display its icon while retaining accessibility support.


dogi:
**Review — rating: 63/100**

The consolidation is the right call for #15520 — folding the select-all checkbox and the search row into the header reclaims a whole horizontal band, and several incidental changes are genuine improvements. `layout_gravity="center"` → `center_vertical` is the correct gravity in a horizontal `LinearLayout`; dropping `android:text`/`android:textColor` from the `filter` `ImageButton` removes attributes an `ImageView` subclass never renders; and I checked the one change that looked like a silent regression — `android:visibility="gone"` on `tv_add` actually *matches* runtime state, since `CourseSelectionController.refreshActionVisibility()` sets `tvAddToLib` GONE whenever `currentSelectedCount == 0`. The XML was previously lying about the initial frame; now it doesn't.

Four things need work.

**1. `?attr/colorOnPrimary` doesn't resolve in this app.** It's used three times here (`tv_fragment_info` textColor, `selectAllCourse` textColor and `buttonTint`). `colorOnPrimary` is a MaterialComponents attribute; it is not defined anywhere in `app/src/main/res`, and the theme chain for this screen never introduces it — `DashboardActivity` → `@style/FullscreenTheme` → `MaterialDrawerTheme.Light.DarkToolbar.TranslucentStatus` → MaterialDrawer **6.1.1** (AppCompat-era) → `Theme.AppCompat.*`. The only theme that would supply it, `AppTheme.MaterialComponents` (`styles.xml:210`, a `.Bridge` parent), isn't referenced by any activity. With the attribute unresolved the colors simply aren't applied and these views fall back to default dark text. CI can't catch this — the material dependency declares the `attr` symbol, so it compiles clean. This is the same issue as #15441, which I've just sent back for it.

**2. A light-mode contrast regression that's real even if you fix #1.** The checkbox and search input move from `@color/secondary_bg` (`#FFFFFF`) onto the header's `@color/colorPrimary` (`md_blue_700`), but `et_search` keeps `android:textColor="@color/daynight_textColor"`, which is `@color/md_black_1000` in `values/colors.xml`. So in light mode you get black search text and a `#808080` hint on medium blue. Night mode is fine (`colorPrimary` is `#18191a`, `daynight_textColor` resolves light), so this will look correct on a dark-theme device and wrong on a light one.

**3. `layout_search.xml` is shared by six layouts, and this edits it.** `grep -rl 'layout="@layout/layout_search"'` returns courses (portrait + land), library (portrait + land), **voices**, and **survey**. Removing `paddingTop`/`paddingBottom` (`padding_normal`), changing `paddingEnd` from `padding_very_large` to `padding_small`, and setting `app:hintEnabled="false"` therefore changes the search row on Voices, Survey, and Library too — screens this PR isn't consolidating and where the row still sits on its own background, so the tighter vertical padding is uncompensated. Either scope those changes to the courses header (a courses-specific search layout, or override the padding at the `<include>` site) or state that the other four screens were checked.

**4. The single row is now crowded.** `tv_fragment_info` loses both `maxLines="1"` and `minWidth="150dp"` and becomes `wrap_content`, sharing one horizontal row with the checkbox, a weighted search, and the 40dp filter button. A longer localized `txt_myCourses` (ar/ne) can now wrap to two lines or squeeze the weighted search toward zero on a narrow phone. Keeping `maxLines="1"` plus `ellipsize="end"` on the title would bound it.

**On the overlap with #15441:** these two edit the *same lines* of `layout/fragment_my_course.xml` and `layout-land/fragment_my_course.xml` — the `tv_fragment_info` block, the container's `padding_large` → `padding_small` line, and the `filter` `ImageButton` block — so whichever lands second will need a rebase. I'm deliberately **not** marking either a duplicate: #15441's actual deliverable is bringing the *library* layouts in line with courses, which this PR never touches, so neither subsumes the other. Rate and land them independently.

**Verdict:** needs another round — drop `?attr/colorOnPrimary`, fix the light-mode contrast on the blue header, and scope the `layout_search.xml` changes.

---

dogi:
**Review — rating: 84/100**

Folds the select-all checkbox and the search field up into the courses header, deleting the separate `secondary_bg` row below it, and adds a courses-specific `layout_search_courses.xml`. Portrait and landscape are updated in step. I chased the two ways this could have broken and both are handled correctly:

`tv_add` gains `android:visibility="gone"`, which looks at first glance like it hides the "Join selected" action. It doesn't — `CourseSelectionController` owns that view's visibility end to end (GONE at lines 36/83/116, VISIBLE at 109, plus `BaseRecyclerFragment` forcing GONE), so the XML default was never what the user saw. Making it `gone` in the layout actually removes a brief flash of the button before the controller runs. Good change.

More importantly, the new layout keeps the `tl_tags` block. That is load-bearing: `CoursesFragment.kt:187` does `requireView().findViewById<View>(R.id.tl_tags).visibility = View.GONE` with no null guard, so dropping that ID while swapping the include would have been an NPE on every entry into the courses screen. Retaining it — and hardcoding `visibility="gone"` rather than relying on the runtime call as `layout_search.xml` does — is the right call and avoids a flash there too. I also confirmed `layout_search.xml` is still referenced by `fragment_my_library.xml`, `fragment_voices.xml`, `fragment_survey.xml` and the landscape library layout, so nothing is orphaned, and there is no `layout-sw600dp/` variant on master to leave stale.

The colour adaptation is right for the new context: the checkbox's `buttonTint`/`textColor` move to white now that it sits on `colorPrimary` instead of `secondary_bg`, and `maxLines="1"` is correctly paired with `ellipsize="end"` (on master `maxLines` alone just hard-clipped).

**Nits:**
- The inner `LinearLayout` wrapping the second `TextInputLayout` is redundant — `layout_weight="1"` on a sole child inside a `layout_weight="1"` parent. It can be flattened.
- Worth an eyeball on the hint: `android:textColor` is explicitly white on the `EditText`, but the hint colour is set as `android:textColorHint` on the `TextInputLayout` while `app:hintEnabled="false"`. With the floating hint disabled it's the `EditText`'s own hint that renders, and a default dark hint on the blue header would be hard to read.
- `tv_fragment_info` drops `layout_weight="1"` for `wrap_content`, so the title now claims its natural width before the search field gets the remainder. Fine in English; a longer localized title will squeeze the search box more than before.
- `layout_search_courses.xml` (65 lines) runs closely parallel to `layout_search.xml` (59 lines), so both now need maintaining together. The divergence is genuine (white-on-blue colours, hint handling, gone-by-default tags), so a second file is defensible — but a style or theme overlay on the existing include would have kept it to one.

⚠️ **Conflict:** #15446 rewrites both `fragment_my_course.xml` variants with a competing header design (its own search pill and chip row), and #15441 edits the same two files. All three touch this header. As with #15441 I've deliberately not marked anything `duplicate` — these are distinct jobs colliding on shared lines, not redundant takes — but the merge order needs to be chosen rather than discovered.

---

Okuro3499 (changes requested):
I think the current header placement reads better visually, compared to here where everything is squeezed it even hides the search 
https://github.com/user-attachments/assets/12e8de8d-e212-43aa-9985-07af2e6a3de3

PR #15525, opened 2026-08-10
dashboard: smoother responsive layout handling (fixes #15524)

### LINKED ISSUE #15524: Dashboard page is unresponsive and layout appears squeezed
## Summary
The dashboard page becomes unresponsive and the UI elements are squeezed together, making the page difficult to use.

## Steps to Reproduce
1. Open the app and navigate to the Dashboard page.
2. Observe the page layout and responsiveness.
https://github.com/user-attachments/assets/e80ec701-82dc-4e0a-8968-d86b75fba7a1

## Actual Behavior
- Dashboard does not respond properly.
- Content appears compressed/squeezed.
- Overall usability is significantly impacted.

## Expected Behavior
- Dashboard should remain responsive across supported screen sizes.
- UI elements should maintain proper spacing and alignment.

Okuro3499:
fixes #15524
[Screen_recording_20260810_195754.webm](https://github.com/user-attachments/assets/1fa36966-7eb9-4c80-a2ff-d1944612c7c0)

## Summary by CodeRabbit

- **New Features**
  - Added a redesigned dashboard with clearer Library, Courses, Teams, and My Life cards.
  - Added a “Sync now” action and relative last-sync status display.
  - Added responsive landscape and tablet layouts with profile, badges, and dashboard navigation.
- **Style**
  - Refined spacing, colors, badges, chips, typography, and dark-theme support.
- **Localization**
  - Added sync-status translations for Arabic, Spanish, French, Nepali, Somali, and Swahili.


J-S-webskas (commented):
Layout looks good.

dogi:
**Review — rating: 77/100**

46 files, +1194/−688, so I sampled: all four Kotlin files and the manifest read in full, and the resource layer traced by which tokens each new color/string actually feeds rather than line by line.

**The resource discipline here is excellent, and it's worth contrasting with #15446.** Exactly one new string, `dashboard_sync_status`, and it lands in the default plus all five locales *and* all six `-sw360dp` variants — twelve files, no English fallbacks. Dark mode is handled with a split that I checked rather than assumed: the six structural tokens (`dashboard_page_bg`, `dashboard_ink_1/2/3`, `dashboard_hairline`, `dashboard_sync_banner_bg`) get `values-night` overrides, and the eight left theme-invariant are legitimately invariant — `dashboard_tile_*` are saturated accents (`#D32F2F`, `#F57C00`, `#388E3C`, `#7B1FA2`) used as narrow stripes via `dashboard_tile_color_block_width`, not text backgrounds, so there's no light-tile/light-ink collision in night mode. Dimens are split across `values/`, `values-land/`, `values-sw360dp/` and `values-sw600dp/`.

The code cleanup is real too: `LinearLayout.LayoutParams(250, 100)` — hardcoded pixels, and the wrong `LayoutParams` type for a `FlexboxLayout` parent — becomes `FlexboxLayout.LayoutParams` built from dimension resources; the zebra-striping helpers (`setTextColor`, `setBackgroundColor`, `setTextViewProperties`) are deleted rather than left orphaned; `createCourseChip` inflates `ItemCourseHomeBinding` instead of hand-assembling a `TextView`. And `dashboard_sync_status` is a proper format string with a placeholder, replacing the `last_synced_colon + time` concatenation — the better pattern.

**The manifest change is correct but has a consequence that isn't addressed.** Dropping `orientation|screenSize` from `DashboardActivity`'s `configChanges` is *necessary*: with orientation declared, Android never re-inflates on rotation, so none of the new `layout-land` / `layout-sw600dp-land` / `layout-normal-land` variants this PR adds would ever be applied. So the change is deliberate and right. But it also means `DashboardActivity` is now **recreated on every rotation**, and nothing in this PR saves instance state.

That directly collides with **#15519**, which is currently sitting on `ready`. That PR adds `isSyncStatusDismissed` as a plain activity field with no `onSaveInstanceState`, and part of why I passed it was that `DashboardActivity` declared `configChanges="orientation|screenSize|keyboardHidden"` and therefore survived rotation. Once this PR lands, a dismissed sync banner comes back on every rotation. Whichever of the two merges second needs to move that flag into `onSaveInstanceState` or a ViewModel. I'm flagging it on #15519 as well so it doesn't get lost.

More broadly: this is the change most in need of hands-on rotation testing across the new variants — portrait/landscape, phone/sw600dp — since CI builds the layouts but never inflates them.

**Two things to confirm:**

1. **`lateinit var params` has an initialization-order dependency.** It's assigned inside `initView(view)`, and every `flexboxLayout.addView(…, params)` will throw `UninitializedPropertyAccessException` if it runs first. The `onResume` → `refreshMyLifeList` → `myLifeListInit` path is safe because `onViewCreated` precedes `onResume`, but this is a base class with subclasses calling `initView` themselves, so the invariant is implicit. `private val params by lazy { … }` removes the ordering question entirely.
2. **`createCourseChip` drops the old type branching.** `setTextViewProperties` switched on `ItemType.LIBRARY` / `COURSE` / `MEETUP`; the replacement always wires `TakeCourseFragment`, and the `EventsDetailFragment` import is gone. That looks safe because the only caller was the courses loop, so `MEETUP` was dead there — but please confirm no meetup chip reached that path.

**Nit:** `binding.dashboardSyncNow.setOnClickListener { logSyncInSharedPrefs() }` — worth double-checking that method actually initiates a sync rather than only recording one; the name reads like bookkeeping, and a "sync now" button that doesn't sync would be an unpleasant bug. Separately, `railSyncStatus ?: return` is a nice touch for the `card_profile_bell` variants that omit the view.

**Verdict:** needs another round — the redesign itself is well built and the localization/theming is the best in the queue; it's the rotation-recreation fallout that needs handling and coordination with #15519.

---

dogi:
**Review — rating: 84/100**

46 files, +1195/−688. I read all five Kotlin files in full and sampled the layout/resource hunks for consistency. There's real substance here beyond the visual work, and three things I expected to be problems turned out to be correct once checked.

`params` moving from `LinearLayout.LayoutParams(250, 100)` to `FlexboxLayout.LayoutParams` with `dashboard_chip_width`/`dashboard_chip_gap` is a genuine fix, not just tidying: the parent is a `FlexboxLayout`, so the old `LinearLayout.LayoutParams` was the wrong type *and* carried raw pixel values that ignored display density. Replacing `getString(last_synced_colon) + time` with the parameterised `getString(dashboard_sync_status, timeText)` is the right i18n move — concatenation can't reorder for RTL, and this is exactly the kind of string Arabic needs to control.

Localization of the new string is complete and I verified the whole matrix: `dashboard_sync_status` appears 12 times — `values/` plus five locales, and again across the six new `-sw360dp` buckets — with a deliberate short/long split (`Sync · %1$s` narrow, `Last sync · %1$s` wider) carried consistently through Arabic, Spanish, French, Nepali and Somali. Nothing falls back to English.

Two removals I checked before believing them. Deleting the `ItemType.MEETUP` and `ItemType.LIBRARY` branches from `setTextViewProperties` looks like dropped functionality, but it isn't — the only `DashboardItem` constructions in the codebase are `ItemType.COURSE` (`BaseDashboardFragment:243`) and `ItemType.LIFE` (`:332`), so both branches were already unreachable and `createCourseChip` loses nothing. And `last_synced_colon` is *not* orphaned by the switch to a format string; `SettingsActivity:336` still uses it. Dropping the `itemCnt % 2` zebra striping also lets the index parameter disappear from several signatures, which is a nice simplification.

**The riskiest line is the manifest change, and it needs to be called out explicitly:**

```diff
-android:configChanges="orientation|screenSize|keyboardHidden"
+android:configChanges="keyboardHidden"
```

This is *necessary* — without it Android never re-selects resources on rotation and none of the new `layout-land/`, `layout-sw600dp-land/`, `layout-*-land/` variants would ever be applied, so the PR's core goal depends on it. But it also converts the app's primary activity from handling rotation itself to being fully destroyed and recreated on every rotation, and `DashboardActivity` hosts essentially every dashboard fragment. That means each rotation now re-runs `onResume` work (`checkPendingSurveys()`, `updateRailSyncStatus()`) and rebuilds all child fragments, so any latent state that was previously preserved by accident — scroll positions, open dialogs, unsaved input, in-flight transactions — is now exposed. Worth deliberate rotation testing across the dashboard's child fragments, and worth a line in the PR description so it isn't discovered later as a mystery regression.

**Nits:**
- `R.drawable.light_rect` and `R.color.dashboard_item_alternative` are now referenced only from the code this PR deletes, so both become unused. Given #14662 and #14663 are open specifically to purge unused resources, deleting them here avoids adding to that pile. `ItemType.MEETUP`/`LIBRARY` are likewise now unused enum entries.
- Six new `strings.xml` buckets to hold one string means `dashboard_sync_status` lives in 12 files, so every future edit is a 12-file change. Correct today, but `autoSizeTextType` or a single wording would sidestep the proliferation.
- `SettingsActivity:336` still builds its label by concatenation; converting it to the same format-string pattern would finish the thought.
- `params` is one shared `LayoutParams` instance handed to every child — pre-existing, and safe while nothing mutates it, just worth knowing.

On the overlap with #15549: superficial, and I've marked neither as a duplicate. That PR only changes the values of `no_resources_added_yet` / `no_courses_joined_yet` / `no_teams_joined_yet`, which this PR never touches, and although you modify `BaseDashboardFragment.kt`, `renderPlaceholder` appears nowhere in this diff and all three of its call sites fall outside your hunks. The two are independent; worst case is a trivial textual merge in `strings.xml`.

**Verdict:** mergeable — CI green (build default + lite, test default). Please exercise rotation across the dashboard fragments before it lands, given the `configChanges` change, and ideally drop the two now-unused resources.

---

Okuro3499:
Screenshot_20260812_192622

dogi (approved):

PR #15526, opened 2026-08-10
actions: smoother workflow automerge drain cancelling (fixes #15561)

### LINKED ISSUE #15561: Add permissions and quota-aware automerge drain cancellation
The release workflow lacks explicit GitHub Actions permissions and doesn't handle Play Store daily quota exhaustion. When the Play Store rejects a release for exceeding the daily save quota, the automerge drain keeps running doomed upload build cycles. Add explicit `contents: write` / `actions: write` permissions, give the playstore fallback step an id, and add a quota-aware step that detects the `Daily save quota exceeded` error in the job log and cancels all pending automerge workflow runs, exiting safely when the error cannot be confirmed to avoid false positives.

dogi:
## Summary
Enhances the release workflow with explicit GitHub Actions permissions and adds intelligent handling for Play Store daily quota exhaustion. When the Play Store rejects a release due to exceeding daily save quota, the workflow now cancels pending automerge drain runs to prevent wasted build cycles on doomed uploads.

## Key Changes

- **Explicit permissions**: Added `contents: write` (for GitHub release artifacts) and `actions: write` (for reading job logs and cancelling runs) to the release workflow
- **Playstore fallback step ID**: Added `id: playstore_fallback` to the fallback upload step for conditional tracking
- **Quota-aware drain cancellation**: New step that:
  - Runs only when both primary and fallback Play Store uploads fail
  - Reads the job's archived log to detect the specific `Daily save quota exceeded` error
  - Cancels all pending automerge workflow runs when quota exhaustion is confirmed
  - Logs warnings and step summary for manual recovery (hand-upload AAB, then re-dispatch automerge)
  - Safely exits without cancelling if the error cannot be confirmed, preventing false positives

## Implementation Details

The new step uses the GitHub CLI to:
1. Locate the current job by name or runner name
2. Poll the archived job log (up to 50 seconds) for the upload action output
3. Match the exact error message to avoid cancelling on unrelated failures
4. Query and cancel all non-completed automerge workflow runs
5. Document the incident in the job summary for visibility

This prevents the automerge drain from spending the rest of the queue on builds that cannot publish until the Play Store quota resets at midnight PT.

https://claude.ai/code/session_01HFnVs3tyUKKdK88hidtKNn

## Summary by CodeRabbit

- **Release Improvements**
  - Improved release publishing reliability when daily Play Store upload limits are reached.
  - Releases now report quota-related conditions clearly and safely pause automated follow-up actions.
  - Other publishing failures continue to be handled without disruption.


dogi:
**Review — rating: 80/100**

Careful, fail-open automation. I traced the conditions and the scopes against the existing `release.yml`:

- **The `permissions` block is correct and is a net tightening.** `release.yml` previously declared no `permissions:`, so every step ran with the repo/org default token scope. Declaring it sets all unlisted scopes to `none`, so the question is whether the two listed cover what the workflow does — they do. Release creation goes through `dogi/upload-release-action@v2.9.0` with `repo_token: secrets.GITHUB_TOKEN` (`contents: write`), `actions/upload-artifact@v7` uses the runtime token rather than a `GITHUB_TOKEN` scope, and `setup-gradle` plus the Play Store and Discord steps need no GH scopes. `actions: write` is exactly what the new cancel/log-read needs. Getting there with the default token instead of a PAT is the right call.
- **Step conditions line up.** The new step gates on `steps.playstore.outcome == 'failure' && steps.playstore_fallback.outcome == 'failure'`, and `playstore_fallback`'s own `if:` is identical to the first two clauses — so whenever the primary upload fails on a non-dispatch `lite` build, the fallback has actually run and its `outcome` is a real `failure`/`success` rather than `skipped`. No dead branch.
- **The shell is written defensively.** Every `gh api` is `|| true`-guarded, the self-log read retries five times using `grep -qF 'upload-google-play'` as a readiness probe, and both the unreadable-log and wrong-error paths `exit 0` with a warning instead of failing the release. Under the default `bash -e` I don't see an unguarded command that could abort the step. `${DRAIN##*/}` correctly yields `automerge.yml`. `PLAYSTORE_RELEASE_NAME` is defined at `:84` as `code (name)`, so the interpolation into the message is well-formed.

**The caveat that matters most: none of this is exercised by the green checks on this PR.** `release.yml` only triggers on a `master` push or manual dispatch, so `build`/`test` passing says nothing about it — and the new step's trigger is narrower still, requiring a real release where Google's daily quota is exhausted *and* the fallback also fails. So three assumptions ship unverified: that `gh api .../actions/jobs/{id}/logs` follows its redirect and yields text, that a job can read its own log while still running, and that the error line is literally `Error: Daily save quota exceeded`. All three fail safe (the drain is simply left alone), but the feature could be silently inert for a long time before anyone notices. A `workflow_dispatch` smoke path, or at minimum a note on how this was hand-verified, would raise confidence a lot.

**Nit:** 45 lines of bash embedded in YAML runs against the convention this repo already established — `CLAUDE.md` notes the automerge logic lives in `.github/scripts/automerge.sh`. Moving this to `.github/scripts/` would make it shellcheck-able and testable in isolation, which is worth more here than usual precisely because CI can't reach it. Also, matching on a third-party action's error prose is inherently brittle; if `dogi/upload-google-play` ever rewords it, detection stops working silently — a `::warning::` on the "did not match" path would at least leave a breadcrumb.

**Verdict:** mergeable — correct scopes, sound conditions, and it degrades safely; the untestability and the inline script are worth a follow-up rather than a block.

---

Okuro3499 (approved):

PR #15527, opened 2026-08-10
life: smoother repository dashboard seeding (fixes #15521)

### LINKED ISSUE #15521: myLife Section Shows Duplicate Items
## Description
The myLife section displays duplicate/double items. Items that should appear only once are being rendered twice, causing a confusing and cluttered user experience.

## Steps to Reproduce
1. Open the myPlanet app
2. Navigate to the **myLife** section
3. Observe the list of items displayed

## Expected Behavior
Each item should appear only once in the myLife section.

## Actual Behavior
Items are duplicated — each item appears twice in the list.

https://github.com/user-attachments/assets/b7479626-e88d-406e-8fee-f40df6d766b5

Okuro3499:
fixes #15521
[Screen_recording_20260810_214902.webm](https://github.com/user-attachments/assets/0a9e2098-9375-42ea-af08-e351d203e6ad)


dogi:
**Review — rating: 86/100**

Well-diagnosed race. Three changes that each address a different facet, and I verified all three:

**The removed `myLifeListInit` call is the actual trigger, and removing it is safe.** `myLifeListInit` had two call sites in `BaseDashboardFragment`: `:318` inside `refreshMyLifeList()`, which `onResume()` invokes, and `:388` during flexbox setup. On first display both ran concurrently, so two coroutines raced into `getMyLifeForDashboard` → `seedMyLifeIfEmpty`. Worth noting the `:388` copy also lacked the `myLifeFlex.removeAllViews()` that `refreshMyLifeList()` does, so it was a double-render source as well as the race trigger. Since `onResume` fires after `onViewCreated`, population is still guaranteed — the feature doesn't lose its initialization.

**The `Mutex` is correctly scoped, which is the part that could easily have been wrong.** A per-instance `Mutex` only serializes anything if there's one instance, and `RepositoryModule:101-103` binds `LifeRepository` with `@Binds @Singleton`, so there is. That matters because two independent entry points remain — `LifeFragment.kt:95` calls `seedMyLifeIfEmpty` directly and `LifeRepositoryImpl:108` calls it internally from the cache-miss path — so the check-then-insert genuinely needed guarding beyond just deleting the duplicate invocation.

**The `dedupKey()` reorder is the right diagnosis.** Every seeded row gets a fresh `UUID.randomUUID()` as `_id`, so with `_id` first in the precedence chain the duplicates looked distinct and sailed through the `distinctBy { it.dedupKey() }` at `:112`. Moving `imageId → title → _id` collapses them. Nice side effect: users who already have duplicate rows get them hidden at read time immediately.

Which leads to the one real gap: **this prevents new duplicates and masks existing ones, but never deletes them.** Anyone already affected by #15521 keeps the duplicate rows in Room forever, correct-looking only because `distinctBy` filters them on read — and they'll still be counted by `countByUserId`, so their seeding stays permanently "non-empty". A one-time cleanup (delete rows sharing an `imageId`/`title` per user, keeping the lowest weight) would actually finish the fix for existing installs.

**Nit:** two smaller points. A Room `withTransaction` around the count-then-insert, or a unique index on `(userId, imageId)`, would make this robust at the storage layer rather than relying on the in-process lock — worth considering since the DAO is the real invariant holder. And the reorder means two legitimately distinct entries sharing an `imageId` (or a `title` when `imageId` is blank) now collapse to one; harmless for the fixed default set, but worth knowing if custom entries are ever addable.

`LifeRepositoryImplTest.kt` already covers `seedMyLifeIfEmpty` for the empty and non-empty cases and still passes, so no regression there. Adding two cases would lock in this PR specifically: two concurrent `async` calls against the singleton asserting a single `insertAll`, and a `dedupKey` precedence case.

**Verdict:** mergeable — the race fix is correct and correctly scoped; the leftover-duplicate cleanup is the follow-up worth filing.

---

dogi (approved):

PR #15528, opened 2026-08-10
resources: smoother guest all selecting (fixes #15514)

### LINKED ISSUE #15514: Hide "Select all" in Library for guest users
## Summary
Hide the **Select all** option in the Library for guest users.

## Problem
Guest users should not be presented with bulk-selection functionality in the Library. Showing **Select all** can be confusing and may expose actions that are not relevant or allowed for guest access.

## Proposed change
- Detect when the current user is a guest user
- Hide or disable the **Select all** control in the Library for guest users
- Keep the existing behavior unchanged for authenticated users with the appropriate permissions

Image

J-S-webskas:
Recording_20260810_124925

## Summary by CodeRabbit

* **Bug Fixes**
  * Guest users can no longer add resources to their library while selecting resources.
  * Guest users no longer see “Select All” when library resources are available.
  * The “Add to Library” control remains visible in empty states.
  * Authorized users can continue using available resource-selection controls.


dogi:
**Review — rating: 68/100**

You found the right bug, and one of the two changed lines is exactly the fix. `setupGuestUserRestrictions()` at `:239` already hides `selectAll` for guests with this same `userModel?.isGuest() == true` check — the problem was that the visibility update at `:400` later set `selectAll.visibility = View.VISIBLE` unconditionally and **overrode** it once the list loaded, which is why guests saw the checkbox reappear. Re-applying the guest check there is correct and necessary. Using `isGuest()` rather than the ad-hoc `id?.startsWith("guest")` that `RequestsAdapter` and `VoicesAdapter` still use is the right call too.

**But the same edit applied to the empty-state branch at `:391` is a regression.** That branch is the `!hasAnyLibraryData && listSize == 0` case, and it deliberately hid `selectAll` because there is nothing to select. It also hides `etSearch`, `tvAddToLib`, `tvSelected`, `filter`, `clearTags` and `tvDelete`. With this change a member opening an empty library now sees a lone "Select all" checkbox with every other control hidden and no rows beneath it. That line should stay `View.GONE` unconditionally — the guest check only belongs in the non-empty branch, since for the empty case the answer is "hidden for everyone".

**The `hideButton()` reorder is inert.** Moving it after `val allSelected = adapterLibrary.areAllSelected()` reads like an ordering fix, but `hideButton()` (`:374`) only reads `selectedItems?.size` and toggles `enabled`/`visibility` on `tvDelete` and `tvAddToLib` — it never mutates selection state, so `areAllSelected()` returns the same value either way. Harmless, but it adds diff noise to a two-line fix and implies a bug that isn't there. I'd drop it unless you saw concrete misbehavior, in which case the real cause is elsewhere and worth naming.

Worth considering while you're in here: the guest ternary is now duplicated in both branches, so the two can drift. Calling `setupGuestUserRestrictions()` at the end of the visibility update — or extracting `applySelectAllVisibility(hasData: Boolean)` — keeps the rule in one place, which is what let the original override slip through.

**Verdict:** needs another round — revert `:391` to `View.GONE`, keep the `:400` fix, and drop the inert reorder. Then it's a clean two-line change.

---

dogi:
**Re-review — Rating: 76/100** (was 68)

All three asks from the last round are done, and one of them you took further than I suggested. Build (default + lite) and test (default) green.

- The empty-state branch is back to `View.GONE` unconditionally, so a member opening an empty library no longer gets a lone "Select all" with every other control hidden. ✅
- The inert `hideButton()` reorder is gone, so the diff is back to the actual change. ✅
- The `:400` guest check is kept, which is the real fix for #15514. ✅
- On consolidation, you went past the suggestion and deleted `setupGuestUserRestrictions()` outright, so `checkList()` is now the single owner of `selectAll`'s visibility. That's better than the two-places-in-sync arrangement I proposed — the duplicate source is what let the original override slip through in the first place.

**The new `tvAddToLib` line reintroduces the same class of bug, in the other direction.** `hideButton()` already owns that button, and it keys off selection count:

```kotlin
val count = selectedItems?.size ?: 0
tvAddToLib.isEnabled = count != 0
if (count != 0) { … else tvAddToLib.visibility = View.VISIBLE }
else            { … else tvAddToLib.visibility = View.GONE }
```

With nothing selected, "Add to library" is meant to be hidden — there is nothing to add. But `checkList()` now unconditionally sets it visible for any non-guest, and before this PR `checkList()`'s non-empty branch didn't touch `tvAddToLib` at all, so `hideButton()`'s `GONE` stood. Any data-driven `checkList()` (initial load, refresh, cleared selection) now leaves a member looking at an "Add to library" button with zero rows selected — visible, and greyed out from whatever `isEnabled` was last set to. That's structurally the same failure as #15514: one method writes a visibility that another method later overrides.

The fix is to put the guest rule where the button already lives rather than adding a second writer:

```kotlin
private fun hideButton() {
    val count = selectedItems?.size ?: 0
    val isGuest = userModel?.isGuest() == true
    tvDelete?.isEnabled = count != 0
    tvAddToLib.isEnabled = count != 0
    if (count != 0 && !isGuest) { … VISIBLE } else { … GONE }
}
```

…and drop the `tvAddToLib` line from `checkList()`. That gives you one owner per view, which is what the commit message is aiming for.

**One thing to double-check: the timing guarantee got weaker.** `setupGuestUserRestrictions()` used to be called immediately after `userModel = userRepository.getUserModel()` inside the coroutine, so the guest rule was guaranteed to apply as soon as the user resolved. With that call deleted, gating happens only inside `checkList()` — so it now depends on `checkList()` running *after* `userModel` is populated. If any `checkList()` path fires first (cached or empty data during `onViewCreated`), `userModel` is null, `userModel?.isGuest() == true` evaluates false, and a guest sees `selectAll` until the next invocation. Worth confirming a post-load `checkList()` always happens; if not, calling the visibility update at the end of that coroutine restores the guarantee cheaply.

**Nits:** `userModel?.isGuest() == true` is evaluated twice on consecutive lines — a single `val isGuest` reads better and can't drift. And `0898815`/`0f97000` are two commits three minutes apart with identical messages; worth squashing before merge.

No conflict with #15523 despite both touching select-all — that one is the courses layouts, this is `ResourcesFragment.kt`.

**Verdict:** needs another round, but a short one — move the guest check into `hideButton()` and remove the `tvAddToLib` line from `checkList()`. The `selectAll` half of this is correct and done.

---

dogi:
**Review — rating: 40/100**

The intent is right — `selectAll` genuinely should be hidden for guests, and folding the check into `checkList()` is a better home for it than a one-shot `setupGuestUserRestrictions()` that ran once and never re-applied. But the implementation crashes the Resources screen.

`val guest = userModel?.isGuest()` is a `Boolean?`, and `hideButton()` dereferences it with `!guest!!`. `userModel` is declared `var userModel: UserEntity? = null` and is only assigned inside `lifecycleScope.launch { userModel = userRepository.getUserModel() }` — a suspend call that suspends. `hideButton()` is called unconditionally from `onViewCreated` (`ResourcesFragment.kt:181`) *before* that launch block, so on first entry to the fragment `userModel` is still `null`, `guest` is `null`, and `guest!!` throws. That's a crash every time the Resources/Library screen is opened fresh, not an edge case. CI didn't catch it because the repo runs `testDefaultDebugUnitTest` only and has no instrumented tests.

The fix is a one-liner — `if (count != 0 && userModel?.isGuest() != true)` — and it's worth noting the code you deleted was already null-safe (`userModel?.isGuest() == true`); this replaced a safe check with `!!`, which is the pattern `docs/CODE_STYLE_GUIDE.md` steers away from precisely because of this failure mode. Since `hideButton()` can run before the user loads, the guest state also needs re-applying once `getUserModel()` returns, otherwise a guest briefly gets the member UI.

Second, smaller regression: dropping `tvAddToLib.visibility = View.GONE` from the empty branch of `checkList()` isn't fully covered by `hideButton()`. `hideButton()` does hide it when `count == 0`, but the filtering path calls `checkList(filteredList.size)` without calling `hideButton()`, so filtering down to an empty result while items are still selected leaves "add to library" visible above an empty list.

**Verdict:** needs another round — blocking NPE on the Resources screen.

For the record, this is not a duplicate of #15523 despite the similar "select all" phrasing in both titles: that one is the courses search header, this one is `ui/resources/ResourcesFragment.kt`. No path overlap, so both can proceed independently.

---

Okuro3499 (approved):

dogi:
**Re-review — Rating: 78/100** (was 40)

The blocking NPE is genuinely fixed. `hideButton()` now reads `if(count != 0 && userModel?.isGuest() != true)` and `checkList()` derives `val isGuest = userModel?.isGuest() == true` — both null-safe, no `!!` anywhere, so opening Resources before `getUserModel()` resolves no longer throws. You also fixed the follow-on I raised: the `lifecycleScope.launch` now re-applies the guest state once the user loads, guarded by `::adapterLibrary.isInitialized && _binding != null`, so a guest no longer gets a flash of member UI. And the second regression is addressed — `hideButton()` is now paired with every `checkList(...)` call, including the filtering path at `checkList(filteredList.size)`, so filtering to an empty result with items selected no longer strands "add to library" above an empty list. CI green (build default + lite, test default).

**Two things stand between this and merge, neither a defect in your logic.**

**The branch is behind master and the divergence lands on the exact method you edited.** Your diff still carries `checkList(listSize: Int = if (::adapterLibrary.isInitialized) adapterLibrary.getLibraryList().size else 0)` as context, but on master that default is now `adapterLibrary.currentList.size` — `getLibraryList()` no longer exists on `ResourcesAdapter`. Master's `ResourcesFragment` is also roughly 70 lines longer than the version you branched from, with `hideButton()` at 440 and `checkList()` at 453. This needs a rebase, and it won't be a no-op one.

**That rebase is where the new coupling gets fragile.** Removing `tvAddToLib.visibility = View.GONE` from `checkList()`'s empty branch means correctness now depends on every `checkList()` call being manually followed by `hideButton()`. On your base you paired all of them. On current master the call sites are `checkList()` at 165, 228 and 395 and `hideButton()` at 180, 409, 564, 822 and 842 — the one at 228 has no adjacent `hideButton()`, so after rebasing there will be a path that hides nothing. Rather than re-pairing by hand, have `checkList()` end by calling `hideButton()` itself. One invariant in one place, and it can't drift as the file keeps growing.

⚠️ **Overlaps two other open PRs in this file.** #15591 (`ready`) rewrites the same `else` branch of `checkList()` — it changes `etSearch.visibility` to `layoutSearch.visibility` and adds a view-toggle line, right where you're adding the `isGuest` conditional on `selectAll`. #15267 (`change`) also rewrites both `checkList()` and `hideButton()`. Whichever lands first, the others need real conflict resolution, not an auto-merge. Worth agreeing an order.

**Nit:** the launch block uses `lifecycleScope` rather than `viewLifecycleOwner.lifecycleScope` while touching `_binding` and views. The `_binding != null` guard covers it, so this isn't a live bug, and the scope choice predates your change — but `viewLifecycleOwner.lifecycleScope` is what this wants.

**Verdict:** needs another round — fold `hideButton()` into `checkList()` and rebase onto master. The guest-visibility logic itself is correct now.

---

dogi:
**Re-review — Rating: 84/100** (was 78)

Both things I asked for are done, and done the right way.

**`hideButton()` is now folded into `checkList()`** rather than paired at each call site. That's the durable fix — the invariant lives in one place, the four hand-maintained pairings are gone, and the empty-branch `tvAddToLib.visibility = View.GONE` can safely be dropped because `hideButton()` is now guaranteed to run and owns that decision. New `checkList()` call sites can't forget it.

**The branch is rebased.** The default parameter now reads `adapterLibrary.currentList.size`, matching master, so the stale `getLibraryList()` reference from the previous round is resolved and this applies cleanly.

The guest logic itself is unchanged from what I approved last round and still null-safe throughout — `userModel?.isGuest() != true` in `hideButton()`, `userModel?.isGuest() == true` in `checkList()`, no `!!` anywhere, and the re-apply after `getUserModel()` resolves is still guarded by `::adapterLibrary.isInitialized && _binding != null`.

Dropping `tvAddToLib.isEnabled` from `changeButtonStatus()` is also correct — `hideButton()` already sets it from the same `count`, so that was a duplicated assignment with two owners. Worth a quick check that every `changeButtonStatus()` call site is reached by `hideButton()` or `checkList()`; if one isn't, the enabled state would go stale there.

**CI note:** `myPlanet build (default)` and `(lite)` are green, but `myPlanet test (default)` was still `in_progress` when I looked, so I can't call the run complete. Worth confirming it lands green before merging.

⚠️ **The conflicts from last round still apply.** #15591 (`ready`) rewrites the same `else` branch of `checkList()` — it swaps `etSearch.visibility` for `layoutSearch.visibility` and adds a view-toggle line, right where you set `selectAll.visibility` from `isGuest`. #15267 (`change`) rewrites both `checkList()` and `hideButton()` wholesale. Whichever lands first, the others need real conflict resolution.

**Verdict:** mergeable, pending the test job going green.

---

dogi (approved):

PR #15529, opened 2026-08-10
login: smoother user repository dao querying (fixes #15791)

### LINKED ISSUE #15791: Refactor: push down userDao.getAll() filtering to Room queries
Replaces inefficient `userDao.getAll()` fetching and memory filtering with specific Room DAO queries in `UserRepositoryImpl`. Fixes N+1 and cartesian product issues during data sync.

dogi:
Replaces inefficient `userDao.getAll()` fetching and memory filtering with specific Room DAO queries in `UserRepositoryImpl`. Fixes N+1 and cartesian product issues during data sync.

---
*PR created automatically by Jules for task [1313930823151028484](https://jules.google.com/task/1313930823151028484) started by @dogi*

## Summary by CodeRabbit

  * Improved user retrieval and synchronization efficiency, especially for large datasets.
  * Reduced unnecessary loading and processing when handling user records.
  * Improved guest-user migration, health-record mapping, duplicate cleanup, and pending-sync processing.


dogi:
**Review — rating: 66/100**

The direction is right and matches what `CLAUDE.md` asks for — filtering belongs in the DAO, not in Kotlin after loading the whole table. The best part is the sync path at `:1247`: replacing `existingUsersList = userDao.getAll()` with targeted fetches keyed off the ids and names actually present in `documentList` is exactly the correct reasoning, since the maps built below are only ever probed by those ids/names. Pushing the grouping into SQL for `getDuplicateUsers` is a good win too, and making `users` nullable in `buildUserFromJson`/`migrateGuestUser` so the batch path keeps its in-memory list while single lookups go targeted is clean.

But translating Kotlin predicates into SQL drifted in several places, and none of it is covered by a test.

**1. `LIKE 'guest_%'` — the underscore is a wildcard.** In `getGuestUserByName` and `getGuestUsersByNames`, `_` matches any single character, so the pattern means "guest, any char, anything" rather than the literal `guest_` prefix that `_id?.startsWith("guest_")` checked. It doesn't misfire today — guest ids are built as `"guest_$username"` at `:158`, so the intended rows still match — but it would also match `guestX…` or `guest-…`. Escape it: `LIKE 'guest\_%' ESCAPE '\'`, or `substr(_id, 1, 6) = 'guest_'`.

**2. `LIKE` is case-insensitive; `startsWith` was not.** `getSyncedUsers` uses `id NOT LIKE 'guest%'` where the old code was `!it.id.startsWith("guest")`. An id of `GUEST123` was previously *included* and is now *excluded*. Same asymmetry in the two guest queries. Low practical risk given ids are generated lowercase, but it's a silent semantic change; `substr`/`GLOB` avoid it.

**3. Chunking can now return the same row twice.** `getUsersByIds` chunks at 500 and appends, but `getUsersByAnyIds` matches `id IN (...) OR _id IN (...)` — so a user whose `id` falls in one chunk and whose `_id` falls in another is added once per chunk. The old single pass over `getAll()` could never duplicate. Needs a `.distinctBy { it.id }` on the accumulated result. The same concatenation pattern at `:1247` merges id-matches and name-matches, which can also double up (harmless there only because the results feed into maps).

**4. `getDuplicateUsers` drops NULL-name duplicates.** `name` is `String?` (`UserEntity.kt:25`), and `name IN (SELECT name …)` never matches NULL, so users with a NULL name are excluded — whereas the old `allUsers.groupBy { it.name }` grouped them together and `cleanupDuplicateUsers` would have deduped them. Narrow, but it's a cleanup routine quietly doing less than before.

**5. Eight new queries, zero tests.** `app/src/test/java/org/ole/planet/myplanet/data/room/dao/` already exists and `room-testing` is a declared dependency, so DAO-level tests are the established pattern here. That matters more than usual on this PR precisely because the findings above are all "the SQL doesn't quite mean what the Kotlin meant" — which is exactly what a handful of DAO tests would have caught. I also can't confirm the unit suite: `myPlanet test (default)` was still in progress when I reviewed, with only the two builds green.

**Nit:** `TRIM(_id) != ''` trims spaces only, while Kotlin `isNullOrBlank()` treats tabs/newlines as blank — immaterial in practice, just noting the pair aren't identical. `getPendingSyncUsers` has a `LIMIT` with no `ORDER BY`, so which rows you get is formally unspecified (matching the old unordered `getAll().take(limit)`, so not a regression — but an explicit `ORDER BY` would make it deterministic). And `UserDao.kt` picks up a stray blank line before the closing brace and still has no trailing newline.

**Verdict:** needs another round — fix the `LIKE` escaping, dedupe the chunked results, and add DAO tests for the new queries. The refactor is worth landing once the SQL provably matches the predicates it replaced.

---

dogi:
**Review — rating: 72/100**

The SQL here is careful work, and a couple of the choices are subtler than they look. But the refactor the title promises isn't actually wired up, and that's what holds it back.

**The pushdown doesn't happen.** The diff adds eight DAO queries and a test file — and changes no call site. `UserRepositoryImpl` still calls `userDao.getAll()` in roughly ten places and filters in Kotlin, including the exact predicate this PR reproduces in SQL:

```kotlin
// UserRepositoryImpl.kt:129 — unchanged by this PR
override suspend fun getSyncedUsers(): List<UserEntity> {
    return userDao.getAll()
        .filter { !it._id.isNullOrBlank() && !it.id.startsWith("guest") }
        .map { it }
}
```

So no query is pushed down, no full-table load is avoided, and the branch adds eight DAO methods with zero production callers. Four of them — `getUsersByAnyIds`, `getGuestUserByName`, `getGuestUsersByNames`, `getDuplicateUsers` — have no counterpart anywhere in `UserRepository`, so they're speculative rather than pending. That matters in a repo where #14662 and #14663 are open specifically to purge unused code. Either land the wiring here (`getSyncedUsers`, `getUsersForHealthSync`, `getPendingSyncUsers` and `getUsersByIds` all have obvious repository counterparts ready to swap), or retitle this as "add DAO queries" and say what's calling them next.

**What's genuinely good**, and worth keeping intact when you do wire it:

- The translations are faithful to the Kotlin rather than approximations. `!it.id.startsWith("guest")` becomes `SUBSTR(id, 1, 5) != 'guest'`, which preserves Kotlin's case-sensitivity. The obvious-looking `NOT LIKE 'guest%'` would *not* have — SQLite's `LIKE` is case-insensitive for ASCII, so it would have silently started excluding `GUEST123`. Your test asserts that row is included, which documents the decision instead of leaving it to be rediscovered.
- `SUBSTR(_id, 1, 6) = 'guest_'` dodges a real trap: `LIKE 'guest_%'` treats `_` as a single-character wildcard and would also match `guestX…`. Using `SUBSTR` with `=` is correct here.
- `getDuplicateUsers` handles the NULL case properly. `IFNULL(name, '')` on both sides is what makes the outer `IN` match NULL-named rows at all, since `IN` never matches a bare `NULL` — easy to get wrong, and the test covers it.
- `name = :name` correctly uses `=` rather than `IS`, matching `CLAUDE.md`'s rule since the parameter is non-null.
- A real in-memory Room test with six cases covering blank/NULL `_id`, case sensitivity, and NULL grouping. CI runs it — `testDefaultDebugUnitTest` is green.

**Problems to fix alongside the wiring:**

- `TRIM(_id) != ''` is not quite `!isNullOrBlank()`. Kotlin's `isBlank()` treats any Unicode whitespace as blank, while SQLite's single-argument `TRIM` strips only spaces — so an `_id` of `"\t"` is excluded by the Kotlin filter but included by the query. Narrow, but faithfulness is this PR's whole value proposition.
- `getPendingSyncUsers(limit)` applies `LIMIT` with no `ORDER BY`, so which rows come back is unspecified. For a repeated sync batch that risks some rows never being picked. `ORDER BY id` makes it deterministic.
- Guest exclusion now has two idioms in the codebase: `SUBSTR(id, 1, 5) != 'guest'` here versus `_id NOT LIKE 'guest%'` in `AchievementDao`. Those differ on case, so it's worth settling on one deliberately.

**Nits:** the test uses `runBlocking` where `docs/TESTING.md` calls for `runTest` with `MainDispatcherRule` (no virtual time needed here, so it works — just off-convention), `import org.junit.Assert.*` is a wildcard import, and `UserDao.kt` still has no trailing newline.

**Verdict:** needs another round — the queries are well built and well tested, but nothing calls them, so the refactor in the title hasn't landed. Wire the four with existing counterparts and this becomes a straightforward approve.

---

dogi:
**Review — rating: 74/100**

This is the right refactor and it's mostly done well — replacing eight `getAll()`-then-filter-in-Kotlin paths with real Room queries is exactly the DAO discipline `CLAUDE.md` asks for, `cleanupDuplicateUsers()` going from "load every user" to a `HAVING COUNT(*) > 1` query is a large win, and making `buildUserFromJson`/`migrateGuestUser` fall back to targeted lookups instead of a full scan is the right shape. The new `UserDaoTest` exercising the real SQL against an in-memory database is what makes this reviewable at all, and it's pointed at the right edge cases.

I walked each translation for semantic equivalence and they hold up:

- `SUBSTR(id, 1, 5) != 'guest'` matches Kotlin's case-sensitive `startsWith`, and the test asserting `GUEST123` is *included* documents that deliberately rather than by accident. Short ids behave identically under both.
- `SUBSTR(_id, 1, 6) = 'guest_'` yields `NULL` for a null `_id`, so those rows drop out — same as `_id?.startsWith("guest_") == true`.
- `name = :name` is correct here rather than `IS` since the parameter is non-null, and it correctly refuses to match `NULL` rows, matching the old `it.name == userName`.
- `getDuplicateUsers()` groups by `IFNULL(name, '')`, which merges null-named and empty-string-named users into one SQL group where the Kotlin `groupBy { it.name }` keeps them separate — but the surviving `if (users.size > 1)` guard means a null/empty pair can never be mistaken for duplicates of each other. Safe, just slightly over-fetching.

The blocking issue is the chunk size. `getUsersByAnyIds` binds `:userIds` **twice** (`id IN (:userIds) OR _id IN (:userIds)`), and Room expands the list independently at each occurrence, so a `chunked(500)` batch binds 1000 variables. `SQLITE_MAX_VARIABLE_NUMBER` is 999 on the SQLite shipped with Android 11 and below, and `minSdk` here is 26 — so any full chunk raises `SQLiteException: too many SQL variables` on those devices. Both call sites are the ones most likely to exceed 500: the bulk user-sync path and `getUsersByIds`. Given myPlanet targets older low-end hardware with offline community-sized user tables, this will surface in the field rather than in testing. Either drop those chunks to 250, or restructure to a single bound list — `WHERE id IN (:userIds) UNION SELECT * FROM users WHERE _id IN (:userIds)` — and keep 500. `getUsersByIds`/`getGuestUsersByNames` bind once, so 500 is fine for them.

**Nit:** `getPendingSyncUsers` applies `LIMIT` with no `ORDER BY`; SQLite will generally return rowid order so this matches the old `take(limit)`, but an explicit ordering would make the batching guaranteed rather than incidental. `UserDao.kt` is also still missing its trailing newline.

**Verdict:** needs another round — one chunk-size fix and it's mergeable; the rest of the refactor is sound and well covered.

---

dogi:
**Follow-up to my last review — the blocker is cleared.** Merged `master` in and closed out the remaining items; all three checks are green on `8e9112d`.

**Merge conflict.** `master` deleted `searchUsers()` and `getHealthRecordsAndAssociatedUsers()` from `UserRepository` as unused code while this branch had edited both — I took the deletions.

One thing worth flagging, because taking `master`'s side verbatim would have quietly undone part of this PR: `master` didn't only delete the health-record method, it *moved* it to `HealthRepositoryImpl.getPatientHealthRecords()`, and that copy still does `userDao.getAll().filter { it.id in userIdSet }` — the exact pattern this PR exists to remove. Left alone, the merge would have reintroduced a full-table scan and left `UserDao.getUsersByIds` with no production caller at all. So the chunked pushdown now lives there instead.

**Review items:**

- **Chunk size (was blocking)** — resolved in `3f15f51`. Both `getUsersByAnyIds` call sites are at 400, so the doubly-bound `:userIds` expands to 800 variables, safely under the 999 `SQLITE_MAX_VARIABLE_NUMBER` on Android 11 and below. `getUsersByIds`/`getGuestUsersByNames` bind once and stay at 500.
- **`getPendingSyncUsers` `LIMIT` with no `ORDER BY`** — added `ORDER BY id`, plus a DAO test asserting the batch is deterministic rather than relying on incidental rowid order.
- **Trailing newline on `UserDao.kt`** — fixed in `3f15f51`.

I left the `TRIM(_id) != ''` vs Kotlin `isBlank()` difference alone — it only diverges on tab/newline-only `_id` values, which the sync path doesn't produce.

That clears everything from my last round. Verified against CI rather than locally: this environment has no Android SDK, so `./gradlew` can't run here — `myPlanet build (default)`, `myPlanet build (lite)`, and `myPlanet test (default)` all pass on the merge commit.

---

dogi:
**Review — rating: 88/100**

Exactly the right refactor: eight call sites stop materialising the entire `users` table into Kotlin objects just to filter it. I checked each new query against the predicate it replaces rather than trusting the titles, and all eight are faithful — including the two easy ones to get wrong. `SUBSTR(id, 1, 5) != 'guest'` preserves Kotlin's case-sensitive `startsWith("guest")` (SQLite `!=` on TEXT is case-sensitive absent `COLLATE NOCASE`, and `getSyncedUsers_excludesGuestsAndNullOrBlank_Id` pins that with the `GUEST123` case). `getDuplicateUsers` groups on `IFNULL(name, '')` so NULL-named rows count as duplicates of each other, and because `cleanupDuplicateUsers` then re-groups on the raw `name` and re-checks `size > 1`, the NULL/`''` conflation can only over-fetch, never cause a false delete — I traced that path specifically.

The chunk sizes are the detail I most expected to be wrong, and they're right. `getUsersByAnyIds` binds `:userIds` **twice** (`id IN (…) OR _id IN (…)`), so Room expands it to 800 host parameters at `chunked(400)` — under SQLite's 999 default — while the single-binding queries correctly use 500. That asymmetry looks like an inconsistency and isn't; worth a comment in the code so nobody "tidies" the 400 up to 500 later. `distinctBy { it.id }` after chunked accumulation is also necessary and present, since a user can match by `id` in one chunk and `_id` in another.

I verified the `bulkInsert` prefetch is actually complete rather than merely plausible: `usersById` and `guestUsersByName` are only ever keyed by `id`/`userName` drawn from `documentList` (lines 1191-1198), which is precisely what `idsToFetch`/`namesToFetch` collect — so narrowing from `getAll()` loses no lookup. Same for `buildUserFromJson`, where `getUserByAnyId` resolves to `WHERE id = :id OR _id = :id LIMIT 1`, matching the old `firstOrNull { it.id == id || it._id == id }`. Index coverage backs the hot paths too: `id` is the PK and `_id`/`name` are both indexed, so the `IN` and name lookups are index-driven. `UserDaoTest` is real coverage — in-memory Room via Robolectric, matching the project's test-source-set convention — and CI is green (build default + lite, test default).

Three things worth noting, none blocking:

- **`getPendingSyncUsers` gains `ORDER BY id`.** The old `.take(limit)` used `getAll()`'s arbitrary order. Deterministic is better, but it's an undeclared semantic change: the same lowest-id pending users now win the limit window every call.
- **`TRIM()` is not `isBlank()`.** SQLite's one-argument `TRIM` strips spaces only, while Kotlin's `isNullOrBlank()` covers all whitespace. An `_id` of `"\t"` would now be treated as synced. Contrived for CouchDB ids, but it is a real divergence in `getSyncedUsers`/`getUsersForHealthSync`/`getPendingSyncUsers`.
- **`getDuplicateUsers` can't use the `name` index** — wrapping the column in `IFNULL` makes it non-sargable, so it's a full scan plus a temp b-tree for the GROUP BY. Still far cheaper than the old full materialisation, so this is a note, not an objection.

**Nit:** the chunk-accumulate-distinct pattern is spelled out three times (`getUsersByIds`, health, `bulkInsert`); one private helper would carry the 400-vs-500 rationale in a single place.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #15547, opened 2026-08-11
enterprises: smoother finances date picking (fixes #15518)

### LINKED ISSUE #15518: Community Finances: Date Picker Does Not Preselect Last Selected Date on Reopen
## Description

In the Community Finances screen, when filtering transactions by date, the date picker resets to the current date each time it is reopened after an initial date has been selected. It should instead preselect the last chosen date to provide a smoother and more intuitive experience when updating the date filter.

## Steps to Reproduce
1. Navigate to **Community Finances**.
2. Use the date filter and open the date picker.
3. Select a date (other than today).
4. Close the date picker.
5. Reopen the date picker.

https://github.com/user-attachments/assets/c21cd6ce-071a-463c-8bf7-2cd3587a7ebe

## Expected Behavior
The date picker should preselect the previously selected date when reopened, allowing users to easily adjust or confirm the date without having to navigate back to it.

ragilzakaria:
https://github.com/user-attachments/assets/c69b77b1-e255-4f06-8021-05dc25cacd54

fixes #15518 

## Summary by CodeRabbit

* **Bug Fixes**
  * Date pickers now open with the currently selected from and to dates.
  * The to-date picker falls back to an appropriate date when no to date is set.
  * The calendar displays the correct year, month, and day when opened.


Okuro3499 (approved):

dogi:
**Review — rating: 85/100**

Small, well-targeted fix: the date picker now opens on the previously chosen date instead of today. The precedence is thoughtfully chosen — for the *to* date it falls back to the **from** date rather than today when no end date has been picked yet, which is genuinely the more useful starting point for a range.

I verified the two ways this could misbehave. `parseDate` is `private fun parseDate(dateString: String): Calendar?` and catches `ParseException`, returning null, so every `parseDate(text) ?: now` is safe against malformed or empty text — no crash path from stale or hand-edited field contents. And the new initial date can't fight the existing `minDate` constraint: the block below the dialog sets `datePicker.minDate` to the from date when picking a to date, and in every path `initialDate` is either the previously chosen to date (which was itself constrained to ≥ the from date) or the from date itself, so initial ≥ min holds. The views read here (`tvFromDateCalendar` for the from date, `etToDate` for the to date) are the same ones the surrounding code already uses, so the TextView/EditText asymmetry is pre-existing convention, not a mix-up.

**Nits, all cleanup:**
- The from-date text is now read and parsed up to twice per invocation — once for `initialDate`, once again in the `if (!isFromDate)` block for `minDate` — and the same `isNotEmpty` + `parseDate` expression is written three times across the method. A single `val fromDate: Calendar? = parseDate(binding.tvFromDateCalendar.text.toString())` hoisted to the top would collapse all three.
- `if (fromDateText.isNotEmpty()) parseDate(fromDateText) ?: now else now` is redundant: `parseDate("")` already returns null because `SimpleDateFormat.parse("")` throws `ParseException`, so this reduces to `parseDate(fromDateText) ?: now`. Same for the to-date branch.
- One benign edge: if a user moves the from date *later* than an already-selected to date, reopening the to-date picker hands it an `initialDate` below `minDate`. `DatePickerDialog` clamps to `minDate` rather than throwing, so the behaviour is fine — just worth knowing that's the path, and it hints at a separate question (whether changing the from date should clear a now-invalid to date) that's outside this PR.

---

dogi (approved):

PR #15549, opened 2026-08-11
dashboard: smoother placeholder wording (fixes #15522)

### LINKED ISSUE #15522: Update dashboard placeholders to be more positive
Update the dashboard placeholder copy to sound more positive, welcoming, and encouraging.

## Current behavior
Some dashboard placeholders read as neutral 

Image

## Desired behavior
Replace placeholder text with friendlier, more optimistic wording that better matches the product tone.

Example "You can add resources" instead of "You haven't added any resources yet"

ragilzakaria:
Screenshot_20260811-190648

fixes #15522


Okuro3499 (approved):

dogi:
**Review — rating: 92/100**

Reframes three dashboard empty-state placeholders from negative to invitational. Small, focused, and complete — no code, no layouts, no key renames, so the risk surface is just the copy itself.

I checked the things that actually go wrong in string PRs. Locale coverage is complete: master has exactly six `strings.xml` buckets (`values/` plus `ar`, `es`, `fr`, `ne`, `so`), all six define these three keys, and all six are updated here — nothing left showing the old wording in a single locale. The keys are live, not vestigial: each is rendered by `BaseDashboardFragment.renderPlaceholder` at lines 189, 232 and 255. Escaping is right too — the old English and French values carried `\'` for `haven\'t` and `n\'avez`, and the replacements contain no apostrophes, so dropping the escapes is correct rather than an oversight. The five translations all read idiomatically and match the English intent ("Puedes unirte a cursos", "Vous pouvez rejoindre des cours", "Waxaad ku biiri kartaa koorsooyin").

On the dashboard-area overlap with #15525: I checked, and it's superficial. #15525 touches `values/strings.xml` and the same five locale files but never these three keys, and although it modifies `BaseDashboardFragment.kt`, `renderPlaceholder` appears nowhere in its diff and all three call sites sit outside its hunks. So these two are independent and neither supersedes the other — no `duplicate` here, and your strings stay live after that redesign lands. Worst case is a trivial textual merge in `strings.xml`.

**Nit:** the key names are now stale — `no_resources_added_yet` / `no_courses_joined_yet` / `no_teams_joined_yet` describe the old negative phrasing, while the values no longer say "yet" or "no". Renaming would churn the three call sites for no functional gain, so leaving them is the pragmatic call, but the next reader will find the names slightly misleading.

One copy thought, take it or leave it: the new strings say what the user *can* do but no longer say what the current state *is*. For an empty state that's usually the right trade and it's exactly what #15522 asked for, so I wouldn't hold this up — just noting that something like "No courses yet — you can join courses" keeps both halves.

---

dogi (approved):

PR #15550, opened 2026-08-11
teams: smoother voices repository replying (fixes #15517)

### LINKED ISSUE #15517: Replies in community not being added
## Issue
Replies to community posts are not being successfully added. When users attempt to respond to posts in the community, the replies fail to be stored or displayed.

https://github.com/user-attachments/assets/0742ef0b-22a5-420e-bb8f-4c345a65a6d5

## Expected Behavior
- Users should be able to add replies to community posts
- Replies should persist in the database
- Replies should be displayed immediately after creation

## Steps to Reproduce
1. Navigate to a community post
2. Attempt to add a reply
3. Observe that the reply does not appear

Okuro3499:
fixes #15517
[Screen_recording_20260811_171619.webm](https://github.com/user-attachments/assets/fb7662b6-7627-4c40-9cff-a6b2ec5712ea)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reply handling so replies correctly reference their parent posts.
  * Ensured the current session user is reflected in the Voices interface when available.

* **Tests**
  * Added coverage verifying correct parent references are stored when posting replies.


dogi:
**Review — rating: 86/100**

The one-line change in `postReply` is the whole fix, and I traced both ends of it rather than taking the diff at face value.

`replyTo` is looked up locally by the parent's **local** id: `VoicesAdapter.updateReplyCount` does `val newsId = news?.id ?: return` (line 692) and hands that to `getReplyCount`, which lands on `NewsDao`'s `WHERE replyTo = :newsId COLLATE NOCASE`. So the old `news._id ?: news.id` wrote the *server* doc id into a column that's queried by *local* id — replies simply wouldn't be counted or listed whenever those two differed. Switching to `news.id` makes the write agree with the read. That's the right fix, in the right place.

It's also a narrower change than it first looks, which is reassuring. Sync-down constructs news as `News().apply { id = underscoreId }` and then sets `_id = underscoreId`, so for every server-originated post `id == _id` and the old and new expressions are identical. For a locally-created post that hasn't uploaded yet, `_id` is null and the old code already fell back to `id`. The only window where behaviour actually changes is a locally-created post that has since been assigned an `_id` — precisely the case your test pins down.

The test is well built: it constructs the divergent case explicitly (`id = "local-uuid-1234"`, `_id = "server-doc-id-5678"`), captures the upserted row with a MockK slot, and asserts on `replyTo`. The name states the intent, so the next person to "simplify" this back to `_id ?: id` gets a failing test explaining why not.

**One thing worth confirming before merge.** `replyTo` crosses the sync boundary in both directions untranslated — pulled in verbatim at `news.replyTo = JsonUtils.getString("replyTo", doc)` (line 410) and pushed out verbatim at `object.addProperty("replyTo", news.replyTo)` (line 461). Cross-device and web threading therefore depends on `replyTo` holding a value the server also recognises, which is true for every synced parent precisely because `id == _id` there. In the divergent window above, though, the uploaded reply would carry a local UUID the server has no document for. Given that's the same window this fix targets, it'd be worth confirming that such a reply still threads correctly on the Planet side — or that the window can't produce an uploadable reply at all. Local behaviour is unambiguously improved either way; I'd just rather that question be answered deliberately than discovered from a web-side report.

**Nit:** `(binding.rvNews.adapter as? VoicesAdapter)?.setCurrentUser(user)` fails silently in two ways — the safe cast swallows a wrong/absent adapter, and it runs inside the user-loading coroutine, so if the adapter is attached after that completes the current user is never set. A non-null adapter reference or setting the user as a constructor/property before attach would make the ordering explicit instead of timing-dependent.

---

dogi (approved):

PR #15554, opened 2026-08-11
dashboard: smoother activities chart landscaping (fixes #15552)

### LINKED ISSUE #15552: login activity chart not responsive in landscape
The login activity chart is not responsive when the device is in landscape orientation.

**Steps to reproduce**
Open the app and navigate to the login activity chart.
Rotate the device to landscape mode.
Observe the chart layout.
https://github.com/user-attachments/assets/76f63beb-0ac9-4cfa-bf6d-40ec47be8ea9

ragilzakaria:
https://github.com/user-attachments/assets/2ae6ebc6-396c-431f-bdc8-027c7db03bd8

fixes #15552 

## Summary by CodeRabbit

- **New Features**
  - Added an Activities chart with clearer month labels, improved formatting, and wrapped legend text.
  - Added a dedicated landscape layout for viewing activity data.
  - Added an empty-state message when no activity is available.

- **Improvements**
  - Activities screens now support vertical scrolling and better spacing across screen sizes.
  - Activity counts are grouped and displayed more accurately by month.

- **Tests**
  - Added coverage for monthly activity calculations and date filtering.


dogi:
**Review — rating: 84/100**

Wrapping the chart in a `NestedScrollView` with `fillViewport="true"` is the right fix for content that can't fit in landscape, and the chart configuration changes are all sound MPAndroidChart hygiene rather than cosmetic churn: `axisLeft.axisMinimum = 0f` is correct for counts that can't go negative, `granularity = 1f` on both axes stops fractional duplicate labels on inherently discrete month/count data, `axisRight.isEnabled = false` drops a redundant mirrored axis, `setFitBars(true)` prevents the first and last bars being clipped at the axis bounds, and `legend.isWordWrapEnabled = true` earns its place precisely in the landscape case this PR is about. Replacing the hardcoded `50dp`/`80dp`/`16dp` margins with dimension resources is a real improvement too.

I checked that the new test can actually run as written, since it instantiates a bare `ActivitiesFragment()` with no Robolectric runner. It can: `getMonth` is `DateFormatSymbols().months[month]`, pure JVM with no `Context` dependency, and `computeMonthlyCounts` is a pure function of its arguments. CI agrees — `test (default)` is green. Covering the `loginTime = null` entry and the out-of-range row is good edge selection.

**Worth addressing:**

- `layout-land/fragment_activities.xml` is a near-copy of the portrait layout, differing only in chart height (220dp vs 320dp) and padding. A `chart_height` dimen with a `values-land/dimens.xml` override would get the same result with one layout to maintain instead of two that have to be kept in sync by hand. The chart heights and `18sp`/`16sp` text sizes also stay hardcoded while the rest of the PR moves to dimens.
- Widening `computeMonthlyCounts` from `private` to `internal` purely for the test is a smell on a Fragment. The function is pure — list plus range in, map out — so it belongs in a ViewModel or a util object, which is also where this repo's testing conventions put logic. Failing that, `@VisibleForTesting` would at least document why the visibility is loosened.
- The new test locks in a pre-existing quirk without covering it. `computeMonthlyCounts` buckets on `calendar.get(Calendar.MONTH)` and discards the year, while the caller's window is a rolling `Calendar.getInstance().add(Calendar.YEAR, -1)` → `now`. That window isn't month-aligned, so the current month appears at *both* ends: today being 11 August, a login on 20 Aug 2025 and one on 5 Aug 2026 are both in range and both land in bucket 7, merging into a single bar. Your test's Dec-31→Dec-31 range happens to sidestep this. Not a regression — the PR doesn't touch that logic — but since you're adding the first test for this function, a same-month-different-year case would pin down whether the merge is intended or a bug worth its own issue.

**Nit:** `DateFormatSymbols()` resolves against the JVM default locale rather than the app's configured locale, so month labels won't follow an in-app language override. Pre-existing, but relevant while these labels are being polished.

No overlap with #15555 despite both being orientation work — that one is entirely `fragment_user_profile.xml` variants, no shared files.

---

dogi:
**Review — rating: 86/100**

Clean, self-contained fix. The root cause of the cropping was a `FrameLayout` holding a fixed 300dp chart plus a 50dp top margin with no way to scroll, so in landscape the chart simply ran off the bottom; converting both layouts to `NestedScrollView` with `fillViewport="true"` and adding a `layout-land` variant at 220dp is the right shape, and swapping the hardcoded 16/50/80dp margins for `@dimen` values brings it in line with the rest of the resources.

The chart configuration changes are real improvements rather than churn: `axisLeft.axisMinimum = 0f` stops the y-axis from floating off zero on low counts, `granularity = 1f` on both axes prevents MPAndroidChart from emitting duplicate integer labels when the range is small, and disabling `axisRight` removes a redundant duplicate scale that was costing horizontal space — which is exactly the constrained dimension in landscape. `setFitBars(true)` fixes the half-clipped first/last bar that the old config would produce.

I checked the binding-compatibility risk: both layouts keep `chart_label`, `chart` and `emptyState` with identical ids and view types, and the portrait root type change from `FrameLayout` to `NestedScrollView` alters `FragmentActivitiesBinding.getRoot()`'s type — build (default + lite) and test (default) are green, so nothing was assigning that root to a `FrameLayout`. The new `ActivitiesFragmentTest` covers the parts worth covering: empty input, out-of-range filtering, null `loginTime`, and month grouping. Range construction and the data all use the same default-timezone `Calendar`, so there's no timezone skew between the fixture and the assertion.

**Nit:** `computeMonthlyCounts` was widened from `private` to `internal` purely for the test — defensible here since it buys genuine coverage of the filtering logic, but it is production visibility spent on tests. The test also never clears `Calendar.MILLISECOND` after `Calendar.getInstance()`, so the fixture timestamps carry a random sub-second component; harmless at these boundaries, but `calendar.clear()` first would make it deterministic.

**Verdict:** mergeable. For the record it doesn't collide with #14960 despite both being login-flavoured landscape work — this touches only `ui/dashboard/ActivitiesFragment.kt` and `fragment_activities.xml`.

---

Okuro3499 (approved):

dogi:
**Promoted to `merge`** — Okuro3499 approved, my review stands at 86/100, no changes since.

---

dogi (approved):

PR #15555, opened 2026-08-11
login: smoother user profile landscaping (fixes #15551)

### LINKED ISSUE #15551: User profile page is not responsive
## Summary
The user profile page is not responsive and does not adapt correctly to smaller screen sizes.

## Steps to reproduce
1. Open the MyPlanet app.
2. Navigate to the user profile page in landscape.
https://github.com/user-attachments/assets/fc72c3a9-d71a-404b-8df8-81272c85a4c2

ragilzakaria:
https://github.com/user-attachments/assets/d10dbbf5-714b-471f-a420-abb49d31d18a

fixes #15551 

## Summary by CodeRabbit

- **UI Improvements**
  - Refined the user profile screen layout for landscape and large-screen devices.
  - Improved spacing, alignment, sizing, and visual consistency across profile details and statistics.
  - Added responsive dimensions for larger screens.
  - Enhanced profile image and edit controls.
  - Improved scrolling behavior and statistics presentation on smaller layouts.


dogi:
**Review — rating: 74/100**

Several of the cleanups here are genuinely sharp catches, but the PR leaves the resource tree in a worse state than it needs to be and quietly shrinks the tablet layout. Both are small to fix.

**What's good, and not obvious:**
- Removing `android:layout_alignParentBottom="true"` from the FAB is a real catch — that's a `RelativeLayout` attribute sitting inside a `ConstraintLayout`, where it does nothing but generate lint noise.
- `tools:context` corrected from `.ui.userprofile.UserProfileFragment` to `.ui.user.UserProfileFragment`. Tools-only, but the old value pointed at a package that doesn't exist, so IDE previews were broken.
- The FAB tint moves from `@color/disable_color` to `@color/profile_pic`, which I confirmed is defined in both `values/colors.xml` (#BDBDBD) and `values-night/colors.xml` (#757575) — so it's theme-aware, and it stops an interactive control being painted in a "disabled" colour.
- `@style/primaryButton` replaces inline `background`/`textColor`/`textAllCaps` on the edit button, and `rv_stat` moves from `wrap_content` + weight to the correct `0dp` + weight idiom. `fillViewport="true"` on the portrait scroll view is right.

**Four identical landscape layouts.** After this PR, `layout-land/`, `layout-large-land/`, `layout-normal-land/` and `layout-xlarge-land/` all contain byte-identical content — every one resolves to git blob `c48a8dc6a5`. `layout-land/` on its own already covers every landscape configuration, so the three size-qualified copies are pure duplication, and worse than inert: they *win* resource resolution on large, normal and xlarge devices, so the next person who edits `layout-land/` will find their change silently not applying on most tablets. `-large-land` and `-normal-land` were already identical to each other before this PR, so the opportunity here was to consolidate; adding a fourth copy instead moves the wrong way, particularly with #14662 and #14663 open to purge redundant resources. Please delete the three qualified variants and keep only `layout-land/`.

**The xlarge layout gets visibly smaller.** `layout-xlarge-land/` was the one variant with deliberate tablet tuning — a `250dp` image and `24sp` text — and collapsing it into the shared layout replaces those with `@dimen/user_image_size` and `@dimen/text_size_large`/`text_size_mid`. I checked the buckets: `user_image_size` is defined **only** in `values/dimens.xml` at `150dp`, with no size-qualified override, and `text_size_large` is `18sp` / `text_size_mid` is `16sp` (the only override anywhere is `values-sw320dp` scaling *down* to 15sp). So on extra-large tablets the profile image drops 250dp → 150dp and the detail rows drop 24sp → 16sp, with nothing compensating. Consolidating is still the right instinct — the fix is to add `values-xlarge/dimens.xml` (or `values-sw720dp`) overrides for `user_image_size` and the text sizes so one layout scales, rather than one layout flattening every screen to phone proportions.

**Nit:** the `6dp`/`4dp`/`2dp` margins stay hardcoded while the rest of the PR moves to dimension resources.

For the record, no overlap with #15554 despite both being orientation work — that PR is entirely `fragment_activities.xml` and its fragment, no shared files, so neither supersedes the other and I've marked no duplicate.

**Verdict:** needs another round — delete the three redundant landscape variants and restore tablet scaling via `values-xlarge` dimens. The structural rewrite and the attribute cleanups are good and should stay.

---

dogi:
**Review — rating: 72/100**

The core move is right and overdue: collapsing `layout-normal-land`, `layout-large-land` and `layout-xlarge-land` into a single `layout-land` deletes ~260 lines of near-identical XML, and expressing the size differences as qualified dimensions instead of forked layouts is the correct modern pattern. Several incidental fixes are good too — `layout_alignParentBottom` was a `RelativeLayout` attribute doing nothing inside a `ConstraintLayout`, the `tools:context` package was stale (`.ui.userprofile` → `.ui.user`), the FAB tint now matches portrait's `@color/profile_pic`, and switching the weighted children from `wrap_content` to `0dp` is a genuine correctness fix for `LinearLayout` weighting rather than a cosmetic one.

I checked the resource-resolution risk first, since a dimension defined only under a qualifier links fine but throws `Resources$NotFoundException` at runtime elsewhere. That's not an issue here — `values/dimens.xml` already carries defaults for all of them (`user_image_size` 150dp, `text_size_large` 18sp, `text_size_mid` 16sp, plus `_2dp`/`_10dp`), and `@style/primaryButton` and `@color/profile_pic` both exist. So nothing crashes.

The problem is the blast radius of the new `values-xlarge/dimens.xml` and `values-sw720dp/dimens.xml`. Those override **shared** dimensions, not profile-specific ones. `text_size_mid` appears in 38 layout files, and `user_image_size` is used by `fragment_voices.xml`, `fragment_discussion_list.xml`, `item_my_life.xml` and `fragment_vital_sign.xml` besides this screen. So a PR scoped to the user profile page silently bumps body text from 16sp to 20sp across 38 layouts and avatars from 150dp to 250dp on four other screens, for every tablet — and unlike the old `layout-xlarge-land`, which was landscape-only, these `values-*` folders apply in **portrait as well**. Enlarging text app-wide on screens nobody re-tested is a real risk of introducing exactly the kind of overflow this PR set out to fix.

The fix is to stop overriding the shared names: introduce profile-scoped dimensions (`profile_image_size`, `profile_name_text_size`, …) and qualify those instead, leaving `text_size_mid`/`user_image_size` alone. I'd also drop one of the two folders — `values-xlarge` and `values-sw720dp` are byte-identical, and `xlarge` is the deprecated qualifier that `sw720dp` replaced, so keeping both just doubles the maintenance surface.

**Nit:** `txt_dob`, `txtLanguage` and `txtLevel` go from a hardcoded 20sp to `text_size_mid` (16sp default), which reads as deliberate normalization against the neighbouring fields that already used `text_size_mid` — worth confirming that shrink is intended, since it's a visible change on phones in landscape.

**Verdict:** needs another round — the consolidation is good, the shared-dimension overrides need scoping to this screen.

---

dogi:
**Review — rating: 85/100**

Good consolidation. Three landscape variants (`layout-normal-land`, `layout-large-land`, `layout-xlarge-land`) collapse into one `layout-land`, with the size differences lifted into `values/` and `values-sw720dp/` dimens. That's the right direction — `normal`/`large`/`xlarge` are deprecated buckets and `swNNNdp` is the supported replacement — and it removes two files that were byte-identical duplicates of each other.

The detail I most wanted to check is the one this actually fixes: `UserProfileFragment` was passing `R.dimen.user_image_size` (150dp) to Glide while `layout-xlarge-land` rendered the `CircleImageView` at a hardcoded 250dp, so tablets were decoding an undersized bitmap into a larger view. Switching both the Kotlin and the layouts onto the same `profile_image_size` token — 150dp in `values/`, 250dp in `values-sw720dp/` — makes the decode target track the actual view size per device class. That's a real fix, not just tidying. Introducing a separate `profile_image_size` rather than reusing `user_image_size` is also correct: the latter is still live in `MembersDetailFragment`, `fragment_voices`, `fragment_discussion_list` and `fragment_vital_sign`, so the profile screen can now scale without dragging those along.

Two other cleanups are right for reasons worth stating. `android:layout_alignParentBottom` was a `RelativeLayout` attribute sitting on a FAB inside a `ConstraintLayout`, where it does nothing — removing it is dead-attribute cleanup, and the FAB's real anchoring comes from the `layout_constraintBottom_toBottomOf="@+id/image"` that stays. And `tools:context` was pointing at `.ui.userprofile.UserProfileFragment`, a package that doesn't exist; the class lives in `ui/user/`. Tooling-only, but it had been stale. `fillViewport="true"` on both `NestedScrollView`s is the correct companion to the `wrap_content` child. CI green (build default + lite, test default).

Undeclared side effects, all defensible but none mentioned in the description:

- **7" tablets in landscape lose a little text size.** They previously matched `layout-large-land` at 20sp for name/dob/language/level; they now fall to `values/` at 18sp name / 16sp detail, because `values-sw720dp` only catches 10" and up. If that band matters, a `values-sw600dp` tier would restore it.
- **Tablet *portrait* changes too.** `values-sw720dp` is orientation-agnostic, so a 10" tablet in portrait now gets a 250dp image and 24sp name where it previously used the 150dp portrait layout. Sensible, but it's a portrait change in a PR framed around landscape.
- **Phone landscape restacks the columns.** The surviving layout is the renamed `xlarge-land` one, and `rv_stat` goes from `layout_weight="3"` to `1.2`, so the stats list drops from roughly three-quarters of the width to about half. That looks like the intended fix for the cramped profile column, but it is the change most visible to the most users and deserves a line in the description.

**Nit:** the landscape `CircleImageView` uses `padding_small` while portrait keeps `padding_normal`, and the landscape `TextView`s carry `android:gravity="center"` at `wrap_content` inside a `center_horizontal` parent, where it has no effect.

**Verdict:** mergeable.

---

Okuro3499 (approved):

dogi (approved):

PR #15559, opened 2026-08-11
exam: redesign UI with elapsed timer and cards (fixes #15558)

### LINKED ISSUE #15558: redesign exam UI
Modernise exam UI

Okuro3499:
fixes #15558
[Screen_recording_20260811_205405.webm](https://github.com/user-attachments/assets/41cd3308-b9bf-4cb7-a725-0fb0b80c16af)

## Summary by CodeRabbit

- **New Features**
  - Added an elapsed-time display that persists during exam navigation and screen recreation.
  - Added clearer question progress indicators, exam titles, and dedicated “Submit Answer” and “Finish and Submit” actions.
  - Added responsive exam layouts for different screen sizes.

- **UI Improvements**
  - Refreshed answer choices with card-style selection states, improved spacing, and accessible sizing.
  - Added a timer icon, elapsed-time chip, progress bar, and improved exam header and action area.

- **Localization**
  - Added translations for question numbering, elapsed time, and finish-and-submit actions in supported languages.

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/ui/exam/ExamTakingFragment.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/exam/ExamTakingFragment.kt
@@ -7,6 +7,7 @@ import android.graphics.drawable.StateListDrawable
 import android.os.Bundle
 import android.os.Handler
 import android.os.Looper
+import android.os.SystemClock
 import android.text.Editable
 import android.text.TextUtils
 import android.text.TextWatcher
@@ -62,12 +63,11 @@ class ExamTakingFragment : BaseExamFragment(), View.OnClickListener, CompoundBut
     private var isExplicitSubmission = false
     private var examTakingTextWatcher: TextWatcher? = null
     private val answerCache = mutableMapOf<String, AnswerData>()
-    private var elapsedSeconds = 0
+    private var startElapsedRealtime = 0L
     private var isTimerRunning = false
     private val timerHandler = Handler(Looper.getMainLooper())
     private val elapsedTimeRunnable = object : Runnable {
         override fun run() {
-            elapsedSeconds++
             updateElapsedTimeText()
             timerHandler.postDelayed(this, 1000L)
         }
@@ -95,7 +95,8 @@ class ExamTakingFragment : BaseExamFragment(), View.OnClickListener, CompoundBut
 
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
-        elapsedSeconds = savedInstanceState?.getInt(KEY_ELAPSED_SECONDS) ?: 0
+        val resumedElapsedSeconds = savedInstanceState?.getInt(KEY_ELAPSED_SECONDS) ?: 0
+        startElapsedRealtime = SystemClock.elapsedRealtime() - resumedElapsedSeconds * 1000L
         updateElapsedTimeText()
         initializeExamData()
         setupListeners()
@@ -104,7 +105,20 @@ class ExamTakingFragment : BaseExamFragment(), View.OnClickListener, CompoundBut
 
     override fun onSaveInstanceState(outState: Bundle) {
         super.onSaveInstanceState(outState)
-        outState.putInt(KEY_ELAPSED_SECONDS, elapsedSeconds)
+        outState.putInt(KEY_ELAPSED_SECONDS, currentElapsedSeconds())
+    }
+
+    override fun onPause() {
+        super.onPause()
+        timerHandler.removeCallbacks(elapsedTimeRunnable)
+    }
+
+    override fun onResume() {
+        super.onResume()
+        if (isTimerRunning) {
+            updateElapsedTimeText()
+            timerHandler.postDelayed(elapsedTimeRunnable, 1000L)
+        }
     }
 
     private fun startTimer() {
@@ -118,9 +132,13 @@ class ExamTakingFragment : BaseExamFragment(), View.OnClickListener, CompoundBut
         timerHandler.removeCallbacks(elapsedTimeRunnable)
     }
 
+    private fun currentElapsedSeconds(): Int =
+        ((SystemClock.elapsedRealtime() - startElapsedRealtime) / 1000L).toInt()
+
     private fun updateElapsedTimeText() {
-        val minutes = elapsedSeconds / 60
-        val seconds = elapsedSeconds % 60
+        val totalSeconds = currentElapsedSeconds()
+        val minutes = totalSeconds / 60
+        val seconds = totalSeconds % 60
         binding.tvElapsedTime.text = String.format(Locale.US, "%02d:%02d", minutes, seconds)
     }
 
--- a/app/src/main/java/org/ole/planet/myplanet/ui/exam/ExamTakingFragment.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/exam/ExamTakingFragment.kt
@@ -5,6 +5,8 @@ import android.graphics.Typeface
 import android.graphics.drawable.GradientDrawable
 import android.graphics.drawable.StateListDrawable
 import android.os.Bundle
+import android.os.Handler
+import android.os.Looper
 import android.text.Editable
 import android.text.TextUtils
 import android.text.TextWatcher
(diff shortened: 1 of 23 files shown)


dogi:
**Review — rating: 78/100**

A lot of this is careful work, including two fixes that go beyond the visual brief. It's held just under the line by the timer — the headline feature — being subtly inaccurate, plus two cleanup misses.

**Genuinely good, and easy to miss:**

- The inflate change is a real bug fix, not a style tweak. `item_checkbox.xml` gains `layout_width="match_parent"` and `layout_marginBottom`, and those attributes are *only* honoured if the inflater has a parent to generate LayoutParams from — inflating with `null` silently discards them. I checked the sibling paths: `addRadioButton` (line 582) and the radio branch of `addCompoundButton` already passed `binding.groupChoices, false`, so the checkbox branch was the one asymmetric call, and this PR fixes exactly it. Without that, the new card styling would have applied to radios but not checkboxes.
- `btnBack` and `btnNext` both had `contentDescription="@string/btn_back"` — a copy-paste bug that made TalkBack announce "back" for the next button. Now `@string/previous` / `@string/next`. Also drops the inert `android:text` from both `ImageButton`s.
- The root goes from `FrameLayout` wrapping three nested `LinearLayout`s to a single `LinearLayout`, a real reduction in hierarchy depth on a screen that inflates a view per choice.
- Theming is done properly: `choice_button_tint` is a `state_checked` selector, `bg_choice_card` a state-list with distinct selected/unselected shapes, and `choice_border_unselected` is defined in **both** `values/` (#33000000) and `values-night/` (#33FFFFFF). The `exam_chip_*` colours are translucent white, correct for sitting on `colorPrimary` in either theme. All four new strings are translated into all five locales.
- `layout-w640dp` is the right qualifier — width-driven rather than `-land`, so it picks up tablets and wide landscape alike — and the two-column split (question weight 2, answers weight 3) is a real structural change that justifies a separate file.

**The timer under-counts.** It accumulates ticks rather than measuring time:

```kotlin
override fun run() {
    elapsedSeconds++
    updateElapsedTimeText()
    timerHandler.postDelayed(this, 1000L)
}
```

`postDelayed` guarantees *at least* 1000 ms, never exactly, so every scheduling delay is lost permanently — and this app ships an `ANRWatchdog`, so main-thread contention is a known reality. The bigger loss is lifecycle: there's no `onPause`/`onResume` handling, only `onViewCreated`/`onDestroyView`, so on Android 12+ a cached-and-frozen process stops delivering the callbacks entirely and the display silently loses that entire period. Recording a start timestamp with `SystemClock.elapsedRealtime()`, persisting it in `onSaveInstanceState` alongside the count, and computing `now - start` on each tick makes the value correct regardless of tick delivery — and the ticks then only drive the refresh, not the measurement. While you're there, decide deliberately whether backgrounded time should count; a `Handler` posting every second for the life of the fragment is also a needless wakeup on the low-end hardware this app targets. Worth noting the value is display-only — it's never persisted or submitted — so if time-on-task was meant to be captured, that part isn't wired.

**Two cleanup misses:**

- `R.string.Q` and `R.string.Q1` are now orphaned. Their only two usages were `ExamTakingFragment` lines 102 and 331, both replaced here by `exam_question_of`, and nothing else references them — so they're twelve dead entries across the six `strings.xml` files. (`finish` and `submit` do stay in use elsewhere, so leave those.) With #14662 and #14663 open to purge unused resources, worth removing them here.
- `values-sw320dp/strings.xml` only helps English. Android resolves the locale qualifier at higher precedence than `smallestWidth`, so on a small Arabic device `values-ar/strings.xml` wins over `values-sw320dp/strings.xml` and the compact `%1$d / %2$d` form never applies. Either add the override per locale (as #15525 does for its `-sw360dp` strings) or drop the file, since the long form is what non-English small screens will get either way.

**Nits:** `contentDescription` on the `elapsed_time_chip` `LinearLayout` won't reliably be announced as a unit without `focusable="true"`/`importantForAccessibility`, and it describes the label without the value. The `layout-w640dp` variant duplicates the header and footer blocks verbatim — `<include>` would keep them in sync. `String.format(Locale.US, …)` keeps ASCII digits in locales that otherwise use localized numerals; defensible for a clock, just a deliberate inconsistency.

**Verdict:** needs another round — make the elapsed time timestamp-based and decide its background behaviour, then drop the orphaned `Q`/`Q1` strings and resolve the `sw320dp` locale gap. The layout, theming, accessibility and inflate fixes are solid and should stay as they are.

---

dogi:

`a355cc5` addresses all three items from the last round, and it does each one properly rather than papering over it. Build (default + lite) and test (default) green on the new head.

**1. The timer is now measured, not accumulated.** `elapsedSeconds++` is gone from the runnable, which is reduced to a pure refresh, and the value comes from a single subtraction:

```kotlin
private fun currentElapsedSeconds(): Int =
    ((SystemClock.elapsedRealtime() - startElapsedRealtime) / 1000L).toInt()
```

Tick delay no longer costs anything, because ticks don't carry the measurement. The restore path is right too — `startElapsedRealtime = SystemClock.elapsedRealtime() - resumedElapsedSeconds * 1000L` reconstructs the origin from the saved count, so rotation resumes at the correct value instead of restarting. `elapsedRealtime()` is also the correct clock here: it's monotonic and unaffected by wall-clock changes, so a user changing the device time mid-exam can't skew or reverse the timer.

**2. `onPause`/`onResume` handling is correct, and it answers the design question I raised.** Ticks stop while backgrounded (no per-second wakeups on low-end hardware), but because measurement is timestamp-based, time spent backgrounded still counts — which is the right semantics for something labelled "elapsed time". `onResume` calls `updateElapsedTimeText()` before re-posting, so the display jumps straight to the correct value rather than showing a stale number for a second.

**3. `Q`/`Q1` are gone from all six locale files**, and the `sw320dp` compact variant now exists per locale — `values-ar-sw320dp`, `-es-`, `-fr-`, `-ne-`, `-so-` alongside the existing `values-sw320dp`. That closes the precedence gap exactly: previously `values-ar` outranked `values-sw320dp` so only English got the compact form; now every locale has a same-precedence match on a small screen.

**One small thing introduced by the new lifecycle code.** On first creation the runnable gets posted twice. `onViewCreated` calls `startTimer()`, which posts and sets `isTimerRunning = true`; `onResume` then fires, sees the flag already true, and posts the same `Runnable` instance again. `Handler` doesn't dedupe, so two self-perpetuating chains run and `updateElapsedTimeText()` fires twice a second. It's harmless to correctness now that the text is derived from a timestamp rather than incremented — and it self-corrects after the first pause, since `removeCallbacks` clears both chains and `onResume` re-posts one. Still worth tidying: let `startTimer()` set the flag only and leave the posting to `onResume`, so there's a single owner of the scheduling.

**Nits from last time that are still open** (all genuinely minor): `contentDescription` on the `elapsed_time_chip` `LinearLayout` won't reliably be announced without `focusable="true"`, and it names the label without the value; the `layout-w640dp` variant still duplicates the header and footer blocks verbatim where `<include>` would keep them in sync; and `Locale.US` keeps ASCII digits in locales that otherwise use localized numerals.

**Verdict:** mergeable. Not moving to `merge` because the trusted-reviewer gate can't be satisfied here — @Okuro3499 is the author and can't approve his own PR — and there's no star tag on this one, so it needs a separate reviewer's sign-off.

---

Okuro3499 (commented):
@dogi 
```
Why exam_question_of is duplicated across values-{lang}-sw320dp/strings.xml                                                                                                   
                                                                                                                                                                                
Android resolves resource qualifiers in a fixed precedence order, and locale outranks smallest-width (sw320dp). When picking a resource, Android matches locale first, then   
narrows by width among only the locale-matching candidates.                                                                                                                   
                                                                                                                                                                                
That means a single generic values-sw320dp/strings.xml (no locale) only serves the default locale. For any of our 5 translated locales (ar, es, fr, ne, so), a device in that 
locale would match its locale-specific file (values-ar, values-es, etc.) before width is even considered — skipping the generic sw320dp override entirely and falling back to 
the full, longer translated string ("Question %1$d of %2$d" / "Pregunta %1$d de %2$d" / …) even on narrow screens.                                                            
                                                                                                                                                                                
So each values-{lang}-sw320dp/strings.xml exists to make the compact %1$d / %2$d format win for that locale on narrow-width devices. The value happens to be identical across 
languages today (it's just digits and a separator, nothing to translate), but the file itself is still required per locale — it's not redundant duplication, it's how         
Android's locale-before-width matching rule has to be satisfied. 
```

J-S-webskas (approved):
Currently, the survey designs are working well. I did find one spelling mistake for the survey "Belmont - Todos Commectados Community Interview - team 4": heathcare is spelled as healh care

Image

PR #15560, opened 2026-08-11
resources: smoother repository pending downloading (fixes #15557)

### LINKED ISSUE #15557: IllegalStateException
```
java.lang.IllegalStateException: Couldn't read row 50, col 0 from CursorWindow.  Make sure the Cursor is initialized correctly before accessing data from it.
at android.database.CursorWindow.nativeGetString(Native Method)
at android.database.CursorWindow.getString(CursorWindow.java:505)
at android.database.AbstractWindowedCursor.getString(AbstractWindowedCursor.java:54)
at androidx.sqlite.driver.SupportSQLiteStatement$RowSQLiteStatement.getText(SupportSQLiteStatement.android.kt:370)
at org.ole.planet.myplanet.data.room.dao.MyLibraryDao_Impl.getPendingDownloadsForUserPatternFlow$lambda$0(MyLibraryDao_Impl.kt:6466)
at org.ole.planet.myplanet.data.room.dao.MyLibraryDao_Impl$$ExternalSyntheticLambda7.invoke(D8$$SyntheticClass:0)
at androidx.room.util.DBUtil__DBUtil_androidKt$performSuspending$lambda$1$$inlined$internalPerform$1.invokeSuspend(DBUtil.kt:173)
at androidx.room.util.DBUtil__DBUtil_androidKt$performSuspending$lambda$1$$inlined$internalPerform$1.invoke(DBUtil.kt:8)
at androidx.room.util.DBUtil__DBUtil_androidKt$performSuspending$lambda$1$$inlined$internalPerform$1.invoke(DBUtil.kt:4)
at androidx.room.coroutines.PassthroughConnectionPool$useConnection$2.invokeSuspend(PassthroughConnectionPool.kt:59)
at androidx.room.coroutines.PassthroughConnectionPool$useConnection$2.invoke(PassthroughConnectionPool.kt:8)
at androidx.room.coroutines.PassthroughConnectionPool$useConnection$2.invoke(PassthroughConnectionPool.kt:4)
at kotlinx.coroutines.intrinsics.UndispatchedKt.startUndispatched(Undispatched.kt:66)
at kotlinx.coroutines.intrinsics.UndispatchedKt.startUndispatchedOrReturn(Undispatched.kt:43)
at kotlinx.coroutines.BuildersKt__Builders_commonKt.withContext(Builders.common.kt:497)
at kotlinx.coroutines.BuildersKt.withContext(dex-id-f8807928e301f3e00b2f3e943cc1a1c45bcd9b4b:1)
at androidx.room.coroutines.PassthroughConnectionPool.useConnection(PassthroughConnectionPool.kt:59)
at androidx.room.RoomConnectionManager.useConnection(RoomConnectionManager.android.kt:138)
at androidx.room.RoomDatabase.useConnection(RoomDatabase.android.kt:619)
at androidx.room.util.DBUtil__DBUtil_androidKt$performSuspending$$inlined$compatCoroutineExecute$DBUtil__DBUtil_androidKt$1.invokeSuspend(DBUtil.android.kt:261)
at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:34)
at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:100)
at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1100)
at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
at java.lang.Thread.run(Thread.java:1572)
Suppressed: kotlinx.coroutines.internal.DiagnosticCoroutineContextException: [StandaloneCoroutine{Cancelling}@3f62be6, Dispatchers.Main.immediate]
```

Okuro3499:
fixes #15557
Change getPendingDownloads to return Flow<List<String>> (IDs only) instead of Flow<List<MyLibrary>> to reduce memory overhead when only resource IDs are needed.

## Summary by CodeRabbit

* **Improvements**
  * Improved handling of pending downloads by using resource identifiers, supporting more efficient download tracking.

* **Tests**
  * Updated pending-download validation to reflect identifier-based download tracking.

* **Chores**
  * Updated the Android app version to 0.64.33.


dogi:
**Review — rating: 88/100**

Narrows `getPendingDownloadsForUserPatternFlow` from `SELECT *` to `SELECT id`, changing the flow type from `Flow<List<MyLibrary>>` to `Flow<List<String>>` through the DAO, repository interface, impl and test. Small and correct.

I checked the two things that could have made this unsafe. The projection is type-safe: `MyLibrary.id` is `@PrimaryKey var id: String = ""`, non-null, so Room's single-column mapping to `List<String>` can't produce nulls in a list Kotlin believes is non-nullable. And no caller loses anything — there is exactly one consumer, `DashboardViewModel:200`:

```kotlin
resourcesRepository.getPendingDownloads(userId ?: "").map {},
```

That `.map {}` discards the payload outright; the flow is used purely as a change signal. So nothing was ever reading a `MyLibrary` field off this query, which makes the type change safe and the win real: every table invalidation previously hydrated full entities — all columns, including the Gson-backed `@TypeConverters` fields — to build objects that were immediately thrown away.

Worth being clear about where the saving is and isn't. `userIdPattern` produces a leading-wildcard pattern (`%"userId"%`), so `userId LIKE :userPattern` can't use an index and this stays a full table scan either way. The improvement is entirely in per-row hydration and allocation, not lookup — which is still the right fix, just not a change in query complexity.

**One suggestion, given how the result is actually used.** Since the only consumer maps the payload to `Unit`, even a list of ids is more than needed — it still allocates N strings per emission. `SELECT COUNT(*)` returning `Flow<Int>` would allocate one boxed `Int` and serve that consumer exactly as well. Returning ids is the more general contract and a reasonable deliberate choice, so I'd take this as-is; but if the intent is purely "something changed", a count is the cheaper endpoint, and if the intent is a future caller that needs the ids, that caller doesn't exist yet.

The test update is right — `listOf("lib1")` in place of `listOf(mockk<MyLibrary>())` keeps the pattern assertion intact while matching the new type. Worth double-checking whether the `MyLibrary` import is still used elsewhere in `ResourcesRepositoryImplTest` now that this case doesn't need it.

**Verdict:** mergeable — CI green (build default + lite, test default). Not `merge`: @Okuro3499 is the author, so the trusted-reviewer approval has to come from someone else, and there's no star tag on this PR.

---

dogi (approved):

PR #15579, opened 2026-08-12
resources: smoother grid thumbnail previewing (fixes #15574)

### LINKED ISSUE #15574: Show resource previews in the list instead of the colored backgrounds
## Summary
Currently, the resource list displays colored backgrounds as placeholders for resources. This issue proposes replacing the colored backgrounds with actual resource previews (e.g., thumbnails or cover images) to give users a more informative and visually rich experience when browsing resources.

## Desired Behavior
Each resource in the list should display a meaningful preview — such as a thumbnail image, cover page, or generated preview — instead of a plain colored background.

## Benefits
- Improves the user experience by giving visual context for each resource at a glance.
- Makes it easier for users to identify and select the resource they are looking for.
- Aligns the UI with more modern content browsing patterns.

Okuro3499:
fixes #15574
[resources.webm](https://github.com/user-attachments/assets/f9e0cb6c-5556-4ca6-9836-a8b35361eb7c)

## Summary by CodeRabbit

* **New Features**
  * Added visual previews for downloaded images, videos, and first-page PDF documents in library grid and list views.
  * Added fallback icons when previews are unavailable, unsupported, or fail to load.
  * Added consistent rounded styling for library cover previews.

* **Bug Fixes**
  * Improved preview loading by cancelling outdated work during scrolling.
  * Improved PDF preview reuse, cache management, and memory handling.
  * Standardized file extension handling and resource file lookup for more reliable previews.


dogi:
**Review — rating: 70/100**

The scaffolding here is the best part and it's genuinely well done: `dispatcherProvider` injected rather than hard-coded per `CLAUDE.md`, an adapter scope cancelled in `onDetachedFromRecyclerView` and rebuilt in `onAttachedToRecyclerView`, per-holder job cancellation on rebind and in `onViewRecycled`, and a clean fall-through to the existing coloured container + type icon whenever a preview can't be produced. Returning the `Job` from `bindCover` so the holder owns cancellation is the right shape, and returning `null` on the non-PDF branches correctly kills a stale PDF job when a holder is rebound to an image. Build (default + lite) and test (default) green.

Four things need another pass.

**PDF bitmaps are rendered at full page resolution for an 84dp thumbnail.** `createBitmap(page.width * scale, page.height * scale)` with `scale = 2` uses `PdfRenderer` page dimensions in points, so a US Letter page is 612×792 → 1224×1584 at ARGB_8888 ≈ 7.7 MB per thumbnail, cached into an `LruCache` sized `maxMemory/8`. That's a handful of PDFs to fill the budget, for images displayed in an 84dp grid cover or a 44dp list thumbnail. Deriving the scale from the target view size (or a fixed small width) would cut this by one to two orders of magnitude on exactly the low-end hardware myPlanet targets.

**The previews won't respect the rounded corners.** Neither `item_library_grid.xml` nor `item_library_list.xml` gained `android:clipToOutline="true"` on the cover container, so a `centerCrop` `match_parent` `ImageView` paints square corners over `bg_cover_top_rounded` / `bg_rounded_thumbnail`. Worth noting that #15582 does add `clipToOutline` for the course covers, so as it stands the two lists will render inconsistently. One caveat when you add it: `bg_rounded_thumbnail` is a uniform 8dp radius and will clip correctly, but `bg_cover_top_rounded` sets only `topLeftRadius`/`topRightRadius`, and `GradientDrawable` can only express a uniform radius as an `Outline` — with per-corner radii the clip degrades to a plain rectangle. The grid cover will need a different approach (a rounded `Glide` transform, or a `ShapeableImageView`).

**A pending Glide request can overwrite a PDF thumbnail.** `showTypeIconOnly` calls `Glide.with(context).clear(ivPreview)`, but the PDF branch goes straight to `ivPreview.setImageBitmap(bitmap)` without clearing. Glide only cancels a prior request on a view when a new request targets that same view, so a holder rebound from an image to a PDF keeps the old load in flight; when it lands it replaces the rendered page. A `clear()` at the top of the PDF path closes it.

**This is the third copy of the same code.** `showPdfPreview` is near-verbatim `InlineResourceAdapter.showPdfPreview` (`ui/courses/InlineResourceAdapter.kt:241`) — same `scale = 2`, same `LruCache`, same cache-key-then-`?.also { put }` structure, same swallowed `catch` — and the `File(dir, "ole/${'$'}{model.library.id}/${'$'}address")` path is already built in `VoicesAdapter` (lines 917 and 933) and `ResourceViewerFragment`. With #15582 adding cover loading to a third adapter, this is the moment to extract a shared preview loader and a `FileUtils` path helper rather than fork it again. The bitmap sizing above is inherited from that copy, so fixing it centrally fixes both call sites.

**Nit:** rendering onto a fresh `createBitmap` leaves a transparent background — PDFs that rely on the page's white backdrop will composite oddly; `eraseColor(Color.WHITE)` before `render` is the usual guard. `DefaultDispatcherProvider()` as a default argument on the `BaseAdapterFactory` interface also quietly re-introduces a concrete dependency the injection was meant to remove, and `catch (e: Exception)` binds `e` without using it.

**Verdict:** needs another round. Not a duplicate of #15582 despite the adjacent subject matter — that one loads course cover attachments in `CoursesAdapter` via Glide only, with no overlapping files.

---

dogi:
**Review — rating: 76/100**

Strong bones. Hoisting PDF rendering into a shared `PdfThumbnailLoader` is the right call, and three details in it are genuinely better than what `InlineResourceAdapter` had: the cache key is `path + lastModified + length + targetWidthPx`, so a re-downloaded file invalidates *and* the grid (84dp), list (44dp) and inline (240dp) variants can't collide on one entry; rendering to `targetWidthPx` replaces the old hardcoded `scale = 2`, which allocated a bitmap at twice the PDF's intrinsic page size no matter how small the view was; and `bitmap.eraseColor(Color.WHITE)` fixes pages with transparent backgrounds, which `PdfRenderer` otherwise leaves as uninitialised memory. `FileUtils.getLibraryFile` is good deduplication and quietly fixes a real defect — `InlineResourceAdapter.onCurrentListChanged` previously built `File(dir, "ole/${prev.id}/${prev.resourceLocalAddress}")` with a nullable address, so a null interpolated the literal string `"null"` into the path and the cache eviction silently matched nothing. The cancellation story is sound: per-holder `previewJob`, cancelled from both `setPreviewJob` and `onViewRecycled`, with `adapterScope` recreated in `onAttachedToRecyclerView`. `DispatcherProvider` is injected rather than hardcoded, per the style guide. CI green (build default + lite, test default).

**The blocker is a stale-thumbnail bug on the PDF path.** `bindCover` resets the cover for every branch *except* PDF:

```kotlin
mimeType?.contains("pdf") == true -> {
    val targetWidthPx = (coverWidthDp * context.resources.displayMetrics.density).toInt()
    adapterScope.launch { showPdfPreview(ivPreview, ivTypeIcon, file, targetWidthPx) }
}
```

Nothing touches `ivPreview` synchronously. The `Glide.clear` and the visibility reset live inside `showPdfPreview`, which only runs after `adapterScope.launch` dispatches on `Dispatchers.Main` — at minimum one frame later, and much longer on a cache miss, since that path does `ParcelFileDescriptor.open` + `PdfRenderer` on IO. So a holder recycled from an image or video resource onto a PDF keeps painting **the previous item's photo**, center-cropped, until the render lands. In a mixed library that's the common scroll case, and it's the feature this PR exists to add. Calling `showTypeIconOnly(ivPreview, ivTypeIcon)` before the `launch` fixes it in one line.

Three smaller things:

- **Uppercase extensions get no preview at all.** `Utilities.getMimeType` → `FileUtils.getFileExtension` → `File(it).extension`, which preserves case, and `MimeTypeMap.getMimeTypeFromExtension` only matches lowercase. A resource stored as `report.PDF` or `photo.JPG` falls through to the `else` branch. That's a pre-existing weakness in the util, but this PR makes it load-bearing for a user-visible feature — a `.lowercase()` in `getFileExtension` would fix it everywhere.
- **The `PdfThumbnailLoader` cache is process-wide with no trim hook.** It's an `object`, so up to `maxMemory/16` of bitmaps persists for the process lifetime and across activity restarts, where the old per-adapter cache was evicted in `onDetachedFromRecyclerView`. Sharing across adapters is the point, so this is a deliberate trade — but an `onTrimMemory`/`ComponentCallbacks2` hook to `evictAll()` would make it safe on low-RAM devices, which are squarely this app's target.
- **The two layouts round corners differently.** Grid uses `ShapeableImageView` with `ShapeAppearance.MyPlanet.CoverTopRounded`; list uses a plain `ImageView` under `clipToOutline="true"`. The list one does work — `bg_rounded_thumbnail` is a `<shape>` with `<corners>`, so it supplies an outline — but two mechanisms for one job invites drift.

**Nit:** `DiskCacheStrategy.ALL` on a `load(file)` of a local file writes the source back to Glide's disk cache; `RESOURCE` caches just the transformed output. Also worth noting the feature lands only on `ResourcesAdapter` — `TeamResourcesAdapter` is a separate class, so team resources keep the flat colour tiles. And no tests: `FileUtils.getLibraryFile` and the cache-key construction are both trivially unit-testable without an Android PDF stack.

**Verdict:** needs another round — one synchronous reset in the PDF branch and this is ready.

---

J-S-webskas (dismissed):
I really like the idea of the thumbnail. But the thumbails don't show unless the user clicks a video and exit. Sometimes, the thumbnail doesn't show for some video after exiting.

https://github.com/user-attachments/assets/d0f96be7-3aec-430b-9d3d-7ee14b9fec31

Okuro3499:
> I really like the idea of the thumbnail. But the thumbails don't show unless the user clicks a video and exit. Sometimes, the thumbnail doesn't show for some video after exiting.
> 
>  Recording_20260813_135949.mp4

@J-S-webskas This happens because the thumbnail only appears once the download has finished — that's the only point at which it becomes available. Videos download progressively as they play, so you can exit the video before the download finishes and it will keep downloading in the background, but the thumbnail won't display until that download is complete

J-S-webskas (approved):

dogi:
**Review — rating: 82/100**

Strong work. The memory story is the real content here, and it's better than the title suggests.

**Three separate wins on bitmap memory:**

1. **One shared cache instead of one per adapter.** `InlineResourceAdapter` held its own `LruCache` sized at `maxMemory / 8`; every additional adapter instance meant another cache of that size. `PdfThumbnailLoader` is a single object cache at `maxMemory / 16`, shared with `ResourcesAdapter`.
2. **Bitmaps sized to the view, not the page.** The old render was `page.width * 2` — for a letter-size PDF that's a ~1224px-wide bitmap held in cache regardless of where it was shown. Now it's scaled to 84dp (grid), 44dp (list), or 240dp (inline). That's the bulk of the saving.
3. **`onTrimMemory` eviction.** A process-lifetime bitmap cache needs a pressure hook, and `TRIM_MEMORY_RUNNING_LOW` is the right threshold.

Not recycling on `evictAll()` is also correct — a view may still be displaying an evicted bitmap, and on API 26+ the pixel data is native-heap and GC-managed.

**Two fixes that aren't in the description:**

- `bitmap.eraseColor(Color.WHITE)` before `page.render(...)`. `PdfRenderer` composites onto whatever's in the bitmap and doesn't clear it, so PDFs with transparent backgrounds previously rendered over uninitialized pixels. This is a real rendering fix.
- `getFileExtension` now returns `.lowercase()`. I checked every caller: `BaseContainerFragment:233-236` compares against `"mp4"`, `"mp3"`, `"aac"`, `"wav"`, so a file named `CLIP.MP4` was silently not detected as video. `ResourceDetailFragment:166` already called `.lowercase()` on the result (now redundant, harmless), and `FileUtilsTest` only asserts lowercase inputs. So this fixes a real case-sensitivity bug with no caller breakage — worth a line in the PR description, since it's easy to miss in a ten-file diff.

**The cache key is right** — `absolutePath + lastModified + length + targetWidthPx` means a re-downloaded resource or a different cover size can't serve a stale bitmap. That's also why dropping the manual `bitmapCache` invalidation in `invalidateCaches` is safe.

**Lifecycle handling is correct.** `adapterScope` is recreated in `onAttachedToRecyclerView` if previously cancelled, cancelled in `onDetachedFromRecyclerView`, and `onViewRecycled` cancels the per-holder preview job so a recycled holder can't receive a late bitmap for the wrong row.

**One thing to fix — main-thread disk I/O on every bind:**

```kotlin
suspend fun firstPageBitmap(file: File, dispatcherProvider: DispatcherProvider, targetWidthPx: Int): Bitmap? {
    if (targetWidthPx <= 0) return null
    val cacheKey = "${file.absolutePath}_${file.lastModified()}_${file.length()}_$targetWidthPx"
    cache.get(cacheKey)?.let { return it }
    return withContext(dispatcherProvider.io) { … }
}
```

`lastModified()` and `length()` are `stat` calls, and they run *before* the dispatcher hop — on `dispatcherProvider.main`, since that's the scope `showPdfPreview` launches in. So every bind does two filesystem syscalls on the UI thread, including cache hits, which are the common case while scrolling. That's the exact class of work this PR is removing elsewhere, and StrictMode would flag it.

Moving the whole body inside `withContext(dispatcherProvider.io)` fixes it — `LruCache` is internally synchronized, so reading it off the main thread is fine.

**Nits:**

- `BaseAdapterFactory.createResourcesAdapter` gains `dispatcherProvider` as a required parameter positioned between two defaulted ones (`viewMode`, `onEditClick`). Fine for named-argument callers, awkward for anyone implementing the interface.
- No tests for `PdfThumbnailLoader`. Rendering needs a real PDF, but the `targetWidthPx <= 0` guard and the cache-key composition are testable if the key builder is extracted.
- 10 commits behind `master`. Merges clean — I test-merged it — just needs the merge before landing.

**Verdict:** `ready`. Move the two `stat` calls behind the IO dispatcher and this is unambiguous; everything else here is a clear improvement over what it replaces.

---

dogi (approved):

PR #15582, opened 2026-08-12
courses: smoother grid cover imaging (fixes #15575)

### LINKED ISSUE #15575: Display course covers in the courses list
Courses list should display cover images/thumbnails for each course instead of showing only text entries.

### Goal
- Show course covers in the courses list.
- Make the list visually richer and easier to scan.
- Ensure missing covers have a sensible fallback state.

Okuro3499:
fixes #15575
Screenshot_20260812_172331
Screenshot_20260812_172324

## Summary by CodeRabbit

- **New Features**
  - Course cards now display personalized cover images in both grid and list views.
  - Covers support local and online loading with placeholders and fallback imagery.
  - Updated course covers refresh automatically when their revision changes.
  - Rounded cover layouts improve presentation and maintain existing subject icons when no cover is available.


dogi:
**Review — rating: 85/100**

Tight and correct. The local-file-then-remote-attachment fallback is the right ordering for an offline-first app, and the `GlideUrl` + `LazyHeaders` construction matches how the rest of the codebase authenticates against CouchDB rather than inventing a new path.

The detail I checked most closely is cache correctness, and both halves hold up. `MyCourse.courseRev` is `@ColumnInfo(name = "_rev")`, i.e. the CouchDB document revision, so `.signature(ObjectKey(course.courseRev.orEmpty()))` invalidates Glide's cached bitmap exactly when the course document changes — that's the correct key, and it's easy to get wrong by signing with the course id instead. Equally important, `coverFileName` and `courseRev` were both added to the adapter's content selector, so a cover that changes server-side actually produces a rebind instead of being swallowed by `DiffUtil`; adding the field to the model without that would have been a silent staleness bug. `MyCourse.getCoverImageFile` and `UrlUtils.getCourseImageUrl` both already exist, and the subject-icon fallback is preserved whenever neither source resolves.

Three small things, none blocking:

- `clipToOutline="true"` does what you want on `item_course_list.xml`, where `bg_rounded_thumbnail` is a uniform 8dp radius. It won't on the grid: `bg_cover_top_rounded` sets only `topLeftRadius`/`topRightRadius`, and `GradientDrawable` can only express a uniform corner radius as an `Outline`, so with per-corner radii the clip falls back to a plain rectangle and the cover's top corners stay square. A rounded `Glide` transform or a `ShapeableImageView` would be needed there.
- No `.placeholder()` on the load. Since `ivSubjectIcon` is hidden as soon as a model resolves, a remote cover leaves the container blank for the duration of the fetch — the subject icon would be a better placeholder than nothing.
- The `model == null` path hides `ivCover` without calling `Glide.with(context).clear(ivCover)`, so a request from the previous binding stays in flight against a recycled view. Harmless while the view is `GONE`, but clearing is the cheaper habit.

**Verdict:** mergeable.

Worth flagging for whoever picks these up: this and #15579 are independent features, not duplicates — different adapters, different issues, no shared files — but between them, plus the existing `InlineResourceAdapter` and `VoicesAdapter`, the codebase will have four separate implementations of "resolve a local resource file, else load remotely, else show a fallback icon". This PR's approach is the cleanest of them and is the natural basis for a shared helper.

---

dogi (approved):

PR #15587, opened 2026-08-13
resources: smoother viewer view modelling (fixes #15729)

### LINKED ISSUE #15729: Refactor: Move IO-bound work in ResourceViewer to ViewModel
This PR addresses the issue of moving IO-bound work (`lifecycleScope.launch { withContext(io) { ... } }`) out of the fragment and into the `ResourceViewerViewModel`.

**Changes made:**
- Created `getExternalFilesDir`, `downloadResource`, and `extractPdfText` as suspend functions in `ResourceViewerViewModel` that run on `dispatcherProvider.io`.
- Updated `ResourceViewerFragment` to call these ViewModel functions from its `viewLifecycleOwner.lifecycleScope.launch` default dispatcher.
- This represents one vertical slice as requested in the issue description.

Tests have been run and passed.

dogi:
This PR addresses the issue of moving IO-bound work (`lifecycleScope.launch { withContext(io) { ... } }`) out of the fragment and into the `ResourceViewerViewModel`.

**Changes made:**
- Created `getExternalFilesDir`, `downloadResource`, and `extractPdfText` as suspend functions in `ResourceViewerViewModel` that run on `dispatcherProvider.io`.
- Updated `ResourceViewerFragment` to call these ViewModel functions from its `viewLifecycleOwner.lifecycleScope.launch` default dispatcher.
- This represents one vertical slice as requested in the issue description.

Tests have been run and passed.

---
*PR created automatically by Jules for task [4727003253466733739](https://jules.google.com/task/4727003253466733739) started by @dogi*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resource viewing by moving file access, downloading, and PDF text extraction into background processing.
  * Improved reliability when downloading resources and extracting PDF text, including safer handling of failures.

* **Chores**
  * Updated the Android app version to 0.64.67.


dogi:
**Review — rating: 72/100**

The best thing in here isn't in the title. `extractPdfText`'s launcher changes from `lifecycleScope.launch(dispatcherProvider.io)` to `viewLifecycleOwner.lifecycleScope.launch` — the old code used the *fragment's* scope, which outlives the view across a detach/reattach, so a PDF extraction in flight could resume and write `pdfText`/`isExtractingText` against a torn-down view. That's a real lifecycle fix, and folding the trailing `withContext(dispatcherProvider.main) { isExtractingText = false }` into the now-Main-dispatched launch body is the correct simplification of it. The three extracted methods keep `withContext(dispatcherProvider.io)` rather than hardcoding `Dispatchers.IO`, per the style guide. CI green (build default + lite, test default).

**But it pushes `Context` into the ViewModel, which is the wrong direction for this codebase.** All three new methods take a `Context` parameter — `getExternalFilesDir(context)`, `downloadResource(context, url)`, `extractPdfText(context, file)`. Passing it per-call rather than retaining it does dodge the classic leak, so this isn't a bug. The problem is that #15629, open in the same review queue right now, is explicitly *removing* Android `Context` from the layers underneath the UI. Landing this as-is sets the opposite precedent in a file that PR will likely have to revisit. `@ApplicationContext` injected into the ViewModel's constructor gets the same IO off the fragment without threading `Context` through three public method signatures — and it's what `extractPdfText` actually wants anyway, since it immediately calls `context.applicationContext`.

Two concrete cleanups before this lands:

- **Fully-qualified PDFBox references.** `com.tom_roush.pdfbox.android.PDFBoxResourceLoader.init(...)`, `com.tom_roush.pdfbox.pdmodel.PDDocument.load(file)` and `com.tom_roush.pdfbox.text.PDFTextStripper()` are all spelled out inline in a file that has a perfectly good import block. The mirror image of this is that `ResourceViewerFragment.kt:50-52` still imports those exact three classes, and after this PR removes lines 434-436 they are its only usages — so the change leaves three dead imports in the fragment while writing FQNs in the ViewModel. Moving the imports across fixes both halves.
- **Import ordering.** The new `java.io.File` / `kotlinx.coroutines.withContext` / `DownloadUtils` / `FileUtils` imports are inserted above `javax.inject.Inject`, breaking the file's existing sorted block.

**Nit:** `PDDocument.load(file)` is still closed only on the success path — if `PDFTextStripper().getText()` throws, the `catch` swallows it and the document never closes, leaking the file handle. Pre-existing, but the code is being moved anyway and `PDDocument` is `Closeable`, so `.use { }` is a one-line fix. The `catch (e: Exception)` also binds `e` without using it. And no tests: extracting these into the ViewModel is what makes them testable, so `downloadResource`'s guard logic is worth a case with `TestDispatcherProvider` now that it's reachable.

**Verdict:** needs another round — the lifecycle fix is worth keeping, but constructor-inject `@ApplicationContext` instead of threading it through the signatures, and clean up the imports on both sides.

---

dogi:
**Re-review — Rating: 62/100**

CI green on all three checks. The `ResourceViewer` change itself is sound, but it's roughly 47 lines inside a 232-line diff spanning 39 main-source files, and the extraction introduces three fully-qualified names where the original had proper imports.

**37 of the 39 files contain only import reordering**

I checked each file for changes outside import lines. `BaseDashboardFragment`, `ServiceModule`, `UserEntity`, `CoursesRepository(Impl)`, `HealthRepository(Impl)`, `LifeRepositoryImpl`, `PersonalsRepositoryImpl`, `ResourcesRepositoryImpl`, `RetryRepositoryImpl`, `UserRepository(Impl)`, `ThemeManager`, `UploadManager`, `SyncManager`, `TransactionSyncManager`, `LeadersViewModel`, `CoursesFragment`, `DashboardViewModel` and the rest all come back **zero** — same imports removed and re-added in a different order.

Tidying import order is fine, but bundled here it means a reviewer has to diff 39 files to find the two that matter, and the PR will conflict with anything else touching those files — which, given the current queue, is most of them. Splitting the sweep into its own PR would make both halves trivially reviewable.

**The extraction dropped three imports in favour of FQNs**

```kotlin
com.tom_roush.pdfbox.android.PDFBoxResourceLoader.init(context.applicationContext)
val document = com.tom_roush.pdfbox.pdmodel.PDDocument.load(file)
val text = com.tom_roush.pdfbox.text.PDFTextStripper().getText(document).trim()
```

`ResourceViewerFragment` imported all three and used the short names; moving the code into `ResourceViewerViewModel` left the imports behind. Add them to the ViewModel and this reads as `PDFBoxResourceLoader.init(...)`, `PDDocument.load(file)`, `PDFTextStripper()`.

**`Context` passed into ViewModel methods.** `getExternalFilesDir(context)`, `downloadResource(context, url)` and `extractPdfText(context, file)` all take a `Context` from the caller. Every one of them is application-context-safe, so `@ApplicationContext private val context: Context` in the constructor would let all three lose the parameter — the same shape #15658 settled on for `ConfigurationsRepository`. As written the fragment still has to supply what it was supposed to stop knowing about.

**Import ordering in the new ViewModel:** `java.io.File`, `kotlinx.coroutines.withContext` and the two `org.ole.planet.myplanet.utils.*` entries are inserted above `javax.inject.Inject`.

**The core change is right.** Hoisting PDF text extraction, the `getExternalFilesDir` call and the download check into ViewModel methods that own their own `withContext(io)` is correct, and switching `lifecycleScope` to `viewLifecycleOwner.lifecycleScope` for the extraction coroutine fixes a real scope bug — `isExtractingText = false` was previously assignable after the view was gone.

**Verdict:** good change buried in an unrelated sweep. Split out the import reordering, restore the three PDFBox imports, and inject `Context` rather than passing it — then this is a small, obviously-correct PR.

---

dogi:
**Review — rating: 74/100**

Sound, narrowly-scoped move of three IO blocks out of `ResourceViewerFragment` into `ResourceViewerViewModel`. CI green on all three checks. The context injected is `@ApplicationContext`, so holding it in a ViewModel is safe, and `PDFBoxResourceLoader.init(context)` is equivalent to the old `requireContext().applicationContext`. Worth noting for anyone reading the title as a perf fix: the work was *already* off the main thread via `withContext(dispatcherProvider.io)` — this buys testability and layering, not throughput. That's still worth doing.

Two things to fix.

**1. Fully-qualified names instead of imports** — `ResourceViewerViewModel.downloadResource`:

```kotlin
if (!org.ole.planet.myplanet.utils.FileUtils.checkFileExist(context, url)) {
    org.ole.planet.myplanet.utils.DownloadUtils.openDownloadService(context, arrayListOf(url), false)
}
```

Both need an `import` and the short name. The file already imports from `org.ole.planet.myplanet.utils` (`DispatcherProvider`), so there's no collision to work around — add `FileUtils` and `DownloadUtils` alongside it and the body reads as two plain calls.

**2. `PDDocument` leaks when extraction throws** — `extractPdfText`:

```kotlin
val document = PDDocument.load(file)
val text = PDFTextStripper().getText(document).trim()
document.close()
```

If `getText()` throws — corrupt or encrypted PDF, which is exactly what the `catch` is there for — `close()` is skipped and the native buffers stay held until GC. The old code had the same shape, but this PR is rewriting the block, so it's cheap to fix properly:

```kotlin
PDDocument.load(file).use { PDFTextStripper().getText(it).trim() }
```

`PDDocument` implements `Closeable`, so `use` handles both paths and the line count drops.

**Nit:** `isExtractingText` (fragment line 101) is written at both ends of this function and read nowhere in the file — it's a dead flag. Since you're already inside this function, deleting it removes two lines and one piece of misleading state.

**Nit:** branch is 3 commits behind `master`; no conflicts, just needs a merge before it lands.

**Verdict:** the refactor itself is right and I'd take it as-is on correctness grounds. Fix the two FQNs (project convention is import + short name, not inline qualification) and switch to `use {}`, and this goes straight to `ready`.

---

dogi:

All three findings addressed, and each one properly rather than minimally.

**FQNs → imports.** `FileUtils` and `DownloadUtils` are now imported and the call sites read as plain calls. The imports also landed in the right slot — `DispatcherProvider`, `DownloadUtils`, `FileUtils` in order within the `utils` group.

**`PDDocument` no longer leaks:**

```kotlin
PDDocument.load(file).use {
    PDFTextStripper().getText(it).trim()
}
```

This is the right shape. `use` returns the lambda's value so the function still yields the trimmed text, and on a throw it closes the document *before* propagating to the outer `catch`, which is exactly the corrupt/encrypted-PDF path the `catch` exists for. `load` throwing on its own is fine too — no document to close in that case.

**Dead flag gone.** `isExtractingText` is removed at the declaration and both write sites, rather than left in place "just in case". That's the version I'd hoped for.

**Verified unchanged and still correct:** the injected `Context` is `@ApplicationContext`, so holding it in a ViewModel is safe and `PDFBoxResourceLoader.init(context)` matches the old `requireContext().applicationContext`.

**Nit:** the branch is 3 commits behind `master`. No conflicts — just needs a merge before it lands.

**Observation, not a request:** removing the flag makes visible something that was already true — there's no UI signal while a large PDF is being extracted, so tapping read-aloud early speaks an empty string. `isExtractingText` was never read, so this isn't a regression and nothing here caused it, but it's the natural follow-up if anyone reports it.

**Verdict:** `ready`. Clean scoped refactor, correct resource handling, CI green on all three checks.

---

Okuro3499 (approved):

PR #15590, opened 2026-08-13
courses: smoother empty state controling (fixes #15571)

### LINKED ISSUE #15571: Hide progress filter, search, and layout switch when Courses list is empty
## Summary
When there are no courses to display, showing filter/search/view controls is unnecessary and can confuse users.

## Expected behavior
If the courses list is empty, do not show:
- Progress filter
- Search input
- Layout switch

https://github.com/user-attachments/assets/baff4262-6995-41d5-ab80-354aa5331d4b

ragilzakaria:
fixes #15571 
Screenshot_20260813-201740

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the My Course view’s handling of search, chip filters, and list/grid controls across screen sizes and orientations.
  * Ensured the correct course view state is restored when opening the screen.
  * Improved visibility toggling for course controls when switching between list and grid views.


dogi:
**Review — rating: 81/100**

The fix is right on both halves. `setListVisible` was hiding `etSearch` and the filter button while leaving the pill container, the chip-filter scroller and the grid/list toggle on screen — so an empty course list left a stripe of padded, empty chrome behind. Hiding the containers instead of the leaf views is the correct level to operate at, and the ids are added consistently to all three `fragment_my_course.xml` variants (portrait, `layout-land`, `layout-sw600dp`), which is the part most easily forgotten. Putting `@+id/layout_view_toggle` inside `view_grid_list_toggle.xml` rather than on each `<include>` is the better choice for a shared layout. The `checkList()` call added after `selectionController.clearAll(null)` is the other half: that block is the initial-load coroutine, so without it the first render evaluated visibility from stale state and only corrected on the next `coursesState` emission. CI green (build default + lite, test default).

**The defensive scaffolding is unnecessary and one piece of it is wrong.** Every lookup gets a `?:` fallback *and* a second `if (x == null)` fallback at the use site:

```kotlin
layoutSearch = rootView.findViewById(R.id.layout_search) ?: (etSearch.parent as? View)
```

None of those branches can execute. `CourseFilterController` is constructed in exactly one place (`CoursesFragment.kt:197`), `CoursesFragment` inflates `fragment_my_course.xml`, and this PR adds `layout_search` and `scroll_chip_filter` to all three variants of it; `layout_view_toggle` resolves because all six library/course layouts `<include>` `view_grid_list_toggle`. So it's six dead branches guarding ids the same commit guarantees.

Worse, the `etSearch.parent` fallback wouldn't be equivalent if it ever did run. In `layout_search_pill.xml` the root `LinearLayout` holds `tl_tags`/`flexbox_tags` *and* a separate inner container that wraps the search icon and `et_search` (line 52). `etSearch.parent` is that inner box, not the pill root — so the fallback would hide the search field but leave the pill's padding and tag area visible, which is the exact bug this PR set out to fix. Dead code that silently does the wrong thing is worse than no fallback: if someone later drops the id from a layout, this fails quietly instead of loudly. Deleting all six branches makes the method four lines and strictly more correct.

⚠️ **Textual conflict with #15591:** both add the identical `android:id="@+id/layout_view_toggle"` line to `view_grid_list_toggle.xml`. Same line, same content, so whichever lands second needs at most a trivial resolution — but they're separate fixes for separate issues, not duplicates, and both should land.

**Verdict:** mergeable — I'd strip the unreachable fallbacks first, but they don't change runtime behaviour.

---

Okuro3499 (approved):

dogi (approved):

PR #15591, opened 2026-08-13
resources: smoother list grid toggling (fixes #15572)

### LINKED ISSUE #15572: Hide search and layout switch when resources list is empty
## Summary
When there are no resources to display, showing search/view controls is unnecessary and can confuse users.

## Expected behavior
If the resources list is empty, do not show:
- Search input
- Layout switch
https://github.com/user-attachments/assets/f22b8fe0-5851-41bc-b4cb-9aa03277a6ad

ragilzakaria:
Screenshot_20260813-203035

fixes #15572

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6525
-        versionName = "0.65.25"
+        versionCode = 6526
+        versionName = "0.65.26"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/ui/resources/ResourcesFragment.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/resources/ResourcesFragment.kt
@@ -71,11 +71,13 @@ class ResourcesFragment : BaseRecyclerFragment<MyLibrary?>(), OnLibraryItemSelec
     private val binding get() = _binding!!
     private val tvAddToLib get() = binding.tvAdd
     private val tvSelected get() = binding.tvSelected
+    private val layoutSearch get() = binding.layoutSearch.root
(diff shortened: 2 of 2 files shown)


dogi:
**Review — rating: 84/100**

The resources-side twin of #15590, and the tighter of the two. Same real bug: `checkList()` toggled `etSearch` alone, so an empty library left the search pill's padding and the grid/list toggle rendered over nothing. Switching to `binding.layoutSearch.root` is exactly right — `layout_search_pill.xml`'s root `LinearLayout` also holds `tl_tags`/`flexbox_tags`, so hiding the `EditText` alone was always going to leave chrome behind, and ViewBinding already exposes the include's root here so no new id was needed. Both branches of `checkList()` are updated symmetrically, which is the thing to get wrong in a visibility fix. CI green (build default + lite, test default).

Two notes, neither blocking:

- **The fallback is unreachable.** `?: (toggleGridButton?.parent as? View)` can't fire: this PR puts `@+id/layout_view_toggle` on the `LinearLayout` inside `view_grid_list_toggle.xml`, and all three `fragment_my_library.xml` variants `<include>` that layout, so the lookup always resolves. Unlike the equivalent fallback in #15590 this one at least *is* semantically correct — `toggle_grid` is a direct child of that `LinearLayout`, so `.parent` is the same view — but it's still a branch that exists to handle a state the same commit rules out.
- **`findViewById` in a getter.** `layoutViewToggle` re-walks the view tree on every access, and `checkList()` touches it once per call. It matches the local convention (`toggleGridButton` at line 242 does the same), so I don't think it should change in isolation — but giving the `<include>` in `fragment_my_library.xml` an id would let both of these become plain `binding.*` lookups and drop the `findViewById` and the fallback together.

⚠️ **Textual conflict with #15590:** both add the identical `android:id="@+id/layout_view_toggle"` line to `view_grid_list_toggle.xml`. Identical content, so resolution is trivial — they fix different issues (#15572 vs #15571) on different screens and both should land, but sequence them deliberately rather than letting merge order decide.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #15592, opened 2026-08-13
teams: smoother survey title ordering (fixes #15576)

### LINKED ISSUE #15576: survey: fix order by title
https://github.com/user-attachments/assets/7708dc71-e9b2-4692-919a-5df595fff675

When we go to survey, we can sort or order by title. but if I click Order by Title for the first time, the order is from Z-A, not A-Z. Maybe it is better if either:
- make two order by title like order by date, which are ascending and descending, or;
- for the first time clicking order by title, it should make A-Z

ragilzakaria:
https://github.com/user-attachments/assets/d2a640c4-8782-43b5-aaf6-b0dc504af46a

fixes #15576 

## Summary by CodeRabbit

- **New Features**
  - Added separate ascending and descending title-sorting options for surveys.
  - Updated sorting labels across supported languages, including Arabic, Spanish, French, Nepali, and Somali.

- **Bug Fixes**
  - Improved survey sorting controls by clearly distinguishing title order options.

- **Chores**
  - Incremented the Android app version to 0.64.51.


dogi:
**Review — rating: 86/100**

Right fix for the right reason. The old design hid the sort direction behind re-selecting an already-selected spinner item — an interaction with no affordance, no indicator of current direction, and which needed a custom `onSameItemSelected` hook on `CustomSpinner` to work at all. Two explicit entries make the state visible and idempotent, and `SortOption` already had `TITLE_ASC`/`TITLE_DESC`, so this is pure wiring rather than new sort logic.

I checked the thing most likely to break: `R.array.sort_by_date` grows from three items to four, so any other spinner bound to it would silently gain an unhandled index. It has exactly two consumers — `fragment_survey.xml:65` (`android:entries`) and `SurveyFragment.kt:128` (the `ArrayAdapter`) — both the same spinner, so extending the array is safe. All six locales are updated (`values`, `ar`, `es`, `fr`, `ne`, `so`), and in each one the new items are in ASC-then-DESC order, matching the `when` branches `2 -> TITLE_ASC` / `3 -> TITLE_DESC`. Getting that ordering inverted in one translation file would have been an invisible bug in that locale only, so it's worth saying it's consistent across all six. The Somali entry also quietly corrects "Ururinta Magaca" (by *name*) to "Ururinta Cinwaanka" (by *title*). CI green (build default + lite, test default).

**Two pieces of code go dead and neither is removed:**

- `SurveysViewModel.toggleTitleSort()` (line 114) loses both its production callers — `SurveyFragment` lines 152 and 164 were the only ones. It survives only because `SurveysViewModelTest.kt:117-132` still exercises it, so the suite stays green while testing something nothing calls. Either drop the method and that test, or keep it deliberately. Worth being precise here: `CoursesViewModel.toggleTitleSort()` is a *different* method and is still live via `CoursesFragment.kt:326`, so don't let a cleanup sweep take that one too.
- `CustomSpinner.onSameItemSelected()` (line 15) had exactly one caller, the block this PR deletes. The whole re-selection mechanism on `CustomSpinner` is now unreferenced.

**Nit:** within one spinner the date pair reads Desc-then-Asc while the new title pair reads Asc-then-Desc. Harmless, but the four entries would scan better in a consistent direction order.

**Verdict:** mergeable — the dead code is worth a follow-up but doesn't affect behaviour.

---

Okuro3499 (approved):

dogi (approved):

PR #15594, opened 2026-08-13
enterprises: smoother finances landscaping (fixes #15577)

### LINKED ISSUE #15577: community: finance menu is cropped and cant be scrolled in landscape mode
Image

ragilzakaria:
https://github.com/user-attachments/assets/22d8906a-31d9-4e00-8f43-22c65173118e

fixes #15577

## Summary by CodeRabbit

* **New Features**
  * Added finance date-range filtering, reset controls, balance warnings, transaction headings, totals, and ascending/descending title sorting.
  * Added a scrollable transaction-entry dialog with date selection and image attachment controls.

* **Bug Fixes**
  * Improved scrolling on transaction-entry and finance screens, especially on smaller displays.
  * Preserved existing transaction entry and finance tracking functionality.

* **Localization**
  * Added transaction date-range labels, selection prompts, and expanded sorting options in Arabic, Spanish, French, Nepali, Somali, and English.


dogi:
**Review — rating: 72/100**

The bug is real and the diagnosis is right. `fragment_finance.xml` stacked a date-range row, a caution line, a column-header row and a totals bar above `rv_finance`, which was `layout_height="match_parent"` inside a plain `LinearLayout` — so in landscape the header block consumed the viewport and the list had nowhere to go, with nothing scrollable. Fixing it with one `NestedScrollView` + `fillViewport="true"` rather than adding a `layout-land` duplicate is the better call: one layout to maintain instead of two that drift. I checked the binding impact, since the `data_layout` id moves from the outer container to the inner `LinearLayout` — `EnterprisesFinancesFragment` touches `binding.dataLayout` at lines 345/349/353 and `binding.rvFinance` at 239/240/347/351/355, all of which still resolve to the same views, and toggling `dataLayout` still hides exactly the same content. `add_transaction.xml` is inflated via `AddTransactionBinding.inflate` into `AlertDialog.setView`, so a `ScrollView` root at `wrap_content` is the right shape there — the dialog still sizes to content but scrolls once the fields plus the 120dp image preview overflow a landscape dialog. CI green (build default + lite, test default).

**The cost is that `rv_finance` stops recycling.** `wrap_content` height inside a scrolling parent plus `android:nestedScrollingEnabled="false"` means `LinearLayoutManager` measures with an unspecified height and lays out *every* row up front — no view reuse, no windowing. The `nestedScrollingEnabled="false"` is the correct companion for this pattern, so this looks deliberate, but on a community finance ledger the row count is open-ended: it grows with every transaction ever recorded and there's no date-range default bounding it (the From/To filters start empty and there's a Reset button). On the low-end offline devices this app targets, a few hundred transaction rows all inflated and retained is exactly the case RecyclerView exists to avoid.

The fix that keeps both properties is to stop treating the header as layout and make it content: move the date row, caution, column headers and totals bar into the `RecyclerView` as header view types (or a `ConcatAdapter`), drop the `NestedScrollView`, and let `rv_finance` scroll everything at `match_parent`. That fixes landscape cropping *and* keeps recycling. If instead you know the row count is small and bounded, say so on the PR and I'll take this as-is — but that's a claim that should be written down rather than implied.

**Nits:**
- `add_transaction.xml` loses its `<?xml version="1.0" encoding="utf-8"?>` prolog. Harmless to the build, but every other layout in the project has one and this reads as an accident of the rewrite.
- The date fields carry hardcoded English — `android:hint="From Date"`, `"To Date"`, `contentDescription="Select From Date"`/`"Select To Date"`. Pre-existing, but this PR rewrites those exact lines in a project shipping five translations, so it's a cheap moment to move them into `strings.xml`.

**Verdict:** needs another round — the scroll fix is correct, but not at the price of laying out the whole ledger at once.

---

dogi:
**Review — rating: 74/100**

Moving the filter/totals header into a `ConcatAdapter` row so it scrolls with the transactions is the right approach, and the supporting work is tidy: `add_transaction.xml` → `dialog_add_transaction.xml` matches the project's `dialog_*` naming convention, `HeaderState` cleanly replaces reading UI state back out of `TextView`s, and all four new strings landed in every one of the six locale files.

**The empty state regressed.** Before:

```kotlin
if (results.isNotEmpty()) { … rvFinance VISIBLE }
else if (fromDate.isNullOrEmpty() && toDate.isNullOrEmpty()) { tvNodata VISIBLE; rvFinance GONE }
else { tvNodata VISIBLE; rvFinance GONE }   // filtered to nothing → still says "no data"
```

After:

```kotlin
if (results.isNotEmpty() || headerState.fromDate.isNotEmpty() || headerState.toDate.isNotEmpty()) {
    binding.tvNodata.visibility = View.GONE
    binding.rvFinance.visibility = View.VISIBLE
} else { … }
```

So filtering to a date range with no transactions now shows an empty list and **no** "no data" message — the user gets a header and blank space with no indication that the filter matched nothing.

I can see why it was written this way: the header now lives inside `rvFinance`, so the old `rvFinance.visibility = GONE` would hide the filter controls and strand the user with no way to clear them. That constraint is real, but the fix conflates two independent decisions. They want to be separate:

- `rvFinance` — always `VISIBLE`, because it now carries the header.
- `tvNodata` — `VISIBLE` whenever `results.isEmpty()`, filtered or not.

**`notifyItemChanged(0)` will animate the header.** The default `ItemAnimator` cross-fades on a change notification, and the header is re-notified on every date pick, every sort toggle, every reset, and every `calculateTotal` — which runs on each list update. That's a visible flicker of the whole filter bar on routine refreshes. Either pass a payload (`notifyItemChanged(0, PAYLOAD_HEADER)`, which suppresses the change animation) or set `supportsChangeAnimations = false` on the RecyclerView's animator.

**Listeners are re-registered on every bind.** `bindHeader` attaches six `setOnClickListener`s each time it runs. Functionally fine — they replace rather than stack — but combined with the notify frequency above it's six lambda allocations per refresh. Attaching them once in `onCreateViewHolder` (against `headerState`, which the fragment already owns) would separate one-time wiring from per-bind state.

**Nit — unrelated change bundled in:** the Spanish and French `note_is_required` strings are reworded ("Se requiere una nota" → "La nota es obligatoria", "La note est requise" → "Une note est requise"). Probably an improvement, but it's a translation edit inside a layout-restructuring PR and it isn't mentioned anywhere.

I checked that `dataLayout` is fully removed — no dangling references in `fragment_finance.xml` or the enterprises package. Clean removal.

**Verdict:** `change`. The header extraction is good work worth landing. Split the `rvFinance` and `tvNodata` visibility decisions so a filter that matches nothing still says so, and quiet the header animation.

---

dogi:
**Re-review — Rating: 84/100** (was 74)

All three findings landed, and two of them landed better than I asked.

**The empty state is split correctly now.** I said `rvFinance` and `tvNodata` needed to be independent decisions — `rvFinance` always visible because it carries the header, `tvNodata` visible whenever the result set is empty regardless of filtering. That's exactly the shape:

```kotlin
binding.rvFinance.visibility = View.VISIBLE
binding.tvNodata.visibility = if (results.isEmpty()) View.VISIBLE else View.GONE
```

So a date range matching nothing now says so instead of showing blank space.

**The header animation is quieted twice over** — `(binding.rvFinance.itemAnimator as? SimpleItemAnimator)?.supportsChangeAnimations = false` *and* the `PAYLOAD_HEADER` on every `notifyItemChanged`. I offered those as alternatives; taking both is belt-and-braces rather than a fault. Worth knowing the payload can't do more than it currently does, though: `FinanceHeaderAdapter` overrides both `onBindViewHolder` overloads with identical bodies (`onBindHeader(holder.binding)`), so there's no partial-bind path for a payload to select. Either give the payload overload a narrower bind, or drop it and rely on the animator setting alone.

**Listeners moved to `onCreateViewHolder`**, via `onCreateHeader`, with `bindHeader` reduced to pure state application. That's the separation I was after — one-time wiring vs per-bind state — and it removes the six lambda allocations per refresh.

**One regression to restore before merging.** Dropping `showNoData(binding.tvNodata, transactions.size, "finances")` from the transactions collector also dropped the message it was setting. `showNoData` does two things:

```kotlin
v.visibility = if (count == 0) View.VISIBLE else View.GONE
val messageRes = noDataMessages[source] ?: R.string.no_data_available_please_check_and_try_again
textView.setText(messageRes)
```

and `noDataMessages["finances"]` is `R.string.no_finance_record`. The new `fragment_finance.xml` hardcodes `android:text="@string/no_data_available"` instead, so the screen went from "No finance record" to the generic "No data available". The fix is to put the call back in `updatedFinanceList` in place of the manual visibility line — it handles both the visibility and the right string, so it's a swap rather than an addition.

**`PAYLOAD_HEADER` is declared as an instance property:**

```kotlin
private val PAYLOAD_HEADER = "PAYLOAD_HEADER"
```

SCREAMING_SNAKE_CASE signals a compile-time constant; as written it's allocated per fragment. `private const val` in a `companion object` matches the convention and the intent.

**Two merges to sequence, and one of them isn't mechanical.**

Master has since landed `a8c444f` ("sync: smoother time logger date formatting"), which converted this exact file from `SimpleDateFormat`/`ParseException` to `java.time` — `DateTimeFormatter.ofPattern(...).withZone(...)`, `dateFormatter.format(this.toInstant())`, `DateTimeParseException`. This branch still has the old `SimpleDateFormat` version of `formatToString` and `parseDate`. Resolve toward master's `java.time` implementation; a naive "keep ours" reintroduces `SimpleDateFormat`.

And **#15768 in this same batch conflicts directly** — I confirmed it. That PR's `resetFilterAndSort()` calls `updateToDateState(false)` and touches `binding.imgDate` and `isAsc`, all three of which this PR deletes in favour of `HeaderState`. Whichever lands second needs its reset rewritten against `headerState`, not a textual merge. Given this PR is the larger restructure, landing it first and rebasing #15768 onto it is the cheaper order.

**Still bundled in:** the Spanish and French `note_is_required` rewordings are unchanged from last round and still unmentioned. Fine to keep — just call them out in the description so a translator reviewing later knows they were intentional.

CI green on all three checks.

**Verdict:** `ready`. Restore the `showNoData` call and make `PAYLOAD_HEADER` a `const`, then merge master in — the `java.time` conflict is the one to be careful with.

---

Okuro3499 (approved):

dogi (approved):

PR #15595, opened 2026-08-13
teams: smoother member requests view modelling (fixes #15578)

### LINKED ISSUE #15578: enterprise: make Join Requests not visible to non-member
Image

I'm requested to join a enterprise, it still says requested, but I can see the join request. i think the join request should only be visible to admin or at least the members.

ragilzakaria:
Screenshot_20260813-215920

fixes #15578

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/ui/teams/members/MembersFragment.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/teams/members/MembersFragment.kt
@@ -88,7 +88,7 @@ class MembersFragment : BaseTeamFragment() {
                 launch {
                     requestsViewModel.uiState.collect { state ->
                         requestsAdapter?.setData(state.members, state.isLeader, state.memberCount)
-                        val hasRequests = state.members.isNotEmpty()
+                        val hasRequests = state.isMember && state.members.isNotEmpty()
                         binding.llRequestsSection.visibility = if (hasRequests) View.VISIBLE else View.GONE
                         if (hasRequests) {
                             binding.tvRequestsHeader.text = getString(R.string.join_requests) + " (${state.members.size})"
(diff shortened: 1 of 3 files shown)


dogi:
**Review — rating: 84/100**

Correct and appropriately paranoid for what is really a privacy fix — join requests expose other users' names, and they were rendering for anyone who could open the team. Three things I checked specifically:

**Parameter order.** The two adjacent repository calls have *opposite* signatures — `isMember(userId: String?, teamId: String)` at `TeamsRepository.kt:97` versus `isTeamLeader(teamId: String, userId: String?)` at line 98 — and this PR calls both correctly. Worth noting for anyone reading it later that a swap wouldn't compile in either direction, since `user?.id` is `String?` and each function's non-null slot rejects it. So the inconsistent ordering is a readability wart in the repository API rather than a live hazard, but it's the sort of thing that reads as a bug on a skim.

**The mid-record field insertion.** `isMember` goes in as the *third* property of `RequestsUiState`, ahead of `memberCount`, and the existing construction was positional. Converting that call to named arguments is the right defensive move — a `Boolean`/`Int` mismatch would have failed the build here, but named args mean the next field added to this state class can't silently shift anything.

**The other writer of `_uiState`.** `respondToRequest` builds its optimistic state with `originalState.copy(...)`, so `isMember` carries through untouched and accepting or rejecting a request can't collapse the section. That was the most plausible place for this change to break, and it doesn't.

The default `isMember = false` also means the section stays hidden until `fetchMembers` resolves, so it fails closed rather than flashing requests during load — the right default for a visibility gate. Skipping `getRequestedMembers` entirely for non-members is a nice secondary win: the data is never fetched, not merely not displayed.

**The gap is that the new behaviour is the one thing not tested.** `RequestsViewModelTest` gains `coEvery { teamsRepository.isMember(any(), any()) } returns true` in `setUp`, which exists purely to keep the *existing* cases green — every test still runs the member path. There's no case pinning `isMember = false` → `members` empty, `isMember` false in the emitted state, and `getRequestedMembers` never called (`coVerify(exactly = 0)`). This ViewModel already has a test class and the project runs `testDefaultDebugUnitTest` on every push, so that's a handful of lines for the assertion that actually guards the privacy behaviour. Without it, someone flipping the `if (isMember)` back to an unconditional fetch breaks nothing.

**Nit:** `hasRequests = state.isMember && state.members.isNotEmpty()` is redundant — the ViewModel already forces `members` to `emptyList()` when `isMember` is false, so the first clause can never change the result. Harmless belt-and-braces, but it does leave two sources of truth for one rule.

**Verdict:** mergeable — please add the non-member test case, since it's the whole point of the change.

---

PR #15596, opened 2026-08-13
teams: smoother voices replying (fixes #15792)

### LINKED ISSUE #15792: Refactor VoicesAdapter to remove UserRepository dependency
Refactor VoicesAdapter to remove UserRepository dependency

Removed `UserRepository` from `VoicesAdapter` constructor and replaced it with a pre-parsed `leadersList: List<UserEntity>`. The leaders list is now parsed inline at the adapter construction sites (`VoicesFragment`, `ReplyActivity`, `TeamsVoicesFragment`) using `userRepository.parseLeadersJson` and the result is passed directly to the adapter. This removes a data layer dependency from the UI adapter layer.

dogi:
Refactor VoicesAdapter to remove UserRepository dependency

Removed `UserRepository` from `VoicesAdapter` constructor and replaced it with a pre-parsed `leadersList: List<UserEntity>`. The leaders list is now parsed inline at the adapter construction sites (`VoicesFragment`, `ReplyActivity`, `TeamsVoicesFragment`) using `userRepository.parseLeadersJson` and the result is passed directly to the adapter. This removes a data layer dependency from the UI adapter layer.

---
*PR created automatically by Jules for task [13368324851304894580](https://jules.google.com/task/13368324851304894580) started by @dogi*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the display of community leaders in Voices and Teams Voices.
  * Ensured leader information loads consistently when viewing voices, replies, and team voices.
  * Improved the reliability of leader information shown alongside voice content.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6526
-        versionName = "0.65.26"
+        versionCode = 6527
+        versionName = "0.65.27"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/ui/teams/voices/TeamsVoicesFragment.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/teams/voices/TeamsVoicesFragment.kt
@@ -249,8 +249,7 @@ class TeamsVoicesFragment : BaseTeamFragment() {
                     onAnimateTyping = VoicesAdapterHelper.createOnAnimateTyping(viewLifecycleOwner.lifecycleScope, dispatcherProvider),
                     labelManager = labelManager,
                     voicesRepository = voicesRepository,
-                    userRepository = userRepository,
-                    getCommunityLeadersFn = { sharedPrefManager.getCommunityLeaders() },
+                    leadersList = userRepository.parseLeadersJson(sharedPrefManager.getCommunityLeaders()),
                     setRepliedNewsIdFn = { sharedPrefManager.setRepliedNewsId(it) }
                 )
             }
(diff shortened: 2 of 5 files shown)


dogi:
**Review — rating: 82/100**

Right direction. An adapter holding a repository plus a `() -> String` prefs accessor means it can't be constructed or reasoned about without the data layer; passing in the already-parsed `List<UserEntity>` makes `VoicesAdapter` take data instead of the means to fetch data. All three construction sites (`VoicesFragment`, `TeamsVoicesFragment`, `ReplyActivity`) are updated consistently, and `parseLeadersJson` is a plain non-suspend `org.json` parse wrapped in try/catch — it returns an empty list on unset prefs rather than throwing — so hoisting the call to the caller is safe. CI green (build default + lite, test default).

Four things, none blocking:

- **`import org.ole.planet.myplanet.repository.UserRepository` at `VoicesAdapter.kt:37` is now dead.** Removing the constructor field and the line-173 usage leaves nothing referencing it in that file.
- **The cleanup is half-applied.** `voicesRepository` stays on the constructor (line 65, used at 485 and 745), so the adapter still reaches into the data layer — just through a different door. That's fine as a scoped change, but the PR title reads like the adapter is now repository-free and it isn't.
- **The expression is copy-pasted three times.** `userRepository.parseLeadersJson(sharedPrefManager.getCommunityLeaders())` appears verbatim at each construction site, and `VoicesAdapterHelper` already exists precisely to hold shared adapter-construction logic (`createOnAnimateTyping`). One helper there would keep the three sites honest with each other.
- **The snapshot moves earlier.** `by lazy` computed the leaders on first access during binding; it's now computed when the adapter is constructed. Both are compute-once, so neither refreshes after a later sync — that staleness is pre-existing, not introduced here. But the window widens: an adapter built during fragment setup while a login sync is still writing community leaders would previously have picked up the fresh value at first bind and now won't. Worth a thought given `VoicesFragment` explicitly handles a `fromLogin` path. The flip side is a small win — you also drop `by lazy`'s default synchronised access on every bind.

⚠️ **Conflicts with #15608** ("Optimize VoicesAdapter bind path"), which edits the same file and the same bind path where `leadersList` is consumed. Not duplicates — different jobs — but whichever lands second will need a real rebase, not a trivial one, so sequence them deliberately.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #15597, opened 2026-08-13
all: smoother regex normalizing (fixes #15730)

### LINKED ISSUE #15730: Hoist per-call Regex compilations and extract normalization logic
- Hoisted regex compilations in `TTSManager.kt` to companion object constants.
- Avoided `.toRegex()` compilation in `VersionUtils.kt` by using string replacements.
- Extracted and centralized `DIACRITICS_REGEX` into `Utilities.kt`.
- Refactored `MyLibrary.kt`, `ChatViewModel.kt`, and `SurveysViewModel.kt` to use the unified `Utilities.normalizeText()` helper.

dogi:
- Hoisted regex compilations in `TTSManager.kt` to companion object constants.
- Avoided `.toRegex()` compilation in `VersionUtils.kt` by using string replacements.
- Extracted and centralized `DIACRITICS_REGEX` into `Utilities.kt`.
- Refactored `MyLibrary.kt`, `ChatViewModel.kt`, and `SurveysViewModel.kt` to use the unified `Utilities.normalizeText()` helper.

---
*PR created automatically by Jules for task [6477484253693433030](https://jules.google.com/task/6477484253693433030) started by @dogi*


dogi:
**Review — rating: 87/100**

I checked all ten hoisted patterns in `stripMarkdown` character by character — pattern, `RegexOption.MULTILINE` flag, replacement string and application order all survive unchanged, so the transformation is behaviour-preserving. That one matters because `stripMarkdown` runs per TTS utterance and was compiling ten `Regex` objects on every call. `Utilities.normalizeText` is the same story from the other direction: `ChatViewModel` maps it over every conversation query and response (lines 154-159) and both view models run it per query token, so hoisting that one regex is a tight-loop win.

`VersionUtils` is correct and worth spelling out, because it looks like a semantic change and isn't: Kotlin's `String.replace(String, String)` is a *literal* replacement, unlike Java's `replaceAll`. So `raw.replace("v".toRegex(), "")` → `raw.replace("v", "")` and `"\\.".toRegex()` → `"."` both preserve exact behaviour while dropping two per-call regex compilations. The two `DIACRITICS_REGEX` constants deleted from `ChatViewModel` and `SurveysViewModel` were genuinely dead — nothing else in either file references them, which the green build confirms.

**The `MyLibrary` change is the most valuable thing here and the PR undersells it as extraction.** `titleNormal` is an indexed column (`@Entity(... indices = [Index("titleNormal")])`) and it had *two different* normalisations writing to it. `ResourcesRepositoryImpl` writes it via `Utilities.normalizeText` at lines 148 and 234, and the search path normalises the query the same way before `titleNormal LIKE ?` (line 111) and `titleNormal.startsWith(normalizedQuery)` (line 121). But `MyLibrary.kt:238` — the sync/deserialise path — did NFD → strip diacritics → `lowercase(Locale.ROOT)`, i.e. the opposite order and a different locale from everything else. So resources arriving through sync were indexed under one scheme and searched under another. For plain Latin text the two orders coincide, so this mostly hid; in any locale where case mapping differs it wouldn't. Unifying on `normalizeText` closes that. Please say so in the description — it's a search fix, not a tidy-up.

Two consequences worth acting on:

- **Existing rows keep the old value.** `titleNormal` is persisted, so anything already stored by the old sync path stays normalised the old way until the record is re-synced. Given the project's drop-and-resync strategy this resolves on the next schema bump, but until then the two schemes coexist in one table. Worth deciding whether that needs a forced resync.
- **`normalizeText` uses `Locale.getDefault()` for a persisted index.** That's pre-existing, and this PR makes it *more* consistent rather than less — but a stored search index whose contents depend on the device locale at write time is fragile (the Turkish dotted/dotless I is the standard example, and it changes both what's written and what's queried). `Locale.ROOT` would make the index locale-stable. Follow-up, not this PR.

**Nits:** `TABLE_PIPE_REGEX = Regex("\\|")` is a regex for a single literal character — `replace('|', ' ')` uses the `Char` overload and skips the matcher entirely. And `SurveysViewModel` is left with a double blank line where the constant used to be.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #15598, opened 2026-08-13
courses: smoother repository flowing (fixes #15731)

### LINKED ISSUE #15731: Refactor Flow chains in CoursesRepositoryImpl
Refactored Flow chains in CoursesRepositoryImpl to avoid redundant emissions by utilizing `distinctUntilChanged` and optimized off-thread execution by replacing inline `withContext` usage with the idiomatic `flowOn` operator.

dogi:
Refactored Flow chains in CoursesRepositoryImpl to avoid redundant emissions by utilizing `distinctUntilChanged` and optimized off-thread execution by replacing inline `withContext` usage with the idiomatic `flowOn` operator.

---
*PR created automatically by Jules for task [344420228444936317](https://jules.google.com/task/344420228444936317) started by @dogi*


dogi:
**Review — rating: 76/100**

The `flowOn` work is correct and is the better half of this PR. Replacing the `withContext(dispatcherProvider.io)` *inside* `map` with `.flowOn(dispatcherProvider.io)` on the chain is the canonical idiom — the old form worked but confined only the block body, while `flowOn` covers the upstream operators too, and it removes a suspension boundary per emission. The composition with `getCourseByCourseIdFlow` is also well-formed rather than accidental: that function now ends in `.flowOn(default)`, so `mapCourse` runs on Default, and the downstream aggregation in `getCourseDetailModelFlow` picks up `.flowOn(io)` for the repository fan-out. Each `flowOn` governs the segment above it, so the two don't fight. `dispatcherProvider.default` exists on the interface, so nothing is hardcoded. CI green (build default + lite, test default).

**`distinctUntilChanged()` on `getMyCoursesFlow` does nothing.** `MyCourse` is declared `open class MyCourse(` at `MyCourse.kt:16` — not a `data class` — and the file defines no `equals` override. `distinctUntilChanged()` with no selector falls back to `Any.equals`, which for `List<MyCourse>` means an element-wise comparison that bottoms out in reference identity. Room constructs fresh entity instances on every invalidation, and `mapCourses(...).filter { ... }` builds a new list on top of that, so consecutive emissions can never compare equal. The operator therefore walks the whole list on every emission and suppresses nothing.

That's the operator whose entire purpose is to suppress redundant emissions, so it should either work or come out. Two ways to make it real:

```kotlin
    old.size == new.size && old.zip(new).all { (a, b) -> a.id == b.id && a.courseTitle == b.courseTitle /* … */ }
}
```

or give `MyCourse` a value-based `equals`/`hashCode` — but note that's a wider change, since `MyCourse` is a Room `@Entity` used as a `ListAdapter` payload elsewhere, and adding `equals` would silently alter `DiffUtil` behaviour across the courses UI. The explicit-comparator form is the contained option.

**No tests.** `CoursesRepositoryImplTest` already exists and the project ships `TestDispatcherProvider` precisely so dispatcher-confinement changes are assertable. A case that collects `getMyCoursesFlow` across two identical DAO emissions and asserts one downstream emission would have caught the no-op above before review.

**Verdict:** needs another round — keep the `flowOn` changes, make `distinctUntilChanged` actually distinct or drop it.

---

dogi:
**Re-review — Rating: 74/100**

CI green on all three checks. The `withContext`-inside-`map` → `flowOn` conversion is exactly right, but the custom `distinctUntilChanged` predicate will swallow real updates.

**The `flowOn` work is correct.** `withContext(dispatcherProvider.io)` nested inside `.map { }` is an anti-pattern — it dispatches per emission and doesn't move the upstream producer. Replacing it with `.flowOn(dispatcherProvider.io)` on the chain shifts everything upstream in one place, which is what the operator exists for. Four chains converted consistently, with `.default` for the CPU-bound mapping chains and `.io` for the one that hits repositories. Good distinction.

**The equality predicate compares four fields out of many**

```kotlin
    old.size == new.size && old.zip(new).all { (a, b) ->
        a.id == b.id && a.courseTitle == b.courseTitle && a.userId == b.userId && a.courseSteps?.size == b.courseSteps?.size
    }
}
```

Returning `true` means "these are equal — **drop the new emission**." So any change confined to a field not listed here never reaches the UI: `description`, `createdDate`, `numberOfSteps`, `gradeLevel`, `subjectLevel`, `_rev`, and the contents of `courseSteps` (only the count is compared, so an edited step title looks identical).

`courseDao.observeAll()` re-emits on any write to the courses table, so this is the path a sync uses to push updated course metadata into an open list. With this predicate, a synced description change silently doesn't render until something else forces a rebuild.

Two ways out, both fine:

- Compare `_rev` instead of guessing at fields. It's CouchDB's change marker, so `a.id == b.id && a._rev == b._rev` catches every server-side change in one comparison, and you can keep the local-only fields alongside it.
- Or drop the predicate. `MyCourse` is an `open class` `@Entity` without `equals`, so plain `distinctUntilChanged()` falls back to identity and would filter almost nothing — which is why the custom one exists — but it's worth checking whether the re-emission was actually costly enough to need filtering, given `mapCourses` now runs on `default` rather than the collector's thread.

Under-inclusive equality is the failure mode that's hardest to notice later: nothing crashes, the screen is just stale.

**Verdict:** the dispatcher work is a clear improvement and worth landing. Widen the equality check — `_rev` is the cheap correct answer — and this is ready.

---

dogi:

CI green on all three checks. The equality predicate now keys on the revision marker:

```kotlin
a.id == b.id && a.courseRev == b.courseRev && a.userId == b.userId
```

That closes the hole. `courseRev` changes on every server-side write, so a synced update to `description`, `numberOfSteps`, or a step's contents now produces a different comparison and the emission passes through — where the previous field-by-field list would have declared them equal and dropped it. One field replacing four, and strictly more correct.

The `flowOn` work was already right and is unchanged: `withContext` inside `.map { }` gone in favour of `flowOn` on the chain, with `.default` for the CPU-bound mappings and `.io` for the chain that hits repositories.

**Verdict:** correct on both counts. Marking `ready`.

---

Okuro3499 (approved):

PR #15599, opened 2026-08-13
courses: smoother resources caching (fixes #15793)

### LINKED ISSUE #15793: Cache list adapters in ResourcesFragment and CoursesFragment
Cached list adapters in `ResourcesFragment` and `CoursesFragment` to avoid unnecessary rebuilds and loss of user state.

dogi:
Cached list adapters in `ResourcesFragment` and `CoursesFragment` to avoid unnecessary rebuilds and loss of user state.

---
*PR created automatically by Jules for task [14447802560373906274](https://jules.google.com/task/14447802560373906274) started by @dogi*


dogi:
**Review — rating: 68/100**

The goal is sound — rebuilding the adapter on every reload throws away `RecyclerView` state and forces a full rebind, and reuse composes well with the `onAttachedToRecyclerView`/`onDetachedFromRecyclerView` scope handling `ResourcesAdapter` already has. `setViewMode` is also the right refresh hook: I checked both implementations and each short-circuits on an unchanged mode and calls `notifyDataSetChanged()` otherwise, so the view-mode path genuinely rebinds rather than silently deferring to the next submit. `setListener(...)` is still re-invoked outside the guard, so the callback target stays current. CI green (build default + lite, test default).

**But `viewMode` is not the only thing that can change between calls, and it's the only thing refreshed.** Both adapters take their other inputs as immutable constructor `val`s, and there are no setters for them:

- `ResourcesAdapter.isGuest` (line 30) drives `checkbox.visibility = if (isGuest) View.GONE else View.VISIBLE` at line 209 and gates selection at 236 and 246; `CoursesAdapter.isGuest` (line 41) branches at line 310.
- `ResourcesAdapter.currentUserName` and both adapters' `context` are likewise frozen at first construction.

The code immediately above the guard re-reads the user on every call — `userModel = userSessionManager.getUserModel()` in `CoursesFragment`, `val user = userRepository.getUserModel()` in `ResourcesFragment`. That re-fetch only makes sense if the user can change between invocations, and this PR then throws the fresh value away on every call after the first. A guest who signs up through `BecomeMemberActivity` and returns to the same fragment instance keeps guest-mode chrome: checkboxes stay hidden, selection stays disabled. That is the same failure mode #15528 is currently open to fix from the other direction, so it's a live concern in this queue, not a hypothetical.

The fix mirrors what's already there — give `isGuest`/`currentUserName` the `setViewMode` treatment (backing `var`, no-op on unchanged, `notifyDataSetChanged()` otherwise) and refresh them in the `else` branch alongside the view mode.

**The `Mutex` should come out.** `withLock` guards a block containing no suspension points, mutating a field only ever touched from main-thread coroutines in these two fragments. Coroutines on `Dispatchers.Main` can't interleave inside a non-suspending block, so there is no race for it to prevent — the `!::adapterCourses.isInitialized` check is already the whole guard. What it does add is a suspension point around the block and a non-reentrant lock: if this path is ever re-entered from inside the critical section it deadlocks instead of recursing. That's a real footgun bought for no protection.

**Nit:** the cached `context = requireActivity()` now outlives `onDestroyView`. For a back-stack pop within the same Activity that's harmless, and non-retained fragments are recreated alongside the Activity on configuration change, so I don't think there's a live leak here — but an adapter holding an Activity across view-lifecycle boundaries is worth a comment saying why it's safe.

**Verdict:** needs another round — refresh `isGuest`/`currentUserName` with the view mode and drop the mutex.

---

dogi:

CI green on all three checks. Caching the adapter across `onCreateView` is the right idea and the `isGuest` handling is done properly — but the diff keeps a `notifyDataSetChanged()` that a `ready` PR is removing from the same method.

**Direct conflict with #15654**

That PR (currently `ready`) changes `CoursesAdapter.setViewMode` from `notifyDataSetChanged()` to `notifyItemRangeChanged(0, itemCount)`. This PR rewrites the same method and keeps `notifyDataSetChanged()`:

```kotlin
fun setViewMode(mode: ListViewMode, isGuest: Boolean, onChanged: (() -> Unit)? = null) {
    …
    if (changed) { notifyDataSetChanged() }
}
```

They can't both land as written, and the merge order determines which behaviour survives — silently. Since #15654 is approved and this one changes the signature, the cleanest resolution is to rebase this onto it and keep `notifyItemRangeChanged(0, itemCount)` inside the new `if (changed)` guard. That composes: the guard avoids the notify entirely when nothing changed, and the range notify avoids the full-adapter reset when something did.

**The `isGuest` handling itself is good.** Making it a `var` and folding it into `setViewMode` is what makes adapter caching possible at all — a cached adapter would otherwise keep a stale guest flag after a user change. Tracking `changed` across both fields so the notify fires once rather than twice is the right shape.

**On the leak comment:**

```kotlin
// The adapter caches the Context (Activity) which outlives onCreateView,
// but Fragments and their host Activities are re-created together so this is safe from leaks.
```

That reasoning holds, and it's worth having written down. The case it's really covering is back-stack navigation — `onDestroyView` runs while the fragment instance and its adapter survive, but the Activity is still alive, so the retained `Context` is current. On a configuration change the fragment is destroyed alongside the activity (absent `setRetainInstance`), taking the adapter with it. Both paths are safe.

**Verdict:** correct approach, and adapter caching is a real improvement for back-stack returns. Rebase onto #15654 so the `notifyDataSetChanged()` removal isn't undone by merge order.

---

dogi:

CI green on all three checks. The conflict I flagged is still present, and the diff has grown slightly rather than resolving it.

**Still `notifyDataSetChanged()`, and #15654 is `ready`**

`CoursesAdapter.setViewMode` in this branch:

```kotlin
fun setViewMode(mode: ListViewMode, isGuest: Boolean, onChanged: (() -> Unit)? = null) {
    …
    if (changed) { notifyDataSetChanged() }
}
```

#15654 changes that same call to `notifyItemRangeChanged(0, itemCount)` and is already approved. Whichever merges second silently decides which survives — and if this one lands after, the approved improvement is reverted without anyone reviewing the reversion.

`ResourcesAdapter.setViewMode` now also grows a `currentUserName` parameter alongside `isGuest`, so the signature diverges further from what #15654 touches. That widens the conflict rather than narrowing it.

The resolution is unchanged and small: rebase onto #15654 and keep `notifyItemRangeChanged(0, itemCount)` inside the new `if (changed)` guard. They compose well — the guard skips the notify entirely when nothing changed, and the range notify avoids the full-adapter reset when something did.

**The rest is still good.** Making `isGuest` a `var` folded into `setViewMode` is what makes adapter caching viable at all, since a cached adapter would otherwise keep a stale guest flag across a user change. Tracking `changed` across both fields so the notify fires once is right, and the same treatment for `currentUserName` in `ResourcesAdapter` is consistent.

The leak reasoning in the comment holds: on back-stack navigation the fragment and adapter survive `onDestroyView` while the Activity is still alive, and on configuration change the fragment is destroyed alongside the Activity, so the retained `Context` is never stale.

**Verdict:** correct approach, unresolved merge hazard. Rebase onto #15654 and this is ready.

---

dogi:
**Re-review — Rating: 78/100** (was 76)

My previous finding is fixed: `setViewMode` no longer calls `notifyDataSetChanged()`, and both adapters now use `notifyItemRangeChanged(0, itemCount)`. Thanks for that.

The PR has also grown a second, larger change — caching the adapters across `onCreateView` — so most of this review is about the new material.

**The caching is worth having, and the leak comment is correct.** I checked the reasoning rather than taking the comment at face value:

```kotlin
// The adapter caches the Context (Activity) which outlives onCreateView,
// but Fragments and their host Activities are re-created together so this is safe from leaks.
```

That holds up. The only path where `::adapterCourses.isInitialized` is true on a second `onCreateView` is a surviving Fragment instance — back stack, or detach/attach — and in every one of those the host Activity is the same instance. A configuration change destroys the Fragment along with the Activity, so `isInitialized` is false and a fresh adapter is built against the new Activity. No stale-context path. Keeping the adapter also keeps `AsyncListDiffer`'s backing list, so returning to the fragment no longer re-diffs from empty — that's the real win here.

Making `isGuest` (and `currentUserName`) mutable follows from the caching: a cached adapter would otherwise carry stale identity if the user changed while the Fragment instance survived. Right call.

**Blocking — the view-mode toggle is now asynchronous, and it didn't need to be:**

```kotlin
override fun onViewModeChanged(mode: ListViewMode) {
    prefManager.setLibraryViewMode(mode)
    updateToggleUi(mode)
    if (::adapterLibrary.isInitialized) {
            val user = userRepository.getUserModel()      // suspend — hits Room
            adapterLibrary.setViewMode(mode, user?.isGuest() == true, user?.name)
        }
    }
}
```

`getUserModel()` is `suspend` on both `UserRepository` and `UserSessionManager`, so a grid/list toggle now waits on a DB read before the list re-lays out, while `updateToggleUi(mode)` has already flipped the button synchronously. Two consequences:

1. The toggle button and the list are briefly inconsistent — worse the colder the read.
2. Two quick taps launch two coroutines with no ordering guarantee. If the second resolves first, the adapter ends up in the opposite mode from the button. Rare, but it's a real interleaving, not a theoretical one.

And the reason it's there doesn't hold: **toggling grid/list cannot change guest status or username.** Those only change on login/logout, which recreates the Activity and therefore the adapter. The identity refresh belongs in `onCreateView`'s else branch — where it already is, and where the user has just been read anyway — not on every toggle.

The fix is small: keep a synchronous `setViewMode(mode)` for the toggle path, and use the wider `setViewMode(mode, isGuest, currentUserName)` only at the `onCreateView` re-entry point. That also un-conflates the method, which currently does three things under a name that promises one.

**Still conflicts with #15654.** #15654 makes the identical `notifyDataSetChanged()` → `notifyItemRangeChanged(0, itemCount)` change in *both* `CoursesAdapter` and `ResourcesAdapter`, and it's labelled `ready`. This PR now changes the same lines *and* the method signature, so whichever lands second needs a manual fixup — and if #15599 lands first, #15654's remaining hunks still apply, but if #15654 lands first, this PR's signature change has to be reapplied by hand. #15654 has other content (the `PAYLOAD_SELECTION` re-indent), so it isn't simply subsumed. Worth deciding an order rather than discovering it at merge time.

**Nit:** two stray blank lines added — `CoursesFragment.kt` after the `viewModel` declaration, `ResourcesFragment.kt` in the same spot. Pure diff noise; please drop them.

**Verdict:** the `notifyDataSetChanged` fix and the adapter caching are both good and I'd take them. Move the identity refresh out of `onViewModeChanged` so the toggle stays synchronous, drop the two blank lines, and settle the ordering with #15654 — then this is `ready`.

---

dogi:

Fixed exactly as suggested, and the split is cleaner than what I proposed.

**The toggle is synchronous again.** `onViewModeChanged` no longer launches a coroutine to re-read the user before applying the mode:

```kotlin
if (::adapterCourses.isInitialized) {
    adapterCourses.setViewMode(mode)
}
```

That closes both problems in one move — no lag between `updateToggleUi` and the list, and no way for two quick taps to resolve out of order.

**`updateIdentity` is the right seam.** Splitting the identity refresh into its own method rather than overloading `setViewMode` means each call site says what it's doing, and `setViewMode` is back to a single responsibility. It's called only from the `onCreateView` re-entry path, where `userModel` has already been resolved — so the synchronous call is safe there:

```kotlin
adapterCourses.setViewMode(sharedPrefManager.getCourseViewMode())
adapterCourses.updateIdentity(userModel?.isGuest() ?: true)
```

`ResourcesAdapter.updateIdentity` correctly keeps the `changed` accumulator, since it guards two fields behind one notification; `CoursesAdapter` drops it for the single field. Right call in both.

**Stray blank lines gone** in both fragments — and the `ResourcesFragment` one took a trailing-whitespace line with it.

**Two nits, neither blocking:**

- `onChanged: (() -> Unit)? = null` is dead on both adapters. I grepped every call site: nothing passes it. Pre-existing, but you're rewriting both signatures anyway, so it's a free deletion.
- `setViewMode` still uses a `var changed` accumulator for a single condition. With `isGuest` moved out, `if (viewMode != mode) { viewMode = mode; notifyItemRangeChanged(0, itemCount) }` says the same thing in two lines.

**Still needs sequencing with #15654.** That PR (labelled `ready`) makes the same `notifyDataSetChanged()` → `notifyItemRangeChanged(0, itemCount)` change in both adapters, plus a `PAYLOAD_SELECTION` re-indent this PR doesn't have. Not a defect in either — they just overlap, and #15654 isn't subsumed, so someone needs to pick an order.

**Verdict:** `ready`. Adapter caching preserves the differ's list across `onCreateView`, the notify is incremental, and the toggle path is back to being immediate. CI green on all three checks.

---

Okuro3499 (approved):

PR #15601, opened 2026-08-13
sync: smoother file uploading streaming (fixes #15794)

### LINKED ISSUE #15794: Stream file uploads instead of allocating whole files in memory
Replaced `file.readBytes().toRequestBody()` with OkHttp's streaming `file.asRequestBody()` extension function in `UploadManager.kt` (for team image attachments) and `AchievementUploader.kt` (for CV PDF attachments). This change prevents large files from being fully loaded into memory before uploading, effectively addressing high memory usage and out-of-memory risks during syncs while preserving MIME types, URL encoding, and retry behaviors.

dogi:
Replaced `file.readBytes().toRequestBody()` with OkHttp's streaming `file.asRequestBody()` extension function in `UploadManager.kt` (for team image attachments) and `AchievementUploader.kt` (for CV PDF attachments). This change prevents large files from being fully loaded into memory before uploading, effectively addressing high memory usage and out-of-memory risks during syncs while preserving MIME types, URL encoding, and retry behaviors.

---
*PR created automatically by Jules for task [12183164733263910298](https://jules.google.com/task/12183164733263910298) started by @dogi*

## Summary by CodeRabbit

* **Improvements**
  * File uploads now use a more memory-efficient approach, reducing the need to load entire files into memory.
  * Updated attachment, team image, news image, and achievement uploads accordingly.

* **Release**
  * Android app version updated to 0.65.29.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6528
-        versionName = "0.65.28"
+        versionCode = 6529
+        versionName = "0.65.29"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/repository/UploadRepositoryImpl.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/UploadRepositoryImpl.kt
@@ -5,7 +5,7 @@ import java.io.File
 import javax.inject.Inject
 import javax.inject.Singleton
 import okhttp3.MediaType.Companion.toMediaTypeOrNull
-import okhttp3.RequestBody.Companion.toRequestBody
+import okhttp3.RequestBody.Companion.asRequestBody
 import org.ole.planet.myplanet.data.api.ApiInterface
 import org.ole.planet.myplanet.data.room.dao.AnswerDao
 import org.ole.planet.myplanet.data.room.dao.ExamDao
@@ -14,7 +14,6 @@ import org.ole.planet.myplanet.model.MembershipDoc
 import org.ole.planet.myplanet.model.StepExam
 import org.ole.planet.myplanet.model.Submission
 import org.ole.planet.myplanet.services.FileUploader
-import org.ole.planet.myplanet.utils.FileUtils
 import kotlinx.coroutines.withContext
 import org.ole.planet.myplanet.utils.DispatcherProvider
 import org.ole.planet.myplanet.utils.UrlUtils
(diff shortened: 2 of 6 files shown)


dogi:
**Review — rating: 79/100**

The substitution is correct and the reasoning behind it holds up under the two checks that usually sink this change. `File.asRequestBody()` reports `contentLength()` as `file.length()`, so `Content-Length` is still set and the request doesn't silently fall back to chunked transfer encoding — which matters here because these are CouchDB attachment `PUT`s. And it's not a one-shot body, so OkHttp can still replay it across redirects and retries exactly as the `ByteArray` version could. What changes is that the bytes now move through an 8 KiB buffer instead of being materialised as a single heap allocation, which is a real win on the low-RAM devices this app targets — a multi-megabyte team photo or CV no longer needs a contiguous array. CI green (build default + lite, test default).

**The sweep is half done, and one of the misses is in a file this PR already edits.** `UploadManager.kt:454-455`, about 86 lines below the line you changed, still does:

```kotlin
val fileBody = FileUtils.fullyReadFileToBytes(imageFile)
    .toRequestBody("application/octet-stream".toMediaTypeOrNull())
```

That's the resource-image attachment upload — plausibly the largest file of the four, since it's an arbitrary user-supplied image. `UploadRepositoryImpl.kt:118` does the same thing via `val body = FileUtils.fullyReadFileToBytes(file)`. And `FileUtils.fullyReadFileToBytes` is itself just `f.readBytes()` (line 48) whose only two callers are those upload paths — so converting both lets the helper be deleted in the same change, leaving no remaining way to accidentally re-introduce the pattern. As it stands the PR title promises the general fix and delivers two of four sites.

**Nit:** after this change `AchievementUploader.kt:53` is the only thing that used `toRequestBody` in that file, so the import at line 10 becomes dead. `UploadManager` keeps its import legitimately, because of line 455 — which is exactly the line that should also be converted.

One behavioural note worth knowing rather than acting on: streaming reads the file lazily, so a file truncated or deleted between the `exists()` check and the request now fails mid-flight with a content-length mismatch, where `readBytes()` took a snapshot up front. For photo and CV uploads that window is negligible, and the memory win is worth it.

**Verdict:** needs another round — finish the sweep (`UploadManager:454`, `UploadRepositoryImpl:118`, drop `fullyReadFileToBytes`) and this is straightforwardly mergeable.

---

dogi:
**Review — rating: 88/100**

CI green on all three checks. This is the most valuable change I've seen in the perf batch — it's not an optimization, it's an OOM fix.

```kotlin
- val body = FileUtils.fullyReadFileToBytes(file).toRequestBody("application/octet-stream".toMediaTypeOrNull())
+ val body = file.asRequestBody("application/octet-stream".toMediaTypeOrNull())
```

`readBytes()` materializes the entire file as a `ByteArray` on the heap before the upload begins. On the devices this app targets, uploading a video or a large PDF attachment that way is a plausible `OutOfMemoryError` — and the failure would look like a random upload crash rather than a memory problem. `asRequestBody` streams from the `File` as OkHttp writes the request, so peak memory is a buffer rather than the whole file.

Applied consistently across all four sites: `UploadRepositoryImpl.uploadAttachment`, the two in `UploadManager` (team image, resource image), and `AchievementUploader`'s CV upload.

**Two details that make this correct rather than just smaller:**

`asRequestBody` reports `contentLength()` from `file.length()`, so requests stay non-chunked and CouchDB still gets a `Content-Length` — worth knowing, because a `RequestBody` without a known length would have silently switched these to chunked transfer encoding.

Removing `FileUtils.fullyReadFileToBytes` and its test rather than leaving it orphaned is the right cleanup, and the compiler confirms no other callers.

**Minor:** because the body is now read at write time rather than call time, a file deleted or truncated between the call and the network write will fail differently than before — an IO error during upload instead of at read. Both end in the same catch; just noting the timing shift.

**Verdict:** fixes a real memory hazard, applied uniformly, dead helper removed. Marking `ready`.

---

Okuro3499 (approved):

PR #15602, opened 2026-08-13
resources: smoother content item callback diffing (fixes #15702)

### LINKED ISSUE #15702: Fix broken `areContentsTheSame` in `ResourcesAdapter`
Fix broken `areContentsTheSame` in `ResourcesAdapter` by adding an explicit `contentSelector` over the scalar fields, mirroring `CoursesAdapter`. This prevents full UI rebinds on every refresh.

Tracked in PR #15602.

dogi:
Fix broken `areContentsTheSame` in `ResourcesAdapter` by adding an explicit `contentSelector` over the scalar fields, mirroring `CoursesAdapter`. This prevents full UI rebinds on every refresh.

---
*PR created automatically by Jules for task [3695330053528556453](https://jules.google.com/task/3695330053528556453) started by @dogi*


dogi:
**Review — rating: 84/100**

The diagnosis is exactly right, and I traced it rather than taking the title's word. `DiffUtils.standardItemCallback` falls back to `oldItem == newItem` when no `contentSelector` is supplied. `ResourceListModel` *is* a `data class`, so that looks fine at a glance — but its generated `equals` delegates to each property, and `library: MyLibrary` is a Room `@Entity` declared as a plain `open class` with no `equals` override. Room builds a fresh `MyLibrary` on every emission, so that one field compares by reference and drags the whole comparison to `false` every time. `areContentsTheSame` was therefore constant-false and every visible row rebound on every submit. The explicit `contentSelector` sidesteps it, and it works because the fields it selects are value-comparable: `ResourceItem` and `TagItem` are both `data class`es, so `tags: List<TagItem>` compares element-wise rather than by reference. That was the detail most likely to quietly defeat the fix, and it holds.

**One thing to tighten before this lands.** Making `areContentsTheSame` sometimes-true is the moment field coverage starts to matter — until now, blanket rebinding accidentally kept everything fresh. Two rendered values read from `library`, which the selector never touches:

- `buildMetaLine` (line 193) reads `model.library.language` and derives its type label from `classify(model.library)`.
- `LibraryTypeClassifier.classify` (line 16) reads `library.resourceLocalAddress`, `library.resourceRemoteAddress` and `library.mediaType` — which drive both the cover colour and the type icon.

In practice `item._rev` covers the server-driven cases, since any CouchDB revision to `mediaType`/`language`/`resourceRemoteAddress` bumps it, and local download completion flips `isOffline`/`isLocallyOffline`, which are both selected. So I don't think there's a live staleness bug today. But the coverage is transitive and accidental rather than stated, and it'll break the first time someone mutates a `library` field without touching `_rev` or the offline flags. Adding `it.library.mediaType`, `it.library.language` and `it.library.resourceLocalAddress` makes the invariant explicit and costs nothing.

**Nit:** no test. `DiffUtils` is pure Kotlin with no Android dependencies, so a case asserting that two `ResourceListModel`s differing only in `library` identity now compare equal — and that changing `title` makes them differ — would pin this against regression cheaply.

**Verdict:** mergeable.

---

Okuro3499 (dismissed):
```
File: app/src/main/java/org/ole/planet/myplanet/ui/resources/ResourcesAdapter.kt, ITEM_CALLBACK (contentSelector, lines 49-61)                                                   
    - Problem: model.library is entirely excluded from content comparison, even though library.language, library.addedBy, and whatever fields LibraryTypeClassifier.classify reads   
  are used directly in bindGrid/bindList/bindClicks. Changes to those fields alone (no accompanying item.*/tags/isOpened/isLocallyOffline change) will be silently missed by         
  DiffUtil, leaving stale type icon, color, language text, or edit-menu eligibility on screen.
```

dogi:

CI green on all three checks. The broken `areContentsTheSame` now actually compares contents.

Adding a `contentSelector` to `DiffUtils.standardItemCallback` gives DiffUtil something to compare beyond identity, so a resource whose data changed while keeping its id will now rebind instead of being silently treated as unchanged. That was the defect.

**The field list is well chosen.** Including `_rev` is the key move — it's CouchDB's change marker, so any server-side field that isn't enumerated here still triggers a diff when the document is re-synced. That makes the explicit list a fast path rather than an exhaustive contract, which is the right way to build one of these: the enumerated fields catch purely local changes (`isOpened`, `isLocallyOffline`, `tags`), and `_rev` catches everything remote.

Keeping the existing `payloadSelector` alongside means the targeted `PAYLOAD_SELECTION` rebinds still apply where they can, so this doesn't push everything back to full binds.

**Minor:** `it.item.isOffline` and `it.isLocallyOffline` are both in the content list and both in the payload predicate — correct, but it does mean a change to either produces both a content difference and a payload, so RecyclerView takes the payload path. That's the intended behaviour; noting it because the duplication looks redundant at first glance and isn't.

**Verdict:** fixes the reported defect without giving up partial rebinds. Marking `ready`.

---

Okuro3499 (approved):

PR #15604, opened 2026-08-13
all: less database service module is more (fixes #15732)

### LINKED ISSUE #15732: Remove vestigial DatabaseService and DatabaseModule
Removes the unused `DatabaseService` and its related files, cleaning up dead code and adhering to the direct DAO injection pattern.

dogi:
Removes the unused `DatabaseService` and its related files, cleaning up dead code and adhering to the direct DAO injection pattern.

---
*PR created automatically by Jules for task [4904099468036069023](https://jules.google.com/task/4904099468036069023) started by @dogi*


dogi:
**Review — rating: 88/100**

I verified "vestigial" rather than taking it on trust, and it holds: after this PR the only references to `DatabaseService` anywhere in `main` or `test` are the files it deletes or edits — `DatabaseModule.kt` (deleted), `DatabaseServiceTest.kt` (deleted), and the two stale `mockk` declarations in `SyncManagerTest`/`TransactionSyncManagerTest`, which were never even passed to the constructors they sat beside. No production code injects it.

The one capability worth checking before deleting a DB wrapper is the table-wipe path, since `DatabaseService.clearAll()` was the only wrapper around `room.clearAllTables()`. That's safe here — `ConfigurationsRepositoryImpl.kt:201` already calls `appDatabase.clearAllTables()` directly, so the reset flow doesn't route through the deleted class. `withRoomAsync` and `executeRoomTransactionAsync` are likewise superseded by Room's own `withTransaction`, which is what the repositories already use.

Deleting `DatabaseServiceTest.kt` alongside is correct, not a coverage loss — those three cases tested the wrapper's own delegation, so they'd have nothing left to assert. Updating `CLAUDE.md` in the same commit is the right instinct and more thorough than most cleanups manage: the module list, the repository-pattern paragraph, the DAO rules note and the key-file table all get the reference removed. CI green (build default + lite, test default).

**Nit — the doc counts this PR invalidates aren't updated.** `di/` currently holds exactly 10 files and `CLAUDE.md` says so in three places; after deleting `DatabaseModule.kt` it's 9. All three need a decrement:

- the package table: `| di/ | Hilt dependency injection | 10 | …` → 9
- the DI section header: `**Module Structure (8 modules):**` → 7
- the inventory summary: `| di/ | 10 | Dependency injection (8 modules + 2 entry points) |` → `9 | … (7 modules + 2 entry points)`

For a PR whose whole purpose is removing things the docs already called vestigial, leaving three counts describing the deleted file is the same class of drift. (Separately and not your doing: the `data/` row claims 40 files where the directory actually holds 47, so that number was already stale before this change — worth a follow-up rather than folding in here.)

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #15605, opened 2026-08-13
teams: smoother voices repository reply bulk querying (fixes #15710)

### LINKED ISSUE #15710: Refactor `collectNewsAndReplies` using recursive CTE
Replaced the expensive recursive database calls in `VoicesRepositoryImpl.collectNewsAndReplies` for deleting news threads with a cycle-safe, recursive CTE returning only the IDs directly in `NewsDao`.

Tracked in PR #15605.

dogi:
Replaced the expensive recursive database calls in `VoicesRepositoryImpl.collectNewsAndReplies` for deleting news threads with a cycle-safe, recursive CTE returning only the IDs directly in `NewsDao`. Added a test for the recursive behavior in `NewsDaoTest.kt` and updated the repository mocks in `VoicesRepositoryImplTest.kt`.

---
*PR created automatically by Jules for task [3519019509259584427](https://jules.google.com/task/3519019509259584427) started by @dogi*


dogi:
**Review — rating: 88/100**

This replaces one DB round-trip *per node in the thread* with a single query, and it fixes two latent failure modes in the process rather than just being faster.

The old `collectNewsAndReplies` recursed in Kotlin, issuing `getDirectReplies` at every level, so a deep thread meant both unbounded stack depth and a query per reply. Worse, `replyTo` is an ordinary column with no constraint preventing a cycle, and on a cycle that recursion never terminates. The CTE uses `UNION` rather than `UNION ALL`, which dedupes and therefore reaches a fixpoint even on cyclic data — the recursion simply stops producing new rows. That's a genuine robustness gain, not just a rewrite.

I checked the two places this kind of translation usually goes wrong:

- **Case sensitivity.** `NewsDao` uses `COLLATE NOCASE` in eight places, and the adjacent `getReplyCount` matches `replyTo = :newsId COLLATE NOCASE`. But `getDirectReplies` (line 57) is plain `WHERE replyTo = :newsId`, and the new CTE's `news.replyTo = thread.id` is likewise case-sensitive — so this preserves the behaviour it replaces exactly. Worth knowing that `getReplyCount` and `getDirectReplies` already disagree on collation, meaning the displayed reply count and the set actually deleted can diverge on mixed-case ids. Pre-existing and out of scope here, but someone should decide which is right.
- **The base case.** The old code seeded `mutableListOf(newsId)` unconditionally, returning the id even when no such row exists; `SELECT :newsId` does the same. Faithful.

The tests are real: `getNewsAndRepliesIds_fetchesAllRecursiveReplies` runs against in-memory Room with three levels of nesting plus an unrelated leaf to prove the join doesn't over-collect, and the repository test is simplified rather than deleted. CI green (build default + lite, test default).

**Nit:** `NewsDao.getDirectReplies` is now dead — `VoicesRepositoryImpl.kt:279`, the line this PR removes, was its only production caller. Worth deleting it and its remaining test mocks in the same change. Also, `collectNewsAndReplies` is now a one-line passthrough to the DAO; keeping the named wrapper is defensible for the comment it carries, but it could equally be inlined. And since cycle-safety is now a property of the implementation, a test with `a.replyTo = b` and `b.replyTo = a` would pin it — that behaviour is currently accidental-looking to a reader who doesn't know `UNION` from `UNION ALL`.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #15607, opened 2026-08-13
all: smoother notification repository destination view modelling (fixes #15606)

### LINKED ISSUE #15606: Make tapping notifications open the correct destination screen
## Summary
Tapping an item in the in-app Notifications list should open the screen that matches the notification type instead of leaving the user in the list or navigating inconsistently.

## Expected behavior
Implement notification click handling so these notification types navigate to the correct destination:
- join request → the team's **Join Requests** page
- team update → the relevant team screen
- voice reply → `ReplyActivity`

Okuro3499:
fixes #15606
[Screen_recording_20260813_183924.webm](https://github.com/user-attachments/assets/7684bec6-830a-4d9c-8ad9-f65f8b945fee)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved notification classification for team updates, join requests, chats, tasks, and resources.
  * Fixed notification links and related content opening the correct destination.
  * Added support for English and Spanish team join notifications.
  * Improved fallback handling for blank team names and types.
  * Preserved notification read status updates after navigation.

* **Tests**
  * Added coverage for notification parsing, routing, classification, and team information fallbacks.


dogi:
**Review — rating: 62/100**

CI green on all three checks. The core insight is right: `parseNotification` never populated `relatedId`, so server-sourced notifications had nothing to navigate with, and adding `chat` / `team_join` / `voice_reply` routing closes real gaps. But collapsing the click handler onto `notification.relatedId` treats three different kinds of id as interchangeable, and two of them aren't team ids.

**1. `task` navigation regresses for locally-generated notifications**

`relatedId` on a local `task` notification is a **task id** — `NotificationsViewModel.kt:62-66` collects exactly those relatedIds and feeds them to `getTaskTeamNamesByTaskIds(taskIds)`. The old handler resolved it properly:

```kotlin
getTaskDetails(relatedId) -> teamTaskDao.getById(it) -> JSONObject(task.link).optString("teams")
```

The new path does `openTeam(notification.relatedId, TasksPage)`, handing a task id to `TeamDetailFragment` as `teamId`. It only works for server `newTask` docs, where `extractIdFromLink` pulls a team id out of the link.

**2. `join_request` navigation regresses the same way**

`relatedId` here is a **join-request doc id**, not a team id — `TeamsRepositoryImpl.getJoinRequestInfo` does `teamDao.getById(requestId)` and then reads `req.teamId` off it, so the two are distinct fields on distinct docs. It can also arrive `join_request_`-prefixed, which both `NotificationsRepositoryImpl.kt:169` and `DashboardActivity.kt:563` still strip. `openTeam(notification.relatedId, JoinRequestsPage)` passes it through unresolved and unstripped.

Note the display path still reads `relatedId` with the old meaning (`getJoinRequestDetailsBatch(joinRequestIds)` at `NotificationsViewModel.kt:77-81`), so after this change the same field means "request id" to the formatter and "team id" to the click handler. Either keep the resolution in the click path, or normalize `relatedId` to a team id at parse time and migrate the formatter and batch lookups with it — but the two readers have to agree.

**3. Dead code left behind**

Once the fragment stops calling them, `NotificationsViewModel.getTaskDetails` / `getJoinRequestTeamId` have no callers, and that orphans `NotificationsRepository(.Impl).getTaskDetails` / `getJoinRequestTeamId` and the `TaskNotificationResult` model with them. Worth removing in the same PR if the navigation rework stands. (`TeamsRepository.getJoinRequestTeamId` is a different method and is still used by `DashboardViewModel.kt:216` — that one stays.)

**4. `resolveType` leans harder on localized message sniffing**

The new `type == "team"` branch matches English phrases plus one Spanish one. The app ships ar/es/fr/ne/so, so a team chat notification rendered in Arabic or Nepali falls through to `team_join` and opens team home instead of the chat page. Message sniffing predates this PR, but the new branch short-circuits ahead of everything else, so it becomes the primary classifier for the most common server type. Keying off structural fields in the doc would be sturdier than phrase lists.

**Reads correctly:**
- `extractIdFromLink` bounds are right — `viewIndex in 0 until segments.lastIndex` rejects both "no `view` segment" (-1) and a trailing `view` with nothing after it.
- Passing `teamName = ""` / `teamType = ""` from `openTeam` is coherent precisely because of the `takeIf { it.isNotBlank() }` guards added to `getEffectiveTeamName`/`getEffectiveTeamType` and `renderPlaceholder` — those three changes are a set, and the new tests cover the fallback both ways.
- Widening `getEffectiveTeamName`/`Type` to `open` is test-only, matching the existing `shouldQueryTeamLocally` pattern in the same test file.

**Verdict:** the direction is good and the server-notification fix is worth having, but items 1 and 2 break click-through for notifications the app generates itself — the same behaviour the PR set out to improve. Resolve the id-kind mismatch (and pick one meaning for `relatedId` across both readers), then this is close to ready.

---

J-S-webskas (commented):
Currently I can only trigger the resources and storage notification, which isn't the ones tested. But the okuro account's notifications don't seem to be showing up for my phone, even when I log in, so I don't think I can test this one.

Okuro3499:
> Currently I can only trigger the resources and storage notification, which isn't the ones tested. But the okuro account's notifications don't seem to be showing up for my phone, even when I log in, so I don't think I can test this one.

@J-S-webskas 
You need to do a fresh install and sync for this to work as it starts from how it is saved during sync

dogi:
**Review — rating: 82/100**

Good work, well reasoned, and the comments explain the non-obvious parts rather than restating the code. One consequence needs surfacing before this ships.

**The classification design is right.** The key insight is documented where it belongs:

> Raw type "team" covers join requests, team-membership changes, and chat posts alike, and the server renders `message` in the recipient's locale, so it can't be classified reliably by sniffing English/Spanish phrases. `linkParams.activeTab == "applicantTab"` is a locale-independent signal…

Preferring a structural signal over message text, and having `resolveType` consult `subType` first with the heuristic only as fallback, is the correct ordering. `extractIdFromLink`'s bounds check (`viewIndex in 0 until segments.lastIndex`) is correct — it guarantees `viewIndex + 1` is in range rather than assuming it.

Making `getEffectiveTeamName`/`getEffectiveTeamType` use `takeIf { it.isNotBlank() }` is a real fix, and it's load-bearing for the rest of the PR: `openTeam` now passes `teamName = ""` and `teamType = ""`, so without the blank guard those empty arguments would win over the team's actual values. Good that both halves are in the same PR.

272 lines of tests across the repository, ViewModel, and `BaseTeamFragment`.

**⚠️ This bumps the Room schema version, which wipes every user's local database.**

```kotlin
-    version = 7,
+    version = 8,
```

`RoomModule` builds with `fallbackToDestructiveMigration(true)`, so on upgrade the local DB is dropped and re-pulled from CouchDB. Per the project's own guidance, that rebuild **also discards unsynced local writes** — pending submissions, drafts, anything not yet uploaded — for any device that hasn't synced before installing the update.

That's a heavy price for one nullable column, and it isn't mentioned in the PR description. Two things worth doing:

1. **Say it in the description**, so whoever cuts the release knows this one needs a forced-upload window rather than being folded into a routine build.
2. **Consider whether `subType` needs to be persisted at all.** It's derived from `linkParams.activeTab` at parse time, and `resolveType` already falls back to message-sniffing when it's absent. If the fallback is acceptable for the window between upgrade and first sync, the column buys accuracy at the cost of a wipe; if the column stays, that trade should be a conscious one.

**Known gap worth stating in the code.** The comment correctly argues that message text can't be classified reliably because it's server-localized — and then the fallback does exactly that, in English and Spanish only:

```kotlin
lower.contains("requested to join") || lower.contains("wants to join") ||
    lower.contains("solicitado unirse") -> "join_request"
```

The app ships six locales. For `ar`, `fr`, `ne`, and `so`, any `"team"` notification without `activeTab` falls through to `"team_join"` — so a chat notification in French routes to the team overview instead of the chat page. The `subType` path fixes this for join requests specifically; chat has no structural signal yet. Worth a `TODO` naming what server-side field would close it, so the next person doesn't add four more languages' worth of substrings.

**Nit:** `resolveAndOpenTeam`'s fallback — `resolve(relatedId) ?: relatedId` — is documented, which is good, but it means an unresolvable task id gets passed to `TeamDetailFragment` as a team id. That'll open an empty team rather than failing visibly. A log line on the fallback path would make the diagnosis obvious if it ever happens.

**Verdict:** `ready` on the code. Please add the schema-bump consequence to the PR description — that's the part most likely to cause a support issue, and it's invisible in the diff.

---

dogi (approved):

PR #15608, opened 2026-08-13
teams: smoother voices payload diffing (fixes #15733)

### LINKED ISSUE #15733: Optimize VoicesAdapter bind path
Optimized `VoicesAdapter` bind performance.

1.  **Hoisted JSON Parsing:** Extracted heavy Gson parsing overhead out of `onBindViewHolder`. Cached `parsedSharedTeamName` and `parsedImagesArray` inside the `News` model and populated them at submit time (`preParseNews`), preventing rapid scrolling allocations.
2.  **Fine-Grained Payloads:** Extended `DiffUtil.ItemCallback` and `getChangePayload`. Separated payloads for labels (`PAYLOAD_LABELS_CHANGED`) and images (`PAYLOAD_IMAGES_CHANGED`), bypassing redundant `resetViews` and `loadImage` calls when small updates (like labels or edit replies) ar

dogi:
Optimized `VoicesAdapter` bind performance.

1.  **Hoisted JSON Parsing:** Extracted heavy Gson parsing overhead out of `onBindViewHolder`. Cached `parsedSharedTeamName` and `parsedImagesArray` inside the `News` model and populated them at submit time (`preParseNews`), preventing rapid scrolling allocations.
2.  **Fine-Grained Payloads:** Extended `DiffUtil.ItemCallback` and `getChangePayload`. Separated payloads for labels (`PAYLOAD_LABELS_CHANGED`) and images (`PAYLOAD_IMAGES_CHANGED`), bypassing redundant `resetViews` and `loadImage` calls when small updates (like labels or edit replies) are made.
3.  **O(1) User Lookups:** Replaced `currentList.forEachIndexed` loops inside `getUserFn` callbacks with an `O(1)` map structure `userIdPositions` that rebuilds automatically on `onCurrentListChanged`.

---
*PR created automatically by Jules for task [4608294796161424433](https://jules.google.com/task/4608294796161424433) started by @dogi*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Voices list updates so labels, images, shared team names, and editing controls refresh correctly.
  * Fixed Voice item position tracking when entries are added, removed, or reordered.
  * Improved loading and display of library images.
* **Performance**
  * Faster, more efficient updates when Voice content changes.
  * Reduced unnecessary image reloads while editing Voice entries.


dogi:
**Review — rating: 72/100**

The performance work is genuinely good. `userIdPositions` replaces a full `currentList.forEachIndexed` scan on every user-fetch callback — with many rows and many distinct authors that was O(rows × users) of pure waste, and rebuilding the index in `onCurrentListChanged` is the right hook since `ListAdapter` invokes it synchronously after the diff lands, so the cached indices always match the committed list. Caching `parsedSharedTeamName` in `prepareSubmitList` removes a `JsonUtils.extractSharedTeamName` parse from every bind, and `parsedImagesArray` follows the existing `rawImageUrls`/`imageUrls` invalidation pattern already in that method, so it's consistent with how the class already works. Splitting labels and images out of `PAYLOAD_TEAM_LEADER_CHANGED`/`PAYLOAD_USER_FETCHED` into their own payloads is a real precision gain — lumping label changes under a "team leader changed" payload was actively misleading. CI green (build default + lite, test default).

**The `if (news.chat)` gate is a behaviour change, not an optimisation, and it's inconsistent.** In `PAYLOAD_EDIT_ACTION` the call becomes conditional, but the full-bind path at line 325 still calls `handleChat(holder, news)` unconditionally. So the two paths now disagree about when chat is refreshed. That matters because `handleChat` owns visibility in both directions:

```kotlin
if (news.newsId?.isNotEmpty() == true) {
    …
    holder.binding.recyclerGchat.visibility = View.VISIBLE
    holder.binding.sharedChat.visibility = View.VISIBLE
} else {
    holder.binding.recyclerGchat.visibility = View.GONE
    holder.binding.sharedChat.visibility = View.GONE
}
```

It gates on `newsId`, not on `chat` — and `news.chat` appears nowhere else in `VoicesAdapter`, so this PR introduces a dependency on a field the adapter has never consulted, whose relationship to `newsId` is unstated. Partial rebinds don't call `resetViews`, so the hide branch is the only thing that can clear chat views on that path, and it's now skipped for anything where `chat` is false. I don't think it corrupts state today, because `PAYLOAD_EDIT_ACTION` fires on `message`/`isEdited`/`time`/`sharedBy`/`replyTo` and none of those change conversations — but `handleChat` is cheap for a non-chat row, so the gate buys nothing measurable while making the two call sites disagree. Either drop it, or gate line 325 the same way and say why `chat` is the right predicate rather than `newsId`.

Two smaller things:

- **`loadImage` now depends on a precompute having run.** It reads `news?.parsedImagesArray` where it previously read the computed `imagesArray` property. `prepareSubmitList` wraps that assignment in a `try`/`catch (IllegalStateException)`, so if that catch ever fires the field stays null and images silently stop rendering, where the computed property would still have worked. Falling back — `news?.parsedImagesArray ?: news?.imagesArray` — keeps the fast path and removes the failure mode. (That catch is itself vestigial, incidentally: its comment refers to Realm managing the object, and Realm is gone.)
- **`PAYLOAD_LABELS_CHANGED`'s handler is byte-identical to `PAYLOAD_TEAM_LEADER_CHANGED`'s** — three duplicated lines that should share a branch or a helper.

⚠️ **Conflicts with #15596** (`ready`), which rewrites this adapter's constructor and the `leadersList` field. Both touch `VoicesAdapter` substantially, so whichever lands second needs a real rebase. Worth sequencing.

**Nit:** no tests. `getChangePayload` is pure logic over two `News` instances — the new label/image payload split is exactly the kind of branching that a few direct assertions would pin cheaply.

**Verdict:** needs another round — justify or drop the `news.chat` gate, and the rest is ready.

---

dogi:
**Re-review — Rating: 78/100**

CI green on all three checks. Three solid improvements here, held by one missing fallback.

**What's good:**

Splitting `PAYLOAD_TEAM_LEADER_CHANGED` into `PAYLOAD_LABELS_CHANGED` and `PAYLOAD_IMAGES_CHANGED` means an image change no longer forces a user-info rebind and vice versa — the payload now describes what actually changed. Extracting `updateLabels(holder, news)` removes four copies of the same three-line block. And maintaining `userIdPositions` in `onCurrentListChanged` replaces a `currentList.forEachIndexed` scan per lookup with an O(1) map read.

**`parsedSharedTeamName` has no recompute fallback**

```kotlin
val sharedTeamName = news.parsedSharedTeamName ?: ""     // :309 and :328
```

Compare the images path two hundred lines down, which does the defensive thing:

```kotlin
val imagesToLoad = news?.parsedImagesArray ?: news?.imagesArray    // :872
```

The population site makes this matter rather than being theoretical — `it.parsedSharedTeamName = JsonUtils.extractSharedTeamName(it)` at `:642` sits inside a `try { … } catch (e: IllegalStateException) { e.printStackTrace() }`. If that catch ever fires, the field stays null and the row renders an **empty team name** instead of falling back to computing it. A silently blank label rather than a visible failure.

`?: JsonUtils.extractSharedTeamName(news)` at both sites matches the images pattern and costs nothing on the hot path, since the cache hit still short-circuits.

**The comment on that catch is stale:**

```kotlin
// If Realm manages the object, and we are on a different thread, mutating @Ignore fields might throw.
```

This codebase has no Realm — Room is the only store and the `Realm*` prefixes are gone. So either the catch is guarding against something that can no longer happen and should go, or it's guarding against something real and the comment should say what. Right now it justifies swallowing an exception with a reason that doesn't apply.

**Consistent with #15643**, which uses the same cache-on-`@Ignore`-field approach for `parsedViewIn` — and notably *does* keep `?: gson.fromJson(...)` fallbacks at every read site. Worth matching that.

**Verdict:** the payload split and the position map are real wins. Add the recompute fallback at the two `parsedSharedTeamName` reads and sort out the Realm comment, then this is ready.

---

dogi:
**Review — rating: 74/100**

The core of this is right, and two of the three changes are clear wins.

**The payload split is a real bug fix, not just an optimization.** A labels change was emitting `PAYLOAD_TEAM_LEADER_CHANGED`, which meant the leader-changed handler ran for an unrelated reason. Giving labels their own `PAYLOAD_LABELS_CHANGED` fixes that, and pulling the image fields out of `PAYLOAD_USER_FETCHED` into `PAYLOAD_IMAGES_CHANGED` is correct — I checked `loadImage`, and it reads nothing user-derived, so the two really are independent. Dropping `loadImage` from the `PAYLOAD_EDIT_ACTION` branch is also safe: an edit that swaps attachments changes `images`, which now emits `PAYLOAD_IMAGES_CHANGED` in the same payload list, and `onBindViewHolder` iterates all payloads.

**`userIdPositions` is a sound trade.** Old cost was O(users × n); new is one O(n) rebuild per list swap plus O(1) per user callback. `onCurrentListChanged` fires after the diff is dispatched, and the direct `super.submitList(...)` calls at lines 354 and 556 go through it too, so the map can't drift.

**The problem is `preParseNews`.** Removing the `try/catch` is defensible on its face — Realm is long gone, so the stated reason for the catch no longer holds. But the catch was `catch (e: IllegalStateException)`, and that is precisely what this new line can throw:

```kotlin
it.parsedSharedTeamName = JsonUtils.extractSharedTeamName(it)
```

`extractSharedTeamName` wraps only the `gson.fromJson` call in its own `try`. The line after it is unguarded:

```kotlin
if (ar != null && ar.size() > 1) {
    val ob = ar[0].asJsonObject   // IllegalStateException if element 0 isn't an object
```

So the PR adds a call that throws `IllegalStateException` and deletes the handler for `IllegalStateException` in the same commit.

`it.parsedImagesArray = it.imagesArray` has the same shape — the getter is `gson.fromJson(images, JsonArray::class.java)`, which throws `JsonSyntaxException` on malformed input.

What makes this worse than the status quo is the change in *when* it runs. Both parses used to happen lazily in `onBindViewHolder`, so a malformed record only ever hurt the row being scrolled into view. `preParseNews` runs over every item at submit time, so one bad `viewIn` or `images` value anywhere in the list now takes down the whole list submission. Eager parsing is the right call for scroll performance — it just needs to fail per-item rather than per-list. Wrapping the two new assignments in a `try/catch` that leaves the cached field null (the read sites already fall back) restores that.

**Caching gap — `parsedImagesArray` misses the cases it most wants to hit:**

```kotlin
if (it.rawImages != it.images) {
    it.parsedImagesArray = it.imagesArray
    it.rawImages = it.images
}
```

When `images` is null, `rawImages` is also null on a fresh object, so the guard is false, the branch never runs, and `loadImage`'s `?: news?.imagesArray` fallback re-enters the getter on every bind. When `images` is `""`, `fromJson` returns null, so `parsedImagesArray` is null and the fallback fires every bind as well. `null` is doing double duty as both "not cached yet" and "parsed to nothing". A separate `imagesParsed: Boolean` flag, or keying the guard on `parsedImagesArray == null || rawImages != images`, closes it. Neither case is expensive on its own, but they're the majority of rows, which makes them the ones worth caching.

`parsedSharedTeamName` doesn't have this issue — `extractSharedTeamName` returns non-null `""`, so the elvis is a genuine cold-path guard.

**Observation, not a blocker:** three more mutable `@Ignore` fields on the `News` entity as adapter-side scratch space. It follows the existing `parsedViewIn`/`rawViewIn` precedent and the values are pure functions of the entity's own fields, so sharing an instance between the parent and children lists is harmless. Worth keeping an eye on as a pattern before it grows further.

**Verdict:** the payload work and the position map I'd take as-is. Guard the two new eager parses so a single malformed record can't fail an entire `submitList`, and tighten the `parsedImagesArray` guard, and this is `ready`.

---

dogi:

Both blocking findings are fixed, and the error handling is better than what was there before the PR started.

**Per-item isolation restored, and widened.** `preParseNews` now wraps its whole body in `catch (e: Exception)`, so a malformed record degrades to uncached fields instead of failing the entire `submitList`. Worth noting this is *stronger* than the `catch (e: IllegalStateException)` you originally removed: the eager `it.parsedImagesArray = it.imagesArray` can throw `JsonSyntaxException` on bad `images` JSON, which the old narrow catch would have missed. Catching `Exception` covers both that and the `asJsonObject` path inside `extractSharedTeamName`. The comment explains the *why*, which is the part that stops someone deleting it again.

**The `parsedImagesArray` guard is right now:**

```kotlin
if (it.parsedImagesArray == null || it.rawImages != it.images) {
```

This closes the case that mattered. With `images == null`, the old `rawImages != images` guard was false on a fresh object, so the branch never ran and every bind fell through to the `imagesArray` getter. Now the first pass caches the empty `JsonArray` and subsequent passes skip. Since most rows have no images, that was the majority of binds.

**Two small residuals, neither blocking:**

- `images == ""` still re-parses. The getter's `if (images == null) JsonArray() else gson.fromJson(images, ...)` returns *null* for an empty string, so `parsedImagesArray` stays null and the `== null` clause retriggers each pass, with the bind-path elvis falling through too. In practice this shouldn't arise — `VoicesRepositoryImpl` writes `images` via `toJson`, which produces `"[]"` rather than `""`, and `"[]"` caches correctly — so it's a legacy-record edge at most.

- The `parsedSharedTeamName` fallback re-enters the throwing call. If `extractSharedTeamName` is what raised inside `preParseNews`, the field stays null, and `onBindViewHolder`'s `news.parsedSharedTeamName ?: JsonUtils.extractSharedTeamName(news)` calls the same function again — this time with no catch around it. That's the pre-existing behavior on `master`, so it's not a regression and the common path is now protected; but assigning `""` on failure (the function's own non-null "not shared" value) would make the guard total rather than partial.

**Verified still correct from last round:** the payload split is a genuine fix, not just an optimization — labels were emitting `PAYLOAD_TEAM_LEADER_CHANGED`, and images are independent of user data, so pulling them out of `PAYLOAD_USER_FETCHED` is sound. `userIdPositions` stays in step because `onCurrentListChanged` fires after the diff dispatches, including for the direct `super.submitList(...)` calls at lines 354 and 556.

**Verdict:** `ready`. The eager parsing now fails per-item instead of per-list, the cache actually caches the common case, and CI is green on all three checks.

---

Okuro3499 (approved):

PR #15609, opened 2026-08-13
all: smoother data room converters type token caching (fixes #15705)

### LINKED ISSUE #15705: Cache TypeToken in Converters to improve database read performance
Addresses issue where `toStringList()`, `toConversationList()`, and `toAttachmentList()` allocated an anonymous `TypeToken` on every row conversion. Resolving generic `Type` values in a companion object and reusing them.

Tracked in PR #15609.

dogi:
Addresses issue where `toStringList()`, `toConversationList()`, and `toAttachmentList()` allocated an anonymous `TypeToken` on every row conversion. Resolving generic `Type` values in a companion object and reusing them. Included round trip tests for lists and null inputs.

---
*PR created automatically by Jules for task [8780174892482025904](https://jules.google.com/task/8780174892482025904) started by @dogi*


dogi:
**Review — rating: 90/100**

Well-chosen target. `object : TypeToken<List<String>>() {}.type` isn't just an allocation — instantiating an anonymous `TypeToken` subclass forces a `getGenericSuperclass()` reflection walk to recover the erased type argument, and `Converters` runs on *every row read* of a column carrying one of these types. Room may also hold more than one `Converters` instance, so putting the three `Type` values in a `companion object` makes them process-wide rather than per-instance. The hoisted values are identical `Type` objects, `Type` is immutable, `fromJson(String, Type)` is thread-safe, and companion initialisation is handled by class init — so there's no behavioural or concurrency change, only the reflection and allocation removed from the read path.

I checked that the sweep is complete where it matters. `Converters.kt` has exactly eight `@TypeConverter` functions; the three `to*List` methods are the only ones using `TypeToken` at all (the `Date` pair round-trips through `Long`), and all three are hoisted. There are five other `object : TypeToken` sites in the codebase — `ChatDetailFragment:627`, `CoursesProgressFragment:45`, `LifeRepositoryImpl:90`, `SharedPrefManager:74`, `ChatApiService:40` — but every one of those is per-user-action rather than per-row, so leaving them is the right scoping call, not an oversight. Worth a line in the description saying so, since at a glance it reads like a partial sweep.

`ConvertersTest` is a genuine addition — six cases covering round-trip and the null/empty/whitespace guard for all three converters, in a file that previously had no tests. Using field-level assertions for `Conversation`/`Attachment` rather than `assertEquals` on the lists is the right choice given those classes have no value equality; the `List<String>` case can and does use full equality. Plain JUnit with no Robolectric needed, since Gson is pure JVM. CI green (build default + lite, test default).

**Nit:** the tests cover the `to*` direction thoroughly but never call `fromStringList(null)`, and nothing covers malformed input. That second one is worth knowing rather than fixing here: none of the three converters catches `JsonSyntaxException`, so a corrupted column value propagates the exception out of a Room query. Pre-existing, and arguably the honest behaviour given the drop-and-resync migration strategy, but it's now the only untested branch of a file that otherwise has full coverage.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #15610, opened 2026-08-13
courses: smoother submissions repository dao bulk inserting (fixes #15707)

### LINKED ISSUE #15707: Optimize survey submission creation by using bulk chunked queries and inserts
Replaced the per-user `getOrCreateSubmission` loop in `createBulkSurveySubmissions` with a batched approach. It queries existing submissions in chunks of 500 (avoiding SQLite's 999 parameter limit), filters out users who already have a submission, and bulk inserts entirely new submissions for the rest using `upsertAll`.

Tracked in PR #15610.

dogi:
Optimize survey submission creation by using bulk chunked queries and inserts

Replaced the per-user `getOrCreateSubmission` loop in `createBulkSurveySubmissions`
with a highly efficient batched approach. It now queries existing submissions in
chunks of 500 (avoiding SQLite's 999 parameter limit) using a new DAO query
`getPendingByUsersAndParent`, filters out users who already have a submission,
and bulk inserts entirely new submissions for the rest using `upsertAll`. Tests
have been updated to cover all edge cases (empty list, all new, mixed, and duplicates).

---
*PR created automatically by Jules for task [1947861779757725074](https://jules.google.com/task/1947861779757725074) started by @dogi*


dogi:
**Review — rating: 89/100**

Correct on every point I checked, including the three that usually break this kind of inlining.

**The construction is faithful.** The new `Submission().apply { … }` block is byte-for-byte the same eight assignments in the same order as `getOrCreateSubmission`'s creation path (lines 559-568) — `id`, `userId`, `parentId`, `status = "pending"`, `type = "survey"`, `startTime`, `lastUpdateTime`, `answers`. Nothing is dropped.

**The write is equivalent.** `getOrCreateSubmission` finished with `saveSubmission`, which is `submissionDao.upsertAll(listOf(submission))` plus an `answerDao.upsertAll` guarded by `if (answerEntities.isNotEmpty())`. New submissions are created with `answers = mutableListOf()`, so that branch could never fire — meaning the bulk `submissionDao.upsertAll(newSubmissions)` is exactly what the per-user path did, not an approximation of it.

**The lookup matches.** `getPendingByUsersAndParent` is the set-valued generalisation of `getLatestPendingByUserAndParent`: same `status = 'pending'` predicate, same `parentId` equality, same `lastUpdateTime DESC` ordering. So a user with only a *completed* submission still gets a new pending one, exactly as before.

And the chunking is right, which is worth calling out because two other open PRs in this queue got it wrong. `:userIds` is bound once, so `chunked(500)` yields 501 host parameters — comfortably under SQLite's 999 limit. Dedup across chunks also works, since chunk N+1's query sees chunk N's committed inserts. The four replacement tests are better than what they replace: the old one stubbed `getOrCreateSubmission` on the repository under test, whereas these assert against the DAO and cover empty / all-new / mixed / all-existing. CI green (build default + lite, test default).

**Two nits:**

- **The chunking boundary is the one new branch with no test.** All four cases use two or three users, so the `chunked(500)` loop only ever runs once. A case with 501 ids asserting two `getPendingByUsersAndParent` calls would pin the behaviour that this PR actually introduces.
- **`userIds` isn't de-duplicated.** A repeated id *within* one chunk passes `chunk.filter { it !in existingUserIds }` twice and produces two submissions with different UUIDs. The old per-user loop couldn't do that, because the second `getOrCreateSubmission` would have found the first. Across chunks it's safe for the reason above, so this is intra-chunk only — and `userIds.distinct().chunked(500)` closes it for free.

The `ORDER BY lastUpdateTime DESC` is unnecessary given only the userId set is consumed, but it mirrors the singular query it generalises, so I'd leave it.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #15611, opened 2026-08-13
teams: smoother repository dao querying (fixes #15706)

### LINKED ISSUE #15706: Refactor TeamDao queries to prevent full-table scans
Replaced full-table `getAll()` scans in `getJoinRequestsInfo`, `getTeamNamesByIds`, and `getResourceIds` with targeted SQL queries using `_id IN (:ids)` and specific docType filters. Added `.distinctUntilChanged()` to `getMyTeamDetailsFlow` to suppress redundant recompositions. Chunked SQL `IN` queries by 500 to avoid exceeding SQLite max parameter limits.

Tracked in PR #15611.

dogi:
Replaced full-table `getAll()` scans in `getJoinRequestsInfo`, `getTeamNamesByIds`, and `getResourceIds` with targeted SQL queries using `_id IN (:ids)` and specific docType filters.
Added `.distinctUntilChanged()` to `getMyTeamDetailsFlow` to suppress redundant recompositions triggered by unrelated database writes.
Chunked SQL `IN` queries by 500 to avoid exceeding SQLite max parameter limits.

---
*PR created automatically by Jules for task [2525456066979484370](https://jules.google.com/task/2525456066979484370) started by @dogi*


dogi:
**Review — rating: 86/100**

Three separate `teamDao.getAll()` full-table materialisations replaced with targeted queries, and each translation holds up.

**`getByIds` is exactly equivalent, for a reason worth stating.** The old predicate was `(it._id ?: it.id) in requestIds`, which reads as "match `_id`, falling back to `id`" — and a `WHERE _id IN (:ids)` query would then miss rows with a null `_id`. But `MyTeam._id` is declared `@PrimaryKey @JvmField var _id: String = ""` — non-null — so that elvis could never take its right branch. The old code was effectively `it._id in requestIds` all along, and the new query matches it exactly. (Which means the `?: it.id` you kept in `.distinctBy { it._id ?: it.id }` is likewise dead; worth dropping so the next reader doesn't infer a nullable `_id` that doesn't exist.)

**The chunking is correct.** `:ids` is bound once, so `chunked(500)` yields 501 host parameters, well under SQLite's 999 limit — and `distinctBy` is genuinely needed rather than defensive padding, since duplicate entries in the input list can land in different chunks and match the same row twice.

**`getResourceIdsByTeamId` is faithful and better than a like-for-like port.** `resourceId?.isNotBlank() == true` → `IS NOT NULL AND TRIM(resourceId) != ''`, `docType.isNullOrBlank()` → `docType IS NULL OR TRIM(docType) = ''`, and the two literal comparisons stay case-sensitive as SQLite's `=` on TEXT already is. Selecting `resourceId` directly rather than `SELECT *` also means the query returns only the column the caller wanted, which the `mapNotNull` was previously doing after hydrating whole entities.

**And `distinctUntilChanged()` actually works here** — unlike the equivalent addition in #15598, which I sent back precisely because `MyCourse` is an `open class` with no `equals`. `TeamDetails` is a `data class` whose fields are all `String?`/`Long`/`TeamStatus?`, so the generated `equals` compares by value and the operator can genuinely suppress duplicate emissions.

CI green (build default + lite, test default).

**Nits:**

- `.flowOn(dispatcherProvider.default).distinctUntilChanged()` puts the comparison *downstream* of `flowOn`, so the list-equality walk runs on the collector's dispatcher rather than on Default. Swapping the two moves that work off the collector as well.
- `TRIM()` strips spaces only, while Kotlin's `isNotBlank()` covers all whitespace — a `resourceId` of `"\t"` now survives the filter. Contrived for synced data, and `TRIM` is the right call versus a bare `!= ''`, but it's a real edge difference.
- No tests. This repo already has `UserDaoTest` and `NewsDaoTest` doing exactly this against in-memory Room, and `getResourceIdsByTeamId` has four interacting predicates — the highest-value case being a row with a `docType` that is none of null/blank/`resourceLink`/`link`, to prove it's excluded.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #15612, opened 2026-08-13
courses: smoother survey submission syncing (fixes #15593)

### LINKED ISSUE #15593: First survey submission is not counted or submitted
## Description

When a user attempts to submit a survey for the first time, the submission is not counted or recorded. The survey appears to complete, but the response is neither saved nor reflected in the system.

## Steps to Reproduce
1. Open the app and navigate to a survey.
2. Fill out all required fields in the survey.
3. Submit the survey (first-time submission).
4. Check whether the submission has been recorded/counted.
https://github.com/user-attachments/assets/7ad51729-65f8-4304-a0a3-f4c149236bb0

## Expected Behavior
The first survey submission should be successfully recorded and counted in the system, just like any subsequent submission.

Okuro3499:
fixes #15593
[Screen_recording_20260813_194656.webm](https://github.com/user-attachments/assets/fcb04b8b-6ebe-461a-81da-674a52e3ab1a)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved offline submission tracking, including submissions without a remote identifier.
  * Ensured updated submission statuses are saved correctly.
  * Refreshed survey displays when survey information changes.

* **Chores**
  * Updated the Android app version to 0.64.52.


dogi:
**Review — rating: 72/100**

CI green on all three checks. Two unrelated halves in one PR, and they land very differently.

**The DAO half is the good part — one of these is a silent data-loss fix**

`_id = ''` never matches a `NULL` in SQL, so `getPendingSubmissions` was skipping every locally-created submission whose `_id` was null rather than empty-string. Those rows are exactly the ones that have never been to the server, so they'd sit unuploaded indefinitely with no user-visible signal. Same reasoning for `countPendingOfflineSubmissions`. The fix also matches the shape already used two lines down in `getPendingExamResults` (`_id IS NULL OR _id = ''`), so the file is now self-consistent instead of half-right.

Adding `isUpdated = 1` to `updateStatus` / `updateStatusAndLastUpdate` is right too, and I checked the direction of the risk: both callers (`SubmissionsRepositoryImpl.kt:511` and `:527`) are local user-driven writes — answering a question, explicitly setting status — not sync-apply paths. So there's no echo loop where a downloaded row gets re-marked dirty and pushed straight back, and `markUploaded` clears the flag on the way out.

**The `SurveyFragment` half is the problem**

```kotlin
viewModel.surveyInfos.collect { infos -> …; getAdapter().notifyDataSetChanged() }
viewModel.bindingData.collect { data -> …; getAdapter().notifyDataSetChanged() }
```

The underlying complaint is real: `surveyInfoMap` and `bindingDataMap` are side-channel maps the adapter reads during bind, so mutating them notifies nothing and the rows keep showing stale data. Something has to trigger a rebind. But `notifyDataSetChanged()` on a `ListAdapter` is the bluntest available option — it discards item animations, forces a full rebind of every visible row, and interleaves with the `AsyncListDiffer`'s own notifications that the adapter is simultaneously issuing from `submitList`. Two collectors each firing it on every emission, including the initial one, means several full rebinds during ordinary screen setup.

`notifyItemRangeChanged(0, itemCount, SOME_PAYLOAD)` would be a drop-in improvement. The structurally right fix is to fold `surveyInfo` and `bindingData` into the list item type so they flow through `submitList` and the existing `DiffUtil` callback, which removes the side-channel maps entirely.

Worth flagging that this cuts directly against work already in flight: #15650 is currently held at `change` for adding `notifyDataSetChanged()` to `BaseRecyclerFragment`, and #15602 / #15608 / #15618 / #15621 / #15652 / #15653 / #15654 are all moving adapters toward payload-based diffing. Landing this adds churn those PRs would have to come back and remove.

**Minor:** `getAdapter()` is memoized behind `adapterMutex` in this fragment, so calling it per emission doesn't rebuild the adapter — no issue there.

**Verdict:** the `SubmissionDao` changes are worth landing on their own merit and I'd take them as-is. Swap the two `notifyDataSetChanged()` calls for a payload rebind — or split the DAO fixes into their own PR so they aren't held up by the adapter question.

---

J-S-webskas (commented):
So far, the survey correctly updates the latest time that it is taken:

https://github.com/user-attachments/assets/8657f872-dd79-4f4d-ae9d-bc7458de8a24

Okuro3499:
> So far, the survey correctly updates the latest time that it is taken:
> 
>  Recording_20260813_142522.mp4
@J-S-webskas 
approve

Okuro3499:
> **Review — rating: 72/100**
> 
> CI green on all three checks. Two unrelated halves in one PR, and they land very differently.
> 
> **The DAO half is the good part — one of these is a silent data-loss fix**
> 
> `_id = ''` never matches a `NULL` in SQL, so `getPendingSubmissions` was skipping every locally-created submission whose `_id` was null rather than empty-string. Those rows are exactly the ones that have never been to the server, so they'd sit unuploaded indefinitely with no user-visible signal. Same reasoning for `countPendingOfflineSubmissions`. The fix also matches the shape already used two lines down in `getPendingExamResults` (`_id IS NULL OR _id = ''`), so the file is now self-consistent instead of half-right.
> 
> Adding `isUpdated = 1` to `updateStatus` / `updateStatusAndLastUpdate` is right too, and I checked the direction of the risk: both callers (`SubmissionsRepositoryImpl.kt:511` and `:527`) are local user-driven writes — answering a question, explicitly setting status — not sync-apply paths. So there's no echo loop where a downloaded row gets re-marked dirty and pushed straight back, and `markUploaded` clears the flag on the way out.
> 
> **The `SurveyFragment` half is the problem**
> 
> ```kotlin
> viewModel.surveyInfos.collect { infos -> …; getAdapter().notifyDataSetChanged() }
> viewModel.bindingData.collect { data -> …; getAdapter().notifyDataSetChanged() }
> ```
> 
> The underlying complaint is real: `surveyInfoMap` and `bindingDataMap` are side-channel maps the adapter reads during bind, so mutating them notifies nothing and the rows keep showing stale data. Something has to trigger a rebind. But `notifyDataSetChanged()` on a `ListAdapter` is the bluntest available option — it discards item animations, forces a full rebind of every visible row, and interleaves with the `AsyncListDiffer`'s own notifications that the adapter is simultaneously issuing from `submitList`. Two collectors each firing it on every emission, including the initial one, means several full rebinds during ordinary screen setup.
> 
> `notifyItemRangeChanged(0, itemCount, SOME_PAYLOAD)` would be a drop-in improvement. The structurally right fix is to fold `surveyInfo` and `bindingData` into the list item type so they flow through `submitList` and the existing `DiffUtil` callback, which removes the side-channel maps entirely.
> 
> Worth flagging that this cuts directly against work already in flight: #15650 is currently held at `change` for adding `notifyDataSetChanged()` to `BaseRecyclerFragment`, and #15602 / #15608 / #15618 / #15621 / #15652 / #15653 / #15654 are all moving adapters toward payload-based diffing. Landing this adds churn those PRs would have to come back and remove.
> 
> **Minor:** `getAdapter()` is memoized behind `adapterMutex` in this fragment, so calling it per emission doesn't rebuild the adapter — no issue there.
> 
> **Verdict:** the `SubmissionDao` changes are worth landing on their own merit and I'd take them as-is. Swap the two `notifyDataSetChanged()` calls for a payload rebind — or split the DAO fixes into their own PR so they aren't held up by the adapter question.
> 
> _Generated by [Claude Code](https://claude.ai/code)_

claude suggests this to be done in a new pr so I won't be making changes on this branch
Screenshot 2026-08-14 at 17 00 03

J-S-webskas (approved):

dogi:
**Review — rating: 78/100**

Two real bug fixes in the DAO, and one blunt instrument in the fragment.

**The NULL handling fix is correct and overdue.**

```sql
-- before
WHERE (isUpdated = 1 OR _id = '')
-- after
WHERE (isUpdated = 1 OR _id IS NULL OR _id = '')
```

`_id = ''` never matches `NULL` in SQL, so submissions that had never been assigned a remote id were invisible to both `countPendingOfflineSubmissions` and `getPendingSubmissions` — i.e. silently excluded from the pending count *and* from the upload queue. That's data that never syncs, with no symptom other than a wrong badge number.

It also brings these two in line with `getPendingExamResults`, three lines below, which already reads `(_id IS NULL OR _id = '')`. The inconsistency between them was the tell.

**`isUpdated = 1` on the status updates is safe.** I chased the echo-loop risk — if `updateStatus` were called anywhere in the upload path, marking the row dirty would re-queue it forever against `markUploaded`'s `isUpdated = 0`. Both callers are local user-initiated writes, not upload-path writes, so there's no loop. Worth stating explicitly since it's the non-obvious risk in this change.

**The `notifyDataSetChanged()` calls are the weak part:**

```kotlin
viewModel.surveyInfos.collect { infos ->
    surveyInfoMap.clear()
    surveyInfoMap.putAll(infos)
    getAdapter().notifyDataSetChanged()
}
```

A notification is genuinely *needed* here — `surveyInfoMap` and `bindingDataMap` are side maps the adapter reads during bind, so mutating them without notifying leaves stale rows. No argument with the intent.

But `notifyDataSetChanged()` discards every view holder, kills item animations, and drops scroll position, and this fires on every emission of two separate flows. It's also directly against where the codebase is heading: **#15654 and #15599 are both open specifically to replace `notifyDataSetChanged()` with `notifyItemRangeChanged(0, itemCount)`** in the courses and resources adapters. Adding two new ones while those are in flight will just generate a third cleanup PR.

`notifyItemRangeChanged(0, itemCount)` is a drop-in here and preserves holders and scroll state. Payload-based would be better still, since each map feeds a specific part of the row, but the range call is the cheap correct move.

**Nit:** both collectors do `clear()` then `putAll()` then notify unconditionally, so a flow emission with identical content still triggers a full rebind. If these flows aren't already `distinctUntilChanged`, that's avoidable churn.

**Verdict:** `change` — and only for the notify. The DAO half is `ready`-grade and fixes submissions that were quietly never uploading; if you'd rather land that immediately, splitting the `SubmissionDao` changes into their own PR would get the fix out without waiting on the adapter question.

---

dogi:
Promoted `change` → `merge`.

The `SubmissionDao` half stands on its own: adding `_id IS NULL` to `countPendingOfflineSubmissions` and `getPendingSubmissions` fixes submissions that had never been assigned a remote id being invisible to both the pending count and the upload queue, and `isUpdated = 1` on the two status updates is safe — I traced both callers and neither is on the upload path, so there's no echo loop against `markUploaded`'s `isUpdated = 0`.

The `notifyDataSetChanged()` point is tracked separately in #15700 rather than holding this PR — the notification is needed for correctness, it's just blunter than it should be.

---

dogi (approved):

PR #15613, opened 2026-08-13
sync: smoother time logger date formatting (fixes #15709)

### LINKED ISSUE #15709: Refactor: Optimize date format allocations to thread-safe formatters
Resolves the issue of per-row and inline allocations of `SimpleDateFormat` across `NotificationsAdapter`, `TeamCalendarFragment`, `EnterprisesFinancesFragment`, `SyncManager`, and `SyncTimeLogger`. These ad-hoc instances were expensive to construct and non-thread-safe. By utilizing fixed properties populated with `java.time.format.DateTimeFormatter` (and `ThreadLocal` wrapper for legacy endpoints), this improves memory churn during heavy recycler view scroll updates and sync cycles.

Tracked in PR #15613.

dogi:
Resolves the issue of per-row and inline allocations of `SimpleDateFormat` across `NotificationsAdapter`, `TeamCalendarFragment`, `EnterprisesFinancesFragment`, `SyncManager`, and `SyncTimeLogger`. These ad-hoc instances were expensive to construct and non-thread-safe. By utilizing fixed properties populated with `java.time.format.DateTimeFormatter` (and `ThreadLocal` wrapper for legacy endpoints), this patch improves memory churn during heavy recycler view scroll updates and sync cycles.

---
*PR created automatically by Jules for task [16586369772961802935](https://jules.google.com/task/16586369772961802935) started by @dogi*


dogi:
**Review — rating: 70/100**

Three of the five sites are clean wins. `SyncManager` was allocating a `SimpleDateFormat` on every log line — twice per sync — and a hoisted `DateTimeFormatter` field fixes that with no semantic change, since both `ofPattern(p)` and `SimpleDateFormat(p)` resolve to `Locale.getDefault(Locale.Category.FORMAT)`. `SyncTimeLogger` is the best of them: `DateTimeFormatter` is genuinely immutable and thread-safe, so it removes the `ThreadLocal` machinery outright rather than working around `SimpleDateFormat`'s mutability, and `Locale.US` is preserved. `Calendar.toInstant()` in `TeamCalendarFragment` is fine at minSdk 26.

**But two sites delete deliberate locale-change handling.** `NotificationsAdapter.getDateFormat()` and `TeamCalendarFragment.getDateFormat()` both did this:

```kotlin
if (cached != null && lastLocale == currentLocale) return cached
```

That isn't incidental caching — it's explicit invalidation for when `Locale.getDefault()` changes, and someone wrote it because this app ships five languages with in-app language switching. Both are replaced by a `private val` that snapshots the locale once at construction. If the adapter and fragment are always recreated on a locale change then the new code is fine — but in that case the old invalidation was already dead and the PR should say so, because right now it reads as an optimisation that quietly removed a guard. Either justify it in the description or capture the locale at format time.

**And one site moves in the opposite direction from the other four.** `EnterprisesFinancesFragment` goes *from* a plain `SimpleDateFormat` field *to* `ThreadLocal.withInitial { SimpleDateFormat(...) }` — while `SyncTimeLogger`, in the same PR, deletes exactly that pattern. The three call sites (`formatToString`, `parseDate`, `filterDataByDateRange`) are all on the UI path, so there's no concurrency to defend against; what the `ThreadLocal` adds is a per-access map lookup and four `.get()!!` non-null assertions. I take it the reason is that this site also *parses*, which `DateTimeFormatter` makes more awkward when the result must be a `Date` — but then the honest fix is to leave it alone, or convert it to `LocalDate` parsing, not to adopt the idiom the rest of the PR is removing.

**Nits:** `val formatter = dateFormatter` in `showMeetupDialog` is now pointless indirection. The new `java.time.*` imports are inserted above `java.text.*`, which is out of alphabetical order in every file touched. And `SimpleDateFormat`/`Date` look dead in `NotificationsAdapter` and `SyncTimeLogger` once their last usages go — worth checking each file's imports.

**Verdict:** needs another round — keep the three clean conversions, justify or restore the locale invalidation, and make `EnterprisesFinancesFragment` consistent with the rest.

---

dogi:

CI green on all three checks. The thread-safety angle is the real win here and it's worth being explicit about, because the title only mentions allocations.

`SimpleDateFormat` is **not thread-safe** — a shared instance used from two threads produces silently wrong output, not an exception. `EnterprisesFinancesFragment`, `NotificationsAdapter`, `TeamCalendarFragment` and `SyncTimeLogger` all held one as a field. `DateTimeFormatter` is immutable and safe to share, so those become correct rather than merely cheaper.

`SyncManager` was constructing a fresh `SimpleDateFormat` per log line; hoisting it to a field removes that allocation from the sync path.

**The conversions are done correctly.** `dateFormatter.format(this.toInstant())` works because the formatter carries `.withZone(ZoneId.systemDefault())` — formatting an `Instant` with a date-only pattern needs a zone, and it has one. `parseDate` goes through `LocalDate.parse(dateString, dateFormatter)` then `atStartOfDay(zone).toInstant().toEpochMilli()`, which lands on the same instant `SimpleDateFormat.parse` produced for a midnight-anchored date.

**One behavioural change worth stating in the description:** `SimpleDateFormat` parses leniently by default. `"2024-13-45"` used to roll over into a valid date, and trailing garbage like `"2024-01-01xyz"` used to parse fine. `LocalDate.parse` rejects both with `DateTimeParseException`, which the catch converts to `null` / a logged stack trace. That's stricter and almost certainly what you want, but any stored value that only parsed by leniency will now be treated as invalid.

**Minor:** in `filterDataByDateRange` the `catch (e: IllegalArgumentException)` is now unreachable — it existed to catch the explicit `throw IllegalArgumentException(...)` on a null parse result, and `LocalDate.parse` throws `DateTimeParseException` (a `RuntimeException`, not an `IllegalArgumentException`) instead.

**Verdict:** correct conversions, and it retires a genuine thread-safety hazard rather than just an allocation. Marking `ready`.

---

Okuro3499 (approved):

PR #15614, opened 2026-08-13
Memoize `UrlUtils.header`

dogi:
This PR optimizes `UrlUtils.header` by memoizing the result based on the user and password pair, as retrieved via `SharedPrefManager`. It avoids continuous string concatenations and Base64 encodings on every access, which occurs in high-frequency scenarios like sync operations and `onBindViewHolder` Glide-header paths. A new test validates that the basicAuthHeader generation correctly utilizes the cache.

---
*PR created automatically by Jules for task [1852748558311848263](https://jules.google.com/task/1852748558311848263) started by @dogi*

## Summary by CodeRabbit

* **Performance**
  * Improved authorization handling by reusing previously generated headers when credentials remain unchanged.
  * Authorization headers are regenerated automatically when credentials change.

* **Bug Fixes**
  * Ensured cached authorization data is cleared when resetting URL utility state.


dogi:
**Review — rating: 62/100**

Adding the cache fields to `resetForTesting()` is the right instinct — static mutable state on an `object` leaks between tests otherwise, and you covered that. But the memoization itself doesn't hold up on either correctness or value.

**The cache is unsynchronized shared mutable state on a singleton.** `cachedUser`, `cachedPwd` and `cachedHeader` are plain `private var` on an `object`, with no `@Volatile` and no lock — and `spmInstance` directly above them *is* `@Volatile`, so the file already acknowledges that this class is touched from multiple threads. `UrlUtils.header` is read on request paths, so concurrent access from `Dispatchers.IO` coroutines is the normal case.

The write order is `cachedUser`, then `cachedPwd`, then `cachedHeader`. Without a happens-before edge, a second thread can observe the first two writes and not the third, which means it passes the `cachedUser == user && cachedPwd == pwd && cachedHeader != null` guard while `cachedHeader` still holds the **previous** credentials' value — and returns a stale `Basic` header. Credentials only change on server switch or re-login, so the window is narrow, but the failure mode is 401s that look like a server problem rather than a client cache bug. If the cache stays, it wants to be one immutable `@Volatile` holder so the triple publishes atomically:

```kotlin
private data class HeaderCache(val user: String, val pwd: String, val header: String)
@Volatile private var cache: HeaderCache? = null
```

**And the caching skips the part that would justify it.** `spm.getUrlUser()` and `spm.getUrlPwd()` are still called on *every* access — they run before the cache check, because they're the cache key. What's saved is only `basicAuthHeader`, which is a string concatenation plus `Base64.encodeToString(…, NO_WRAP)` over a short credential string. That's nanoseconds. So the change takes on a data race and three fields of mutable static state to avoid an operation that doesn't register, while leaving the two `SharedPreferences` lookups in place. If those lookups were the real cost, the cache would need to key off something else entirely — a preference-change listener, or invalidation on login/server switch.

**Nit on the test:** `verify(exactly = 1) { UrlUtils.basicAuthHeader("user1", "pass1") }` depends on `mockkObject(UrlUtils)` spying rather than stubbing, which works but couples the test to that setup. More to the point, it only exercises the single-threaded path — the behaviour actually at risk here is concurrent, and that's not covered.

**Verdict:** needs another round — I'd drop this one unless there's a profile showing `basicAuthHeader` on a hot path; if it stays, publish the cache atomically through a single `@Volatile` holder.

---

dogi:
**Review — rating: 74/100**

CI green on all three checks. The caching is thread-correct, but it keeps credentials resident for the process lifetime to save work that isn't expensive.

**The concurrency is right.** `@Volatile` on `headerCache`, a single read into a local before the comparison, and an immutable `HeaderCache` holding all three values together — so a reader can't observe a half-updated cache. Two threads racing to populate it both compute the same value, so the benign duplicate write costs nothing. `resetForTesting()` clears it alongside `spmInstance`.

**But the cache stores the plaintext password**

```kotlin
private data class HeaderCache(val user: String, val pwd: String, val header: String)
```

`UrlUtils` is a long-lived `object`, so after the first request this holds the user's password in a `String` for as long as the process runs. Previously `getUrlPwd()` was read and the intermediate discarded each call; now there's a permanent heap-resident copy, plus the header itself (which is just base64 of `user:pwd` — reversible, not a hash).

The credentials already live in `SharedPreferences`, so this isn't a new secret so much as a new place it sits — but it's the kind of thing that shows up in a heap dump or a crash report attachment, and CLAUDE.md's security section already flags this project's handling of these particular credentials.

Storing a hash for the invalidation check avoids it entirely, since the field only exists to detect change:

```kotlin
private data class HeaderCache(val credentialsHash: Int, val header: String)
…
val hash = 31 * user.hashCode() + pwd.hashCode()
```

The header still has to be retained — that's the point of the cache — but the password itself doesn't.

**Worth asking whether the memoization earns its keep.** `basicAuthHeader` is a base64 encode of a short string: sub-microsecond, no I/O. Against that, this adds a volatile field, a cache-validity branch on every access, and a lifetime-resident credential. If profiling showed `UrlUtils.header` as hot, the cost is far more likely to be the two `SharedPreferences` reads (`getUrlUser()` / `getUrlPwd()`) that still happen on every call — and those aren't cached here, so the hot path is unchanged.

**Verdict:** correctly implemented, but store a hash rather than the password — and it's worth confirming there's a measurement behind this, since the base64 encode is unlikely to be what's costing anything.

---

dogi:

CI green on all three checks. The credential retention is gone:

```kotlin
private data class HeaderCache(val credentialsHash: Int, val header: String)
…
val hash = 31 * user.hashCode() + pwd.hashCode()
if (currentCache != null && currentCache.credentialsHash == hash) return currentCache.header
```

The cache now keeps only what it needs to detect a change, so no plaintext password sits in a process-lifetime static. The header itself still has to be retained — that's the cache — but that was always going to be true.

Thread-safety is unchanged and still correct: `@Volatile`, a single read into a local before comparison, and an immutable record so no half-updated state is observable. A hash collision would serve a stale header, but at 2⁻³² for a credential change that's not a practical concern, and `resetForTesting()` still clears it.

The open question from last time stands but isn't blocking: `basicAuthHeader` is a base64 encode of a short string, and the two `SharedPreferences` reads on either side of the cache check aren't cached — so if `UrlUtils.header` showed up in a profile, this probably isn't where the time went. Worth a sentence in the description if there's a measurement behind it.

**Verdict:** correct, and no longer retains the password. Marking `ready`.

---

Okuro3499 (dismissed):
```
app/src/main/java/org/ole/planet/myplanet/utils/UrlUtils.kt:41-48 — Cache validity is keyed on 31 * user.hashCode() + pwd.hashCode() instead of the actual values. Fix: compare  
  currentCache.user == user && currentCache.pwd == pwd (as ed17a686a already did), storing user/pwd in HeaderCache instead of a hash. If the plaintext-in-memory concern is genuinely
  wanted, that's better solved by not adding a second copy at all (see suggestion below), not by hashing
```

dogi:
**Re-review — Rating: 74/100** (was 86)

The score drops because the branch has moved back to the version I flagged earlier. Head `d14f856` stores the plaintext password again:

```kotlin
private data class HeaderCache(val user: String, val pwd: String, val header: String)

@Volatile
private var headerCache: HeaderCache? = null
```

An earlier revision had replaced this with a hash of the credentials:

```kotlin
private data class HeaderCache(val credentialsHash: Int, val header: String)
val hash = 31 * user.hashCode() + pwd.hashCode()
if (currentCache != null && currentCache.credentialsHash == hash) return currentCache.header
```

That version is what took this from 74 to 86, and it's gone from the current head. The two most recent commits both read "Memoize `UrlUtils.header` securely and atomically", but the code they land holds the CouchDB password as a plaintext `String` in a process-lifetime `@Volatile` field on a singleton — which is the opposite of what "securely" describes.

To be clear about the severity: the Base64 header itself is trivially reversible, so this isn't a new secret being exposed. What changes is the *retention* — a plaintext copy now lives in a long-lived object field for the life of the process, reachable from any heap dump, rather than existing only as a transient local. The hash version got the same memoization with none of that.

If the revert was deliberate — say the hash comparison was suspected of collisions — that's worth stating in the PR, and the answer would be to widen the hash rather than store the credential. A 32-bit `hashCode` collision would serve a stale header, so it's a fair concern; `user.hashCode()` combined with a SHA-256 of the password, or simply comparing against the *derived header string* rather than the inputs, both avoid it. Comparing the computed header is probably simplest — you're about to produce it anyway only on a miss.

**Otherwise unchanged and still fine:** `@Volatile` with read-into-local before comparison is a correct benign-race memoization (worst case two threads each compute the same header), `resetForTesting()` clears the cache, and the new `testHeaderCaching` covers both the hit and the credential-change path.

**Nit:** the test's import block puts `io.mockk.verify` before `io.mockk.mockkObject` / `io.mockk.mockkStatic`.

**Verdict:** `change`. Restore the credential-independent cache key — hash, or compare the derived header — and this goes straight back to `ready`.

---

dogi:
**Re-review — Rating: 72/100** (was 74)

The security problem is fixed — the plaintext password is no longer retained. But the fix chosen makes the memoization cost more than the computation it was meant to avoid, so the PR now works against its own purpose.

**Fixed:** `HeaderCache` holds `userHash: Int` and `pwdHash: String` instead of the credentials, so nothing recoverable sits in the `@Volatile` field. The test import order is corrected too.

**The problem — this is now almost certainly slower than not caching at all.**

```kotlin
private fun sha256(input: String): String {
    val bytes = input.toByteArray()
    val md = MessageDigest.getInstance("SHA-256")
    val digest = md.digest(bytes)
    return digest.joinToString("") { "%02x".format(it) }
}
```

Every read of `UrlUtils.header` — cache hit or miss — now performs:

- a `MessageDigest.getInstance("SHA-256")` provider lookup and instance allocation,
- a SHA-256 digest,
- and **32 separate `String.format("%02x", …)` calls**, each of which parses the format string at runtime.

Compare that to what it's avoiding: `basicAuthHeader` is one string concatenation and one `Base64.encodeToString` over a short byte array. The hex loop alone is heavier than that; `String.format` is one of the slowest ways to produce hex on the JVM.

`header` is read on every authenticated request, so this is squarely in the hot path the PR exists to optimise.

**The earlier revision had this right.** The version that scored 86 compared cheap integer hashes:

```kotlin
private data class HeaderCache(val credentialsHash: Int, val header: String)
val hash = 31 * user.hashCode() + pwd.hashCode()
```

That satisfies both goals at once — `hashCode()` is lossy and 32-bit, so it retains nothing recoverable, and it costs two cheap hash computations instead of a digest. Going back to it is the right move.

If a 32-bit collision genuinely worries you, note what a collision actually costs here: serving a stale `Basic` header, which fails authentication and gets corrected on the next credential change. That's a recoverable, self-correcting failure, not a security hole — and it requires the *combined* hash of username and password to collide, which won't happen in a two-entry cache with one live credential pair.

**Smaller points:**

- `userHash = user.hashCode()` hashes a value that isn't secret. If you keep any hashing, store `user` plainly — it's collision-free, costs nothing, and only the password needs protecting.
- Import order: `java.security.MessageDigest` is placed before `java.net.URLEncoder`; `java.net` sorts first.

**Verdict:** `change`. Restore the `hashCode()`-based cache key. The security concern was *retention*, and `hashCode()` already solves it without putting a digest and 32 `String.format` calls on every request.

---

dogi:

Third revision, and this one lands. The cache key is now `(user: String, pwdHash: Int)`:

```kotlin
private data class HeaderCache(val user: String, val pwdHash: Int, val header: String)
```

That resolves both problems at once. Revision 1 kept the password in a process-lifetime static; revision 2 fixed that with SHA-256 but paid `MessageDigest.getInstance` plus 32 `String.format("%02x")` calls on **every** read of `header`, cache hit included — more work than the single concat + Base64 encode it was memoizing. `hashCode()` costs neither: `String` caches its hash after first computation and `SharedPreferences` hands back the same interned instance from its in-memory map, so the steady-state cost is a field read, a short string compare and an int compare.

It's also better than what I suggested last round. I proposed folding both into one int (`31 * user.hashCode() + pwd.hashCode()`), which admits a collision on the *username* — that would serve one user's header to another. Keeping the username exact and hashing only the password confines any collision to the same user, where the worst case is a stale header that fails auth. Right call.

Thread-safety is correct: `headerCache` is `@Volatile`, the payload is an immutable `data class`, and the getter reads into a local (`val currentCache = headerCache`) before testing its fields, so there's no torn read between the check and the return. Two threads racing a miss recompute independently and one write wins — both writes are equivalent, so that's benign. `resetForTesting()` clears the cache alongside `spmInstance`, which keeps the existing tests hermetic.

**The test doesn't cover the field it was added for.** `testHeaderCaching` changes user *and* password together:

```kotlin
every { mockSpm.getUrlUser() } returns "user2"
every { mockSpm.getUrlPwd() } returns "pass2"
```

The `user != user` comparison alone already misses on that, so `pwdHash` could be dropped from the cache key entirely and this test would still pass. The interesting case — and the one that would be a live auth bug — is a password change with the username unchanged: same `user1`, `pass1` → `pass2`, assert a fresh `basicAuthHeader("user1", "pass2")` call and a new header. That's the scenario a user hits after rotating their PIN. Worth adding; it's three lines on top of what's already there.

Nit, not blocking: a `pwd.hashCode()` collision under an unchanged username pins a wrong header until the credentials change again, rather than self-correcting. At 32 bits against the handful of passwords a device sees in its lifetime this is not a practical concern, and the alternatives (retaining the password, or hashing it properly) are the two things this revision is deliberately avoiding. Noting it only so the trade-off is on the record.

CI green on all four checks (build default + lite, test shards 1/2 and 2/2), merges cleanly against master.

**Verdict:** `ready`. Add the same-user-new-password test when convenient — it doesn't need to gate the merge.

---

Okuro3499 (dismissed):
```
Problem: Cache validity is checked via currentCache.pwdHash == pHash (a 32-bit String.hashCode() comparison) instead of comparing the actual password. If the password changes to a
different string that happens to produce the same hashCode() (collisions are not exotic for String.hashCode(), a simple 31-multiplier polynomial hash), the getter silently        
returns the old Basic-Auth header for the new password — i.e., it can send stale/wrong credentials to the server without any error, log, or signal. This is exactly the kind of    
silent-wrong-behavior bug that must never exist on an auth path.
```

dogi:
**Re-review — Rating: 70/100** (was 86)

The cache key went back to storing the password in plaintext:

```kotlin
private data class HeaderCache(val user: String, val pwd: String, val header: String)
```

That's the thing this PR spent three revisions moving away from, and the revision I rated 86 had already solved it better than my own suggestion. `HeaderCache(user: String, pwdHash: Int, header: String)` matched the username exactly — so no cross-user collision was possible — and reduced the password to a 32-bit lossy hash that nothing can be recovered from. It cost the same: `String` caches its `hashCode` after first computation, and `SharedPreferences` returns the same interned instance from its in-memory map, so the steady-state comparison was a field read, a short string compare and an int compare. Full string equality on the password isn't faster than that; it's the same shape with a retained secret attached.

The commit message is `Memoize UrlUtils.header securely and robustly`, which is the opposite of what the diff does. Worth fixing before squash — a message that asserts a security property the code doesn't have is the thing a later reader trusts instead of re-checking.

**On whether the retention actually matters:** `SharedPrefManager` already holds the password, so this isn't a new secret in the process. What changes is lifetime and reachability. `UrlUtils` is a Kotlin `object` — alive for the whole process, no eviction, and the only thing that clears `headerCache` is `resetForTesting()`, which is `@VisibleForTesting` and never runs in production. So a heap dump or memory scrape gets one more copy at a fixed, well-known location. Given CLAUDE.md's security section is explicitly about pulling credentials *out* of easily-recoverable places, adding a static plaintext copy runs the wrong way — especially when the alternative in this PR's own history was free.

Everything else about the implementation is still right, and I want to be clear I'm not asking for a rewrite. `@Volatile` on the reference, an immutable `data class` payload, and reading into a local (`val currentCache = headerCache`) before testing its fields means there's no torn read between check and return. Two threads racing a miss recompute independently and one write wins, and both writes are equivalent. `resetForTesting()` clearing the cache keeps the existing tests hermetic. That's all unchanged and all correct.

**The test still doesn't cover the case the key exists for.** `testHeaderCaching` changes user and password together, so the `user != user` comparison alone already forces a miss — the `pwd` field could be dropped from the data class and this test would still pass. The scenario worth pinning is the same username with a rotated password, which is what a user hits after changing their PIN.

**Verdict:** `change` — restore `pwdHash: Int` from the previous revision, and fix the commit message. That version was ready to merge.

---

dogi:
**Re-review — Rating: 88/100** (was 70)

Both things landed. The cache key is back to `HeaderCache(user: String, pwdHash: Int, header: String)` — exact username match so a collision can never cross users, password reduced to a 32-bit lossy hash so nothing recoverable is retained, and no slower than full string equality since `String` caches its `hashCode` and `SharedPreferences` hands back the same interned instance.

And the test now covers the case I'd asked about twice — same username, rotated password:

```kotlin
every { mockSpm.getUrlUser() } returns "user2"
every { mockSpm.getUrlPwd() } returns "pass3"

val header4 = UrlUtils.header
assertEquals("Basic dXNlcjI6cGFzczM=", header4)
…
verify(exactly = 1) { UrlUtils.basicAuthHeader("user2", "pass3") }
```

That's the one a user actually hits after changing their PIN, and it's the only case that can distinguish "the `pwdHash` field is doing work" from "the username comparison alone happens to catch it". With all three `verify(exactly = 1)` assertions in place, dropping any field from the cache key would now fail a test.

**One file to delete before this merges.** `app/src/main/java/org/ole/planet/myplanet/utils/UrlUtils.kt.orig` is committed — 162 lines, a `git merge` conflict backup. The build doesn't notice it because the Kotlin source set matches `**/*.kt` and `.kt.orig` doesn't, which is why CI is green, but it shouldn't be in the tree. It's also the least convenient possible copy to leave behind: line 15 is

```kotlin
private data class HeaderCache(val user: String, val pwd: String, val header: String)
```

— the plaintext-password version this PR just moved away from. Anyone grepping for credential handling later finds two implementations and has to work out which one is live.

`.gitignore` has no `*.orig` or `*.rej` rule, so this can happen again on the next conflicted merge. Worth adding both lines while you're in there.

Everything else I checked before still holds: `@Volatile` on the reference with an immutable `data class` payload, the getter reading into a local before testing its fields so there's no torn read, benign write races between two threads that miss, and `resetForTesting()` clearing the cache so the rest of `UrlUtilsTest` stays hermetic.

**Verdict:** `ready` — `git rm` the `.orig` file and this is done. The commit history is seven revisions of the same idea; worth squashing to the final one, and the message on `df4abe8` is accurate this time.

---

Okuro3499 (changes requested):
```
UrlUtils.kt:41,43 compares pwd.hashCode() instead of the plaintext pwd, which lets a hash collision serve a stale/wrong 
  auth header — a real correctness regression
```

PR #15615, opened 2026-08-13
sync: less upload shelf api interface is more (fixes #15734)

### LINKED ISSUE #15734: Remove unused ApiInterface from UploadToShelfService
Removed the unused `ApiInterface` dependency from `UploadToShelfService`'s constructor and the DI module since `ApiInterface` was never used within the service. Also removed it from the unit test initialization.

dogi:
Removed the unused `ApiInterface` dependency from `UploadToShelfService`'s constructor and the DI module since `ApiInterface` was never used within the service. Also removed it from the unit test initialization.

---
*PR created automatically by Jules for task [12692888376142547936](https://jules.google.com/task/12692888376142547936) started by @dogi*


dogi:
**Review — rating: 90/100**

Clean, complete removal. The dependency is dropped from all four places it appeared — the constructor, the `@Provides` argument list in `ServiceModule`, the import in `UploadToShelfService`, and the `mockk` in `UploadToShelfServiceTest` — so nothing is left dangling. A green build is sufficient proof the field was genuinely unused, since an unused constructor parameter that was actually referenced would fail to compile once removed.

I checked the one thing that could have gone wrong quietly: `ServiceModule` still references `ApiInterface` three times for its other providers, so removing this usage doesn't strand the import. CI green (build default + lite, test default).

**Nit — unrelated to your change, but visible from it:** `UploadToShelfService` carries both an `@Inject constructor` *and* a hand-written `@Provides` in `ServiceModule` that constructs it manually. Hilt only needs one; the `@Provides` wins, which makes the `@Inject constructor` annotation decorative and means every future dependency change has to be made in two places — as this PR just demonstrated. Deleting the `@Provides` and letting constructor injection do the work would make the next one a single-line edit. Worth its own small PR rather than folding in here.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #15616, opened 2026-08-13
courses: smoother progress repository data fetching (fixes #15735)

### LINKED ISSUE #15735: Refactor submissions grouping in fetchCourseData to use HashSet
Replaces the linear course-ID scans in `ProgressRepositoryImpl.fetchCourseData` with a more robust `HashSet` lookup. 

The previous implementation grouped every submission via `courseIds.firstOrNull { submission.parentId?.contains(courseId) == true }`, which resulted in O(N * M) string scans and allowed for ambiguous substring matches (e.g. `course1` incorrectly matching a `parentId` containing `course10`).

The new implementation:
1. Parses the course ID once from the established `examId@courseId` format.
2. Validates it against a prebuilt `HashSet` for O(1) lookups.
3. Groups in a single pas

dogi:
Replaces the linear course-ID scans in `ProgressRepositoryImpl.fetchCourseData` with a more robust `HashSet` lookup. 

The previous implementation grouped every submission via `courseIds.firstOrNull { submission.parentId?.contains(courseId) == true }`, which resulted in O(N * M) string scans and allowed for ambiguous substring matches (e.g. `course1` incorrectly matching a `parentId` containing `course10`).

The new implementation:
1. Parses the course ID once from the established `examId@courseId` format.
2. Validates it against a prebuilt `HashSet` for O(1) lookups.
3. Groups in a single pass.
4. Maintains an explicit, tested fallback for malformed or legacy parent IDs.
5. Adds unit tests specifically covering normal, malformed, missing, and substring-collision parent IDs.

---
*PR created automatically by Jules for task [11656045341491080307](https://jules.google.com/task/11656045341491080307) started by @dogi*


dogi:
**Review — rating: 76/100**

This is more than the performance win the title advertises — it fixes a real correctness bug, and that deserves to be the headline.

The old grouping was `courseIds.firstOrNull { submission.parentId?.contains(it) == true }`, a substring scan in list order. With `courseIds = ["course1", "course10"]` and a submission whose `parentId` is `exam10@course10`, `contains("course1")` matches first and the submission is filed under the **wrong course**. Splitting on `@` and looking up `parts[1]` in a `HashSet` resolves that exactly, because it matches the whole segment rather than a substring — and the `examId@courseId` shape is the format `SubmissionsRepositoryImpl` actually constructs. Hoisting `courseIdsSet` out of the `groupBy` also turns an O(submissions × courses) scan into O(submissions).

**Two things to address.**

**The test doesn't test the fix.** `testFetchCourseData_SubmissionsGrouping` sets up the right four cases — well-formed, legacy, null, and the `exam10@course10` collision — but `examDao.getByCourseIds` is stubbed to `emptyList()`, so no exams exist and the only assertions are `data.size() == 2` and `mistakes == 0` for both courses. Those pass whether `sub4` is grouped under `course1` or `course10`, which is precisely the distinction the change exists to get right. The fixture names the bug and the assertions can't see it. To actually pin it, the test needs exams and steps configured so that a mis-grouped submission produces a different `mistakes`/progress value per course — otherwise this is a regression guard that will stay green through a regression.

**The fallback's tie-breaking changed silently.** When the `@`-split doesn't yield a known course id, the code falls back to `courseIdsSet.firstOrNull { parentId.contains(it) }` — iterating the *set*, where it previously iterated the ordered `courseIds` list. For a legacy `parentId` that contains two course ids as substrings, the winner is now determined by hash iteration order rather than list order. It's stable for a given set of ids, so this isn't flakiness, but it is a different answer from before on exactly the malformed inputs the fallback exists to handle. Iterating `courseIds` in the fallback while keeping the set for the fast path would preserve the old tie-break and cost nothing.

**Nit:** `parts.size == 2` sends any `parentId` containing more than one `@` down the substring path. That's a reasonable conservative choice, but a comment saying so would help — it currently reads as though the split is expected to always yield two parts.

**Verdict:** needs another round — the code change is good, but make the test assert the grouping it was written for.

---

dogi:
**Review — rating: 84/100**

The title undersells this. It reads as a hash-lookup optimization, but it's a **correctness fix**, and the test proves it.

The old grouping key was `courseIds.firstOrNull { submission.parentId?.contains(courseId) == true }` — first courseId that appears *anywhere* in `parentId`. With `courseIds = ["course1", "course10"]` and `parentId = "exam10@course10"`, `"course1"` matches as a substring and wins on ordering, so every submission for `course10` was silently attributed to `course1`. That's inflated mistake counts on one course and zeros on the other. `testFetchCourseData_SubmissionsGrouping` sets up exactly that pair and asserts 15/20 rather than 35/0. Good test — it fails on the old code for the right reason, and the inline comments say what the failure would look like.

The design is appropriately conservative: well-formed `examId@courseId` gets exact-segment matching, anything else (`"course1_legacy"`, no `@` at all) keeps the legacy substring behavior, and `sub2` in the test pins that down so the fallback doesn't rot.

**One thing to tighten** — the fallback is reached in a case where it can only do harm:

```kotlin
if (parts.size == 2 && courseIdsSet.contains(parts[1])) {
    parts[1]
} else {
    courseIds.firstOrNull { parentId.contains(it) }
}
```

`getExamSubmissionsByUser(userId)` returns the user's exam submissions across *all* courses, not just the ones in `courseIds`. So a well-formed `parentId` whose course segment simply isn't in this list — a submission for a course the user has since dropped, say — fails the `contains` check and falls into the substring scan. For a well-formed id the answer there is unambiguously `null`; the substring scan can't find anything correct, but it *can* find something wrong, which is the same collision class the PR just fixed. It's also the O(n) branch, and on a user with submissions spread across many courses it'll be the branch that runs most.

Splitting the conditions fixes both at once:

```kotlin
if (parts.size == 2) {
    parts[1].takeIf(courseIdsSet::contains)
} else {
    courseIds.firstOrNull { parentId.contains(it) }
}
```

Now the well-formed path is O(1) on hits *and* misses, the substring scan is reserved for genuinely malformed ids, and the collision fix covers the whole well-formed space instead of just the enrolled subset. The existing tests should all still pass unchanged.

**Nits, neither blocking:**

- `examDao.getByCourseIds(courseIds).map { it }` and `submissionDao.getExamSubmissionsByUser(userId).map { it }` are identity maps that copy both lists for nothing. Pre-existing, but you're editing the second one already.
- The test indexes `data[0]` / `data[1]`, which assumes the output array preserves `mycourses` order. It's deterministic today; asserting by looking up `courseId` would survive a future ordering change.
- Stray double blank line before `testFetchCourseData_SubmissionsGrouping`.

**Verdict:** `ready`. Correct, CI green on all three checks, and it ships with a test that captures the actual bug rather than just the refactor. The `parts.size == 2` split is worth doing — either here or as a follow-up — but it narrows a pre-existing bug rather than introducing one, so I'm not holding the PR for it.

---

Okuro3499 (approved):

PR #15617, opened 2026-08-13
sync: smoother transaction checkpoint applying (fixes #15736)

### LINKED ISSUE #15736: Refactor TransactionSyncManager intermediate sync checkpoints to use apply()
Fixes a performance issue where the sync batch loop blocked on synchronous `SharedPreferences.commit()` writes on every iteration. 

- Modifies `TransactionSyncManager.kt` lines 198 and 313.
- Replaces intermediate `putInt(checkpointKey, skip).commit()` with `.apply()`.
- The final `.remove(checkpointKey).commit()` at the end of the sync process remains untouched for durability.
- Updates `TransactionSyncManagerCheckpointTest.kt` to mock `.apply()` to fix failing CI tests.

dogi:
Fixes a performance issue where the sync batch loop blocked on synchronous `SharedPreferences.commit()` writes on every iteration. 

- Modifies `TransactionSyncManager.kt` lines 198 and 313.
- Replaces intermediate `putInt(checkpointKey, skip).commit()` with `.apply()`.
- The final `.remove(checkpointKey).commit()` at the end of the sync process remains untouched for durability.
- Updates `TransactionSyncManagerCheckpointTest.kt` to mock `.apply()` to fix failing CI tests.

---
*PR created automatically by Jules for task [7678460224176550553](https://jules.google.com/task/7678460224176550553) started by @dogi*


dogi:
**Review — rating: 80/100**

The cost being removed is real. `commit()` does a synchronous write-and-fsync of the whole preferences file, and this runs *twice per batch* inside the pagination loop — once at line 195 before the page is fetched, once at line 310 after it's committed to Room. Over a full sync that's hundreds of blocking disk flushes on exactly the storage-constrained devices this app targets. The ignored `Boolean` return means nothing is lost at the call sites, and `apply()` still updates the in-memory map synchronously, so a *reader* in the same process sees the new checkpoint immediately either way.

The durability question resolves in your favour too, though the PR should say so out loud. `apply()` can lose its disk write if the process is hard-killed, which is precisely the interruption a checkpoint exists for — but the consequence is bounded: resume restarts from a stale `skip` and re-processes one page of upserts, which are `onConflict = REPLACE` and therefore idempotent. So the worst case is redundant work, not corruption. That's an acceptable trade for dropping the fsyncs, and it's worth stating in the description rather than leaving a reviewer to derive it.

**The comment directly above the changed line now contradicts the code:**

```kotlin
// Persist progress immediately after a batch is committed so an interruption
// resumes past it rather than re-processing the just-inserted page.
```

"Persist progress immediately" is the one guarantee `apply()` gives up. Leaving that sitting on top of an `apply()` will mislead the next reader into believing the checkpoint is crash-durable when it isn't. Please reword it to say the checkpoint is best-effort and re-processing a page is safe because the writes are idempotent.

Worth considering, if you'd rather keep the guarantee where it counts: the two sites aren't equally important. Line 195 writes `skip` *before* the page is even fetched — nothing is at stake there and `apply()` is clearly right. Line 310 is the one the comment is about, and it's the only place where durability buys anything. Keeping `commit()` at 310 and `apply()` at 195 halves the fsyncs while preserving the post-batch guarantee.

**Nit:** the test now stubs both `editor.commit()` and `editor.apply()`. If `commit()` is no longer reachable from this path, that stub is dead and should go — otherwise a future regression back to `commit()` passes silently. The test also only captures the checkpoint *values*, never asserting which flush method was used, so nothing pins the behaviour this PR changes.

**Verdict:** mergeable — please fix the comment before merging, since it now asserts the opposite of what the code does.

---

Okuro3499 (approved):

PR #15618, opened 2026-08-13
enterprises: smoother members reports payload diffing (fixes #15737)

### LINKED ISSUE #15737: Add payload-only rebinds for MembersAdapter and EnterprisesReportsAdapter
- Add specific payloads to `MembersAdapter` and `EnterprisesReportsAdapter` to avoid full visual rebinds when toggling chrome visibility (like menus or edit/delete buttons).
- Ensure that `DiffUtil` remains the primary diffing source, but targeted `notifyItemRangeChanged(..., payload)` is used when simple flags toggle.
- Add Robolectric tests for both adapters to verify the correct payload is emitted and processed.

dogi:
- Add specific payloads to `MembersAdapter` and `EnterprisesReportsAdapter` to avoid full visual rebinds when toggling chrome visibility (like menus or edit/delete buttons).
- Ensure that `DiffUtil` remains the primary diffing source, but targeted `notifyItemRangeChanged(..., payload)` is used when simple flags toggle.
- Add Robolectric tests for both adapters to verify the correct payload is emitted and processed.

---
*PR created automatically by Jules for task [14665757403789768820](https://jules.google.com/task/14665757403789768820) started by @dogi*


dogi:
**Review — rating: 78/100**

The core change is right. Both adapters were calling `notifyItemRangeChanged(0, itemCount)` with no payload on a leader/membership flip, forcing a full rebind of every visible row to toggle a couple of view visibilities. Passing a payload and handling it in the three-arg `onBindViewHolder` is exactly the fix, and `MembersAdapter`'s version is the better of the two — iterating `payloads` with a `when` rather than the old `payloads[0] as Bundle` correctly handles RecyclerView accumulating several payloads onto one position, and drops an unchecked cast that would have thrown the moment a non-`Bundle` payload arrived. Which is precisely what this PR introduces, so that rewrite wasn't optional. Adding real Robolectric tests for two previously untested adapters is a genuine improvement.

**Both overrides silently skip binding when a payload isn't recognised.** In `EnterprisesReportsAdapter`:

```kotlin
if (payloads.isNotEmpty()) {
    payloads.forEach { payload -> if (payload == PAYLOAD_KEY_NON_TEAM_MEMBER_CHANGED) { … } }
} else {
}
```

A non-empty `payloads` containing nothing recognised falls through every branch and the row is left showing stale content — no full bind, no partial update. `MembersAdapter` has the same shape: a `when` with no `else`. Neither is reachable today, since each adapter has exactly one payload source (`EnterprisesReportsAdapter.diffCallback` defines no `getChangePayload` at all), but this PR is establishing the pattern that other adapters will copy, and the failure mode is invisible — a stale row, not a crash. The safe shape tracks whether anything handled the payload and falls back to the full bind otherwise.

**The tests are non-vacuous only by accident of `AsyncListDiffer`'s fast paths.** Every assertion sits inside a `submitList` commit callback, and those callbacks only run synchronously here because of two implementation details: the first `submitList` on an empty differ short-circuits and invokes the callback inline, and the second one passes *the same list instance*, hitting `AsyncListDiffer`'s `newList == mList` identity check. Change either — copy the list, or add an element so the diff actually runs — and the callback is posted to Robolectric's paused main looper, never executes, and every assertion inside it is silently skipped while the test still reports green. That's the worst kind of test: one that stops checking without telling you. Hoisting the assertions out of the callback, or configuring the adapter with an `AsyncDifferConfig` using a direct executor, makes them robust.

**Nit:** `EnterprisesReportsAdapter`'s partial branch duplicates the `nonTeamMember` visibility block verbatim from the full `onBindViewHolder` rather than calling a shared helper, so the two will drift the first time that rule gains a condition. I also couldn't verify from the diff that `nonTeamMember` affects *only* `edit`/`delete` visibility in the full bind — if it also gates a click listener further down, the partial path misses it. Worth confirming.

**Verdict:** needs another round — add the unrecognised-payload fallback in both adapters and make the tests independent of differ fast paths.

---

dogi:
**Review — rating: 78/100**

Careful work. I checked the thing that actually matters for a payload refactor — whether each payload branch does *everything* the full bind would have done for that state change — and in both adapters it does:

- `nonTeamMember` is read in exactly one place (`EnterprisesReportsAdapter:114`), and `setNonTeamMemberVisibility` is a clean extraction of it.
- `isLoggedInUserTeamLeader` is read in exactly one place (`MembersAdapter:161`, inside `checkUserAndShowOverflowMenu`), and that function handles both directions — the `else` branch sets `icMore` back to `GONE`, so a leader losing the role gets the menu hidden rather than left behind.

The `MembersAdapter` restructure deserves a specific mention. The old handler was `payloads[0] as Bundle` — an unchecked cast that would have thrown `ClassCastException` the moment a non-Bundle payload appeared. Introducing `PAYLOAD_KEY_LOGGED_IN_USER_LEADER_CHANGED` is exactly that moment, and the PR converts to a `when` over every payload in the same commit. Right order, no window where the cast could fire.

Test coverage is proportionate — 257 lines for 52 lines of production change, covering payload emission, both visibility directions on each adapter, and the unknown-payload fallback. CI green on all three checks.

**Must fix — fully-qualified name** in `EnterprisesReportsAdapterTest`:

```kotlin
assertEquals(context.getString(org.ole.planet.myplanet.R.string.team_financial_report, "Test Team"), ...)
```

Project convention is `import` + short name. The wrinkle is that `setUp` also references `androidx.appcompat.R.style.Theme_AppCompat`, so you can't leave both bare — import the project's `R` and use `R.string.team_financial_report` here, and alias the other (`import androidx.appcompat.R as AppCompatR`) so both read short.

**Worth tightening — the `handled` flag silently drops payloads:**

```kotlin
payloads.forEach { payload -> if (payload == PAYLOAD_…) { …; handled = true } }
if (!handled) { super.onBindViewHolder(holder, position, payloads) }
```

If `payloads` ever holds one recognized entry and one unrecognized entry, `handled` is true and the unrecognized one is discarded with no rebind. That can't happen today — neither `DiffUtils.itemCallback` here passes a `payloadSelector`, so the only payloads are the ones these adapters emit themselves. But the helper does support `getChangePayload`, so the day someone adds a selector to either diff callback, the failure is a silently stale row rather than anything that shows up in a test. Inverting it — track whether *any* payload went unrecognized, and full-bind if so — costs nothing and removes the trap.

**Nits:**

- The `handled` block is byte-identical across the two adapters. A small shared extension over `MutableList<Any>` would carry the corrected semantics above to both at once.
- `submitList(newList) { notifyItemRangeChanged(0, itemCount, PAYLOAD) }` re-binds every visible row right after the diff already rebound the rows it changed. Correct, just slightly redundant; pre-existing shape, not something this PR introduced.

**Verdict:** the refactor is `ready`-grade — correct, complete, and well tested. Holding at `change` only for the FQN, which is a one-line fix. Do that (and ideally invert the `handled` logic while you're in there) and this goes straight through.

---

dogi:

Both findings addressed, and the inversion is done properly in both adapters.

**FQN resolved with the alias.** `import androidx.appcompat.R as AppCompatR` alongside `import org.ole.planet.myplanet.R` lets both read short, which was the awkward part — you can't just import two `R`s. `AppCompatR.style.Theme_AppCompat` and `R.string.team_financial_report` are both clean now.

**`handled` → `unhandled` is correct in both handlers.** `EnterprisesReportsAdapter` is the simple case, and `MembersAdapter` catches the subtle one:

```kotlin
is Bundle -> {
    if (payload.containsKey(PAYLOAD_KEY_LEADER)) { … } else { unhandled = true }
}
PAYLOAD_KEY_LOGGED_IN_USER_LEADER_CHANGED -> { checkUserAndShowOverflowMenu(holder.binding, position) }
else -> unhandled = true
```

A `Bundle` arriving *without* the leader key now falls through to a full bind rather than being silently swallowed — that's the case the old flag would have missed even after the outer inversion, and it's handled.

Worth naming the consequence so it isn't mistaken for a bug later: with a mixed payload list, the recognized handlers run *and* then `super.onBindViewHolder` does a full bind on top. That's redundant work, but it's the safe direction — the full bind is authoritative and overwrites whatever the partial handlers set. Correctness over a saved rebind is the right trade here.

**Still holds from last round:** each payload branch is complete. `nonTeamMember` is read only in `setNonTeamMemberVisibility`, `isLoggedInUserTeamLeader` only in `checkUserAndShowOverflowMenu`, and that function handles both directions — its `else` sets `icMore` back to `GONE`, so a demoted leader loses the menu.

**Two nits, neither blocking:**

- `import androidx.appcompat.R as AppCompatR` sits after `org.junit.runner.RunWith`. Aliased or not, it sorts by its real package, so it belongs at the top of the `androidx` group beside `ApplicationProvider`. (`org.ole.planet.myplanet.R` is correctly placed.)
- No test covers the mixed case you just fixed — one payload list holding both a recognized entry and an unrecognized one, asserting the full bind still runs. It's the branch the inversion exists for, and the existing `withUnknownPayload` test is most of the scaffolding already.

**Verdict:** `ready`. Complete payload branches, a safe fallback, 257 lines of Robolectric coverage, CI green on all three checks.

---

Okuro3499 (approved):

PR #15620, opened 2026-08-13
sync: smoother repository json tree mapping (fixes #15738)

### LINKED ISSUE #15738: Replace toJson/fromJson round-trips with toJsonTree
Replaces `toJson → fromJson/parseString` round-trips with `toJsonTree` in SyncRepositoryImpl and MyTeam as requested.

dogi:
Replaces `toJson → fromJson/parseString` round-trips with `toJsonTree` in SyncRepositoryImpl and MyTeam as requested.

---
*PR created automatically by Jules for task [16055201641869781857](https://jules.google.com/task/16055201641869781857) started by @dogi*


dogi:
**Review — rating: 72/100**

The `SyncRepositoryImpl` change is a clean win and I'd take it as-is. `gson.fromJson(gson.toJson(batch), JsonArray::class.java)` serialises a list to a string and immediately re-parses it; `gson.toJsonTree(batch)` builds the same `JsonArray` directly, and `JsonObject.add` takes a `JsonElement`, so the types line up. For a homogeneous `List<String>` the reflective `src.getClass()` overload is safe.

**The two `MyTeam` sites are a different case, and I don't think `toJsonTree` is the right substitution there.** In both, `` `object` `` is *already* a `JsonObject` built up by `addProperty` calls immediately above. So the old code serialised a `JsonObject` to a string and re-parsed it into a `JsonObject` — which means one of two things is true, and the PR should establish which:

1. **It was pointless**, in which case the correct fix is `return \`object\`` with no Gson call at all. `toJsonTree` still dispatches through the `JsonElement` type adapter and walks the tree to build a copy; returning the object directly skips even that.
2. **It was load-bearing null-stripping**, in which case `toJsonTree` may not preserve it. `JsonUtils.gson` is a bare `Gson()` (JsonUtils.kt:13), so `serializeNulls` is `false`. That means `gson.toJson(jsonObject)` **omits** properties whose value is `JsonNull` — and `MyTeam` feeds nullable fields straight into `addProperty("name", team.name)`, `addProperty("docType", team.docType)` and friends, each of which stores `JsonNull` when the source is null. The round-trip was therefore dropping those keys from the document. Gson's `JsonTreeWriter` is well known not to honour `serializeNulls` the way `JsonWriter` does, so `toJsonTree` would keep them — and these objects are serialised for upload to CouchDB, where the difference between "key absent" and `"key": null` is a real difference in the stored document.

I'm not certain enough about Gson 2.14's `JsonTreeWriter` to call that a definite regression, which is exactly the problem: this needs a test, not a reviewer's recollection. A three-line assertion that serialising a `MyTeam` with a null `name`/`docType` produces the same key set before and after would settle it. If the keys should be dropped, keep an explicit filter rather than relying on a serialisation side effect; if they shouldn't, note that the PR intentionally changes the uploaded payload.

**Nit:** `JsonParser` appears on three lines in `MyTeam.kt` — the import plus the two usages this PR removes — so the import goes dead here.

**Verdict:** needs another round — split out the `SyncRepositoryImpl` change, which is unambiguously good, and settle the null-key question for `MyTeam` with a test.

---

dogi:
**Review — rating: 82/100**

Both changes are correct. I verified the equivalence rather than taking it on faith, because the `MyTeam` one is subtler than it looks.

**`SyncRepositoryImpl`** — `gson.fromJson(gson.toJson(batch), JsonArray::class.java)` → `gson.toJsonTree(batch)` is unambiguously right. Serializing a list to a string and immediately reparsing it into a tree is pure waste, and this runs per batch inside the sync loop. `JsonObject.add` takes a `JsonElement`, so the types line up with nothing lost.

**`MyTeam.serialize`** — this one hinges on two things that happen to hold:

1. The round-trip you're replacing wasn't a no-op — it was a *null strip*. `JsonUtils.gson` is a plain `Gson()`, so `serializeNulls` is false, `toJson` omits null-valued members, and reparsing gives back an object without them. Your comment names this correctly, which is more than the original code did.
2. The replacement filters `object.keySet()` — top level only. That's equivalent **because `serialize(team)` builds a strictly flat object**: every member is an `addProperty` scalar, with no nested `JsonObject` or `JsonArray` anywhere. Nesting only enters in the three-arg `serialize(team, courses, coursesResourcesMap)` overload, which calls this one first and *then* attaches the arrays — after the strip point. So there's no nested `JsonNull` for a shallow filter to miss.

Worth stating plainly since it's the load-bearing assumption: **if anyone later adds a nested object to `serialize(team)`, the shallow strip silently stops matching the old behavior.** A comment at the top of the function, or a `require`-style note, would make that trap visible to the next person.

Also correct: materializing the filter into a list before removing. Iterating `keySet()` while calling `remove` would have thrown `ConcurrentModificationException`, and `filter` returning a fresh `List` avoids it.

**Cleanups worth doing:**

- The strip block is duplicated verbatim, comment and all, at both return sites. A `private fun JsonObject.stripNulls(): JsonObject` used as `return \`object\`.stripNulls()` collapses six lines to two and makes the third early return — the `isDeletePending` path, which doesn't strip — a deliberate-looking choice rather than an oversight. (It's fine as-is: that path only carries `_deleted` plus `_id`/`_rev` via `JsonUtils.addString`, which already skips null and empty, so no nulls can reach it.)
- The PR title says "Replace toJson/fromJson round-trips with toJsonTree", but the `MyTeam` half doesn't use `toJsonTree` — it does explicit stripping. Two different techniques for two different shapes, which is the right call; the title just undersells it.

**Test gaps** — `MyTeamTest` is a reasonable start but leaves the interesting cases open:

- It asserts `_id` survives, but `_id` is added through `JsonUtils.addString`, which is a *different* path from the `addProperty` calls the strip actually operates on. So nothing currently proves a non-null `addProperty` value is preserved. Setting `team.name = "X"` and asserting `serialized.get("name").asString == "X"` would close that — right now the test would still pass if the strip removed everything.
- The `docType == "resourceLink"` early return is the other line you changed and has no coverage at all.
- No test for the `toJsonTree` change.

**Verdict:** `ready`. Both replacements are behavior-preserving — I checked the Gson config and the object shape that make them so — and they remove real per-upload and per-batch waste. The helper extraction and the `name` assertion are worth folding in, but neither blocks.

---

Okuro3499 (approved):

PR #15621, opened 2026-08-13
life: smoother list adapter caching (fixes #15739)

### LINKED ISSUE #15739: fix: cache LifeFragment adapter and touch helper
Fixes stacking drag-and-drop callbacks on `LifeFragment` due to recreated `ItemTouchHelper` and redundant flow observers.

dogi:
Fixes stacking drag-and-drop callbacks on `LifeFragment` due to recreated `ItemTouchHelper` and redundant flow observers.

---
*PR created automatically by Jules for task [11189212974883818473](https://jules.google.com/task/11189212974883818473) started by @dogi*

## Summary by CodeRabbit

- **New Features**
  - Updated the Android app version to 0.64.78 (build 6478).

- **Bug Fixes**
  - Improved the Life section’s list handling to prevent duplicate loading and observers.
  - Preserved list reordering and visibility behavior while improving reliability when reopening the screen.


dogi:
**Review — rating: 84/100**

This is the same adapter-caching pattern as #15599, which I sent back earlier today — but scoped correctly, and the differences are what make it work.

The reason #15599 fails is that caching its adapter also freezes `isGuest` and `currentUserName`, which are immutable constructor `val`s the surrounding code re-reads on every call. `LifeAdapter` has no equivalent: its constructor takes a `Context`, the fragment as `OnStartDragListener`, and two lambdas. The lambdas resolve `requireContext()` *lazily* at invocation time rather than capturing it, so the toasts can't go stale, and there's no user- or permission-dependent state to drift. Nothing needs refreshing on re-entry, so the guard is safe here in a way it isn't there.

Two details are right that would be easy to get wrong:

- **`itemTouchHelper?.attachToRecyclerView(recyclerView)` stays outside the guard.** The `RecyclerView` is recreated on every view lifecycle, so re-attaching unconditionally is required — and `attachToRecyclerView` tears down its callbacks on the previous view first, so repeat calls don't stack.
- **`isObserverAttached` is reset in `onDestroyView`.** `collectWhenStarted` is scoped to `viewLifecycleOwner`, so the collection is cancelled when the view dies and *must* be re-established on the next `onCreateView`. A guard that didn't reset would leave the list permanently unpopulated after a back-stack pop. Moving `viewModel.loadMyLifeList()` inside the same guard is consistent — once per view lifecycle rather than once per `getAdapter()` call.

CI green (build default + lite, test default).

**Nits:**

- `lifeAdapter` is `lateinit` and never cleared, so it and the `Context` passed to its constructor outlive `onDestroyView`. For a back-stack pop within the same Activity that's harmless, and non-retained fragments are recreated alongside the Activity on configuration change — but it's worth a comment, since an adapter surviving the view lifecycle is the kind of thing a later reader will assume is a bug.
- `onDestroyView` doesn't call `itemTouchHelper?.attachToRecyclerView(null)`, so between view destruction and the next `getAdapter()` the helper still holds the dead `RecyclerView`. The next attach releases it, so this is a bounded window rather than a leak, but detaching explicitly is one line.
- The description should say *why* the guard is needed — i.e. that `getAdapter()` is invoked more than once per view lifecycle. If it isn't, `isObserverAttached` is dead weight; if it is, that's the actual bug being fixed and it deserves naming.

**Verdict:** mergeable.

---

Okuro3499 (approved):

Okuro3499 (approved):

PR #15622, opened 2026-08-13
life: smoother health examination dispatcher providing (fixes #15740)

### LINKED ISSUE #15740: Precompute HealthExaminationAdapter display data before binding
Extract conditions JSON parsing and string formatting out of `onBindViewHolder` into `HealthExaminationItem` wrapper mapped in `submitExaminations`. Improves frame rate during RecyclerView scroll.

dogi:
Extract conditions JSON parsing and string formatting out of `onBindViewHolder` into `HealthExaminationItem` wrapper mapped in `submitExaminations`. Improves frame rate during RecyclerView scroll.

---
*PR created automatically by Jules for task [17391694092291995245](https://jules.google.com/task/17391694092291995245) started by @dogi*

## Summary by CodeRabbit

* **Improvements**
  * Health examination lists now load and refresh more smoothly.
  * Examination dates and creator names are displayed consistently.
  * Self-examination and encrypted-data indicators are calculated reliably.
  * Examination details remain available when condition data contains malformed entries.
  * List updates now keep displayed information synchronized with the complete health record set.


dogi:
**Review — rating: 76/100**

The right idea, and it picks up two bug fixes along the way that the description doesn't claim.

Moving the work out of `onBindViewHolder` matters more here than in a typical precompute PR, because the per-bind work included `getEncryptedDataAsJson(user)` — actual decryption — plus a Gson parse of `conditions` and a date format, all re-run on every rebind and every scroll pass. Precomputing once per list is the correct shape, and wrapping it in a typed `HealthExaminationItem` rather than a parallel map is the right way to do it.

Two things improve correctness as a side effect, both worth keeping:

- **`showConditions` could crash.** The old version did `JsonUtils.gson.fromJson(realmExamination?.conditions, …)` and then `conditionsMap.keySet()` with no null guard, so a null or malformed `conditions` threw when the dialog opened. The new `try`/`catch` leaves the text empty instead.
- **The diff callback had a real gap.** `areContentsTheSame` never compared `conditions`, so a change to conditions alone wouldn't rebind. Comparing `conditionsText` closes that.

**Two things to fix before this lands.**

**Precomputation is eager and unconfined.** `submitExaminations` now decrypts *every* record in the list at once, on whatever thread the caller is on — and `MyHealthFragment` calls it straight from its collect block on the main thread. Previously decryption was lazy and bounded to visible rows. For a personal exam list that's probably tens of records and imperceptible, but it's decryption on the main thread and it scales with history length, so it belongs behind `withContext(dispatcherProvider.default)` or in the ViewModel that already produces `list`.

**Precomputed display names never refresh.** `displayDate` depends on both `userModel` and `userMap`, but `updateUserMap()` (line ~48) only reassigns those fields — it doesn't recompute the items or clear `displayNameCache`. Under the old code the lookup happened at bind time, so a `userMap` arriving after the list still produced correct names on the next rebind. Now, if `submitExaminations` runs before `updateUserMap`, every row is stuck with the `createdBy`-split fallback, *and* that fallback gets memoised into `displayNameCache` so a later correct lookup never happens. Please confirm the call ordering in `MyHealthFragment`, and have `updateUserMap` clear the cache and re-run the precompute over the current list.

**Nit:** `binding.txtDate.tag = formatDate(realmExamination.date, "MMM dd, yyyy")` still formats on every bind, which is the exact cost this PR set out to remove — the raw formatted date is already computed in `submitExaminations` as `formattedDate` and just isn't carried on the item. Adding a `rawDate` field finishes the job.

**Verdict:** needs another round — confine the eager decryption, and make `updateUserMap` invalidate the precomputed names.

---

dogi:
**Review — rating: 70/100**

The premise is sound and the biggest win is real. `getEncryptedDataAsJson` runs `AndroidDecrypter.decrypt(...)` followed by a Gson parse, and it was being called once per `onBindViewHolder` — i.e. per scroll frame, per row. Hoisting that into a precomputed `HealthExaminationItem` is the right fix, and it's the part of this PR I'd most want to keep.

Two latent bugs get fixed along the way, both worth noting because neither is mentioned in the description:

- **Stale position in the click listener.** `holder.itemView.setOnClickListener { showAlert(binding, position, encrypted) }` captured `position` and re-resolved it via `getItem(position)` when tapped. After any list change the captured index points at a different record — a user could open someone else's examination. Capturing the item instead removes the whole class.
- **NPE in `showConditions`.** `JsonUtils.gson.fromJson(realmExamination?.conditions, ...)` returns `null` when `conditions` is null or empty, and the very next line called `conditionsMap.keySet()` on it. The replacement null-checks and wraps in `try/catch`.

Dropping the `if (realmExamination != null)` guards is also correct — `getItem` returns a non-null item and `item.examination` is non-null, so they were dead.

**What's blocking:**

**1. Hard-coded dispatchers.** `CLAUDE.md` is explicit: inject `DispatcherProvider` rather than referencing `Dispatchers.*` directly, so tests can substitute deterministic dispatchers.

```kotlin
val items = withContext(Dispatchers.Default) { … }
withContext(Dispatchers.Main) { submitList(items) }
```

This adapter takes constructor params already, so threading a `DispatcherProvider` through is mechanical — and it's what makes the next point testable.

**2. Localized strings are frozen at submit time.** `displayDate` is built with `context.getString(R.string.two_strings, …)` / `R.string.self_examination` inside the mapping pass and stored on the item. Previously `onBindViewHolder` rebuilt them on every bind, so a locale change followed by a rebind picked up the new language. Now the cached string survives until something calls `submitExaminations` again. The app ships six locales and `MainApplication` manages locale switching, so this is reachable. Storing the *parts* (`formattedDate`, resolved `name`, `isSelfExamination`) and doing the cheap `getString` in `onBindViewHolder` keeps the expensive work hoisted without freezing the localization.

**3. Decrypted health data is now retained for the whole list.** `encryptedData: JsonObject?` holds the parsed plaintext of every examination for as long as the adapter's list is alive, where before it was produced per bind and dropped. For medical records that's a meaningful change in exposure window — a heap dump or background snapshot now contains every record rather than the handful on screen. If the dialog is the only consumer, decrypting lazily on tap and keeping only a `hasEncryptedData: Boolean` on the item gets the same UI behavior with the old exposure profile.

**4. No tests.** 175 lines rewritten across an adapter and a fragment, on the health path, with zero test coverage added. `submitExaminations` is now a pure-ish `suspend` mapping function — which is exactly the shape that's easy to test, once the dispatcher is injected. At minimum: the self-vs-other `displayDate` branch, the `displayNameCache` fallback chain, and the malformed-`conditions` path that used to NPE.

**Smaller points:**

- `conditionsText` is computed for every item at submit time, but conditions parsing was never in the bind path — it only ran when the dialog opened. So this adds work for every record while helping scroll performance not at all. Harmless off-thread, but it's the one field that didn't need hoisting.
- Indentation on `private val colorGrey50 by lazy { … }` is 8 spaces where the surrounding lines are 4.
- `DIFF_CALLBACK` compares `displayDate` *and* `examination.date`, `conditionsText` *and* the fields it derives from. Redundant rather than wrong — the derived fields alone would do.
- Moving `displayNameCache` from a field to a local is a quiet improvement: as a field it grew unbounded across the adapter's life.

**Verdict:** the hoisting is the right idea and the two incidental bug fixes are worth having. Inject `DispatcherProvider`, keep the `getString` calls in `onBindViewHolder` so locale changes still land, reconsider holding decrypted records for the list's lifetime, and add tests for the mapping function. That's a solid `ready` — it just isn't there yet.

---

dogi:
**Re-review — Rating: 76/100** (was 70)

Three of my four findings are fixed, and fixed well — the restructure into `formattedDate` + `resolvedName` + `hasEncryptedData` is exactly the right decomposition. Two new problems came in with it, one of them user-visible.

**What's fixed:**

- **`DispatcherProvider` injected**, replacing `Dispatchers.Default` / `Dispatchers.Main`.
- **Localization unfrozen.** `getString(R.string.two_strings, …)` and `getString(R.string.self_examination, …)` are back in `onBindViewHolder`, with only the locale-independent parts cached. A language change now lands on the next bind.
- **Decrypted records no longer retained.** `encryptedData: JsonObject?` → `hasEncryptedData: Boolean`, with decryption deferred to the tap. The exposure window is back to what it was, and the hoist that motivated the PR is intact — `getEncryptedDataAsJson` still runs once per item in `submitExaminations` and no longer per bind.
- **`conditionsText` moved back into `showAlert`**, so it's parsed on tap rather than for every record.

**Regression — records with no encrypted data are no longer clickable.**

```kotlin
hasEncryptedData = encrypted != null && encrypted.keySet().isNotEmpty()
```

`getEncryptedDataAsJson` returns `JsonObject()` — non-null but empty — whenever `data` is empty:

```kotlin
return if (!TextUtils.isEmpty(data)) JsonUtils.gson.fromJson(...) else JsonObject()
```

So on `master` the old guard `if (encrypted != null)` was true for **every** row whenever a user was logged in, and every row opened the dialog. The new `keySet().isNotEmpty()` check makes rows with empty `data` inert.

That matters more than "an empty notes section is useless", because the dialog isn't only notes — it shows the full vitals block (temperature, pulse, BP, height, weight, vision, hearing) *and* carries the neutral **Edit** button into `HealthExaminationActivity`. A record with no encrypted notes still has all of that, and is now unreachable and uneditable.

`hasEncryptedData = encrypted != null` restores the old behavior. If the intent was to gate on something, `userModel != null` is the condition that actually determines whether the dialog can render.

**Two fully-qualified names.** Both need an import:

```kotlin
// HealthExaminationAdapter constructor
private val dispatcherProvider: org.ole.planet.myplanet.utils.DispatcherProvider

// MyHealthFragment
@Inject lateinit var dispatcherProvider: org.ole.planet.myplanet.utils.DispatcherProvider
```

**Smaller points:**

- `rawDate` and `formattedDate` are assigned the same value and never diverge — `rawDate = formattedDate, formattedDate = formattedDate`. One of them can go; `rawDate` exists only to feed `binding.txtDate.tag`, which could read `formattedDate` directly.
- The `@Inject lateinit var dispatcherProvider` in `MyHealthFragment` is declared on one line while the neighbouring `@Inject lateinit var realtimeSyncManager` uses two. Minor, but they sit adjacent.
- Indentation on `private val colorGrey50 by lazy { … }` is still 8 spaces against the surrounding 4.
- Still no tests. `submitExaminations` is now a much better testing target than last round — pure mapping, injected dispatcher, no resource lookups inside it. The self-vs-other branch and the `displayNameCache` fallback chain are two cheap cases.

**Verdict:** real progress, and the architecture is now right. Restore the click guard, import the two `DispatcherProvider` references, and this goes to `ready` — the remaining items are nits.

---

dogi:

Every item from last round is fixed, including the nits.

**The click regression is resolved.** `hasEncryptedData = encrypted != null` restores the original semantics — `getEncryptedDataAsJson` returns a non-null empty `JsonObject` when `data` is empty, so every row is clickable again whenever a user is logged in. Records with no encrypted notes can reach their vitals and the Edit button as before.

**Both FQNs imported**, `rawDate` removed now that it was provably identical to `formattedDate`, the `@Inject lateinit var` reformatted to match its neighbour, and the stray 8-space indent on `colorGrey50` corrected.

Taken with the previous round, the whole shape is now right: the expensive `AndroidDecrypter.decrypt` + Gson parse runs once per item in `submitExaminations` rather than per bind; localized strings are built in `onBindViewHolder` so a locale change lands; decrypted content isn't retained on the item; and `conditionsText` is parsed on tap rather than for every record. That's the version I'd merge.

**Two misordered imports** — both new lines landed at the end of the `utils` group instead of alphabetically:

- `HealthExaminationAdapter.kt` — `DispatcherProvider` sits after `TimeUtils.formatDate`; it belongs right after `DiffUtils`.
- `MyHealthFragment.kt` — `DispatcherProvider` sits after `TimeUtils`; it belongs before `ImageUtils`.

Nits, not blockers, but worth folding into the next push since you'll likely have one.

**Still no tests**, which is the one thing that's been open across all three rounds. Not holding the PR for it — I've let comparable changes through — but `submitExaminations` is now about as testable as this code will ever get: pure mapping, injected dispatcher, no resource lookups inside it. The self-vs-other branch and the `displayNameCache` fallback chain (`getFullName()` → colon-split → raw id) are two cheap cases, and the fallback chain is the kind of thing that breaks silently.

**CI note:** all three checks were still `in_progress` when I reviewed, so this rating is on the diff rather than on a green run. The delta since the last green head is two imports, one boolean, and a removed field, so I don't expect trouble — but if anything goes red, that's where to look.

**Verdict:** `ready`.

---

Okuro3499 (approved):

PR #15623, opened 2026-08-13
teams: smoother voices repository view modelling (fixes #15741)

### LINKED ISSUE #15741: Refactor VoicesRepository to remove cross-feature lookups
Refactor VoicesRepository to remove cross-feature lookups

Removed `getUserById` and `getLibraryResource` from `VoicesRepository` and its implementation to maintain strict separation of concerns, eliminating cross-feature dependencies within the repository layer. The lookups have been correctly rerouted via dependency injection of `UserRepository` and `ResourcesRepository` directly into `VoicesViewModel` and `TeamsVoicesViewModel`. The corresponding test suites were updated to reflect these dependency changes.

dogi:
Refactor VoicesRepository to remove cross-feature lookups

Removed `getUserById` and `getLibraryResource` from `VoicesRepository` and its implementation to maintain strict separation of concerns, eliminating cross-feature dependencies within the repository layer. The lookups have been correctly rerouted via dependency injection of `UserRepository` and `ResourcesRepository` directly into `VoicesViewModel` and `TeamsVoicesViewModel`. The corresponding test suites were updated to reflect these dependency changes.

---
*PR created automatically by Jules for task [7599419300872196835](https://jules.google.com/task/7599419300872196835) started by @dogi*


dogi:
**Review — rating: 88/100**

Correct layering fix. `VoicesRepository` was exposing `getUserById` and `getLibraryResource`, neither of which is voices data — the first was a straight proxy to `UserRepository`, the second a lone `MyLibraryDao` call. Having the two ViewModels inject `UserRepository` and `ResourcesRepository` directly is the right shape, and it lets `VoicesRepositoryImpl` drop its `dagger.Lazy<UserRepository>`. That `Lazy` was there to defer a dependency cycle; removing the dependency eliminates the cycle instead of postponing it, which is the better outcome.

I checked the substitution that could have silently changed behaviour. `getLibraryResource` was `myLibraryDao.getByUnderscoreId(resourceId)` — matching `_id` only — while the replacement `ResourcesRepositoryImpl.getLibraryItemByResourceId` (line 164) tries `getByResourceId` first and *falls back* to `getByUnderscoreId`. So it's a strict superset: every row the old call could find is still found. `MyLibraryDao` has three same-shaped single-row lookups keyed on `id`, `resourceId` and `_id`, so landing on the wrong one was the live risk here, and this doesn't. The only behavioural difference is precedence — if some row's `resourceId` equalled a *different* row's `_id`, the new call would prefer the former. That's contrived for CouchDB-sourced ids, and consolidating on the repository's canonical lookup (already used at lines 297 and 498) is worth it.

Test changes are right too: deleting `getUserById delegates to userRepository` is correct rather than a coverage loss, since the method it asserted on no longer exists, and the constructor updates are mechanical. CI green (build default + lite, test default).

**Nit:** both ViewModel test files declare the new mocks with inline fully-qualified names — `private val userRepository: org.ole.planet.myplanet.repository.UserRepository = mockk(relaxed = true)` — while every other dependency in those files is imported. Worth adding the imports for consistency.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #15626, opened 2026-08-13
courses: smoother steps filter coroutine scoping (fixes #15795)

### LINKED ISSUE #15795: Refactor CoroutineScopes and adapter IO in courses UI
Fixes hand-rolled `CoroutineScope`s and adapter IO in the courses UI by moving away from hardcoded scopes and default DispatcherProviders. This allows properly scoped structural concurrency and cleaner dependency injection.

dogi:
Fixes hand-rolled `CoroutineScope`s and adapter IO in the courses UI by moving away from hardcoded scopes and default DispatcherProviders. This allows properly scoped structural concurrency and cleaner dependency injection.

---
*PR created automatically by Jules for task [13942332907888503601](https://jules.google.com/task/13942332907888503601) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6528
-        versionName = "0.65.28"
+        versionCode = 6532
+        versionName = "0.65.32"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6532
-        versionName = "0.65.32"
+        versionCode = 6533
+        versionName = "0.65.33"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/base/BaseDashboardFragment.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/base/BaseDashboardFragment.kt
@@ -13,9 +13,7 @@ import android.widget.TextView
 import androidx.appcompat.app.AlertDialog
 import androidx.core.content.ContextCompat
 import androidx.fragment.app.viewModels
-import androidx.lifecycle.Lifecycle
 import androidx.lifecycle.lifecycleScope
-import androidx.lifecycle.repeatOnLifecycle
 import androidx.recyclerview.widget.LinearLayoutManager
 import com.google.android.flexbox.FlexDirection
 import com.google.android.flexbox.FlexboxLayout
@@ -69,7 +67,6 @@ open class BaseDashboardFragment : DashboardPluginFragment() {
     }
     private var di: DialogUtils.CustomProgressDialog? = null
 
-
     @Inject
     lateinit var lifeRepository: LifeRepository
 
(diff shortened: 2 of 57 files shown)


dogi:
**Review — rating: 72/100**

Three of the four changes here are right, and one is the exact opposite of what the PR is trying to do.

**The good parts.** Having `CourseFilterController` take a `CoroutineScope` instead of building its own from `DefaultDispatcherProvider()` is correct — `CoursesFragment` passes `viewLifecycleOwner.lifecycleScope`, so the debounce job now dies with the view instead of relying on someone remembering to call `clear()`. Extracting `ResourcePreviewHelper` is a real improvement: four inline `withContext(io)` bodies doing `PdfRenderer`, `MediaMetadataRetriever`, OpenCSV and buffered-reader work had no business inside an adapter, and they're now isolated behind a class that takes an injected `DispatcherProvider`. And removing the `= DefaultDispatcherProvider()` default parameters from `ANRWatchdog` and `CameraUtils.capturePhoto`/`savePicture` is exactly right — a default that silently constructs the real provider means callers never notice they aren't injecting one.

**Which makes this the odd one out:**

```kotlin
private var adapterScope = CoroutineScope(SupervisorJob() + kotlinx.coroutines.Dispatchers.Main)
```

`InlineResourceAdapter` loses `dispatcherProvider` from its constructor, and its two scope constructions switch from `dispatcherProvider.main` to a hard-coded, fully-qualified `Dispatchers.Main`. That's the pattern `CLAUDE.md` calls out directly — "Inject `DispatcherProvider` (don't hard-code `Dispatchers.*`) so tests can substitute deterministic dispatchers" — and it's the same practice this PR removes from three other files in the same diff. It also makes the adapter untestable with `TestDispatcherProvider`: there's no longer any way for a test to control the dispatcher its coroutines run on.

The adapter still needs a `DispatcherProvider` for its scope even though the IO moved to the helper. Keep it on the constructor alongside `previewHelper`, or have the helper expose the provider — either way, don't reach for `Dispatchers.Main` directly.

**`clear()` is now dead.** It's an empty method whose entire body is `// Scope cancellation is now handled by the owner`. If callers remain, they're calling nothing; if none remain, it should be deleted. A no-op method with an explanatory comment is worse than either — it reads as a lifecycle hook that still does something. Delete it and its call sites.

**No tests, which is the disappointing part.** `ResourcePreviewHelper` was extracted precisely so this logic becomes testable, and all four methods take a `File` and return a value — `getCsvPreview` (5-row cap), `getTextPreview` (8-line cap) and `getAudioPreview`'s duration formatting are all straightforward to cover with temp files and a `TestDispatcherProvider`. Extracting for testability and then not testing leaves only the refactor.

**Nit:** `String.format("%d:%02d", …)` in `getAudioPreview` has no `Locale` argument, so it uses the default locale — in locales with non-ASCII digits that produces unexpected output. Carried over rather than introduced, but it's landing in a brand-new file where lint will flag it.

**Verdict:** needs another round — put `DispatcherProvider` back on `InlineResourceAdapter` instead of hard-coding `Dispatchers.Main`, and delete the no-op `clear()`.

---

dogi:

CI green on all three checks. Good extraction — `InlineResourceAdapter` was doing three jobs and now does one.

Pulling PDF rendering, audio duration extraction, and CSV preview parsing into `ResourcePreviewHelper` takes `PdfRenderer`, `ParcelFileDescriptor`, `MediaMetadataRetriever`, `FileReader` and the OpenCSV builders out of an adapter that has no business knowing about them. The adapter keeps what it should — the `LruCache` bookkeeping and view binding — and the call sites read as `previewHelper.getPdfPreview(file)?.also { bitmapCache.put(cacheKey, it) }`, which is exactly the right division.

**The 63-line `ResourcePreviewHelperTest` is the proof this was worth doing.** None of that media-decoding logic was reachable from a test while it was buried in adapter bind methods; now it is. That's the concrete payoff, more than the line count.

The caching stays in the adapter and the IO stays in the helper, so the fast path still short-circuits before any dispatcher hop — worth noting, because moving the cache lookups into the helper too would have made every cached bind pay a `withContext`.

**Minor:** `ANRWatchdog.kt` (2 lines) and `CameraUtils.kt` (4 lines) are touched with no obvious connection to resource previews. If those are incidental import fixes they're harmless, but they're unexplained in a PR about the courses UI — worth a word in the description or splitting out.

**Verdict:** clean separation, newly testable, tests included. Marking `ready`.

---

Okuro3499 (approved):

PR #15627, opened 2026-08-13
enterprises: smoother glide request managing (fixes #15774)

### LINKED ISSUE #15774: Add Glide `onViewRecycled` clears + lifecycle-safe request managers
- Added `onViewRecycled` in `CoursesAdapter`, `EnterprisesFinancesAdapter`, and `EnterprisesReportsAdapter` to clear Glide requests and null click listeners.
- Changed `Glide.with(this)` to `Glide.with(requireView())` in `EnterprisesFinancesFragment`, `EnterprisesReportsFragment`, and `ResourceViewerFragment` for proper view lifecycle binding.

dogi:
- Added `onViewRecycled` in `CoursesAdapter`, `EnterprisesFinancesAdapter`, and `EnterprisesReportsAdapter` to clear Glide requests and null click listeners.
- Changed `Glide.with(this)` to `Glide.with(requireView())` in `EnterprisesFinancesFragment`, `EnterprisesReportsFragment`, and `ResourceViewerFragment` for proper view lifecycle binding.

---
*PR created automatically by Jules for task [2560109261817049029](https://jules.google.com/task/2560109261817049029) started by @dogi*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved scrolling performance and reliability in course, finance, and report lists.
  * Prevented outdated images and click actions from appearing when list items are reused.
  * Reduced the risk of stale view references and unintended interactions in recycled items.


dogi:
**Review — rating: 70/100**

**The `onViewRecycled` half is right.** `Glide.with(context).clear(imageView)` on recycle cancels the in-flight request and releases the target, which is the standard fix for the "wrong thumbnail flashes on a recycled row" class of bug — the same failure mode I flagged on #15579. Adding it to `CoursesAdapter`, `EnterprisesFinancesAdapter` and `EnterprisesReportsAdapter` is worth doing. Nulling the click listeners is harmless since `onBindViewHolder` re-sets them, and it drops a lambda reference while the holder sits in the pool.

**The `Glide.with(this)` → `Glide.with(requireView())` half looks backwards.** In a Fragment, `Glide.with(this)` binds the `RequestManager` directly to that Fragment's lifecycle — the most precise binding available. `Glide.with(View)` doesn't improve on it: it resolves the manager by walking up from the view to find the containing Fragment, and *falls back to the Activity* when it can't. That's the coarser lifecycle, not the finer one, and Glide's own guidance is to prefer the Fragment overload when you're in a Fragment and reserve the View overload for code that only has a View. So the PR title claims lifecycle-safety that this half arguably reduces.

It also adds a failure mode that wasn't there. `requireView()` throws `IllegalStateException` outside the view lifecycle, where `Glide.with(this)` doesn't. Two of the five call sites are inside `registerForActivityResult` callbacks (`EnterprisesFinancesFragment:73`, `EnterprisesReportsFragment:105`) — an image-picker result arriving after the view is gone now crashes instead of no-op'ing. Result callbacks are delivered at STARTED so a view normally exists, but "normally" is doing real work in that sentence, and the old form didn't depend on it at all.

If there's a concrete leak these five sites were causing, please say what it was — I can't see one, and `with(Fragment)` is the stronger binding. Otherwise I'd revert that half and keep the recycling clears, which stand on their own.

**Nit:** `EnterprisesReportsAdapter` gains an `onViewRecycled` here while #15618 gives the same class a payload-aware `onBindViewHolder`. Both touch that file; whichever lands second needs a rebase.

**Verdict:** needs another round — split the `onViewRecycled` clears out and land them; justify or drop the `requireView()` switch.

---

dogi:
**Re-review — Rating: 80/100**

CI green on all three checks. The `onViewRecycled` clears are correct and worth having — a recycled holder no longer shows the previous row's image while the new load is in flight, which is the visible symptom this fixes.

**The title over-promises relative to the diff.** "lifecycle-safe request managers" implies a change to how `RequestManager`s are obtained, but every call is `Glide.with(context)` against the adapter's stored `Context` — the same acquisition as before. Only the clearing is new. Either the lifecycle half was descoped and the title didn't follow, or it's still to come; worth reconciling so reviewers aren't looking for a change that isn't there.

**That's also where the one real risk sits.** `CoursesAdapter.kt:40` holds `private val context: Context`, and adapters here are constructed with an Activity. `RequestManagerRetriever.get(Activity)` calls `assertNotDestroyed` and throws `IllegalArgumentException` — that check runs when you *obtain* the manager, before `clear()` is ever reached. `onViewRecycled` can fire during teardown (pool clearing, `adapter = null`), so on a destroyed Activity this path can throw rather than quietly no-op.

`Glide.with(holder.itemView).clear(...)` is the idiom for exactly this reason — it resolves from the view and degrades to the application-scoped manager when there's no live host. I'd switch the three call sites; it costs nothing and removes the question.

**The listener clearing is defensive noise.** `onBindViewHolder` assigns every listener on every bind, so nulling them on recycle doesn't prevent a stale handler from firing — the view isn't visible between recycle and rebind. It isn't harmful, just three lines that imply a leak that isn't there. If there *is* a specific leak motivating it, the PR description is the place to say so.

**Verdict:** the Glide clears are a genuine fix and the reason to land this. Switch `Glide.with(context)` to `Glide.with(holder.itemView)` and either drop the listener nulling or justify it. Marking `ready` on the strength of the core change — the `Glide.with` swap is small enough to fold in, but I'd rather it went in before merge than after.

---

Okuro3499 (dismissed):
```
CoursesAdapter.onViewRecycled (lines 238–249) introduces a confirmed regression: it nulls holder.itemView's click listener, but that listener is only ever set once, in            
  GridViewHolder/ListViewHolder's init {} block (lines 364, 430) — never in bind() or bindPayloads(). Once any course row is recycled and rebound (which happens on ordinary         
  scrolling in any list longer than one screen), tapping that item permanently does nothing. This breaks the core "open a course" interaction for the CoursesAdapter, which is a     
  worse regression than the memory-leak/phantom-click issue the change is meant to fix.
```

dogi:
**Review — rating: 86/100**

Clean and correct. Three adapters gain `onViewRecycled` overrides that release what a recycled holder shouldn't keep.

**The Glide clears are the substantive part.** Without `Glide.with(context).clear(view)`, an in-flight load started for row 5 can complete after that holder has been rebound to row 40 and briefly paint the wrong image before the new load lands — the classic recycled-holder flash. Clearing on recycle cancels the request and drops Glide's reference to the target, which also lets the bitmap be reclaimed instead of being pinned by a request that no longer matters.

`onViewRecycled` runs on the main thread, which is what `Glide.clear` requires, so the placement is right.

**Nulling the listeners is worth having too.** `financeImage`, `reportImage`, `edit`, and `delete` all get their listeners re-attached in `onBindViewHolder`, so clearing them on recycle can't drop a live handler — it just guarantees a holder sitting in the pool isn't holding a lambda that captures the previous row's `Transaction` or `MyTeam`. Cheap, and it makes the capture lifetime obvious.

The `when (holder)` dispatch in `CoursesAdapter` correctly handles both `GridViewHolder` and `ListViewHolder`, matching how that adapter binds two view types.

**Minor observations, neither blocking:**

- `CoursesAdapter.onViewRecycled` has identical bodies in both `when` branches. Since both view holders expose `binding.ivCover`, an `or`-style branch or a small helper would say "same handling for both" more directly — cosmetic.
- No tests. That's proportionate here: `onViewRecycled` behavior is genuinely awkward to assert without an attached RecyclerView driving recycling, and the change is small and mechanical. Worth noting only because `EnterprisesReportsAdapterTest` already exists (added by #15618) and instantiates holders directly, so a `Glide.clear` verification would have somewhere to live if you wanted one.
- The overlap with **#15618** is real but harmless — that PR adds `onBindViewHolder(holder, position, payloads)` to `EnterprisesReportsAdapter` while this adds `onViewRecycled` to the same class. Different methods, adjacent lines; expect a trivial conflict depending on merge order.

**Verdict:** `ready`. Small, focused, fixes a real recycling artifact, and CI is green on all three checks.

---

Okuro3499 (approved):

PR #15628, opened 2026-08-13
dashboard: smoother bell view modelling (fixes #15744)

### LINKED ISSUE #15744: Refactor BellDashboardFragment survey-reminder logic into ViewModel
Moved data-fetching and repository dependencies (SurveysRepository,
SubmissionsRepository, UserRepository, CoursesRepository, TimeProvider)
out of BellDashboardFragment and into BellDashboardViewModel.

Introduced a SurveyPrompt StateFlow to drive the survey reminder UI,
preventing cross-feature repository leaks within the dashboard fragment.

dogi:
Moved data-fetching and repository dependencies (SurveysRepository,
SubmissionsRepository, UserRepository, CoursesRepository, TimeProvider)
out of BellDashboardFragment and into BellDashboardViewModel.

Introduced a SurveyPrompt StateFlow to drive the survey reminder UI,
preventing cross-feature repository leaks within the dashboard fragment.

---
*PR created automatically by Jules for task [15372804692671881163](https://jules.google.com/task/15372804692671881163) started by @dogi*

## Summary by CodeRabbit

* **New Features**
  * Survey prompts now support scheduled reminders with reminder-specific messaging.
  * Multiple due survey reminders can be presented together.
  * Survey dialogs can be dismissed with a neutral action while preserving reminder scheduling options.

* **Bug Fixes**
  * Reduced duplicate survey prompts by preventing recently displayed or already scheduled surveys from reappearing too soon.

* **Tests**
  * Added coverage for survey prompt timing, reminder suppression, and multiple reminder groups.


dogi:
**Review — rating: 70/100**

The direction is right — survey-reminder scheduling, the one-hour interval guard, reminder-group parsing and pending-status filtering are all business logic that had no business living in a Fragment. Moving them behind `BellDashboardViewModel`, and giving the fragment `getUserModel()`/`isCourseCertified()` pass-throughs so it stops injecting four repositories, is the correct shape. Unifying the two dialog paths behind one `SurveyPrompt` with an `isReminder` flag also removes a genuine duplication — `showPendingSurveysReminder` and `checkPendingSurveys` were building the same dialog two different ways.

**But `StateFlow` is the wrong carrier for a one-shot UI event, and it costs a real behaviour change.**

`handleDueReminders` loops over reminder groups and assigns inside the loop:

```kotlin
for (surveyIds in remindersToShow) {
    …
    if (pendingSurveys.isNotEmpty()) {
        _surveyPrompt.value = SurveyPrompt(pendingSurveys, surveyTitles, isReminder = true)
    }
}
```

`StateFlow` conflates, so when more than one reminder group is due only the **last** one ever reaches the collector — the earlier groups are silently dropped. The code this replaces called `showPendingSurveysReminder(pendingSurveys)` once per group, so multiple due reminders each got a dialog. That's a straightforward regression, and it's invisible until a user has two reminders fire together.

The same choice creates a second problem: `StateFlow` replays its current value to new collectors. `collectWhenStarted` re-subscribes on every STARTED transition, so a rotation or a trip through recents between `showSurveyListDialog(...)` and `clearSurveyPrompt()` re-delivers the same prompt and shows the dialog again. The `clear` call narrows the window but doesn't close it.

Both go away with `MutableSharedFlow(replay = 0)` and `emit()`, which queues each event and replays none — and there's precedent in this codebase already (`RequestsViewModel._successAction`, `RealtimeSyncManager.dataUpdateFlow`). `clearSurveyPrompt()` then disappears too.

**Also worth noting:** `dueRemindersFlow()` moves from `collectWhenStarted` in the fragment to a plain `viewModelScope.launch` in `init`, so it now collects while the fragment is stopped rather than only while STARTED. That's defensible for a ViewModel, but it interacts badly with conflation — prompts computed in the background pile up and all but the newest are lost.

**No tests.** This is the part that stands out: the whole point of moving this logic is that it becomes testable, and the interval guard, the `isReminderScheduled` short-circuit, the comma-splitting and the `status == "pending"` filter are all now reachable from a plain unit test with `TestDispatcherProvider` and a fake `TimeProvider` — which this repo already has. The multi-group case above would have been caught by one.

**Nits:** check whether `userRepository`, `surveysRepository`, `submissionsRepository`, `coursesRepository` and `timeProvider` are still referenced in the fragment — several look dead now. The new imports in the ViewModel are inserted out of order (a `java.util` import lands mid-`org.ole` block), and the diff leaves stray double blank lines where the companion object was removed and after `isCourseCertified`.

**Verdict:** needs another round — switch `_surveyPrompt` to a replay-0 `SharedFlow` so multiple due reminders aren't dropped, and add tests for the logic you've just made testable.

---

dogi:
**Review — rating: 86/100**

CI green on all three checks. Proper ViewModel extraction rather than relocation — the same shape as #15694, and the distinction matters.

`checkPendingSurveys`, `handleDueReminders` and `scheduleSurveyReminder` move into `BellDashboardViewModel` along with the `SURVEY_DIALOG_INTERVAL_MS` constant, so the fragment stops injecting `surveysRepository`, `submissionsRepository`, `userRepository` and `timeProvider`. The reminder logic — interval throttling, dedupe against already-scheduled ids, mapping submissions to titles — is genuine business logic that had no business living in a fragment, and it was untestable there.

**The 141 lines of `BellDashboardViewModelTest` are the proof.** That logic has several branches worth pinning: the one-hour throttle, the `isReminderScheduled` short-circuit, the singular/plural title, and the empty-list early return. None of it was reachable from a test before.

The fragment drops 86 lines and keeps only what it should — dialog presentation and the `collectWhenStarted` wiring.

**Minor:** `viewModel.checkPendingSurveys(user?.id)` passes the user id in from the fragment, which still holds `user` from `viewModel.getUserModel()`. Since the ViewModel can fetch the user itself, having it own that too would let the fragment stop tracking `user` for this purpose — small, and arguably clearer as an explicit parameter. Not blocking either way.

**Heads-up on overlap:** #15635 is sweeping `repeatOnLifecycle` boilerplate to `collectWhenStarted` across ten fragments and #15648 is reworking `TeamsTasksFragment`; this PR is confined to `BellDashboardFragment`, so no direct conflict, but the dashboard area is busy right now.

**Verdict:** correct extraction, real test coverage for logic that had none, fragment genuinely shrinks. Marking `ready`.

---

Okuro3499 (approved):

PR #15629, opened 2026-08-13
teams: smoother submissions repositories streamlining (fixes #15796)

### LINKED ISSUE #15796: Refactor repository interfaces to remove Android Context dependency
Removes the `Context` parameter from repository interface serialization methods by injecting `@ApplicationContext` into the implementations instead. This makes the repositories more testable and decoupled from the Android framework. also cleans up unused upload serializer config variants.

dogi:
Removes the `Context` parameter from repository interface serialization methods by injecting `@ApplicationContext` into the implementations instead. This makes the repositories more testable and decoupled from the Android framework. also cleans up unused upload serializer config variants.

---
*PR created automatically by Jules for task [15862076471608288046](https://jules.google.com/task/15862076471608288046) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6529
-        versionName = "0.65.29"
+        versionCode = 6530
+        versionName = "0.65.30"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/model/ApkLog.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/model/ApkLog.kt
@@ -27,7 +27,7 @@ open class ApkLog {
     companion object {
         const val ERROR_TYPE_CRASH = "crash"
 
-        fun serialize(log: ApkLog): JsonObject {
+        fun serialize(log: ApkLog, customDeviceName: String): JsonObject {
             val `object` = JsonObject()
             `object`.addProperty("type", log.type)
             `object`.addProperty("error", log.error)
@@ -38,7 +38,7 @@ open class ApkLog {
             `object`.addProperty("createdOn", log.createdOn)
             `object`.addProperty("androidId", NetworkUtils.getUniqueIdentifier())
             `object`.addProperty("deviceName", NetworkUtils.getDeviceName())
-            `object`.addProperty("customDeviceName", NetworkUtils.getCustomDeviceName(org.ole.planet.myplanet.MainApplication.context))
+            `object`.addProperty("customDeviceName", customDeviceName)
             `object`.addProperty("parentCode", log.parentCode)
             return `object`
         }
--- a/app/src/main/java/org/ole/planet/myplanet/model/ApkLog.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/model/ApkLog.kt
@@ -1,6 +1,5 @@
 package org.ole.planet.myplanet.model
 
-import android.content.Context
 import androidx.room.Entity
 import androidx.room.PrimaryKey
 import com.google.gson.JsonObject
@@ -28,7 +27,7 @@ open class ApkLog {
     companion object {
         const val ERROR_TYPE_CRASH = "crash"
 
-        fun serialize(log: ApkLog, context: Context): JsonObject {
+        fun serialize(log: ApkLog): JsonObject {
             val `object` = JsonObject()
             `object`.addProperty("type", log.type)
             `object`.addProperty("error", log.error)
@@ -39,7 +38,7 @@ open class ApkLog {
             `object`.addProperty("createdOn", log.createdOn)
             `object`.addProperty("androidId", NetworkUtils.getUniqueIdentifier())
             `object`.addProperty("deviceName", NetworkUtils.getDeviceName())
-            `object`.addProperty("customDeviceName", NetworkUtils.getCustomDeviceName(context))
+            `object`.addProperty("customDeviceName", NetworkUtils.getCustomDeviceName(org.ole.planet.myplanet.MainApplication.context))
             `object`.addProperty("parentCode", log.parentCode)
             return `object`
         }
--- a/app/src/main/java/org/ole/planet/myplanet/repository/SubmissionsRepository.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/SubmissionsRepository.kt
@@ -57,7 +57,7 @@ interface SubmissionsRepository {
     suspend fun bulkInsertFromSync(jsonArray: JsonArray)
     suspend fun insertSubmission(submission: JsonObject)
     suspend fun getExamUploadPayload(submission: Submission): JsonObject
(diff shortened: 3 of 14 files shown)


dogi:
**Review — rating: 74/100**

Most of this is the right change. Collapsing `UploadSerializer` from four variants to two is a genuine simplification — `WithContext` and `AsyncContext` existed only to thread a `Context` to serializers, and removing them shrinks three `when` blocks in `UploadCoordinator` and the config surface with them. `serializeSubmission` losing its `context` parameter is a clean win: the build only stays green if the body never used it, so that was dead weight in a public interface method. And giving `TeamsRepositoryImpl` an injected `@ApplicationContext` is exactly how a repository should obtain a Context — once at construction, not per call.

**The `ApkLog` change goes the wrong way, though.** It doesn't remove the Context dependency, it conceals it:

```kotlin
`object`.addProperty("customDeviceName", NetworkUtils.getCustomDeviceName(org.ole.planet.myplanet.MainApplication.context))
```

The parameter is gone from the signature but the requirement isn't — it's now a static reach into the Application singleton. That's worse than what it replaced on three counts. The signature no longer tells the truth about what the function needs; `ApkLog.serialize` becomes untestable without mocking a static, where previously a test could just pass a Context; and it introduces a hard dependency from a model class onto the Application object. The inline fully-qualified reference rather than an import reads like it was uncomfortable to write, which is usually the tell.

`ApkLog` is a companion-object serializer so it genuinely can't take constructor injection — but it doesn't need a `Context`, it needs one *string*. The caller is `UploadConfigs`, which is already `@Inject`-constructed and can take `@ApplicationContext` like `TeamsRepositoryImpl` now does:

```kotlin
serializer = UploadSerializer.Simple { log -> ApkLog.serialize(log, NetworkUtils.getCustomDeviceName(context)) }
```

That keeps `UploadSerializer.Simple`, keeps the Context at the injectable layer, and leaves `ApkLog.serialize` a pure function of its inputs — which is what "remove the Context dependency" should mean.

**Nits:** several `import android.content.Context` lines are likely dead now — `SubmissionsRepository.kt` and `TeamsSyncRepository.kt` no longer name the type in any signature, and `UploadConfig.kt` loses its only two uses with the deleted variants. Worth checking each. Also, `TeamsRepositoryImpl`'s new `context` is added as the *first* positional constructor parameter, which is why three test files needed edits; adding it last would have left them untouched.

**Verdict:** needs another round — pass the device name into `ApkLog.serialize` rather than reaching for `MainApplication.context`. Everything else here is good.

---

dogi:

CI green on all three checks. The core refactor is right and the payoff is visible: dropping `Context` from `SubmissionsRepository.serializeSubmission`, `TeamsSyncRepository.serializeTeamActivities`, and `ApkLog.serialize` lets both `UploadSerializer.WithContext` and `AsyncContext` disappear, which collapses three `when` blocks in `UploadCoordinator` from four branches to two. That's the good kind of deletion — the abstraction shrinks because the need for it is gone, not because it was papered over.

Passing `customDeviceName: String` into `ApkLog.serialize` instead of a `Context` is the right call too: the function needed one string, not an Android handle.

**Fully-qualified names — please import and use short names**

Three sites, and this codebase should have none:

```kotlin
// TeamsRepositoryImpl
@dagger.hilt.android.qualifiers.ApplicationContext private val context: Context,

// UploadConfigs — FQN annotation *and* FQN type
@dagger.hilt.android.qualifiers.ApplicationContext private val context: android.content.Context,

// UploadConfigs
serializer = UploadSerializer.Simple { log -> ApkLog.serialize(log, org.ole.planet.myplanet.utils.NetworkUtils.getCustomDeviceName(context)) },
```

`TeamsRepositoryImpl` already imports `Context`, so that one is purely the annotation. `UploadConfigs` needs `import dagger.hilt.android.qualifiers.ApplicationContext`, `import android.content.Context`, and `import org.ole.planet.myplanet.utils.NetworkUtils` — after which those lines read as `@ApplicationContext private val context: Context` and `NetworkUtils.getCustomDeviceName(context)`. Every other injection point in the DI layer already uses the short form, so these stand out.

**Scope worth stating plainly.** The title says "repository interfaces", and that's accurate — but `TeamsRepositoryImpl` now takes `@ApplicationContext context` as a constructor field, so the *implementation* still depends on `Context`; it moved from a per-call parameter to a held reference. That's a defensible design (Android-free interface, impl injects what it needs), and I don't think it should change. It's just worth a line in the description so nobody reads this as "repositories no longer touch Context".

**Conflict with #15649, which is currently `ready`.** That PR rewrites `UploadConfigs.kt` and still uses `UploadSerializer.WithContext(ApkLog::serialize)` and `UploadSerializer.WithContext { log, context -> … }` — the two constructs this PR deletes. They cannot both merge as-is. #15649 is the smaller rebase of the two, so landing it first and rebasing this one onto it is the cheaper order; the reverse means #15649 has to be substantially rewritten.

**Verdict:** genuine layering improvement and the deletion of two serializer variants is the right outcome. Fix the three FQN sites, note the impl-holds-Context detail in the description, and sequence against #15649.

---

dogi:

CI green on all three checks. All three FQN sites are fixed:

```kotlin
import dagger.hilt.android.qualifiers.ApplicationContext
import android.content.Context
import org.ole.planet.myplanet.utils.NetworkUtils
…
@ApplicationContext private val context: Context,
serializer = UploadSerializer.Simple { log -> ApkLog.serialize(log, NetworkUtils.getCustomDeviceName(context)) },
```

The substance was already right and is unchanged: dropping `Context` from `SubmissionsRepository.serializeSubmission`, `TeamsSyncRepository.serializeTeamActivities` and `ApkLog.serialize` lets both `UploadSerializer.WithContext` and `AsyncContext` be deleted, which collapses three `when` blocks in `UploadCoordinator` from four branches to two. Passing `customDeviceName: String` instead of a `Context` is the right granularity — the function needed one string.

**Sequencing still stands:** #15649 (currently `ready`) rewrites `UploadConfigs.kt` and still uses `UploadSerializer.WithContext(ApkLog::serialize)` and `UploadSerializer.WithContext { log, context -> … }`, both of which this PR deletes. They can't both merge as written. #15649 is the smaller rebase, so landing it first and rebasing this onto it remains the cheaper order.

Worth still adding a line to the description noting that `TeamsRepositoryImpl` now holds `@ApplicationContext context` as a field — the *interfaces* are Android-free, which is what the title claims, but the implementation still depends on `Context`. That's a fine design; it just shouldn't read as "repositories no longer touch Context".

**Verdict:** clean now. Marking `ready`, to land after #15649.

---

Okuro3499 (changes requested):
```
UploadSerializer no longer needs two parallel variants (Simple/WithContext, Async/AsyncContext) just to thread a Context through, which simplifies             
  UploadCoordinator's when dispatch (UploadCoordinator.kt:112-129, 352-357). Not solving an urgent bug — it's a structural simplification, and a reasonable one since the removed    
  variants were pure boilerplate. However, the branch doesn't fully deliver on its own title: Context isn't removed from the repository layer, it's just pushed up a level —         
  TeamsRepositoryImpl and UploadConfigs now inject @ApplicationContext directly (TeamsRepositoryImpl.kt:61, UploadConfigs.kt:28) where before they only needed it transiently at call
  time. Net Context-dependency count in this slice of the codebase is roughly unchanged.
```

Okuro3499 (approved):

PR #15630, opened 2026-08-13
all: smoother user repository dao save searching (fixes #15745)

### LINKED ISSUE #15745: Refactor: Inject UserRepository into Health and Submissions repositories
Replaces direct `UserDao` injections in `HealthRepositoryImpl` and `SubmissionsRepositoryImpl` with `dagger.Lazy<UserRepository>`. This routes user operations through the domain boundary and prevents cross-domain write risks. Adds `searchUsers` and `saveUser` APIs to `UserRepository` to support these delegations. Test mocks were properly migrated.

dogi:
Replaces direct `UserDao` injections in `HealthRepositoryImpl` and `SubmissionsRepositoryImpl` with `dagger.Lazy<UserRepository>`. This routes user operations through the domain boundary and prevents cross-domain write risks. Adds `searchUsers` and `saveUser` APIs to `UserRepository` to support these delegations. Test mocks were properly migrated.

---
*PR created automatically by Jules for task [1961428334386748199](https://jules.google.com/task/1961428334386748199) started by @dogi*

## Summary by CodeRabbit

* **New Features**
  * Added user search support through the application’s user repository.
  * Added direct user saving and improved retrieval for users identified by multiple IDs.
* **Bug Fixes**
  * Improved handling of duplicate user IDs and large user lookup requests.
  * Ensured health records and submission details consistently retrieve and save user information.
* **Tests**
  * Updated coverage for user lookup, persistence, searching, and submission serialization scenarios.


dogi:
**Review — rating: 68/100**

Every substitution is behaviour-preserving, and I checked each rather than assuming. `UserRepositoryImpl.getUserById` (line 104) is literally `userDao.getById(userId)`, the same call being replaced — which mattered, because `UserRepository` also exposes `getUserByAnyId`, and picking that one would have changed the matching semantics. `searchUsers` and `saveUser` are thin wrappers over `userDao.search`/`userDao.upsert`, and `getAllUsers` over `userDao.getAll()`. Tests are updated properly, including stubbing `userRepositoryLazy.get()` in both `@Before` blocks. The `dagger.Lazy` is also genuinely required rather than cargo-culted: `UserRepositoryImpl` injects `HealthRepository` at line 83 and calls it at 852/867, so this creates a real bidirectional dependency that would otherwise fail at KSP time.

**That cycle is the problem, though.** The change doesn't just need `Lazy` — it *creates* the situation that needs it. `UserRepositoryImpl` ↔ `HealthRepositoryImpl` now depend on each other, resolved by deferring one side's construction. And #15623, open in this same queue from the same author, **removes** a `dagger.Lazy<UserRepository>` from `VoicesRepositoryImpl` on exactly the opposite reasoning — that eliminating the cross-repository dependency is better than deferring the cycle. I marked that one `ready` earlier today. These two PRs can't both be the house style; worth settling which it is before either lands.

**It also conflicts with #15529, and loses the exchange.** That PR is already `ready` at 88/100, and it rewrites two of the same methods:

- `getUserMap` — #15529 replaces `userDao.getAll().filter { it.id in userIdSet }` with chunked `userDao.getUsersByIds(chunk)`, removing a full-table materialisation. This PR rewrites the same block to `userRepositoryLazy.get().getAllUsers().filter { … }`, which **keeps** the full scan and just routes it through another layer.
- `getPatientsSortedBy` — same story, `getAll()` preserved.

So landing this after #15529 would undo a verified optimisation unless someone rebases carefully, and the merge won't flag it because both sides "work".

Lastly, the premise is worth questioning. `CLAUDE.md` states the convention as implementations injecting "the **Room DAOs** they need (plus `ApiInterface`, `DispatcherProvider`, other repositories as needed)" — so a repository holding a DAO isn't a violation to be refactored away. Adding `searchUsers` and `saveUser` to the `UserRepository` interface purely to avoid a `UserDao` reference widens that interface for no behavioural gain.

**Verdict:** needs another round — sequence this against #15529 (whose `getUserMap`/`getPatientsSortedBy` versions are better), and reconcile the `dagger.Lazy` direction with #15623 before adopting a repository cycle here.

---

dogi:
**Re-review — Rating: 74/100**

CI green on all three checks. The DAO-coupling removal is right, and the new batch query is a genuine performance fix — but delegating one branch of `searchPatients` while leaving the other local introduces an inconsistency, and there are two FQN sites.

**The batch query is well done.** Replacing `userDao.getAll().filter { it.id in set || it._id in set }` with a chunked `IN` query removes a full-table scan from `getUsersByIds`. The host-variable arithmetic is also correct, which is worth saying explicitly since `userIds` is bound *twice*:

```sql
SELECT * FROM users WHERE id IN (:userIds) OR _id IN (:userIds)
```

At `chunked(400)` that's 800 bound variables against SQLite's 999 limit — comfortable headroom, and notably more careful than #15680, which lands at 1001. `distinctBy { it.id }` correctly collapses rows that match on both columns.

**`searchPatients` now sorts differently depending on the query**

```kotlin
val users = if (query.isBlank()) {
    userRepositoryLazy.get().getUsersSortedBy(sortField, descending)   // → sortUsers
} else {
    sortPatients(userRepositoryLazy.get().searchUsers(query), sortField, descending)
}
```

`sortUsers` (`UserRepositoryImpl.kt:358`) handles 16 fields. `sortPatients` (`HealthRepositoryImpl.kt:168`) handles exactly two — `joinDate` and `name` — and falls through to `else -> users` for everything else.

So after this change, sorting by `firstName`, `email`, `age`, `gender` and so on works when the search box is empty and **silently does nothing** as soon as the user types. Previously both branches went through `sortPatients`, so they were at least consistently limited. Route the search branch through the same repository sort and the asymmetry disappears — `sortPatients` then has no callers and can go with it, since it's a strict subset of `sortUsers`.

**Use imports, not fully-qualified names**

```kotlin
private val userRepositoryLazy: dagger.Lazy<UserRepository>   // HealthRepositoryImpl
private val userRepositoryLazy: dagger.Lazy<UserRepository>   // SubmissionsRepositoryImpl
```

Both files should `import dagger.Lazy` and declare `Lazy<UserRepository>`. `UploadManager` already does exactly that (`private val teamsRepository: Lazy<TeamsRepository>`), so there's local precedent for both the import and the name — the `Lazy` suffix on the property puts the DI mechanism in the field name for no benefit. `userRepository` reads better and matches the neighbours.

**Context:** this is one of the three PRs in the `HealthRepositoryImpl` dependency triangle with #15623 and #15637. The `Lazy<>` wrapper here is what keeps it from being a Dagger cycle, so it's load-bearing — worth a one-line comment saying so, since a future cleanup that "simplifies" it to a direct injection would fail at KSP time with a confusing message.

**Verdict:** good direction, and the batch query alone justifies the change. Fix the two FQNs and make both `searchPatients` branches sort through the same function, then this is ready.

---

dogi:
**Review — rating: 80/100**

Good work, and the parts most likely to be wrong are right. I checked each of them rather than assuming.

**The headline win is real.** `getUsersByIds` was `userDao.getAll().filter { it.id in userIdSet || it._id in userIdSet }` — materializing the entire `users` table to pick out a handful. Replacing it with an indexed `IN` query is the correct fix, and the same pattern is removed from `HealthRepositoryImpl.getHealthRecord` at the same time.

**The chunk size is correct, and that's not obvious.** The new query binds the collection twice:

```sql
SELECT * FROM users WHERE id IN (:userIds) OR _id IN (:userIds)
```

Room expands each occurrence into its own placeholder run, so a chunk of 400 costs **800** host variables, not 400. That clears SQLite's 999 ceiling with room to spare. This repo has tripped over that limit more than once, so please add a comment saying exactly that — the danger is someone later "optimizing" 400 up to 900 on the reasonable-looking assumption that 900 < 999, and shipping 1800 bindings. One line above the `chunked(400)` prevents it.

**The `Lazy` is genuinely required, and the cycle claim checks out.** `UserRepositoryImpl`'s constructor holds `private val healthRepository: HealthRepository` directly, so `HealthRepositoryImpl → UserRepository → HealthRepository` is a real Dagger cycle and `Lazy` is the right break. Worth flagging that the comment on the *Submissions* side is looser — `UserRepositoryImpl` has no `SubmissionsRepository` dependency, so there's no cycle there. The parenthetical ("we follow this pattern consistently") is the actual reason; the "which creates a cycle" clause reads as if it applies to that class. Reword so the next person doesn't go looking for a cycle that isn't there.

**Cross-chunk duplicates are handled.** A user whose `id` lands in one chunk and whose `_id` lands in another would be returned twice; `distinctBy { it.id }` catches it. That's safe here because `UserEntity.id` is `@PrimaryKey var id: String = ""` — non-null, so no null-collapse risk.

**Ordering changed and it's currently benign.** Results used to come back in `getAll()` table order; they now come back in chunk order. All three callers — `SubmissionViewModel:76`, `NotificationsRepositoryImpl:234`, `HealthRepositoryImpl:196` — immediately build a map keyed by id, so none of them can observe it. Fine today, worth knowing if a list-returning caller shows up later.

**`sortPatients` → `sortUsers` is a superset, not an equivalence.** The deleted `sortPatients` handled `joinDate` and `name`, and returned the list *unsorted* for anything else. `sortUsers` handles those two identically, plus `firstName`, `lastName`, `middleName`, `email`, `planetCode`, `parentCode`, `level`, `language`, and `gender`. So health screens passing any of those nine now get sorted output where they previously got insertion order. Almost certainly an improvement — just not the no-op the "refactor" framing implies, and worth a line in the PR description.

`searchPatients` with a blank query is equivalent: `getUsersSortedBy` → `sortUsers(getAllUsers(), …)` and `getAllUsers()` is `userDao.getAll().map { it }`, matching the old `userDao.getAll()`. (That `.map { it }` is a full identity copy for nothing — pre-existing, but easy to drop while you're here.)

**The gap: no new tests.** The test diff is entirely constructor-signature updates to existing suites — zero new `@Test` methods for a new DAO query, new chunking logic, and two new interface methods (`searchUsers`, `saveUser`). The one that would earn its keep is a chunk-boundary case: 401+ ids exercising the multi-chunk path and the `distinctBy` dedup, which is the only place this change can silently lose or duplicate rows. `searchUsers` is a two-liner and could ride along cheaply.

**Verdict:** `ready`. The query, the chunk arithmetic, the cycle break, and the dedup all hold up under checking, and CI is green. Add the chunk-size comment and a boundary test — the first matters more than the second, given how this codebase has met the 999 limit before.

---

Okuro3499 (approved):

PR #15631, opened 2026-08-13
resources: smoother viewer markdown utils loading (fixes #15775)

### LINKED ISSUE #15775: Move text-resource file reading and Markdown setup off the main thread
Fixes blocking view initialization in `ResourceViewerFragment` by moving text file reading and Markdown AST parsing to a background IO coroutine.

dogi:
Fixes blocking view initialization in `ResourceViewerFragment` by moving text file reading and Markdown AST parsing to a background IO coroutine.

---
*PR created automatically by Jules for task [17626145000222997628](https://jules.google.com/task/17626145000222997628) started by @dogi*

## Summary by CodeRabbit

* **New Features**
  * Improved Markdown rendering in the resource viewer, including formatted text and interactive links.
  * Preserved direct display for non-Markdown text files.

* **Bug Fixes**
  * Improved text loading responsiveness by moving file reading and Markdown processing off the main thread.
  * Startup now continues gracefully when optional secure storage or cache initialization fails.

* **Tests**
  * Added coverage verifying text resource display and graceful handling of secure storage unavailability.


dogi:
**Review — rating: 72/100**

The problem is real and the decomposition is right. `file.readText()` plus Markwon's `toMarkdown()` on the main thread is a genuine ANR source — an arbitrary-length markdown resource parsed synchronously during view setup on slow storage. Splitting `MarkdownUtils` into `parseMarkdown` (returns a `Spanned`, safe off-main) and `setParsedMarkdown` (applies it to the `TextView`) is exactly the split Markwon supports for this purpose, and making `setupTextViewer` `suspend` is the honest way to express it.

**`requireContext()` is called inside the IO block, and that's a crash path.**

```kotlin
val (text, spanned) = withContext(dispatcherProvider.io) {
    val rawText = file.readText()
    if (type == ResourceType.MARKDOWN) {
        Pair(rawText, MarkdownUtils.parseMarkdown(requireContext(), rawText))
```

`requireContext()` throws `IllegalStateException` when the fragment isn't attached, and the whole point of this change is that there's now a real window — a file read — during which the user can navigate away. Hoisting it to `val ctx = requireContext()` before `withContext` fixes it and costs nothing.

**Markwon is built twice per render, on the main thread.** `parseMarkdown` calls `create(context)`, then `setParsedMarkdown` calls `create(textView.context)` again. Constructing the instance — plugins, theme, image loader — is a meaningful part of the cost this PR is trying to move off the main thread, and the second construction happens right back on it. It also means the `Spanned` is applied by a *different* `Markwon` instance than the one that produced it; same configuration so it should behave, but Markwon's contract pairs `setParsedMarkdown` with the instance that parsed. Returning the instance alongside the `Spanned`, or caching it, resolves both.

**The test doesn't test what its name says.** `setupTextViewer_movesIoOffMainThread` asserts only `textContent.text == "Test Content"`, which would pass whether or not the read happened off the main thread. And the mock provider is assigned *after* `Robolectric.buildActivity(...).setup()`:

```kotlin
val activityController = Robolectric.buildActivity(...).setup()
…
fragment.dispatcherProvider = mockDispatcherProvider
```

By that point `onViewCreated` has run and `setupTextViewer` has already been launched with the Hilt-injected provider, so `mockDispatcherProvider` most likely never participates. Assigning `@Inject lateinit` fields from a test is also a sign the seam is in the wrong place — the dispatcher should come from a Hilt test module replacement, not post-hoc field assignment. As written this is close to a vacuous test on the behaviour it names.

**Nit:** `android.text.Spanned` is written fully qualified in both new `MarkdownUtils` signatures instead of imported. `@Config(sdk = [33])` also pins this test to one SDK where the rest of the suite doesn't.

**Verdict:** needs another round — hoist `requireContext()` out of the IO block, avoid the second `create()`, and either make the test actually observe the dispatcher or rename it to what it checks.

---

dogi:

CI green on all three checks. Much tighter than the version that carried `change` — the scope is now three files, and the threading is done correctly.

**The split is in the right place.** Markwon's expensive step is parsing, and `markwon.toMarkdown(rawText)` now happens inside `withContext(dispatcherProvider.io)` alongside `file.readText()`, with only `setParsedMarkdown` left on the main thread. That's exactly the API Markwon provides for this — `setParsedMarkdown(TextView, Spanned)` exists so the parse can be hoisted off the UI thread — so this isn't a workaround, it's the documented path.

`MarkdownUtils.create(ctx)` staying on the main thread is fine: building the plugin chain is cheap relative to parsing, and the resulting instance is needed on both sides of the dispatch.

**The detach race is handled.** `val ctx = requireContext()` is captured *before* the `withContext`, so the IO block never calls a throwing fragment accessor off the main thread. That's the same defect I flagged on #15639, and it's avoided here.

Worth noting the views are captured as locals too (`textContent` resolved before the suspend), so the resumption can't NPE through a nulled binding — at worst it writes text into a detached view, which is harmless. An `if (!isAdded) return` after the `withContext` would make that explicit rather than incidental, but it isn't a defect.

**Tests:** 73 lines of new `ResourceViewerFragmentTest` coverage for a path that had none. Good addition for a change that alters when work happens relative to the lifecycle.

**Minor:** `setParsedMarkdown` and the existing `setMarkdownText` now both set `movementMethod = CustomLinkMovementMethod()` — a shared private helper would keep the two from drifting if link handling ever changes.

**Verdict:** correct threading, right API, race avoided, tests added. Marking `ready`.

---

Okuro3499 (dismissed):
```
The production change (ResourceViewerFragment.kt:471-495, MarkdownUtils.kt:81-84) is correct and follows the standard Markwon pattern (parse off-main, render on-main). But the    
  test added in this branch (ResourceViewerFragmentTest.kt) is a regression in test quality relative to this same branch's own earlier commit: commit c3c2ddce7/69066853e had a      
  deterministic test using an injected TestDispatcherProvider (StandardTestDispatcher + runTest + testScheduler.advanceUntilIdle()) — the pattern this repo's docs/TESTING.md        
  establishes precisely to avoid timing-dependent tests. The final commit (7d8d3ac94) throws that out and replaces it with ShadowLooper.idleMainLooper() + Thread.sleep(100) to "wait
  for the coroutine," using the real Hilt-provided DispatcherProvider instead of a controllable test one. That's exactly the kind of sleep-based synchronization the project's       
  existing test infra (MainDispatcherRule, TestDispatcherProvider) exists to eliminate.                                                                                              
```

dogi:
**Review — rating: 76/100**

The production change is right, and it uses the API that exists for exactly this.

`Markwon.toMarkdown(String): Spanned` followed by `Markwon.setParsedMarkdown(TextView, Spanned)` is Markwon's sanctioned background-parsing split — parse off the main thread, apply on it. That's what this does, so a large markdown resource no longer blocks the UI thread while it's being parsed. `MarkdownUtils.create(ctx)` stays on the main thread before the hop, which is fine, and `setParsedMarkdown` reattaches `CustomLinkMovementMethod` the same way `setMarkdownText` did.

Making `setupTextViewer` `suspend` is the honest signature for that.

**Blocking — fully-qualified name in the test:**

```kotlin
tempFile = File(ApplicationProvider.getApplicationContext<android.content.Context>().getExternalFilesDir(null), "ole/test.txt")
```

`android.content.Intent` is already imported two lines up; `android.content.Context` needs the same treatment so this reads `getApplicationContext<Context>()`.

**Also in the test file:**

- `import kotlinx.coroutines.test.*` — a wildcard import. Every other test in this repo lists its imports explicitly, and the star hides which of `runTest`, `StandardTestDispatcher`, and `testScheduler` are actually in use.
- `import kotlinx.coroutines.Dispatchers` is unused — the test uses `StandardTestDispatcher()` throughout.
- Import ordering is scrambled: `java.io.File` sits after `org.robolectric.annotation.Config`, and `dagger.hilt.android.testing.BindValue` plus `org.ole.planet.myplanet.utils.DispatcherProvider` are appended at the very end, below the other `org.ole` imports.

**On the test itself** — worth saying it's a good test to have. Driving a real `ResourceViewerActivity` through Robolectric with a `@BindValue` `TestDispatcherProvider` and `@UninstallModules(DispatcherModule::class)` is the right way to prove the coroutine actually completes and the text lands. It asserts the `TEXT` path; the `MARKDOWN` path — the one this PR changes — is uncovered. A second case with `ResourceType.MARKDOWN` and a `# Heading` fixture asserting the rendered text would exercise the code the PR exists for.

**Nit:** `MarkdownUtils.setMarkdownText` and the new `setParsedMarkdown` now duplicate the `movementMethod` assignment. Having `setMarkdownText` delegate to `setParsedMarkdown(textView, markwon, markwon.toMarkdown(text))` would keep one place that knows how to finish a markdown TextView.

**Verdict:** `change`, for the FQN and the import hygiene in the test — the production change itself is `ready`-grade. Adding the markdown-path test case while you're in there would be worth more than either.

---

dogi:

Everything addressed, plus the two optional suggestions.

**Fixed:**

- `android.content.Context` is imported; the inline FQN is gone.
- `kotlinx.coroutines.test.*` replaced with explicit `StandardTestDispatcher` and `runTest`.
- Unused `kotlinx.coroutines.Dispatchers` import removed.
- Import block reordered — `BindValue` sits with the other Hilt testing imports, and the `org.ole` group is contiguous.

**Both optional items taken:**

- `setMarkdownText` now delegates rather than duplicating the finish step:

  ```kotlin
  fun setMarkdownText(textView: TextView, markdown: String) {
      val markwon = create(textView.context)
      setParsedMarkdown(textView, markwon, markwon.toMarkdown(markdown))
  }
  ```

  One place now knows how to attach `CustomLinkMovementMethod`, and both entry points go through it.

- `setupTextViewer_rendersMarkdownContentFromFile` covers the markdown branch — the path this PR actually changes. The previous test only exercised `TEXT`, which would have passed with the markdown parsing broken.

**The production change remains right:** `Markwon.toMarkdown(String)` off the main thread followed by `setParsedMarkdown(TextView, Spanned)` on it is Markwon's own background-parsing contract, so a large markdown resource no longer blocks the UI thread while parsing.

**Nit:** `import java.io.File` sits at the very end of the block, after `org.robolectric.annotation.Config`. Elsewhere in the project `java.*` and `javax.*` come before `kotlinx` and `org`.

**Merge note:** 7 commits behind `master`; I test-merged it and it's clean.

**Verdict:** `ready`.

---

Okuro3499 (approved):

dogi:
@Okuro3499 look also into https://claude.ai/code/session_017DTytfwRi65Bgz73qRnk4A

PR #15632, opened 2026-08-13
courses: smoother progress repository state mapping (fixes #15746)

### LINKED ISSUE #15746: Refactor `ProgressRepository` to return typed `CourseProgressState`
This refactors the `ProgressRepository` to return a strongly typed map of `CourseProgressState` instead of raw `JsonObject` representations. This improves type safety and removes raw JSON extraction from UI layers like `CoursesAdapter` and `CoursesViewModel`.

Note: The existing `getCourseProgress` method is retained but marked as deprecated for backwards compatibility until all usages are migrated. The new data class is named `CourseProgressState` to avoid conflicts with the existing Room entity `CourseProgress`.

dogi:
This refactors the `ProgressRepository` to return a strongly typed map of `CourseProgressState` instead of raw `JsonObject` representations. This improves type safety and removes raw JSON extraction from UI layers like `CoursesAdapter` and `CoursesViewModel`.

Note: The existing `getCourseProgress` method is retained but marked as deprecated for backwards compatibility until all usages are migrated. The new data class is named `CourseProgressState` to avoid conflicts with the existing Room entity `CourseProgress`.

---
*PR created automatically by Jules for task [8861795488637832506](https://jules.google.com/task/8861795488637832506) started by @dogi*


dogi:
**Review — rating: 74/100**

The typed model is the right destination. `HashMap<String?, JsonObject>` as an in-process return type meant every consumer went through string-keyed `getInt("current", progress)` lookups with no compiler help; `CourseProgressState(current, max)` replaces that with real fields, and tightening the key from `String?` to `String` is a bonus. I diffed the two implementations line by line and they're structurally identical — same DAO calls, same `groupBy`, same `for (courseId in courseIds)` loop that puts an entry for *every* requested id. That last detail mattered: `CoursesAdapter.progressState` derives `hasProgress = progress != null`, so if the new map populated a different key set the progress bar would appear on courses that previously showed none. It doesn't. Semantics are preserved exactly.

**But the migration is left half-finished, and that's a worse state than either endpoint.** `TakeCourseViewModel.kt:64` still calls `progressRepository.getCourseProgress(listOf(courseId), userModel?.id)` — the method this PR deprecates. So after this lands:

- both representations of the same data are live in production code,
- the two implementations are near-duplicates that must be kept in sync by hand,
- and the build now emits a deprecation warning at that call site.

`@Deprecated` is the right tool for published API with consumers you can't change. This is an internal interface with exactly one remaining in-repo caller. Migrating `TakeCourseViewModel` to `getCourseProgressTyped` and deleting `getCourseProgress` outright is a smaller diff than the deprecation machinery, and leaves one implementation instead of two.

If you'd rather keep the JSON-shaped method for now, at least make it delegate so there's a single source of truth:

```kotlin
override suspend fun getCourseProgress(courseIds: List<String>, userId: String?) =
    getCourseProgressTyped(courseIds, userId).entries.associateTo(HashMap()) { (id, state) ->
        id to JsonObject().apply {
            addProperty("current", state.current)
            addProperty("max", state.max)
        }
    }
```

**Nit:** `org.ole.planet.myplanet.model.CourseProgressState` is written out fully qualified in the interface signature and three times in the impl, while `CoursesAdapter` and `CoursesViewModel` import it normally. Worth importing it in the repository files too.

**Verdict:** needs another round — migrate `TakeCourseViewModel:64` and delete the old method rather than deprecating it.

---

dogi:

CI green on all three checks. Clean, well-carried-through typed-model refactor — this is the right direction and the edges are handled.

**The core change is a real improvement.** `HashMap<String?, JsonObject>` was JSON used as an in-memory data structure: every read went through `getInt("current", progress)` with no type safety and no compiler help if a key were misspelled. `Map<String, CourseProgressState>` makes `progress.current` and `progress.max` ordinary field accesses, and it lets the nullable key type go away too.

The follow-through is complete — `CoursesAdapter`, `CoursesViewModel`, and `TakeCourseViewModel` all move to field access, the now-unused `JsonUtils.getInt` import is dropped from `CoursesAdapter`, and both affected test files are updated.

**The JSON boundary is handled correctly.** `fetchCourseData` still has to emit JSON, so it rebuilds the object from the typed state:

```kotlin
if (progressState != null) { … obj.add("progress", progressObj) } else { obj.add("progress", null) }
```

That preserves the previous shape exactly — the old code passed a possibly-null `JsonObject` to `add`, which Gson stores as `JsonNull`, and the explicit null branch does the same. Easy place to accidentally emit an absent key instead of a null one; it didn't.

Keeping the conversion at the one place that actually needs JSON, rather than carrying `JsonObject` through the ViewModel and adapter, is the right boundary.

**Import ordering** — two files place the new import out of alphabetical order:

```kotlin
// CoursesAdapter.kt — model.* inserted above callback.*
import org.ole.planet.myplanet.model.CourseProgressState
import org.ole.planet.myplanet.callback.OnCourseItemSelectedListener

// CoursesViewModel.kt — model.CourseProgressState placed after model.Tag
```

Both should sort into their existing blocks. Cosmetic, but the surrounding imports are otherwise ordered.

**Minor:** `CoursesUiState` still carries `map: HashMap<String?, JsonObject>` alongside the newly typed `progressMap`. Out of scope here, but it's the same pattern one field over, and a natural follow-up while the context is fresh.

**Verdict:** correct, complete, and tests moved with it. Marking `ready`.

---

Okuro3499 (approved):

PR #15633, opened 2026-08-13
Refactor UserDao and Repositories for Narrow Queries

dogi:
Replaced multiple `getAll().filter { }` patterns in the repositories with targeted SQL queries within `UserDao`. This includes retrieving users by IDs, fetching synced users, retrieving users for health sync with a specific projection, and getting pending sync users within limits. Addressed code review feedback and user formatting requirements. Updated test to mock the newly created narrow queries.

---
*PR created automatically by Jules for task [3087866265564946628](https://jules.google.com/task/3087866265564946628) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/data/room/dao/UserDao.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/data/room/dao/UserDao.kt
@@ -9,23 +9,20 @@ import org.ole.planet.myplanet.model.UserEntity
 interface UserDao {
     @Query("SELECT * FROM users WHERE id = :id OR _id = :id LIMIT 1")
     suspend fun getById(id: String): UserEntity?
-
     @Query("SELECT * FROM users") suspend fun getAll(): List<UserEntity>
     @Query("SELECT * FROM users WHERE id IN (:ids) OR _id IN (:ids)") suspend fun getByIds(ids: List<String>): List<UserEntity>
     @Query("SELECT * FROM users WHERE _id IS NOT NULL AND _id != '' AND id NOT LIKE 'guest%'") suspend fun getSyncedUsers(): List<UserEntity>
-@Query("SELECT * FROM users WHERE _id IS NOT NULL AND _id != ''") suspend fun getUsersForHealthSync(): List<UserEntity>
+    @Query("SELECT id, _id, planetCode FROM users WHERE _id IS NOT NULL AND _id != ''") suspend fun getUsersForHealthSync(): List<org.ole.planet.myplanet.model.UserHealthSync>
     @Query("SELECT * FROM users WHERE _id IS NULL OR _id = '' OR isUpdated = 1 LIMIT :limit") suspend fun getPendingSyncUsers(limit: Int): List<UserEntity>
     @Query("SELECT * FROM users WHERE name = :name LIMIT 1") suspend fun getByName(name: String): UserEntity?
     @Query("SELECT * FROM users WHERE name = :name COLLATE NOCASE LIMIT 1") suspend fun getByNameIgnoreCase(name: String): UserEntity?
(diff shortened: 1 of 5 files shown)


dogi:
**Review — rating: 55/100 (duplicate)**

This is the same change as #15529: the same six new `UserDao` queries, the same eight rewrites in `UserRepositoryImpl` (`getUsersByIds`, `getSyncedUsers`, `getUsersForHealthSync`, `getAllUsers`, `getPendingSyncUsers`, `migrateGuestUser`, `buildUserFromJson`, `cleanupDuplicateUsers`, `bulkInsert`), the same `HealthRepositoryImpl` user-map fix, and byte-for-byte the same three `coEvery` swaps in `UserRepositoryBulkInsertTest`. Both edit the same lines, so only one can merge.

**Two ideas here are better than #15529** and should be carried over rather than lost:

- **`UserHealthSync` as a column projection.** `SELECT id, _id, planetCode` beats #15529's `SELECT *` for a path that only needs three columns.
- **SQL-side patient sorting.** #15529 leaves `HealthRepositoryImpl.getPatientsSortedBy` on `getAll()` + in-memory `sortPatients`; pushing it into `ORDER BY` is the right direction. I checked `sortPatients` (lines 168-175) and the `when` branches match exactly — `joinDate`, `name`, else-unsorted — so `else -> userDao.getAll()` is faithful.

**But several of the SQL translations are unfaithful in ways #15529 gets right**, and two are outright bugs:

- **`_id LIKE 'guest_%'` — `_` is a LIKE wildcard.** It matches any single character, so this also matches `guestX123`, `guests_…`, and so on. The Kotlin it replaces was `_id?.startsWith("guest_")`, a literal underscore. This is in `getGuestUsersByNames`, which drives `migrateGuestUser` — so it can select the wrong guest row and then `deleteById` it. #15529 uses `SUBSTR(_id, 1, 6) = 'guest_'`, which is exact. Fixing this needs `ESCAPE` or `SUBSTR`.
- **`id NOT LIKE 'guest%'` is case-insensitive.** SQLite's `LIKE` ignores ASCII case by default, so a user with id `GUEST123` is excluded here, while the original `!it.id.startsWith("guest")` includes it. #15529 uses `SUBSTR(id, 1, 5) != 'guest'` and pins this exact case in a test.
- **`_id != ''` isn't `isNullOrBlank()`.** A whitespace-only `_id` counts as synced here; #15529's `TRIM(_id) != ''` excludes it, matching the Kotlin.
- **`ORDER BY name ASC` doesn't reproduce `sortPatients`.** That function sorts on `value(it.name)`, which lowercases first. SQLite's default collation is binary, so `Zebra` sorts before `apple` in SQL and after it in Kotlin. Needs `COLLATE NOCASE`.
- **`getUsersForHealthSync` changes which fields are populated.** `mapToLightweightUser` (lines 134-140) copies `id`, `name`, `planetCode`; the replacement copies `id`, `_id`, `planetCode`. `name` is silently dropped from every returned user.
- **No chunking.** `getByIds(userIds)` binds the list twice (`id IN (:ids) OR _id IN (:ids)`), and in `bulkInsert` the ids come from a whole sync batch. Past ~500 documents that's over SQLite's 999-variable limit and throws in the sync path. #15529 chunks at 400 precisely because of the double binding, and at 500 for the single-binding queries. The union `getByIds(...) + getGuestUsersByNames(...)` also isn't de-duplicated, where #15529 applies `distinctBy { it.id }`.

#15529 also ships `UserDaoTest.kt` — seven cases pinning the SQL semantics, including the `GUEST123` case this PR gets wrong — where this one only adjusts the existing bulk-insert mocks.

Both edit the same lines, so only one can merge. **Recommend closing in favour of #15529**, and opening a small follow-up for the two genuinely better ideas above (the `UserHealthSync` projection and SQL-side patient sorting, with `COLLATE NOCASE`). Marking `duplicate`.

---

PR #15636, opened 2026-08-13
courses: smoother view pager listening (fixes #15704)

### LINKED ISSUE #15704: Fix ViewPager listener leaks
Fixes long-lived listener leaks by keeping a reference to the anonymous listener/callback and explicitly removing it in `onDestroy` / `onDestroyView`.

Tracked in PR #15636.

dogi:
Fixes long-lived listener leaks by keeping a reference to the anonymous listener/callback and explicitly removing it in `onDestroy` / `onDestroyView`.

---
*PR created automatically by Jules for task [16090928890726272274](https://jules.google.com/task/16090928890726272274) started by @dogi*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved screen lifecycle handling during course navigation and onboarding.
  * Reduced the risk of lingering page-change listeners after leaving these screens.

* **Chores**
  * Updated the Android app version to 0.64.60.


dogi:
**Review — rating: 84/100**

Correct on both sides. Hoisting the anonymous callbacks into fields and unregistering them symmetrically is the right shape: `ViewPager2` holds registered `OnPageChangeCallback`s strongly, the `TakeCourseFragment` one captures the fragment through `this@TakeCourseFragment.onPageSelected(position)`, and the v1 `OnPageChangeListener` in `OnboardingActivity` is the same story. The `?.let { }` guards are also right rather than defensive noise — the `TakeCourseFragment` registration sits inside `if (coursesPagerAdapter == null)`, so on a path where that block doesn't run the field is genuinely null and the unregister correctly no-ops. Doing the unregister in `onDestroyView` before `binding` goes out of scope is the correct ordering. CI green (build default + lite, test default).

One thing worth stating plainly: neither of these is a *leak* in the sense the title implies. In both cases the `ViewPager` is part of the view hierarchy being destroyed, so the listener and everything it captures become garbage at the same moment — nothing outlives its owner. What this fixes is register/unregister asymmetry, which is worth fixing on its own terms (it's what makes double-registration bugs possible later), but the description shouldn't send someone hunting for a retained-Activity leak that isn't there.

**Two nits:**

- **Field lifetimes are now asymmetric in `TakeCourseFragment`.** `pageChangeCallback` is nulled in `onDestroyView` but `coursesPagerAdapter` isn't — and the registration is guarded on `coursesPagerAdapter == null`. So after a back-stack pop the guard is false, the block doesn't run, and the recreated `ViewPager2` gets neither an adapter nor a callback. That behaviour predates this PR, so it isn't a regression, but the two fields should be cleared together; as it stands, anyone who later nulls `coursesPagerAdapter` in `onDestroyView` will silently change registration behaviour.
- **`OnboardingActivity.onDestroy` calls `super.onDestroy()` first**, then touches `binding.pagerIntroduction`. Safe for an Activity, but cleanup conventionally goes before the `super` call.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #15640, opened 2026-08-13
all: smoother feedback repository saving (fixes #15747)

### LINKED ISSUE #15747: Refactor Feedback save logic
Refactored `FeedbackFragment` to use a single `createAndSaveFeedback` repository method instead of building the `Feedback` entity in the UI and saving it separately.

dogi:
Refactored `FeedbackFragment` to use a single `createAndSaveFeedback` repository method instead of building the `Feedback` entity in the UI and saving it separately.

---
*PR created automatically by Jules for task [10436953021432051587](https://jules.google.com/task/10436953021432051587) started by @dogi*


dogi:
**Review — rating: 82/100**

Sound encapsulation. Create-then-save was a two-call sequence the caller had to get right, and folding it into `createAndSaveFeedback` makes it one operation the repository owns. It also quietly fixes a placement bug: `createFeedback` was being invoked *outside* the `launch`, so its `UUID`/`Date`/Gson work ran synchronously on the main thread while only the save was in the coroutine. Both are inside now.

**The interface grew where it should have shrunk.** `FeedbackFragment` lines 111 and 113 were the *only* callers of `createFeedback` and `saveFeedback` anywhere outside `FeedbackRepositoryImpl` — I checked `main` and `test`. So after this change both are externally unused, reachable only from `createAndSaveFeedback` inside the impl, yet all three remain on the public interface. The refactor is a good opportunity to drop `createFeedback` and `saveFeedback` from `FeedbackRepository` and make them private, leaving one method instead of three near-identical six-parameter signatures.

**Two things worth considering while you're in here:**

- **No failure path.** `createAndSaveFeedback` returns `Unit`, and the fragment unconditionally toasts "feedback saved", notifies the listener and dismisses. If the save throws, the exception propagates out of `viewLifecycleOwner.lifecycleScope` and crashes. That's pre-existing rather than introduced, but this codebase already uses `Result<Unit>` for exactly this shape elsewhere (`ResourcesRepository.removeResourcesFromShelf`), and returning one here would let the fragment stop claiming success on failure.
- **The fragment injects the repository directly** rather than going through a ViewModel, while `ui/feedback/` already has `FeedbackListViewModel`. Not this PR's job to fix, but worth noting it pulls in the opposite direction from #15625, which is moving `ResourcesFragment`'s writes behind its ViewModel.

**Nit:** no test. `createAndSaveFeedback` is a two-line composition, but it's a new interface method and `FeedbackRepositoryImpl` is trivially mockable — a case asserting `saveFeedback` receives what `createFeedback` produced would be cheap.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #15641, opened 2026-08-13
all: smoother network utils flowing (fixes #15701)

### LINKED ISSUE #15701: Fix `NetworkUtils` WhileSubscribed timeout
Adds a 5000ms stop timeout to the `SharingStarted.WhileSubscribed()` policy in `NetworkUtils.kt`. This optimizes the `ConnectivityManager` network callback management by keeping the upstream flow alive during brief periods where there are momentarily zero subscribers, such as during Fragment navigation.

Tracked in PR #15641.

dogi:
Adds a 5000ms stop timeout to the `SharingStarted.WhileSubscribed()` policy in `NetworkUtils.kt`. This optimizes the `ConnectivityManager` network callback management by keeping the upstream flow alive during brief periods where there are momentarily zero subscribers, such as during Fragment navigation.

---
*PR created automatically by Jules for task [470395011056008471](https://jules.google.com/task/470395011056008471) started by @dogi*


dogi:
**Review — rating: 80/100**

`WhileSubscribed(5_000)` is the right default and I'd take this, but the "Fix" in the title claims more than the change delivers, and it's worth being precise about why.

`WhileSubscribed()` with no argument means `stopTimeoutMillis = 0` — the upstream is cancelled the instant the last collector goes away and restarted when a new one arrives. The classic reason that's a bug is a config change: the last collector detaches, a new one attaches milliseconds later, and any expensive upstream setup is torn down and rebuilt in between. The canonical example is a `callbackFlow` that registers a system callback.

That isn't the situation here. `_currentNetwork` is a `MutableStateFlow` (line 48), so the upstream being restarted is a bare `.map { it.isConnected() }` over an in-memory value — cancelling and restarting it allocates nothing and releases nothing. The actual `ConnectivityManager` registration is managed separately through the `isListening` flag (lines 62-74) and is entirely independent of how many collectors this derived flow has. `stateIn` also retains its cached `value` across upstream restarts, so no collector ever observes a gap or a stale reading.

So this is idiomatic alignment rather than a defect fix: it stops a trivial operator from being torn down across rotation, at no cost. Fine to merge, but the description shouldn't imply a leak or a dropped-emission bug was being fixed — someone reading the history later will go looking for one.

**Nit:** since `isNetworkConnectedFlow` is `by lazy` on an `object` and lives for the process, a short comment explaining why 5 s was chosen (surviving configuration changes) would be more useful than the bare magic number.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #15642, opened 2026-08-13
sync: smoother status dashboard collecting (fixes #15708)

### LINKED ISSUE #15708: Fix: Make long-running sync collectors idempotent across lifecycle restarts
Using `distinctUntilChanged` on flows within `collectWhenStarted` does not prevent redundant emissions when the lifecycle returns to the `STARTED` state, because the collection is completely restarted and the operator's distinct state is lost.

This adds explicit state variables (`lastSyncStatus`, `lastHandledUpdate`) in `SyncActivity`, `DashboardActivity`, and `MyHealthFragment` to correctly track and suppress redundant status emissions across view re-subscriptions.

Tracked in PR #15642.

dogi:
Fix: Make long-running sync collectors idempotent across lifecycle restarts

Using `distinctUntilChanged` on flows within `collectWhenStarted` does not prevent
redundant emissions when the lifecycle returns to the `STARTED` state, because the
collection is completely restarted and the operator's distinct state is lost.

This adds explicit state variables (`lastSyncStatus`, `lastHandledUpdate`) in
`SyncActivity`, `DashboardActivity`, and `MyHealthFragment` to correctly track and
suppress redundant status emissions across view re-subscriptions.

---
*PR created automatically by Jules for task [3619798755365208288](https://jules.google.com/task/3619798755365208288) started by @dogi*


dogi:
**Review — rating: 62/100**

The premise is sound for two of the three call sites and wrong for the third, and the third is the one where it does damage.

**`syncStatus` — the guards are legitimate.** `SyncManager.syncStatus` is a `StateFlow`, and `collectWhenStarted` re-subscribes on every STOPPED→STARTED transition, at which point the current value is replayed. So `updateLastSyncStatus()` really did re-run on every return to the foreground with no new sync. Tracking `lastSyncStatus` suppresses that. Worth noting the guard is redundant *within* a subscription — `StateFlow` already refuses to emit two equal consecutive values — so its only effect is across re-subscription, which is exactly the stated goal.

**`MyHealthFragment` is a different case, and the guard is both useless and lossy.**

`RealtimeSyncManager` declares:

```kotlin
private val _dataUpdateFlow = MutableSharedFlow<TableDataUpdate>(extraBufferCapacity = 1)
```

`replay` defaults to **0**, so this flow never replays to a new subscriber. There is no re-subscription replay to suppress here — the guard cannot fire for the reason the PR exists.

What it *can* do is drop real events. `TableDataUpdate` is a `data class`, so `update != lastHandledUpdate` compares by value across `table`, `newItemsCount`, `updatedItemsCount` and `shouldRefreshUI`. Two consecutive single-record health syncs both produce `TableDataUpdate("health", 0, 1, true)` — identical by value, genuinely distinct events. The second is silently swallowed and the health UI never refreshes. That converts a correct event stream into a lossy one, in the one place where a missed refresh means the user sees stale medical data.

That guard should come out. If something is genuinely double-firing on this flow, the cause is elsewhere and worth finding.

**The fix is also partial on its own terms.** All three `last*` fields are instance state, so they reset to `null` when the Activity or Fragment is recreated — a rotation replays the status and re-runs the handler exactly as before. If replay-on-resubscribe is the real problem for `syncStatus`, a more durable shape is to react to something monotonic (the last-sync timestamp you're already reading in `updateLastSyncStatus`) rather than to the status value, so the check survives recreation.

**Nit:** the two `syncStatus` guards are written differently for no reason — `DashboardActivity` nests the work inside `if (status != lastSyncStatus) { … }` while `SyncActivity` uses an early `return@collectWhenStarted`. The early return reads better; worth using it in both.

**Verdict:** needs another round — drop the `MyHealthFragment` guard, and consider whether the two `syncStatus` guards should key off the sync timestamp so they survive recreation.

---

dogi:
**Re-review — Rating: 84/100** (was 62)

CI green on all three checks. This is a different — and correct — change from the one I reviewed before. The old version guarded a `MutableSharedFlow(replay = 0)` in `MyHealthFragment`, where the guard was both a no-op and lossy. This version targets `syncManager.syncStatus`, and the reasoning holds up.

**Why the guard is right here and was wrong before.** `syncStatus` is a `StateFlow<SyncStatus>` (`SyncManager.kt:82`), and `Syncing`/`Success`/`Error` are all data classes. `StateFlow` conflates by equality, so setting the same value twice never emits — meaning the collector can only ever see a duplicate through **collector restart**. That's exactly what `collectWhenStarted` does: `repeatOnLifecycle` cancels on STOP and re-collects on START, and a fresh `StateFlow` collector immediately receives the current value. So every background→foreground trip re-delivered the last `Success` and re-ran `updateLastSyncStatus()`.

Holding `lastSyncStatus` in an Activity field is the right shape precisely because the field outlives the collector. And because `StateFlow` already suppresses equal consecutive emissions, the guard cannot drop a real event — the only thing it filters is the replay. That's the distinction the earlier version got wrong.

**Worth a comment in the code**, because the obvious "simplification" reintroduces the bug: `syncStatus.distinctUntilChanged()` would *not* fix this. The operator's state lives inside the flow collection, so it's rebuilt on every `repeatOnLifecycle` restart and would happily re-deliver the replayed value. The field is load-bearing.

I traced the state transitions for the case that worries me — a genuine second success:
`Success("x")` → replay suppressed → `Syncing(...)` → not equal, handled, field updated → `Success("x")` → not equal to `Syncing`, handled. Correct. A repeat can only be swallowed if no intermediate state occurred, which `StateFlow` would have swallowed anyway.

**Minor:** the same three lines now appear in `DashboardActivity` and `SyncActivity`. If a third collector needs it, a small `collectDistinctAcrossRestarts` helper holding the last value outside the `repeatOnLifecycle` block would centralize the subtlety — and give the explanation one home instead of two.

**Verdict:** correct fix for a real, reproducible bug. Marking `ready`.

---

Okuro3499 (approved):

PR #15643, opened 2026-08-13
all: smoother repositories json parsing (fixes #15748)

### LINKED ISSUE #15748: Refactor JSON parsing inside repository loops
This PR fixes several places where JSON string columns were being read or parsed repeatedly per item in a loop.
- `SubmissionsRepositoryImpl.kt`: Hoisted `question.choices` parsing outside of `choices.joinToString` loop. Extracted repeated `JsonUtils.getString` calls for `parentId` and `_rev`.
- `FeedbackRepositoryImpl.kt`: Extracted `feedback.messages` reading and `_id` extraction to local variables to prevent re-reading properties multiple times.
- `VoicesRepositoryImpl.kt`: Cached `news.viewIn` parsing in a transient `parsedViewIn` property to avoid parsing it repeatedly in `getCommunityNew

dogi:
This PR fixes several places where JSON string columns were being read or parsed repeatedly per item in a loop.
- `SubmissionsRepositoryImpl.kt`: Hoisted `question.choices` parsing outside of `choices.joinToString` loop. Extracted repeated `JsonUtils.getString` calls for `parentId` and `_rev`.
- `FeedbackRepositoryImpl.kt`: Extracted `feedback.messages` reading and `_id` extraction to local variables to prevent re-reading properties multiple times.
- `VoicesRepositoryImpl.kt`: Cached `news.viewIn` parsing in a transient `parsedViewIn` property to avoid parsing it repeatedly in `getCommunityNews`, and extracted it locally in other single-item methods.
- No changes to API or external behavior were made, resolving issue 70/100.

---
*PR created automatically by Jules for task [8105532455419006369](https://jules.google.com/task/8105532455419006369) started by @dogi*


dogi:
**Review — rating: 65/100**

Most of this is a good set of parse-once fixes, and several are real. `JsonParser.parseString(question.choices).asJsonArray` was being re-parsed **inside** the `joinToString` lambda — once per selected choice id — and hoisting it out is a genuine win. Same for `_rev`/`parentId` in `upsertRoomSubmissionsFromSync` (parsed twice and four times respectively), `_id` in `insertNewsList` and `mapToFeedback`, and `JsonUtils.gson.toJson(JsonUtils.getJsonObject("user", doc))` where `user` had already been extracted one line above. The `News.parsedViewIn` cache is the best of them — `viewIn` was previously parsed twice per item per emission, once in `isVisibleToUser` and again in `calculateSortDate`. The null/empty guards before `gson.fromJson` in `deletePost`, `insertFeedback` and the reply lookup are correct defensive fixes.

**But the `getCommunityNews` rewrite drops posts that should be visible.**

```kotlin
allNews.mapNotNull { news ->
    val viewInStr = news.viewIn
    if (viewInStr.isNullOrEmpty()) return@mapNotNull null
    …
    if (isVisibleToUser(news, userIdentifier)) { … } else null
}
```

That early return fires *before* `isVisibleToUser` is consulted — but `isVisibleToUser` opens with:

```kotlin
if (news.viewableBy.equals("community", ignoreCase = true)) {
    return true
}
val viewIn = news.viewIn ?: return false
```

A post marked `viewableBy == "community"` is visible **regardless of `viewIn`**, and that check comes first. So any community post with a null or empty `viewIn` was shown before this change and is silently filtered out after it — on the community news feed, which is precisely where `viewableBy == "community"` posts live. The old `filter { isVisibleToUser(...) }` had no such shortcut.

The fix is to drop the early return and let `isVisibleToUser` decide, populating `parsedViewIn` only when `viewIn` is non-empty:

```kotlin
allNews.mapNotNull { news ->
    news.viewIn?.takeIf { it.isNotEmpty() }?.let { s ->
        news.parsedViewIn = try { gson.fromJson(s, JsonArray::class.java) } catch (e: Exception) { null }
    }
    if (isVisibleToUser(news, userIdentifier)) {
        news.sortDate = news.calculateSortDate()
        news
    } else null
}
```

**Nits:**

- **Indentation is broken** in the `SubmissionsRepositoryImpl` choices hunk — `val choicesArray` sits at a deeper indent than the block it's in, and `val choiceObject` is misaligned under it. Worth running the formatter over that file.
- **`buildNewsFromJson(doc, existing = null, underscoreId)`** puts a required parameter *after* a defaulted one, which forces every caller to supply both positionally. Moving `underscoreId` ahead of `existing` would read better.

**Verdict:** needs another round — the `viewableBy == "community"` path is being filtered out of the community feed.

---

dogi:
**Re-review — Rating: 84/100** (was 65)

CI green on all three checks. The defect I flagged before is fixed, and the surrounding null-safety work is a genuine improvement.

**The visibility bug is gone.** The previous version returned `null` from `mapNotNull` before `isVisibleToUser` ever ran, which silently dropped `viewableBy == "community"` posts from the feed. This version parses as a side effect and then evaluates visibility for **every** item:

```kotlin
allNews.mapNotNull { news ->
    news.viewIn?.takeIf { it.isNotEmpty() }?.let { s -> news.parsedViewIn = … }
    if (isVisibleToUser(news, userIdentifier)) { news.sortDate = news.calculateSortDate(); news } else null
}
```

No early exit, so no dropped posts.

**The caching now does what the title claims.** `isVisibleToUser` and `calculateSortDate` both read `parsedViewIn ?: gson.fromJson(...)`, so `viewIn` is parsed once per item instead of twice. I checked the field is `@Ignore`d (`News.kt:66-67`) and already present on the base, so there's no Room column and no schema bump implied.

**The null-safety fixes are the quiet win.** `FeedbackRepositoryImpl.kt:85` previously did `gson.fromJson(feedback.messages, JsonArray::class.java)` and then `.add(obj)` on the result — with a null or empty `messages` that's a null return followed by an NPE. Guarding to an empty `JsonArray` fixes a real crash. Same shape applied to three `viewIn` reads in `VoicesRepositoryImpl`.

The `mapToFeedback(json, existing, id)` and `buildNewsFromJson(doc, underscoreId, existing)` signature changes are also right — both were re-extracting `_id` from JSON that the caller had already parsed, and `news.user = JsonUtils.gson.toJson(user)` drops a duplicate `getJsonObject("user", doc)` call.

**One design note, not blocking:** caching parsed state by mutating the Room entity inside `mapNotNull` blurs the line between persistence model and view model. It's safe here — these instances come straight from a DAO query and aren't shared — but `parsedViewIn` is now a mutable field whose population depends on which code path loaded the entity, so a future caller reading it without going through this flow will silently get `null`. The `?: gson.fromJson(...)` fallbacks make that correct rather than broken, which is the right call.

**Verdict:** the reported bug is fixed, the optimization is real, and it fixes an NPE on the way through. Marking `ready`.

---

Okuro3499 (approved):

PR #15644, opened 2026-08-13
Refactor TeamFragment and MyHealthFragment to use textChanges Flow

dogi:
Refactor TeamFragment and MyHealthFragment to use textChanges Flow

- Replaced raw TextWatcher bindings with textChanges().debounce() flows.
- Collected the flows safely in the UI using collectWhenStarted.
- Handled lifecycle issues when binding flow to AlertDialog inputs by managing a Job explicitly.
- Removed manual watcher bookkeeping logic.

---
*PR created automatically by Jules for task [11360845641091705820](https://jules.google.com/task/11360845641091705820) started by @dogi*

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved patient search responsiveness by removing redundant delays.
  - Search now waits briefly until typing pauses, reducing unnecessary requests.
  - Search activity is properly canceled when dialogs close or screens are destroyed.
  - Team search now updates reliably in line with the screen lifecycle.

- **Tests**
  - Updated search test naming to reflect the current behavior.

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/ui/health/HealthViewModel.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/health/HealthViewModel.kt
@@ -59,7 +59,6 @@ class HealthViewModel @Inject constructor(
     fun searchPatients(query: String, sortBy: String = "joinDate", descending: Boolean = true) {
         searchJob?.cancel()
         searchJob = viewModelScope.launch {
-            delay(300)
             val loadingJob = launch {
                 delay(100)
                 _isListLoading.value = true
--- a/app/src/main/java/org/ole/planet/myplanet/ui/health/MyHealthFragment.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/health/MyHealthFragment.kt
@@ -16,9 +16,6 @@ import androidx.appcompat.widget.AppCompatSpinner
 import androidx.core.content.ContextCompat
 import androidx.fragment.app.Fragment
 import androidx.fragment.app.viewModels
-import androidx.lifecycle.lifecycleScope
-import kotlinx.coroutines.FlowPreview
-import kotlinx.coroutines.flow.debounce
 import androidx.recyclerview.widget.DividerItemDecoration
 import androidx.recyclerview.widget.LinearLayoutManager
 import androidx.recyclerview.widget.RecyclerView
@@ -26,6 +23,9 @@ import dagger.hilt.android.AndroidEntryPoint
 import java.util.Calendar
 import java.util.Locale
 import javax.inject.Inject
+import kotlinx.coroutines.FlowPreview
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.flow.debounce
 import org.ole.planet.myplanet.R
 import org.ole.planet.myplanet.databinding.AlertHealthListBinding
 import org.ole.planet.myplanet.databinding.AlertMyPersonalBinding
@@ -38,10 +38,6 @@ import org.ole.planet.myplanet.utils.TimeUtils
 import org.ole.planet.myplanet.utils.Utilities
 import org.ole.planet.myplanet.utils.collectWhenStarted
 import org.ole.planet.myplanet.utils.textChanges
-import kotlinx.coroutines.Job
-import kotlinx.coroutines.flow.launchIn
-import kotlinx.coroutines.flow.onEach
-import kotlinx.coroutines.launch
 
 @AndroidEntryPoint
 class MyHealthFragment : Fragment() {
@@ -312,12 +308,11 @@ class MyHealthFragment : Fragment() {
     @OptIn(FlowPreview::class)
     private fun setTextWatcher(etSearch: EditText, btnAddMember: Button, rv: RecyclerView) {
         searchJob?.cancel()
-        searchJob = etSearch.textChanges()
-            .debounce(300)
-            .onEach { editable ->
-                viewModel.searchPatients(editable?.toString() ?: "", "joinDate", true)
-            }
-            .launchIn(viewLifecycleOwner.lifecycleScope)
+        searchJob = collectWhenStarted(
+            etSearch.textChanges().debounce(300)
+        ) { editable ->
+            viewModel.searchPatients(editable?.toString() ?: "", "joinDate", true)
+        }
     }
 
     override fun onResume() {
--- a/app/src/main/java/org/ole/planet/myplanet/ui/health/MyHealthFragment.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/health/MyHealthFragment.kt
@@ -276,7 +276,7 @@ class MyHealthFragment : Fragment() {
         alertHealthListBinding?.let { binding ->
             binding.list.layoutManager = LinearLayoutManager(requireContext())
             binding.list.adapter = adapter
-            setTextWatcher(binding.etSearch, binding.btnAddMember, binding.list)
+            observeSearchInput(binding.etSearch)
             sortList(binding.spnSort, binding.list)
             dialog = AlertDialog.Builder(requireActivity(), R.style.AlertDialogTheme)
                 .setTitle(getString(R.string.select_health_member)).setView(binding.root)
@@ -306,10 +306,10 @@ class MyHealthFragment : Fragment() {
     }
 
     @OptIn(FlowPreview::class)
-    private fun setTextWatcher(etSearch: EditText, btnAddMember: Button, rv: RecyclerView) {
+    private fun observeSearchInput(etSearch: EditText) {
         searchJob?.cancel()
         searchJob = collectWhenStarted(
-            etSearch.textChanges()
+            etSearch.textChanges().debounce(300)
         ) { editable ->
             viewModel.searchPatients(editable?.toString() ?: "", "joinDate", true)
         }
--- a/app/src/main/java/org/ole/planet/myplanet/ui/health/MyHealthFragment.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/health/MyHealthFragment.kt
@@ -309,7 +309,7 @@ class MyHealthFragment : Fragment() {
     private fun setTextWatcher(etSearch: EditText, btnAddMember: Button, rv: RecyclerView) {
         searchJob?.cancel()
         searchJob = collectWhenStarted(
-            etSearch.textChanges().debounce(300)
+            etSearch.textChanges()
         ) { editable ->
             viewModel.searchPatients(editable?.toString() ?: "", "joinDate", true)
         }
--- a/app/src/main/java/org/ole/planet/myplanet/ui/health/MyHealthFragment.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/health/MyHealthFragment.kt
@@ -4,7 +4,6 @@ import android.app.DatePickerDialog
 import android.content.Intent
 import android.os.Bundle
 import android.text.TextUtils
-import android.text.TextWatcher
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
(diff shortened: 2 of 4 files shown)


dogi:
**Review — rating: 76/100**

The `MyHealthFragment` half is a genuine improvement. `doAfterTextChanged` fired `viewModel.searchPatients(...)` on *every keystroke*, and that call runs a DB query and an in-memory sort — adding `.debounce(300)` is the real win here, not the Flow conversion itself. Replacing manual `removeTextChangedListener` bookkeeping with a cancellable `Job` is cleaner, and both the `dialog?.setOnDismissListener { searchJob?.cancel() }` and the `searchJob?.cancel()` at the top of `setTextWatcher` are correct — the dialog can be reopened, and without that guard you'd stack collectors.

**But the PR uses two opposite idioms for the same job, in the same commit.** `MyHealthFragment` moves *to* `.launchIn(viewLifecycleOwner.lifecycleScope)`, while `TeamFragment` moves *away* from exactly that pattern to `collectWhenStarted`. Both are search-field debounce collectors. One of them should change.

There's a defensible argument for the split — `collectWhenStarted` cancels and re-subscribes on every STOPPED→STARTED transition, which resets the debounce window and re-registers the underlying `TextWatcher`, and for a search box inside a dialog that churn buys nothing. If that's the reasoning, `TeamFragment` is the one going the wrong way. Either way it needs to be stated, because as it stands the diff reads like two different authors.

**An unrelated semantic change is bundled in.** `observeTeamData` goes from `collectLatestWhenStarted` to `collectWhenStarted`. That's `collectLatest` → `collect`: the previous handler is no longer cancelled when a new value arrives. The body is `submitList` plus two synchronous view updates, so nothing suspends and the practical difference is nil today — but it's a real behavioural change with nothing to do with "use textChanges Flow", and it silently removes the protection if that handler ever gains a suspending call. Either revert it or explain it.

**Nit:** the new imports in `MyHealthFragment` land in two wrong places — `androidx.lifecycle.lifecycleScope`, `kotlinx.coroutines.FlowPreview` and `kotlinx.coroutines.flow.debounce` are inserted between `androidx.fragment.app.viewModels` and `androidx.recyclerview.*`, and a second cluster (`Job`, `launchIn`, `onEach`, `launch`) is appended below the `org.ole.*` block. `TeamFragment`'s imports are tidy by comparison.

**Verdict:** needs another round — settle on one collector idiom across both fragments, and separate out the `collectLatest` change.

---

dogi:

CI green on all three checks. Both concerns from the previous pass are addressed, and the `MyHealthFragment` conversion is a real behavioural improvement rather than a stylistic one.

**`collectWhenStarted` instead of `launchIn(lifecycleScope)` is the substantive fix.** `viewLifecycleOwner.lifecycleScope` keeps collecting while the fragment is stopped, so a backgrounded fragment would still run `searchTeams` on every debounced emission. `collectWhenStarted` wraps `repeatOnLifecycle(STARTED)`, so collection suspends when the fragment isn't visible. That's the difference between "same code, nicer shape" and an actual change.

**`MyHealthFragment` gains debouncing it didn't have.** The old `doAfterTextChanged` fired `viewModel.searchPatients(...)` on every keystroke — each one a DB query. `textChanges().debounce(300)` collapses a burst of typing into one search. For a patient list that's the difference between a dozen queries and one.

**The dialog lifecycle is handled properly**, which is the part I'd expect to go wrong. The `EditText` lives in a dialog while `collectWhenStarted` is scoped to the *fragment*, so a dismissed dialog would otherwise leave a collector running against a detached view. All three exits are covered:

- `setTextWatcher` cancels any prior `searchJob` before starting a new one, so reopening the dialog can't stack collectors
- `dialog?.setOnDismissListener { searchJob?.cancel() }` handles dismissal while the fragment stays STARTED
- `onDestroyView` cancels and nulls

That's more careful than the `removeTextChangedListener` it replaces, which only covered the third case.

**Minor:** `@OptIn(FlowPreview::class)` is likely unnecessary now — `Flow.debounce(Long)` left `@FlowPreview` in coroutines 1.6, and this project is on 1.11. It's pre-existing in `TeamFragment` rather than introduced here, so removing it (and the `FlowPreview` import) from both files would be a small tidy, not a fix.

**Verdict:** correct, and the lifecycle handling around the dialog is the detail that makes it safe. Marking `ready`.

---

Okuro3499 (dismissed):
```
File: app/src/main/java/org/ole/planet/myplanet/ui/health/MyHealthFragment.kt:311-315 (setTextWatcher), in combination with                                                      
  app/src/main/java/org/ole/planet/myplanet/ui/health/HealthViewModel.kt:59-72 (searchPatients).                                                                                     
  Problem: Double debouncing — the Fragment's Flow debounces 300ms before calling searchPatients, which itself debounces another 300ms internally before running the query. Search   
  now takes ~2x as long to respond to typing as before this branch. 
```

dogi:
**Review — rating: 78/100**

Replacing the manual `TextWatcher` with a `textChanges()` Flow is the right direction, and the lifecycle handling is better than what it replaces. One thing got dropped on the way.

**The lifecycle side is a real improvement.** The old code held a `TextWatcher` field and detached it in `onDestroyView` via `alertHealthListBinding?.etSearch?.removeTextChangedListener(...)` — which silently did nothing whenever `alertHealthListBinding` was already null. A `Job` cancelled in both `onDestroyView` *and* the dialog's `setOnDismissListener` can't miss that way, and it stops the search running against a dismissed dialog.

`TeamFragment`'s `onEach`/`launchIn` → `collectWhenStarted` conversion is a straight simplification and drops two now-unused imports.

**The debounce is missing from `MyHealthFragment`.** `TeamFragment` gets:

```kotlin
collectWhenStarted(binding.etSearch.textChanges().debounce(300)) { … }
```

but `MyHealthFragment` gets:

```kotlin
@OptIn(FlowPreview::class)
private fun setTextWatcher(etSearch: EditText, btnAddMember: Button, rv: RecyclerView) {
    searchJob?.cancel()
    searchJob = collectWhenStarted(etSearch.textChanges()) { editable ->
        viewModel.searchPatients(editable?.toString() ?: "", "joinDate", true)
    }
}
```

No `debounce`. Two leftovers show it was meant to be there:

- `import kotlinx.coroutines.flow.debounce` is added to the file and never used.
- `@OptIn(FlowPreview::class)` is only required *because* of `debounce`, so as written the annotation is unnecessary too.

This isn't a regression — the old `doAfterTextChanged` fired per keystroke as well — but `searchPatients` runs `userDao.search(query)` and then sorts the result, so every character typed is a query plus a sort. The sibling change in this same PR debounces exactly this pattern; the health search should match it.

**Nits:**

- `setTextWatcher` no longer sets a `TextWatcher`. `observeSearchInput` or similar would stop the name pointing at machinery that's gone.
- `btnAddMember` and `rv` are still parameters of `setTextWatcher` but neither is referenced in the new body. Worth checking whether they were already unused before this change — if so, this is a free cleanup while the signature is being touched.

**Verdict:** `change`. Add `.debounce(300)` to the health search (which also justifies the `@OptIn` and the import), and this matches its own `TeamFragment` half and goes to `ready`.

---

dogi:

Both findings fixed.

**The debounce is back**, so the health search now matches the `TeamFragment` half of this PR:

```kotlin
searchJob = collectWhenStarted(
    etSearch.textChanges().debounce(300)
) { editable ->
    viewModel.searchPatients(editable?.toString() ?: "", "joinDate", true)
}
```

That also retires the two loose ends the omission left behind — the `kotlinx.coroutines.flow.debounce` import now has a user, and `@OptIn(FlowPreview::class)` is required rather than vestigial. `searchPatients` runs a DAO query plus a sort, so this is the difference between one query per pause and one per keystroke.

**The rename landed too:** `setTextWatcher` → `observeSearchInput`, and the two parameters that were no longer referenced (`btnAddMember`, `rv`) are gone, so the signature matches what the function does.

**Still good from the first pass:** cancelling the `Job` in both `onDestroyView` and the dialog's `setOnDismissListener` closes the gap the old `TextWatcher` had, where `alertHealthListBinding?.etSearch?.removeTextChangedListener(...)` silently did nothing whenever the binding was already null.

**Merge note:** 7 commits behind `master`, merges clean.

**Verdict:** `ready`.

---

Okuro3499 (dismissed):
```
File/function: app/src/main/java/org/ole/planet/myplanet/ui/health/MyHealthFragment.kt:311-315 (setTextWatcher), interacting with                                                
  app/src/main/java/org/ole/planet/myplanet/ui/health/HealthViewModel.kt:59-71 (searchPatients).                                                                                     
  Problem: Double debouncing — the Fragment's textChanges().debounce(300) delays the call to searchPatients, which then applies its own delay(300) internally before querying.       
  Combined latency (~600ms) is a real, user-visible regression versus pre-branch (~300ms) behavior, introduced by this branch.
```

dogi:
**Re-review — Rating: 88/100** (was 86)

The new commits fix a real timing bug I missed last round. With `.debounce(300)` on the Fragment's `textChanges()` flow *and* `delay(300)` at the top of `HealthViewModel.searchPatients`, every keystroke was costing 600ms before the query ran. Dropping the ViewModel's delay is the right half to remove — debouncing is a property of the input source, not of the search operation, and `searchPatients` has exactly one caller (`MyHealthFragment:304`), so nothing else was depending on that implicit delay. The `searchJob?.cancel()` guard inside the ViewModel stays, which is the part that actually matters for a stale in-flight query.

`TeamFragment` moving from `launchIn(viewLifecycleOwner.lifecycleScope)` to `collectWhenStarted` is a genuine upgrade, not just a cosmetic swap — `lifecycleScope` keeps collecting while the fragment is merely stopped, so a background keystroke (or a restored `EditText` value) would fire a search against a view that isn't visible. `repeatOnLifecycle(STARTED)` gates that properly, and it matches what `MyHealthFragment` now does.

**One unused import.** Dropping `btnAddMember: Button` and `rv: RecyclerView` from the old `setTextWatcher` signature left `android.widget.Button` with no remaining reference in `MyHealthFragment` — the import line is now its only occurrence in the file. `RecyclerView` is still used elsewhere so that one stays. Please delete the `Button` import.

**`setupTextWatcher()` no longer sets a text watcher.** You renamed the equivalent in `MyHealthFragment` to `observeSearchInput`, which is the accurate name; the `TeamFragment` version kept the old one in the same diff. Worth making them match while you're in both files.

**The cancellation is threefold.** `searchJob` is cancelled at the head of `observeSearchInput`, again in `dialog?.setOnDismissListener`, and again in `onDestroyView`. `collectWhenStarted` already ties the job to `viewLifecycleOwner`, so `onDestroyView` is redundant, and the re-entry guard is redundant with the dismiss listener given `observeSearchInput` only runs when the dialog is built. Only the dismiss listener earns its place — it covers the window where the dialog closes but the view is still alive. Not worth blocking on; just noting the other two are noise you can delete.

**Minor, worth knowing about:** `textChanges()` ends with `.onStart { emit(text?.toString()) }`, so it replays the field's current contents whenever collection begins. Under `repeatOnLifecycle(STARTED)` that means a fresh (empty-query) search fires 300ms after every return to foreground, not just on first open. It's idempotent so nothing breaks, but it's an extra round-trip to the repository on each resume. If that shows up in profiling, `.drop(1)` after `onStart` — or `distinctUntilChanged()` — would settle it.

**On coverage:** renaming `searchPatients updates patientList with debounce` to drop "with debounce" is the honest move now that the ViewModel doesn't debounce. It does mean the 300ms behaviour is no longer covered anywhere, since it lives in a Fragment and there are no Fragment tests in this project. That's a pre-existing structural gap, not something this PR should fix.

CI green on build (default + lite) and test (default). Merges cleanly.

**Verdict:** `ready`. Drop the `Button` import and rename `setupTextWatcher` when you get a chance — neither needs to gate the merge.

---

Okuro3499 (changes requested):
```
the header getter still validates the cache with                 
currentCache.pwdHash == pHash (pwd.hashCode()) instead of comparing the actual password string
```

PR #15645, opened 2026-08-13
courses: smoother surveys sort views modelling (fixes #15749)

### LINKED ISSUE #15749: Refactor ViewModel CPU-bound sorting off main dispatcher
Refactor ViewModel CPU-bound sorting off main dispatcher

- In `CoursesViewModel` and `SurveysViewModel`, move `sortByTitle`, `sortByDate`, and `applyFilterAndSort` logic off the main dispatcher.
- Wrap sorting and filtering inside `withContext(dispatcherProvider.default)`.
- Use a `Job` to provide single-flight and cancel-previous semantics to avoid race conditions.
- Update `CoursesViewModelTest` and `SurveysViewModelTest` to advance scheduler accurately due to dispatcher usage.

dogi:
Refactor ViewModel CPU-bound sorting off main dispatcher

- In `CoursesViewModel` and `SurveysViewModel`, move `sortByTitle`, `sortByDate`, and `applyFilterAndSort` logic off the main dispatcher.
- Wrap sorting and filtering inside `withContext(dispatcherProvider.default)`.
- Use a `Job` to provide single-flight and cancel-previous semantics to avoid race conditions.
- Update `CoursesViewModelTest` and `SurveysViewModelTest` to advance scheduler accurately due to dispatcher usage.

---
*PR created automatically by Jules for task [7791877893281724971](https://jules.google.com/task/7791877893281724971) started by @dogi*


dogi:
**Review — rating: 78/100**

The core change is right. `sortCourses` and `applyFilterAndSort` were doing real per-element work on the caller's thread — `lowercase(Locale.getDefault())` per item, `getSortDate` per item, plus a full sort — invoked straight from spinner listeners on Main. Moving both to `dispatcherProvider.default` is correct, and injecting the provider into `SurveysViewModel` rather than reaching for `Dispatchers.Default` is the right call (two other PRs in this queue today hard-code it instead). The `sortJob`/`filterSortJob` cancellation is a genuine bonus — rapid toggling now abandons superseded work instead of queueing it.

**The test updates deserve specific credit.** Making this work asynchronous silently breaks every existing assertion that read `viewModel.surveys.value` immediately after a `sort()` call, and you found and fixed all eight of them with `advanceUntilIdle()`. That's the part of an async refactor that usually gets missed.

**`CoursesViewModel.applySort()` introduces a race, though.** It snapshots the list, sorts off-thread, then writes back:

```kotlin
val currentCourses = _coursesState.value.courses
val sortedCourses = withContext(dispatcherProvider.default) { sortCourses(currentCourses) }
_coursesState.value = _coursesState.value.copy(courses = sortedCourses)
```

If `loadCourses` completes between the read and the write, that write clobbers the freshly loaded courses with a sort of the *previous* list. `copy()` preserves the other fields, but `courses` is overwritten wholesale. `sortJob?.cancel()` guards against a second sort, not against a concurrent load — and this race didn't exist before, because the old version was synchronous on Main and therefore atomic with respect to other main-thread state updates.

`SurveysViewModel` is less exposed because it reads `rawSurveys` *inside* the `withContext`, so it picks up the current value at execution time — but the same write-back pattern applies. Guarding with a generation counter, or re-reading and re-checking identity before assigning, would close both.

**One of the two new tests doesn't test anything:**

```kotlin
@Test
fun testToggleSort_AppliesSortOffMainThread() = runTest {
    viewModel.toggleTitleSort()
    io.mockk.verify { dispatcherProvider.default }
}
```

`verify { dispatcherProvider.default }` asserts only that the property getter was read — not that sorting ran off the main thread, and not that the result is sorted. It would pass if `sortCourses` returned the list untouched. The `SurveysViewModelTest` changes in this same PR show what a real assertion looks like; this one should assert the ordering of `_coursesState.value.courses` after `advanceUntilIdle()`.

**Nit:** `private val dispatcherProvider: org.ole.planet.myplanet.utils.DispatcherProvider` is fully qualified in the `SurveysViewModel` constructor rather than imported.

**Verdict:** needs another round — guard the write-back in `applySort` against a concurrent load, and make the new Courses test assert the sort.

---

dogi:
**Re-review — Rating: 80/100** (was 78)

CI green on all three checks. The concurrency work is now correct — the stale-result handling in particular is better than I usually see for this pattern. One style rule is the only thing holding it.

**The stale-result guard is the right shape.** Moving the sort off the main thread makes the state update asynchronous, which opens a race: a sort started against list A can finish after the list has been replaced by B, clobbering it with stale data. Both ViewModels close it properly:

```kotlin
sortJob?.cancel()
sortJob = viewModelScope.launch {
    val currentCourses = _coursesState.value.courses
    val sortedCourses = withContext(dispatcherProvider.default) { sortCourses(currentCourses) }
    if (_coursesState.value.courses === currentCourses) { … }
}
```

The `cancel()` handles the common case, and the reference-identity check covers the window where cancellation loses the race to a completed `withContext`. Using `===` rather than `==` is right too — a structurally equal but freshly loaded list *should* invalidate the result. `SurveysViewModel` applies the same guard against `rawSurveys`. The 111 new lines of `CoursesViewModelTest` coverage suggest this was worked through rather than assumed.

**Use an import, not a fully-qualified name**

```kotlin
private val dispatcherProvider: org.ole.planet.myplanet.utils.DispatcherProvider
```

`SurveysViewModel` needs `import org.ole.planet.myplanet.utils.DispatcherProvider` and the short name in the constructor. `CoursesViewModel` — the other file in this same PR — already imports it and writes `dispatcherProvider.default`, so the two halves of the diff disagree with each other on style. This is the only thing keeping the PR at `change`; it's a one-line fix.

**Minor:** `import kotlinx.coroutines.Job` is inserted after `coroutineScope` in `CoursesViewModel`; `Job` sorts before the lowercase entries in the same block.

**Worth confirming, not blocking:** the payoff here scales with how expensive `getSortDate(it)` is, since `sortedByDescending` calls it per comparison rather than once per element. If it's a cheap field read, this trades real complexity — job tracking, cancellation, identity guards, async state updates — for a sub-millisecond win on a few hundred items. If it parses or computes anything, the change is clearly worth it. Either way the code is correct; it's a question of whether the ceremony is earning its keep, and a `sortedBy` with a precomputed key would sidestep the question entirely.

**Verdict:** correct and well-guarded. Fix the FQN and this is ready.

---

dogi:
**Re-review — Rating: 86/100** (was 80)

CI green on all three checks. The FQN is gone — `SurveysViewModel` now has `import org.ole.planet.myplanet.utils.DispatcherProvider` and declares `private val dispatcherProvider: DispatcherProvider`, matching `CoursesViewModel` in the same PR.

Everything else was already right and is unchanged. Restating the part worth keeping in mind for anyone reviewing similar work later: moving the sort off the main thread makes the state update asynchronous, which opens a window where a sort started against list A can land after the list has been replaced by B. Both ViewModels close it with `cancel()` plus a reference-identity check on the captured list, and `===` is the correct comparison — a structurally equal but freshly loaded list *should* invalidate a pending result.

**Verdict:** clean. Marking `ready`.

---

Okuro3499 (approved):

PR #15639, opened 2026-08-13
all: smoother base container dispatcher providing (fixes #15638)

Okuro3499:
fixes #15638
Consolidate dispatcherProvider injection into BaseContainerFragment, removing duplicate injections from subclasses. Move file I/O operations to background dispatcher and use Provider<SharedPreferences> in DownloadService to avoid early initialization.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved download and file checks by performing them in the background, helping keep the app responsive.
  - Resources now open automatically only after confirming the required files are available.
  - Improved reliability when downloading and opening HTML resources.

- **Refactor**
  - Streamlined background task and preference handling to improve stability and maintainability.

PR #15646, opened 2026-08-13
teams: smoother base voices tasks dispatcher providing (fixes #15799)

### LINKED ISSUE #15799: Drop redundant dispatcher ceremony from BaseTeamFragment and siblings
Refactored `BaseTeamFragment` to remove the redundant `dispatcherProvider.io` launch and `withContext(main)` block for `loadTeamDetails()`, as Room suspend calls safely run off the main thread.
Added a `Job` to track the team load to prevent double-fetching on rapid re-entry, matching `TeamDetailFragment`.
Removed the unused `dispatcherProvider` injection from `BaseTeamFragment` and its `override` in `TeamsVoicesFragment`. 
Cleaned up sibling redundant dispatcher launches in `TeamsTasksFragment`, `TeamsVoicesFragment` (for `isTeamLeaderFn`), and `ProcessUserDataActivity`.

dogi:
Refactored `BaseTeamFragment` to remove the redundant `dispatcherProvider.io` launch and `withContext(main)` block for `loadTeamDetails()`, as Room suspend calls safely run off the main thread.
Added a `Job` to track the team load to prevent double-fetching on rapid re-entry, matching `TeamDetailFragment`.
Removed the unused `dispatcherProvider` injection from `BaseTeamFragment` and its `override` in `TeamsVoicesFragment`. 
Cleaned up sibling redundant dispatcher launches in `TeamsTasksFragment`, `TeamsVoicesFragment` (for `isTeamLeaderFn`), and `ProcessUserDataActivity`.

---
*PR created automatically by Jules for task [15476049036712947539](https://jules.google.com/task/15476049036712947539) started by @dogi*


dogi:
**Review — rating: 70/100**

The principle is correct and worth applying. `lifecycleScope.launch(dispatcherProvider.io) { … withContext(dispatcherProvider.main) { updateUi() } }` is ceremony when everything inside is a `suspend` repository call — Room's generated DAOs confine themselves, so the hop to IO and back buys nothing and only adds two dispatches. `BaseTeamFragment.loadTeamDetails` is the clearest case: `getTeamById` and `isMember` are both suspend repository calls, so launching on Main and assigning `_teamFlow`/`_isMemberFlow` directly is simpler *and* removes a thread hop. Adding `loadTeamJob?.cancel()` is a real bonus that isn't in the title — overlapping loads could previously race on `team` and the two StateFlows.

`TeamsVoicesFragment` follows correctly (dropping `override` now that the base no longer declares the field), and `ProcessUserDataActivity` is fine — `lifecycleScope` already dispatches on `Dispatchers.Main.immediate`, so `launch(dispatcherProvider.main)` was redundant.

**But one site trades the injected provider for a hard-coded dispatcher:**

```kotlin
val (taskList, fetchedNames, currentSnapshot) = withContext(kotlinx.coroutines.Dispatchers.IO) {
```

`TeamsTasksFragment` genuinely *does* need to leave the main thread there — that block runs `myTasks()`/`completedTasks()` filtering plus assignee-name lookups, which is real work — so keeping the `withContext` is right. Hard-coding `Dispatchers.IO` for it is not. `CLAUDE.md` is explicit that `DispatcherProvider` should be injected rather than `Dispatchers.*` hard-coded, precisely so tests can substitute `TestDispatcherProvider`, and this is the second PR today doing it (#15626 does the same to `InlineResourceAdapter`).

This one looks like collateral damage rather than intent: removing `@Inject open lateinit var dispatcherProvider` from `BaseTeamFragment` took away the property `TeamsTasksFragment` was relying on, and the hard-coded dispatcher is the patch. The `import javax.inject.Inject` added to that file with no accompanying `@Inject` field looks like a half-finished attempt at the right fix — declaring its own `@Inject lateinit var dispatcherProvider`, exactly as `TeamsVoicesFragment` now does, and using `dispatcherProvider.io`.

**Nit:** worth double-checking whether any other `BaseTeamFragment` subclass referenced the removed `dispatcherProvider`. The build is green so nothing is unresolved, but subclasses that silently lost access are where more hard-coded dispatchers tend to appear next.

**Verdict:** needs another round — give `TeamsTasksFragment` its own injected `DispatcherProvider` instead of `Dispatchers.IO`. The rest of the change is good.

---

dogi:
**Re-review — Rating: 80/100** (was 70)

CI green on all three checks. The dispatcher removals are justified rather than cosmetic, and killing the `open lateinit var` wart is overdue.

**The `launch(dispatcherProvider.io)` → `launch { }` swap is sound.** The `loadTeamDetails` body only calls `teamsRepository.getTeamById` and `teamsRepository.isMember` — suspend repository functions bottoming out in Room DAOs, which Room dispatches to its own executor regardless of the calling dispatcher. So there was nothing to move off Main, and the paired `withContext(dispatcherProvider.main)` wrapping the state writes becomes redundant once the body already runs there. Both halves go together correctly.

Worth contrasting with #15657, where the same "the dispatcher is redundant" reasoning was applied to a body containing `ResponseBody.string()` — genuinely blocking work that still needed IO. Here every call in the block suspends, which is what makes the removal safe.

**`loadTeamJob?.cancel()` is a real improvement beyond the stated scope.** Re-entering `loadTeamDetails` previously stacked concurrent loads that raced to write `_teamFlow`; now the older one is cancelled first.

**Dropping `@Inject open lateinit var dispatcherProvider` from `BaseTeamFragment`** removes the `open`/`override` pair on an injected field that I flagged as a wart during #15639. Re-injecting it in the two leaf fragments that actually use it is the right trade — and I confirmed `TeamsTasksFragment` genuinely needs it (`withContext(dispatcherProvider.io)` at line 311 wraps a heavier block), so that injection isn't dead weight.

**Conflict with #15639 — these two pull in opposite directions.** #15639 hoists `dispatcherProvider` *up* into `BaseContainerFragment` (the grandparent of `BaseTeamFragment`), so every subclass inherits one. This PR pushes it *down*, removing it from `BaseTeamFragment` and re-declaring it in leaves. Both are defensible, but they can't both land as written — and if #15639 merges first, the leaf injections here become shadowing redeclarations. Worth deciding which direction the codebase wants before either goes in; I'd lean toward this one, since only a couple of fragments actually need it.

**Fully-qualified name left behind.** `TeamsVoicesFragment` still has `kotlinx.coroutines.withTimeoutOrNull(2000)` — and this PR removed the `kotlinx.coroutines.withContext` FQN from the very next line, so the remaining one now sits alone. It's pre-existing rather than introduced, so I'm not blocking on it the way I did for the FQN introduced in #15645, but it's a one-line cleanup in code you're already editing: add `import kotlinx.coroutines.withTimeoutOrNull`.

**Import ordering** in `TeamsTasksFragment`: `javax.inject.Inject` is inserted above `java.util.Calendar` (`java` sorts before `javax`), and `utils.DispatcherProvider` lands after `utils.TimeUtils.formatDate`. `BaseTeamFragment` puts `kotlinx.coroutines.Job` after the `flow.*` entries too.

**Verdict:** correct reasoning, and the job cancellation is a bonus fix. Marking `ready` — but resolve the direction question against #15639 before merging either, and the `withTimeoutOrNull` import is free while you're in there.

---

Okuro3499 (approved):

PR #15647, opened 2026-08-13
chat: smoother detail ui state view modelling (fixes #15776)

### LINKED ISSUE #15776: Refactor ChatDetailFragment to single UiState and remove repo injection
1. Introduced `ChatUiState` in `ChatViewModel` using `combine` and `stateIn` to merge seven disparate `StateFlow` streams (`selectedChatHistory`, `selectedAiProvider`, `selectedId`, `selectedRev`, `aiProviders`, `aiProvidersLoading`, `aiProvidersError`) into one.
2. Removed direct `ChatRepository` and `UserRepository` injection from `ChatDetailFragment`. Proxy methods were introduced in `ChatViewModel` to perform database operations, strictly adhering to the architectural separation of concerns.
3. Updated `ChatDetailFragment` to observe the single `chatUiState` flow using `collectWhenStarted`

dogi:
1. Introduced `ChatUiState` in `ChatViewModel` using `combine` and `stateIn` to merge seven disparate `StateFlow` streams (`selectedChatHistory`, `selectedAiProvider`, `selectedId`, `selectedRev`, `aiProviders`, `aiProvidersLoading`, `aiProvidersError`) into one.
2. Removed direct `ChatRepository` and `UserRepository` injection from `ChatDetailFragment`. Proxy methods were introduced in `ChatViewModel` to perform database operations, strictly adhering to the architectural separation of concerns.
3. Updated `ChatDetailFragment` to observe the single `chatUiState` flow using `collectWhenStarted`. Diffing mechanism utilizing locally scoped variables within the collector prevents redundant layout redraws.

---
*PR created automatically by Jules for task [17013242129277349947](https://jules.google.com/task/17013242129277349947) started by @dogi*


dogi:
**Review — rating: 62/100**

Removing `chatRepository` and `userRepository` from the fragment and routing through `sharedViewModel` is the right direction, and consolidating seven collectors into one is a reasonable goal. But the consolidation drops a call path.

**AI-provider observation no longer runs when the chat is opened from a news item.** The old code called `observeAiProviders()` unconditionally, before the `if (newsId != null)` branch. The new `observeUiState()` absorbs that responsibility — but it's only invoked in the **else** branch:

```kotlin
if (newsId != null) {
    loadNewsConversations(newsId, newsRev, newsConversations)
} else {
    observeUiState()
}
```

So on the news path (`newsId != null`), nothing observes `aiProviders`, `aiProvidersLoading` or `aiProvidersError` any more. `updateAIButtons` never fires, the "fetching AI providers" dialog never shows or dismisses, and the cached-provider fallback on error never runs — even though `checkAiProviders()` is still called and still sets all three. `observeUiState()` needs to be called unconditionally, with only the chat-history handling gated on `newsId`.

**The five `last*` locals don't do what they look like they do.** They're declared inside `observeUiState()`, so they're re-initialised to `null` every time the collector restarts — and `collectWhenStarted` restarts on every STOPPED→STARTED transition, at which point the `StateFlow` replays its current value. Every branch therefore re-fires on each return to the foreground, including:

```kotlin
mAdapter.clearData()
sharedViewModel.clearPaginationState()
binding.editGchatMessage.text.clear()
```

which discards whatever the user had typed. That behaviour existed before this PR too — the old per-field `collect` calls replayed identically — so it isn't a regression. But adding five variables that give the appearance of deduplication without providing any across re-subscription makes the code harder to reason about, not easier. If dedup is wanted, `distinctUntilChanged` on the individual flows (or `distinctUntilChangedBy` on the combined state) does it properly and survives re-subscription.

**The seven-way `combine` gives up type safety.** Kotlin only provides typed `combine` overloads up to five flows; past that you get the `Array<Any?>` form, which is why this needs `@Suppress("UNCHECKED_CAST")` and positional `args[0] as List<Conversation>?`. Reordering the arguments would compile fine and fail at runtime with a `ClassCastException`. Nesting two `combine`s — or splitting into an `AiProviderState` and a `SelectionState` — keeps the compiler checking it.

**Nits:** `_id` and `_rev` are assigned on every emission while the other five fields are guarded, which is inconsistent even if harmless. The new imports (`SharingStarted`, `stateIn`, `combine`, `AiProvider`, `ChatResult`) are inserted mid-block, breaking the ordering.

**Verdict:** needs another round — call `observeUiState()` on both branches, and replace the seven-way untyped `combine`.

---

dogi:
**Re-review — Rating: 66/100** (was 62)

CI green on all three checks. The defect I flagged before is fixed — `observeUiState()` is now called unconditionally, with `if (newsId == null)` guarding only the chat-history branch, so AI providers are observed on both paths. The `ChatUiState` consolidation is a genuine improvement over five separate collectors. But the manual change-tracking introduced to support it has a lifecycle bug.

**The `last*` guards suppress the state that repopulates recreated views**

```kotlin
private var lastChatHistory: List<Conversation>? = null      // :71-75, fragment fields
private var lastAiProvider: String? = null
private var lastAiProvidersLoading: Boolean? = null
private var lastAiProvidersError: Boolean? = null
private var lastAiProviders: Map<String, Boolean>? = null
```

None of these is reset in `onDestroyView` (`:703-711` clears the binding, the text watcher, `loadingJob` and the speech recognizer — not these).

`collectWhenStarted` is scoped to the **view** lifecycle, so on rotation or a back-stack return the collector is torn down and re-established, and `chatUiState` — a `StateFlow` — immediately replays its current value to the new collector. That replay is the only thing that populates the freshly inflated views. But every guard evaluates false:

```kotlin
if (state.aiProviders != lastAiProviders) { … }   // false: lastAiProviders survived the view
```

So the new views never receive the provider list, the chat history, or the loading/error state. The fragment comes back blank until something changes the state again.

This is precisely the inverse of #15642, and the contrast is worth holding onto. There, a field surviving the collector restart was *correct*, because `StateFlow` conflation meant the replay was the only duplicate and suppressing it avoided redundant work. Here the replay is the only delivery the new views will ever get, so suppressing it is the bug. The deciding question is whether the thing being guarded is an *action* (suppress the replay) or a *rendering* (never suppress it).

Simplest fix: null all five in `onDestroyView`. Alternatively drop the manual tracking and put `distinctUntilChanged()` on the flow — because that operator's state is rebuilt with each collection, it gives the right behaviour on both axes for free.

Note `state.selectedId != _id` and `state.selectedRev != _rev` are fine as written — those compare against the fragment's live values rather than a shadow copy, so they self-correct.

**Five new pass-through methods on the ViewModel.** `fetchAiProviders`, `getLatestRev`, `sendNewChatRequest`, `sendContinueChatRequest`, and `getUserById` all just forward to repositories, and the fragment still calls them from its own scope. That does achieve the title's goal — the fragment no longer injects repositories — but it's relocation rather than encapsulation, the same shape as #15655. Fine as a step; worth saying so in the description.

**Import block needs sorting:**

```kotlin
import kotlinx.coroutines.withContext
import kotlinx.coroutines.flow.SharingStarted      ← flow.* belongs above withContext
import kotlinx.coroutines.flow.stateIn
import kotlinx.coroutines.flow.combine             ← after stateIn
import org.ole.planet.myplanet.model.AiProvider
import org.ole.planet.myplanet.repository.ChatResult   ← between two model.* entries
import org.ole.planet.myplanet.model.ChatHistory
```

Plus a double blank line before `data class ChatUiState`.

**Verdict:** the previous bug is genuinely fixed and the single-UiState shape is better than what it replaces. Reset or remove the `last*` shadows — as written, rotating the chat screen leaves it empty.

---

dogi:
**Re-review — Rating: 86/100** (was 66)

CI green on all three checks. The lifecycle bug is fixed — all five shadows are cleared in `onDestroyView`:

```kotlin
lastChatHistory = null
lastAiProvider = null
lastAiProvidersLoading = null
lastAiProvidersError = null
lastAiProviders = null
```

That's what makes the guards correct rather than harmful. `collectWhenStarted` is view-scoped, so a recreated view gets a fresh collector and `chatUiState` replays its current value; with the shadows cleared, every `if (state.X != lastX)` now evaluates true and the new views actually get populated. Previously the surviving fields suppressed exactly the emission the new views depended on, and the chat screen came back blank.

Worth restating the distinction this PR and #15642 sit on either side of, since both use a field that outlives the collector: suppressing the replay is right when the guard protects an **action** (#15642's `updateLastSyncStatus`), and wrong when it protects a **rendering**. Clearing on view destruction is what lets the same mechanism serve both.

The rest was already good and is unchanged: `observeUiState()` called unconditionally with `if (newsId == null)` scoping only the chat-history branch, and the single `ChatUiState` replacing five separate collectors.

**Still worth tidying, not blocking:** the `ChatViewModel` import block has `kotlinx.coroutines.flow.*` entries after `withContext` and `combine` after `stateIn`, plus `repository.ChatResult` between two `model.*` imports. And the five pass-through methods (`fetchAiProviders`, `getLatestRev`, `sendNewChatRequest`, `sendContinueChatRequest`, `getUserById`) achieve the title's goal but are relocation rather than encapsulation — fine as a step, worth a line in the description.

**Verdict:** the rotation bug is resolved. Marking `ready`.

---

Okuro3499 (dismissed):
```
 Problem: None of the new public surface is tested: the chatUiState combined flow (does it actually emit a coherent merged state when multiple sources change, does the             
  WhileSubscribed(5000) restart behave correctly), nor the five new proxy methods (fetchAiProviders, getLatestRev, sendNewChatRequest, sendContinueChatRequest, getUserById). Given  
  CLAUDE.md's documented near-total ViewModel test coverage and that these are exactly the kind of "encapsulate repository calls" behavior this refactor is supposed to guarantee,   
  shipping this without tests undercuts the stated goal.XSXXV                                       
```

dogi:
**Re-review — Rating: 86/100** (unchanged)

Re-checked against the current head; the code is unchanged from the revision I approved, so this is a straight re-affirmation. The PR appears to be back in `review` because `master` advanced, not because anything here moved.

The fix I was holding for is still in place — all five shadow fields cleared in `onDestroyView` (lines 709-713):

```kotlin
lastChatHistory = null
lastAiProvider = null
lastAiProvidersLoading = null
lastAiProvidersError = null
lastAiProviders = null
```

That's what makes the change-detection guards safe here. The fields exist to suppress redundant *rendering* work, and a rendering guard that survives view destruction is wrong — the new view starts blank, the guard says "unchanged", and the UI stays empty. Clearing them on view teardown means the first emission after `onCreateView` always paints. (The mirror case is a guard protecting an *action* rather than a rendering, where surviving the restart is the correct behavior — worth keeping the distinction in mind, since both patterns appear in this codebase.)

`ChatViewModelTest` at 106 lines covers the state transitions.

**Merge note:** 7 commits behind `master`. I test-merged it — clean, no conflicts. Just needs the merge.

**Verdict:** `ready`, as before.

---

Okuro3499 (approved):

PR #15648, opened 2026-08-13
teams: smoother tasks view modelling (fixes #15777)

### LINKED ISSUE #15777: Refactor Teams Tasks to ViewModel
Moves data operations for teams tasks out of the fragment into the view model, and returns UI state flows.

dogi:
Moves data operations for teams tasks out of the fragment into the view model, and returns UI state flows.

---
*PR created automatically by Jules for task [6685021020648376908](https://jules.google.com/task/6685021020648376908) started by @dogi*


dogi:
**Review — rating: 66/100**

The architecture is right, and one choice here is notably better than a sibling PR: using `MutableSharedFlow<TaskActionEvent>` for the toasts and tab-switching is the correct tool for one-shot UI events. #15628 does the same job with a `StateFlow` and consequently drops events through conflation — this doesn't. Removing `getAssignee` from `TeamsRepository` is also right; it was a pure proxy to `userRepository.getUserById` that had no business on the teams interface.

**The event flow has no buffer, though.** `MutableSharedFlow<TaskActionEvent>()` defaults to `replay = 0` *and* `extraBufferCapacity = 0`, which makes `emit()` a **suspending rendezvous** — it blocks until a subscriber actually receives the value. The only collector lives inside `repeatOnLifecycle(STARTED)`, so if the user rotates, backgrounds the app, or simply dismisses the dialog between the repository write and the emit, the emit parks. The write has already happened, but the toast and the tab switch are deferred until a collector reappears — which surfaces as a stale toast much later, or never. `RealtimeSyncManager` in this codebase already models the fix: `MutableSharedFlow(extraBufferCapacity = 1)`.

**The test file is the weaker half of this PR.** It introduces problems rather than coverage:

- **It uses Mockito.** `CLAUDE.md` is explicit — "MockK (not Mockito)" — and every other test in this repo uses MockK. This adds a second mocking framework for one file.
- **The `dispatcherProvider` mock is unusable.** `mock(DispatcherProvider::class.java)` returns `null` for `.io`, so any test that actually invoked `createOrUpdateTask`, `deleteTask`, `setTaskCompletion` or `assignTask` would fail on `viewModelScope.launch(null)`. It doesn't blow up today only because no test exercises them.
- **No tests for the moved logic.** That's the point of moving it to a ViewModel — `createOrUpdateTask`'s created-vs-updated branch and the corresponding event payloads are now unit-testable and untested.
- `StandardTestDispatcher`, `TestCoroutineScheduler` and `DefaultDispatcherProvider` are imported and unused; `InstantTaskExecutorRule` is a LiveData rule in a ViewModel that uses only StateFlow/SharedFlow; and `testDispatcher` is written as `kotlinx.coroutines.Dispatchers.Unconfined` despite `Dispatchers` being imported.

Swapping to MockK with a `TestDispatcherProvider` — both of which this repo already provides — would make the mock work and let the new methods actually be covered.

**Nit:** `TeamsTasksFragment` gains `import org.ole.planet.myplanet.ui.teams.tasks.TaskActionEvent`, which is its own package.

**Verdict:** needs another round — give the `SharedFlow` an `extraBufferCapacity`, and rebuild the test setup on MockK + `TestDispatcherProvider` with at least one case per new ViewModel method.

---

dogi:
**Re-review — Rating: 70/100** (was 66)

CI green on all three checks. The structural direction is right — task CRUD belongs in the ViewModel, and dropping `TeamsRepository.getAssignee` (a pure pass-through to `userRepository.getUserById`) removes a pointless cross-repository hop. Three things need attention before this lands.

**1. One-shot events can be lost**

```kotlin
private val _taskActionEvents = MutableSharedFlow<TaskActionEvent>(extraBufferCapacity = 1)
```

`replay = 0` with no active subscriber means the emission is **dropped**, not queued. `extraBufferCapacity` only prevents `emit` from suspending when a subscriber exists but is behind; it doesn't hold values for a collector that hasn't attached yet.

That matters because these are exactly the events that must not vanish. The work runs in `viewModelScope`, which survives configuration changes, while the fragment's collector does not — so a rotation (or a background trip) during `deleteTask` lands the `TaskDeleted` emission in the gap between the old collector being torn down and the new one attaching. The delete succeeds; the UI never learns about it.

For guaranteed-delivery one-shot events the standard shape is a channel:

```kotlin
private val _taskActionEvents = Channel<TaskActionEvent>(Channel.BUFFERED)
val taskActionEvents = _taskActionEvents.receiveAsFlow()
```

That buffers while no collector is attached and delivers to exactly one collector when it returns.

**2. `fetchAssigneeNames` is an N+1, and the batch API now exists**

```kotlin
suspend fun fetchAssigneeNames(assigneesToFetch: List<String>): Map<String, String> {
    return assigneesToFetch.mapNotNull { id -> userRepository.getUserById(id)?.name?.let { name -> id to name } }.toMap()
}
```

One query per assignee. #15630 adds `UserRepository.getUsersByIds(List<String>)` backed by a chunked `IN` query — a single round trip for the whole list. Worth depending on that and mapping the result, especially since a task list refresh calls this for every distinct assignee on screen.

**3. The import block needs a tidy**

```kotlin
import androidx.lifecycle.ViewModel
                                          ← blank line inside the import block
import androidx.lifecycle.viewModelScope
…
import org.ole.planet.myplanet.model.TeamTask
import dagger.hilt.android.lifecycle.HiltViewModel     ← org.ole.* above dagger.*
import java.util.Calendar
```

plus a second out-of-order group later (`org.ole.planet.myplanet.repository.*` and `model.UserEntity` landing after `utils.TimeUtils`), and two consecutive blank lines both after the imports and after the `taskActionEvents` declaration. Worth sorting into single ordered groups.

**Smaller things:**

- `viewModelScope.launch(dispatcherProvider.io)` on every action — these bodies only call suspend repository functions that bottom out in Room, which dispatches itself. That's the same redundant ceremony #15646 is removing elsewhere in this batch; plain `viewModelScope.launch { }` matches.
- `object TaskDeleted : TaskActionEvent()` can be `data object` on Kotlin 2.4, which gets you a sensible `toString`/`equals` for free.
- `getJoinedMembers` and `getAssignee` are pass-throughs to the repositories. Acceptable as the seam that keeps the fragment repository-free, but worth knowing they add no logic.

**Verdict:** good direction and the `getAssignee` removal is a genuine simplification. The event-loss window is the one that will produce a real bug report; fix that and the N+1, and this is close.

---

dogi:
**Review — rating: 82/100**

This is the best-executed ViewModel extraction I've seen in this queue. Three things it gets right that similar PRs usually don't:

**1. The event channel is the correct primitive.**

```kotlin
private val _taskActionEvents = Channel<TaskActionEvent>(Channel.BUFFERED)
val taskActionEvents: Flow<TaskActionEvent> = _taskActionEvents.receiveAsFlow()
```

collected inside `repeatOnLifecycle(Lifecycle.State.STARTED)`. This is the pairing that actually works for one-shot events. A `SharedFlow(replay = 0)` would *drop* a toast emitted while the fragment is stopped mid-rotation; a `SharedFlow` with replay would *re-fire* it on every collector restart. `Channel` + `receiveAsFlow` buffers while no collector is attached and delivers exactly once when one reattaches. Given how much of this codebase wrestles with replay-on-restart, it's worth naming as the pattern to copy.

**2. It removes an N+1 query.** The fragment previously did:

```kotlin
assigneesToFetch.mapNotNull { id -> userRepository.getUserById(id)?.name?.let { name -> id to name } }
```

one round trip per assignee. The ViewModel replaces it with a single batched call:

```kotlin
userRepository.getUsersByIds(assigneesToFetch).mapNotNull { user -> user.name?.let { user.id to it } }.toMap()
```

`getUsersByIds` is the chunked `IN` query, so a task list with 30 distinct assignees goes from 30 queries to one.

**3. It deletes a pointless indirection rather than moving it.** `TeamsRepository.getAssignee(userId)` was nothing but `userRepository.getUserById(userId)`. Dropping it from the interface *and* the 1437-line `TeamsRepositoryImpl`, and calling `userRepository` directly from the ViewModel, is the right call — one less method on the largest file in the project.

The extraction is also complete: `teamsRepository.` now appears **zero** times in `TeamsTasksFragment`. Not a partial migration. 162 lines of tests come with it.

**One thing to fix — `dispatcherProvider` is dead:**

```kotlin
class TeamsTasksViewModel @Inject constructor(
    private val teamsRepository: TeamsRepository,
    private val dispatcherProvider: DispatcherProvider   // ← never referenced
) : ViewModel()
```

It occurs exactly once in the file — the constructor line. Every new method uses a bare `viewModelScope.launch`, so nothing consumes it, and its import is dead with it. Please delete both. This matters more than usual right now: **#15667 through #15679 are fifteen open PRs whose entire content is removing unused constructor properties from other classes**, and it would be a shame to land a new one alongside them.

(If the intent was to dispatch repository calls off the main thread, note that Room DAO `suspend` functions already run on Room's own executors, so `viewModelScope.launch` is correct as written and the parameter genuinely isn't needed.)

**Smaller points:**

- `createOrUpdateTask` silently does nothing when `teamTask != null` but `teamTask.id == null` — the `let` block is skipped, so no repository call, no event, no toast, and the dialog closes as if it worked. The old code had the same hole via `?: return@launch`, so it's not new, but the `let` makes it easier to miss.
- Deadline validation is now split: the fragment checks `deadline.value == null` before calling, and the ViewModel calls `getDeadlineMillis()` internally. Since the ViewModel owns the deadline state, it could own the validation too and emit a `TaskActionEvent` for the error — that would make the whole operation testable in one place.

**Verdict:** `ready`. Delete the unused `dispatcherProvider` parameter and its import; everything else here I'd take as-is.

---

Okuro3499 (approved):

PR #15651, opened 2026-08-13
all: smoother view models loading (fixes #15751)

### LINKED ISSUE #15751: Refactor: Parallelize independent loads in ViewModels
Parallelizes independent suspend loads in `EventsDetailViewModel`, `HealthViewModel`, and `UserProfileViewModel` using `coroutineScope` and `async`. This removes sequential bottlenecks and significantly cuts time-to-first-paint for these screens.

dogi:
Parallelizes independent suspend loads in `EventsDetailViewModel`, `HealthViewModel`, and `UserProfileViewModel` using `coroutineScope` and `async`. This removes sequential bottlenecks and significantly cuts time-to-first-paint for these screens.

---
*PR created automatically by Jules for task [7039477272440110400](https://jules.google.com/task/7039477272440110400) started by @dogi*


dogi:
**Review — rating: 82/100**

CI green on all three checks. Small, correct, and the parallelism is real rather than nominal — which is the part worth checking, since `async` inside `viewModelScope` doesn't parallelize on its own.

`viewModelScope` runs on `Dispatchers.Main.immediate`, so each `async` body starts inline on the main thread and only yields at its first real suspension. That makes the win entirely dependent on the repository calls suspending rather than blocking — and they do, since these all bottom out in `suspend` DAO methods that Room dispatches to its own executor. So async1 suspends, async2 starts, async3 starts, and the three reads genuinely overlap. Worth stating explicitly because the same code shape would be a no-op (or worse, main-thread blocking) against a repository that did work before its first suspension point.

**Structured concurrency is handled correctly in all three:**
- `EventsDetailViewModel` — every `async` started is awaited on both branches, so `coroutineScope` has no orphaned child in either path. The blank-`meetUpId` branch correctly leaves `_meetup` and `_members` untouched, as before.
- `HealthViewModel` — both deferreds awaited immediately after starting; assignment order into `HealthData` unchanged.
- `UserProfileViewModel` — three reads overlapped, and the assignment order (`maxOpenedResource`, `lastVisit`, `numberOfResourceOpen`) matches the original, so observers see the same sequence.

Failure behaviour is also unchanged: an exception in any child cancels its siblings and propagates out of `coroutineScope` to `viewModelScope`, which is where it went before too.

**Two optional tidies, neither blocking:**

The `coroutineScope { }` wrapper is redundant — `launch { }` already provides a `CoroutineScope` receiver, and a failing `async` cancels the parent job identically either way. It does make the join point visually explicit, so keeping it is a defensible style call; just noting it isn't load-bearing.

`HealthViewModel` sets `_isLoading.value = false` after the block, so a throw from either repository leaves the spinner stuck on. That's pre-existing (the old sequential code had the same hole), but since you're already in the function, `try { … } finally { _isLoading.value = false }` closes it for free.

**Verdict:** correct refactor, no behaviour change, no tests needed for it. Marking `ready`.

---

Okuro3499 (approved):

PR #15652, opened 2026-08-13
Refactor `CoursesPagerAdapter` and `TeamPagerAdapter` manual diffing to `AsyncListDiffer`

dogi:
Replaced manual main-thread diffing in `CoursesPagerAdapter` and `TeamPagerAdapter` with `AsyncListDiffer` for consistency and performance improvements. Also verified that `UserArrayAdapter.kt` is already correctly using the `ListAdapter` pattern.

---
*PR created automatically by Jules for task [4781656101433460063](https://jules.google.com/task/4781656101433460063) started by @dogi*

## Summary by CodeRabbit

* **New Features**
  * Course completion now guides learners through required surveys and available ratings.
  * Survey lists support ascending and descending title sorting.
  * Guest users no longer see unavailable resource selection or library actions.
  * Activities charts now offer clearer monthly labels and improved empty-state displays.

* **Bug Fixes**
  * Improved restoration of selected course and team pages.
  * Improved landscape layouts for courses, libraries, activities, and profiles.
  * Refined rating controls, profile image sizing, and list visibility behavior.

* **Chores**
  * Updated the app version to 0.64.51.


dogi:
**Review — rating: 62/100**

CI green on all three checks. The mechanical translation is careful — but swapping a *synchronous* diff for an *asynchronous* one changes when `getItemCount()` becomes correct, and there's a caller that depends on it being correct immediately.

**Blocking: `setCurrentItem` now runs against a stale page count**

`TakeCourseFragment.kt:135-138`:

```kotlin
coursesPagerAdapter?.submitList(steps.mapNotNull { it?.id })

binding.viewPager2.isUserInputEnabled = false
binding.viewPager2.setCurrentItem(position, false)
```

The old `submitList` mutated `steps` and dispatched updates before returning, so by line 138 `getItemCount()` already reflected the new list. `AsyncListDiffer` only takes the synchronous fast path when its backing list is still `null` — i.e. the very first submit. Every later submit hands the work to a background thread and returns immediately, leaving `getItemCount()` reporting the *previous* count when `setCurrentItem(position, false)` runs one line down.

That path is reachable: the adapter is deliberately retained across invocations (`if (coursesPagerAdapter == null)` at :121, nulled only in `onDestroyView` at :463), and `onResume` re-enters this function once `steps` is initialized. So on a reload where the restored `position` is beyond the old step count, ViewPager2 clamps it, and nothing re-navigates once the diff lands — the user is left on the wrong step.

The fix is already idiomatic in this codebase — `AsyncListDiffer.submitList` takes a commit callback, and `CoursesFragment.kt:134` uses exactly that shape:

```kotlin
differ.submitList(listOf(null) + newSteps) { /* sequence setCurrentItem here */ }
```

`TeamPagerAdapter` is less exposed — `TabLayoutMediator` re-reads `getPageTitle` off adapter notifications, and `TeamDetailFragment.kt:225` already hedges with `?: pageConfigs.getOrNull(position)` — but it inherits the same window.

**Worth reconsidering the premise.** These are pager adapters holding a handful of items — 2–8 pages. `DiffUtil` over that many elements is microseconds, so the previous synchronous diff cost nothing and had no staleness window at all. `AsyncListDiffer` buys background diffing that isn't needed here and charges for it in ordering complexity. If the motivation is retiring the custom `DiffUtils` helper rather than performance, that's worth stating in the description, because on the merits the async machinery is a downgrade for this particular use.

**Also: the structural early-return is gone.** `if (newSteps == steps) return` compared by value; `AsyncListDiffer.submitList` only short-circuits on *reference* equality. An equal-but-freshly-built list — which is what `steps.mapNotNull { it?.id }` produces on every call — now schedules a full background diff where it used to return instantly. That's precisely the churn #15653 is trying to eliminate elsewhere, so the two PRs are pulling in opposite directions on this file.

**Two things I checked that are fine:**

The index translation in `createFragment` is correct. `currentList[position]` equals the old `steps[position - 1]` given the `null` prefix, and the guard rewrite `position < steps.size` → `position + 1 < currentList.size` is exactly equivalent. Easy place to slip; it didn't.

The `null` sentinel at index 0 is safe despite `DiffUtil.ItemCallback`'s non-null contract — `AsyncListDiffer` wraps the callback and null-guards before delegating, returning `oldItem == null && newItem == null` from `areItemsTheSame`, and `areContentsTheSame` is only reached once `areItemsTheSame` returned true, so the asymmetric-null `AssertionError` branch is unreachable. No change needed there; flagging it so nobody "fixes" a non-bug later.

**Verdict:** sequence the `setCurrentItem` through a commit callback (or keep the synchronous diff for these two adapters) and this is sound. As it stands the step-restore path can land on the wrong page.

---

dogi:

CI green on all three checks. The staleness window is closed exactly where it needed to be:

```kotlin
coursesPagerAdapter?.submitList(steps.mapNotNull { it?.id }) {
    binding.viewPager2.isUserInputEnabled = false
    binding.viewPager2.setCurrentItem(position, false)
    updateStepDisplay(position)
    …
}
```

Moving the whole post-submit block into `AsyncListDiffer`'s commit callback means `setCurrentItem(position)` now runs after the diff has been applied and `getItemCount()` reflects the new list. Previously it executed on the line after `submitList`, against a count that was still the old one on every submit past the first — so restoring a step position beyond the previous count got clamped to the wrong page.

Sequencing `setCourseData()` and `checkSurveyCompletion()` inside the callback too keeps them consistent with the pager state rather than partly ahead of it.

The parts I checked last time still hold: the index translation in `createFragment` is equivalent to the original (`currentList[position]` ≡ old `steps[position - 1]`, and `position + 1 < currentList.size` ≡ `position < steps.size`), and the `null` sentinel at index 0 remains safe because `AsyncListDiffer` null-guards before delegating to the callback.

**One open item from before, not blocking:** `AsyncListDiffer.submitList` short-circuits only on reference equality, so `steps.mapNotNull { it?.id }` — a freshly built list each call — still schedules a background diff where the old `if (newSteps == steps) return` returned instantly. Worth a cheap structural guard before `submitList` if this path runs often.

**Verdict:** the navigation bug is fixed and the sequencing is now explicit. Marking `ready`.

---

Okuro3499 (changes requested):
```
1. CoursesPagerAdapter.kt:29-43, submitList() — For a course with zero steps, the first call to submitList(emptyList()) hits if (newSteps == currentSteps) { commitCallback?.run();
  return } (both are emptyList<String>()), so it returns before ever calling differ.submitList(...). Unlike TeamPagerAdapter, CoursesPagerAdapter has no init {                      
  differ.submitList(...) } seeding, so differ.currentList stays at its default empty list forever. getItemCount() then returns 0, and the ViewPager2 shows no pages at all — not even
  CourseDetailFragment — for any course that currently has no steps (a state TakeCourseFragment.bindCourse explicitly anticipates via if (steps.isEmpty())). Previously              
getItemCount() was steps.size + 1, so the detail page always rendered.                                                                                                             
2. TeamDetailFragment.kt:206-246 (setupViewPager) and :373 (refreshTeamDetails) — These call currentAdapter.updatePages(pageConfigs) with no commitCallback, then separately do    
  binding.viewPager2.post { selectPage(restorePageId, false) }. Before this refactor, updatePages diffed synchronously on the main thread, so by the next post tick the adapter was  
  already up to date. Now updatePages on an existing adapter triggers AsyncListDiffer's background-thread diff path, which posts its result back to the main thread on its own       
  schedule — there's no guarantee it beats viewPager2.post{}. So selectPage() can call setCurrentItem(index, ...) against an index computed from the new pageConfigs while the       
  adapter is still serving the old list (stale item count/order), causing the wrong tab to be selected or a visible snap/jump. This is exactly the class of bug the second commit    
  already fixed for TakeCourseFragment via commitCallback — it just wasn't applied here. Reproducible on leaving a team (tab set shrinks) and on refreshTeamDetails().               
```

dogi:
**Re-review — Rating: 78/100** (was 86)

The code is unchanged and still correct. The drop is entirely about mergeability: **this no longer merges into `master`, and the obvious conflict resolution breaks the build.**

I test-merged all eighteen PRs currently in review against `master` (`658f49b`); this is the only one that conflicts, on `TakeCourseFragment.kt`.

**What happened.** `master` picked up `c5141b6` — *"courses: smoother completion rating (fixes #15439) (#15442)"* — which reworked 68 lines of this exact file. Among other things it **deleted `checkSurveyCompletion()` entirely**; the method no longer exists anywhere in `TakeCourseFragment`. Your diff moves that call into the new commit callback:

```kotlin
coursesPagerAdapter?.submitList(steps.mapNotNull { it?.id }) {
    binding.viewPager2.isUserInputEnabled = false
    binding.viewPager2.setCurrentItem(position, false)
    updateStepDisplay(position)
    if (position == 0) {
        binding.previousStep.visibility = View.GONE
    }
    setCourseData()
    checkSurveyCompletion()   // ← removed from master by #15442
}
```

So resolving in favour of this branch gives an unresolved reference, and resolving in favour of `master` throws away the fix.

**The resolution you want** keeps the callback structure and drops the one deleted call:

```kotlin
coursesPagerAdapter?.submitList(steps.mapNotNull { it?.id }) {
    binding.viewPager2.isUserInputEnabled = false
    binding.viewPager2.setCurrentItem(position, false)
    updateStepDisplay(position)
    if (position == 0) {
        binding.previousStep.visibility = View.GONE
    }
    setCourseData()
}
```

`master`'s current body is the same four statements plus `setCourseData()`, just executed eagerly instead of in the callback — so this is a clean superset once `checkSurveyCompletion()` is dropped.

**Why the callback still matters, unchanged from my last review:** `submitList` computes its diff asynchronously, so `setCurrentItem(position, false)` was running against the *previous* item count. `AsyncListDiffer` only fast-paths synchronously when its backing list is null — i.e. the very first submission — so on every subsequent load the ViewPager was told to jump to a position the adapter didn't yet report. Sequencing that work in the commit callback is the correct fix, and the same reasoning applies to the `TeamPagerAdapter` / `TeamDetailFragment` half of this PR.

**Verdict:** `change` — purely to get a merge that compiles. The four files are otherwise the same work I rated 86, and once rebased with `checkSurveyCompletion()` dropped, it goes straight back there.

---

dogi:
**Re-review — Rating: 70/100** (was 78)

Still conflicts, and the reason has changed since my last review — it's now bigger than a rebase.

**Most of the `TakeCourseFragment` diff is already in `master`.** Commit `c5141b6` (#15442, "courses: smoother completion rating") landed the entire course-rating feature this branch adds. On `master` today:

```
32:  import org.ole.planet.myplanet.ui.ratings.RatingsFragment
48:      lateinit var ratingsRepository: RatingsRepository
386:     R.id.finish_step -> onFinishStep()
391:     private suspend fun showCourseRatingDialogAndFinish() {
397:         val summary = ratingsRepository.getRatingSummary("course", cId, userId)
407:         val ratingDialog = RatingsFragment.newInstance("course", cId, title)
428:         showCourseRatingDialogAndFinish()
```

`onFinishStep`, `showCourseRatingDialogAndFinish`, the `ratingsRepository` injection, the `RatingsFragment` import, the `MANDATORY_SURVEY_COURSE_ID` extraction, and the removal of `checkSurveyCompletion` — all of it is upstream already. This PR re-adds the same code, which is why git can't auto-merge.

**What's actually still yours** is the one thing the PR is titled for. `master` still has:

```kotlin
coursesPagerAdapter?.submitList(steps.mapNotNull { it?.id })

binding.viewPager2.isUserInputEnabled = false
binding.viewPager2.setCurrentItem(position, false)
updateStepDisplay(position)
if (position == 0) { binding.previousStep.visibility = View.GONE }
setCourseData()
```

— eager, not sequenced behind the diff. Moving those five statements into the `submitList` commit callback is the novel contribution here, alongside the `CoursesPagerAdapter`, `TeamPagerAdapter`, and `TeamDetailFragment` changes.

**Suggested resolution:** rebase onto `master` and drop the entire rating-feature portion of the `TakeCourseFragment` diff, keeping only:

```kotlin
coursesPagerAdapter?.submitList(steps.mapNotNull { it?.id }) {
    binding.viewPager2.isUserInputEnabled = false
    binding.viewPager2.setCurrentItem(position, false)
    updateStepDisplay(position)
    if (position == 0) {
        binding.previousStep.visibility = View.GONE
    }
    setCourseData()
}
```

That should reduce the file to a handful of changed lines and make the conflict disappear.

**The reasoning behind the fix is unchanged and still worth landing.** `submitList` diffs asynchronously, so `setCurrentItem(position, false)` was running against the previous item count — `AsyncListDiffer` only fast-paths synchronously when its backing list is null, i.e. the very first submission. Every later load told the ViewPager to jump to a position the adapter didn't yet report. The same applies to the `TeamPagerAdapter` half.

**Nit for the rebase:** the current callback has a stray blank line before its closing brace and another after it.

**Related:** #15699 adds the equivalent rating-on-completion prompt for *resources* in `ResourceViewerFragment`, using the same `getRatingSummary` / `userRating ?: existingRating` shape. Not a conflict — different file — but worth knowing the two exist so the has-rated check doesn't drift apart.

**Verdict:** `change`. The adapter work is worth having; it just needs separating from code that's already merged.

---

dogi:
**Re-review — Rating: 45/100** (was 70)

The `AsyncListDiffer` conversion introduces a crash, and the PR has grown to 34 files under a title describing two of them.

**`CoursesPagerAdapter` will throw on the second `submitList`.**

The sentinel `null` for the course-detail page is still there:

```kotlin
differ.submitList(listOf(null) + newSteps, commitCallback)
```

but the callback it's now handed rejects nulls:

```kotlin
private val differ = AsyncListDiffer(this, object : DiffUtil.ItemCallback<String?>() {
    override fun areItemsTheSame(oldItem: String, newItem: String): Boolean {
```

`DiffUtil.ItemCallback<T>` declares its parameters `@NonNull`, so Kotlin sees `T & Any` — with `T = String?` that's `String`, which is why this compiles as an override. But a public override with non-null declared parameters gets `Intrinsics.checkNotNullParameter` emitted at its top, and `DiffUtil` is Java calling it with whatever the list holds. Index 0 holds `null`.

It survives the first call because `AsyncListDiffer` fast-paths an empty old list into insert-all without diffing. The second `submitList` — a course whose steps arrive in two passes, or navigating to a different course — runs the real diff, and `DiffUtil.calculateDiff` compares `oldList[0]` against `newList[0]` in its first snake. NPE, on the differ's background thread.

The old code avoided this by construction rather than by luck. `DiffUtils.calculateDiff` takes Kotlin lambdas — `areItemsTheSame: (oldItem: T, newItem: T) -> Boolean` with `T` inferred as `String?` from the very same `listOf(null) + steps` — so nulls were legal all the way through. Swapping a Kotlin generic helper for a Java `@NonNull` interface is where the property was lost.

CI can't see this: there's no test for `CoursesPagerAdapter`, and green here only means it compiles. The minimal fix is to stop putting `null` in the diffed list — a non-null sentinel (`"__course_detail__"`) keeps every index calculation identical and makes the callback honest. Typing the differ over a small sealed wrapper would be tidier still.

For what it's worth, I checked the index arithmetic and it's correct. With `currentList = [null, s0, s1, …]`, `currentList[position]` equals the old `steps[position - 1]`, `currentList[position + 1]` equals the old `steps[position]`, and the guard `position + 1 < currentList.size` reduces to the old `position < steps.size`. `TeamPagerAdapter` has no sentinel and no null problem — that half of the conversion is clean.

**The PR bumps the app version.** `app/build.gradle` goes `6444` → `6451` / `0.64.44` → `0.64.51`. Version bumps belong to `.github/scripts/automerge.sh`, which does them as part of draining the queue; master is already well past this number, which is why `build.gradle` is one of the three conflicting files. Drop that hunk.

**Scope has outgrown the title.** "Refactor `CoursesPagerAdapter` and `TeamPagerAdapter` manual diffing to `AsyncListDiffer`" now covers:

- deleting `layout-large-land/fragment_user_profile.xml` and `layout-normal-land/fragment_user_profile.xml` (262 lines) and reworking `layout-land/fragment_user_profile.xml`
- a new `layout-land/fragment_activities.xml`
- new `values-sw720dp/dimens.xml` plus `values/dimens.xml` additions
- a string in all six locales
- changes to `RatingsFragment`, `SurveyFragment`, `UserProfileFragment`, `ActivitiesFragment`, `TeamDetailFragment` and `CourseFilterController`
- a new 78-line `ActivitiesFragmentTest.kt`

All eight commits carry the identical message, so the history doesn't separate them either. The landscape-layout consolidation is plausibly good work, but it can't be reviewed on its merits while bundled with a pager refactor — and it's what makes this PR expensive to rebase. Splitting the layout work into its own PR would let both move.

**The `TakeCourseFragment` duplication I flagged before is now worse.** Master's `c5141b6` (#15442) already contains `ratingsRepository`, the `RatingsFragment` import, `showCourseRatingDialogAndFinish`, and `onFinishStep` — this branch re-adds all of it. Since then master also landed `02368af` (#15636, "smoother view pager listening"), which touches the same pager-callback area this PR restructures. The genuinely novel contribution to that file remains the `submitList` commit-callback, which is a handful of lines. Dropping the rest should collapse that conflict rather than require resolving it.

Conflicts against master in `app/build.gradle`, `TakeCourseFragment.kt` and `ResourcesFragment.kt` — the latter from #15599, #15591, #15625 and #15579 all landing since this branch was cut.

**Verdict:** `change`. Fix the null sentinel before anything else — that's a live crash. Then drop the version bump, split the landscape-layout work into its own PR, and reduce the `TakeCourseFragment` hunk to just the commit-callback so it merges against what's already upstream.

---

PR #15654, opened 2026-08-13
resources: smoother payload notifying (fixes #15753)

### LINKED ISSUE #15753: Refactor adapter notify calls to use targeted payloads
Replaces expensive `notifyDataSetChanged` calls with `notifyItemRangeChanged` and targeted notify payloads to optimize list adapter updates for grid/list mode switching. Also improves performance of `markItemAsOffline` and `setOpenedResourceIds` in `ResourcesAdapter` to avoid unnecessary loops.

dogi:
Replaces expensive `notifyDataSetChanged` calls with `notifyItemRangeChanged` and targeted notify payloads to optimize list adapter updates for grid/list mode switching. Also improves performance of `markItemAsOffline` and `setOpenedResourceIds` in `ResourcesAdapter` to avoid unnecessary loops.

---
*PR created automatically by Jules for task [5091109218306089579](https://jules.google.com/task/5091109218306089579) started by @dogi*


dogi:
**Review — rating: 76/100**

CI green on all three checks. Two of the three hunks are clear wins; the third is a pessimization dressed as an optimization.

**The `notifyDataSetChanged()` removals are the right call.** Both `setViewMode` implementations already guard on `if (viewMode == mode) return`, so the notify only fires on a genuine grid↔list switch, and `notifyItemRangeChanged(0, itemCount)` expresses that without nuking RecyclerView's entire view state. The view-type change is handled correctly too — a scrapped holder is only reused when `holder.itemViewType` matches what `getItemViewType(position)` now returns, so the switch still produces fresh holders of the new type. Good direction, and the opposite of what #15650 and #15612 are currently proposing.

**`markItemAsOffline` is a straight improvement** — `indexOfFirst` stops at the match instead of scanning the whole list to notify a single position. Identical behaviour given unique resource ids.

**`setOpenedResourceIds` goes the wrong way**

```kotlin
val changedIds = oldOpenedResourceIds.subtract(newOpenedResourceIds) union newOpenedResourceIds.subtract(oldOpenedResourceIds)
changedIds.forEach { id ->
    val index = currentList.indexOfFirst { it.item.id == id }
    …
}
```

The old loop was a single O(n) pass over `currentList` with two O(1) set lookups per item and zero allocations. The new version allocates three intermediate sets and then does `indexOfFirst` — an O(n) scan — once per changed id, making it **O(k·n)**.

That's never faster than what it replaced, and it's materially slower in the case that matters: the first population of `openedResourceIds`, where `k` equals the full set and the whole thing degrades to O(n²) over the resource list. For the incremental case (one resource opened, k=1) it's a wash on time and still allocates three sets the old code didn't.

If the goal is fewer wasted `notifyItemChanged` calls, the old loop already made exactly the same calls — it computed the symmetric difference implicitly, per item, in one pass. I'd just revert this hunk. If you want it genuinely faster, the win is a `Map<String, Int>` id→index maintained alongside `currentList`, which would help `markItemAsOffline` too.

**Nit:** `a.subtract(b) union b.subtract(a)` mixes call syntax and infix syntax for the same kind of operation. It parses correctly — infix binds looser than the dot calls, so the grouping is the intended symmetric difference — but `(a - b) + (b - a)` reads better and is the same thing.

**Verdict:** land the two `notifyItemRangeChanged` swaps and `markItemAsOffline` as-is. Revert or rework `setOpenedResourceIds` — as written it's slower and allocates more than the code it replaces, which cuts against the PR's stated purpose.

---

dogi:

CI green on all three checks. The `setOpenedResourceIds` rewrite is gone, which was the right call — it turned an O(n) single pass into O(k·n) plus three intermediate set allocations, and was never faster than what it replaced.

What remains is the part that was always worth landing:

- Both `setViewMode` implementations swap `notifyDataSetChanged()` for `notifyItemRangeChanged(0, itemCount)`. The `if (viewMode == mode) return` guard means these only fire on a real grid↔list switch, and RecyclerView still produces fresh holders for the changed view type, since a scrapped holder is only reused when its `itemViewType` matches.
- `markItemAsOffline` uses `indexOfFirst` instead of scanning the whole list to notify one position — same behaviour given unique resource ids, less work.

Both are movements away from `notifyDataSetChanged()`, which is the direction the adapter work across #15602 / #15618 / #15621 has been pushing.

**Verdict:** scoped down to the changes that pay for themselves. Marking `ready`.

---

Okuro3499 (approved):

PR #65, opened 2018-06-09
changed sync to image button and made sync and settings buttons bigger (fixes #64)

onlyserenity:
* Adjusted width, height, padding and other attributes with(in) layouts that are related to "sync" and "gear" imagebuttons

PR #15655, opened 2026-08-13
courses: smoother take view modelling (fixes #15800)

### LINKED ISSUE #15800: Refactor TakeCourseFragment to use TakeCourseViewModel
Moves direct `coursesRepository` dependencies from `TakeCourseFragment` to `TakeCourseViewModel` to improve architecture and testing per issue #65. The `coursesRepository` field was removed from the fragment, and all calls were updated to use corresponding delegating methods added to `TakeCourseViewModel`. No regression was introduced.

dogi:
Moves direct `coursesRepository` dependencies from `TakeCourseFragment` to `TakeCourseViewModel` to improve architecture and testing per issue #65. The `coursesRepository` field was removed from the fragment, and all calls were updated to use corresponding delegating methods added to `TakeCourseViewModel`. No regression was introduced.

---
*PR created automatically by Jules for task [13336096109315558517](https://jules.google.com/task/13336096109315558517) started by @dogi*


dogi:
**Review — rating: 74/100**

CI green on all three checks. Directionally consistent with the codebase's layering — UI shouldn't hold a repository — but as written this relocates the dependency without collecting any of the benefits that justify the move.

**The ViewModel is pure pass-through.** All seven new methods are `suspend fun x(…) = coursesRepository.x(…)`, and every call site still runs in `viewLifecycleOwner.lifecycleScope.launch` with results assigned to fragment fields (`currentCourseProgress`, `isNextStepLocked`). So:

- work is still scoped to the *view*, not the ViewModel — it's cancelled and restarted on rotation exactly as before
- state still lives in the fragment, so nothing survives a config change
- `TakeCourseViewModel` gains no state, no `StateFlow`, no `viewModelScope` usage

The net effect is that `CoursesRepository` is now reached through one more hop. That's a reasonable *intermediate* step if the follow-up moves these into `viewModelScope` with hoisted state — but it's worth saying so in the description, because on its own the 34 lines of delegation don't change behaviour or resilience.

**Two wrappers quietly narrow the repository's contract**

```kotlin
suspend fun getCurrentProgress(steps: List<CourseStep?>, userId: String?, courseId: String?): Int {
    return courseId?.let { coursesRepository.getCurrentProgress(steps, userId, it) } ?: 0
}

suspend fun isStepCompleted(stepId: String?, userId: String?): Boolean {
    return stepId?.let { coursesRepository.isStepCompleted(it, userId) } ?: false
}
```

Both repository methods already accept nullables and delegate onward (`CoursesRepositoryImpl.kt:550-556` → `progressRepository` / `submissionsRepository`), so the null case previously reached the repository and got whatever answer it gives. The wrappers now short-circuit to `0` / `false` without asking.

`isStepCompleted` is the reachable one — `stepId` comes from `steps.getOrNull(position - 1)?.id`, which is null at the boundary positions. The answer is very likely `false` either way, so I'd call this latent rather than broken, but it's an unforced divergence: a pass-through wrapper should mirror the signature it wraps and let the repository keep owning the null semantics.

**Nit:** stray double blank line before `logCourseVisit`.

**Verdict:** no defects that will bite today, but the PR as scoped is indirection rather than refactoring. Either finish the move — pull these into `viewModelScope` and expose state as `StateFlow` so rotation actually gets cheaper — or keep it as a mechanical step and mirror the repository's nullable signatures instead of narrowing them.

---

dogi:
**Re-review — Rating: 80/100** (was 74)

CI green on all three checks. The contract-narrowing I flagged is fixed — both wrappers now mirror the repository signatures instead of short-circuiting:

```kotlin
suspend fun getCurrentProgress(steps: List<CourseStep?>?, userId: String?, courseId: String?): Int =
    coursesRepository.getCurrentProgress(steps, userId, courseId)

suspend fun isStepCompleted(stepId: String?, userId: String?): Boolean =
    coursesRepository.isStepCompleted(stepId, userId)
```

Previously these returned `0` / `false` for null inputs without asking the repository, which quietly diverged from what the repository would have answered. Now the null semantics stay in one place. The stray double blank line is gone too.

**What remains is the shape, not a defect.** These are still seven pass-through methods, and the fragment still runs them in `viewLifecycleOwner.lifecycleScope` with results assigned to fragment fields — so nothing yet survives a configuration change. As a mechanical step that gets `CoursesRepository` out of the fragment and creates the seam for hoisting state later, that's fine; it just isn't the win the word "Refactor…to use TakeCourseViewModel" suggests on its own. Worth one line in the description saying it's the first half.

**Verdict:** the correctness concern is resolved and nothing here is wrong. Marking `ready` — the state-hoisting follow-up is where the actual benefit arrives.

---

Okuro3499 (approved):

PR #15657, opened 2026-08-13
login: smoother configurations repository io wrapping (fixes #15778)

### LINKED ISSUE #15778: refactor: Drop redundant withContext(io) around suspend Retrofit calls in ConfigurationsRepositoryImpl
Removed redundant `withContext(dispatcherProvider.io)` around suspend Retrofit calls in `ConfigurationsRepositoryImpl` since they natively dispatch off the caller thread. Kept the `withContext(dispatcherProvider.io)` for actual blocking file I/O operations (`Sha256Utils().getCheckSumFromFile(f)`).

dogi:
Removed redundant `withContext(dispatcherProvider.io)` around suspend Retrofit calls in `ConfigurationsRepositoryImpl` since they natively dispatch off the caller thread. Kept the `withContext(dispatcherProvider.io)` for actual blocking file I/O operations (`Sha256Utils().getCheckSumFromFile(f)`).

---
*PR created automatically by Jules for task [6006691664478212922](https://jules.google.com/task/6006691664478212922) started by @dogi*


dogi:
**Review — rating: 62/100**

CI green on all three checks. The premise is right — Retrofit's `suspend` functions already dispatch onto OkHttp's own executor, so wrapping them in `withContext(io)` buys nothing. But the removal takes the dispatcher away from work that genuinely needs it.

**`ResponseBody.string()` is blocking IO, and it's now unprotected**

All three endpoints return `Response<ResponseBody>`:

```kotlin
suspend fun healthAccess(@Url url: String?): Response<ResponseBody>
suspend fun getChecksum(@Url url: String?): Response<ResponseBody>
suspend fun isPlanetAvailable(@Url serverUrl: String?): Response<ResponseBody>
```

Retrofit does *not* buffer a raw `ResponseBody` — that's the whole point of the type. The suspend call resumes once the response is available, and `response.body()?.string()` then reads the stream, blocking whatever thread calls it. Previously that read sat inside `withContext(dispatcherProvider.io)`. Now it runs on the caller's dispatcher, in both `checkServerAvailability` and `checkCheckSum`.

At least one caller is on Main. `SyncActivity.isServerReachable` (`SyncActivity.kt:327-344`) calls `checkServerAvailability(url)` and then `startSync(type)` — a UI-touching call in an Activity, so this is running in a Main-dispatched coroutine. That makes the `.string()` read a blocking network read on the main thread: `NetworkOnMainThreadException` on a fresh connection, or an ANR at best. `DialogUtils.kt:178` calls `checkCheckSum` from a caller-supplied `scope` with the same exposure.

Worth noting you already applied the correct pattern one hunk away — `Sha256Utils().getCheckSumFromFile(f)` got its own narrow `withContext(dispatcherProvider.io)` precisely because it's blocking work that isn't a Retrofit call. `.string()` is in exactly that category and needs the same treatment:

```kotlin
val response = apiInterface.isPlanetAvailable(url)   // fine bare — Retrofit dispatches
val ss = withContext(dispatcherProvider.io) { response.body()?.string() }
```

`healthAccess` is safe as changed — nothing reads its body, only `response.code()`.

**Minor:** `catch (e: Exception)` in `checkServerAvailability` binds `e` and never uses it. Pre-existing, and it swallows genuine failures indistinguishably from "server said no", but since you're in the function it's a cheap `catch (_: Exception)` or a log line.

**Verdict:** the `withContext` removals around the Retrofit calls themselves are correct and worth landing. Put a narrow `withContext(dispatcherProvider.io)` back around the two `.string()` reads — same shape you used for the checksum — and this is ready.

---

dogi:

CI green on all three checks. The blocking-read problem is fixed exactly where it needed to be.

```kotlin
val response = apiInterface.isPlanetAvailable(url)          // bare — Retrofit dispatches itself
val ss = withContext(dispatcherProvider.io) { response.body()?.string() }   // wrapped — real blocking read
```

Both `checkServerAvailability` and `checkCheckSum` now keep the narrow `withContext` around `.string()` while leaving the suspend Retrofit calls unwrapped, which is the distinction the original version missed. `healthAccess` correctly stays bare — only `response.code()` is read there, no body.

The `Sha256Utils().getCheckSumFromFile(f)` wrapper is retained, so the file hash still runs off the caller's thread too.

`catch (_: Exception)` also picks up the unused-binding nit. Worth noting the underlying concern remains — a swallowed exception makes "server unreachable" and "request threw" indistinguishable — but that's pre-existing and out of scope here.

This now reads the same way #15682 does, which is the reference for this pattern: Retrofit calls bare, body reads and file I/O on a dispatcher.

**Verdict:** correct on both axes. Marking `ready`.

---

Okuro3499 (approved):

PR #15661, opened 2026-08-13
all: smoother gson injecting (fixes #15801)

### LINKED ISSUE #15801: Inject Hilt-provided Gson into repositories
This submission introduces a `@DefaultGson` qualifier to dependency inject `Gson` instances into repositories. It targets `HealthRepositoryImpl`, `ChatRepositoryImpl`, `SubmissionsRepositoryImpl`, `ConfigurationsRepositoryImpl`, and `VoicesRepositoryImpl` to remove their direct coupling with the singleton `JsonUtils.gson`. This enhances testability and ensures we maintain bit-identical configuration compared to the previously used singleton. Associated unit tests have been updated, and all tests are passing.

dogi:
This submission introduces a `@DefaultGson` qualifier to dependency inject `Gson` instances into repositories. It targets `HealthRepositoryImpl`, `ChatRepositoryImpl`, `SubmissionsRepositoryImpl`, `ConfigurationsRepositoryImpl`, and `VoicesRepositoryImpl` to remove their direct coupling with the singleton `JsonUtils.gson`. This enhances testability and ensures we maintain bit-identical configuration compared to the previously used singleton. Associated unit tests have been updated, and all tests are passing.

---
*PR created automatically by Jules for task [5566190821418426847](https://jules.google.com/task/5566190821418426847) started by @dogi*

## Summary by CodeRabbit

* **New Features**
  * Added health-profile retrieval and updates.
  * Improved handling of pending photo and news-log uploads.
  * Added automatic server URL updates when an alternative is available.

* **Bug Fixes**
  * Improved bulk survey submission processing, including duplicate prevention and malformed data handling.
  * Improved reliability when sharing, deleting, importing, and syncing news.
  * Standardized JSON processing across chat, configuration, health, submissions, and voice features.

* **Tests**
  * Updated automated coverage for the affected workflows.


dogi:
**Review — rating: 70/100**

CI green on all three checks. The central judgement call here is right, and it's the one most people would get wrong — but one file gets caught by it.

**The important thing done correctly:** `JsonUtils.gson` is a bare `Gson()` (`JsonUtils.kt:12-14`), while `NetworkModule`'s existing unqualified provider is configured:

```kotlin
GsonBuilder()
    .excludeFieldsWithModifiers(Modifier.FINAL, Modifier.TRANSIENT, Modifier.STATIC)
    .serializeNulls()
    .create()
```

Injecting *that* one would have silently changed serialization at every converted site — nulls appearing in payloads, final/transient fields dropping out. Adding a separate `@DefaultGson` qualifier bound to a plain `Gson()` keeps the swap byte-for-byte equivalent. That's the correct call and it's clearly deliberate.

**But `VoicesRepositoryImpl` didn't use the static — it already injected the configured Gson**

```diff
-    private val gson: Gson,
+    @DefaultGson private val gson: Gson,
```

That parameter was unqualified, so it resolved to the `serializeNulls()` instance, and there are **six pre-existing call sites using it** — `VoicesRepositoryImpl.kt` lines 67, 144, 195, 213, 241, 257 — none of which were `JsonUtils.gson`. Adding the qualifier repoints all six to the plain `Gson()`.

The observable difference is `serializeNulls`: `JsonObject` members holding `JsonNull` are written as `"key": null` under the configured instance and **skipped entirely** under a plain one. Lines 213 and 257 write `news.viewIn`, and line 67 writes `news.images` — values that get persisted and later uploaded, so a dropped key is a real payload change. (The `JsonArray` cases are less exposed, since array elements aren't subject to the same skipping; it's object members that vanish.) `excludeFieldsWithModifiers` doesn't apply here either way, since these are Gson tree types rather than reflected POJOs.

Whether it actually bites depends on whether these documents carry explicit nulls — but it's an unintended change either way. The PR's purpose is retiring a static, not switching which configured Gson an already-injected field resolves to, and nothing in the description flags it.

Cleanest fix: leave that constructor parameter unqualified and add a second `@DefaultGson` field for the sites that were previously `JsonUtils.gson`. Needing two Gsons in one class is itself a signal that the two configurations deserve reconciling — but that's a bigger question than this PR.

**Naming makes this trap likelier.** `@DefaultGson` labels the *plain* instance "default" while the unqualified binding — the one Dagger actually treats as the default — is the configured one. `@PlainGson` or `@UnconfiguredGson` would put the surprising instance behind the surprising name.

**Scope note:** `JsonUtils.gson` still has callers in `Converters.kt` (6), `BaseVoicesFragment`, `ChatApiService`, `LoginSyncManager`, and several models, so the static isn't retired yet. Fine as an incremental step — worth saying so, since the title reads like a completed migration.

**Nit:** `import org.ole.planet.myplanet.di.DefaultGson` is placed after the `model.*` imports in both `ChatRepositoryImpl` and `ConfigurationsRepositoryImpl` — out of alphabetical order.

**Verdict:** four of the five files are clean, behaviour-preserving, and test-updated. Revert the qualifier on `VoicesRepositoryImpl`'s existing parameter (or confirm and document that those six sites should change), then this is ready.

---

dogi:
**Review — rating: 76/100**

The substitution is safe, and I verified the part that could have silently broken uploads.

`JsonUtils.gson` is `Gson()` — a plain instance with default settings. `providePlainGson()` is also `Gson()`. Identical configuration, so all ~20 call-site swaps are behavior-preserving.

**The important thing this PR gets right** is *not* reusing the existing binding. `NetworkModule` already provides an unqualified `Gson` built with `.serializeNulls()`. Injecting that instead would have started emitting explicit `null` members into chat requests, submission payloads, and news documents — a wire-format change with no visible local symptom. `VoicesRepositoryImpl` is the proof this was deliberate: it now takes *both*, leaves every pre-existing `gson` use alone, and converts only the `JsonUtils.gson` sites to `plainGson`.

Also good: each of the five touched files is migrated to zero remaining `JsonUtils.gson` references. Half-converted files are worse than unconverted ones, and this avoids that.

**Blocking — import ordering is broken in five of the six production files:**

- `ChatRepositoryImpl` — `di.PlainGson` placed after `model.News`
- `ConfigurationsRepositoryImpl` — `di.PlainGson` after `model.MyPlanet`
- `HealthRepositoryImpl` — `di.PlainGson` after `utils.AndroidDecrypter`
- `SubmissionsRepositoryImpl` — `com.google.gson.Gson` after `com.google.gson.JsonObject`, and `di.PlainGson` after `model.UserEntity`
- `VoicesRepositoryImpl` — `di.PlainGson` inserted *between* `model.TeamNotification` and `model.UserEntity`, splitting the model block

Every one of these should sort ahead of where it landed. The last is the worst — it breaks up a contiguous group. This is mechanical to fix and the repo maintains ordered imports everywhere else.

**Design concern worth addressing before this pattern spreads.** The qualifier is on the wrong binding. Right now:

```kotlin
@Provides @Singleton fun provideGson(): Gson = GsonBuilder().serializeNulls().create()   // unqualified
@Provides @Singleton @PlainGson fun providePlainGson(): Gson = Gson()                    // qualified
```

The *default* injection point gets the *non-default* Gson. Anyone who writes `private val gson: Gson` in a new class gets `serializeNulls` without knowing it, which is precisely the trap this PR carefully sidestepped in `VoicesRepositoryImpl`. Qualifying both — `@SerializeNullsGson` alongside `@PlainGson`, with no unqualified binding — makes every injection site state which semantics it wants, and turns a silent wrong-default into a compile error. That's a bigger change than this PR, but it's the shape this is heading toward and it's cheaper to fix at two call sites than at twenty.

Relatedly: `VoicesRepositoryImpl`'s constructor now has `gson` and `plainGson` adjacent, distinguished only by name. Renaming the existing one to `serializeNullsGson` would make a future mix-up much harder.

**Scope note:** the title says "into repositories", but this converts 5 of the 12 repositories that use `JsonUtils.gson`, and about 20 of 101 references app-wide. The commit message says "incrementally", which is the accurate framing — worth moving that word into the title or description so reviewers don't read it as a completed migration.

**Merge-ordering warning:** this collides with **#15630** on four files — `HealthRepositoryImpl`, `SubmissionsRepositoryImpl`, and both of their test files. #15630 replaces `userDao` with `Lazy<UserRepository>` in the same two constructors and edits the same test setup blocks. No semantic conflict, but whichever lands second needs a rebase. #15630 is currently `ready`; landing it first and rebasing this one is the cheaper order.

**Verdict:** the substitution itself is correct and the care around the `serializeNulls` binding is exactly right. Fix the imports — that alone is what's holding this at `change` — and it's `ready`. The qualifier inversion is worth a follow-up issue rather than blocking here.

---

dogi:

All six import placements fixed — that was the only blocking item.

- `ChatRepositoryImpl` — `di.PlainGson` now ahead of `model.AiProvider`
- `ConfigurationsRepositoryImpl` — ahead of `model.MyPlanet`
- `HealthRepositoryImpl` — ahead of `model.HealthExamination`
- `SubmissionsRepositoryImpl` — `com.google.gson.Gson` ahead of `JsonArray`, and `di.PlainGson` ahead of `model.Answer`
- `VoicesRepositoryImpl` — `di.PlainGson` ahead of `model.MyLibrary`, so the model block is contiguous again

The delta is imports only, so the substance I verified last round is untouched: `@PlainGson` provides a bare `Gson()`, identical to `JsonUtils.gson`, making every call-site swap behavior-preserving; the existing unqualified `serializeNulls` binding is deliberately *not* reused, which is what keeps chat requests, submission payloads, and news documents on the wire format they had; and all five touched files are migrated to zero remaining `JsonUtils.gson` references rather than left half-converted. CI green on all three checks.

**Carried forward as follow-ups, none blocking:**

- The qualifier is still on the wrong binding — the unqualified `Gson` is the `serializeNulls` one, so a new class injecting a bare `Gson` silently gets non-default semantics. Qualifying both (`@SerializeNullsGson` + `@PlainGson`, no unqualified binding) turns that into a compile error instead of a wire-format surprise. Worth its own issue.
- `VoicesRepositoryImpl` still has `gson` and `plainGson` adjacent in the constructor, distinguished only by name. Renaming the first to `serializeNullsGson` would make a future mix-up hard.
- Title still reads as a completed migration; this is 5 of 12 repositories and ~20 of 101 references. The commit message's "incrementally" is the accurate word.

**Merge ordering unchanged:** collides with **#15630** on `HealthRepositoryImpl`, `SubmissionsRepositoryImpl`, and both test files. #15630 is `ready` and swaps `userDao` for `Lazy<UserRepository>` in the same two constructors — landing it first and rebasing this is still the cheaper order.

**Verdict:** `ready`.

---

Okuro3499 (approved):

PR #15664, opened 2026-08-13
sync: less manager courses repository is more (fixes #15802)

### LINKED ISSUE #15802: 🧹 Remove unused coursesRepository property from SyncManager
🎯 **What:** The code health issue addressed
Removed the unused private property `coursesRepository` from `SyncManager.kt` and updated the `SyncManagerTest` instantiations to reflect this change.

💡 **Why:** How this improves maintainability
Removing unused dependencies from classes, particularly those using Dependency Injection, clears up the constructor signature, removes unnecessary imports, and prevents the creation of dependencies that aren't being used by the class.

✅ **Verification:** How you confirmed the change is safe
Ran `./gradlew testDefaultDebugUnitTest` to ensure that `SyncManag

dogi:
🎯 **What:** The code health issue addressed
Removed the unused private property `coursesRepository` from `SyncManager.kt` and updated the `SyncManagerTest` instantiations to reflect this change.

💡 **Why:** How this improves maintainability
Removing unused dependencies from classes, particularly those using Dependency Injection, clears up the constructor signature, removes unnecessary imports, and prevents the creation of dependencies that aren't being used by the class.

✅ **Verification:** How you confirmed the change is safe
Ran `./gradlew testDefaultDebugUnitTest` to ensure that `SyncManagerTest` and the rest of the application compiled successfully and passed existing tests. Verified via code review that no functional changes were made.

✨ **Result:** The improvement achieved
A cleaner `SyncManager` class without unneeded dependencies, making the code more readable and reducing the overhead in its instantiation.

---
*PR created automatically by Jules for task [12554540692289871793](https://jules.google.com/task/12554540692289871793) started by @dogi*


dogi:
**Review — rating: 86/100**

CI green on all three checks. Clean removal — `SyncManager` never referenced `coursesRepository`, the import goes with it, and `SyncManagerTest` drops the corresponding mock.

I checked the one way a change like this can bite: removing an injection point can stop a `@Singleton` from ever being constructed, so any side effect in its `init` would silently disappear. `CoursesRepositoryImpl` has no `init` block, and it's injected in plenty of other places, so nothing is lost beyond the unused field.

**One process note:** #15665 and #15671 also edit `SyncManager`'s constructor and `SyncManagerTest`, each removing a different property. All three are correct in isolation, but as separate PRs against the same parameter list they'll conflict pairwise — whichever lands first forces the other two to rebase and re-run CI. One PR removing all three unused repositories would be the same net change with a third of the churn.

---

Okuro3499 (approved):

PR #15665, opened 2026-08-13
sync: less manager teams repository is more (fixes #15755)

dogi:
🎯 **What:** Removed the unused `teamsRepository` private dependency from `SyncManager.kt` and its associated import. Also updated `SyncManagerTest.kt` to reflect this change in the constructor invocation.
💡 **Why:** This improves the maintainability and readability of the codebase by getting rid of dead code and unneeded dependencies inside a Hilt-injected class.
✅ **Verification:** Ran `SyncManagerTest` unit tests using `./gradlew testDefaultDebugUnitTest` to confirm the change does not break compilation and all relevant tests pass.
✨ **Result:** A cleaner constructor and reduced boilerplate in the `SyncManager` and its tests.

---
*PR created automatically by Jules for task [11086869338243385648](https://jules.google.com/task/11086869338243385648) started by @dogi*

PR #15659, opened 2026-08-13
sync: smoother upload repository api routing (fixes #15779)

dogi:
This change completes the routing of direct `ApiInterface` network calls in the upload flow (`UploadManager`, `PhotoUploader`, `AchievementUploader`) through the `UploadRepository`. This centralizes data operations, improves testability by allowing interface mocking rather than raw network interfaces, and brings the uploaders in line with the project's repository pattern. Existing tests were updated to reflect the new constructor signatures and mock behavior.

---
*PR created automatically by Jules for task [4155697220269861960](https://jules.google.com/task/4155697220269861960) started by @dogi*

## Summary by CodeRabbit

* **Improvements**
  * Streamlined resource uploads across team images, news attachments, achievements, CVs, and photos.
  * Upload operations now provide consistent request handling and JSON responses.
  * Existing upload workflows and error handling remain unchanged.

* **Tests**
  * Updated upload coverage to reflect the improved upload flow and verify successful photo submissions.

PR #15666, opened 2026-08-13
🧹 [Code Health] Remove unused `apiInterface` from `UploadToShelfService`

dogi:
🎯 **What:** The code health issue addressed was the presence of an unused private property, `apiInterface`, in the `UploadToShelfService` constructor.
💡 **Why:** Removing unused code improves maintainability and readability by reducing clutter. By removing this dependency, `UploadToShelfService` becomes easier to understand, test, and inject since it requires one less parameter. It simplifies the constructor and removes an unneeded import.
✅ **Verification:** I confirmed the change is safe by updating the corresponding dependency injection module (`ServiceModule.kt`) and unit test (`UploadToShelfServiceTest.kt`) to align with the new constructor signature. I ran the full test suite (`./gradlew testDefaultDebugUnitTest`) to ensure no functionality is broken.
✨ **Result:** The codebase is cleaner, and the redundant dependency injection parameter is gone.

---
*PR created automatically by Jules for task [2326338919881779006](https://jules.google.com/task/2326338919881779006) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/di/ServiceModule.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/di/ServiceModule.kt
@@ -60,10 +60,9 @@ object ServiceModule {
         userSyncRepository: UserSyncRepository,
         healthRepository: HealthRepository,
         @ApplicationScope appScope: CoroutineScope,
-        dispatcherProvider: DispatcherProvider,
-        apiInterface: ApiInterface
+        dispatcherProvider: DispatcherProvider
     ): UploadToShelfService {
-        return UploadToShelfService(context, preferences, sharedPrefManager, userRepository, userSyncRepository, healthRepository, appScope, dispatcherProvider, apiInterface)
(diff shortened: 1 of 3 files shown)


dogi:
**Review — rating: 84/100 (duplicate)**

CI green on all three checks, and the change itself is correct — `UploadToShelfService` genuinely doesn't use `apiInterface`, the `ServiceModule` provider is updated to match, and the test is adjusted alongside.

The problem is that **#15615 already does exactly this**. I diffed the two branches against the current base and they are byte-identical — same three files, same hunks, no difference at all. #15615 is the lower number and is already labelled `ready` from an earlier review pass.

**Recommend closing this in favour of #15615.** Nothing is lost by doing so; that PR carries the identical change and is further along.

Worth noting for whatever is generating this batch: #15664, #15665 and #15671 all edit `SyncManager`'s constructor, and #15667, #15668, #15669 all edit `UploadManager`'s constructor (as does #15615's sibling #15659, already `ready`). Those aren't duplicates of each other — each removes a different property — but as separate PRs against the same parameter list they'll conflict pairwise on rebase, and the last few will need re-running CI for no reason. Grouping per-file would land the same cleanup in two PRs instead of seven.

---

PR #15670, opened 2026-08-13
sync: less transaction sync manager application scope is more (fixes #15803)

### LINKED ISSUE #15803: 🧹 Remove unused `applicationScope` from `TransactionSyncManager`
🎯 **What:** Removed the unused private property `applicationScope` from `TransactionSyncManager`.
💡 **Why:** The variable `@ApplicationScope private val applicationScope: CoroutineScope` was injected via constructor but was never used inside the class. Removing it makes the code cleaner, avoids an unnecessary dependency, and improves maintainability.
✅ **Verification:** Updated the corresponding Dagger Hilt provider in `ServiceModule.kt` and test invocations. Verified via full local `./gradlew test` (targeting `org.ole.planet.myplanet.services.sync.*`) that there are no compilation errors and

dogi:
🎯 **What:** Removed the unused private property `applicationScope` from `TransactionSyncManager`.
💡 **Why:** The variable `@ApplicationScope private val applicationScope: CoroutineScope` was injected via constructor but was never used inside the class. Removing it makes the code cleaner, avoids an unnecessary dependency, and improves maintainability.
✅ **Verification:** Updated the corresponding Dagger Hilt provider in `ServiceModule.kt` and test invocations. Verified via full local `./gradlew test` (targeting `org.ole.planet.myplanet.services.sync.*`) that there are no compilation errors and no broken tests.
✨ **Result:** A cleaner constructor with no unused property overhead, ensuring dependency injection is strictly necessary and improving code health.

---
*PR created automatically by Jules for task [13128348686959656874](https://jules.google.com/task/13128348686959656874) started by @dogi*

## Summary by CodeRabbit

* **Release**
  * Updated the Android app version to 0.65.22 (build 6522).

* **Refactor**
  * Improved transaction synchronization setup and lifecycle handling.
  * No changes to synchronization behavior are expected.


dogi:
**Review — rating: 84/100**

CI green on all three checks. Correct removal, and this one carried slightly more risk than its siblings because `TransactionSyncManager` isn't constructor-injected — `ServiceModule` builds it by hand with 23 positional arguments, so dropping a parameter means editing that call site too.

I checked the argument alignment, since a positional shift there would compile fine if two neighbouring parameters shared a type. The removed `scope: CoroutineScope` sat between `surveysRepository` and `dispatcherProvider`, and `CoroutineScope` / `DispatcherProvider` are unrelated types, so any misalignment would have failed the build rather than silently rewiring dependencies. The remaining 22 arguments line up.

Also worth noting: dropping this injection doesn't destroy the application scope. `@ApplicationScope CoroutineScope` is provided by `DispatcherModule` and injected in many other places, so no in-flight work is affected — `TransactionSyncManager` simply never used it. Both of its test files are updated in step.

**The real finding is what this exposes:** a 23-argument constructor built by hand in a DI module. Every dependency change forces an edit to a positional call that only the type checker is guarding, and it'll keep growing. `TransactionSyncManager` is a strong candidate for `@Inject constructor` (removing the `ServiceModule` provider entirely), or for grouping its ~20 repositories behind a small number of cohesive interfaces. Out of scope here, but this PR is the third or fourth to touch that call site.

---

Okuro3499 (approved):

PR #15674, opened 2026-08-13
sync: less upload shelf service constructor is more (fixes #15805)

### LINKED ISSUE #15805: 🧹 Remove unused constructor properties in UploadToShelfService
🎯 **What:** Removed unused `sharedPrefManager` and `apiInterface` dependencies from the `UploadToShelfService` class, its corresponding Dagger module (`ServiceModule`), and unit tests.
💡 **Why:** These dependencies were not being utilized in the class, meaning they were safely removable without changing behavior. Their removal clarifies what dependencies the class actually requires and slims down the DI graph.
✅ **Verification:** Verified by a local Gradle build (`assembleDebug`), as well as successfully running unit tests (`testDebugUnitTest`), to ensure there are no compilation errors or reg

dogi:
🎯 **What:** Removed unused `sharedPrefManager` and `apiInterface` dependencies from the `UploadToShelfService` class, its corresponding Dagger module (`ServiceModule`), and unit tests.
💡 **Why:** These dependencies were not being utilized in the class, meaning they were safely removable without changing behavior. Their removal clarifies what dependencies the class actually requires and slims down the DI graph.
✅ **Verification:** Verified by a local Gradle build (`assembleDebug`), as well as successfully running unit tests (`testDebugUnitTest`), to ensure there are no compilation errors or regressions.
✨ **Result:** Improved maintainability, cleaner code, and accurate constructor parameters for `UploadToShelfService`.

---
*PR created automatically by Jules for task [12947670196855048810](https://jules.google.com/task/12947670196855048810) started by @dogi*


dogi:
**Review — rating: 84/100**

CI green on all three checks. The removals are correct — neither `sharedPrefManager` nor `apiInterface` was referenced in `UploadToShelfService`, the `ServiceModule` provider signature and its construction call are updated together, and `UploadToShelfServiceTest` drops both mocks.

**This collides with #15615, which is already `ready`**

Both PRs touch the same three files, and both remove `apiInterface` from `UploadToShelfService` and its provider. This one additionally removes `sharedPrefManager`, so it's a strict superset — #15615's entire diff is contained in it.

They cannot both land: whichever merges first leaves the other with a conflicting hunk against a parameter list that no longer matches. My recommendation is to **keep this one and close #15615**, since it's the more complete cleanup and closing the subset costs nothing. That's the opposite of the usual "lower number wins" tiebreak, but only because one genuinely subsumes the other rather than merely overlapping.

Worth noting #15666 was already closed as a duplicate of #15615 earlier — so this is the third PR generated against the same unused `apiInterface` field. Whatever is producing these batches doesn't appear to be checking the open queue before filing, which is also how #15664/#15665/#15671/#15672 ended up as four separate PRs against one constructor.

**On the `ServiceModule` call site:** the provider builds `UploadToShelfService` positionally, so removing two parameters means the remaining seven arguments have to stay aligned. I checked — `sharedPrefManager` sat between `preferences` and `userRepository`, and `apiInterface` was last; the surrounding types are all distinct, so a misalignment would have failed the build rather than silently rewiring. It's fine, but this is the same fragility #15670 ran into with `TransactionSyncManager`: hand-written positional construction in a DI module that every dependency change has to edit. `@Inject constructor` would remove the whole category.

**Verdict:** correct and safe on its own merits. Marking `ready`, with the recommendation that it supersede #15615 rather than both being merged.

---

Okuro3499 (approved):

PR #15676, opened 2026-08-13
all: less dialog utils indeterminate is more (fixes #15762)

### LINKED ISSUE #15762: 🧹 Remove unused `setIndeterminate` from `DialogUtils.kt`
🎯 **What:** Removed the unused private function `setIndeterminate` from `DialogUtils.kt`.
💡 **Why:** The function was unused dead code that cluttered the utility class. Removing it improves maintainability and readability without affecting existing behavior.
✅ **Verification:** Verified that tests pass via Gradle (`./gradlew testDefaultDebugUnitTest`) and that the file compiles successfully. No existing functionality is broken.
✨ **Result:** Improved code health by removing dead code.

dogi:
🎯 **What:** Removed the unused private function `setIndeterminate` from `DialogUtils.kt`.
💡 **Why:** The function was unused dead code that cluttered the utility class. Removing it improves maintainability and readability without affecting existing behavior.
✅ **Verification:** Verified that tests pass via Gradle (`./gradlew testDefaultDebugUnitTest`) and that the file compiles successfully. No existing functionality is broken.
✨ **Result:** Improved code health by removing dead code.

---
*PR created automatically by Jules for task [8005285338405452722](https://jules.google.com/task/8005285338405452722) started by @dogi*


dogi:
**Review — rating: 86/100**

CI green on all three checks. Genuinely dead code, and I checked the one thing that would have made deleting it the wrong move.

The concern with removing an unused private helper is that it might be unused *by mistake* — that someone meant to call it and the missing call is the real bug, in which case deleting it cements the defect. That's not the case here. `DialogUtils` sets `progressBar.isIndeterminate` in three other places: inline at lines 257 and 259, and through the public `setIndeterminateMode(Boolean)` at line 277, which `BaseResourceFragment.kt:83` calls and `BaseResourceFragmentTest` covers. The determinate path is fully served without this helper.

Also worth noting it was misnamed — `setIndeterminate()` set `isIndeterminate = false`, i.e. it made the bar *determinate*. A private no-arg method whose name asserts the opposite of what it does is exactly the kind of thing that causes a mistaken call later, so deleting it is a small readability win beyond the line count.

Only main-source change in the batch that isn't a constructor parameter, so no conflicts with its siblings.

---

Okuro3499 (approved):

PR #15679, opened 2026-08-13
sync: less transaction manager teams repository is more (fixes #15764)

### LINKED ISSUE #15764: 🧹 [code health improvement] Remove unused teamsRepository from TransactionSyncManager
🎯 **What:** Removed the unused `teamsRepository` private property from the `TransactionSyncManager` constructor and updated associated tests and DI modules.
💡 **Why:** `teamsRepository` was not being used within `TransactionSyncManager`. Removing it cleans up the constructor, simplifies dependency injection, and improves overall code readability and maintainability.
✅ **Verification:** Ran the test suite (`./gradlew testDefaultDebugUnitTest --tests "org.ole.planet.myplanet.services.sync.*"`) which passed successfully, confirming that this removal breaks no existing logic or tests.
✨ **Result:*

dogi:
🎯 **What:** Removed the unused `teamsRepository` private property from the `TransactionSyncManager` constructor and updated associated tests and DI modules.
💡 **Why:** `teamsRepository` was not being used within `TransactionSyncManager`. Removing it cleans up the constructor, simplifies dependency injection, and improves overall code readability and maintainability.
✅ **Verification:** Ran the test suite (`./gradlew testDefaultDebugUnitTest --tests "org.ole.planet.myplanet.services.sync.*"`) which passed successfully, confirming that this removal breaks no existing logic or tests.
✨ **Result:** A cleaner constructor with fewer dependencies for `TransactionSyncManager`.

---
*PR created automatically by Jules for task [4396016470013737949](https://jules.google.com/task/4396016470013737949) started by @dogi*


dogi:
**Review — rating: 84/100**

CI green on all three checks. The removal is correct — `teamsRepository` was unreferenced in `TransactionSyncManager`, and both test files are updated.

**Direct conflict with #15670, which is already `ready`**

Both PRs rewrite the same `return TransactionSyncManager(...)` line in `ServiceModule`. This branch's "before" text still contains `scope`, which #15670 removes — so it was branched from a base without that change, and the two hunks overlap exactly. Whichever merges first, the other needs a rebase; they cannot both apply cleanly.

Since #15670 is already approved and removes a different parameter, the low-friction path is to land it first and rebase this one onto it — or fold both removals into a single commit against that call site.

**On the positional call:** dropping `teamsRepository` shifts everything after it, and `ServiceModule` constructs this class by hand rather than via `@Inject constructor`. I checked the alignment — `teamsRepository: Lazy<TeamsRepository>` sat between `activitiesRepository` and `teamsSyncRepository`, and the neighbouring types are distinct, so a misalignment would fail the build rather than silently rewire dependencies. Safe as written.

But this is now the second PR in two batches to edit that call, and the pattern is the finding: a hand-written 20+ argument construction in a DI module where the type checker is the only thing guarding argument order. Two adjacent parameters of the same type would make a silent mis-wiring possible, and the list is long enough that it's plausible. `TransactionSyncManager` wants `@Inject constructor` with the `ServiceModule` provider deleted — that removes the conflict class entirely, not just this instance of it.

`TeamsRepositoryImpl` has no `init` block and keeps other injectors, so nothing construction-time is stranded.

**Verdict:** correct and safe. Marking `ready`, with the note that it needs a rebase after #15670 (or should merge with it).

---

Okuro3499 (approved):

PR #15680, opened 2026-08-13
courses: smoother removed log dao deleting (fixes #15780)

### LINKED ISSUE #15780: ⚡ perf: optimize removed log deletion using batched transaction
💡 **What:** Replaced chunked database deletions loop in `CoursesRepositoryImpl` with a single `@Transaction` method added to `RemovedLogDao`.

🎯 **Why:** To prevent an N+1 query performance bottleneck caused by creating a new database transaction for every chunked delete when syncing large sets of courses or resources. It respects the SQLite 999 parameter limit natively while grouping execution to one transaction.

📊 **Measured Improvement:** In benchmark tests simulating deleting 10,000 logs:
- Baseline (chunking + N separate transactions): 7 ms
- Optimized (transactional chunking inside DAO)

dogi:
💡 **What:** Replaced chunked database deletions loop in `CoursesRepositoryImpl` with a single `@Transaction` method added to `RemovedLogDao`.

🎯 **Why:** To prevent an N+1 query performance bottleneck caused by creating a new database transaction for every chunked delete when syncing large sets of courses or resources. It respects the SQLite 999 parameter limit natively while grouping execution to one transaction.

📊 **Measured Improvement:** In benchmark tests simulating deleting 10,000 logs:
- Baseline (chunking + N separate transactions): 7 ms
- Optimized (transactional chunking inside DAO): 0 ms (Too fast for `measureTimeMillis`)

---
*PR created automatically by Jules for task [302577740642398329](https://jules.google.com/task/302577740642398329) started by @dogi*


dogi:
**Review — rating: 72/100**

CI green on all three checks. Two genuine wins here, and one number that's still wrong.

**The best part isn't in the title.** `ResourcesRepositoryImpl.kt:422` previously called `deleteByTypeUserAndDocs("resources", userId, resourceIds)` with **no chunking at all** — an unbounded id list going straight into an `IN (...)` clause. Routing it through the helper fixes a real crash waiting to happen on a large resource set. Worth calling out in the PR description, because it's a bug fix rather than an optimization.

Wrapping the chunk loop in a single transaction is also a real improvement: N chunked `DELETE`s become one commit instead of N, so one fsync rather than several.

**The chunk size is still over the limit**

```kotlin
@Query("DELETE FROM removed_log WHERE type = :type AND userId = :userId AND docId IN (:docIds)")
```

That's `docIds.size + 2` bound parameters — `type` and `userId` are host variables too. `SQLITE_MAX_VARIABLE_NUMBER` is **999** on Android API 26–30 (it only rises to 32766 with the SQLite bundled from API 31), and minSdk here is 26. So:

- old `chunked(1000)` → 1002 variables — over
- new `chunked(999)` → 1001 variables — **still over**

Moving 1000 → 999 reads like an attempt to respect the limit, but it lands one short of fixing it. The safe value is **997**, or better, follow the precedent already in this codebase: `ResourcesRepositoryImpl.kt:535` and `:574` both use `chunked(900)`, which leaves comfortable headroom without needing anyone to recount host variables when a `WHERE` clause grows.

Note this also means the pre-existing `CoursesRepositoryImpl` code was already capable of throwing `too many SQL variables` on an older device with >1000 removed course ids — so tightening it properly is worth doing, not just tidier.

**Use an import, not a fully-qualified annotation**

```kotlin
+    @androidx.room.Transaction
```

`RemovedLogDao` already imports `androidx.room.Dao`, `Insert`, `OnConflictStrategy`, and `Query` — add `Transaction` to that list and write `@Transaction`. Inline FQNs shouldn't appear in this codebase; the import block is right there and the annotation sits oddly beside its four short-named siblings.

**Minor:** `deleteByTypeUserAndDocsTransaction` puts the implementation mechanism in the name. `deleteByTypeUserAndDocsChunked` describes what a caller actually cares about — or just keep the transactional version as the only public entry point and make the raw one private.

**Verdict:** the transaction batching and the `ResourcesRepositoryImpl` fix are both worth landing. Change 999 → 997 (or 900, matching the sibling file) and swap the FQN annotation for an import, then this is ready.

---

dogi:

CI green on all three checks. Both findings addressed, and the chunk size went to the better of the two options I offered.

`import androidx.room.Transaction` is in place beside its four `androidx.room` siblings, and the annotation reads `@Transaction`. `chunked(900)` matches the precedent already set in `ResourcesRepositoryImpl.kt:535` and `:574` — 902 bound variables against SQLite's 999 ceiling, with enough headroom that nobody has to recount host variables if the `WHERE` clause grows a column. Picking the existing convention over a hand-computed 997 is the more durable call.

`deleteByTypeUserAndDocsChunked` also drops the implementation-mechanism suffix in favour of the property callers care about.

The substance was already good and is unchanged: the transaction turns N chunked deletes into one commit, and routing `ResourcesRepositoryImpl.kt:422` through the helper fixes a genuinely unbounded `IN (...)` that had no chunking at all — still the most valuable line in the diff and still worth a mention in the description.

**Verdict:** clean. Marking `ready`.

---

Okuro3499 (approved):

PR #15681, opened 2026-08-13
all: smoother user repository name unifying (fixes #15781)

### LINKED ISSUE #15781: ⚡ Optimize deduplication memory consumption in UserRepository
💡 **What:** Replaced the `groupBy` logic in `UserRepositoryImpl.cleanupDuplicateUsers` with a custom deduplication map strategy that tracks seen users and only allocates lists for duplicate names.

🎯 **Why:** `allUsers.groupBy { it.name }` allocates a new `List` for every unique user in the database, even if they aren't duplicates. When cleaning up thousands of users, this results in significant unnecessary heap allocations and garbage collection overhead.

📊 **Measured Improvement:** 
I created a benchmark script allocating 100,000 unique users and 2,000 duplicates.
**Baseline (Old Logic):**

dogi:
💡 **What:** Replaced the `groupBy` logic in `UserRepositoryImpl.cleanupDuplicateUsers` with a custom deduplication map strategy that tracks seen users and only allocates lists for duplicate names.

🎯 **Why:** `allUsers.groupBy { it.name }` allocates a new `List` for every unique user in the database, even if they aren't duplicates. When cleaning up thousands of users, this results in significant unnecessary heap allocations and garbage collection overhead.

📊 **Measured Improvement:** 
I created a benchmark script allocating 100,000 unique users and 2,000 duplicates.
**Baseline (Old Logic):** 532 ms (total across 5 iterations)
**Optimized (New Logic):** 136 ms (total across 5 iterations)
This demonstrates a ~74% reduction in processing time alongside the unmeasured but substantial reduction in memory allocations.

---
*PR created automatically by Jules for task [12659101296967493745](https://jules.google.com/task/12659101296967493745) started by @dogi*


dogi:
**Review — rating: 68/100**

CI green on all three checks. The rewrite is behaviourally correct — I traced it — but it optimizes the small half of the problem at a real cost in readability, and it reaches for fully-qualified names that are already imported.

**Correctness holds.** Walking the grouping by hand: first user for a name goes into `seenUsers` and `put` returns null, so nothing else happens; the second triggers `put` returning the first, and `duplicates[name] = mutableListOf(existing, user)`; the third onward hits the `duplicateList != null` branch and appends. Group membership and ordering come out identical to `groupBy`, so the subsequent sort and `drop(1)` delete exactly the same rows as before. No issue there.

**Use the import — `UserEntity` is already imported in this file**

```kotlin
val seenUsers = HashMap<String?, org.ole.planet.myplanet.model.UserEntity>()
val duplicates = HashMap<String?, MutableList<org.ole.planet.myplanet.model.UserEntity>>()
```

`UserRepositoryImpl.kt:51` is `import org.ole.planet.myplanet.model.UserEntity`, and the short name is used 46 times elsewhere in this same file. There is no ambiguity to disambiguate and nothing to gain — these two lines should read `HashMap<String?, UserEntity>()` and `HashMap<String?, MutableList<UserEntity>>()`. Same rule for the rest of the codebase: import it and use the short name.

**The memory win is smaller than the diff implies.** `seenUsers` still holds an entry for *every distinct name*, so the map itself hasn't shrunk — what's saved is the per-singleton `ArrayList` that `groupBy` allocates for names with only one user. That's real, but it's a thin slice of the actual footprint: `userDao.getAll()` has already materialized every `UserEntity` in the table, and those objects dominate. Trading three obvious lines for twenty to shave the smaller allocation is a poor exchange at the sizes involved.

If duplicate cleanup is genuinely heavy on large user tables, the win is not loading the table at all:

```sql
SELECT * FROM users WHERE name IN (
    SELECT name FROM users GROUP BY name HAVING COUNT(*) > 1
)
```

That returns only the rows that can possibly be deleted — usually a handful — and leaves both the grouping and the peak memory to SQLite. The Kotlin side then keeps the readable `groupBy`, because it's operating on a tiny list.

**Verdict:** correct but not worth the readability cost as scoped. Fix the two FQNs regardless. I'd rather see the duplicate detection pushed into the query, with `groupBy` restored — that addresses the memory concern the title raises, instead of the fraction of it that survives after `getAll()`.

---

dogi:

CI green on all three checks. This took the better path rather than the minimum one — the hand-rolled two-map grouping is gone entirely, replaced by a query that answers the actual question:

```sql
SELECT * FROM users WHERE name IN (SELECT name FROM users GROUP BY name HAVING COUNT(*) > 1)
```

That returns only rows that can possibly be deleted — usually a handful — so `getAll()` no longer materializes the entire user table just to discard most of it. The memory concern in the title is now addressed at its source rather than at the margin, and the Kotlin side goes back to a readable `groupBy` because it's operating on a tiny list. The two FQNs went with the rewrite.

**Two small observations, neither blocking:**

`if (users.size > 1)` inside the loop is now always true — the query guarantees it. Harmless as a belt-and-braces check, though it can go.

There's a subtle behavioural narrowing worth knowing about: `name IN (subquery)` never matches when `name` is `NULL`, since `NULL IN (…)` is never true in SQL. The old `groupBy { it.name }` bucketed null-named users together and would have deduplicated among them. I'd call that an improvement — collapsing users because they *both* lack a name is not obviously right — but it is a change, so worth a line in the description rather than a surprise later.

**Verdict:** correct, and the right fix rather than a smaller one. Marking `ready`.

---

Okuro3499 (approved):

PR #15683, opened 2026-08-13
courses: smoother download dialog handling (fixes #15435)

### LINKED ISSUE #15435: Do not show this dialog in Courses and My Course

dogi:
- Added `shouldShowDownloadDialog` to `BaseResourceFragment`
- Used `shouldShowDownloadDialog` in `BaseRecyclerFragment` to prevent unnecessary DB queries
- Disabled the dialog cleanly in `CoursesFragment` and `TeamCoursesFragment`
- Removed hacky empty overrides for `showDownloadDialog`

---
*PR created automatically by Jules for task [2186841099224776187](https://jules.google.com/task/2186841099224776187) started by @dogi*


dogi:
**Review — rating: 84/100**

CI green on all three checks. This is the change #15663 should have been — and it's worth being explicit about the difference, because the two look superficially similar.

#15663 added a bare `return` to an already-empty override, which compiled to nothing. This PR replaces the empty-override pattern outright with a declared property, and in doing so fixes something the overrides genuinely did *not* fix:

```kotlin
if (isMyCourseLib && courseLib == null && !isSurvey) {
    val userId = sharedPrefManager.getUserId().ifEmpty { "--" }
    val libraryList = resourcesRepository.getLibraryListForUser(userId)   // ← ran regardless
    showDownloadDialog(libraryList)                                        // ← then no-oped
}
```

The empty overrides suppressed the *dialog*, but `BaseRecyclerFragment` still ran `getLibraryListForUser(userId)` first and threw the result away. Hoisting the check to the call site skips the query entirely for fragments that opted out. That's the actual win, and it's not visible from the PR title.

The design is also better than what it replaces. `open val shouldShowDownloadDialog` declares intent in one line at the top of each fragment, where an empty method body with an explanatory comment relied on the reader noticing it 180 lines down. Keeping the belt-and-braces `if (!shouldShowDownloadDialog) return` inside `showDownloadDialog` means any future call site is covered without a third edit.

I checked that the two removed overrides were the only ones — they were, so no fragment silently regains the dialog, and everything else keeps the `true` default.

**One thing to clean up:** `MyLibrary` is now an unused import in `CoursesFragment`. It appears exactly once in the post-change file, on its own import line — the removed override signature was its only use. `TeamCoursesFragment` still uses it three times, so that one's fine.

**Minor:** `shouldShowDownloadDialog` is declared above the `@Inject` fields in `BaseResourceFragment`, which reads a little oddly against the surrounding convention. No functional issue — the property has a constant initializer and is only read after construction, so there's no open-val-during-init hazard.

**Verdict:** real fix, better structure than the empty-override approach, and it supersedes #15663 — I'd close that one in favour of this. Drop the dead `MyLibrary` import and this is ready; marking `ready` on the assumption that's a trivial follow-up.

---

Okuro3499 (approved):

PR #15685, opened 2026-08-13
🧪 Add unit tests for TeamTask.serialize

dogi:
🎯 **What:** Added unit tests for the `TeamTask.serialize` method which was previously untested.

📊 **Coverage:**
- Tested the "happy path" where a fully populated `TeamTask` and `UserEntity` are provided.
- Tested the edge case where `UserEntity` is null and the task's `_id` is empty to verify null handling and `TextUtils.isEmpty` branches.
- Mocked Android's `TextUtils` to allow running as a fast local JVM unit test without Robolectric.

✨ **Result:** Improved test coverage and reliability for the `TeamTask` model serialization, making future refactoring safer.

---
*PR created automatically by Jules for task [2397686028500798667](https://jules.google.com/task/2397686028500798667) started by @dogi*

### DIFF
--- a/app/src/test/java/org/ole/planet/myplanet/model/TeamTaskTest.kt
+++ b/app/src/test/java/org/ole/planet/myplanet/model/TeamTaskTest.kt
@@ -1,0 +1,101 @@
+package org.ole.planet.myplanet.model
+
+import android.text.TextUtils
+import com.google.gson.JsonObject
+import io.mockk.every
+import io.mockk.mockk
+import io.mockk.mockkStatic
+import io.mockk.unmockkAll
+import org.junit.After
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertFalse
+import org.junit.Assert.assertTrue
+import org.junit.Before
+import org.junit.Test
+
+class TeamTaskTest {
+
+    @Before
+    fun setup() {
+        mockkStatic(TextUtils::class)
+        every { TextUtils.isEmpty(any()) } answers {
+            val str = arg<CharSequence?>(0)
+            str == null || str.isEmpty()
+        }
+    }
+
+    @After
+    fun tearDown() {
+        unmockkAll()
+    }
+
+    @Test
+    fun testSerialize_withUser() {
+        val task = TeamTask().apply {
+            _id = "task123"
+            _rev = "rev123"
+            title = "Task Title"
+            deadline = 1620000000000L
+            description = "Task Description"
+            completed = true
+            completedTime = 1620000000000L
+            sync = "{\"syncKey\":\"syncValue\"}"
+            link = "{\"linkKey\":\"linkValue\"}"
+        }
+
+        val userEntity = mockk<UserEntity>()
+        val userJson = JsonObject().apply { addProperty("userName", "John Doe") }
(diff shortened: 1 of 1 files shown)


dogi:
**Review — rating: 82/100**

CI green on all three checks. Solid coverage for `TeamTask.serialize`, using the stack this project actually standardises on — MockK with `mockkStatic` for `TextUtils`, `@Before`/`@After` with `unmockkAll`, no Mockito.

**File collision with #15686 — these two cannot both merge**

#15686 creates the **same file** at the same path with the same class name:

```
app/src/test/java/org/ole/planet/myplanet/model/TeamTaskTest.kt   →  class TeamTaskTest
```

That's an add/add conflict in git, and even resolved by hand it wouldn't compile — two `TeamTaskTest` classes in one source set. They're not duplicates in content (this one covers `serialize`, #15686 covers `fromJson`), so nothing should be thrown away; they need to be **one file**.

They also differ in approach, which has to be reconciled rather than merged mechanically: this PR mocks `TextUtils` statically via MockK, while #15686 runs under `RobolectricTestRunner` and gets the real framework implementation. Both are legitimate and both are used elsewhere in this repo — but one file can only have one runner. Robolectric on the class plus real `TextUtils` is the more faithful option and removes the need for `mockkStatic` entirely; MockK-only is faster. Worth an explicit choice.

Since this is the lower number and is otherwise ready, the cheapest path is: land this one, then rebase #15686 to add its `fromJson` cases to the existing file rather than creating it. I've marked #15686 `change` with that instruction.

**Minor:** with `mockkStatic(TextUtils::class)` in play, it's worth an assertion or two that exercise the empty/blank branches of `serialize` specifically — that's the behaviour the mock exists to control, and it's where a regression would hide.

**Verdict:** good test, right conventions, adds coverage where there was none. Marking `ready` — merge this before #15686.

---

dogi:
**Review — rating: 60/100 (duplicate)**

The tests themselves are correct. I checked them against `TeamTask.serialize`: the `_id = ""` case correctly asserts that *both* `_id` and `_rev` are omitted, and the `assignee`-as-empty-string vs `assignee`-as-object asymmetry is pinned down properly. As standalone work there's nothing wrong with it.

The problem is #15686.

**It adds the same file — `app/src/test/java/org/ole/planet/myplanet/model/TeamTaskTest.kt` — as a new file, exactly as this PR does.** Two new files at one path is an add/add conflict: whichever merges second cannot merge without manual resolution.

And the overlap is total, not partial. #15686 contains `testSerialize_withUser` and `testSerialize_withoutUser` with the same assertions as this PR, then adds three `fromJson` tests on top — 166 lines against this PR's 101. **Everything here is already in #15686.**

**Recommendation: close this in favour of #15686.** Nothing is lost — the serialize coverage survives verbatim — and it removes a guaranteed merge conflict. Merging this one first only forces #15686 to rebase for no gain, since its file already supersedes this one.

If you'd rather keep this PR as the landing spot instead, the equivalent move is to close #15686 and cherry-pick its three `fromJson` tests here. Either way one of the two needs to go before either can merge.

I'm flagging this rather than acting on it — closing the PR is the maintainers' call.

---

PR #15686, opened 2026-08-13
teams: smoother task json testing (fixes #15783)

### LINKED ISSUE #15783: 🧪 test: Add unit tests for TeamTask fromJson mapping
🎯 **What:** The testing gap addressed was an untested `TeamTask.fromJson` parsing method. It implicitly accessed fields from generic `JsonObject` instances which is error-prone.
  📊 **Coverage:** The tests now comprehensively cover: 1) the "happy path" with valid and fully populated data, 2) handling scenarios with missing nested `_id` values (specifically assignee), and 3) handling cases with completely empty data objects to ensure defaults are applied without crashing.
  ✨ **Result:** Enhanced test suite coverage for data models and increased regression safety for future modifications to `Te

dogi:
🎯 **What:** The testing gap addressed was an untested `TeamTask.fromJson` parsing method. It implicitly accessed fields from generic `JsonObject` instances which is error-prone.
  📊 **Coverage:** The tests now comprehensively cover: 1) the "happy path" with valid and fully populated data, 2) handling scenarios with missing nested `_id` values (specifically assignee), and 3) handling cases with completely empty data objects to ensure defaults are applied without crashing.
  ✨ **Result:** Enhanced test suite coverage for data models and increased regression safety for future modifications to `TeamTask`.

---
*PR created automatically by Jules for task [6682418883245391685](https://jules.google.com/task/6682418883245391685) started by @dogi*


dogi:
**Review — rating: 74/100**

CI green on all three checks. The `fromJson` coverage is worth having and the Robolectric approach is sound — but this PR and #15685 both **create the same file**, so they cannot both merge.

**The collision**

Both add `app/src/test/java/org/ole/planet/myplanet/model/TeamTaskTest.kt` declaring `class TeamTaskTest`. Git sees an add/add conflict, and even a hand-resolved merge wouldn't compile — one source set can't hold two classes with that fully-qualified name.

The content isn't duplicated: #15685 covers `serialize`, this covers `fromJson` mapping. Both sets of tests should survive; they just have to live in one file.

**The approaches differ in a way that needs a decision, not a merge.** This PR uses `@RunWith(RobolectricTestRunner::class)` so `TextUtils` behaves for real; #15685 uses `mockkStatic(TextUtils::class)` instead. A single class can only have one runner. My preference is this PR's approach — real `TextUtils` beats a mocked one, and it drops the `mockkStatic`/`unmockkAll` scaffolding entirely — but Robolectric is the slower runner, so it's a legitimate trade either way.

**Suggested resolution:** #15685 is the lower number and is otherwise ready, so let it land first, then rebase this one to *add* the `fromJson` tests to the now-existing file rather than creating it. If you'd rather keep Robolectric as the runner, say so on #15685 and swap the two — either order works, but the second PR has to become an edit rather than an add.

The tests themselves are fine: `assertNull` coverage for absent fields is the part most often skipped, and it's here.

**Verdict:** good tests blocked by a file conflict. Rebase onto #15685's file and this is ready.

---

dogi:
**Review — rating: 86/100**

I checked every assertion against `TeamTask.serialize` / `TeamTask.fromJson` rather than trusting green CI, and they all pin real behavior:

- `_id = ""` exercising the `if (!TextUtils.isEmpty(task._id))` guard, asserting **both** `_id` and `_rev` are absent — that guard controls two properties and the test covers both.
- `assignee` as an empty *string* in the null-user case vs. a nested object otherwise. That asymmetry (`addProperty("assignee", "")` vs `add("assignee", user.serialize())`) is exactly the sort of thing that gets "cleaned up" into a null and silently breaks the server contract. Good thing to have locked down.
- `task.teamId` derived from `link.teams` — a field that's easy to break because it's read from a *nested* object rather than the top level.
- The `assignee` object present but `_id` missing → `assignee` stays null, covering the `if (user.has("_id"))` branch.
- Empty-`JsonObject` round trip asserting `link` and `sync` become `"{}"` rather than null or `""`.

`@RunWith(RobolectricTestRunner::class)` is genuinely required here — `serialize` calls `TextUtils.isEmpty`, which is a stub outside Robolectric. Not gratuitous.

**One gap worth filling, because it likely exposes a real bug.** Nothing covers `serialize` when `task.sync` or `task.link` is null:

```kotlin
`object`.add("sync", JsonUtils.gson.fromJson(task.sync, JsonObject::class.java))
```

`gson.fromJson(null, …)` returns null, and `JsonObject.add` converts a null element to `JsonNull`. So a task with no sync/link data serializes to `"sync": null` and gets uploaded that way. These fields come off a Room entity where null is entirely reachable. A test asserting the current behavior would document it; if the intended output is `{}` or an omitted key, the test fails and you've found something. Either outcome is more useful than the current silence.

**Blocking issue that isn't about the code: this collides head-on with #15685.**

Both PRs add `app/src/test/java/org/ole/planet/myplanet/model/TeamTaskTest.kt` as a **new file**. That's an add/add conflict — whichever merges second cannot merge at all without manual resolution.

And it isn't a partial overlap. #15685's `testSerialize_withUser` and `testSerialize_withoutUser` are the same two tests with the same assertions as the first two here. **#15686 is a strict superset of #15685** — same serialize coverage, plus three `fromJson` tests and 65 more lines.

The commit message here says "Merge fromJson tests into existing file", which suggests the intent was to land on top of #15685's file. That didn't happen — this is still a standalone new file, so the conflict stands.

**Recommendation: land this one and close #15685.** Everything #15685 contributes is already here, so closing it costs nothing and merging it first only creates a rebase for this PR. I've commented on #15685 to that effect. (Flagging only — I don't close PRs.)

**Verdict:** `ready` on the merits. The tests are accurate, cover the branches that matter, and use Robolectric for a real reason. Resolve the overlap with #15685 before merging, and consider adding the null `sync`/`link` case.

---

Okuro3499 (approved):

PR #15687, opened 2026-08-13
resources: smoother repository inserting (fixes #15812)

### LINKED ISSUE #15812: ⚡ Fix N+1 query bottleneck in ResourcesRepository batch insertions
💡 **What:** Optimized `batchInsertResources` and `batchInsertMyLibrary` in `ResourcesRepositoryImpl` by accumulating `MyLibrary` resources in a list and using a batch `upsertAll` DAO method instead of calling `upsert` N times in a loop.
🎯 **Why:** To resolve a severe N+1 query issue that was degrading performance significantly during batch resource synchronization/insertion.
📊 **Measured Improvement:** The number of SQLite database transactions is dramatically reduced. A custom test (`ResourcesRepositoryImplNPlusOneTest`) verified that a 5-item list previously invoked `upsert` 5 times, whereas

dogi:
💡 **What:** Optimized `batchInsertResources` and `batchInsertMyLibrary` in `ResourcesRepositoryImpl` by accumulating `MyLibrary` resources in a list and using a batch `upsertAll` DAO method instead of calling `upsert` N times in a loop.
🎯 **Why:** To resolve a severe N+1 query issue that was degrading performance significantly during batch resource synchronization/insertion.
📊 **Measured Improvement:** The number of SQLite database transactions is dramatically reduced. A custom test (`ResourcesRepositoryImplNPlusOneTest`) verified that a 5-item list previously invoked `upsert` 5 times, whereas now it invokes `upsertAll` exactly 1 time. This equates to an O(1) query profile compared to the previous O(N) profile for batch operations.

---
*PR created automatically by Jules for task [17033819639281705526](https://jules.google.com/task/17033819639281705526) started by @dogi*

## Summary by CodeRabbit

* **Improvements**
  * Improved the efficiency of importing multiple library items and resources.
  * Updated the app version to 0.65.36.


dogi:
**Review — rating: 84/100**

CI green on all three checks. Straightforward N+1 fix, correctly done in both places.

Collecting into `librariesToUpsert` and issuing a single `myLibraryDao.upsertAll(...)` replaces one Room transaction per document with one per batch. On a sync pulling hundreds of resources that's the difference between hundreds of commits and two.

**Two things I checked that could have made this wrong:**

The `existingItems[resourceId] = library` bookkeeping still happens inside the loop, so a later document referencing an id seen earlier in the same batch still observes the in-memory version. Deferring only the *write* preserves that.

Error handling is unchanged in effect: the `try/catch` sits inside `forEach`, so a malformed document is skipped and the loop continues — it never aborts before reaching the batched write. The one real difference is atomicity, and it moves in the safer direction: previously a crash midway through left a partial set of rows committed, now the batch either lands or doesn't.

Unlike the `IN (...)` chunking discussions on #15680 and #15630, there's no host-variable ceiling to worry about here — Room's `@Upsert` with a collection issues per-row statements inside one transaction rather than a single parameterized query, so an unbounded list is fine. Memory is bounded by `documents`, which is already fully materialized.

**Minor:** both call sites are now the same shape (declare list, fill in loop, guarded `upsertAll`). If a third appears, it's worth extracting.

**Verdict:** real fix, no behavioural surprises. Marking `ready`.

---

Okuro3499 (dismissed):
```
The test only verifies getByIds is called once; it asserts nothing about upsertAll/upsert call counts. So if a future edit reintroduces the per-item myLibraryDao.upsert()
call (the exact N+1 pattern this PR removes), no test would catch the regression.
```

dogi:
**Re-review — Rating: 86/100** (was 84)

The production change is byte-identical to what I rated 84; the new commits add the tests it was missing, and they assert the right things.

```kotlin
coVerify(exactly = 1) { myLibraryDao.upsertAll(match { it.size == 5 }) }
coVerify(exactly = 0) { myLibraryDao.upsert(any()) }
```

That's the pair that matters. The positive check alone would pass if the code batched *and* also kept calling `upsert` per row; the `exactly = 0` on the single-row path is what actually pins the N+1 fix down. Both `batchInsertResources` and `batchInsertMyLibrary` get the same treatment, so neither can regress independently.

**One thing to fix:** `com.google.gson.JsonObject()` is written out in full at lines 424 and 443, and the file has no `JsonObject` import. Add `import com.google.gson.JsonObject` and use the short name at both sites.

**A refinement to what I said last round, now that the tests made me look again.** I described the atomicity change as moving purely in the safer direction. That's right for consistency, and it's right for the *parse* failures the `try/catch` inside `forEach` handles — a malformed document is still skipped and the loop still reaches the batched write. But the DB write itself moved out from under that `catch`:

```kotlin
} catch (e: Exception) {
    e.printStackTrace()
}
}
if (librariesToUpsert.isNotEmpty()) {
    myLibraryDao.upsertAll(librariesToUpsert)   // ← no longer inside the per-doc catch
}
```

Before, a row-level DB failure was swallowed and the remaining documents still persisted. Now one bad row fails the whole batch and throws out of the function, so `processedCount` / `savedIds` never reach the caller. For a sync pulling from CouchDB that's a defensible trade — a partially-applied batch is arguably worse than a retried one — but it is a resilience change on top of the performance one, and it deserves a line in the PR description so whoever handles the next sync failure report knows where the behaviour moved.

If you'd rather keep both properties, wrapping the `upsertAll` in its own `try/catch` and returning the count on failure preserves the old resilience; chunking via the existing `processInBatches` helper (used in `UploadManager.uploadTeams`) would additionally bound the blast radius to one chunk instead of the whole batch. Neither is necessary for this to land.

I re-checked the two things that could have made the batching wrong and both still hold: `getByIds` is already `chunked(900)` at lines 535 and 574, so there's no host-variable ceiling in play, and Room's `@Upsert` over a collection issues per-row statements inside one transaction rather than one giant parameterized query — so an unbounded `librariesToUpsert` is fine on that axis. Memory is still bounded by `documents`, which the caller has already materialised.

**Verdict:** `ready`. Import `JsonObject` and drop the two fully-qualified constructions — doesn't need another review round.

---

Okuro3499 (approved):

PR #15682, opened 2026-08-13
sync: smoother upload repository attachment dispatcher providing (fixes #15782)

dogi:
💡 **What:**
Injected `DispatcherProvider` into `UploadRepositoryImpl` and wrapped the blocking I/O calls (`file.toURI().toURL().openConnection()` and `FileUtils.fullyReadFileToBytes(file)`) in `UploadRepositoryImpl.uploadAttachment` with `withContext(dispatcherProvider.io)`. Mocked `DispatcherProvider` was injected into the test setup using `UnconfinedTestDispatcher`.

🎯 **Why:**
`uploadAttachment` is a `suspend` function, but it was executing blocking file reads and URL connections synchronously. This blocked the underlying thread from executing other coroutines, which negatively impacts overall application performance and responsiveness, especially on the `main` or `default` thread pool.

📊 **Measured Improvement:**
Although I was unable to compile the benchmark runner properly in the environment due to missing dependencies (`kotlinc` not found in `bash` / environment classpath issues), moving blocking file I/O and network operations to `Dispatchers.IO` is a fundamental coroutine best practice. By offloading these operations, the calling dispatcher (e.g. `Dispatchers.Main` or `Dispatchers.Default`) is no longer blocked for the duration of the disk read (e.g., reading a multi-megabyte file) or the initial network connection setup. This guarantees UI fluidity if called from the Main thread, and prevents thread starvation if called from the Default thread pool.

---
*PR created automatically by Jules for task [8639373676185993072](https://jules.google.com/task/8639373676185993072) started by @dogi*

## Summary by CodeRabbit

* **Performance**
  * Improved attachment uploads by processing files and preparing upload requests in the background.
  * Reduced the impact of file processing on app responsiveness.

* **Reliability**
  * Improved dispatcher handling for more consistent upload processing.
  * Upload submission behavior remains unchanged.

* **Maintenance**
  * Updated the app to version 0.65.13.

PR #15693, opened 2026-08-14
login: smoother configurations repository provisioning (fixes #15811)

### LINKED ISSUE #15811: Refactor SyncActivity to extract provisioning logic into ConfigurationsRepository
Extract `SyncActivity` provisioning logic out of the UI layer.

This PR extracts the `clearInternalStorage` and `downloadAdditionalResources` logic out of `SyncActivity.kt` into a single, unified `provisionApp()` method inside `ConfigurationsRepository`. 
The new `provisionApp(hasWritePermission)` method handles checking the `getFirstRun` flag, deleting `FileUtils.getOlePath()` if permission is granted, and parsing concatenated links to start the download service. `SyncActivity` is updated to invoke this method during `onSyncComplete`.

dogi:
Extract `SyncActivity` provisioning logic out of the UI layer.

This PR extracts the `clearInternalStorage` and `downloadAdditionalResources` logic out of `SyncActivity.kt` into a single, unified `provisionApp()` method inside `ConfigurationsRepository`. 
The new `provisionApp(hasWritePermission)` method handles checking the `getFirstRun` flag, deleting `FileUtils.getOlePath()` if permission is granted, and parsing concatenated links to start the download service. `SyncActivity` is updated to invoke this method during `onSyncComplete`.

---
*PR created automatically by Jules for task [13302662138937293453](https://jules.google.com/task/13302662138937293453) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6543
-        versionName = "0.65.43"
+        versionCode = 6544
+        versionName = "0.65.44"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/repository/ConfigurationsRepository.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/ConfigurationsRepository.kt
@@ -13,7 +13,8 @@ interface ConfigurationsRepository {
     suspend fun getMinApk(url: String, pin: String): ConfigurationResult
     fun getPlanetType(): String?
     suspend fun ensureServerUrlUpdated()
-    fun provisionApp(hasWritePermission: Boolean)
+    suspend fun clearFirstRunStorage(hasWritePermission: Boolean)
+    suspend fun startQueuedDownloads()
 
     interface CheckVersionCallback {
         fun onUpdateAvailable(info: MyPlanet?, cancelable: Boolean)
--- a/app/src/main/java/org/ole/planet/myplanet/repository/ConfigurationsRepository.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/ConfigurationsRepository.kt
@@ -13,8 +13,8 @@ interface ConfigurationsRepository {
     suspend fun getMinApk(url: String, pin: String): ConfigurationResult
     fun getPlanetType(): String?
     suspend fun ensureServerUrlUpdated()
-    suspend fun clearFirstRunStorage(hasWritePermission: Boolean)
-    suspend fun startQueuedDownloads()
+    suspend fun clearFirstRunStorageAndSetFlag(hasWritePermission: Boolean)
+    suspend fun getQueuedDownloads(): ArrayList<String>?
 
     interface CheckVersionCallback {
         fun onUpdateAvailable(info: MyPlanet?, cancelable: Boolean)
--- a/app/src/main/java/org/ole/planet/myplanet/repository/ConfigurationsRepository.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/ConfigurationsRepository.kt
@@ -13,6 +13,7 @@ interface ConfigurationsRepository {
     suspend fun getMinApk(url: String, pin: String): ConfigurationResult
     fun getPlanetType(): String?
     suspend fun ensureServerUrlUpdated()
+    fun provisionApp(hasWritePermission: Boolean)
 
     interface CheckVersionCallback {
         fun onUpdateAvailable(info: MyPlanet?, cancelable: Boolean)
--- a/app/src/main/java/org/ole/planet/myplanet/repository/ConfigurationsRepository.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/ConfigurationsRepository.kt
@@ -14,7 +14,7 @@ interface ConfigurationsRepository {
     fun getPlanetType(): String?
     suspend fun ensureServerUrlUpdated()
     suspend fun clearFirstRunStorageAndSetFlag(hasWritePermission: Boolean)
-    suspend fun getQueuedDownloads(): ArrayList<String>?
+    suspend fun getQueuedDownloads(): List<String>
 
     interface CheckVersionCallback {
         fun onUpdateAvailable(info: MyPlanet?, cancelable: Boolean)
--- a/app/src/main/java/org/ole/planet/myplanet/repository/ConfigurationsRepositoryImpl.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/ConfigurationsRepositoryImpl.kt
@@ -5,11 +5,13 @@ import androidx.core.content.edit
 import androidx.core.net.toUri
 import com.google.gson.JsonObject
 import dagger.hilt.android.qualifiers.ApplicationContext
+import java.io.File
 import java.io.IOException
 import java.net.ConnectException
 import java.net.SocketTimeoutException
 import java.net.UnknownHostException
 import java.util.concurrent.ConcurrentHashMap
+import kotlinx.serialization.json.Json
 import javax.inject.Inject
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.TimeoutCancellationException
@@ -36,6 +38,7 @@ import org.ole.planet.myplanet.utils.NetworkUtils
 import org.ole.planet.myplanet.utils.Sha256Utils
 import org.ole.planet.myplanet.utils.TimeProvider
 import org.ole.planet.myplanet.utils.UrlUtils
+import org.ole.planet.myplanet.utils.DownloadUtils
 import org.ole.planet.myplanet.utils.VersionUtils
 
 class ConfigurationsRepositoryImpl @Inject constructor(
@@ -367,6 +370,26 @@ class ConfigurationsRepositoryImpl @Inject constructor(
 
     override fun getPlanetType(): String? {
         return sharedPrefManager.getRawString("planetType")
+    }
+
+    override fun provisionApp(hasWritePermission: Boolean) {
+        if (hasWritePermission && sharedPrefManager.getFirstRun()) {
+            val myDir = File(FileUtils.getOlePath(context))
+            if (myDir.isDirectory) {
+                val children = myDir.list()
+                if (children != null) {
+                    for (i in children.indices) {
+                        File(myDir, children[i]).delete()
+                    }
+                }
+            }
+            sharedPrefManager.setFirstRun(false)
+        }
+        val storedJsonConcatenatedLinks = sharedPrefManager.getConcatenatedLinks()
+        if (storedJsonConcatenatedLinks != null) {
+            val storedConcatenatedLinks: ArrayList<String> = Json.decodeFromString(storedJsonConcatenatedLinks)
+            DownloadUtils.openDownloadService(context, storedConcatenatedLinks, true)
+        }
     }
 
     private fun buildCouchdbUrl(currentUrl: String, pin: String): String {
--- a/app/src/main/java/org/ole/planet/myplanet/repository/ConfigurationsRepositoryImpl.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/ConfigurationsRepositoryImpl.kt
@@ -11,7 +11,6 @@ import java.net.ConnectException
 import java.net.SocketTimeoutException
 import java.net.UnknownHostException
 import java.util.concurrent.ConcurrentHashMap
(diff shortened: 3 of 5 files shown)


dogi:
**Review — rating: 64/100**

CI green on all three checks. Moving provisioning out of `SyncActivity` is worth doing, but this merges two operations that ran at different times into one call, which changes when the storage wipe happens.

**The first-run wipe moves from before the sync to after it**

Previously these were two separate methods invoked at two points:

- `clearInternalStorage()` — called in the sync-trigger path (`SyncActivity.kt:~698`), *before* `isServerReachable(processedUrl, "sync")`
- `downloadAdditionalResources()` — called in the post-sync success path (`~531`)

Now both live in `provisionApp(hasWritePermission)`, called only at the post-sync site, and the pre-sync call is deleted. So on a first run the `ole/` directory is emptied *after* the sync completes rather than before it begins.

That matters if anything wrote into `FileUtils.getOlePath(context)` during the sync — the wipe would delete it. The window is narrow (`getFirstRun()` gates it, and a genuinely fresh install has an empty directory), but the case it was presumably guarding — a reinstall over leftover data — is exactly the case where the sync repopulates that directory first and then gets wiped. Worth confirming this reordering is intentional, and saying so in the description if it is.

**Blocking file I/O in a non-suspend repository method**

```kotlin
override fun provisionApp(hasWritePermission: Boolean) {
    …
    val myDir = File(FileUtils.getOlePath(context))
    if (myDir.isDirectory) { … File(myDir, children[i]).delete() … }
```

`isDirectory`, `list()` and a `delete()` loop are all disk I/O, and the method is a plain `fun` called from the activity. That was already true of `clearInternalStorage`, so it isn't a regression — but relocating it into the repository layer makes it the repository's contract, and this codebase's whole `DispatcherProvider` discipline says otherwise. `suspend fun provisionApp(...)` with a `withContext(dispatcherProvider.io)` body would fix it while it's being moved, and #15682 (merged this week) is the reference for that shape.

**`provisionApp` bundles two unrelated things.** Clearing stale storage on first run and kicking off queued downloads share only their call site. Two methods — `clearFirstRunStorage(hasWritePermission)` and `startQueuedDownloads()` — would keep the ordering explicit at the call site and let the activity decide when each runs, which is what the current split was doing.

**Import ordering:** `kotlinx.serialization.json.Json` is inserted between `java.util.concurrent.ConcurrentHashMap` and `javax.inject.Inject`, and `utils.DownloadUtils` lands after `utils.UrlUtils`.

**Worth checking:** `File`, `FileUtils`, `Json` and `openDownloadService` may now be unused imports in `SyncActivity` if the two removed methods were their only consumers.

**Verdict:** right instinct, and `ConfigurationsRepository` is the right home. Confirm or restore the wipe's ordering, make the method `suspend` with the I/O on a dispatcher, and consider keeping the two operations separate.

---

dogi:
**Review — rating: 72/100**

One of these two moves is a genuine fix; the other is in the wrong layer. Plus some leftovers.

**`clearFirstRunStorage` is the good one, and for a reason the description doesn't mention.** The old call site ran on the main thread:

```kotlin
isSync = true
if (checkPermission(Manifest.permission.WRITE_EXTERNAL_STORAGE) && prefData.getFirstRun()) {
    clearInternalStorage()   // directory listing + N file deletes, on the UI thread
}
lifecycleScope.launch { isServerReachable(processedUrl, "sync") }
```

Enumerating and deleting a directory's contents synchronously on the UI thread is an ANR waiting for a slow device with a full OLE folder. The new version does it under `dispatcherProvider.io`, and because the two calls are sequential suspends inside one `launch`, the ordering relative to `isServerReachable` is preserved. That's worth having on its own.

**`startQueuedDownloads` is in the wrong place.** It ends with:

```kotlin
withContext(dispatcherProvider.main) {
    DownloadUtils.openDownloadService(context, storedConcatenatedLinks, true)
}
```

A repository starting an Android Service inverts the project's layering — repositories sit *below* services, not above them. It's also inconsistent with where this exact call already lives elsewhere: `ResourceViewerViewModel.downloadResource` makes the same `openDownloadService` call from the ViewModel layer. Reading prefs and decoding the JSON is legitimately repository work; kicking off the service isn't. Returning the decoded `List<String>` and letting `SyncActivity` start the service keeps the layering and still gets the parse off the main thread.

(I checked `DownloadRepository` as an alternative home — it's a single `downloadFileResponse` HTTP abstraction, so it isn't one either.)

**Two stale imports left in `SyncActivity`.** The PR correctly dropped `FileUtils` and `openDownloadService`, but both methods it removed used two more that are now unreferenced:

- `import java.io.File` — zero remaining `File(` usages
- `import kotlinx.serialization.json.Json` — zero remaining `Json.` usages

**Two misordered imports in `ConfigurationsRepositoryImpl`:**

```kotlin
import java.util.concurrent.ConcurrentHashMap
import kotlinx.serialization.json.Json     // belongs after the kotlinx.coroutines block
import javax.inject.Inject
```

```kotlin
import org.ole.planet.myplanet.utils.UrlUtils
import org.ole.planet.myplanet.utils.DownloadUtils    // belongs before NetworkUtils
import org.ole.planet.myplanet.utils.VersionUtils
```

**Naming:** `clearFirstRunStorage` also does `sharedPrefManager.setFirstRun(false)`. The flag flip is the part a caller would be surprised to lose, and the name only advertises the deletion.

**Nit (pre-existing, moved as-is):** `File(myDir, children[i]).delete()` only removes top-level entries — `delete()` fails silently on a non-empty subdirectory, so nested content survives a "clear". `deleteRecursively()` is what the name implies. Changing it is arguably out of scope for a move, but it's worth knowing the method doesn't do what it says.

**No tests.** `ConfigurationsRepositoryImplTest` already exists, and `startQueuedDownloads`'s decode path (null prefs, malformed JSON, populated list) is straightforward to cover once the service call is out of it — which is a second argument for moving that call out.

**Merge note:** `ConfigurationsRepositoryImpl.kt` is also edited by **#15661** (`ready`), which adds a `@PlainGson` constructor parameter and touches the same import block.

**Verdict:** `change`. Keep `clearFirstRunStorage` — the off-main-thread move is a real improvement. Have `startQueuedDownloads` return the links instead of starting the service, clear the two stale `SyncActivity` imports, and fix the two import positions.

---

dogi:

Every point addressed, including the one that mattered most.

**The layering inversion is gone.** `startQueuedDownloads()` — which started an Android Service from inside a repository — is now `getQueuedDownloads(): ArrayList<String>?`, and `SyncActivity` starts the service:

```kotlin
val links = configurationsRepository.getQueuedDownloads()
if (!links.isNullOrEmpty()) {
    openDownloadService(context, links, true)
}
```

The repository does what a repository should — read the preference, decode the JSON, off the main thread — and the service start stays in the UI layer where the codebase already puts it. As a bonus this is now genuinely testable: it returns a value instead of a side effect.

**Also fixed:**

- `clearFirstRunStorage` → `clearFirstRunStorageAndSetFlag`, so the `setFirstRun(false)` side effect is visible in the name.
- `myDir.listFiles()?.forEach { it.deleteRecursively() }` replaces the index loop over `list()` with `File(myDir, name).delete()`. The old version silently left non-empty subdirectories behind — `delete()` fails on those — so "clear internal storage" now actually clears it. That was a pre-existing bug you didn't have to fix.
- Both stale `SyncActivity` imports removed — `java.io.File` and `kotlinx.serialization.json.Json` are gone, and I confirmed zero remaining usages of either.
- `kotlinx.serialization.json.Json` repositioned after the `kotlinx.coroutines` block, and the `DownloadUtils` import dropped along with the code that needed it.

**Still the good part from last round:** moving the first-run directory clear off the main thread. It ran synchronously in `startSync` before the coroutine, enumerating and deleting a directory on the UI thread, and the sequencing relative to `isServerReachable` is preserved because both are now sequential suspends in the same `launch`.

**Nit:** still no tests, but `getQueuedDownloads` is now a pure `prefs → List<String>?` function and `ConfigurationsRepositoryImplTest` already exists — null prefs, malformed JSON, and a populated list are three cheap cases.

**Merge note:** `ConfigurationsRepositoryImpl.kt` is also touched by **#15661** (`ready`), which adds a `@PlainGson` constructor parameter and edits the same import block. Trivial conflict, just needs sequencing.

**Verdict:** `ready`.

---

Okuro3499 (changes requested):
```
No tests cover the two new methods (clearFirstRunStorageAndSetFlag, getQueuedDownloads), despite this refactor's stated purpose being to make this logic testable outside the Activity.
```

dogi:
**Re-review — Rating: 82/100** (was 72)

Every substantive item from last round landed, and the layering complaint was fixed properly rather than papered over.

`startQueuedDownloads` is now `getQueuedDownloads()` and returns the decoded links instead of starting a Service from inside a repository. `DownloadUtils` and `openDownloadService` no longer appear anywhere in `ConfigurationsRepositoryImpl`; `SyncActivity` owns the service start again. That keeps repositories below services where the project puts them, and the JSON decode still happens off the main thread under `dispatcherProvider.io`. The caller even gained an `isNullOrEmpty()` guard the old `downloadAdditionalResources()` didn't have — an empty stored list used to start the download service with nothing to do.

`java.io.File` and `kotlinx.serialization.json.Json` are gone from `SyncActivity`, and both import-position problems in `ConfigurationsRepositoryImpl` are fixed — `java.io.File` sorts into the `java.*` block at line 8 and `kotlinx.serialization.json.Json` sits after the `kotlinx.coroutines` block at line 22. Tests appeared where there were none.

You also took the `deleteRecursively()` nit, which I'd flagged as arguably out of scope. It's the right semantics — `delete()` fails silently on a non-empty subdirectory, so the old "clear internal storage" left nested content behind — but it does mean this PR broadens a destructive operation, and the PR description still describes itself as an extraction. Worth a line there so it isn't a surprise later.

**One stale import survived.** `SyncActivity` still has `import org.ole.planet.myplanet.utils.FileUtils` at line 63, and `FileUtils` now appears nowhere else in the file — `clearInternalStorage()` was its only user. I listed it as already-dropped last round; it wasn't. Please remove it.

**`ArrayList<String>?` shouldn't be the interface type.** `getQueuedDownloads(): ArrayList<String>?` commits the contract to a concrete mutable implementation. `List<String>` is the idiomatic return type, and returning `emptyList()` rather than `null` would let the call site read `if (links.isNotEmpty())` without the null branch. The decode still works — `Json.decodeFromString<List<String>>` infers fine.

**The decode is unguarded.** `Json.decodeFromString(storedJsonConcatenatedLinks)` throws `SerializationException` on malformed stored prefs, and there's no `try/catch` between it and the `lifecycleScope.launch` in `SyncActivity`. The old code had the same hole, so this isn't a regression — but the whole point of moving it into a repository with a test file is that it's now trivial to guard and to cover. A `runCatching { … }.getOrNull()` and one more test case closes it.

**The deletion test doesn't test the deletion, and says so.** From `clearFirstRunStorageAndSetFlag sets firstRun to false when conditions are met`:

```kotlin
val mockFile = mockk<java.io.File>(relaxed = true)
every { mockFile.isDirectory } returns true
every { mockFile.listFiles() } returns arrayOf()

// We will not mock File constructor as it causes StackOverflow.
// ...
// File("mock_path").isDirectory will just return false on most systems,
// which skips the listFiles block. We mainly care about setFirstRun.
```

`mockFile` is constructed, stubbed twice, and never passed to anything — it's dead code, and the comment is scaffolding explaining a limitation rather than documenting behaviour. Neither should ship. The test that remains asserts only `setFirstRun(false)`, which means the one genuinely risky thing in this PR — the switch to recursive deletion — has no coverage at all. Two things would fix that cheaply: delete the unused mock and the comment block, and take the directory path as a parameter (or inject a small file-ops seam) so a real `@get:Rule TemporaryFolder` can exercise the delete. Failing that, at least add the two negative cases — `hasWritePermission = false` and `getFirstRun() = false` should both leave `setFirstRun` uncalled, and neither is covered today.

**Minor:** the new tests use `io.mockk.mockkObject`, `io.mockk.unmockkObject` and `mockk<java.io.File>` inline while `just` and `runs` were added as imports in the same diff. Import them and use short names, consistent with the rest of the file. And `clearFirstRunStorageAndSetFlag` addresses my naming complaint by advertising both jobs rather than resolving the fact that it does two — a caller can't clear storage without also flipping the flag. Fine as-is; just noting the "AndX" suffix is usually a hint to split.

**You'll need to merge master before this lands.** `ConfigurationsRepository.kt` conflicts because master has since added `suspend fun ensureServerUrlUpdated()` on the exact line your two declarations replace. Resolving it carelessly deletes that method — keep all three. `ConfigurationsRepositoryImpl.kt` is also touched by #15661, and `SyncActivity.kt` by both #15603 and #15698, so whichever of these lands last will need a rebase.

**Verdict:** `ready`. Drop the `FileUtils` import, widen the return type to `List<String>`, and clean the dead mock out of the new test — none of which needs another review round.

---

Okuro3499 (approved):

PR #15694, opened 2026-08-14
Refactor Collections, Activities, StorageCategoryDetail, and ResourceDetail Fragments to use ViewModels

dogi:
Extracts logic from four remaining screens (`CollectionsFragment`, `ActivitiesFragment`, `StorageCategoryDetailFragment`, and `ResourceDetailFragment`) into explicitly scoped Hilt ViewModels:

- `CollectionsViewModel`: Handles loading Tag data from DB via `TagsRepository`.
- `ActivitiesViewModel`: Handles long-running `getOfflineLogins` stream via `ActivitiesRepository`.
- `StorageCategoryDetailViewModel`: Manages storage items logic (load and delete via `ResourcesRepository`).
- `ResourceDetailViewModel`: Handles `MyLibrary` logic, rating tracking, and `UserEntity` injection via `ResourcesRepository`, `RatingsRepository`, and `UserRepository`.

All fragments are updated to strictly observe state flows using `collectLatestWhenStarted`. Cleaned up direct IO Coroutine usage and obsolete DAO/Repository injections inside these fragments.

---
*PR created automatically by Jules for task [17118903205406121772](https://jules.google.com/task/17118903205406121772) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/ui/dashboard/ActivitiesFragment.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/dashboard/ActivitiesFragment.kt
@@ -6,7 +6,7 @@ import android.view.View
 import android.view.ViewGroup
 import androidx.core.content.res.ResourcesCompat
 import androidx.fragment.app.Fragment
-import androidx.lifecycle.lifecycleScope
+import androidx.fragment.app.viewModels
 import com.github.mikephil.charting.data.BarData
 import com.github.mikephil.charting.data.BarDataSet
 import com.github.mikephil.charting.data.BarEntry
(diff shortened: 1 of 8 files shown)


dogi:
**Review — rating: 86/100**

CI green on all three checks. Labelled `enormous`, so I reviewed at hunk level across all eight files and read the four new ViewModels in full rather than tracing every line of `ResourceDetailFragment` — flagging that so the depth is clear.

**This is the ViewModel extraction the other PRs in this queue have been approximating.** The distinction matters: #15655 and #15647 moved repository calls behind `suspend` pass-throughs while the fragment kept running them in its own scope and holding the results in fields. Here the work actually moves:

- state lives in the ViewModel as `StateFlow` (`_library`, `_rating`, `_userModel`, `_isLibraryNotFound`, `_tagsWithChildren`)
- the work runs in `viewModelScope`, so it survives configuration changes instead of being cancelled and restarted
- the fragment collects with the project's `collectLatestWhenStarted` helper rather than raw `lifecycleScope.launch`
- repository injections leave the fragments entirely

`ResourceDetailFragment` losing 153 lines while gaining a 113-line ViewModel is the right shape — the fragment is smaller *and* the logic became testable.

**Two details I checked specifically:**

The old code had a `lifecycleScope.launch { if (!isAdded) return@launch … }` — fragment-scoped rather than view-scoped, which is the leak pattern I've flagged on several PRs this week. It's gone, replaced by ViewModel state, so the `isAdded` guard isn't needed any more either.

`CollectionsViewModel.loadTags` guards on `if (_tagsWithChildren.value == null)`, so returning to the screen doesn't re-query. Small, and exactly the payoff that makes the move worth doing.

**Scope note:** four unrelated screens in one PR (resources collections, dashboard activities, storage category detail, resource detail). They share no files, so there's no internal conflict, and the pattern is identical across all four — which is an argument for reviewing them together. Worth saying in the description that they're independent, so a reviewer knows a problem in one doesn't implicate the others.

**Verdict:** correct pattern, applied consistently, fragments genuinely shrink. Marking `ready`.

---

Okuro3499 (changes requested):
```
  - CollectionsFragment.kt:88-102 — setListAdapter()'s collectLatestWhenStarted block resets selectedItemsList/childMap/list on every STARTED re-entry, wiping user tag selections on
  backgrounding. Fix: guard the one-time initialization (e.g. if (!::list.isInitialized) { selectedItemsList = ArrayList(recentList); ... }), matching the existing                  
  ::adapter.isInitialized guard already in the same block and the pattern used elsewhere in the codebase (e.g. ResourcesFragment.kt).                                                
  - StorageCategoryDetailFragment.kt:136-154 — loadResources()'s collector reassigns items = loaded on every STARTED re-entry, wiping user checkbox selections on backgrounding. Fix:
  guard with something like if (items.isEmpty()) { items = loaded; ... }, or push checked-state changes back into the ViewModel so viewModel.items is the single source of truth     
  instead of a fragment-local mutable copy.   
```

PR #15696, opened 2026-08-14
community: smoother voices showing (fixes #15695)

### LINKED ISSUE #15695: community: deleting shared enterprise message requires two clicks and deletes original post
When a message/post is created within an Enterprise and shared to the Community, attempting to delete the message from the Community tab requires clicking the delete button twice. Additionally, deleting it from the Community completely deletes the original message from the Enterprise chat as well.

### Steps to Reproduce
1. Navigate to **Enterprises** and open an enterprise
2. Add a new message/post and select the option to share it to the **Community**.
3. Navigate to the **Community** tab -> **Messages/Voices**.
4. Locate the shared enterprise message and click the **Delete** (trash) icon, then confirm.
5. Notice that the message remains visible until you click **Delete** and confirm a **second time**.
6. Return to **Enterprises** -> open the same enterprise -> notice that the original message has also been deleted from the Enterprise chat.

### Expected Behavior
- Clicking delete on a shared message in the Community tab should remove/unshare it immediately on the first attempt without requiring a second click.
- Deleting or unsharing the message from the Community feed should only remove it from the Community view and preserve the original message inside the Enterprise chat (or have clear unshare vs delete options).

https://github.com/user-attachments/assets/2130faed-ca9e-445f-9f14-e6804d04c4bb

ragilzakaria:
https://github.com/user-attachments/assets/950406e2-ad9f-47b3-a55b-f354116cdefd

fixes #15695 

## Summary by CodeRabbit

- **Improvements**
  - Community news visibility now handles broader sharing configurations while hiding malformed entries.
  - Community news is consistently sorted by the most recent calculated date.
  - News updates now reflect changes to visibility and sharing settings.
  - Sharing is more resilient when community details are missing or invalid.
  - Deleting posts now correctly removes shared and team content while preserving valid remaining entries.


dogi:
**Review — rating: 84/100**

Several genuine fixes here, and the tests pin the branching that matters.

**The `distinctUntilChanged` predicates were suppressing the very updates this feature produces.** Both flows compared `message`, `isEdited`, `imageUrls`, `images` (and `labels`) — but sharing a post mutates `viewIn` and `sharedBy`, and unsharing mutates them back. So the flow saw "no change" and the UI never refreshed after a share. Adding `viewIn`, `viewableBy`, `viewableId`, and `sharedBy` to both predicates is the fix. This is the same failure mode that bit `CoursesRepositoryImpl` (an equality predicate omitting the field the UI depends on) — worth watching for in any hand-written `distinctUntilChanged`.

**`deletePost` no longer orphans rows.** Previously it always upserted the filtered `viewIn`, so removing the last share target left a post with an empty `viewIn` sitting in the database forever. Now an empty result deletes the post and its replies, and a non-empty one updates and clears `sharedBy`. The three new tests cover exactly the three paths — unshare-keeps-row, team-delete-removes-row-and-replies, community-delete-removes-row — which is the right decomposition.

**Sorting on the update path was missing.** `VoicesFragment.setData` sorted only in the adapter-null branch; every subsequent `submitList` got the unsorted list. Sorting once up front and using it in both branches (and in `showNoData`) fixes it.

**The `"@"` handling is coherent.** Building `communityId` as `""` rather than a bare `"@"` when both codes are empty, while still treating a stored `"@"` as a wildcard on read, keeps already-synced documents working. Good backward-compatibility thinking.

**One thing I'd like confirmed — the wildcard is broad:**

```kotlin
id.isEmpty() || id == "@" || userIdentifier.isEmpty() || userIdentifier == "@" || id.equals(userIdentifier, ignoreCase = true)
```

If `userIdentifier` is empty — which happens when `planetCode` and `parentCode` are both unset — **every** community-section entry matches, so a device in that state sees community posts from all planets rather than none. If that's the intended "not configured yet, show everything" fallback, fine; if the safer default is "show nothing until configured", the two `userIdentifier` clauses should go. Either way it deserves a comment, because it's the kind of condition that reads as defensive but is actually permissive.

**Nits:**

- The list is now sorted twice — `sortedByDescending { it.sortDate }` in the repository flow *and* `sortNews(list)` in the fragment. Harmless, but one of them is redundant, and the repository is the better place to keep it.
- `news.sharedBy = ""` for "not shared" while the field is otherwise a user id; `null` would distinguish "never shared" from "unshared" if anything ever needs to.

**Verdict:** `ready`. Three real bugs fixed with tests that would catch regressions in each. Please just answer the wildcard question — it's the only behavior here I can't tell is deliberate.

---

Okuro3499 (approved):

dogi (approved):

PR #15698, opened 2026-08-14
fix sync planet up to date toast (fixes #15697)

### LINKED ISSUE #15697: sync: "Planet is up to date" toast appears before data synchronization completes
When clicking the **Sync Now** button (or the top-bar sync icon), a popup toast stating *"Planet is up to date."* is displayed immediately at the start of the process, while the progress dialog for syncing account data and resources is still actively running.

### Expected Behavior
- The completion/status message should only appear after the entire synchronization process (account data, resources, and sync tasks) has fully finished.
- The initial version check should not show a completion toast before data sync has even begun.

ragilzakaria:
https://github.com/user-attachments/assets/fd8e0e45-6893-4933-96aa-e4baf96d33e1

fixes #15697

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved synchronization handling when content is already up to date, allowing the process to continue without displaying an unnecessary error.
  * Ensured progress indicators are dismissed appropriately after synchronization checks.
  * Replaced the upload-completion message with a localized notification.

* **Localization**
  * Added upload-completion translations for Arabic, Spanish, French, Nepali, Somali, and English.

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/repository/ConfigurationsRepository.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/ConfigurationsRepository.kt
@@ -20,6 +20,7 @@ interface ConfigurationsRepository {
         fun onUpdateAvailable(info: MyPlanet?, cancelable: Boolean)
         fun onCheckingVersion() {}
         fun onError(msg: String, blockSync: Boolean)
+        fun onUpToDate() {}
     }
 
     sealed class ConfigurationResult {
--- a/app/src/main/java/org/ole/planet/myplanet/repository/ConfigurationsRepositoryImpl.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/ConfigurationsRepositoryImpl.kt
@@ -134,7 +134,7 @@ class ConfigurationsRepositoryImpl @Inject constructor(
                 val rawApkVersion = fetchApkVersionString(spm)
                 val versionStr = gson.fromJson(rawApkVersion, String::class.java)
                 if (versionStr.isNullOrEmpty()) {
-                    callback.onError(context.getString(R.string.planet_is_up_to_date), false)
+                    callback.onUpToDate()
                     return@launch
                 }
 
@@ -467,7 +467,7 @@ class ConfigurationsRepositoryImpl @Inject constructor(
         } else if (currentVersion < info.minapkcode && apkVersion < info.minapkcode) {
             callback.onUpdateAvailable(info, true)
         } else {
-            callback.onError(context.getString(R.string.planet_is_up_to_date), false)
+            callback.onUpToDate()
         }
     }
 
--- a/app/src/main/java/org/ole/planet/myplanet/services/AutoSyncWorker.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/services/AutoSyncWorker.kt
@@ -97,12 +97,20 @@ class AutoSyncWorker @AssistedInject constructor(
         syncContinuation = null
     }
 
+    override fun onUpToDate() {
+        startSyncProcess()
+    }
+
     override fun onError(msg: String, blockSync: Boolean) {
         if (blockSync) {
             syncContinuation?.takeIf { it.isActive }?.resume(Unit)
             syncContinuation = null
             return
         }
+        startSyncProcess()
+    }
+
+    private fun startSyncProcess() {
         workerScope.launch(dispatcherProvider.io) {
             syncManager.start(this@AutoSyncWorker, "upload")
             launch {
--- a/app/src/main/java/org/ole/planet/myplanet/ui/sync/ProcessUserDataActivity.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/sync/ProcessUserDataActivity.kt
@@ -209,7 +209,7 @@ abstract class ProcessUserDataActivity : BasePermissionActivity(), OnSuccessList
         collectWhenStarted(flow.takeWhile { value ->
             if (value is SyncUiState.Success) {
                 safelyDismissDialog()
-                Toast.makeText(this@ProcessUserDataActivity, "upload complete", Toast.LENGTH_SHORT).show()
+                Toast.makeText(this@ProcessUserDataActivity, getString(R.string.upload_complete), Toast.LENGTH_SHORT).show()
                 false
             } else if (value is SyncUiState.Error) {
                 safelyDismissDialog()
--- a/app/src/main/java/org/ole/planet/myplanet/ui/sync/SyncActivity.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/sync/SyncActivity.kt
@@ -748,6 +748,15 @@ abstract class SyncActivity : ProcessUserDataActivity(), ConfigurationsRepositor
         }
     }
 
+    override fun onUpToDate() {
+        lifecycleScope.launch {
+            if (customProgressDialog.isShowing()) {
+                customProgressDialog.dismiss()
+            }
+            continueSyncProcess()
+        }
+    }
+
     override fun onError(msg: String, blockSync: Boolean) {
         lifecycleScope.launch {
             Utilities.toast(this@SyncActivity, msg)
--- a/app/src/main/res/values-ar/strings.xml
+++ b/app/src/main/res/values-ar/strings.xml
@@ -724,6 +724,7 @@
     <string name="invalid_url">عنوان URL غير صالح</string>
     <string name="please_enter_valid_url_to_continue">يرجى إدخال عنوان URL صالح للمتابعة</string>
     <string name="uploading_data_to_server_please_wait">جارٍ تحميل البيانات إلى الخادم، يرجى الانتظار…</string>
+    <string name="upload_complete">اكتمل الرفع</string>
     <string name="check_the_server_address_again_what_i_connected_to_wasn_t_the_planet_server">تحقق مرة أخرى من عنوان الخادم. ما قمت بالاتصال به ليس خادم بلانت</string>
     <string name="device_couldn_t_reach_local_server">لم يتمكن الجهاز من الوصول إلى الخادم. تحقق مما إذا كنت متصلاً بشبكة Wi-Fi الصحيحة للخادم المحلي (المجتمع).</string>
     <string name="device_couldn_t_reach_nation_server">لم يتمكن الجهاز من الوصول إلى الخادم. تحقق مما إذا كنت متصلاً بالإنترنت وحاول مرة أخرى.</string>
@@ -1130,7 +1131,6 @@
     <string name="edit_mission_and_services">تعديل المهمة والخدمات</string>
     <string name="this_team_has_no_description_defined">هذا الفريق ليس لديه وصف محدد</string>
     <string name="contact_preference">contact preference</string>
-    <string name="planet_is_up_to_date">الكوكب محدث</string>
     <string name="new_apk_version_required_but_not_found_on_server">مطلوب إصدار جديد من APK ولكنه غير موجود على الخادم. اتصل بالمشرف.</string>
     <string name="version_not_found">الإصدار غير موجود.</string>
     <string name="connection_failed">فشل الاتصال!</string>
--- a/app/src/main/res/values-es/strings.xml
+++ b/app/src/main/res/values-es/strings.xml
@@ -724,6 +724,7 @@
     <string name="invalid_url">URL inválida</string>
     <string name="please_enter_valid_url_to_continue">Por favor ingresa una URL válida para continuar.</string>
     <string name="uploading_data_to_server_please_wait">Cargando datos al servidor, por favor espera.....</string>
+    <string name="upload_complete">subida completada</string>
     <string name="check_the_server_address_again_what_i_connected_to_wasn_t_the_planet_server">Verifica de nuevo la dirección del servidor. A lo que me conecté no fue el servidor de Planet</string>
     <string name="device_couldn_t_reach_local_server">El dispositivo no pudo alcanzar el servidor. Verifique si está conectado a la red Wi-Fi correcta para el servidor local (comunidad).</string>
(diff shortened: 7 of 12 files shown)


dogi:
**Review — rating: 62/100**

The title says "fix sync planet up to date toast", but the diff also restructures `onError` in a way that changes error handling. That part needs attention.

**`onError` now swallows non-blocking errors entirely.**

Before, every error path did four things — toast the message, open the settings dialog for `Config` errors, dismiss the progress dialog — and *then* branched on `blockSync`. After:

```kotlin
if (blockSync) {
    Utilities.toast(this@SyncActivity, msg)
    if (msg.startsWith("Config")) { settingDialog() }
    if (customProgressDialog.isShowing()) { customProgressDialog.dismiss() }
    if (::syncIconDrawable.isInitialized) { … }
} else {
    continueSyncProcess()
}
```

For `blockSync == false` the message is now discarded with no user-visible trace, and — more concerning — **`customProgressDialog` is no longer dismissed** before `continueSyncProcess()` runs. The old code always dismissed first. Unless `continueSyncProcess()` reliably takes ownership of that dialog, a non-blocking error can leave a spinner on screen.

Suppressing noisy toasts during an otherwise-successful sync is a reasonable goal, but it should be a deliberate, separate decision from dropping the dialog dismissal — and it's not mentioned in the PR description at all.

**The new string is unused.** `upload_complete` is added to all six locale files:

```xml
<string name="upload_complete">upload complete</string>
```

but the call site uses a different key:

```kotlin
Toast.makeText(this@ProcessUserDataActivity, getString(R.string.planet_is_up_to_date), Toast.LENGTH_SHORT).show()
```

So six new translated strings ship with no reference anywhere. Either the toast should say `R.string.upload_complete` (which is what the old hardcoded `"upload complete"` literal meant, and what the translations were written for), or the string should be dropped from all six files. As it stands the translators did work for nothing and the user gets a message about being up to date after an *upload*, which isn't the same event.

**Duplicated completion feedback.** In `SyncActivity`:

```kotlin
Utilities.toast(activityContext, getString(R.string.planet_is_up_to_date))
showSnack(activityContext.findViewById(android.R.id.content), getString(R.string.sync_completed))
```

A toast and a snackbar, back to back, both announcing the same thing in different words. Pick one — the snackbar was already there.

**What's right:** replacing the hardcoded English `"upload complete"` with a resource lookup is the correct instinct, and adding the translation to every one of the six locale files rather than just `values/` is the right discipline.

**Verdict:** `change`. Point the toast at `R.string.upload_complete` (or delete the unused string), drop one of the two completion messages, and either restore the progress-dialog dismissal on the non-blocking path or explain in the description why non-blocking errors should now be silent.

---

dogi:
**Review — rating: 68/100**

Right diagnosis, and the string extraction half is clean. The gate is too blunt.

**The localization work is correct and complete.** `"upload complete"` was a hardcoded literal passed straight to `Toast.makeText`, and it's now `getString(R.string.upload_complete)` with the string added to all six `strings.xml` files. Each one is inserted at the same position — immediately after `uploading_data_to_server_please_wait` — which keeps the files line-parallel and makes future diffs readable. The English value reproduces the old literal exactly, so there's no visible change for English users. Translations read correctly: `اكتمل الرفع`, `subida completada`, `téléversement terminé`, `अपलोड पूरा भयो`. The Somali `dhamaystirka upload-ka` is a noun phrase ("the completion of the upload") where the others are predicates, and the English value is lowercase where neighbouring strings are sentence-case — both cosmetic.

**The `onError` change silences one message too many.** `blockSync` isn't a severity flag — it's control flow. Look at what `SyncActivity.onError` does with it:

```kotlin
if (!blockSync) {
    continueSyncProcess()      // not an error; carry on syncing
} else {
    syncIconDrawable.stop()    // stop here
}
```

So `blockSync = false` means "keep going", and it's used by exactly three call sites in `ConfigurationsRepositoryImpl`:

- line 133 — `planet_is_up_to_date`
- line 139 — `new_apk_version_required_but_not_found_on_server`
- line 442 — `planet_is_up_to_date`

Gating the toast on `blockSync` suppresses all three. Two of them are the bug you're fixing. The third isn't: "new apk version required but not found on server" tells an admin their Planet has a version record the app can't parse, and after this change that condition is completely silent — the sync just proceeds and nobody learns why the update never arrives. That's a diagnosable problem turned invisible.

**The underlying issue is that "up to date" is being reported as an error at all.** `checkVersion` uses `onError(msg, false)` as its "nothing to do, continue" signal, which is why the user taps sync and immediately sees a toast telling them they're up to date — before the sync they just asked for has even started. The clean fix is at the source: give `CheckVersionCallback` a method for that case (`onUpToDate()`, or just let `handleVersionEvaluation` return without a callback when there's nothing to report), and leave `onError` meaning what its name says. Then no gate is needed and the version-not-found warning keeps working.

If you'd rather keep the change local to `SyncActivity`, gate on the message rather than on `blockSync`:

```kotlin
if (msg != getString(R.string.planet_is_up_to_date)) {
    Utilities.toast(this@SyncActivity, msg)
}
```

Uglier, but it suppresses exactly the message that's the problem and nothing else.

**The PR has drifted from its title.** It's called "fix sync planet up to date toast" and the first four commits are all variations on *adding* an up-to-date toast (`b4c51d5`, `2fe22a5`, `4d2f826`), before `c6fe2d2` pivoted to removing one and switching the upload string. The final diff does the opposite of what the title and most of the history describe. Worth retitling and squashing so the record matches the change — a reader coming to this later will otherwise go looking for a toast that isn't there.

**No tests.** `SyncActivity` isn't unit-testable here, but if the fix moves into `ConfigurationsRepositoryImpl` as suggested above, `ConfigurationsRepositoryImplTest` already exists and the "version string empty → no error callback" path is a few lines to cover.

CI green on all three checks; merges cleanly. Note that `SyncActivity.kt` is also edited by #15603 and #15693, so expect a rebase depending on ordering.

**Verdict:** `change`. Keep the `upload_complete` extraction as-is. Narrow the suppression so `new_apk_version_required_but_not_found_on_server` still reaches the user — preferably by not routing the up-to-date case through `onError` in the first place.

---

dogi:

This is the fix at the source, and it's more complete than what I suggested.

Last round I objected that gating the toast on `blockSync` silenced `new_apk_version_required_but_not_found_on_server` alongside the two spurious "Planet is up to date" messages, and said the clean answer was to give `CheckVersionCallback` a method for the up-to-date case and leave `onError` meaning what its name says. That's what landed:

```kotlin
fun onUpToDate() {}
```

Both `onError(planet_is_up_to_date, false)` sites — line 133 and line 442 — now call `onUpToDate()`, and `SyncActivity.onError` is back to an unconditional toast. So the version-not-found warning reaches the user again, and the up-to-date case stops travelling through the error channel at all.

**The `AutoSyncWorker` override is the part I wouldn't have caught from my own suggestion.** The worker used `onError(_, blockSync = false)` as its "nothing to update, get on with syncing" signal:

```kotlin
override fun onError(msg: String, blockSync: Boolean) {
    if (blockSync) { … return }
    startSyncProcess()
}
```

Since `onUpToDate` has a default empty body, adding it without an override here would have left background auto-sync silently doing nothing on the most common path — a bug with no compile error and no failing test. Extracting `startSyncProcess()` and overriding `onUpToDate()` to call it closes that, and sharing the body between the two entry points keeps them from drifting.

I checked the blast radius: `CheckVersionCallback` has exactly two implementors in the codebase — `SyncActivity` and `AutoSyncWorker` — and both override the new method. Nothing is left relying on the default no-op.

`SyncActivity.onUpToDate` reproduces the old non-blocking path minus the toast: dismiss the progress dialog, then `continueSyncProcess()`. It correctly skips the `msg.startsWith("Config")` branch (that message never started with "Config") and skips the `syncIconDrawable` stop (that only ran on the blocking path). Behaviour preserved.

**The tests are the right two.** Both `onUpToDate()` call sites are covered — empty version string, and current version ≥ server version — and each asserts the negative that matters:

```kotlin
verify { callback.onUpToDate() }
verify(exactly = 0) { callback.onError(any(), any()) }
```

Without that second line the tests would pass on an implementation that called both.

**One leftover:** `R.string.planet_is_up_to_date` now has no code references — I grepped the branch and `ConfigurationsRepositoryImpl` is down to zero. It's still declared in all six `strings.xml` files. Worth deleting so the next person doesn't wire it back into an error path.

**Two smaller notes.**

The new tests reach for fully-qualified names inline — `io.mockk.mockkObject`, `io.mockk.unmockkObject`, `org.ole.planet.myplanet.utils.NetworkUtils`, `org.ole.planet.myplanet.utils.Constants`, `android.content.pm.PackageManager`. That matches the existing convention in this particular file rather than introducing it, so I won't push hard — but imports and short names would read better, and the file is already large.

The SDK branch inside the second test is dead either way:

```kotlin
if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.P) {
    longVersionCode = 2L
} else {
    @Suppress("DEPRECATION") versionCode = 2
}
```

Robolectric pins the SDK through `@Config`, so this resolves at a fixed value for the whole run — one branch is unreachable. Setting `longVersionCode` directly is clearer.

The seven commits are a visible record of working the problem out; worth squashing to the final design so the history reads as one decision.

CI green on all three checks; merges cleanly. Note `SyncActivity.kt` is also touched by #15603 (`change`) and #15693 (`merge`), so expect a rebase depending on order.

**Verdict:** `ready`. Delete the orphaned string when convenient — nothing here needs another round.

---

Okuro3499 (changes requested):
```
@ragilzakaria "Planet is up to date" is the message shown by the APK version check, not by data synchronization
Fix direction: rename/reword the string to something like "App is up to date" or suppress the toast for this no-op case, and only show a genuine "sync complete" message tied to SyncManager.syncStatus reaching Success. 
```

dogi:
**Review — rating: 88/100**

Correctly diagnosed: the "up to date" state was informational but was being delivered through `onError`, so it surfaced as an error toast. The fix adds a dedicated `onUpToDate()` callback instead of suppressing the toast at the call site, which is the right layer.

What I verified rather than assumed:

- **No implementer can silently stall.** This was my main worry with a defaulted interface method — an implementer relying on `onError` to continue the sync would now do nothing. `CheckVersionCallback` has exactly two implementers in `main` (`SyncActivity`, `AutoSyncWorker`) and both are overridden here.
- **`SyncActivity.onUpToDate()` is behavior-preserving.** Against the old `onError(msg, blockSync = false)` path it keeps the progress-dialog dismissal and the `continueSyncProcess()` call, and drops only the toast and the `msg.startsWith("Config")` branch that never matched this message.
- **`AutoSyncWorker`** extracts the non-blocking path into `startSyncProcess()` and calls it from both `onUpToDate()` and `onError`, so the refactor removes duplication rather than forking behavior.
- **Both call sites converted, no dangling resource.** `planet_is_up_to_date` had exactly two usages (`ConfigurationsRepositoryImpl` lines 137 and 471) and is removed from `values/` plus all five locale files.
- **The test's dropped `SDK_INT` branch is safe.** `VersionUtils.getVersionCode` reads `longVersionCode` only on P+; this test class carries no `@RunWith(RobolectricTestRunner)`, so `Build.VERSION.SDK_INT` is 0 and only the deprecated `versionCode` field is ever read. The old P+ branch was dead code here, and the `versionCode = 2` vs `latestapkcode = 1` test would fail outright if it weren't.

The two new tests assert `onUpToDate()` fires *and* `onError` does not (`exactly = 0`), which is precisely the regression to lock down. Localizing the hardcoded `"upload complete"` toast with all five translations is a welcome extra. Build (default + lite) and test (both shards) green.

**Nit:** `onUpToDate()` has a default no-op body, but unlike `onCheckingVersion()` it is control-flow-critical — continuing the sync hangs off it. A future implementer who forgets to override gets a silently stalled sync instead of a compile error; making it abstract like `onError`/`onUpdateAvailable` would close that off. Not a blocker, since both current implementers are correct.

**Verdict:** mergeable.

---

dogi:
**Review — rating: 88/100**

This fixes a real category error: "Planet is up to date" is the *success* case, and it was being delivered through `onError`. That meant a normal, healthy sync showed the user an error toast, and every implementer had to treat "everything is fine" as a failure it chose not to block on. Giving it its own `onUpToDate()` is the right shape.

I checked the thing that would break this — a `CheckVersionCallback` implementer left behind on the old path. There are exactly two, `SyncActivity:77` and `AutoSyncWorker:46`, and both override the new method, so nothing silently no-ops through the default body. I also diffed the new `SyncActivity.onUpToDate` against what `onError` did for this message at `SyncActivity.kt:751-769`: dismiss the dialog, then `continueSyncProcess()` because `blockSync` was false. The only thing dropped is the toast (the point of the PR) and the `msg.startsWith("Config")` branch, which never applied to this string. `AutoSyncWorker` is the same story — `startSyncProcess()` is a faithful extraction of the tail `onError` already fell through to.

Two bits of housekeeping in here that are better than they look. `planet_is_up_to_date` is removed from **all six** locale files now that nothing references it — that's the string-level cleanup that gets forgotten more often than not. And the hardcoded `"upload complete"` toast in `ProcessUserDataActivity` becomes `R.string.upload_complete` with translations in all six, which is a genuine localization fix for an app shipping `ar`/`es`/`fr`/`ne`/`so`. Slightly beyond the title's scope, but it's the same sync-toast surface and it's the right direction, so I'd keep it.

The two tests pin both routes into `onUpToDate` — the empty apk-version-string early return and the `handleVersionEvaluation` else-branch — and the now-pointless `context.getString(R.string.planet_is_up_to_date)` stubs are gone from the existing tests.

**Nit:** `onUpToDate()` gets a default empty body, matching `onCheckingVersion()`. Fine for the two implementers that exist, but the failure mode if someone adds a third and forgets is a sync that silently stops rather than a compile error — `onCheckingVersion` is genuinely optional, this one isn't. Worth considering making it abstract. Also each new test repeats ~20 lines of `sharedPrefManager` stubbing verbatim from its neighbours; a shared setup helper would make the next one cheap, though the duplication is pre-existing in that file.

**Verdict:** ready.

---

PR #15658, opened 2026-08-13
all: smoother configurations repository server url updating (fixes #15754)

dogi:
Replaces UserSessionManager and SharedPrefManager direct reads with UserRepository in CourseProgressViewModel, ProgressViewModel, and ResourceViewerViewModel to improve separation of concerns and centralize identity and preference handling through the repository layer. Updated corresponding tests.

---
*PR created automatically by Jules for task [17428781668191913421](https://jules.google.com/task/17428781668191913421) started by @dogi*

## Summary by CodeRabbit

* **New Features**
  * Improved automatic server URL handling by checking configured alternatives and selecting a reachable address.
  * Updated the app version to 0.64.93.

* **Bug Fixes**
  * Improved reliability when loading course and learning progress by retrieving user information through the updated account data flow.

* **Tests**
  * Updated progress-loading coverage to validate the revised user information flow.

PR #15699, opened 2026-08-14
resources: smoother ratings repository viewer prompt coordinating (fixes #15444)

### LINKED ISSUE #15444: show resource (not course) rating dialog after learner view the resource
If the user views a reource they have never rated before, show the rating dialog
Show the dialogue only after viewing the resource (Ex. after the user watches the video, but not when user see the resource details)

J-S-webskas:
https://github.com/user-attachments/assets/8d5f28ce-abc1-4696-8ba6-14f346f799d6

## Known limitation — "finished" is approximate for 3 of the 5 resource types

Only video and audio have a real completion signal. For PDF, image and text, `isResourceFinished` is set unconditionally, and `WebViewActivity` does not consult the flag at all, so for those types the feature is effectively **prompt on first exit**, not prompt when finished. Combined with `setRatingPrompted` firing as the dialog is shown, a learner who opens a PDF by accident spends their one prompt for that resource.

This is an accepted tradeoff for this PR: the prompt is capped at once per resource per user, so the worst case is one missed opportunity rather than repeated nagging. A dwell-time or scroll-position gate for PDF/image/text (and wiring the flag into `WebViewActivity`) is the follow-up that would make "finished" mean finished.

## Summary by CodeRabbit

* **New Features**
  * Resources now track completion for videos, PDFs, and images.
  * Completed resources can prompt users to submit a rating when leaving.
  * Rating prompts are suppressed after submission or dismissal.
  * Rating behavior is consistent across resource viewers and web content.

* **Bug Fixes**
  * Improved audio playback reliability.
  * Prevented rating-related background work from continuing after leaving an image viewer.

* **Tests**
  * Added coverage for rating prompts based on rating and prompt history.

Fixes #15444


dogi:
**Review — rating: 60/100**

The feature is reasonable — prompt for a rating once a video finishes, only if the user has never rated it — and `Player.STATE_ENDED` is the right hook. But two things need fixing, and one of them shouldn't be in this PR at all.

**1. `@RequiresApi(Build.VERSION_CODES.S)` is wrong and unrelated.**

```kotlin
@RequiresApi(Build.VERSION_CODES.S)
@OptIn(UnstableApi::class)
override fun setAuthSession(responseHeader: Map<String, List<String>>) {
```

`Build.VERSION_CODES.S` is API 31; this project's `minSdk` is **26**. Worse, `setAuthSession` is an override of `AuthSessionUpdater.AuthCallback` — the interface declares no such requirement, so annotating the implementation asserts a constraint the contract doesn't have, and every caller would need a version guard to satisfy lint. Nothing in this method is API-31-gated.

It's also entirely unconnected to showing a rating dialog. This has the shape of an accepted IDE quick-fix that came along with the `@OptIn(UnstableApi::class)` suggestion. Please drop it — and check whether the `@OptIn` is actually needed, or whether it was collateral from the same prompt.

**2. The null-safety is misleading, and doesn't do what it looks like it does.**

```kotlin
Player.STATE_ENDED -> {
    val rid = library?.resourceId ?: return
```

`library` is declared `private lateinit var library: MyLibrary` (line 99) — non-nullable. So `library?.` is a safe call on a non-null type, which Kotlin will warn about, and critically **it does not protect against the failure it appears to guard**: an uninitialized `lateinit` throws `UninitializedPropertyAccessException` on access, and a safe call doesn't catch that.

The same exposure exists in the helper, which uses `library.title` with no guard at all. If playback can ever reach `STATE_ENDED` before `library` is assigned, both sites throw. `::library.isInitialized` is the check that actually works here:

```kotlin
Player.STATE_ENDED -> {
    if (!::library.isInitialized) return
    val rid = library.resourceId ?: return
```

**3. The dialog reappears on every replay.** `STATE_ENDED` fires each time playback completes, so a user who watches a video twice without rating gets prompted twice, and looping media prompts on every cycle. A `hasPromptedThisSession` flag would make it once-per-viewing, which is what "show the dialog after finishing" implies.

**What's good:** checking both `summary.userRating` and `summary.existingRating` before prompting is the right condition, the `try/catch` around the ratings lookup means a repository failure degrades to not-prompting rather than crashing playback, and the `isAdded` check before `show()` guards the fragment-detached case.

**Nits:**

- No tests. `showResourceRatingDialogIfNeverRated` is a suspend function over two injected repositories — the has-rated and never-rated branches are straightforward to cover with MockK.
- Injecting both `RatingsRepository` and `UserRepository` directly into a Fragment sits oddly next to `ResourceViewerViewModel`, which already holds `resourcesRepository` and is where this lookup would naturally live.

**Verdict:** `change`. Remove the `@RequiresApi(S)` annotation, replace the ineffective safe call with `::library.isInitialized`, and consider guarding against repeat prompts. The feature itself is worth having.

---

dogi:
**Re-review — Rating: 68/100** (was 60)

All three findings are addressed. `@RequiresApi(Build.VERSION_CODES.S)` is gone — it asserted API 31 on a project with `minSdk 26`, on an override whose interface declares no such requirement. The ineffective `library?.resourceId` safe-call is replaced with `if (!::library.isInitialized) return`, which is the check that actually catches an uninitialized `lateinit`. And repeat prompting is guarded, more thoroughly than the session flag I suggested.

That last one is where the new problems are.

**The prompt flag isn't scoped to the user.** The key is `"prompted_$resourceId"`, and myPlanet is a shared-device app — that's the premise of the whole member/leader model. The code fetches `userId` two lines later to ask whether *this user* has rated:

```kotlin
val userId = userRepository.getUserModel()?.id
val hasRated = … ratingsRepository.getRatingSummary("resource", resourceId, userId) …
```

but the flag that gates the whole thing ignores it. On a classroom tablet, the first learner to finish a video is the only one ever prompted for it; everyone after them is silently skipped, including learners who have never rated anything. `"prompted_${userId}_$resourceId"` fixes it, and `userId` is already in hand.

**It writes to a private `SharedPreferences` file from a Fragment.** `requireContext().getSharedPreferences("rating_prompt_prefs", MODE_PRIVATE)` creates a seventh preferences store outside `SharedPrefManager`, which is the project's single owner for key-value settings and is already injectable here. Two consequences beyond the inconsistency: nothing clears these keys on logout or on "clear data", and the store grows by one boolean per resource ever finished, forever, with no eviction. Routing it through `SharedPrefManager` puts it where the rest of the app's prefs lifecycle already applies. Also, the project uses the `androidx.core.content.edit { }` KTX extension elsewhere rather than `edit().putBoolean(…).apply()`.

**The flag is set before the dialog is shown, and regardless of the outcome.**

```kotlin
if (!hasRated && isAdded) {
    prefs.edit().putBoolean(promptKey, true).apply()
    val ratingDialog = RatingsFragment.newInstance("resource", resourceId, library.title)
    ratingDialog.show(parentFragmentManager, RatingsFragment.TAG)
}
```

Combined with `catch (e: Exception) { false }` on the lookup, a transient repository failure spends the user's one lifetime prompt on a dialog that may have been unnecessary. Marking after a successful `show()` — or better, only once a rating is actually submitted — makes the flag mean what it's for. The `catch` also swallows silently; a `Log`/`printStackTrace` would at least leave a trace.

**Two imports came in with the annotation and stayed after it left.** `android.os.Build` (line 10) and `androidx.annotation.RequiresApi` (line 24) each appear exactly once in the file now — on their own import lines. Please remove both.

**Scope worth stating in the description:** this hooks `Player.STATE_ENDED`, so it fires only for resources that play through ExoPlayer. PDFs, images, text and HTML resources go through the other branches of `ResourceViewerFragment` and never prompt. That may well be the intended first step, but as written the PR title reads as though it covers finishing any resource.

**Still open from last round:**

- No tests. `showResourceRatingDialogIfNeverRated` is a suspend function over two injected repositories; the already-rated, never-rated and already-prompted branches are three straightforward MockK cases, and the user-scoping bug above is exactly the kind a test would have caught.
- `RatingsRepository` and `UserRepository` are injected straight into the Fragment while `ResourceViewerViewModel` sits right there holding `resourcesRepository`. The has-rated lookup belongs in the ViewModel; the Fragment should only be showing the dialog.

**Minor:** `if (!::library.isInitialized) return` runs in both the `STATE_ENDED` branch and again at the top of `showResourceRatingDialogIfNeverRated` — the second is unreachable-but-harmless. And the commit messages (`fix1`, `fixes`, `fix 1`, `fixed 1`) won't tell anyone anything after a squash; worth a real subject line.

**Worth knowing:** master now carries the equivalent course-side flow via #15442 — `TakeCourseFragment.showCourseRatingDialogAndFinish` uses the same `getRatingSummary` plus `userRating`/`existingRating` check. That's two implementations of "has this user rated this thing yet, and if not, prompt". Once this lands, extracting the has-rated test and the prompt-once bookkeeping into one place (a small use case, or `RatingsRepository` itself) would keep them from drifting — and would give the user-scoping fix a single home.

**Verdict:** `change`. The user-scoped key and the two dead imports are the must-fixes; moving the flag into `SharedPrefManager` and marking it after the dialog actually shows are the ones I'd do at the same time.

---

dogi:
**Re-review — Rating: 65/100** (was 68)

The restructure is the right one — the has-rated decision moved into `ResourceViewerViewModel`, the prompt key is now scoped per user (`rating_prompted_${userId}_$resourceId`), and there are tests. Both of my main structural findings from last round are addressed. But CI is red and there's a variable mix-up that undoes the per-user fix.

**`myPlanet test (default)` is failing, and I think I know why.**

```kotlin
val mockUser = mockk<UserEntity> { every { id } returns userId }
```

`UserEntity.id` is declared `@PrimaryKey @JvmField var id: String = ""` — a Java field, not a getter. MockK stubs method calls, so `every { id }` has no invocation to intercept and the stub never registers.

This is the exact trap that two comments in `ResourcesViewModelTest` used to document, before #15770 deleted them:

> `MyLibrary` is a Room entity whose id is a `@JvmField` … neither can be stubbed with mockk; use a real instance.

The fix is the same pattern the rest of the suite uses:

```kotlin
val user = UserEntity().apply { id = userId }
coEvery { userRepository.getUserModel() } returns user
```

**The Fragment checks one id and records another.**

```kotlin
val rid = library.resourceId ?: return
    val showDialog = viewModel.showResourceRatingDialogIfNeverRated(rid)
    if (showDialog) {
        val userId = userRepository.getUserModel()?.id
        val ratingDialog = RatingsFragment.newInstance("resource", resourceId, library.title)
        ratingDialog.show(parentFragmentManager, RatingsFragment.TAG)
        viewModel.setRatingPrompted(userId, resourceId)
    }
}
```

The "have we prompted?" lookup uses `rid` (`library.resourceId`); the dialog and the "mark as prompted" write use `resourceId`, which is a different property — the Fragment's own field, read from `ARG_RESOURCE_ID` at line 151. When those two differ the flag is written under a key that's never read, so the prompt reappears on every playback forever, and the dialog opens against the wrong id. Use one value for all three.

**Two dead constants.** The Fragment companion still declares

```kotlin
private const val PREFS_NAME = "rating_prompt_prefs"
private const val KEY_RATING_PROMPT_PREFIX = "prompted_"
```

but the Fragment no longer touches `SharedPreferences` — the ViewModel owns that now with its own `RATING_PROMPT_PREFIX = "rating_prompted_"`. Both are unreferenced.

**The function name promises something it doesn't do.** `showResourceRatingDialogIfNeverRated` returns a `Boolean` and shows nothing. `shouldPromptForRating` (or `shouldShowResourceRatingDialog`) describes the contract, and it matters here because the caller has to remember to do the showing *and* the marking.

**Dead branch inside it:**

```kotlin
val userId = userRepository.getUserModel()?.id ?: return false
…
val hasRated = if (!userId.isNullOrEmpty()) { … } else { false }
```

`userId` is already a non-null `String` after the elvis return, so `isNullOrEmpty()` is a redundant null check the compiler will warn about, and the `else` branch is only reachable for the empty string. Either fold the emptiness into the elvis (`?.id?.takeIf { it.isNotEmpty() } ?: return false`) or drop the branch.

And the comment above the return describes behaviour that lives elsewhere:

```kotlin
// Mark as prompted after dialog is shown, even if user chooses not to rate. Respects user decision.
return !hasRated
```

Nothing in this function marks anything.

**Storage nits.** `getRawString(key, "false") == "true"` / `setRawString(key, "true")` stores a boolean as a string when `SharedPrefManager` already has typed boolean accessors. And the unbounded key growth I raised last round is unchanged — one entry per (user, resource) pair, never cleared, including on logout or clear-data.

**Two mechanical things.** The test file lives at `app/src/test/java/…/ui/resources/ResourceViewerViewModelTest.kt` but declares `package org.ole.planet.myplanet.ui.viewer`. Gradle doesn't care, but the project convention is to mirror the main package path, and it makes the file hard to find. Separately, `ResourceViewerViewModel.kt` conflicts with master — `b0b6b5b` (#15658) and `03ebe3a` (#15587) both landed there since this branch was cut. Dropping the now-unused `dispatcherProvider` constructor parameter is a legitimate cleanup (I checked — it had no other reference on master), just make sure the merge keeps master's changes.

**Verdict:** `change` — fix the `UserEntity` mock to get CI green, and unify `rid`/`resourceId` so the per-user key you added actually gets used. The rest are cleanups.

---

dogi:
**Re-review — Rating: 84/100** (was 65)

Both blockers are gone and CI is green on all three checks.

The id mix-up is resolved — `rid` is now used for the check, the dialog and the flag write:

```kotlin
val rid = library.resourceId ?: return
…
val showDialog = viewModel.shouldShowResourceRatingDialog(rid)
if (showDialog) {
    val ratingDialog = RatingsFragment.newInstance("resource", rid, library.title)
    …
    viewModel.setRatingPrompted(userId, rid)
}
```

so the per-user key you added is now written under the same id it's read under. The two dead constants (`PREFS_NAME`, `KEY_RATING_PROMPT_PREFIX`) are removed, the function is renamed to `shouldShowResourceRatingDialog` — which is what it does — and the misplaced comment about marking-after-showing is gone.

The test is green and moved to `ui/viewer/` so path and package agree. Worth noting for future reference why it was failing: `UserEntity.id` is `@PrimaryKey @JvmField var id: String`, a Java field with no getter, so `mockk<UserEntity> { every { id } returns … }` has no call to intercept. That's the same trap the two comments in `ResourcesViewModelTest` documented before #15770 deleted them — worth restoring those when that branch comes back around.

**One dead branch left.** The elvis now guarantees a non-blank id:

```kotlin
val userId = userRepository.getUserModel()?.id?.takeIf { it.isNotBlank() } ?: return false
…
val hasRated = if (!userId.isNullOrEmpty()) { … } else { false }
```

`userId` is a non-null, non-blank `String` by that point, so `isNullOrEmpty()` is a redundant null check the compiler will warn about and the `else` branch is unreachable. Collapse it to the `try` block.

**Two things I raised earlier that are still open, neither blocking:**

The prompt flag is stored as a string — `getRawString(key, "false") == "true"` / `setRawString(key, "true")` — where `SharedPrefManager` already has typed boolean accessors. And the key space still grows without bound: one `rating_prompted_${userId}_$resourceId` entry per pair, never cleared on logout or clear-data. Neither hurts today; both are the kind of thing that's much cheaper to fix now than after a year of entries.

**You'll need to merge master in.** `ResourceViewerViewModel.kt` conflicts — `b0b6b5b` (#15658) and `03ebe3a` (#15587) both landed there since this branch was cut. Dropping the `dispatcherProvider` constructor parameter is legitimate (I checked, it had no other reference), just make sure the resolution keeps master's changes to `ensureServerUrlUpdated`.

**Verdict:** `ready`. Collapse the dead branch and merge master in — neither needs another review round.

---

ragilzakaria (dismissed):

Okuro3499 (changes requested):
@J-S-webskas 
I am not able to trigger this for image resource
https://github.com/user-attachments/assets/3b2c1d47-166e-48e2-a143-ed77d6a4b71c

J-S-webskas:
FYI: The rating dialog only shows when user is finished with the resource AND exits AND has not rated before/dialog has not popped up before

dogi:
**Review — rating: 62/100**

The feature is worth having and the three ViewModel tests cover the right cases — already-rated, never-rated, already-prompted — with the SharedPreferences key format pinned. `Player.STATE_ENDED` is exactly the right completion signal for audio and video. What holds this back is a fair amount of scaffolding that doesn't do anything yet, plus one heuristic that inverts the feature's premise.

**The PDF path marks every PDF finished on open.** `onPdfPageChanged(0, 1)` runs right after the first page renders, and with `totalPages = 1` the `currentPage == totalPages - 1` test passes immediately. Open a 200-page PDF, press back one second later, and you get a "rate this resource" prompt for something you didn't read. `onPdfPageChanged` is written for a paged viewer that doesn't exist here — until it does, the honest options are to leave PDFs out of the prompt or to gate on a real signal like scroll extent.

**Five new fields are written but never read.** `pdfTotalPages`, `pdfCurrentPage`, `hasPromptedPdfRating`, `imageStartTime`, and `hasPromptedImageRating` are all introduced by this PR (none exist on master) and none is ever read back. `import kotlinx.coroutines.delay` is added and `delay` never called. Kotlin only warns on these, so CI stays green — but it's ~8 lines of state implying a time-on-image or per-type prompt-tracking design that isn't implemented. Either finish it or drop it.

**The image coroutine is a no-op that adds a race.** `imageRatingJob = viewLifecycleOwner.lifecycleScope.launch { if (type == ResourceType.IMAGE) isResourceFinished = true }` contains no suspension point, so the `launch` buys nothing — and because `lifecycleScope.launch` on Main dispatches to a later loop turn, a fast back press can read `isResourceFinished == false` before it runs. A plain assignment is both simpler and correct.

**An always-true branch in `shouldShowResourceRatingDialog`.** `userId` is established non-null and non-blank by the `?: return false` on the line above, so `if (!userId.isNullOrEmpty())` can never be false and the `else { false }` is unreachable.

**The user is fetched twice per back press.** `handleBackNavigation` calls `userRepository.getUserModel()` for `isRatingPrompted`, then `shouldShowResourceRatingDialog` calls it again internally. Passing the resolved `userId` in would halve the work and remove the possibility of the two reads disagreeing.

**Two conventions this would be the first to break.** `jakarta.inject.Inject` in `ResourceViewerActivity` and `WebViewActivity`: the repo has **0** files using `jakarta.inject` and **148** using `javax.inject.Inject` — including `ResourceViewerFragment`, which this same PR edits. And `junit.framework.TestCase.assertTrue/assertFalse` in the new test is JUnit 3; no test file in the repo uses `junit.framework` today, the house import is `org.junit.Assert`.

**Nits:** the new test's `import androidx.annotation.OptIn` is the AndroidX opt-in annotation, which doesn't satisfy Kotlin's `@RequiresOptIn` on `ExperimentalCoroutinesApi` — you want `kotlin.OptIn`. In `WebViewActivity`'s back callback, `isEnabled = false` before `handleBackNavigation()` is now vestigial and can leave the second back press bypassing the dialog. Neither activity guards against `handleBackNavigation()` running twice, so two quick back presses can show two dialogs. And the prompt flag writes one SharedPreferences key per (user, resource) forever — worth a thought about growth on a device with a large library.

**Verdict:** genuinely useful feature with real tests, but the PDF completion check fires on open, and there's a layer of unused state and a no-op coroutine to clear out first. Also worth rebasing — this branch predates the test sharding, so it's still running the old unsharded job.

---

Okuro3499 (dismissed):

dogi:
**Re-review — Rating: 74/100** (was 62)

Every finding from the last round is genuinely gone, not papered over. The `onPdfPageChanged(0, 1)` call that marked every PDF finished the moment it opened is deleted rather than tweaked; the five write-only fields are gone; the no-op coroutine that raced a fast back press is now a real `handleBackNavigation` with a re-entrancy guard; `jakarta.inject` is back to `javax.inject` (still 0 vs 148 in the repo the other way); and the test is JUnit4 + MockK + `MainDispatcherRule` + `TestDispatcherProvider`, which is exactly the shared infra `docs/TESTING.md` points at. The three tests cover the three real branches of `shouldShowResourceRatingDialog` — already-rated, never-rated, already-prompted — and persisting the prompted flag means dismissing without rating doesn't nag on every subsequent open. I checked that `RESOURCE_ID` is actually populated (`BaseContainerFragment.kt:201`) and `RESOURCE_TITLE` too (`ResourceOpener.kt:45`, `IntentUtils.kt:15`), so neither rating path is dead code, and `setRawString` goes through `pref.edit { }`, which defaults to `apply()` — no main-thread disk write.

**Blocking: `dialog.show()` can throw after `onSaveInstanceState`.** Both copies do `lifecycleScope.launch { ... userRepository.getUserModel() ... viewModel.shouldShowResourceRatingDialog(...) ... dialog.show(supportFragmentManager, RatingsFragment.TAG) }`. `lifecycleScope` only cancels at `onDestroy`, not at `onStop`, so if the user presses back and then backgrounds the app while those two suspend calls are in flight, the coroutine resumes after the state is saved and `show()` throws `IllegalStateException: Can not perform this action after onSaveInstanceState`. Narrow window, but it's a hard crash on the way out of a resource. Guard with `if (supportFragmentManager.isStateSaved) { finish(); return@launch }` before showing (or gate the whole body on `Lifecycle.State.STARTED`).

Two behavioural gaps worth confirming are intentional. `isResourceFinished` is only ever set for video `STATE_ENDED` and for images (immediately, on display) — **PDFs and text resources never mark finished, so they never prompt.** The old bug over-triggered on PDFs; this correction lands on never-triggering for what's probably the most common resource type. Separately, neither `isResourceFinished` nor `RatingsFragment`'s `dismissListener` survives a configuration change (the listener is a plain field, and the fragment is recreated by the FragmentManager), so rotating with the dialog up means dismissing it no longer calls `finish()`. Not stuck — `backNavigationHandled` resets with the activity and a second back press finishes, since the prompted flag is already persisted — but it reads as a dead dismiss button.

**Nit:** switching `initializeAudioPlayer` to `createExoPlayer()` quietly gives audio the video listener. That's mostly an upgrade (audio had no error handling at all), but an audio failure now shows `R.string.video_playback_error`, and audio gets `AUDIO_CONTENT_TYPE_MOVIE`. `videoLoadingOverlay`/`videoLoadingText` are nullable so the loading calls are safe no-ops in the audio layout — I checked. Also: `handleBackNavigation` is ~25 lines duplicated verbatim across the two activities, the caller's `isRatingPrompted` check is repeated inside `shouldShowResourceRatingDialog`, `setRatingPrompted` runs after `show()` rather than before, and the prefs store grows one key per (user, resource) forever.

**Verdict:** a real rework that clears everything from last round — add the `isStateSaved` guard and confirm the PDF/text gap is deliberate, and this is ready.

---

dogi:
## Review — rating 62/100

Good feature idea, and the Android-lifecycle details are handled better than most PRs of this shape — `!supportFragmentManager.isStateSaved` before `show()` avoids the classic `IllegalStateException`, the null-user path is covered, and the tests use `MainDispatcherRule`/`TestDispatcherProvider` per the repo's conventions. Three things need addressing before I would merge it.

### 1. Unbounded SharedPreferences growth

```kotlin
fun setRatingPrompted(userId: String?, resourceId: String?) {
    val key = "${RATING_PROMPT_PREFIX}${userId}_$resourceId"
    sharedPrefManager.setRawString(key, "true")
}
```

`setRawString` writes into the app's main preferences file (`SharedPrefManager.kt:273-274`), so this adds **one permanent key per (user, resource) pair** with no cleanup path. SharedPreferences is parsed fully into memory on first access and the whole file is rewritten on every `apply()`. On a device where a user opens a few hundred resources — the normal case for this app — that is a steadily growing XML file being re-read at every startup and rewritten on every unrelated preference write.

This state belongs in Room. The project already has a `rating` table and `RatingsRepository` keyed by exactly `(type, item, user)`, which is the same key you are constructing by string concatenation here. A `prompted` column on that table, or a small companion table, gives you bounded storage, a real query, and it disappears with the user's data on the drop-and-resync path instead of lingering forever.

### 2. `isResourceFinished` does not mean finished

Only the media path actually waits for completion:

```kotlin
Player.STATE_ENDED -> { isResourceFinished = true }
```

For PDF, image and text it is set inside the *setup* function, at open time:

```kotlin
val pdfFileName = binding.root.findViewById<TextView>(R.id.pdfFileName)
pdfFileName.text = title
isResourceFinished = true      // before renderPdf() has even run
```

So for three of the four resource types the flag means "opened", and the user gets a rating prompt after glancing at a PDF for one second and backing out. `WebViewActivity` does not consult the flag at all — it prompts whenever `RESOURCE_ID` is present.

Either the flag should mean what it says (a scroll-to-end check for PDF/text, a nontrivial dwell time for images), or the feature should be described as "prompt on leaving a resource" and the flag dropped. What it should not do is claim one thing and implement another across four inconsistent paths — that inconsistency will read as a bug to whoever touches this next.

### 3. `backNavigationHandled` can strand the user

```kotlin
private fun handleBackNavigation() {
    if (backNavigationHandled) return
    ...
    backNavigationHandled = true
    lifecycleScope.launch {
        val userId = userRepository.getUserModel()?.id?.takeIf { it.isNotBlank() }
```

The flag is set before an async DB read. While that read is in flight, every further back press returns immediately and does nothing — the activity looks frozen. If `getUserModel()` is slow or stalls, back is permanently dead and the only exit is the task switcher. Since the whole point is a *non-essential* prompt, this should fail toward letting the user leave: resolve the user id before committing to the flag, or race the lookup against a short timeout and `finish()` if it loses.

### Smaller items

- `handleBackNavigation()` is ~25 lines duplicated near-verbatim between `ResourceViewerActivity` and `WebViewActivity`. Worth extracting once the semantics above are settled, since right now the two copies already disagree about the finished check.
- `shouldShowResourceRatingDialog` calls `isRatingPrompted` internally, and both call sites check it again first. Harmless, but pick one.
- The `catch (Exception) { false }` in `shouldShowResourceRatingDialog` makes `hasRated` false, so `return !hasRated` is `true` — a ratings-DB failure *shows* the prompt. Probably the right direction for a non-critical dialog, but it is untested and worth a comment saying it is deliberate.
- `setRatingPrompted` runs after `dialog.show()`; if the process dies in between the user gets prompted twice. Set it first.
- New imports in `ResourceViewerActivity` are appended after `EdgeToEdgeUtils` rather than in the file's existing alphabetical order.

Moving to `change`.

---

dogi:
## Review — rating 50/100 (was 62)

The storage rework is exactly what I asked for, and then the change that made it possible is missing its version bump.

### The Room migration is the right design

`RatingPrompt` with a composite primary key on `(userId, item, type)` replaces the unbounded `rating_prompted_<user>_<resource>` SharedPreferences keys. Bounded, queryable, keyed the same way `Rating` already is, and it disappears correctly on the drop-and-resync path. `isRatingPrompted` as an `EXISTS` query is the right shape. Dropping the now-unused `dispatcherProvider` from `RatingsRepositoryImpl` (and updating its test) is a good tidy-up along the way.

Two smaller notes from last time also handled: `setRatingPrompted` now runs before `dialog.show()`, and `backNavigationHandled` is set after the user lookup with a re-check inside the coroutine, so a null user no longer strands the back button.

### Blocking: `AppDatabase.version` is not bumped

```diff
         Rating::class,
+        RatingPrompt::class,
         TagEntity::class,
```

`AppDatabase.kt` has exactly two hunks in this diff — the import and the entity list. `version = 9` sits at line 128 and is untouched.

Adding an entity changes the schema Room hashes into `room_master_table`. On open, `checkIdentity` compares that hash and throws when it differs at an unchanged version:

> `IllegalStateException: Room cannot verify the data integrity. Looks like you've changed schema but forgot to update the version number.`

`fallbackToDestructiveMigration(true)` does **not** cover this — it lives in `onUpgrade`/`onDowngrade` and only runs when the version number itself changes. So every device with an existing `myplanet_room.db` crashes on first database access after this ships, while a fresh install works fine. That asymmetry is exactly why it survives testing on a wiped emulator.

One line: `version = 9` → `10`. Worth noting #15951 has the identical defect, so if both land in the same release only one bump is needed between them — but each needs to be correct standing alone.

### The "already prompted" test does not test prompting

```kotlin
every { sharedPrefManager.getRawString("rating_prompted_${userId}_${resourceId}", "false") } returns "true"
val result = viewModel.shouldShowResourceRatingDialog(userId, resourceId)
assertFalse(result)
```

`sharedPrefManager` is no longer a constructor argument of `ResourceViewerViewModel` — it is a leftover field in the test that is never passed in, so this stub is dead. The production path now calls `ratingsRepository.isRatingPrompted(...)`, which this test never stubs; the relaxed mock returns `false`, so execution falls through to `getRatingSummary`, which is also unstubbed here and returns a relaxed `RatingSummary` whose `userRating` is non-null. That makes `hasRated` true and the assertion passes — via the *already-rated* branch, not the already-prompted one.

So the assertion is green while the behaviour it names is uncovered, and nothing exercises the new Room-backed `isRatingPrompted` at all. The fix is to stub what the code actually calls:

```kotlin
coEvery { ratingsRepository.isRatingPrompted(userId, resourceId) } returns true
val result = viewModel.shouldShowResourceRatingDialog(userId, resourceId)
assertFalse(result)
coVerify(exactly = 0) { ratingsRepository.getRatingSummary(any(), any(), any()) }  // short-circuits
```

and to delete the `sharedPrefManager` field from the test. A `RatingDao` round-trip test for `setRatingPrompted`/`isRatingPrompted` against an in-memory database would be worth adding too, in the style of `NewsDaoTest`.

### Still open from last time

`WebViewActivity.handleBackNavigation` prompts whenever `RESOURCE_ID` is present and never consults `isResourceFinished`, so web resources behave differently from every other type. And the ~30-line `handleBackNavigation` is still duplicated verbatim across the two activities, with that one behavioural divergence between the copies — which is precisely the failure mode duplication invites.

Crediting the comments on the PDF/text paths (*"Currently, cannot detect scroll-to-bottom"*) — documenting the limitation honestly is much better than the flag silently meaning "opened".

Nit: `RatingPrompt.kt` has no trailing newline.

Moving to `change` — the version bump is the blocker.

---

dogi:
## Re-review — rating 74/100 → `change`

**The blocker is fixed.** `AppDatabase` now goes `version = 9 → 10` alongside the new `RatingPrompt` entity, so the identity-hash mismatch that would have thrown `Room cannot verify the data integrity` on every existing install is gone — destructive fallback fires on the version change and the DB rebuilds cleanly.

The test rewrite also fixes the second problem I raised. The old "already prompted" test stubbed a `SharedPrefManager` the ViewModel no longer took, so it passed through the already-*rated* branch by accident. `ResourceViewerViewModelTest` now covers the three real branches — prompted → false with `coVerify(exactly = 0)` proving the summary query is skipped, rated → false, neither → true. That last `coVerify` is the one that would catch a regression to the old accidental path.

The `RatingPrompt` design remains right: a composite `["userId", "item", "type"]` primary key with `OnConflictStrategy.REPLACE` is the correct shape for "has this user been asked about this item", and it retires the unbounded SharedPreferences key growth. CI green on all five checks.

Four things before this lands.

### `equals`-free, but the import ordering is visibly wrong in two files

`AppDatabase.kt` inserts `import org.ole.planet.myplanet.model.RatingPrompt` between `androidx.room.Database` and `androidx.room.RoomDatabase`, and `ResourceViewerActivity.kt` appends `androidx.activity.OnBackPressedCallback`, `androidx.activity.viewModels`, `javax.inject.Inject` and `org.ole.planet.myplanet.repository.UserRepository` *after* the `org.ole.planet.myplanet.utils.EdgeToEdgeUtils` line. The repo's `kotlin-importing` skill sorts both in one pass.

### `handleBackNavigation` is copy-pasted into two activities

The ~30-line block in `ResourceViewerActivity` and the one in `WebViewActivity` are the same logic — resolve user id, guard `backNavigationHandled`, ask `shouldShowResourceRatingDialog`, show the dialog with a dismiss listener, mark prompted, else finish. Two copies means two places to fix when the prompt policy changes. A shared extension or a small base class would carry it once.

### "Finished" means "opened" for three of the four viewer types

```kotlin
// PDF has only one page. Prompt upon leaving for now. Currently, cannot detect scroll-to-bottom
isResourceFinished = true
```

The same unconditional assignment appears for images and text, and `WebViewActivity` doesn't consult the flag at all — it prompts whenever `resourceId` is non-blank. Only video has a real completion signal (`Player.STATE_ENDED`).

So in practice the feature is "prompt on first exit from any resource", not "prompt when finished". A learner who opens a PDF by mistake and immediately backs out gets a rating dialog for something they never read — and because `setRatingPrompted` runs before the dialog is shown, that's their one and only chance to be asked about that resource. The comments are honest about the limitation, which I appreciate, but it's worth deciding whether prompting on open is the behaviour you want to ship, or whether PDFs/text should gate on something cheap like a scroll-position check or a dwell time.

### `catch (e: Exception)` swallows cancellation

```kotlin
val hasRated = try {
    val summary = ratingsRepository.getRatingSummary("resource", resourceId, userId)
    ...
} catch (e: Exception) { false }
```

`shouldShowResourceRatingDialog` is a suspend function called from `lifecycleScope`; `CancellationException` is an `Exception`. Backing out while the summary query is in flight is absorbed here and reported as "not rated", so the dialog can be shown on a finishing activity. A `catch (e: CancellationException) { throw e }` arm ahead of it matches what #16630, #16655 and #16658 settled on.

Minor: dropping `dispatcherProvider` from `RatingsRepositoryImpl`'s constructor is a fine cleanup but unrelated to this feature — worth a line in the description so reviewers don't have to work out whether it changed threading behaviour (it doesn't; Room's suspend DAOs dispatch themselves).

---

dogi:
## Re-review — rating 74/100 → `change`

The import ordering is fixed in `AppDatabase.kt` — `RatingPrompt` now sits between `Rating` and `RemovedLog` where it belongs. `ResourceViewerActivity.kt` is mostly sorted too, with one left over: `org.ole.planet.myplanet.repository.UserRepository` sits above `org.ole.planet.myplanet.databinding.ActivityResourceViewerBinding`, and `databinding` sorts before `repository`.

Beyond that, this revision is byte-identical to the one I looked at last round, so the three substantive points from that review are all still open. Rather than restate them at length, here's what I think the minimum is:

**One line, and I'd move this to `ready`:**

```kotlin
val hasRated = try {
    val summary = ratingsRepository.getRatingSummary("resource", resourceId, userId)
    summary.userRating != null || summary.existingRating != null
} catch (e: CancellationException) {   // ← add this arm
    throw e
} catch (e: Exception) {
    false
}
```

`shouldShowResourceRatingDialog` is called from `lifecycleScope`, so backing out while the summary query is in flight currently reports "not rated" and can put a dialog on a finishing activity. This is the same taxonomy #16630, #16655, #16658 and #16675 have all now adopted.

**The other two are judgement calls I'd like recorded rather than fixed here, if you disagree with them:**

- The ~30-line `handleBackNavigation` is duplicated verbatim in `ResourceViewerActivity` and `WebViewActivity`. Not a defect, but the prompt policy now lives in two places.
- `isResourceFinished = true` is set unconditionally for PDF, image and text, and `WebViewActivity` doesn't consult the flag at all — so for four of the five resource types the feature is "prompt on first exit", not "prompt when finished". Your comments say as much (*"cannot detect scroll-to-bottom"*), and combined with `setRatingPrompted` firing before the dialog shows, a user who opens a PDF by accident spends their one prompt for that resource. If that's an accepted tradeoff for now, saying so in the PR description closes the question; if it isn't, a dwell-time or scroll-position gate on those three types would.

The Room work remains the strong part — `version = 10` alongside the new entity, the composite `["userId", "item", "type"]` key, and the three ViewModel tests covering prompted / rated / neither with `coVerify(exactly = 0)` on the skipped summary query. CI green on all five checks.

---

dogi:

Both actionable points are fixed, and the second one better than I proposed.

**Cancellation propagates.** The `catch (e: CancellationException) { throw e }` arm is in place ahead of the generic handler, so backing out mid-summary-query no longer reports "not rated" and can no longer put a dialog on a finishing activity.

**The duplication is gone, extracted rather than merely deduplicated.** I'd suggested a shared helper; `ResourceRatingPrompter` is a better answer than that — it makes the prompt policy a named object with a KDoc that states the rule ("when a learner leaves a finished resource, prompt them to rate it once, then finish"), owns the `handled` flag, and leaves both activities with a three-line delegation. The policy now has one home instead of being an emergent property of two copies.

Two details in it that are right and easy to get wrong:

- **`by lazy` on the prompter.** `userRepository` is `@Inject lateinit`, populated during `super.onCreate()`. A direct field initialiser would have read it before injection; deferring construction to the first `handleBackNavigation()` call avoids that entirely.
- **The double-tap guard still holds.** `handled` is only set after the suspending `getUserModel()`, so two fast back presses can both enter the coroutine — but the second `if (handled) return@launch` catches it before the dialog is shown. One dialog, one `setRatingPrompted`.

Making `isResourceFinished` a parameter with a `true` default is also a quiet improvement: `WebViewActivity` not passing it now reads as an explicit "HTML always counts as finished" rather than an omission, which is the tradeoff I asked to see recorded. Combined with the `// cannot detect scroll-to-bottom` comments on the PDF and text paths, the limitation is documented where someone will find it.

Import ordering in `ResourceViewerActivity` is correct now. CI green on all five checks.

Three small things left, none blocking:

- **`ResourceViewerFragment` still injects `UserRepository` and never uses it.** That field was added when the prompt logic lived in the fragment's orbit; now that `ResourceRatingPrompter` resolves the user in the activity, the fragment's `@Inject lateinit var userRepository` has no readers.
- **`WebViewActivity` import order**: `javax.inject.Inject` landed above `java.io.File`, and `java` sorts before `javax`.
- **Nothing tests `ResourceRatingPrompter`.** Most of it is activity-lifecycle glue and the decision logic it calls is covered by `ResourceViewerViewModelTest`, so this is low value — but the ordering it encodes (mark prompted *before* showing, finish on dismiss) is the kind of thing a future edit could invert without any test noticing.

---

Okuro3499 (approved):

dogi:
## Re-review — rating 86/100 (was 84) → `merge`

Four commits since the last round: three renames and the master merge. Both of the small things I left open are closed, and the merge — the only part that could have gone wrong — resolves correctly.

**The merge with #16625 keeps both sides.** This was the risk: master's viewer playback rework and this branch's rating prompt both touch `Player.STATE_ENDED`, both add a `ResourceViewerViewModelTest`. All three conflicts resolved the way they should be:

- `STATE_ENDED` now does both jobs — `isResourceFinished = true` from this branch, then `lastSavedPositionMs = 0L` / `savePlaybackProgress(getMediaKey(), 0L)` from master. Neither behaviour is dropped.
- The ViewModel carries master's playback accessors *and* this branch's three rating functions, not one set at the other's expense.
- The test class is genuinely merged — master's four `calculateEffectivePlaybackPosition` / playback-speed tests sit alongside this branch's three rating-prompt tests, with `ratingsRepository` threaded into the shared constructor call. Seven tests in one class, none lost.

**Both open nits are fixed.** `ResourceViewerFragment` no longer injects the `UserRepository` it never read, and `javax.inject.Inject` now sorts below `java.io.File` in `WebViewActivity`.

**The renames are the right calls.** `RatingPromptLog` matches the `RemovedLog` convention for what the row actually is — one record per (userId, item, type) that a prompt was shown — and `ResourcesExitCoordinator` names the real responsibility: every branch of `handleBackNavigation` finishes the activity, and prompting is the conditional part. That matches `UploadCoordinator` / `ResourceDownloadCoordinator`. Table renamed to match at `version = 11`.

**Dropping `dispatcherProvider` from `RatingsRepositoryImpl` is safe** — I checked master directly: it's declared on line 18 and referenced nowhere in the file. No `withContext` was removed with it, and the Room suspend DAOs handle their own dispatcher, so this is dead-parameter removal, not a discipline regression.

CI green on all five checks.

**Nit:** `RatingDao.isRatingPrompted(..., type: String = "resource")` carries a default that no caller uses — `RatingsRepositoryImpl` passes all three arguments explicitly, and `"resource"` is then hardcoded a third time in `setRatingPrompted`. A shared constant would leave one place to change. Also, `ResourceViewerActivity` now holds an activity-scoped `ResourceViewerViewModel` while `ResourceViewerFragment` keeps its own fragment-scoped one — two instances on the same screen. Harmless, since the prompt state lives in Room rather than in either instance, but worth knowing it's two. And `ResourcesExitCoordinator` still has no test of its own; the ordering it encodes (mark prompted *before* showing, finish on dismiss) is the kind of thing a later edit could invert silently.

**Verdict:** mergeable — Okuro3499 approved on the current head and the PR carries ⭐⭐⭐, so this goes to `merge`.

---

dogi (approved):

PR #15768, opened 2026-08-17
enterprises: smoother finances date filter resetting (fixes #15767)

### LINKED ISSUE #15767: Date filter never resets
Date filter never resets after leaving the finances page it should reset when this happens
https://github.com/user-attachments/assets/2f6514bf-8f83-4d84-98f4-8c185c4f2572

ragilzakaria:
https://github.com/user-attachments/assets/cf0fc2c2-80d4-4480-ab37-f9ab4c8f48d5

fixes #15767 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved finance transaction filtering and sorting reset behavior.
  * Date filters, end-date controls, rotation state, and sort order now reset consistently.
  * Finance transactions refresh when returning to the screen.


Okuro3499 (approved):

dogi:
**Review — rating: 55/100**

The extraction is clean, but I don't think either of the two new hooks fires in this app, which would mean #15767 is still open after this merges.

**`EnterprisesFinancesFragment` lives in ViewPager2, and neither override is a ViewPager2 signal.**

Both hosts build it through a `FragmentStateAdapter`:

- `CommunityPagerAdapter : FragmentStateAdapter(fm)` — position 4
- `TeamPageConfig` → `TeamPagerAdapter : FragmentStateAdapter(parentFragment)`

`setMenuVisibility` is the ViewPager**1** mechanism — it's what `FragmentPagerAdapter`/`FragmentStatePagerAdapter` called to mark the primary item, and it's deprecated alongside them. ViewPager2's `FragmentStateAdapter` drives page transitions with `FragmentTransaction.setMaxLifecycle` instead and never calls it. `onHiddenChanged` is narrower still: it only fires for explicit `FragmentTransaction.hide()`/`show()`, which nothing in this pager does.

That leaves `onDestroyView`, and the reset there is a no-op in effect — `currentStartDate`, `currentEndDate` and `isAsc` are plain fields that aren't written to `onSaveInstanceState`, so if the fragment is destroyed they're gone regardless. And in the case the issue is actually about, `onDestroyView` doesn't run at all: swiping from Finances to the adjacent Reports page keeps the fragment and its view alive inside the offscreen window, so the user comes back to the filter they left. That's the reported bug, and none of the three hooks touches it.

Worth knowing this isn't a novel mistake in the codebase — `CourseStepFragment:266` overrides `setMenuVisibility` too, and it also sits under a ViewPager2 (`TakeCourseFragment:131`). Whatever that override was meant to do is probably not happening either.

**Two hooks that do work here.** `onPause()` is the straightforward one: `setMaxLifecycle` drops a page out of `RESUMED` when it stops being the primary item, so `onPause` is exactly "the user left this page" and it fires for swipes, tab taps, and backgrounding alike. The other is already in the codebase — `TeamDetailFragment:221` calls `binding.viewPager2.registerOnPageChangeCallback(...)`, so the host has a page-change signal you could hang this off. `onPause` keeps the logic in the fragment that owns the state, which I'd prefer.

Please verify before the next round rather than taking my word for it: drop a log line in each of the three overrides, swipe away from Finances and back, and see which ones print.

**Reloading data on the way out is backwards.** Both new hooks call `observeTransactions(...)` immediately after resetting, so the moment the user leaves the page it kicks off a query for a screen they're no longer looking at. Resetting the state is enough — the reload belongs on the way back in, where `onResume` (or the existing `onViewCreated` path) already handles it.

**The explicit arguments are redundant.** `observeTransactions` defaults are:

```kotlin
private fun observeTransactions(
    sortAscending: Boolean = isAsc,
    startDate: Long? = currentStartDate,
    endDate: Long? = currentEndDate,
)
```

and `resetDateFilter()` has just set all three fields to `false`/`null`/`null`. So `observeTransactions(sortAscending = isAsc, startDate = null, endDate = null)` is the same call as `observeTransactions()`. That line is repeated three times; collapsing it — or folding it into a single `resetAndReload()` — removes the duplication and the chance of the two drifting.

**Naming:** `resetDateFilter()` also resets `isAsc` and `imgDate.rotation`, which are sort state, not date-filter state. Either widen the name or split the sort reset out — the `btnReset` button arguably should reset both, but a reader of `onDestroyView` won't expect sort direction in there.

**What's good:** pulling the seven inline statements out of the `btnReset` listener is the right move regardless of how the leave-detection lands, and guarding the view touches with `_binding?.let { }` while keeping the non-view field resets outside it is exactly right — the fields still reset even when the view is gone.

**Verdict:** `change`. The refactor can stay as-is. Replace `setMenuVisibility`/`onHiddenChanged` with `onPause()`, drop the reload from the leave path, and collapse the three identical `observeTransactions(...)` calls.

---

dogi:
**Re-review — Rating: 82/100** (was 55)

Every structural point landed. `setMenuVisibility` and `onHiddenChanged` are gone — both were ViewPager1 / explicit-hide signals that never fire for a fragment built by `FragmentStateAdapter`, which is what `CommunityPagerAdapter` and `TeamPagerAdapter` both are. `onPause()` is the hook that actually works here: `setMaxLifecycle` drops a page out of `RESUMED` the moment it stops being the primary item, so a swipe away triggers it and a swipe back triggers `onResume`.

The reload moved to the right side too. Last round both leave-hooks called `observeTransactions(...)` immediately after resetting, kicking off a query for a screen the user had just left; now `onPause` only resets state and `onResume` does the reload. And `resetFilterAndSort` is the honest name for something that clears `isAsc` and the chevron rotation alongside the dates. The three redundant explicit arguments are gone as well — `observeTransactions()` with its defaults is the same call now that the fields are already nulled.

**`onDestroyView`'s reset is dead now.** `onPause` always runs before `onDestroyView`, so the state is already cleared by the time it's reached — and the fields are plain properties that die with the fragment regardless. Harmless, but it's one more place to keep in sync for no benefit.

**First display runs the query twice.** `onViewCreated` ends with `observeTransactions()` (line 244) and `onResume` follows immediately with another (line 365). I checked whether that stacks collectors, and it doesn't — `EnterprisesFinancesViewModel.getTeamTransactions` opens with `transactionsJob?.cancel()`, so the second call cancels the first and only one flow survives. So this is a wasted round-trip rather than a bug. Guarding `onResume` with a "have we loaded yet" flag, or dropping the `onViewCreated` call and letting `onResume` own it, would settle it.

**Worth a decision rather than a fix: `onPause` also fires when the app is backgrounded.** A user who sets a date range, takes a phone call, and comes back finds the filter cleared and the list reloaded. #15767 is about leaving the finances *page*, and `onPause` conflates that with leaving the app. It may well be acceptable — the filter is transient by design — but it's a behaviour the issue doesn't ask for. If page-precision matters, `TeamDetailFragment:221` already registers a `ViewPager2.OnPageChangeCallback`, which fires only on page changes. I'd merge as-is and revisit if anyone complains.

**Merge-order note, and this one isn't mechanical.** #15594 in this same batch restructures this file into a `ConcatAdapter` header — it deletes `updateToDateState`, removes the `isAsc` field in favour of `HeaderState.isAsc`, and moves `imgDate` into the header layout. `resetFilterAndSort` uses all three. I confirmed the two branches conflict. Landing #15594 first and rewriting this reset against `headerState` is the cheaper order; a textual merge won't produce working code either way.

CI green on all four checks; merges cleanly against master today.

**Verdict:** `ready`. Drop the `onDestroyView` call and pick one owner for the initial load — both are tidy-ups, not blockers.

---

dogi:
Promoted `ready` → `merge`: my 82/100 review covers the current head `fd933b7`, and CI is green on all four checks.

**One thing worth a second look before the queue drains this.** @Okuro3499's approval is on `f1a1739`, and `fd933b7` landed after it — that's the commit that replaced `setMenuVisibility`/`onHiddenChanged` with `onPause()`, which is the entire mechanism of the fix. So the approval predates the rework rather than covering it. Worth a re-confirm.

Also still true from my review: this conflicts with #15594 (`ready`), which deletes `updateToDateState`, the `isAsc` field and `binding.imgDate` in favour of `HeaderState`. `resetFilterAndSort` uses all three. If #15594 goes first, this needs its reset rewritten rather than merged.

---

dogi (approved):

PR #15769, opened 2026-08-17
teams: smoother repository dashboard bell view modelling (fixes #15726)

### LINKED ISSUE #15726: my teams empty navigation
For a new user who has not joined any team if they click myteams or anywhere on the section show them the list of existing teams and ensure that if they are owners or members of any teams the current behavior persists where they are show list of teams they are members of
https://github.com/user-attachments/assets/9cf5fa8f-4ab9-456d-9290-48fd155d57ac

ragilzakaria:
https://github.com/user-attachments/assets/66fdfd0b-9cc2-4d67-950f-b7976e2b7aa1

fixes #15726 

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved dashboard team and enterprise loading when no joined teams or user ID is available.
  - Excludes archived, deleted, pending-deletion, nested, and unsupported team records.
  - Improved navigation from dashboard team and library cards, including empty states.
  - Improved survey prompts, reminders, and course-certification checks.
  - Supports consistent filtering of team and enterprise details.

- **Tests**
  - Added coverage for team filtering, fallback loading, enterprise details, and missing user IDs.


Okuro3499 (changes requested):
if user has teams they are member of they should be shown list of teams they are members of
https://github.com/user-attachments/assets/f1270857-884c-450f-a94b-a02ee183d23b

dogi:
**Review — rating: 66/100**

The behaviour is right — sending a user with no joined teams to the full team list instead of an empty "my teams" screen is the same gap #15770 closed for resources. Two things need work: the mechanism it uses to express that, and two repository changes riding along unannounced.

**`fromDashboard` is being overloaded to mean something it doesn't.**

```kotlin
val hasJoinedTeams = userTeams.isNotEmpty()
val fragment = TeamFragment().apply {
    arguments = Bundle().apply {
        putBoolean("fromDashboard", hasJoinedTeams)
    }
}
```

The user *did* arrive from the dashboard. Setting the flag to `false` to mean "show me all teams" makes it lie about the thing its name describes, and `TeamViewModel.loadTeams` then branches on it as though it were a provenance flag. Anything that later reads `fromDashboard` for its actual meaning — back-stack handling, screen chrome, analytics — gets the wrong answer, and nothing will fail loudly.

It's already inconsistent inside this diff: the empty-state placeholder in `renderMyTeams` hardcodes `putBoolean("fromDashboard", false)` for a case where the user also came from the dashboard. So the flag now carries two different meanings in one file.

#15770 solved the identical problem without lying about anything — it picked between two navigation calls (`openMyFragment` for the filtered view, `openCallFragment` for the full list) based on emptiness. That reads correctly at the call site and needs no ViewModel change. If a flag is genuinely preferable here, give it its own name (`showAllTeams`) and leave `fromDashboard` alone.

**Two repository behaviour changes aren't mentioned anywhere.**

`getMyTeamsFlow` gains a type filter:

```kotlin
it.isRootTeam() &&
    (it.type == "team" || it.type.isNullOrBlank())
```

That changes what *every* caller of `getMyTeamsFlow` sees, not just the dashboard — anything with a `type` other than `"team"` or null now disappears from "my teams". If enterprises were leaking into that list, this is a fix worth having; if some other type exists in the wild, those rows vanish silently. Either way it deserves its own line in the description, because "empty teams navigation" doesn't prepare a reader for it.

`getMyTeamDetailsFlow` gains two more conditions on the membership filter:

```kotlin
it.userId == userId &&
    it.docType == "membership" &&
    !it.isDeletePending &&
    !it.teamId.isNullOrBlank()
```

Excluding delete-pending memberships is very likely correct — a membership queued for removal shouldn't count as joined — but it's a separate bug fix bundled into a navigation PR, and it changes what the teams count shows.

**Neither of those is covered by a test.** The two tests that were added cover the ViewModel paths — joined teams via `getMyTeamDetailsFlow(userId, "team")`, and the null-user path through `getTeamDetails(null)` — which is the right pair for the feature. But the riskiest part of the diff is the filtering change inside `TeamsRepositoryImpl`, and there's nothing exercising it. `TeamsRepositoryImplTest` already exists; a case with a mixed-type team set would pin down what the new predicate includes and excludes.

**`userTeams` carries the same seam as #15770's `userLibrary`.** It's a `protected var` on `BaseDashboardFragment`, assigned only inside `renderMyTeams`, and read at click time — so "not loaded yet" is indistinguishable from "loaded and empty". CodeRabbit raised exactly this on #15770 and it applies verbatim here. The window is short, so it isn't a blocker, but if you're touching both fragments anyway, reading the current list from `DashboardViewModel` at click time removes the ambiguity for both cards at once.

**Merge-order note:** #15770 is labelled `merge` and edits both of these files — it adds `protected var userLibrary` to `BaseDashboardFragment` and restructures the same `declareElements` block in `BellDashboardFragment` with an `openLibraryAction` lambda. These two will collide, and more importantly they'd land two different idioms for the same problem in adjacent lines. Worth settling on one before either merges; #15770's approach is the one that doesn't require a flag.

**What's good:** wiring `myTeamsImageButton` alongside `llHomeTeam` so the whole card is tappable matches what #15770 did for the library card, and the `TeamViewModel` restructure — hoisting `targetType` once and adding the `userId == null` branch — is clean and readable.

**Verdict:** `change`. Replace the `fromDashboard` overload with either the two-navigation-calls approach from #15770 or a distinctly named argument, and split out (or at least document and test) the two `TeamsRepositoryImpl` filter changes.

---

dogi:
**Re-review — Rating: 84/100** (was 66)

Both findings landed, and the first one landed the way I hoped rather than the way I'd have settled for.

**`fromDashboard` no longer lies.** The flag is back to meaning "I came from the dashboard", and the my-teams-vs-all-teams choice is expressed by picking the navigation call instead:

```kotlin
val openTeamsAction = {
    if (userTeams.isNotEmpty()) {
        val fragment = TeamFragment().apply {
            arguments = Bundle().apply { putBoolean("fromDashboard", true) }
        }
        homeItemClickListener?.openMyFragment(fragment)
    } else {
        homeItemClickListener?.openCallFragment(TeamFragment())
    }
}
```

That's the same idiom the already-merged #15770 uses for the library card sitting immediately below it in `declareElements`, so the two now read as one pattern rather than two competing ones. The empty-state placeholder in `renderMyTeams` was fixed too — it went from `putBoolean("fromDashboard", false)` to a plain `openCallFragment(TeamFragment())`, which also cleared the two fully-qualified names (`org.ole.planet.myplanet.ui.teams.TeamFragment`, `android.os.Bundle`) that were there.

**The repository filters are covered now.** That was the riskiest part of the diff — changes to `getMyTeamsFlow` and `getMyTeamDetailsFlow` that affect every caller, not just the dashboard. Three tests were added:

- `getMyTeamsFlow filters out non-root, archived, delete-pending, and non-team types`
- `getMyTeamDetailsFlow filters correctly for team and enterprise types`
- `getMyTeamDetailsFlow returns empty list when user has no joined teams`

The first one is the important one, since it pins down exactly what the new `(it.type == "team" || it.type.isNullOrBlank())` predicate includes and excludes. It's still worth a line in the PR description that this changes what "my teams" returns app-wide — the tests document the behaviour, but a reader of the title won't go looking for them.

**One thing to drop: the version bump.** `app/build.gradle` goes `6546` → `6547`, and `.github/scripts/automerge.sh` does that itself when draining the queue — every merge commit on master carries a +1 from the script. #15519, #15820 and #15824 all currently claim `6547` too, so only one of the four can merge without a `build.gradle` conflict.

**Still open from last round, not blocking:** `userTeams` is a `protected var` on `BaseDashboardFragment` assigned only inside `renderMyTeams`, so "not loaded yet" and "loaded and empty" are the same value at click time. It now sits next to `userLibrary` with the identical shape, which at least makes it a consistent seam rather than a one-off — and it's the natural thing to fix for both cards together by reading from `DashboardViewModel` at click time.

Wiring `myTeamsImageButton` alongside `llHomeTeam` matches what #15770 did for the library card, so the whole card is tappable in both places now.

**Verdict:** `ready`. Drop the `build.gradle` hunk before it goes in the queue.

---

Okuro3499 (dismissed):
@ragilzakaria 
top bar says teams instead of myTeams
Image

dogi:
**Review — rating: 84/100**

Routes the dashboard teams card to the full teams list when the user has no joined teams, and tightens the two "my teams" flows. Build (default + lite) and test (both shards) green.

The change I went in most suspicious of turned out to be safe. Replacing `(it._id ?: it.id) in teamIds` with `it._id in teamIds` looked like it dropped a fallback that could hide teams — but in `MyTeam`, `_id` is `@PrimaryKey var _id: String = ""` (non-null) and `id` is only an alias property (`get() = _id`), so the elvis branch was unreachable and `it.id` resolved to `_id` anyway. Pure simplification, zero behavior change.

Other things I checked:

- **`!it.teamId.isNullOrBlank()` closes a real false positive.** Since `_id` defaults to `""`, a membership row with a blank `teamId` could previously land `""` in `teamIds` and match a team whose `_id` was empty. `mapNotNull` alone didn't cover that.
- **`targetType == "enterprise"` is equivalent to the old `type == "enterprise"`** in the non-dashboard branch (`null` → `"team"`, which never equals `"enterprise"`), so that rename is behavior-preserving.
- **A genuine dead end is fixed:** `fromDashboard && userId == null` previously had no `else`, so the list stayed permanently empty. It now loads all teams (or enterprises).
- `txt_myTeams` exists in `values/` and all five locales; `myTeamsImageButton` resolves through view binding in both flavors.
- Test coverage is good — the repository filter matrix (archived / delete-pending / non-root / enterprise / null-type) plus four ViewModel tests over the new branches.

Two things worth a deliberate decision rather than a silent one:

1. **Scope beyond the title.** The `type` and `isDeletePending` filters on `getMyTeamsFlow` affect its only caller, `DashboardViewModel`, so the dashboard teams card now excludes enterprises and delete-pending teams — and that also shifts the card's count via `setCountText(teams.size, …)`. I think it's the correct semantics for "My Teams", but it's a separate change from what the title describes.
2. **`openTeamsAction` depends on render timing, not on data.** It branches on `userTeams`, which is only populated once `renderMyTeams` runs. A tap before the teams flow first emits sees an empty `userTeams` and opens the full list instead of "My Teams". It self-corrects after the first emission and degrades into exactly the fallback this PR wants, so it's low impact — but reading the count from the ViewModel's `uiState` at click time would make it deterministic.

**Nit:** `(it.type == "team" || it.type.isNullOrBlank())` is now duplicated across both flows, and the inline `if/else` inside `getMyTeamDetailsFlow`'s filter is dense. A small `MyTeam.matchesType(targetType)` helper would keep the two from drifting — `TeamsRepositoryImpl` is already the largest file in the repo.

**Verdict:** mergeable — please just confirm the dashboard-card filtering change in point 1 is intended.

---

Okuro3499 (approved):

dogi (approved):

PR #15770, opened 2026-08-17
dashboard: smoother base resources navigating (fixes #15728)

### LINKED ISSUE #15728: empty resources navigation
For a new user who has not joined any resource if they click myresources or anywhere on the section show them the list of existing resources and ensure that if they joined any resource the current behavior persists where they are show list of resources they joined
https://github.com/user-attachments/assets/eb06f8e9-5b21-41c9-a9a8-bc51f99a04f5

ragilzakaria:
https://github.com/user-attachments/assets/26b2f4b8-8b33-44bb-b9bb-3dd7f8dfcc02

fixes #15728 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved library card navigation for signed-in and guest users.
  * Empty libraries now open the Resources section, while populated libraries open the user’s library view.
  * Library navigation works consistently across the card and image controls.


dogi:
**Review — rating: 80/100**

The navigation distinction is real and the PR uses it correctly. `DashboardActivity.openMyFragment` stuffs `isMyCourseLib = true` into the fragment's arguments before opening it; `openCallFragment` doesn't. So the two entry points genuinely differ — "my resources" versus "all resources" — and sending a user with an empty library to the filtered view was always going to show them an empty screen with no way forward. Routing them to the full catalogue instead is the right answer to #15728.

Both call sites are correct, which is worth saying because the two look inconsistent at a glance. In `BaseDashboardFragment.renderMyLibrary` the switch to `openCallFragment` is unconditional — but it's inside the `if (dbMylibrary.isEmpty())` branch, on the `renderPlaceholder("no resources added yet")` click, so emptiness is already established. In `BellDashboardFragment` the card is reachable in both states, so the `userLibrary.isNotEmpty()` test is needed there. Consistent once you follow which branch each one is in.

Extracting `openLibraryAction` and wiring it to both `llHomeLibrary` and `myLibraryImageButton` is the right shape — the child button consumes its own taps and the parent handles the surrounding area, both landing on one implementation. The `ll_home_library` id is added identically to `layout/` and `layout-sw600dp/`, so the tablet layout doesn't silently miss the new target.

**`userLibrary` is a mutable field with an ordering dependency.** It's only assigned inside `renderMyLibrary`, which runs from the library flow collector, while `openLibraryAction` reads it at click time. Until the first emission arrives, the field is `emptyList()` — indistinguishable from "user genuinely has nothing". A tap in that window sends someone with a full library to the all-resources view. The window is short and the failure is mild, so I'm not holding the PR for it, but the assignment being the first statement in `renderMyLibrary` (before the early return) is what makes it work at all, and that's a subtle thing for the next editor to preserve. Reading the current list from `DashboardViewModel` at click time — or exposing it as a `StateFlow` the action queries — removes the ordering question entirely.

Related: `protected var userLibrary` on an `open` base class is a loose seam. Any subclass can write to it, and nothing signals that it's derived state owned by `renderMyLibrary`. A `protected val` over the ViewModel state would express the ownership properly.

**Please revert the `ResourcesViewModelTest` change.** It's unrelated to this PR, and what it deletes is the useful part:

```kotlin
-        // MyLibrary is a Room entity whose id is a @JvmField and whose isResourceOffline() is
-        // @Ignore'd — neither can be stubbed with mockk; use a real instance.
```

```kotlin
-        // TagEntity is a Room entity whose id is a @JvmField (a Java field, not a getter), so it
-        // cannot be stubbed with mockk `every { id }`; use a real instance instead.
```

Those two comments explain why the surrounding code looks unidiomatic — real instances where mocks would be expected. Without them the next person tries `every { mockLibrary.id }`, gets an unhelpful MockK failure, and rediscovers the `@JvmField` limitation from scratch. That's exactly the knowledge a comment should carry.

**Minor:** the id `ll_home_library` names a `<FrameLayout>` — the `ll_` prefix reads as LinearLayout. And the FrameLayout becomes clickable without an accessibility label of its own; TalkBack should fall through to the inner button, but it's worth a pass with TalkBack on to confirm the card announces something sensible now that the whole area is a touch target.

CI green on all four checks, merges cleanly, and the branch already has master merged in.

**Verdict:** `ready`. Restore the two test comments before merging; the `userLibrary` seam is worth revisiting but doesn't need to block this.

---

Okuro3499 (approved):

dogi:
Promoted `ready` → `merge`: approved by @Okuro3499 at `6ab522f`, still the head, and my 80/100 review covered that commit.

Two things I'd still like, neither worth holding the merge for: restore the two `ResourcesViewModelTest` comments this branch deletes (they document why `@JvmField`/`@Ignore` members can't be mockk-stubbed), and revisit the `userLibrary` seam — CodeRabbit landed on the same point independently, that "not loaded yet" and "loaded and empty" are indistinguishable. Fine as follow-ups.

---

dogi (approved):

PR #15771, opened 2026-08-17
resources: smoother web view nested entry pathing (fixes #15634)

### LINKED ISSUE #15634: Unable to download HTML resources (download keeps failing)
## Summary
Users are unable to download HTML resources in MyPlanet. The download action keeps failing.

## Steps to Reproduce
1. Open MyPlanet.
2. Navigate to an HTML resource.
3. Attempt to download the resource.
https://github.com/user-attachments/assets/6a7bd38b-34e6-44a0-b8d3-625efd8f5f10

Okuro3499:
fixes #15634
Add openWhichFile field to MyLibrary to support HTML resources with non-root entry points (e.g. sudoku/index.html). Preserve nested directory structure when downloading attachments and resolve the correct entry file in WebViewActivity. Includes path traversal protection and unit tests.
[html.webm](https://github.com/user-attachments/assets/64840d23-cf58-4fe8-bb24-304e69cc8796)

## Summary by CodeRabbit

* **New Features**
  * HTML-based resources can open their configured entry file, including nested paths.
  * Resource links preserve nested attachment paths when opened locally.
  * Missing entry-file settings continue to default to `index.html`.

* **Bug Fixes**
  * Improved handling of HTML resource paths while preventing unsafe absolute or traversal-based paths.

* **Tests**
  * Added coverage for nested paths, default entry files, omitted settings, and path safety.


dogi:
**Review — rating: 74/100**

This is the best-constructed PR in the current queue and it fixes a whole class of broken resources, not just the entry point. One thing will crash it on upgrade, and it's a single line.

**The schema change has no version bump.** `MyLibrary` is a Room `@Entity`, and this adds a column:

```kotlin
var openWhichFile: String? = null
```

`AppDatabase.kt` isn't touched — it's still `version = 7`. Room stamps a hash of the compiled schema into `room_master_table` and compares it on every open. When the hash differs but the version doesn't, Room throws `IllegalStateException: Room cannot verify the data integrity. Looks like you've changed schema but forgot to update the version number` — and `fallbackToDestructiveMigration` does **not** rescue it, because that only applies when a version change has no migration path. So every existing install crashes on first launch after this ships.

CI won't catch it: unit tests build the database fresh in memory, so the stored hash always matches, and `exportSchema = false` means there's no schema JSON for the build to diff against. It only shows up on a device that already has data. Bump `version` to 8 in `AppDatabase` and it's handled — CLAUDE.md's Room section spells this out, including that the rebuild re-pulls from CouchDB and discards unsynced local writes.

**The download-path fix is the bigger half of this PR, and the description undersells it.** `getSDPathFromUrl` previously ran the URL through `getFileNameFromUrl`, keeping only the last segment. So an HTML bundle at `resources/123/index.html` + `resources/123/js/game_manager.js` + `resources/123/css/game.css` all landed flat in `ole/123/`, and any page referencing `js/game_manager.js` relatively got a 404. Worse, two files named `index.html` in different subfolders collided on one path. `getResourceRelativePathFromUrl` preserving everything after the id fixes both, and I checked that `createFilePath` already handles it — it splits on `/` and `mkdirs()` the subdirectory before writing, so no plumbing was missing.

Worth calling out in the PR body: this changes where **all** resource attachments land, not only HTML ones. Devices with previously downloaded nested content will look for it at the new nested path, miss, and re-download once. Benign, but it's a bandwidth event on exactly the low-connectivity devices this app targets, so maintainers should know it's coming rather than discover it in the field.

**`resolveHtmlEntryFile` is properly defensive.** Three independent layers — rejecting leading `/` or `\`, rejecting any `..`, then a canonical-path containment check that catches whatever the string checks miss (symlinks, encoding tricks). Returning `File?` and having every caller test `?.exists() == true` means a rejected path degrades to "resource not available" rather than reading something it shouldn't. All three `File(directory, "index.html")` sites are updated — `BaseContainerFragment` at 123 and 192, `WebViewActivity` at 80 — so nothing is left on the old assumption.

The per-segment encoding in `WebViewActivity` is right too:

```kotlin
val encodedEntryPath = entryPath.split("/").joinToString("/") { Uri.encode(it) }
```

Encoding the whole path at once would have escaped the separators; splitting first keeps them intact while still handling spaces and non-ASCII in folder names. And `invariantSeparatorsPath` means it won't break if this ever runs somewhere with a different separator.

**Tests are the strongest part.** Six cases on `resolveHtmlEntryFile` covering null, blank, nested, traversal (both `../` and the `sudoku/../../` form that only the canonical check catches), and absolute paths; two on `getSDPathFromUrl` for nested and single-segment; two on the repository confirming `openWhichFile` is parsed and that it stays null when the CouchDB doc omits it. `.takeIf { it.isNotBlank() }` normalising `""` to null is the right call and it's covered.

**One nit:** `candidate.contains("..")` rejects any path with two consecutive dots anywhere, so a legitimately named `chapter..intro.html` resolves to null and the resource silently reads as not downloaded. The canonical containment check below already catches real traversal, so the substring test could be narrowed to path segments equal to `..` — or dropped. Fail-safe as it stands, so not blocking.

**Verdict:** `change` — bump `AppDatabase.version` to 8 and this is ready. Everything else here I'd merge as-is.

---

dogi:
**Re-review — Rating: 74/100** (unchanged)

CI green on all four checks, merges cleanly, and the substance is unchanged from what I rated 74 — which remains the strongest-tested PR I've looked at in this queue. The one blocker is still open, and it's still a one-line fix.

**`MyLibrary` gains a column and `AppDatabase.version` doesn't move.** The numbers have shifted since last round but the gap hasn't:

- master's `AppDatabase`: `version = 8`
- this branch's `AppDatabase`: `version = 8` — the file isn't touched at all
- master's `MyLibrary`: no `openWhichFile`
- this branch's `MyLibrary`: `+ var openWhichFile: String? = null`

So merging produces a schema that differs from what `version = 8` describes. Room hashes the compiled schema into `room_master_table` and compares on open; a mismatch with an unchanged version throws `IllegalStateException: Room cannot verify the data integrity. Looks like you've changed schema but forgot to update the version number`. `fallbackToDestructiveMigration` doesn't cover it — that only applies when a version *change* has no migration path.

Master is already at 8 and releases ship from master, so any install that has run a v8 build will hit this on first launch after the update. Bump `version` to 9 in `AppDatabase.kt` and it's handled — CLAUDE.md's Room section spells out the drop-and-resync consequence, including that unsynced local writes are discarded.

Nothing in CI can see this: unit tests build the database fresh in memory so the stored hash always matches, and `exportSchema = false` means there's no schema JSON to diff at build time.

Everything else I verified before still stands, and I re-checked the parts that master could have invalidated:

- `resolveHtmlEntryFile` still rejects leading `/` and `\`, rejects `..`, and backstops with a canonical-path containment check — three independent layers, returning `File?` so a rejected path degrades to "not available" rather than reading outside the resource directory.
- `getResourceRelativePathFromUrl` still preserves everything after the resource id, which is what makes nested HTML bundles work at all — previously `resources/123/js/game.js` and `resources/123/index.html` both flattened into `ole/123/`, so relative references 404'd and same-named files in different subfolders collided. `createFilePath` already `mkdirs()` the subdirectory, so no plumbing was missing.
- All three `File(directory, "index.html")` call sites are still updated — `BaseContainerFragment` ×2 and `WebViewActivity` — so nothing is left on the old assumption.
- The per-segment `Uri.encode` in `WebViewActivity` still splits on `/` first, so separators survive while spaces and non-ASCII folder names get escaped.
- Ten tests: six on `resolveHtmlEntryFile` (null, blank, nested, `../`, the `sudoku/../../` form only the canonical check catches, absolute), two on `getSDPathFromUrl`, two on `openWhichFile` parsing including the doc-omits-it case.

**Still worth putting in the PR description:** this changes where *all* resource attachments land, not only HTML ones, so devices with previously downloaded nested content will miss once and re-download. Benign, but it's a bandwidth event on exactly the devices this app targets.

**Nit, unchanged:** `candidate.contains("..")` rejects any path with two consecutive dots, so a file legitimately named `chapter..intro.html` resolves to null. The canonical check already catches real traversal; narrowing the test to path segments equal to `..` would avoid the false positive. Fail-safe as-is.

**Verdict:** `change`, for the same single reason — bump `AppDatabase.version` to 9. Everything else here I'd merge unchanged.

---

dogi:

The blocker is closed. `7945a9c` is exactly the one-line change — `version = 8` → `version = 9` in `AppDatabase.kt`, nothing else in the commit — so the `openWhichFile` column now has a version change to hang the destructive rebuild on, and the Room identity-hash check won't throw on first launch after the update. Master is still at `8`, the branch merges clean into `41d89f4`, and all four checks are green.

I re-verified the parts of the mechanism that a version bump can't tell you about, because the interesting question here was never the column — it was whether a nested entry point survives the whole round trip. It does, and each hop lines up with the next:

- **Download nests.** `getHtmlResourceDownloadUrls` enumerates every `_attachments` key, slashes included, and builds `.../resources/<id>/<name>`. `getSDPathFromUrl` now routes through `getResourceRelativePathFromUrl`, which keeps everything after the resource id, and `createFilePath` already `mkdirs()` the subdirectory — so `sudoku/js/game.js` lands at `ole/<id>/sudoku/js/game.js` instead of colliding in the flat root with every other `game.js` in the bundle.
- **The entry file resolves.** `resolveHtmlEntryFile` defaults to `index.html`, and canonicalises both sides before the containment check, so a symlink can't walk out either.
- **The asset loader serves it.** `WebViewAssetLoader` hands the path handler `Uri.getPath()` with the `/assets/` prefix stripped, which is already percent-decoded — and `WebViewActivity` encodes per-segment after splitting on `/`, so separators survive while spaces and non-ASCII folder names get escaped. Decode and encode are symmetric; the nested URL resolves to the right file.
- **The safety gate lets it through.** `WebViewSafety.isUrlSafe` admits `https` unconditionally and the `appassets.androidplatform.net` URL is `https`, so there's no path-level check to trip over now that the path has more than one segment.
- **No call site left behind.** `git grep index.html` over `app/src/main` at this head returns only the three lines inside `resolveHtmlEntryFile` itself — the KDoc and the default. Both `BaseContainerFragment` sites and the `WebViewActivity` site are converted; nothing else is still hardcoding the root.

Eight new tests, and they cover the cases that actually decide correctness: nested, unset, blank, `../`, the `sudoku/../../` form that only the canonical check catches, absolute, plus the two parse tests including doc-omits-it.

**One real finding, minor.** `getResourceRelativePathFromUrl` decodes each segment with `URLDecoder.decode`, which is the `application/x-www-form-urlencoded` decoder — it turns `+` into a space. Path components don't work that way. An attachment named `my+app/index.html` gets written to `my app/index.html`, while `openWhichFile` comes off the doc raw and still says `my+app/index.html`, so `resolveHtmlEntryFile` misses and the resource opens as a blank WebView. `getFileNameFromUrl` has had the same decoder for a long time, so this isn't a regression on its own — what's new is the raw-vs-decoded comparison, which is where a wrong decode stops being invisible. `Uri.decode` doesn't do the `+` substitution and is already the idiom elsewhere in the file:

```kotlin
segments.subList(idx + 2, segments.size).joinToString("/") { Uri.decode(it) }
```

**Nit, still open from last round.** `candidate.contains("..")` rejects any path with two consecutive dots, so `chapter..intro.html` resolves to null. The canonical containment check already catches real traversal; testing for path segments equal to `..` would drop the false positive. Fail-safe as-is.

**Nit.** When `resolveHtmlEntryFile` returns null or the file is missing, `WebViewActivity` loads nothing at all — empty WebView, no message. That shape predates this PR, but there are now more ways to reach it (a bad `openWhichFile` from the server, a rejected path, the decode mismatch above), so a toast and `finish()` would turn a mystery blank screen into something a user can report.

**Follow-up, pre-existing, worth its own issue.** `MyLibrary.insert` only sets `resourceLocalAddress`/`resourceOffline`/`downloadedRev` for attachment keys with no slash (`if (key.indexOf("/") < 0)`). A bundle whose attachments are *all* nested therefore never gets `resourceOffline = true`, so `isResourceOffline()` stays false forever and `ResourceDetailFragment.updateDownloadButtonState` keeps showing the download affordance for content that's already on disk. Harmless in practice — `openResource` checks the directory and opens it anyway — but it's the same feature area and it'll read as a bug to whoever hits it next.

**Worth a line in the PR description:** this changes where *all* resource attachments land, not just HTML ones. Devices holding previously downloaded nested content will miss once and re-download. Benign, but it's a bandwidth event on exactly the hardware this app targets.

**Verdict:** `ready`. The `Uri.decode` swap is a one-liner I'd take before merge if it's convenient, but it's a narrow edge case and not worth another round on its own.

---

dogi (approved):

PR #15772, opened 2026-08-17
courses: smoother progress scrolling (fixes #15553)

### LINKED ISSUE #15553: Course progress page is not responsive
## Summary
The course progress page is not responsive and does not adapt properly to smaller screen sizes.

## Expected behavior
The course progress page should display correctly across supported mobile and tablet viewport sizes without layout overflow, clipped content, or unusable controls.

## Actual behavior
On smaller screens, the page layout appears broken and is not responsive.

Attached reference:
https://github.com/user-attachments/assets/44e386d7-76fb-498e-9643-124d3f3d7e2e

J-S-webskas:
https://github.com/user-attachments/assets/e5ac3dd9-b0e3-48e2-b3ea-1d60c33f10de

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved scrolling behavior on the course progress screen.
  * Ensured progress content fills the available viewport while allowing the full list to scroll smoothly.
  * Prevented nested scrolling conflicts within the progress list.


dogi:
**Review — rating: 76/100**

The diagnosis is right. `activity_course_progress.xml` had a fixed `match_parent` `LinearLayout` with a `wrap_content` `RecyclerView` last — so the RecyclerView only ever got the leftover space below the 164dp progress ring and three text views, and everything above it was pinned. On a short screen the step grid was squeezed into whatever remained. Wrapping the whole column in a `NestedScrollView` makes the page scroll as one unit, which is what #15553 is asking for.

I diffed the moved block line by line: nothing changed but indentation. No attributes gained, lost, or reordered, and `layout/activity_course_progress.xml` has no `sw600dp` or other configuration variant to keep in sync. Easy to verify, which is worth something on a layout change.

Two things to fix before it lands.

**The scrolling child should be `wrap_content`, not `match_parent`.**

```xml
<androidx.core.widget.NestedScrollView
    android:layout_width="match_parent"
    android:layout_height="match_parent">

    <LinearLayout
        android:layout_width="match_parent"
        android:layout_height="match_parent"     <!-- ← wrap_content -->
```

It happens to work today because `NestedScrollView.measureChild` builds the child's height spec as `UNSPECIFIED` and ignores the declared value entirely. But the declaration says the opposite of what's intended, and it stops being harmless the moment someone adds `android:fillViewport="true"` — at which point the two interact in ways nobody expects. One-word fix.

**The RecyclerView stops recycling.** `rv_progress` is `wrap_content` inside a scroll container now, so it's measured `UNSPECIFIED` and lays out every item at once — with `GridLayoutManager(this, 4)` from `CourseProgressActivity:28`, that means every step view in the course inflated and held. For a short course it's invisible; for a long one it's a memory spike and a stutter on open, and it's the exact thing the RecyclerView was there to avoid. Two ways out:

- Minimal: add `android:nestedScrollingEnabled="false"` to the RecyclerView. Doesn't restore recycling, but it stops the RV and the `NestedScrollView` both trying to handle the same scroll gesture, and it makes the trade-off explicit to the next reader rather than accidental.
- Proper: drop the `NestedScrollView` and let the RecyclerView be the only scroller, moving the progress ring and the three text views in as a header — either a header view type or a `ConcatAdapter`. That gets the whole-page scrolling this PR wants *and* keeps recycling.

The second is more work but it's the shape this screen wants; the first is fine if you'd rather ship the fix now.

**Minor:** without `android:fillViewport="true"`, a course with only a step or two won't stretch the `LinearLayout` to the viewport. The `@color/secondary_bg` is on the root `CoordinatorLayout` so nothing shows through, and the content is already `layout_gravity="center"` horizontally, so the visible effect is small — but if you want the short case to look the same as before, that's the attribute.

Also, the commit message `added nested Scroll View` doesn't mention the issue; the PR title carries `(fixes #15553)`, so it'll link correctly, but a squash message worth reading would help.

**Verdict:** `change`. Both fixes are one line each — `wrap_content` on the `LinearLayout`, and a decision on the RecyclerView. Worth a quick check on a long course before and after, since that's where the recycling change shows.

---

dogi:

All three items landed, and the third one took the option I'd have picked given the size of the change.

- The scrolling child is `wrap_content` now, which is what a `NestedScrollView` child should declare. It happened to work before because `NestedScrollView.measureChild` builds an `UNSPECIFIED` height spec and ignores the declared value — but the declaration said the opposite of the intent, and it stops being harmless the moment `fillViewport` enters the picture. Which it now has.
- `android:fillViewport="true"` is set, so a course with one or two steps stretches the `LinearLayout` to the viewport instead of collapsing to content height.
- `android:nestedScrollingEnabled="false"` on `rv_progress` stops the RecyclerView and the `NestedScrollView` both trying to claim the same scroll gesture, and — as I said last round — makes the recycling trade-off explicit rather than accidental. With `GridLayoutManager(this, 4)` every step view still inflates at once; that's the accepted cost of the minimal fix, and I'm not re-litigating it. If a course with a very large step count ever gets slow to open, moving the header into the adapter as a `ConcatAdapter` row is the upgrade path — #15594 does exactly that for the finance screen and is a good reference.

`android:padding="@dimen/padding_normal"` survived the move onto the inner `LinearLayout`, so the spacing is unchanged, and the rest of the diff is still pure re-indentation — I diffed the moved block again and no attribute was added, dropped or reordered.

**One whitespace nit:** `android:fillViewport="true" >` has a stray space before the closing bracket.

**Verdict:** `ready`. Worth opening the progress screen on a long course once before it ships, since that's the case the recycling change affects.

---

Okuro3499 (approved):

dogi (approved):

PR #15773, opened 2026-08-17
enterprises: smoother finances date filtering (fixes #15766)

### LINKED ISSUE #15766: finances date filter
finances date filter should not go past current date (future dates)
https://github.com/user-attachments/assets/1efbe736-b812-41c0-94cf-f85bc86e2ac6

J-S-webskas:
https://github.com/user-attachments/assets/d5ef7e05-cc1d-4b5b-80ad-97b1ff9bfe47

## Summary by CodeRabbit

* **Bug Fixes**
  * Date selection dialogs now prevent users from choosing dates later than today.


Okuro3499 (approved):

dogi:
**Review — rating: 80/100**

Four lines, and the effective change is one: `datePickerDialog.datePicker.maxDate = maxDay.timeInMillis`. It closes a real hole — before this, both pickers accepted future dates, so a user could filter transactions for a range that hasn't happened and get an empty table with no explanation.

**The second half of the title is already in master.** The PR is titled *"…and toDate has to be greater or equal to fromDate"*, but that constraint was there before this diff, enforced twice over:

```kotlin
if (!isFromDate) {
    val fromDate = parseDate(fromDateText)
    if (fromDate != null) {
        datePickerDialog.datePicker.minDate = fromDate.timeInMillis
    }
}
```

and, in the fromDate callback, clearing a now-invalid toDate:

```kotlin
if (toDateMillis != null && toDateMillis < fromDateMillis) {
    binding.etToDate.setText("")
}
```

Worth retitling to just the max-date fix so the diff and the description agree — otherwise a reviewer goes looking for the ordering logic and can't tell what's new.

**`maxDay` duplicates `now`.** The function already opens with a `Calendar` for today, normalised to midnight:

```kotlin
val now = Calendar.getInstance().apply {
    set(Calendar.HOUR_OF_DAY, 0)
    ...
}

val maxDay = Calendar.getInstance()
```

`now` is today, it's already normalised, and it's four lines up — so `maxDate = now.timeInMillis` does the same job without a second `Calendar.getInstance()`. That also fixes the odd distance between `maxDay`'s declaration and its only use, forty-odd lines later, past the whole `DatePickerDialog` construction. And it makes the normalisation consistent: you deliberately zeroed the time on `now` and didn't on `maxDay`. It happens not to matter — `DatePicker` compares on the date, so today stays selectable either way — but the inconsistency reads like an oversight rather than a decision.

**Ordering edge case, low severity.** `maxDate` is assigned before the `if (!isFromDate)` block that sets `minDate` from the existing fromDate. If a stored fromDate is ever in the future, you'd end up with `minDate > maxDate`, which `DatePicker` handles badly. In practice this is only reachable for a user who picked a future fromDate under the old behaviour and then opens the toDate picker in the same session — which is exactly what this PR stops happening going forward, and the fields aren't persisted across fragment recreation. Setting `maxDate` after the `minDate` block, or clamping, would close it; I wouldn't block on it.

**A design question rather than a defect:** there's no symmetric cap on the fromDate picker. Pick toDate = 10 Jan, then reopen fromDate and you can still select 20 Feb — the callback then silently blanks toDate. Given this PR is about making the range coherent, capping the fromDate picker at the current toDate would prevent the invalid pick instead of discarding the user's other input. Existing behaviour, so out of scope unless you want it.

No tests, which is right for a Fragment in this project — there's no `androidTest` source set and `DatePickerDialog` isn't reachable from the unit suite. Worth a manual pass: confirm tomorrow is greyed out in both pickers, and that picking fromDate after an existing toDate still behaves.

**Merge-order note:** #15768 also edits `EnterprisesFinancesFragment` (adds `resetDateFilter()` plus lifecycle hooks) and is currently `change`. Different functions, and both merge cleanly against master, so they shouldn't collide — but whichever goes second should re-check.

**Verdict:** `ready`. Reuse `now` instead of the extra `Calendar`, and retitle to match what the diff actually does — neither needs another round.

---

dogi (approved):

PR #15820, opened 2026-08-19
teams: smoother task and meetup comment threads managing (fixes #15112)

### LINKED ISSUE #15112: improvement: comment threads on team tasks and meetups
Team tasks and meetups are currently static cards with no way for members to communicate.
It is better to:
- Add an inline comment thread below each Team Task and Meetup card.
- Team members can post updates, ask clarifying questions, and reply to one another offline

ragilzakaria:
https://github.com/user-attachments/assets/7602cb1b-a17c-4e26-8642-3176cf26d49c

fixes #15112 

## Summary by CodeRabbit

- **New Features**
  - Added comments to team meetups and tasks.
  - Users can view comment counts, expand comment lists, add comments, and delete their own comments.
  - Team leaders can delete comments; comment input is restricted for non-members.
  - Updated meetup and task layouts with inline comment sections.
  - Increased app version to 0.65.47 (build 6547).

- **Tests**
  - Added coverage for comment display, permissions, creation, deletion, and repository behavior.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6546
-        versionName = "0.65.46"
+        versionCode = 6547
+        versionName = "0.65.47"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/data/room/dao/NewsDao.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/data/room/dao/NewsDao.kt
@@ -54,6 +54,15 @@ interface NewsDao {
     @Query("SELECT * FROM news WHERE replyTo = :newsId COLLATE NOCASE ORDER BY time DESC")
     suspend fun getReplies(newsId: String): List<News>
 
+    @Query("SELECT * FROM news WHERE replyTo = :parentId COLLATE NOCASE ORDER BY time ASC")
+    fun getCommentsForParentFlow(parentId: String): Flow<List<News>>
+
+    @Query("SELECT * FROM news WHERE replyTo IN (:parentIds) COLLATE NOCASE ORDER BY time ASC")
+    fun getCommentsForParentsFlow(parentIds: List<String>): Flow<List<News>>
+
+    @Query("SELECT * FROM news WHERE replyTo = :parentId COLLATE NOCASE ORDER BY time ASC")
+    suspend fun getCommentsForParent(parentId: String): List<News>
+
     @Query("SELECT * FROM news WHERE replyTo = :newsId")
     suspend fun getDirectReplies(newsId: String): List<News>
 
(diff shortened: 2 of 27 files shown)


dogi:
**Review — rating: 80/100**

A note on depth first, since this is 1438 lines across 27 files: this is a structural review plus targeted checks on the parts most likely to hurt — the new DAO queries, the DI wiring, and the test coverage — not a line-by-line pass. I'll say what I actually verified.

The shape is right. Nine test files land alongside the feature (`InlineCommentsAdapterTest`, `TeamsTasksAdapterTest`, `TeamsTasksViewModelTest`, `TeamCalendarViewModelTest`, `EventsAdapterTest`, plus repository tests), the comment surface goes through `EventsRepository`/`TeamsRepository` rather than reaching into DAOs from the UI, and the new adapter is a separate `InlineCommentsAdapter` instead of more branching inside the task adapter. That's the decomposition CLAUDE.md asks for.

**The multi-parent query has an `IN` clause with no chunking.**

```kotlin
@Query("SELECT * FROM news WHERE replyTo IN (:parentIds) COLLATE NOCASE ORDER BY time ASC")
fun getCommentsForParentsFlow(parentIds: List<String>): Flow<List<News>>
```

SQLite caps host variables at 999 per statement, so this throws once a team has more than 999 tasks or meetups in one call. The repo already handles this elsewhere — `ResourcesRepositoryImpl` chunks the same shape at `resourceIds.chunked(900)` in two places. Whether a team realistically crosses 999 is a judgement call, but the guard is cheap and the precedent is established.

**The `COLLATE NOCASE` on that same query probably isn't doing anything.** Compare the three new queries:

```sql
WHERE replyTo = :parentId    COLLATE NOCASE   -- attaches to the right operand of =, works
WHERE replyTo IN (:parentIds) COLLATE NOCASE  -- attaches to the IN expression's result
```

`COLLATE` is a postfix operator binding to the preceding expression. In the `=` form that's the comparison operand, which is what you want. In the `IN` form the preceding expression is the whole `replyTo IN (…)` predicate, and collating a boolean has no effect on the comparison — so the multi-parent variant silently loses the case-insensitivity its two siblings have. `WHERE replyTo COLLATE NOCASE IN (:parentIds)` is the form that applies it. In practice CouchDB ids are lowercase hex so this may never bite, which is also why it's easy to miss — worth either fixing or dropping the modifier from all three so they agree.

**`deleteById` alongside `deleteByIds`.** The new single-id delete duplicates what `deleteByIds(listOf(id))` already does. One statement fewer to keep in sync if the caller uses the existing one.

**Drop the version bump.** `app/build.gradle` goes `6546` → `6547`; `.github/scripts/automerge.sh` does that itself at merge time. #15519, #15769 and #15824 all claim `6547` as well right now, so only one of the four merges without a `build.gradle` conflict.

**What I checked and found clean:** the DAO additions are `suspend` or `Flow`-returning as the project requires, `ORDER BY time ASC` is right for a comment thread (the existing `getReplies` uses `DESC`, which is correct for its different purpose), and the new flows are non-suspend as Room demands.

**What I did not verify**, and would want a second pair of eyes on before merge: the comment-posting path through `TeamsRepositoryImpl`/`EventsRepositoryImpl` and whether deletes propagate correctly to the sync layer — `News` rows are CouchDB documents, so a local delete needs to be distinguishable from "never existed" for the uploader. The test files suggest this was thought about; I haven't traced it.

CI green on all four checks; merges cleanly against master.

**Verdict:** `ready`. Chunk the `IN` query, settle the `COLLATE` inconsistency, and drop the version bump — none of which needs another round.

---

Okuro3499 (approved):

J-S-webskas (approved):

PR #15692, opened 2026-08-14
sync: smoother repositories interfaces writing (fixes #15786)

dogi:
Extract CouchDB specific sync methods to dedicated SyncRepository interfaces

---
*PR created automatically by Jules for task [3053900976735220140](https://jules.google.com/task/3053900976735220140) started by @dogi*

PR #15824, opened 2026-08-19
gamification achievement hub offline badges streaks (address #15114)

### LINKED ISSUE #15114: improvement: gamification & achievement hub (offline badges & streaks)
An offline reward system with badges, completion certificates, and study streaks (e.g., "5 Courses Completed", "7-Day Study Streak", "Top Team Contributor"). It keeps learners motivated in remote educational environments where external incentives may be limited.

ragilzakaria:
https://github.com/user-attachments/assets/7fc679c4-efa8-4f1b-968d-0b94d533ee9f

fixes #15114 

## Summary by CodeRabbit

* **New Features**
  * Added a gamification dashboard with study streaks, activity statistics, progress tracking, and achievement badges.
  * Added badge category filters and visual states for locked and unlocked achievements.
  * Added course completion certificates with detailed views and sharing options.
  * Added Gamification and Portfolio tabs to the achievement screen.
* **Style**
  * Updated achievement screens with refreshed cards, colors, layouts, and streak visuals.
* **Chores**
  * Updated the app version to 0.65.47.


dogi:
**Review — rating: 84/100**

Same caveat as I gave #15820: at 2310 lines across 40 files this is a structural review with targeted checks on the wiring, threading and conventions — not a line-by-line pass of the badge and streak logic. I'll be explicit about what I verified.

**The DI wiring is right, which is not the default around here.**

```kotlin
@Singleton
class GamificationRepositoryImpl @Inject constructor(
```

with a matching `@Binds @Singleton` in `RepositoryModule`. Getting the scope onto the *class* rather than only the binding is exactly the thing #15656 needed three rounds to land and #15692 replicated across four repositories — worth calling out that it's correct here from the first commit. With one binding key it wouldn't have mattered yet, but it's the version that stays correct when a second interface gets bound to the same impl.

The constructor also reads like the conventions document: DAOs injected directly (`OfflineActivityDao`, `CourseProgressDao`, `TeamTaskDao`, `NewsDao`, `SubmissionDao`, `UserDao`), sibling repositories where the logic already exists (`ProgressRepository`, `ActivitiesRepository`), and injectable `TimeProvider` and `DispatcherProvider` rather than `System.currentTimeMillis()` and hard-coded `Dispatchers.IO`. The `TimeProvider` choice is what makes streak logic testable at all, and `GamificationRepositoryImplTest` exists to use it.

Threading is handled at the boundary rather than sprinkled:

```kotlin
override fun getGamificationSummaryFlow(userId: String, userName: String): Flow<GamificationSummary> = flow {
    emit(getGamificationSummary(userId, userName))
}.flowOn(dispatcherProvider.io)

override suspend fun getGamificationSummary(…) = withContext(dispatcherProvider.io) { … }
```

The domain types live in `model/gamification/` as plain data classes — `GamificationBadge`, `StudyStreakInfo`, `CourseCertificate`, `BadgeCategory`, `GamificationSummary` — with no Room annotations, so this is derived state computed from existing tables rather than a new persisted surface. That's the right call: no `AppDatabase` version bump needed, and nothing new to sync.

**Drop the version bump.** `app/build.gradle` goes `6546` → `6547`. `.github/scripts/automerge.sh` does this at merge time — every merge commit on master carries its +1. #15519, #15769 and #15820 all claim `6547` right now, so at most one of the four merges without a `build.gradle` conflict.

**One thing worth measuring before it ships.** `getGamificationSummary` fans out across six DAOs plus two repositories and computes badges and certificates on every call. If `AchievementFragment` re-collects on each resume, that's a non-trivial burst of queries on a low-end device — which is the target hardware here. Since it's already a `Flow` behind `flowOn(io)`, caching the summary in the ViewModel and refreshing on an explicit signal rather than per-collection would keep it off the critical path. Worth a look at how `GamificationViewModel` exposes it.

**What I checked and found clean:** four test files (`GamificationRepositoryImplTest`, `GamificationViewModelTest`, `GamificationBadgesAdapterTest`, `CourseCertificatesAdapterTest`), so both the computation and the two adapters have coverage; the new drawables and colour selectors are additive; strings are added to `values/strings.xml`.

**What I did not verify:** the streak arithmetic itself — day-boundary handling, timezone behaviour, and whether a streak survives a device clock change. That's the classic source of bugs in this kind of feature and the place I'd focus a second review. The `TimeProvider` injection means it's testable; whether the tests cover the boundary cases is worth confirming. Also unverified: whether the new strings landed in the five translation files — I only checked `values/strings.xml`.

CI green on all four checks; merges cleanly against master.

**Verdict:** `ready`. Drop the version bump, and give the streak day-boundary tests a look before merge.

---

Okuro3499 (approved):

J-S-webskas (approved):

PR #15832, opened 2026-08-19
sync: smoother manager resources cleaning (fixes #15831)

### LINKED ISSUE #15831: mistaken removals
a transient network blip during resource sync causes valid local resources to be deleted, mistaken for server-side removals.

Okuro3499:
fixes #15831
Prevent deletion of valid local resources when one or more sync batches fail, since the collected ID list would be incomplete and could cause unintended data loss.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved synchronization reliability by preventing cleanup of resources when batch fetching or processing encounters failures.
  * Preserved normal cleanup behavior when synchronization completes successfully.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6550
-        versionName = "0.65.50"
+        versionCode = 6551
+        versionName = "0.65.51"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/services/sync/SyncManager.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/services/sync/SyncManager.kt
@@ -270,6 +270,7 @@ class SyncManager @Inject constructor(
         try {
             val newIds: MutableList<String?> = ArrayList()
             var totalRows = 0
+            var hadBatchFailure = false
 
             // Get total count
             logger.startProcess("resource_get_total_count")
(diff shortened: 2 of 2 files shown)


dogi:
**Review — rating: 88/100**

Small diff, real data-loss bug. Worth landing.

`removeDeletedResources(validNewIds)` deletes every local resource whose id isn't in the list gathered during the sync. The existing guard only checked that no id was null or blank:

```kotlin
if (validNewIds.isNotEmpty() && validNewIds.size == newIds.size) {
    resourcesRepository.removeDeletedResources(validNewIds)
}
```

That catches malformed ids but says nothing about *missing* ones. When a batch failed, the loop logged it, advanced `skip`, and carried on — so `newIds` was silently short by a whole page, every id in it was well-formed, the guard passed, and the cleanup deleted resources that exist perfectly well on the server. On a flaky connection that's local content disappearing for no reason the user can see, and it comes back only after a later full sync.

`hadBatchFailure` is the right complement to the existing check: one asks "are the ids we have valid", the other asks "do we have all of them".

**I checked that the flag covers every way out of the loop.** There are exactly two failure paths and both set it — the `response == null` branch and the `catch (e: Exception)` at the bottom. The only other early exit is

```kotlin
if (rows.size() == 0) {
    break
}
```

which is normal end-of-pagination, not a failure, and correctly leaves the flag alone. So the guard is complete rather than covering the obvious cases and missing one.

The `logger.logDetail` on the skip path is a good call — a cleanup that silently doesn't happen is its own kind of confusing, and this leaves a trail in the sync log explaining why.

**The missing piece is a test.** `SyncManagerTest` exists, and this is precisely the kind of fix that deserves a regression guard, because the failure mode is invisible: nothing throws, nothing turns red, resources just quietly vanish. A test that makes one batch return `null` and asserts `removeDeletedResources` is never called would pin the behaviour down permanently. Given the consequence is data loss, I'd want that before the next person refactors this loop.

**Minor:** `skip += batchSize` on the failure paths versus `skip += rows.size()` on success is correct — advancing by the requested page size is what you want when you don't know how many rows a failed page held — but it does mean a failed batch is skipped rather than retried within this sync. That's pre-existing and arguably right, since the cleanup is now correctly suppressed; just noting that "one bad batch" means those resources aren't refreshed until the next sync.

**Verdict:** `ready`. Add the `SyncManagerTest` case when you can — the fix itself is correct and complete.

---

dogi (approved):

PR #15833, opened 2026-08-19
sync: smoother upload teams managing (fixes #15830)

### LINKED ISSUE #15830: duplicated teams
new teams created offline never get marked uploaded, so they're duplicated on CouchDB every sync.

Okuro3499:
fixes #15830
Replace map-based team lookup with index-based matching for bulk upload responses. The previous approach used teamId as a map key, but the response 'id' may not match the request's teamId. Using positional matching aligns with CouchDB's bulk_docs response order guarantee.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved team upload handling when server-generated IDs differ from local IDs.
  * Prevented teams from being incorrectly marked as uploaded.
  * Automatically retries teams that were not included in an incomplete bulk-upload response.


dogi:
**Review — rating: 72/100**

The diagnosis is right and index-matching is the correct mechanism. I don't think this closes #15830 on its own, though — the half that actually marks a team as uploaded is untouched.

**Why the change is right.** CouchDB's `_bulk_docs` returns results in the same order as the submitted `docs` array, so `batch.getOrNull(i)` is the canonical correlation. The old `teamMap[id]` keyed on `teamData.teamId`, which is `team._id`:

```kotlin
TeamUploadData(
    teamId = team._id,
    serialized = MyTeam.serialize(team, courses, coursesResourcesMap),
    …
)
```

and `MyTeam.serialize` emits `_id` through `JsonUtils.addString(object, "_id", team._id)`. For a team that has never been uploaded `_id` is null, so the payload carries no `_id`, CouchDB generates one, and the returned id matches nothing in `teamMap` — `continue`, nothing recorded, re-uploaded next sync. That's the duplicate-teams loop. Matching by index recovers the local record regardless, which is the fix.

It also removes the `org.ole.planet.myplanet.repository.TeamUploadData` fully-qualified name along with the map. Good.

**Where it stops short.** Having recovered `teamData`, the success path still keys the result map by the *server* id:

```kotlin
uploadedTeams[id] = rev
```

and `markTeamsUploaded` resolves that against local rows:

```kotlin
val teamsToUpdate = teamDao.getAll()
    .filter { (it._id ?: it.id) in uploadedTeams.keys }
```

For an existing team, `_id` is the server id and this matches. For the newly created team — the case this PR exists for — the local row still has `_id == null`, so the filter falls back to `it.id` (the locally generated value), which is not the server id CouchDB just returned. No row matches, `updated` stays true, and it uploads again next sync. Same loop, one step later.

`uploadedTeams[teamData.teamId ?: id] = rev` would close it, since `teamData` is exactly the local record you now have in hand. Worth checking on a device with a freshly created team rather than taking my word for it — enable the sync log, create a team, sync twice, and see whether a second document appears server-side.

There's a related gap underneath: even once the row is marked, nothing writes the server-assigned `_id` back to it, so the local record stays `_id == null` and every subsequent upload creates yet another document. That's bigger than this PR and probably its own issue, but it's the reason I'd want the device check before calling #15830 fixed.

**Two smaller things.**

The loop iterates `0 until responseBody.size()` and indexes into `batch`. `getOrNull` makes an over-long response safe, but a *short* response silently leaves the tail of the batch unprocessed — those teams are neither marked uploaded nor queued for retry. A length check with a log line would make that visible instead of silent.

No test. `UploadManagerTest` exists, and the index-vs-id distinction is straightforward to pin down: a batch of two where the response returns server-generated ids, asserting both local records are resolved. That's also the test that would have caught the `uploadedTeams` key issue above.

CI green on all four checks; merges cleanly, and no conflicts with #15832, #15835 or #15837.

**Verdict:** `change` — the index correlation is right, keep it. Key `uploadedTeams` by the local identifier so the marking half works too, and confirm on a device that a newly created team stops re-uploading.

---

dogi:
**Review — rating: 84/100**

The diagnosis is right and the root cause is worth stating explicitly: `TeamUploadData.teamId` is `team._id`, and `MyTeam.serialize` writes `_id` from that same field — so for a team created locally it goes up without a usable `_id`, CouchDB assigns one, and the response `id` matches nothing in `teamMap`. The old `teamMap[id] ?: continue` then swallowed the entry, `uploadedTeams` never got a row for that team, `markTeamsUploaded` never cleared its dirty flag, and the next sync uploaded it again — a fresh server document every time. Matching by index is the correct join here, not a workaround: `_bulk_docs` returns one result per submitted document in submission order, which is the contract CouchDB clients rely on, and it's the only key that survives the server minting the id.

The more valuable part of the change is the distinction it draws, which the comment captures well: the response `id` is now used for *server*-side work (`uploadTeamImageAttachment`, which must address the real document) and `teamData.teamId` for *local* bookkeeping (`uploadedTeams`, which `markTeamsUploaded` resolves against the local record's `_id`). Those two identifiers genuinely differ for new teams and the old code conflated them. The test pins exactly that: `TeamUploadData("localTeam1", …)` against a response of `serverGeneratedId1` asserting `markTeamsUploaded(mapOf("localTeam1" to "rev1"))` fails on master and passes here.

**The split isn't applied to two sibling sites in the same block, though.** `deletedIds.add(id)` still uses the response id, and it feeds `deleteLocalTeamRecords` → `teamDao.deleteById`, which is local bookkeeping and wants `teamData.teamId` by the same argument. It's latent rather than live — a delete-pending team already exists on the server, so `serialize` sends a real `_id` and the two coincide — but it's the identical bug shape sitting two lines away. Likewise `queueTeamRetry(teamData, …, id)` in the error branch passes the response id as `dbId`, while the bulk-failure branch below passes `teamData.teamId` and the existing test asserts `dbId = "team2"`. Worth making both consistent while the reasoning is fresh.

**Nit:** logging the short-response case instead of silently dropping teams is the right call, but the loop then proceeds on the mismatch. Index alignment only holds if a short response is truncated at the tail; an omitted middle entry would pair team *i*'s rev with team *j* and write a wrong rev onto a team — worse than the re-upload being fixed. CouchDB's one-result-per-doc guarantee means this shouldn't arise, and the warning makes it diagnosable if it ever does, but bailing out of the loop on a size mismatch would cost nothing and close it off.

**Verdict:** correct fix for a real duplicate-creating sync bug, with a regression test that pins it — ready; the two leftover keying sites are worth a follow-up.

---

PR #15835, opened 2026-08-19
sync: smoother server url alternative credentials mapping (fixes #15834)

### LINKED ISSUE #15834: breaking auth when the primary server is down
fallback to a clone server extracts credentials from the wrong URL, breaking auth when the primary server is down (the actual scenario the fallback exists for).

Okuro3499:
fixes #15834
Extract credentials from the alternative URL's userinfo instead of the primary URI when the alternative URL contains '@'. Moved altUri parsing before the credential extraction block so getUserInfo receives the correct URI.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6551
-        versionName = "0.65.51"
+        versionCode = 6552
+        versionName = "0.65.52"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/services/sync/ServerUrlMapper.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/services/sync/ServerUrlMapper.kt
@@ -52,11 +52,12 @@ class ServerUrlMapper @Inject constructor(
     }
 
     fun updateUrlPreferences(editor: SharedPreferences.Editor, uri: Uri, alternativeUrl: String, url: String, settings: SharedPreferences) {
+        val altUri = alternativeUrl.toUri()
         val urlUser: String
         val urlPwd: String
 
         if (alternativeUrl.contains("@")) {
-            val userinfo = getUserInfo(uri)
+            val userinfo = getUserInfo(altUri)
             urlUser = userinfo[0]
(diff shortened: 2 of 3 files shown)


dogi:
**Review — rating: 90/100**

One-line fix, correct, with a test that pins exactly the thing that was wrong. This is the shape a bug fix should have.

The branch is entered because the *alternative* URL contains `@`, so the credentials are in the alternative URL — but it was reading them out of the primary one:

```kotlin
if (alternativeUrl.contains("@")) {
    val userinfo = getUserInfo(uri)   // ← primary
```

When the primary URL carries no user info, `getUserInfo` returns empty strings, so `url_user` and `url_pwd` get blanked out and every subsequent authenticated request against the clone fails. Moving `altUri` above the branch and reading from it is the minimal correct change.

**The credential/host pairing is coherent, which is what makes this right rather than just different.** `url_Host` and `url_Scheme` still come from the primary `uri` — I checked the three pre-existing tests and they all assert `url_Host == "primary.com"`, so that's deliberate and unchanged. When `isAlternativeUrl` is set, `UrlUtils.baseUrl` uses `getProcessedAlternativeUrl()`, which this function writes from `altUri`. So credentials extracted from the alternative URL are used against the alternative URL. The two travel together.

**The test is the right test.** It's the only one of the four that sets up a primary with no user info and an alternative that has some:

```kotlin
every { uri.userInfo } returns null
val alternativeUrl = "http://clone_user:clone_pass@alternative.com:5984"
…
verify { editor.putString("url_user", "clone_user") }
verify { editor.putString("url_pwd", "clone_pass") }
```

Under the old code `getUserInfo(uri)` would have produced empty strings and both assertions would fail, so this genuinely guards the regression rather than just exercising the path. It also verifies `processedAlternativeUrl` and the primary host stay put, so a future "fix" that over-corrects and swaps the host too would be caught.

**Tiny nit:** `every { uri.userInfo } returns null` is now unnecessary, since the fixed code never calls it. Harmless, and arguably useful as documentation of the scenario — leaving it is fine.

The class is Robolectric-backed (`@RunWith(RobolectricTestRunner::class)`), so `alternativeUrl.toUri()` parses for real rather than against a stub. That matters here: a plain JUnit test would have thrown on `Uri.parse` and told you nothing.

CI green on all four checks; merges cleanly, and no conflicts with #15832, #15833 or #15837.

**Verdict:** `ready`.

---

dogi (approved):

PR #15837, opened 2026-08-19
sync: smoother user repository security data preserving (fixes #15836)

### LINKED ISSUE #15836: breaking login when server response missing
a server response missing derived_key/salt wipes those fields locally, breaking login

Okuro3499:
fixes #15836
Only overwrite derived_key, salt, password_scheme, and iterations in updateSecurityData when the server provides non-null values. This prevents clearing existing credentials during sync when the server response omits security fields.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6552
-        versionName = "0.65.52"
+        versionCode = 6553
+        versionName = "0.65.53"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/repository/UserRepositoryImpl.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/UserRepositoryImpl.kt
@@ -405,10 +405,10 @@ class UserRepositoryImpl @Inject constructor(
         val user = getUserByName(name) ?: return
         user._id = userId
         user._rev = rev
-        user.derived_key = derivedKey
-        user.salt = salt
-        user.password_scheme = passwordScheme
-        user.iterations = iterations
+        derivedKey?.let { user.derived_key = it }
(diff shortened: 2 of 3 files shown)


dogi:
**Review — rating: 86/100**

Four lines, and they protect the thing that makes offline login work at all.

```kotlin
-        user.derived_key = derivedKey
-        user.salt = salt
-        user.password_scheme = passwordScheme
-        user.iterations = iterations
+        derivedKey?.let { user.derived_key = it }
+        salt?.let { user.salt = it }
+        passwordScheme?.let { user.password_scheme = it }
+        iterations?.let { user.iterations = it }
```

`AndroidDecrypter.androidDecrypter()` verifies passwords with PBKDF2 against the stored `derived_key`, `salt`, `iterations` and `password_scheme`. Those four are the entire offline-auth story — the server is not reachable when they matter. Overwriting them with nulls because a particular server response omitted them doesn't degrade login, it removes it: the user is locked out of their own device until the next successful online sync repopulates the fields, which on a low-connectivity deployment could be days.

Null-coalescing is the right semantics here specifically because there's no legitimate "clear the credential" case. A CouchDB `_users` document that doesn't include the password fields means "I'm not telling you", not "this user has no password" — those documents routinely come back without them depending on the view and the requesting role. Treating absent as no-change is the only reading that matches the data.

`user._id` and `user._rev` are correctly left as unconditional assignments — those *are* meant to track the server, and a null there is a different problem that shouldn't be masked. Good line to draw.

`UserRepositoryImplTest` is updated, which is what I'd want: the whole point is behaviour on a response shape that's hard to hit by hand.

**Worth a look while you're in here, not blocking:** if a response omits these fields on a user who has *never* had them stored locally, the row still ends up without credentials and offline login still won't work for that account — this fix prevents losing what you had, it doesn't establish what you never got. If that's reachable in practice, the useful signal would be a log line when the fields are absent, so the next "I can't log in offline" report has something to go on.

CI green on all four checks; merges cleanly, and no conflicts with #15832, #15833 or #15835.

**Verdict:** `ready`.

---

dogi (approved):

PR #15838, opened 2026-08-19
login: smoother learner registering (fixes #15556)

### LINKED ISSUE #15556: Rotating device to landscape incorrectly triggers submission pickup and error prompt
When rotating the device to landscape orientation, the app sometimes picks up a submission unexpectedly and shows an error prompt.
This should not happen.

**Steps to reproduce**
Open the app and go to a screen with submission flow.
Rotate the device from portrait to landscape.
Observe a submission may be picked up automatically.
An error prompt appears.
https://github.com/user-attachments/assets/41db659b-1ee6-4cd7-8e3a-90f977234e4e

J-S-webskas:
https://github.com/user-attachments/assets/47719ca9-f382-421c-acbf-c4e5ef66547f

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved username validation by clearing previous errors when the field is emptied.
  * Prevented unnecessary validation requests for blank usernames.


dogi:
**Review — rating: 82/100**

Correct fix for the reported symptom, six lines, no side effects I can find.

On rotation the `TextWatcher` fires against an empty username field, which previously went straight into `userRepository.validateUsername("")` and surfaced whatever error that returns — on a field the user has never touched. Returning early on empty input, and clearing any stale `error` on the way out, is the right shape: it treats "empty" as "nothing to say yet" rather than "invalid", which is what an untouched field should mean.

**I checked the stale-job concern and it's already handled.** `usernameValidationJob?.cancel()` now sits *after* the early return, so typing `a` and then deleting it leaves the 300ms job for `"a"` alive. But the coroutine re-reads the field before applying anything:

```kotlin
val error = userRepository.validateUsername(input)
if (activityBecomeMemberBinding.etUsername.text.toString() != input) {
    return@launch
}
```

The field is now empty, `"" != "a"`, and it bails without touching the UI. So no error appears on the cleared field — the guard covers it.

What's left is a wasted call rather than a bug: that job still runs `validateUsername("a")` — which may hit the repository — before discarding the result. Moving `usernameValidationJob?.cancel()` above the `if (input.isEmpty())` block costs nothing and makes the cancellation unconditional, which is also what a reader would assume from the code as written.

**On testing:** none here, and I don't think that's a gap worth pressing. This lives in a `TextWatcher` inside an Activity, the project has no `androidTest` source set, and the debounce-plus-recheck pattern isn't reachable from the unit suite. The behaviour is cheap to confirm by hand instead — open Become a Member, rotate without typing, confirm no error; then type a name, rotate, confirm the validation still runs against the restored text.

**Worth confirming in that same pass:** the early return clears `error` but doesn't reset any other validation state the screen might key off (a submit-button enablement flag, say). If the form tracks validity separately from the `error` property, an empty field should mark it invalid-but-unannotated rather than simply clearing the message.

**Verdict:** `ready`. Move the `cancel()` above the early return when convenient.

---

Okuro3499 (approved):

ragilzakaria (approved):

dogi (approved):

PR #15841, opened 2026-08-19
dashboard: smoother courses shelf handling (fixes #15727)

### LINKED ISSUE #15727: empty courses navigation
For a new user who has not joined any course if they click mycourses or anywhere on the section show them the list of existing courses and ensure that if they joined any course the current behavior persists where they are show list of courses they joined

https://github.com/user-attachments/assets/771e63f4-4763-481d-82d0-248d7fb9485f

J-S-webskas:
https://github.com/user-attachments/assets/44966548-64c6-434e-897d-c8411d8d4b8a

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved dashboard course navigation.
  - The Courses button now opens the course list when courses are available and the course discovery view when none are available.
  - Course listings now consistently retain only valid, nonblank course entries for display.


Okuro3499 (changes requested):
text at the top should not read myCourses as that is not the list being shown
https://github.com/user-attachments/assets/3d73a1f2-daca-4809-a943-8dfa2b2ad8cc

ragilzakaria (dismissed):

dogi:
**Review — rating: 80/100**

This is a consistency fix and the PR body undersells it: `openLibraryAction` at `BellDashboardFragment:327` already branches on `userLibrary.isNotEmpty()` to pick `openMyFragment` vs `openCallFragment`, and the courses button was the odd one out, always going to `openMyFragment`. So a user with no joined courses was landing in the "my courses" view (`isMyCourseLib = true`) instead of the full course list. Extracting `openCoursesAction` as a `val` alongside its sibling and caching `userCourses` in `BaseDashboardFragment` mirrors the library path exactly.

**Nit:** three small things, all in the diff:

- `import org.ole.planet.myplanet.model.MyCourse` is added to `BellDashboardFragment` but nothing in that file references the type — the only `MyCourse` occurrence is the `"isMyCourseLib"` string literal at line 364. `userCourses` is declared in the base class, so the import is dead. Kotlin only warns on unused imports, which is why CI is green.
- `userCourses = filteredCourses` sits *after* `val flexboxLayout = view?.findViewById(...) ?: return`, so if that view is absent the field never gets set and the button silently takes the empty-list branch. `renderMyLibrary` assigns `userLibrary = dbMylibrary` as its first statement, before touching any view — worth matching, it's a one-line move.
- The `CoursesViewModel.kt` hunk only deletes a blank line and has nothing to do with this change; dropping it keeps the diff honest about its scope.

**Verdict:** correct and consistent with the library path; the three nits are all one-liners.

---

Okuro3499 (approved):

dogi (approved):

PR #15843, opened 2026-08-19
all: smoother json utils handling (fixes #15984)

### LINKED ISSUE #15984: perf: prevent eager object and string allocations in JsonUtils
Tracking issue for PR #15843.

Optimizes `JsonUtils` string/object allocations on CouchDB document read hot paths.

---
*PR created automatically by Jules for task [7731212617685698431](https://jules.google.com/task/7731212617685698431) started by @dogi*

dogi:
Optimizes `JsonUtils` string/object allocations on CouchDB document read hot paths.

---
*PR created automatically by Jules for task [7731212617685698431](https://jules.google.com/task/7731212617685698431) started by @dogi*


Okuro3499 (dismissed):

dogi:
**Review — rating: 74/100**

The premise is correct and I verified each change is behavior-preserving:

- **`getInt`/`getFloat`**: for a number primitive the old code evaluated `el.asString.isEmpty()` — allocating a String just to discard it — before falling through to `el.asInt`. The new fast path returns `el.asInt` directly, and every non-number element still takes the original expression. Same results, one less allocation per field. These run per-field per-document during CouchDB sync, so it's a real if small win.
- **`getJsonArray`**: `array.asJsonArray` → `array` is a no-op simplification; `array` is already smart-cast to `JsonArray` at that point.
- **`getJsonElement`**: the old `val default = …` was built on every call even when the field was present, and `safeGet(JsonObject())` allocated a second throwaway. That's the eager allocation in the title, and it's real — `safeGet` takes `default` as an eagerly evaluated parameter.
- The changed exception fallback (always `JsonObject()` → type-dependent) is unobservable: `getJsonElement` has exactly two callers, both inside this file, and each coerces the result to its own type anyway (`array !is JsonArray → JsonArray()`, `el is JsonObject → el else JsonObject()`).

So nothing here is wrong. My objection is that the fix is in the wrong place:

**`safeGet` is already `inline`.** Changing its parameter to a lambda —

```kotlin
private inline fun <T> safeGet(default: () -> T, block: () -> T): T =
    try { block() } catch (e: Exception) { e.printStackTrace(); default() }
```

— removes the eager allocation at *every* call site at zero runtime cost, keeps all the accessors uniform, and needs no duplicated try/catch. As written, this PR fixes one of the three object/array sites and leaves the other two (`getJsonArray`'s `safeGet(JsonArray())` and `getJsonObject`'s `safeGet(JsonObject())`) still allocating eagerly — inconsistent with its own stated goal — while `getJsonElement` becomes the one function in the object that hand-rolls the try/catch and `e.printStackTrace()` that `safeGet` exists to centralize.

**No tests.** `JsonUtilsTest` exists but currently covers none of `getInt`, `getFloat`, `getJsonArray`, or `getJsonElement`. For a change whose entire argument is "identical behavior, fewer allocations", the equivalence is exactly what should be pinned: number, numeric string, empty string, `JsonNull`, missing field, and wrong-type element.

**Nit:** the new `if (isNumber) el.asInt else if (JsonNull || asString.isEmpty()) 0 else el.asInt` has `el.asInt` in two branches; a `when` reads better and makes the fast path obvious.

**Nit:** with both callers now internal, `getJsonElement` could be `private`.

**Verdict:** needs another round — please make `safeGet`'s default a lambda and fix all three sites uniformly instead of bypassing the helper, and add the equivalence tests.

---

ragilzakaria (dismissed):
**Rating:** 85/100

The refactoring to make `safeGet` take a lazy `default: () -> T` lambda is clean and behavior-preserving. Since `safeGet` is `inline`, this avoids allocating lambda instances at runtime while deferring default object instantiations across all accessors. The `when` branch fast-path in `getInt` and `getFloat` also successfully avoids throwaway string allocations on numeric primitives.

**Items to address before merging:**

1. **Missing Test File**:
   - The latest commit message notes *"Adds equivalence tests for number parsing and object/array fallback generation"*, but no test file was included in the commit (only `JsonUtils.kt` was modified).
   - Please add tests in `JsonUtilsTest.kt` covering `getInt`, `getFloat`, `getJsonArray`, and `getJsonObject` across numbers, numeric strings, empty strings, `JsonNull`, and missing keys to lock in equivalence.

2. **Redundant Allocation in `getJsonElement`**:
   ```kotlin
   private fun getJsonElement(fieldName: String, jsonObject: JsonObject, type: Class<*>): JsonElement = safeGet({
       if (type == JsonObject::class.java) JsonObject() else JsonArray()
   }) {
       if (jsonObject.has(fieldName)) jsonObject.get(fieldName)
       else if (type == JsonObject::class.java) JsonObject() else JsonArray()
   }

dogi:
**Re-review — Rating: 90/100** (was 74)

All four items from the last round are addressed, and addressed the right way round:

- **`safeGet`'s default is now a lambda**, so the eager allocation is gone at *every* call site rather than just `getJsonElement`'s — including the two I flagged as inconsistent (`getJsonArray`'s `JsonArray()` and `getJsonObject`'s `JsonObject()`). Since `safeGet` is `inline`, the lambdas don't survive compilation, so this costs nothing at runtime.
- **`getJsonElement` no longer hand-rolls the try/catch** and is `private`. The green build is real evidence there were no external callers, since Kotlin would fail on an inaccessible reference.
- **The `when` replaced the doubled `el.asInt`**, and the number fast path reads clearly.
- **Tests now pin the equivalence.** I checked the case that actually matters: `getInt("wrongType", …)` where the value is a `JsonObject` takes `isJsonPrimitive == false` → `el is JsonNull` false → `el.asString` throws `UnsupportedOperationException` → `safeGet` catches → `0`. So the assertion covers the exception fallback, not just the happy path, and the same holds for `getFloat`. Number, numeric-string, empty-string, `JsonNull`, missing and wrong-type are all covered for all four accessors.

One behavior note I verified rather than assumed: moving the missing-field branch *out* of `safeGet` in `getJsonElement` is safe — `jsonObject.has(fieldName)` can't throw for a non-null receiver, so nothing that used to be caught is now escaping.

**Nit:** `JsonArray` is referenced fully qualified as `com.google.gson.JsonArray()` in the new tests while `JsonObject` is imported; adding the import would make them consistent.

**Verdict:** mergeable.

---

dogi:

The lambda default is in — `safeGet(default: () -> T)` with `{ JsonArray() }` / `{ JsonObject() }` at the call sites — so the container allocation only happens on the exception path it exists for. That was the whole point of the PR and it now actually holds.

Two things beyond that which improve it further:

- **`getInt`/`getFloat` check `isJsonPrimitive && isNumber` first**, so a numeric field goes straight to `asInt`/`asFloat` instead of routing through `el.asString.isEmpty()`. String-encoded numbers still work through the fallback, which the tests pin (`"42"` → 42, `"42.5"` → 42.5f).
- **`getJsonElement` is now private** and no longer wraps the whole body in `safeGet` — the missing-field case returns early without a try/catch, and only `jsonObject.get(fieldName)` is guarded. Narrowing it is safe (both build variants and both test shards compile green, so nothing outside the object called it), and it's a better boundary.

`else array` in place of `else array.asJsonArray` is equivalent — the smart cast already made it a `JsonArray`, and `asJsonArray` returned `this`.

The new tests are aimed well: each of the four getters is checked against a correct value, a string-encoded value, empty, `JsonNull`, missing, and a wrong-typed element. The wrong-type cases are what exercise the `safeGet` path, so the lambda default is covered rather than merely present.

**Nit:** the tests write `com.google.gson.JsonArray()` fully qualified six times while the file already imports `JsonObject` and `JsonNull` from the same package — an `import com.google.gson.JsonArray` would tidy that.

**Nit:** `assertEquals(42.5f, JsonUtils.getFloat("num", obj))` relies on boxed `Float.equals` rather than JUnit's float comparison. It passes because 42.5 is exactly representable, but `assertEquals(42.5f, …, 0.0f)` states the intent and won't surprise anyone who changes the value later.

**Verdict:** mergeable.

---

Okuro3499 (approved):

Okuro3499 (approved):

PR #15844, opened 2026-08-19
resources: smoother viewer video handling (fixes #15985)

### LINKED ISSUE #15985: Fix ExoPlayer leak by releasing before reassignments
Tracking issue for PR #15844.

Fix ExoPlayer leak by releasing before reassignments in `ResourceViewerFragment.kt`. Release the previous `ExoPlayer` instance and set it to null before assigning a fresh player during `prepareVideoPlayer`, `streamVideoFromUrl`, and `initializeAudioPlayer`. Also moved the `OnAudioRecordListener.onRecordStopped` coroutine from `lifecycleScope` to `viewLifecycleOwner.lifecycleScope` to match lifecycle boundaries.

---
*PR created automatically by Jules for task [13276721768784156507](https://jules.google.com/task/13276721768784156507) started by @dogi*

dogi:
Fix ExoPlayer leak by releasing before reassignments in `ResourceViewerFragment.kt`. Release the previous `ExoPlayer` instance and set it to null before assigning a fresh player during `prepareVideoPlayer`, `streamVideoFromUrl`, and `initializeAudioPlayer`. Also moved the `OnAudioRecordListener.onRecordStopped` coroutine from `lifecycleScope` to `viewLifecycleOwner.lifecycleScope` to match lifecycle boundaries.

---
*PR created automatically by Jules for task [13276721768784156507](https://jules.google.com/task/13276721768784156507) started by @dogi*


Okuro3499 (approved):

dogi:
**Review — rating: 84/100**

Correct and low-risk. Releasing before overwriting the only reference to an `ExoPlayer` is unambiguously right, and `exoPlayer?.release()` is a no-op when the field is already null, so none of the three sites can regress. I checked the ordering at each: release → create → `playerView.player = …` is straight-line main-thread code, so the `PlayerView` never observes the released instance.

Context I verified around it:

- `onDestroyView` already does `exoPlayer?.release(); exoPlayer = null`, so teardown was covered. What this fixes is re-initialization *within* one view lifecycle, where the previous player was orphaned.
- The main re-entry path is guarded — `setAuthSession` only calls `streamVideoFromUrl` when `exoPlayer == null`, otherwise it just updates the data-source request properties. So on the paths I can trace, this is defense rather than a leak you'll hit today. That's still worth having at three lines, but the PR description would be stronger if it named a reproducing path.

**Nit — unrelated scope.** The `lifecycleScope` → `viewLifecycleOwner.lifecycleScope` change in `onRecordStopped` has nothing to do with the ExoPlayer leak; it belongs in #15854, which is exactly that migration. It's directionally right and doesn't introduce a new crash class — the same callback already touches `requireContext()` and `binding.fabRecord`, so it would fail on the binding first if the view were gone.

**Nit — adjacent gap, pre-existing.** `audioRecorder` is never stopped in `onDestroyView` (only `authSessionUpdater?.stop()` and the player release are there), so leaving the screen mid-recording keeps the recorder alive and lets `onRecordStopped` fire against a destroyed view. Same family of bug as this PR; worth a follow-up.

**Verdict:** mergeable.

---

PR #15845, opened 2026-08-19
life: smoother health examination view modelling (fixes #15986)

### LINKED ISSUE #15986: Refactor HealthExaminationActivity to use ViewModel for state and lifecycle-aware collection
Tracking issue for PR #15845.

Refactor HealthExaminationActivity to use ViewModel for state and lifecycle-aware collection

Removes the direct `HealthRepository` injection from the Activity by moving the condition-fetching logic inside `HealthExaminationViewModel`. 
Updates `HealthExaminationState` to carry the combined output, which `HealthExaminationActivity` now listens to using `collectWhenStarted` to prevent uncancelled memory and thread leaks while stopped. 
Fixes unit tests to cover the new state property.

---
*PR created automatically by Jules for task [13049281141106073965](https://jules.google.com/task/13049281141106073965) started by @dogi*

dogi:
Refactor HealthExaminationActivity to use ViewModel for state and lifecycle-aware collection

Removes the direct `HealthRepository` injection from the Activity by moving the condition-fetching logic inside `HealthExaminationViewModel`. 
Updates `HealthExaminationState` to carry the combined output, which `HealthExaminationActivity` now listens to using `collectWhenStarted` to prevent uncancelled memory and thread leaks while stopped. 
Fixes unit tests to cover the new state property.

---
*PR created automatically by Jules for task [13049281141106073965](https://jules.google.com/task/13049281141106073965) started by @dogi*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved health examination loading so examination details and condition selections appear consistently.
  * Ensured condition information is available when completing and saving an examination.

* **Chores**
  * Updated the Android app version to 0.65.56.


Okuro3499 (changes requested):
```
Moving the state collection into collectWhenStarted (repeatOnLifecycle STARTED + StateFlow replay) causes validateFields() to run again on every STOP→START transition, 
each time registering an additional doOnTextChanged listener on etBloodpressure without removing the prior one.
```

Okuro3499 (dismissed):
```
Problem: collectWhenStarted(viewModel.state) re-runs initExamination()/validateFields() on every STOPPED→STARTED transition because repeatOnLifecycle replays the current        
StateFlow value from scratch, overwriting user-entered fields and duplicating the doOnTextChanged listener each time.                                                                
```

dogi:
**Review — rating: 65/100**

The ViewModel half is right, but swapping the one-shot collect for `collectWhenStarted` introduces a user-visible regression. `collectWhenStarted` is `repeatOnLifecycle(STARTED)`, so every STOPPED→STARTED transition starts a fresh collection and the `StateFlow` immediately replays its current value — the collector body runs again on **every return to the foreground**. Previously `lifecycleScope.launch { state.collect { … } }` began collecting once and `loadData` sets `_state.value` a single time, so the body ran exactly once.

Two concrete consequences:

1. **Edits are silently discarded.** `initExamination()` calls `setText(…)` on 15 EditTexts (temperature, pulse, BP, height, weight, vision, hearing, and the encrypted note/diagnosis/treatments/… fields). A health worker editing an existing examination who takes a call and comes back gets every field overwritten with the persisted values. This one only bites when `examination != null`, but that's exactly the edit flow.
2. **Listeners stack.** `validateFields()` registers `binding.etBloodpressure.doOnTextChanged { … }`, and `doOnTextChanged` *adds* a TextWatcher on each call. Every foreground return adds another watcher to the same field, all mutating the shared `allowSubmission` flag — a leak plus validation that gets progressively less predictable.

`btnSave.isEnabled = true` re-running is harmless, but the other two aren't.

What's good and should be kept as-is:

- Dropping `@Inject lateinit var healthRepository` from the Activity — the Activity had no business holding a repository.
- Folding `getExaminationConditions(examination)` into the ViewModel's existing `withContext(dispatcherProvider.io)` block. That's the correct layer, and it makes the state atomic: `conditionsMap` previously arrived via a nested `lifecycleScope.launch` *after* the state emission, so `initExamination()` was gated on a second coroutine.
- Removing the nested launch, and updating the test to stub and assert `conditionsMap`.

**Suggested fix:** keep every ViewModel change and separate one-time setup from ongoing observation — register the validation listeners once in `onCreate`, and either consume the first loaded state (`viewModel.state.first { !it.isLoading }`) or guard `initExamination()` behind a flag so a replay can't clobber in-progress input. `collectWhenStarted` is the right tool for observing state; it just can't wrap one-shot initialization.

**Verdict:** needs another round — the collector body isn't replay-safe.

---

ragilzakaria (approved):

dogi:
**Re-review — Rating: 90/100** (was 65)

The replay problem is gone, and the fix took the cleaner of the two options I suggested: `viewModel.state.first { !it.isLoading }` consumes the loaded state exactly once, so the whole initialization block — `initExamination()`, `validateFields()`, `btnSave.isEnabled = true` — runs a single time per Activity instance.

I checked the two failure modes from the last round and both are closed:

- **Edits survive backgrounding.** With no `repeatOnLifecycle`, there's no re-collection on STOPPED→STARTED, so the 15 `setText(…)` calls in `initExamination()` can't overwrite in-progress input any more.
- **Watchers no longer stack.** `doOnTextChanged` in `validateFields()` is reached once, so only one TextWatcher is ever attached to `etBloodpressure`.

I also verified the predicate can't short-circuit on the initial value: `HealthExaminationState.isLoading` defaults to `true` and `loadData` re-asserts `copy(isLoading = true)` before the IO block, so `first { !it.isLoading }` genuinely waits for the loaded state rather than returning the empty one. Because `state` is a `StateFlow`, it also doesn't matter that `loadData` is called before the collector starts — the final value is replayed either way, so there's no race.

Folding `getExaminationConditions(examination)` into the existing `withContext(dispatcherProvider.io)` block is the right layering and makes the state atomic — `conditionsMap` now arrives in the same emission as `examination` instead of via a second coroutine, which is what let the old code call `initExamination()` before conditions were known. The test stubs and asserts the new field.

**Nit:** `import org.ole.planet.myplanet.repository.HealthRepository` in `HealthExaminationActivity` is now unused — the field and its one call site were both removed. Worth deleting with the rest.

**Verdict:** mergeable.

---

dogi:

Both halves are right, and I checked the one thing that could have made `first` wrong: `HealthExaminationState.isLoading` defaults to `true` and `_state` starts at `HealthExaminationState()`, so `state.first { !it.isLoading }` genuinely waits for the loaded snapshot rather than returning the empty initial one.

That replaces a `collect` that re-ran its whole body on every state emission — and launched a *nested* `lifecycleScope.launch` each time to fetch conditions — with a single suspend-until-ready. `initExamination()`, `validateFields()` and enabling the save button now happen exactly once.

Folding `conditionsMap` into `HealthExaminationState` is the better fix than fetching it from the Activity: `getExaminationConditions(examination)` runs inside the same `withContext(dispatcherProvider.io)` block as the rest of the load, so it's off the main thread and consistent with the `examination` it was derived from — no window where one is set and the other isn't. And `HealthRepository` leaves the Activity entirely, which is the layering `CLAUDE.md` asks for.

The test update matches: it stubs `getExaminationConditions(mockExamination)` and asserts `conditionsMap` on the final state, so the new field is covered rather than just carried.

**Nit:** `lifecycleScope.launch { state.first { … } }` runs once for the Activity's lifetime, which is correct here — but it means a later `loadData` call (it does `copy(isLoading = true)` at the top) would not re-drive the UI. Only `onCreate` calls it today, so this is a note for whoever adds a refresh path, not a change request.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #15846, opened 2026-08-19
courses: smoother repository dao querying (fixes #15987)

### LINKED ISSUE #15987: perf(courses): optimize courses shelf queries
Tracking issue for PR #15846.

- Added `getForUserPattern` and `observeForUserPattern` in `CourseDao.kt`
- Added `userIdPattern` function in `CoursesRepositoryImpl.kt` (copied from `ResourcesRepositoryImpl.kt`)
- Switched `getMyCourses` and `getMyCoursesFlow` in `CoursesRepositoryImpl.kt` to use the new native SQL Dao queries.
- Updated unit tests `CoursesRepositoryImplTest.kt` to use mocked behaviour matching new queries.

---
*PR created automatically by Jules for task [16612805223780866624](https://jules.google.com/task/16612805223780866624) started by @dogi*

dogi:
- Added `getForUserPattern` and `observeForUserPattern` in `CourseDao.kt`
- Added `userIdPattern` function in `CoursesRepositoryImpl.kt` (copied from `ResourcesRepositoryImpl.kt`)
- Switched `getMyCourses` and `getMyCoursesFlow` in `CoursesRepositoryImpl.kt` to use the new native SQL Dao queries.
- Updated unit tests `CoursesRepositoryImplTest.kt` to use mocked behaviour matching new queries.

---
*PR created automatically by Jules for task [16612805223780866624](https://jules.google.com/task/16612805223780866624) started by @dogi*


Okuro3499 (dismissed):
```
app/src/test/java/org/ole/planet/myplanet/repository/CoursesRepositoryImplTest.kt: the two updated tests (getMyCourses by userId fetches all and filters, getMyCoursesFlow       
    suppresses redundant emissions) mock away the exact logic this PR changes. Add a CourseDaoTest (in-memory Room.inMemoryDatabaseBuilder, following NewsDaoTest/UserDaoTest        
    convention) that inserts real MyCourse rows and asserts getForUserPattern/observeForUserPattern against actual JSON-serialized userId values, including a case with _ or % in the
    userId, to lock in the escaping behavior. 
```

dogi:
**Review — rating: 87/100**

Pushes shelf membership filtering into SQL instead of loading the whole `courses` table and filtering in Kotlin. I worked through the equivalence rather than taking it on faith:

- **`LIKE '%"id"%'` matches `List.contains` semantics.** `userId` is a `List<String>` persisted as JSON by `Converters`, so the quotes act as delimiters and prevent the prefix collision you'd otherwise get: `%"user1"%` does **not** match `["user10"]`, because after `user1` comes `0`, not the closing quote. That's the failure mode I went looking for and it's handled.
- **The escaping is right, and in the right order.** `\` → `%` → `_`, so the escape character is doubled before the later replacements introduce more of them. Escaping `_` matters here specifically because CouchDB ids routinely contain underscores, and an unescaped `_` is a single-char wildcard.
- **NULL rows behave identically.** `NULL LIKE …` is NULL, so those rows drop out, matching the old `it.userId?.contains(userId) == true`.
- **Nothing was lost in replacing the two-arg overload.** `getMyCourses(userId, courses)` is just `courses.filter { it.userId?.contains(userId) == true }` — a pure userId filter with no sorting, dedupe or title filtering hiding in it. It also stays on the interface for its other callers, so it isn't stranded.
- Room verifies `@Query` column names at compile time through KSP, so the green build is real evidence `userId` resolves.

The win is also bigger than "optimize queries" suggests: `mapCourses` fans out to `courseStepDao.getByCourseIds(...)`, so the old path fetched course steps for **every** course in the table before discarding most of them. Now only the user's courses are mapped. The inline comment explaining the JSON-list rationale is good practice — keep doing that.

**Nit — the tests don't cover the risky part.** Both tests mock the DAO, so neither the `LIKE` semantics nor `userIdPattern` is ever exercised; the mock reconstructs the id with `removeSurrounding("%\"", "\"%")`, which would itself break for any id containing `_`, `%`, or `\`. That's a tell that the escaping path is untested. `room-testing` is already a declared dependency, so an in-memory Room test over a couple of JSON `userId` rows (including `user_1` vs `user1`) would pin the behavior this PR actually depends on. A plain unit test of `userIdPattern` would be the cheap version.

**Nit:** if a userId ever contained `"` or `\`, Gson would store it JSON-escaped and the raw pattern wouldn't match. Not reachable with CouchDB ids — mentioning only so it's a known boundary.

**Verdict:** mergeable.

---

ragilzakaria (dismissed):
**Rating:** 88/100

Great optimization pushing the user shelf filter directly into Room SQLite query execution (`WHERE userId LIKE :userPattern ESCAPE '\'`).
- Prevents loading and mapping steps for all courses in the database when querying a user's enrolled courses.
- Escaping order (`\\` -> `\\%` -> `\\_`) is correct and handles CouchDB underscore IDs properly.
- The added in-memory Room test in `CourseDaoTest.kt` thoroughly validates escaping edge cases (wildcards, underscores, and percentages).

**Required Fix:**
1. **CI Duplication Blocker (`qlty`)**:
   - `qltysh` flagged a blocking duplicate code issue because `userIdPattern` in `CoursesRepositoryImpl.kt` is identical to `userIdPattern` in `ResourcesRepositoryImpl.kt`.
   - Please extract `userIdPattern` to a shared Room/DB utility (e.g. `RoomUtils` or `DbUtils` in `data/room/`) so both repositories share the same implementation and `qlty` passes.

2. **House Style Title**:
   - Update title to: `courses: smoother shelf queries optimizing (fixes #<issue_number>)`.

**Verdict:** Request changes — implementation is solid, just needs the shared `userIdPattern` helper extracted to clear the blocking CI check.

dogi:
**Re-review — Rating: 92/100** (was 87)

The one gap I flagged — that the escaping path was the risky part and nothing exercised it — is now covered by a real in-memory Room test rather than a mock, and it covers exactly the right cases: `user1_test` (underscore as a single-char wildcard), `user1%test` (percent as a multi-char wildcard), and `user1Xtest`. That last row is what pins the prefix-collision behavior: `assertEquals(1, resultUser1.size)` over six rows proves `%"user1"%` matches *only* `["user1", "user2"]` and not the three ids that merely start with `user1`. That's the failure mode this whole approach hinges on, and it's now enforced by the suite instead of by argument.

The flow path also moved into SQL this round (`observeForUserPattern`), which closes the asymmetry from the previous version where `getMyCourses` filtered in the DAO but `getMyCoursesFlow` still loaded the whole `courses` table and filtered in Kotlin. Same `LIKE … ESCAPE '\'` query, so it inherits the semantics the new test pins, and the `distinctUntilChanged` comparator on `(id, courseRev, userId)` is untouched — worth noting because that comparator is what makes the flow usable at all, given `MyCourse` is an `open class` with identity equality.

The win remains bigger than the title suggests: `mapCourses` fans out to `courseStepDao.getByCourseIds(...)`, so both paths previously fetched course steps for every course in the table before discarding most of them.

**Nit:** `CoursesRepositoryImplTest` still fakes the DAO with `removeSurrounding("%\"", "\"%")`, which would itself mis-handle an id containing `_`, `%`, or `\`. It's no longer a coverage hole now that `CourseDaoTest` pins the real semantics, but the fake is a small trap for whoever edits it next — a comment pointing at `CourseDaoTest` would help.

**Nit:** `@Config(sdk = [32])` pins this test to one SDK level; harmless, just worth being consistent with the other Robolectric tests in the suite.

**Verdict:** mergeable.

---

dogi:
**Re-review — Rating: 93/100** (was 92)

Nothing regressed, and the escaping story is now airtight end to end. `userIdPattern` escapes `\`, `%`, and `_` before wrapping in `%"…"%`, both DAO queries declare `ESCAPE '\'`, and `CourseDaoTest` proves the part that matters: `user1` matches only the course whose JSON list contains `"user1"`, while `user1_test`, `user1%test`, and `user1Xtest` each resolve to exactly their own row. Without the escape those last three would have been collateral matches for `user1`, so the test is aimed at precisely the failure mode the change introduces.

Both call sites now push the filter into SQL — `getMyCourses` no longer materializes and maps every course row to discard most of them, and `getMyCoursesFlow` narrows at the query rather than in a `map`, so Room only wakes the flow for rows that can matter.

**Nit:** the pattern is built from the raw `userId` while the column holds Gson-serialized JSON, so a userId containing `"` or `\` would be escaped in storage (`\"`) and wouldn't match the pattern. Not reachable with `org.couchdb.user:<name>` ids, and out of scope here — worth knowing only if user ids ever widen.

**Nit:** `userIdPattern` is duplicated verbatim in `CourseDaoTest`. Reasonable for a DAO test that shouldn't depend on the repository, but if the escaping rule ever changes, the test will keep passing against its own copy. `ResourcesRepositoryImpl` has the same helper shape (`userIdPattern`) — a shared internal util would let both production sites and the test agree by construction.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #15847, opened 2026-08-19
all: smoother map tile utils handling (fixes #15988)

### LINKED ISSUE #15988: Fix MapTileUtils copyAssets behavior
Tracking issue for PR #15847.

Fix MapTileUtils copyAssets behavior

- Check if the map file already exists and is not empty before copying.
- Ensure the destination directory (`osmdroid/`) is created (`mkdirs()`) before writing to prevent `FileNotFoundException`.
- Move `try-catch` inside the file loop so missing assets don't interrupt the rest of the copying process.
- Adapt `MapTileUtilsTest.kt` to the new logic.

---
*PR created automatically by Jules for task [3798495891012307559](https://jules.google.com/task/3798495891012307559) started by @dogi*

dogi:
Fix MapTileUtils copyAssets behavior

- Check if the map file already exists and is not empty before copying.
- Ensure the destination directory (`osmdroid/`) is created (`mkdirs()`) before writing to prevent `FileNotFoundException`.
- Move `try-catch` inside the file loop so missing assets don't interrupt the rest of the copying process.
- Adapt `MapTileUtilsTest.kt` to the new logic.

---
*PR created automatically by Jules for task [3798495891012307559](https://jules.google.com/task/3798495891012307559) started by @dogi*


Okuro3499 (approved):

Okuro3499 (approved):

dogi:
**Review — rating: 82/100**

Three genuine fixes, and the test handling is exemplary — the old tests had the bug baked into their *names* (`copyAssets_failsSilentlyWhenDirectoryDoesNotExist`, `copyAssets_skipsRemainingFilesOnPartialFailure`) and are rewritten to assert the corrected behavior rather than left to rot. That's exactly right for tests that document a defect.

On the changes themselves:

- **try/catch moved inside the loop** — correct. One unreadable asset no longer silently skips every remaining tile.
- **`outFile.parentFile?.mkdirs()`** — this is the substantive fix. Without it, a missing `/osmdroid` directory made `FileOutputStream` throw immediately and, with the old catch outside the loop, nothing was ever copied.
- **skip when `exists() && length() > 0`** — right intent, but see below.

**Nit — the skip can pin a broken file permanently.** The copy writes straight to the destination with no temp-and-rename, so an interrupted copy leaves a truncated but non-empty file. The old code overwrote on every launch — wasteful, but self-healing. With the new guard, that truncated file satisfies `length() > 0` and is kept forever. Comparing against the asset's real size (`assetManager.openFd(s).length`) or writing to a temp file and renaming on success would keep the optimization without the trap.

**Worth knowing before assuming this makes offline maps work.** The destination is `Environment.getExternalStorageDirectory()/osmdroid`, and the manifest declares no `WRITE_EXTERNAL_STORAGE` or `MANAGE_EXTERNAL_STORAGE`, so on API 29+ that write fails on permissions — now caught per-tile and printed. Separately, `OfflineMapsActivity` just calls `Configuration.getInstance().load(...)`, so osmdroid uses its default base path (app-scoped `getExternalFilesDir`), which is not the directory being written here. Both problems predate this PR and it's still a strict improvement over what was there — I'm flagging it only so the fix isn't mistaken for "offline tiles now load."

**Verdict:** mergeable.

---

PR #15848, opened 2026-08-19
teams: smoother surveys item callback diffing (fixes #15989)

### LINKED ISSUE #15989: Remove the codebase's only two notifyDataSetChanged() calls (surveys)
Tracking issue for PR #15848.

This commit removes the remaining `notifyDataSetChanged()` calls from `SurveyFragment.kt` and `SurveysAdapter.kt`. It resolves the issue where separate maps containing survey info and form state were not visible to the adapter's `DiffUtil`, resulting in `notifyDataSetChanged()` being forced to execute whenever the values came in. 

The fix involves introducing a new data class, `SurveyRow`, encapsulating a `StepExam`, its corresponding `SurveyInfo`, and its `SurveyFormState`. `SurveysViewModel` now combines the flows of surveys, infos, and form states into a single `StateFlow<List<SurveyRow>>`.

`SurveyFragment` no longer holds maps manually and now binds to the single list structure via the default list differential logic provided by `DiffUtil` running seamlessly via `submitList`. Tests in `SurveysViewModelTest` were updated to reflect the `SurveyRow` encapsulation logic, explicitly verifying item equivalence via `exam.id`.

---
*PR created automatically by Jules for task [8769693516190306679](https://jules.google.com/task/8769693516190306679) started by @dogi*

dogi:
This commit removes the remaining `notifyDataSetChanged()` calls from `SurveyFragment.kt` and `SurveysAdapter.kt`. It resolves the issue where separate maps containing survey info and form state were not visible to the adapter's `DiffUtil`, resulting in `notifyDataSetChanged()` being forced to execute whenever the values came in. 

The fix involves introducing a new data class, `SurveyRow`, encapsulating a `StepExam`, its corresponding `SurveyInfo`, and its `SurveyFormState`. `SurveysViewModel` now combines the flows of surveys, infos, and form states into a single `StateFlow<List<SurveyRow>>`.

`SurveyFragment` no longer holds maps manually and now binds to the single list structure via the default list differential logic provided by `DiffUtil` running seamlessly via `submitList`. Tests in `SurveysViewModelTest` were updated to reflect the `SurveyRow` encapsulation logic, explicitly verifying item equivalence via `exam.id`.

---
*PR created automatically by Jules for task [8769693516190306679](https://jules.google.com/task/8769693516190306679) started by @dogi*

## Summary by CodeRabbit

* **New Features**
  * Improved survey data handling so survey details and form progress remain synchronized when displayed.
  * Updated the surveys screen to present each survey with its associated information and state.

* **Bug Fixes**
  * Improved consistency when sorting, searching, adopting, and displaying survey status.

* **Chores**
  * Updated the app version to 0.65.67.


Okuro3499 (approved):

dogi:
**Review — rating: 88/100**

Good structural fix, not just a `notifyDataSetChanged()` removal. The old design had the fragment owning two mutable maps that the adapter read through, with the fragment poking the adapter whenever they changed — shared mutable state across a layer boundary. Combining the three flows into `List<SurveyRow>` in the ViewModel makes the adapter a pure function of its list, which is what `ListAdapter` wants.

Things I checked rather than assumed:

- **The DiffUtil comparison actually works.** `areContentsTheSame` now leans on `oldItem.surveyInfo == newItem.surveyInfo && oldItem.formState == newItem.formState`, which is only meaningful if those types have structural equality — both `SurveyInfo` and `SurveyFormState` are `data class`, so they do. Had either been a regular class this would have degenerated into reference comparison and rebound everything anyway, defeating the point.
- **`stateIn(WhileSubscribed(5000))` is safe here.** It makes `surveys.value` empty without an active subscriber, so I looked for production code reading `.value` directly — there is none; `SurveyFragment` collects it via `collectWhenStarted`. The tests needing an explicit collector to populate `.value` is the expected consequence, not a smell.
- `sendSurvey(current.exam)` correctly unwraps, and putting `SurveyRow` in `model/` matches where `SurveyInfo`/`SurveyFormState` already live.

**Nit — use `backgroundScope` in the tests.** `CoroutineScope(testDispatcher).launch { … }` plus a manual `collectJob.cancel()` at the end is fragile: if an assertion fails before that last line, the cancel never runs and `runTest` can fail on leftover coroutines, masking the real assertion failure. `runTest`'s `backgroundScope` exists for exactly this and is cancelled automatically, letting you drop the four `collectJob.cancel()` lines.

**Nit:** `surveyInfos` and `bindingData` are now unused outside the ViewModel — `combine` reads the private `_` backing fields — so both public `StateFlow`s can go private.

**Adjacent, pre-existing:** `updateUIState()` reads `getAdapter().itemCount` right after `submitList`, but `ListAdapter` diffs asynchronously, so the count can be one generation stale when `showNoData` runs. Unchanged by this PR — noting it since it sits in the collector you touched.

**Verdict:** mergeable.

---

PR #15849, opened 2026-08-19
all: smoother notifications repository dao querying (fixes #15990)

### LINKED ISSUE #15990: Batch markNotificationsSynced in a single transaction
Tracking issue for PR #15849.

When synchronizing notifications, `markNotificationsSynced` iterates over a list of `Pair<String, String?>` and calls `notificationDao.markSynced(id, rev)` for each item. Since Room opens a separate transaction and fsyncs for each write statement, this causes N transactions for N items.

This change optimizes the process by creating a new `NotificationDao.markSynced(syncResults: List<Pair<String, String?>>)` default method annotated with `@Transaction`. The repository is updated to use this new batched method, ensuring that all `markSynced` updates are performed in a single database transaction, matching the approach used in `HealthExaminationDao`. Test suites have been updated appropriately to verify this new method.

---
*PR created automatically by Jules for task [10838243413962628543](https://jules.google.com/task/10838243413962628543) started by @dogi*

dogi:
When synchronizing notifications, `markNotificationsSynced` iterates over a list of `Pair<String, String?>` and calls `notificationDao.markSynced(id, rev)` for each item. Since Room opens a separate transaction and fsyncs for each write statement, this causes N transactions for N items.

This change optimizes the process by creating a new `NotificationDao.markSynced(syncResults: List<Pair<String, String?>>)` default method annotated with `@Transaction`. The repository is updated to use this new batched method, ensuring that all `markSynced` updates are performed in a single database transaction, matching the approach used in `HealthExaminationDao`. Test suites have been updated appropriately to verify this new method.

---
*PR created automatically by Jules for task [10838243413962628543](https://jules.google.com/task/10838243413962628543) started by @dogi*


Okuro3499 (approved):

dogi:
**Review — rating: 88/100**

Correct, and the benefit is more than the perf the title claims. Previously each `markSynced` was its own implicit Room transaction, so a process death partway through the loop left some notifications flagged synced and the rest not — the next sync would re-upload an arbitrary subset. Wrapping the loop in `@Transaction` makes it all-or-nothing, so this is an atomicity fix as well as N fsyncs collapsing into one.

Details I checked:

- **`@Transaction` on a default interface method is supported** — Room generates the transaction wrapper around the body rather than requiring a `@Query`, and KSP would have failed the build otherwise, so the green run is real evidence here.
- **The overload is unambiguous** — `markSynced(String, String?)` vs `markSynced(List<Pair<String, String?>>)` differ in arity, so no resolution hazard.
- **Empty input is fine** either way: the repository still early-returns, and the DAO method would simply open and close an empty transaction.
- A single `UPDATE … WHERE id IN (…)` isn't an option here since `rev` differs per row, so loop-inside-transaction is the right shape rather than a `CASE WHEN` contortion.
- The updated tests verify the batch call and, in the empty case, assert `exactly = 0` against *both* overloads — appropriately thorough.

**Nit:** `Pair<String, String?>` is easy to transpose at a call site when both components are strings, and the destructuring `(id, rev)` is purely positional. A tiny `data class NotificationSyncResult(val id: String, val rev: String?)` would make both the DAO signature and `markNotificationsSynced` self-documenting. Not worth blocking over, but it's the kind of type that eventually gets misordered.

The `@Transaction` behavior itself can't be asserted against a mocked DAO — that's inherent and fine, it's Room's contract to keep.

**Verdict:** mergeable.

---

PR #15851, opened 2026-08-19
perf: bound Glide decode size for list thumbnails

dogi:
perf: bound Glide decode size for list thumbnails

Added `.override(targetPx, targetPx)` to Glide image loads in ResourcesAdapter, CoursesAdapter, and InlineResourceAdapter. This bounds the decode size to the actual layout size needed for the rows/grids, significantly improving memory usage by preventing full-resolution image decoding in scrolling lists. targetPx is dynamically computed via dp-to-px conversions matching the XML layout sizes.

---
*PR created automatically by Jules for task [3976867088705420181](https://jules.google.com/task/3976867088705420181) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/ui/courses/CoursesAdapter.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/courses/CoursesAdapter.kt
@@ -289,9 +289,12 @@ class CoursesAdapter(
         }
         ivSubjectIcon.visibility = View.GONE
         ivCover.visibility = View.VISIBLE
+
+        val targetPx = ((if (viewMode == ListViewMode.GRID) 84 else 44) * context.resources.displayMetrics.density).toInt()
         Glide.with(context)
             .load(model)
             .diskCacheStrategy(DiskCacheStrategy.ALL)
+            .override(targetPx, targetPx)
             .signature(ObjectKey(course.courseRev.orEmpty()))
             .centerCrop()
             .error(R.drawable.ole_logo)
--- a/app/src/main/java/org/ole/planet/myplanet/ui/courses/CoursesAdapter.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/courses/CoursesAdapter.kt
@@ -289,12 +289,11 @@ class CoursesAdapter(
         }
         ivSubjectIcon.visibility = View.GONE
         ivCover.visibility = View.VISIBLE
-
-        val targetPx = ((if (viewMode == ListViewMode.GRID) 84 else 44) * context.resources.displayMetrics.density).toInt()
+        // Glide automatically uses SizeDeterminer to bound decode size based on the layout dimensions.
+        // No manual .override() is needed for these views.
         Glide.with(context)
             .load(model)
             .diskCacheStrategy(DiskCacheStrategy.ALL)
-            .override(targetPx, targetPx)
             .signature(ObjectKey(course.courseRev.orEmpty()))
             .centerCrop()
             .error(R.drawable.ole_logo)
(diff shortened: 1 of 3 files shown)


Okuro3499 (dismissed):

dogi:
**Review — rating: 55/100**

Bounding decode size is a real concern in general, but Glide already does it here, and the specific dimensions chosen introduce a visible quality regression. I pulled the layouts for all three targets:

| View | Layout size |
|---|---|
| `iv_resource_preview` (`item_inline_resource`) | `match_parent` × `180dp` |
| `iv_video_thumbnail` (`item_inline_resource`) | `match_parent` × `match_parent` |
| `iv_cover` (`item_course_list` / `item_course_grid`) | `match_parent` × `match_parent` |

Two problems follow:

**1. `.override()` isn't buying what the title claims.** `.into(imageView)` already makes Glide wait for layout and decode at the view's *measured* size — that's what `SizeDeterminer` is for. `.override()` is needed when a view has no determinate size (typically `wrap_content` in an unmeasured parent). None of these do, so the decode was already bounded to the view before this change.

**2. Forcing a square shrinks and distorts non-square targets.** `iv_resource_preview` is full-width × 180dp, and `.override(180dp, 180dp)` decodes a square, which the `centerCrop` ImageView then stretches across the whole screen width — roughly a 2× horizontal upscale on a 360dp-wide phone, i.e. visibly softer than today. `iv_cover` is `match_parent` in both layouts, so its real size comes from the parent cover container, not from the hardcoded 84dp/44dp; wherever the container is larger than the guess, the thumbnail gets upscaled. `centerCrop` also crops to the *override* aspect ratio (1:1) rather than the view's, so the visible framing can change too.

**3. The hardcoded dp values duplicate layout dimensions** and will drift silently the next time a layout is touched.

Worth saying: the `ResourcesAdapter` change is the good one — it derives `targetPx` from the caller-supplied `coverWidthDp`, matching the pattern the existing `showPdfPreview` path already uses. Its only flaw is passing that width as both dimensions instead of letting the height follow the view.

If there's a real decode problem you've measured on a specific screen, the fix is to size from the actual view (or a `dimen` resource shared with the layout) and pass width and height separately — never a square for a non-square target. Otherwise the honest move is to drop `.override()` and rely on Glide's view sizing, since that's already correct here.

**Verdict:** needs another round — as written this ships blurrier thumbnails across courses, resources, and inline course resources for a saving Glide already provides.

---

dogi:
**Re-review — Rating: 84/100** (was 55)

You took the honest option: every `.override()` is gone, and what remains is a two-line comment recording *why* it isn't needed.

```kotlin
// Glide automatically uses SizeDeterminer to bound decode size based on the layout dimensions.
// No manual .override() is needed for these views.
```

That's the right resolution. The decode was already bounded to the measured view by `.into(imageView)`, and the hardcoded squares were shrinking and re-cropping non-square targets — `iv_resource_preview` is full-width × 180dp, `iv_cover` is `match_parent` in both course layouts. Removing the change removes the quality regression, and leaving the reasoning behind stops the next person from re-introducing it after reading "unbounded Glide decode" somewhere. A comment-only diff is a perfectly good outcome for a PR whose premise turned out not to hold.

**Nit:** the comment lives above one Glide call in `CoursesAdapter.showCoverImage`, but the same reasoning applies to `iv_resource_preview` / `iv_video_thumbnail` in the inline-resource adapter and to `ResourcesAdapter` — those were the other two targets in the original change. A line in each, or one comment worded as a general note where cover loading is centralized, would cover the places someone might actually revisit.

**Nit:** "these views" reads oddly above a single call site — "this view" or naming the ImageView would be clearer.

**Verdict:** mergeable.

---

Okuro3499:
this pr only adds comments

PR #15852, opened 2026-08-19
teams: smoother requests dao view modelling (fixes #15992)

### LINKED ISSUE #15992: perf(RequestsViewModel): use getJoinedMemberCount instead of full list
Tracking issue for PR #15852.

Optimize `RequestsViewModel` by using the existing `getJoinedMemberCount` query.

Previously, `RequestsViewModel` was loading the entire list of members and taking the `.size` to find the member count. This caused significant overhead, as it ran individual user queries per member. By switching to the `COUNT` query exposed by `TeamsRepository.getJoinedMemberCount`, we save memory and compute.

Also updated unit tests to use `getJoinedMemberCount`.

---
*PR created automatically by Jules for task [9437900723954636701](https://jules.google.com/task/9437900723954636701) started by @dogi*

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected team member counts to include only distinct, valid users.
  - Excluded deleted, missing, or invalid user records from counts.
  - Improved member request views to display accurate joined-member totals.

- **Tests**
  - Updated member and request scenarios to verify the corrected counting behavior.

dogi:
Optimize `RequestsViewModel` by using the existing `getJoinedMemberCount` query.

Previously, `RequestsViewModel` was loading the entire list of members and taking the `.size` to find the member count. This caused significant overhead, as it ran individual user queries per member. By switching to the `COUNT` query exposed by `TeamsRepository.getJoinedMemberCount`, we save memory and compute.

Also updated unit tests to use `getJoinedMemberCount`.

---
*PR created automatically by Jules for task [9437900723954636701](https://jules.google.com/task/9437900723954636701) started by @dogi*

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected team member counts to include only distinct, valid users.
  - Excluded deleted, missing, or invalid user records from counts.
  - Improved member request views to display accurate joined-member totals.

- **Tests**
  - Updated member and request scenarios to verify the corrected counting behavior.


Okuro3499 (dismissed):

dogi:
**Review — rating: 68/100**

The motivation is right — counting members shouldn't require N per-user lookups — but the two methods aren't equivalent, so this changes a user-visible number.

`getJoinedMembers(teamId)` applies four narrowing steps before you get a size:

```kotlin
teamDao.getByTeamIdAndDocType(teamId, "membership")
    .filter { !it.isDeletePending }      // excludes pending removals
    .mapNotNull { it.userId }            // drops null userIds
    .distinct()                          // dedupes repeat memberships
    .mapNotNull { userRepository.getUserById(it) }   // drops ids with no user row
```

`getJoinedMemberCount(teamId)` is `SELECT COUNT(*) FROM teams WHERE teamId = :teamId AND docType = :docType` — none of those filters. So the new count is `>=` the old one, and over-counts on delete-pending memberships, null `userId` rows, duplicate membership docs, and memberships whose user document hasn't synced yet. That last case is routine in this app: `TransactionSyncManager` pulls tables independently, so a device mid-sync can easily hold membership rows pointing at users it doesn't have, and the Requests screen would then show a member count higher than the members actually listed.

To be fair to the PR: `getJoinedMemberCount` is already used in three places in `TeamDetailFragment`, so this propagates existing behavior rather than inventing it. That makes the good fix a small one in the DAO instead of a revert — fix the query and all four call sites become correct at once:

```sql
SELECT COUNT(DISTINCT userId) FROM teams
WHERE teamId = :teamId AND docType = :docType
  AND isDeletePending = 0 AND userId IS NOT NULL
```

That covers delete-pending, nulls and duplicates. Exact parity would also need the users-table check (`AND userId IN (SELECT id FROM users)` or a join) — whether an unsynced member should count is a product call, but it's worth making deliberately rather than by omission.

**Nit:** the test updates keep the suite green but can't catch this — both methods are mocked, so the stubs simply assert whatever the author picked (`1`, `0`). Nothing validates that the count matches the list.

**Verdict:** needs another round — land the DAO query fix and this swap becomes a clean win everywhere `getJoinedMemberCount` is used.

---

dogi:
**Re-review — Rating: 76/100** (was 68)

Fixing the DAO query rather than reverting the swap was the right call, and three of the four divergences I listed are now closed: `COUNT(DISTINCT userId)` handles duplicate membership docs, `isDeletePending = 0` excludes pending removals, and `userId IS NOT NULL` drops null rows. All three existing `TeamDetailFragment` call sites get the fix for free, which was the point.

The users-table check is the one that doesn't hold up, and it's worth one more pass because it now under-counts:

**`SELECT id FROM users` doesn't cover what `getUserById` covers.** `getJoinedMembers` resolves ids through `UserRepositoryImpl.getUserById` → `UserDao.getById`, whose query is:

```sql
SELECT * FROM users WHERE id = :id OR _id = :id LIMIT 1
```

The new subquery only matches the `id` column, so any user found via the `_id` branch is silently dropped from the count.

That branch is load-bearing here, because membership documents store `_id`, not `id`. In `TeamsRepositoryImpl` the membership rows are created with `userId = user._id` (lines 168 and 175, just above `docType = "membership"`), and `getJoinedMembersWithVisitInfo` reinforces it by matching team userIds against `"org.couchdb.user:${admin.name}"`.

For a fully synced user the two columns agree — `insertUsersFromSync` sets `this.id = id; this._id = id` from the CouchDB `_id`, so `IN (SELECT id …)` happens to work. But `applyJsonToUser` only assigns `id` when it is already blank:

```kotlin
if (id.isNullOrBlank()) {
    id = if (newId.isEmpty()) { UUID.randomUUID().toString() } else { newId }
}
_rev = …
_id = newId
```

So a user who was registered on the device before syncing keeps a UUID in `id` while `_id` becomes `org.couchdb.user:<name>` after upload (see also `user._id = id` in the upload paths). Their membership row holds the `org.couchdb.user:` form, `getUserById` still finds them via `OR _id = :id`, and the new `COUNT` does not. On a shared tablet where members are created locally and synced up — the normal myPlanet flow — that's a member count lower than the list beneath it.

Either of these restores parity:

```sql
AND userId IN (SELECT id FROM users UNION SELECT _id FROM users)
-- or
AND EXISTS (SELECT 1 FROM users u WHERE u.id = teams.userId OR u._id = teams.userId)
```

**Nit:** the tests still mock `getJoinedMemberCount`, so nothing exercises the SQL that all the risk lives in. #15846 in this same queue just added an in-memory `CourseDaoTest` for exactly this reason — a `TeamDao` test with a delete-pending row, a null-`userId` row, a duplicate membership, and one member whose `id` differs from `_id` would pin every clause you just wrote (and would have caught the one above).

**Verdict:** needs another round — one clause, and then the count matches the list everywhere.

---

dogi:

The `EXISTS` clause is in, and it closes the last divergence:

```sql
SELECT COUNT(DISTINCT userId) FROM teams
WHERE teamId = :teamId AND docType = :docType AND isDeletePending = 0 AND userId IS NOT NULL
  AND EXISTS (SELECT 1 FROM users u WHERE u.id = teams.userId OR u._id = teams.userId)
```

That now mirrors `UserDao.getById`'s `WHERE id = :id OR _id = :id` exactly, so the locally-registered-then-synced user — UUID in `id`, `org.couchdb.user:<name>` in `_id`, membership row holding the latter — is counted again. All four differences from `getJoinedMembers` are covered: duplicates by `COUNT(DISTINCT)`, pending removals by `isDeletePending = 0`, null rows explicitly, and unresolvable ids by the `EXISTS`. Room's compile-time verification accepts the correlated subquery, and both test shards are green.

**Nit — `users._id` is indexed but the `OR` may not use it.** `UserEntity` declares `indices = [Index("_id"), …]` and `id` is the primary key, so both columns are individually indexed; SQLite can sometimes satisfy `a = x OR b = x` with an OR-by-union optimization, but it isn't guaranteed. If it doesn't, the subquery scans `users` once per candidate membership row. At myPlanet's scale (tens of members, hundreds of users) that's still far cheaper than the `getJoinedMembers` list materialization this replaces — so it doesn't change the verdict, but if you want it airtight, `AND (userId IN (SELECT id FROM users) OR userId IN (SELECT _id FROM users))` gives the planner two index-friendly subqueries instead of one correlated scan.

**Nit (carried over):** the tests still stub `getJoinedMemberCount`, so all four clauses of that SQL remain unexercised. The count logic is now the interesting part of this PR and the mock is exactly the layer that hides it — a `TeamDao` test with a delete-pending row, a null `userId`, a duplicate membership, and one member whose `id` differs from `_id` would pin it. Not a blocker, but this is the third round on this query, and each round has been a clause that a test would have caught.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #15853, opened 2026-08-19
all: smoother user entity image url handling (fixes #15993)

### LINKED ISSUE #15993: perf(UserEntity): optimize `addImageUrl` by removing JSON string serialization round-trip
Tracking issue for PR #15853.

This PR optimizes the `UserEntity.addImageUrl` method by removing an unnecessary JSON serialization-and-parsing round-trip.

It reads `jsonDoc["_attachments"].asJsonObject` directly, replaces the for-with-break with `entrySet().firstOrNull()?.key`, and drops the now-unused `JsonParser` import. The attachment selection stays first-key-wins, preserving the original behavior while improving performance during sync where this runs once per user document.

---
*PR created automatically by Jules for task [9356263635295215608](https://jules.google.com/task/9356263635295215608) started by @dogi*

dogi:
This PR optimizes the `UserEntity.addImageUrl` method by removing an unnecessary JSON serialization-and-parsing round-trip.

It reads `jsonDoc["_attachments"].asJsonObject` directly, replaces the for-with-break with `entrySet().firstOrNull()?.key`, and drops the now-unused `JsonParser` import. The attachment selection stays first-key-wins, preserving the original behavior while improving performance during sync where this runs once per user document.

---
*PR created automatically by Jules for task [9356263635295215608](https://jules.google.com/task/9356263635295215608) started by @dogi*


Okuro3499 (approved):

dogi:
**Review — rating: 90/100**

Clean removal of a pointless round-trip: the old code serialized `_attachments` to a string with `.toString()` and immediately re-parsed it with `JsonParser.parseString`, producing an object equivalent to the one it already had.

I checked the equivalence rather than eyeballing it:

- **Empty `_attachments` behaves the same.** The old `for ((key1) in entries) { …; break }` never entered the body, leaving `userImage` untouched; `firstOrNull()?.key` returns null and the `if` is skipped. Same outcome.
- **"First" means the same entry.** Gson's `JsonObject` is backed by a `LinkedTreeMap`, so `entrySet()` iterates in insertion order, and the round-trip re-parsed into another insertion-ordered map — so the old and new code pick the same attachment.
- **Exception behavior is unchanged.** `.asJsonObject` on a non-object `_attachments` threw before too, since the old code called it before `.toString()`.
- The `for ((key1) in …)` destructuring bound `Map.Entry.key`, which is exactly what `.key` now reads.

This runs per user during sync (`UserRepositoryImpl:270`), and CouchDB attachment stubs carry digests and lengths, so skipping a serialize+parse of that object per user is a small but real win. Dropping the now-unused `JsonParser` import is a nice touch.

**Nit:** `key1` is a leftover name from the destructuring — as a plain local it'd read better as `attachmentName`. And `keySet().firstOrNull()` would avoid materializing entry objects at all:

```kotlin
jsonDoc["_attachments"].asJsonObject.keySet().firstOrNull()?.let {
    userImage = UrlUtils.getUserImageUrl(id, it)
}
```

**Nit:** `addImageUrl` has no test coverage anywhere. Behavior here is provably identical so nothing is needed for this PR, but it's a pure function over a `JsonObject` — the empty-`_attachments` and multiple-attachment cases would be three cheap assertions if you're touching it anyway.

**Verdict:** mergeable.

---

PR #15854, opened 2026-08-19
all: smoother view lifecycle owning (fixes #16204)

### LINKED ISSUE #16204: Refactor: Update fragment coroutine scopes to viewLifecycleOwner
Tracking issue for PR #15854.

Refactored `lifecycleScope.launch` to `viewLifecycleOwner.lifecycleScope.launch` in multiple Fragment implementations to avoid accessing bindings after the fragment view has been destroyed.

This affects:
- `BellDashboardFragment`
- `CourseStepFragment`
- `CoursesFragment`
- `EditAchievementFragment`
- `AchievementFragment`
- `ResourceViewerFragment`
- `TeamCalendarFragment`
- `ResourcesFragment`
- `ResourceDetailFragment`

`ExamTakingFragment` was excluded as it runs an intentional `NonCancellable` save after view destruction. Base classes were excluded as per specifications.

---
*PR created automatically by Jules for task [14843944453044900124](https://jules.google.com/task/14843944453044900124) started by @dogi*

dogi:
Refactored `lifecycleScope.launch` to `viewLifecycleOwner.lifecycleScope.launch` in multiple Fragment implementations to avoid accessing bindings after the fragment view has been destroyed.

This affects:
- `BellDashboardFragment`
- `CourseStepFragment`
- `CoursesFragment`
- `EditAchievementFragment`
- `AchievementFragment`
- `ResourceViewerFragment`
- `TeamCalendarFragment`
- `ResourcesFragment`
- `ResourceDetailFragment`

`ExamTakingFragment` was excluded as it runs an intentional `NonCancellable` save after view destruction. Base classes were excluded as per specifications.

---
*PR created automatically by Jules for task [14843944453044900124](https://jules.google.com/task/14843944453044900124) started by @dogi*


Okuro3499 (dismissed):
```
CoursesFragment.kt:543 — saveSearchActivity()'s write is launched via viewLifecycleOwner.lifecycleScope but invoked from onPause(), guaranteeing a race with view destruction.   
    Revert to plain lifecycleScope (it's a pure data write, no UI touch), or wrap the DAO call in NonCancellable as done in ExamTakingFragment.onDestroyView.                        
  - EditAchievementFragment.kt:164 — save handler's two sequential repository writes have no UI access inside the coroutine; revert to lifecycleScope (or NonCancellable) so a       
    rotation/back-out mid-save doesn't silently drop the update.                                                                                                                     
  - CourseStepFragment.kt:88 (launchSaveCourseProgress) — pure repository write; revert to lifecycleScope so course-progress persistence isn't racy against view teardown.           
  - CourseStepFragment.kt:267 (setMenuVisibility) — add a view != null (or _binding != null) guard before touching viewLifecycleOwner, since this override can theoretically fire    
    outside the view's lifetime; don't rely solely on isAdded/::step.isInitialized
```

dogi:
**Review — rating: 74/100**

Right direction, and most of the ten sites are fine — but applying one rule uniformly hits two cases where the view lifecycle is the wrong scope, and one where `viewLifecycleOwner` can be touched without a view.

I traced each site for whether the view is guaranteed alive at the point of access. Seven are safe because they fire from click listeners, dialog handlers, or selection actions: `BellDashboardFragment:394`, `ResourcesFragment:838`, `AchievementFragment:197`, `EditAchievementFragment:161`, `TeamCalendarFragment:317`, `CoursesFragment:540`. `CourseStepFragment:88` is also safe — `launchSaveCourseProgress()` is only reached from inside `viewLifecycleOwner.lifecycle.withResumed { }` (line 165) or from within the `setMenuVisibility` launch.

**The one real crash risk: `ResourceDetailFragment.onRatingChanged()`.** This is an external callback from the rating dialog, and `viewLifecycleOwner` is now evaluated *before* the launch while the existing `if (!isAdded) return@launch` guard sits *inside* it. If the callback lands after `onDestroyView`, you get `IllegalStateException: Can't access the Fragment View's LifecycleOwner when getView() is null` where the old code was a harmless no-op. The fact that someone already added an `isAdded` guard here says the lifecycle was known to be dicey. Guarding on the view instead fixes it:

```kotlin
val owner = view?.let { viewLifecycleOwner } ?: return
owner.lifecycleScope.launch { … }
```

Same pattern applies to `ResourcesFragment:132` and `CourseStepFragment:264` — both gate on `isAdded`, which stays true when the fragment is added but its view is gone, so it doesn't establish what `viewLifecycleOwner` requires.

**Two of these are writes, and view scope cancels them.** `openCallFragment` navigates with `replaceFragment(addToBackStack = true)`, so moving forward destroys the view while the fragment survives on the back stack — precisely the case where a view-scoped job dies but a fragment-scoped one wouldn't. That's correct for the UI refreshes here, but it now also cancels `progressRepository.saveCourseProgress` and `userRepository.updateAchievement`. The latter fires from the Save button right after a "saving" toast, so a user who taps save and immediately navigates can lose the write. Persistence should sit on something that outlives the view — `viewModelScope`, or the `@ApplicationScope` already injected elsewhere in this codebase.

**Dead code worth removing instead of modernizing:** `CourseStepFragment.setMenuVisibility` is a ViewPager1 hook. `CoursesPagerAdapter` is a ViewPager2 `FragmentStateAdapter`, which drives fragment state through `setMaxLifecycle` and doesn't call `setMenuVisibility` — so that override looks unreachable. Worth confirming and deleting rather than porting.

**Verdict:** needs another round — guard the `ResourceDetailFragment` callback on `view`, and leave the two persistence calls on a longer-lived scope.

---

ragilzakaria (approved):
**Rating:** 94/100

This is a very solid refactoring to ensure UI coroutines are bound to the fragment's view lifecycle rather than the fragment instance:

- **Lifecycle Safety**: Binds UI updates across `BellDashboardFragment`, `CourseStepFragment`, `ResourcesFragment`, `ResourceDetailFragment`, `TeamCalendarFragment`, `AchievementFragment`, and `ResourceViewerFragment` to `viewLifecycleOwner.lifecycleScope`, eliminating the risk of operating on destroyed views/bindings.
- **Safe View Null-Guarding**: Appropriately added `val owner = view?.let { viewLifecycleOwner } ?: return` guards in async/dialog callbacks (`ResourceDetailFragment.onRatingChanged`, `ResourcesFragment.refreshResourcesData`, `CourseStepFragment.setMenuVisibility`) where events can fire after `onDestroyView`.
- **Data Persistence Preserved**: Critical persistence operations (`updateAchievement`, `saveCourseProgress`) were intentionally kept on longer-lived scopes so in-flight writes aren't cancelled on immediate backward/forward navigation.

**Housekeeping Note:**
- Update PR title to match house style: `ui: smoother coroutine scopes updating (fixes #<issue_number>)`.

**Verdict:** Approved! Ready to merge once retitled in house style.

dogi:

Both problems from the last round are resolved, and resolved differently — which is the right instinct, since they were different problems.

**The three crash-prone sites now use the view guard.** `ResourceDetailFragment.onRatingChanged`, `ResourcesFragment.refreshResourcesData` and `CourseStepFragment.setMenuVisibility` all take `val owner = view?.let { viewLifecycleOwner } ?: return` before launching. That's the correct check: it gates on the same condition `viewLifecycleOwner` itself asserts (`getView() != null`), which the pre-existing `isAdded` guards did not — `isAdded` stays true between `onDestroyView` and `onDestroy`, exactly the window these callbacks can land in.

**The two persistence writes were left alone.** `CourseStepFragment.launchSaveCourseProgress` (`progressRepository.saveCourseProgress`) and `EditAchievementFragment` (`userRepository.updateAchievement`) are no longer in the diff, so they keep the fragment-scoped `lifecycleScope` that survives the `openCallFragment` back-stack navigation. That was the data-loss half of my last comment and it's gone.

I re-checked the sites that stayed converted and they hold up: `BellDashboardFragment:394`, `ResourcesFragment:839`, `TeamCalendarFragment:317` and `AchievementFragment:197` all fire from click or dialog handlers with the view alive. `AchievementFragment` in particular is safe for a reason worth stating — `resourcesRepository.downloadResources` only calls `DownloadUtils.openPriorityDownloadService(...)`, so the download runs in a foreground service and cancelling the coroutine can't interrupt it.

One note on the site that's new this round, `ResourceViewerFragment.onRecordStopped`: it's a write (`translationAudioPath`), so in principle view-scoping it means a cancelled coroutine leaves the recorded file on disk unlinked from the library item. In practice the same callback dereferences `binding` two lines later and `binding` is `_binding!!`, so this listener already can't survive `onDestroyView` — the conversion doesn't introduce a new failure mode, it just picks a different exception. Fine as-is; flagging so it's a known boundary.

**Nit (carried over):** `CourseStepFragment.setMenuVisibility` is a ViewPager1 hook and `CoursesPagerAdapter` is a ViewPager2 `FragmentStateAdapter`, which drives child state through `setMaxLifecycle` and never calls it. Still looks unreachable — worth confirming and deleting rather than keeping a guarded version alive.

**Verdict:** mergeable.

---

dogi:
**Re-review — Rating: 89/100** (was 88)

Still correct, and the guard placement is the part worth confirming: the three sites that can fire outside the view lifecycle get the `view?.let { viewLifecycleOwner } ?: return` guard — `CourseStepFragment.setMenuVisibility`, `ResourceDetailFragment.onRatingChanged`, and `ResourcesFragment.refreshResourcesData` — while the four that can only run with a live view (click listeners in `BellDashboardFragment`, `ResourcesFragment:839`, `TeamCalendarFragment`, `AchievementFragment`, and the recording callback in `ResourceViewerFragment`) use `viewLifecycleOwner.lifecycleScope` directly. That's the right line to draw: accessing `viewLifecycleOwner` before `onCreateView` or after `onDestroyView` throws, so the guard belongs exactly where the caller isn't a view event.

The change itself is the fix that matters — these coroutines touch view binding and adapters, so scoping them to the fragment's lifetime rather than the view's was a use-after-destroy waiting to happen on every one.

**Merge-order note:** #15925 rewrites the same `CourseStepFragment.setMenuVisibility` body, deleting the coroutine entirely in favour of a cached `userHasCourse` flag. If that lands after this, your guard becomes dead code (there's no `viewLifecycleOwner` access left to protect); if it lands first, this hunk conflicts. I've flagged it on that PR too — the two want coordinating rather than letting the merge decide.

**Nit:** `view?.let { viewLifecycleOwner } ?: return` discards the `let` argument, so it reads as though it's doing something with `view` when it's really just a null check. `if (view == null) return` followed by `viewLifecycleOwner.lifecycleScope.launch { … }` says the same thing plainly.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #15855, opened 2026-08-19
sync: smoother download service next url managing (fixes #15994)

### LINKED ISSUE #15994: Refactor `getNextUrl` for better performance
Tracking issue for PR #15855.

This refactor improves the performance of `getNextUrl` in `DownloadService.kt`.

Previously, `getNextUrl` sorted the entire queue, filtered it, mapped the results to `QueuedUrl` wrapper objects, and then extracted the minimum element from the list. This resulted in O(n^2 log n) complexity for large queues because it occurred repeatedly inside the file processing loop.

The refactored code instead filters the `String` set directly, computes the minimum element in O(n) via `.minOrNull()`, and then wraps just that single element in a `QueuedUrl` object.

---
*PR created automatically by Jules for task [11263882347371032895](https://jules.google.com/task/11263882347371032895) started by @dogi*

dogi:
This refactor improves the performance of `getNextUrl` in `DownloadService.kt`.

Previously, `getNextUrl` sorted the entire queue, filtered it, mapped the results to `QueuedUrl` wrapper objects, and then extracted the minimum element from the list. This resulted in O(n^2 log n) complexity for large queues because it occurred repeatedly inside the file processing loop.

The refactored code instead filters the `String` set directly, computes the minimum element in O(n) via `.minOrNull()`, and then wraps just that single element in a `QueuedUrl` object.

---
*PR created automatically by Jules for task [11263882347371032895](https://jules.google.com/task/11263882347371032895) started by @dogi*


Okuro3499 (approved):

dogi:
**Review — rating: 86/100**

Behavior-preserving and a clean win. The equivalence isn't obvious from the diff, so I chased it down:

`QueuedUrl` is `data class QueuedUrl(val url: String, val isPriority: Boolean, val priority: Int = 0)`, and the old `getNextUrl` built every element with the **two-argument** constructor — so `priority` was `0` for the entire queue. `getNextPriorityUrl`'s `maxByOrNull { it.priority }` therefore had nothing to discriminate on and returned the *first* element (Kotlin returns the first maximal element on ties) of a list that had just been `sorted()`. First-of-sorted is exactly `minOrNull()`, so the new code picks the same URL. The empty case also matches: `getNextPriorityUrl(emptyList())` returned null, and `minOrNull()` on an empty list gives null before `?.let` ever runs. Both filters are carried over unchanged.

So it drops an O(n log n) sort and n wrapper allocations for one O(n) pass and a single allocation, with no behavior change. Good.

**Nit — this strands dead code that should go in the same PR.** The companion `getNextPriorityUrl(downloadQueue: List<QueuedUrl>)` at line 556 had exactly one production caller: the `getNextUrl` line you just replaced. Nothing in `main` calls it now — only `DownloadServiceUrlSelectionTest`. (To be clear: the *instance* `getNextPriorityUrl()` at line 168 is still live from `processDownloadQueue`; it's only the List-taking companion overload that's orphaned.)

Worth removing together with it:

- **Its four tests**, which now assert against a function no production path reaches. They're the misleading part — `test getNextPriorityUrl returns item with highest priority` passes `priority` values of 5/10/3 that production never produces, so it reads like priority ranking is a live feature when nothing ever set that field.
- **`QueuedUrl.priority` itself**, which is never assigned anything but its default in production and, once the companion helper is gone, never read either.

That's ~4 lines of production code and one test block, and it makes the remaining code honest about what it does.

**Verdict:** mergeable.

---

PR #15856, opened 2026-08-19
courses: smoother progress view modelling (fixes #15995)

### LINKED ISSUE #15995: Refactor: Move Course Progress JSON mapping into ProgressViewModel
Tracking issue for PR #15856.

Refactors `CoursesProgressFragment` and `ProgressViewModel` to improve performance by offloading JSON mapping to a background thread.

- `ProgressViewModel` now exposes a `StateFlow<List<CoursesProgressRow>>` instead of a raw `JsonArray?`.
- JSON mapping is performed using `withContext(dispatcherProvider.default)` inside the ViewModel.
- The Gson `TypeToken` is initialized once as a `private val`.
- `CoursesProgressFragment` simply passes the emitted list to its adapter.
- Added comprehensive unit tests in `ProgressViewModelTest` to verify the mapping logic.

---
*PR created automatically by Jules for task [2672347155149216592](https://jules.google.com/task/2672347155149216592) started by @dogi*

dogi:
Refactors `CoursesProgressFragment` and `ProgressViewModel` to improve performance by offloading JSON mapping to a background thread.

- `ProgressViewModel` now exposes a `StateFlow<List<CoursesProgressRow>>` instead of a raw `JsonArray?`.
- JSON mapping is performed using `withContext(dispatcherProvider.default)` inside the ViewModel.
- The Gson `TypeToken` is initialized once as a `private val`.
- `CoursesProgressFragment` simply passes the emitted list to its adapter.
- Added comprehensive unit tests in `ProgressViewModelTest` to verify the mapping logic.

---
*PR created automatically by Jules for task [2672347155149216592](https://jules.google.com/task/2672347155149216592) started by @dogi*


Okuro3499 (approved):

dogi:
**Review — rating: 91/100**

Textbook version of this refactor, and it fixes two things beyond the stated reorganization.

- **Parsing was running on the main thread.** The mapping lived inside `collectWhenStarted`, so per-row `gson.fromJson` for `stepMistake` ran on the UI thread on every emission. Moving it into `withContext(dispatcherProvider.default)` is the actual win here, and `Default` is the right pool for CPU-bound work.
- **The Fragment no longer injects `Gson`.** A view has no business deserializing documents, and dropping `@Inject lateinit var gson` from `CoursesProgressFragment` is the right consequence.
- Injecting `DispatcherProvider` rather than reaching for `Dispatchers.Default` follows the convention in `CLAUDE.md`, so the parsing stays deterministic under test.
- Exposing `StateFlow<List<CoursesProgressRow>>` instead of `StateFlow<JsonArray?>` collapses the collector to a single `submitList(courseData)` and takes the nullability out of the UI layer.

Equivalence checks: the mapping body moved verbatim, and hoisting `TypeToken` into a field is a straight improvement (it was already outside the per-row `map`, so no behavior change). One small difference worth knowing is an improvement: previously `courseData?.let { … }` meant a null fetch skipped `submitList` entirely and left stale rows on screen; now it submits an empty list and clears them.

The tests are the right ones — a real `Gson()` rather than a mock, and the new `loadCourseData_parsesJsonCorrectly` covers the nested `progress` object and the `stepMistake` map, which is exactly the logic that moved.

**Nit:** `obj.get("courseId").asString` NPEs if the field is absent (same as before — not introduced here). Now that this logic sits in a unit-testable ViewModel, a malformed-row case would be cheap to pin, and it's the kind of thing an offline-first sync can hand you.

**Verdict:** mergeable.

---

PR #15857, opened 2026-08-19
all: smoother time utils date formatter caching (fixes #15996)

### LINKED ISSUE #15996: perf: memoize DateTimeFormatters in TimeUtils
Tracking issue for PR #15857.

`TimeUtils` dynamically instantiates `DateTimeFormatter`s via `DateTimeFormatter.ofPattern` at ten call sites on every invocation. This parsing is computationally expensive and runs multiple times per row in various list adapters (`EnterprisesReportsAdapter`, `EnterprisesFinancesAdapter`, `HealthExaminationAdapter`) as well as per day cell in the calendar decorator.

This PR optimizes these paths by hoisting the instantiation into a thread-safe, `ConcurrentHashMap`-backed memoization function `formatterFor(pattern, zone, locale)`. It correctly handles the "trap" condition by differentiating between formatters initialized with `Locale.getDefault()` and those initialized with no locale, ensuring zero behavioral changes.

---
*PR created automatically by Jules for task [16922335472144267320](https://jules.google.com/task/16922335472144267320) started by @dogi*

dogi:
`TimeUtils` dynamically instantiates `DateTimeFormatter`s via `DateTimeFormatter.ofPattern` at ten call sites on every invocation. This parsing is computationally expensive and runs multiple times per row in various list adapters (`EnterprisesReportsAdapter`, `EnterprisesFinancesAdapter`, `HealthExaminationAdapter`) as well as per day cell in the calendar decorator.

This PR optimizes these paths by hoisting the instantiation into a thread-safe, `ConcurrentHashMap`-backed memoization function `formatterFor(pattern, zone, locale)`. It correctly handles the "trap" condition by differentiating between formatters initialized with `Locale.getDefault()` and those initialized with no locale, ensuring zero behavioral changes.

---
*PR created automatically by Jules for task [16922335472144267320](https://jules.google.com/task/16922335472144267320) started by @dogi*


Okuro3499 (approved):

dogi:
**Review — rating: 89/100**

Careful work on the part that's easy to get wrong. `DateTimeFormatter.ofPattern` compiles the pattern into a printer/parser tree on every call, `DateTimeFormatter` is immutable and thread-safe, and these sit on a genuine hot path — `formatDate` is called per-bind from `VoicesAdapter`, `EventsAdapter`, `TeamsTasksAdapter`, `HealthUsersAdapter` and others. `EventsAdapter` had already hand-rolled its own `dateCache` for formatted strings, which is decent evidence the cost was noticed before.

Two things I specifically checked:

- **Locale changes still work.** This is the trap with memoizing formatters in an app shipping six languages. `defaultLocale` is `get() = Locale.getDefault()`, so it's re-evaluated per call, and since `locale` is part of `FormatterKey`, switching language yields a new key and a correctly localized formatter rather than a stale cached one.
- **All eight conversions preserve locale and zone.** The subtle part is that the original code mixed `ofPattern(p)` (no locale) with `ofPattern(p, defaultLocale)`, and the rewrite threads `locale = null` in exactly the four no-locale cases while letting the default apply in the rest. I walked each one — `formatterFor(pattern, utcZone)` for the `withZone(utcZone)` sites, `formatterFor(format ?: "", ZoneId.systemDefault())` for `formatDate`, and the bare `formatterFor("dd-MM-yyyy")` where `defaultLocale` was passed explicitly before. No drift. `TimeUtilsTest` already covers this surface, so the green run is meaningful here.

Exception behavior is also unchanged: if `ofPattern` throws, it now throws out of `formatterFor` but still inside each caller's existing `try`, and nothing gets cached.

**Nit:** prefer `formatters.computeIfAbsent(key) { … }` over `getOrPut`. Kotlin's `MutableMap.getOrPut` is get-then-put and not atomic, so two threads racing on first use can both build a formatter with one overwriting the other. Harmless here since the results are interchangeable, but `computeIfAbsent` is atomic and removes any ambiguity about which `getOrPut` overload resolves for a `ConcurrentHashMap`.

**Nit:** the cache has no eviction. It's bounded today — every pattern reaching `formatDate` is a string literal (`"yyyy-MM-dd"`, `"MMM dd, yyyy"`), so the key space is a handful of patterns × zone × locale. Worth remembering that a future caller passing a document-derived pattern would make it grow without limit.

**Verdict:** mergeable.

---

PR #15858, opened 2026-08-19
dashboard: smoother activities view modelling (fixes #16205)

### LINKED ISSUE #16205: Refactor ActivitiesFragment to use ActivitiesViewModel
Tracking issue for PR #15858.

Refactor ActivitiesFragment to use ActivitiesViewModel

- Created a new `ActivitiesViewModel` specifically for the `ActivitiesFragment` to cleanly handle user data resolution and offline login fetching.
- Removed domain injections (`UserSessionManager` and `ActivitiesRepository`) from `ActivitiesFragment`.
- Transitioned `ActivitiesFragment` to collect `viewModel.offlineLogins` safely.
- Kept UI specific logic (MPAndroidChart related configurations and logic) inside the fragment to maintain architectural boundaries.

---
*PR created automatically by Jules for task [4062383564071041160](https://jules.google.com/task/4062383564071041160) started by @dogi*

dogi:
Refactor ActivitiesFragment to use ActivitiesViewModel

- Created a new `ActivitiesViewModel` specifically for the `ActivitiesFragment` to cleanly handle user data resolution and offline login fetching.
- Removed domain injections (`UserSessionManager` and `ActivitiesRepository`) from `ActivitiesFragment`.
- Transitioned `ActivitiesFragment` to collect `viewModel.offlineLogins` safely.
- Kept UI specific logic (MPAndroidChart related configurations and logic) inside the fragment to maintain architectural boundaries.

---
*PR created automatically by Jules for task [4062383564071041160](https://jules.google.com/task/4062383564071041160) started by @dogi*


Okuro3499 (dismissed):
```
ActivitiesViewModel.kt: Add a unit test (e.g. ActivitiesViewModelTest.kt) covering offlineLogins — at minimum: emits repository data when a user is present, and emits nothing   
(or an empty list, per your final design choice) when getUserModel() returns null. This is the one gap that makes the refactor's goal ("move logic somewhere testable")          
incomplete.
```

dogi:
**Review — rating: 78/100**

The direction is right and it fixes a genuine anti-pattern: the old code nested `collectLatestWhenStarted` *inside* `viewLifecycleOwner.lifecycleScope.launch`, where the outer launch existed purely to await the suspending `getUserModel()`. Since `collectLatestWhenStarted` already launches on the view scope with `repeatOnLifecycle`, that was a redundant layer. Pushing the user lookup into the flow and leaving the fragment with a single collect is cleaner, and dropping the `ActivitiesRepository`/`UserSessionManager` injections out of the Fragment is the correct layering. Null-user behavior is preserved too — the old `return@launch` and the new `filterNotNull()` both mean nothing renders.

Three small things keep me from calling it done:

**1. No `stateIn`, so the ViewModel isn't actually buying you caching.** `offlineLogins` is a cold flow, and `collectLatestWhenStarted` re-collects on every STOPPED→STARTED transition — so each return to the foreground re-runs `getUserModel()` and re-subscribes to the Room flow. That matches the old behavior, so it's not a regression, but holding the data across brief backgrounding is the main reason to move a flow into a ViewModel:

```kotlin
val offlineLogins: StateFlow<List<OfflineActivity>> = …
    .stateIn(viewModelScope, SharingStarted.WhileSubscribed(5000), emptyList())
```

#15848 in this same batch does exactly that, so it's the pattern the codebase is converging on.

**2. `flatMapLatest` is ceremony here, and it's what forces the `@OptIn`.** The upstream flow emits exactly one value, so "cancel the previous inner flow" never happens. `emitAll` expresses it directly and drops the experimental opt-in entirely:

```kotlin
val offlineLogins: Flow<List<OfflineActivity>> = flow {
    val userName = userSessionManager.getUserModel()?.name ?: return@flow
    emitAll(activitiesRepository.getOfflineLogins(userName))
}
```

**3. No test for a brand-new ViewModel.** This is now trivially testable — stub `UserSessionManager` and `ActivitiesRepository`, assert the flow emits the logins and emits nothing when the user is null. Most ViewModels in this codebase have one.

**Nit:** `import javax.inject.Inject` sits at the bottom after the `org.ole.planet…` imports; the style guide ordering puts it before them.

**Heads-up on merge order:** #15933 (`computeMonthlyCounts`) and #15937 (`DateFormatSymbols().months`) also touch `ActivitiesFragment`. No functional overlap with this PR — you leave `computeMonthlyCounts` alone — but expect a textual conflict for whichever lands second.

**Verdict:** needs another round — all four items are a handful of lines.

---

ragilzakaria (approved):
**Rating:** 95/100

Clean and proper MVVM architectural refactoring for `ActivitiesFragment`:

- **Layering & Injections**: Completely extracts `UserSessionManager` and `ActivitiesRepository` from `ActivitiesFragment` into the newly created `ActivitiesViewModel`.
- **Reactive StateFlow Caching**: Uses `emitAll` inside the flow builder and converts it with `.stateIn(viewModelScope, SharingStarted.WhileSubscribed(5000), emptyList())`, which provides short-term caching across background/foreground transitions and avoids repeated database reads.
- **Clean UI Consumption**: Replaced the nested `viewLifecycleOwner.lifecycleScope.launch` in the fragment with standard `collectLatestWhenStarted(viewModel.offlineLogins)`.
- **Unit Tests**: Includes `ActivitiesViewModelTest` covering both `user == null` and valid user scenarios.

**Housekeeping Note:**
- Update PR title to match house style: `dashboard: smoother activities view model refactoring (fixes #<issue_number>)`.

**Verdict:** Approved! Clean, well-tested, and ready to merge.

dogi:
**Re-review — Rating: 90/100** (was 78)

All three items are addressed:

- **`stateIn(viewModelScope, WhileSubscribed(5000), emptyList())`** — this is what makes moving the flow into a ViewModel actually pay for itself. A brief backgrounding no longer re-runs `getUserModel()` and re-subscribes to the Room flow; only a gap longer than the 5s grace does.
- **`flatMapLatest` → `flow { … emitAll(…) }`** — same semantics for a single-value upstream, one less operator, and the `@OptIn(ExperimentalCoroutinesApi::class)` is gone from production code as a result.
- **Test added**, and it covers the branch that matters: null user emits only the `emptyList()` seed and nothing else, present user emits seed then repository data. Asserting `emissions.size` is what makes the null case meaningful — it proves the flow completes rather than emitting a stale value.

Import ordering in the new ViewModel is correct too (`javax` before `kotlinx`, `org.ole` last).

One behavior change worth naming, since it isn't a regression but it is different: `stateIn`'s `emptyList()` seed means the fragment now gets one emission *before* data, so `renderChart(computeMonthlyCounts(emptyList(), …))` draws an empty chart first — and for a null user it draws that empty chart instead of leaving the view untouched, where the old `return@launch` rendered nothing at all. For a bar chart that's arguably an improvement over a blank area, so I'd keep it; just be aware the "no user" state is now visually "zero activity" rather than "nothing".

**Heads-up on merge order (unchanged):** #15933 and #15937 both still sit in the queue touching `ActivitiesFragment`. No functional overlap — you leave `computeMonthlyCounts` alone — but expect a textual conflict for whichever lands second.

**Verdict:** mergeable.

---

dogi:
**Re-review — Rating: 91/100** (was 90)

Holds up. The nested-scope pattern is gone — `onViewCreated` had a `viewLifecycleOwner.lifecycleScope.launch { … collectLatestWhenStarted(…) }`, launching a coroutine only to start a lifecycle-aware collector inside it, and that's now one top-level `collectLatestWhenStarted(viewModel.offlineLogins)`.

The ViewModel does the two things that were awkward in the fragment, correctly:

```kotlin
val offlineLogins: StateFlow<List<OfflineActivity>> = flow {
    val userName = userSessionManager.getUserModel()?.name ?: return@flow
    emitAll(activitiesRepository.getOfflineLogins(userName))
}.stateIn(viewModelScope, SharingStarted.WhileSubscribed(5000), emptyList())
```

The suspending user lookup lives inside the flow builder rather than gating the collector, and `WhileSubscribed(5000)` means a rotation or tab switch doesn't re-run the lookup and re-subscribe Room. Both tests are well-aimed — the null-user case asserting exactly one emission, and the populated case documenting that `emissions[0]` is the `StateFlow` seed.

**One behavior change worth naming:** with a null user, `return@flow` completes the flow without emitting, so the `StateFlow` stays at `emptyList()` and the collector still runs — `renderChart` now draws an empty chart where before `return@launch` meant the chart was never rendered at all. Empty is arguably the better state (no stale view), but it is a change, so it belongs in the PR body.

**Merge-order note:** two other PRs in this queue touch `ActivitiesFragment.kt` — #15937 adds a `months` field right below the binding declarations and #15933 rewrites `computeMonthlyCounts`. This PR rewrites the import block and the field block (removing `lifecycleScope`, `Inject`, `launch`, `ActivitiesRepository`, `UserSessionManager`, adding `viewModels`), so it overlaps #15937's region. Git will likely handle it, but whoever merges second should expect to resolve the imports by hand rather than trusting a clean apply.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #15859, opened 2026-08-19
courses: smoother surveys repository exam dao querying (fixes #15997)

### LINKED ISSUE #15997: Optimize `getSurvey` fallback logic with scoped Room query
Tracking issue for PR #15859.

Replaced an inefficient in-memory fallback in `SurveysRepositoryImpl.getSurvey()` that materialized all surveys just to match one name. Added a highly specific Room SQL query `getByTypeAndName` in `ExamDao` with `LIMIT 1` for this purpose. 

Also added a new suite of unit tests in `ExamDaoTest.kt` to cover the new DAO method (match, no match, and filtering by type). Updated existing mocks in `SurveysRepositoryImplTest.kt` to reflect the new repository logic. No speculative database indices were added, in accordance with the issue request.

---
*PR created automatically by Jules for task [2793319731534727212](https://jules.google.com/task/2793319731534727212) started by @dogi*

dogi:
Replaced an inefficient in-memory fallback in `SurveysRepositoryImpl.getSurvey()` that materialized all surveys just to match one name. Added a highly specific Room SQL query `getByTypeAndName` in `ExamDao` with `LIMIT 1` for this purpose. 

Also added a new suite of unit tests in `ExamDaoTest.kt` to cover the new DAO method (match, no match, and filtering by type). Updated existing mocks in `SurveysRepositoryImplTest.kt` to reflect the new repository logic. No speculative database indices were added, in accordance with the issue request.

---
*PR created automatically by Jules for task [2793319731534727212](https://jules.google.com/task/2793319731534727212) started by @dogi*


Okuro3499 (approved):

dogi:
**Review — rating: 92/100**

Best-tested PR I've reviewed in this batch. The change itself is small and clearly right, and the test work is what a DAO change should look like.

The old fallback materialized **every** `type = "surveys"` row just to find one by name — and because `StepExam` goes through `@TypeConverters`, that meant running Gson deserialization per row to then discard nearly all of them. Replacing it with `WHERE type = :type AND name = :name LIMIT 1` is a real win, not a micro-optimization.

Correctness points I checked:

- **`=` rather than `IS` is right here.** `CLAUDE.md` calls for `IS` when a *null argument* should match NULL rows; this parameter is a non-null `String`, so `=` correctly never matches rows with a null `name` — which is exactly what the old `it.name == id` did. Easy rule to misapply in either direction, and it's applied correctly.
- Both old and new are equally non-deterministic if two surveys share a name: the old `getByType` had no `ORDER BY` either, so "first" was already unspecified. No regression — just noting that if duplicate survey names are reachable, an `ORDER BY` would make the pick defined.

**The tests are the highlight.** `ExamDaoTest` uses a real in-memory `AppDatabase` and exercises match / no-match / wrong-type, so it validates the actual SQL rather than a mock of it. That's the difference between a test that proves the query works and one that just restates the stub — and `room-testing` was already a declared dependency waiting to be used. The added repository test for the null path rounds it out.

**Nit — drop `@Config(sdk = [32])`.** #15914 ("all: less robolectric config sdk pins is more", already labelled `merge`) is removing exactly these hardcoded SDK pins across the suite. Adding a new one here means it gets deleted again shortly; leaving it off keeps this test aligned with where the suite is heading.

**Verdict:** mergeable.

---

PR #15860, opened 2026-08-19
resources: smoother feedback personals repositories flowing (fixes #15998)

### LINKED ISSUE #15998: Add distinctUntilChanged guards to unguarded Room flows
Tracking issue for PR #15860.

Appended `.distinctUntilChanged()` guards (with explicit comparators where Room constructs fresh instances) to the five Room flows in `PersonalsRepositoryImpl.kt`, `FeedbackRepositoryImpl.kt`, and `ResourcesRepositoryImpl.kt` to suppress duplicate flow emissions on database writes. Added unit tests verifying flow emission deduplication in each repository test class.

---
*PR created automatically by Jules for task [1639953032540210715](https://jules.google.com/task/1639953032540210715) started by @dogi*

dogi:
Appended `.distinctUntilChanged()` guards (with explicit comparators where Room constructs fresh instances) to the five Room flows in `PersonalsRepositoryImpl.kt`, `FeedbackRepositoryImpl.kt`, and `ResourcesRepositoryImpl.kt` to suppress duplicate flow emissions on database writes. Added unit tests verifying flow emission deduplication in each repository test class.

---
*PR created automatically by Jules for task [1639953032540210715](https://jules.google.com/task/1639953032540210715) started by @dogi*


Okuro3499 (dismissed):
```
Problem: The comparator only checks id, _rev, isUploaded. updatePersonalResource(id, updater) (called from PersonalsFragment.onEditPersonal →                                    
PersonalsViewModel.updatePersonalResource) mutates title/description directly without touching _rev or isUploaded. Room still emits a new list from the DAO flow, but the new    
distinctUntilChanged comparator will treat it as unchanged and swallow the emission. Result: a user edits a personal resource's title/description and the visible list on        
PersonalsFragment silently keeps showing the old title until some unrelated change (upload, delete, app restart) breaks the false-equality. This is a regression relative to     
pre-PR behavior, on the screen's core edit flow. 
```

dogi:
**Review — rating: 76/100**

The premise is sound and matches existing house practice — Room re-emits on any write to a watched table regardless of whether the queried rows changed, so unguarded flows re-render repeatedly during sync. `CoursesRepositoryImpl.getMyCoursesFlow` already carries this exact shape. And using a field comparator rather than plain `distinctUntilChanged()` is *forced* here, not sloppy: these are Room `@Entity` open classes, not data classes, so `equals` is identity and a bare guard would suppress nothing. Using the bare form for `getPendingDownloads` (a `List<String>`) is the right call for that one.

The tests are good too — each guard gets a test that pushes two equal-content emissions and asserts one survives.

Three things to fix before this lands:

**1. It adds three new instances of the pattern #15932 is removing.** That PR is titled "drop allocations in `VoicesRepositoryImpl` `distinctUntilChanged` predicates" and targets exactly this: `old.zip(new)` allocates a list of `Pair`s on every comparison, and this flow fires often *precisely* during the sync storm the guard exists to absorb. Indexing avoids it:

```kotlin
old.size == new.size && old.indices.all { i -> sameItem(old[i], new[i]) }
```

Better to land in that form than to have #15932's cleanup immediately follow you into three more files.

**2. The comparator is copy-pasted twice inside `FeedbackRepositoryImpl`** — the identical five-field lambda on both the manager and owner branches. And the surrounding `old.size == new.size && old.zip(new).all { … }` scaffolding is now duplicated across this PR, `CoursesRepositoryImpl`, and `VoicesRepositoryImpl`. One shared extension would collapse all of it:

```kotlin
inline fun <T> Flow<List<T>>.distinctByContent(crossinline sameItem: (T, T) -> Boolean): Flow<List<T>>
```

Each call site then supplies only the per-entity field comparison, which is the part that actually differs.

**3. Worth stating the invariant these comparators rely on.** Each is an allowlist, so any field that changes *without* `id`/`_rev` changing is silently suppressed and the UI goes stale. That's fine for sync-driven updates, since CouchDB bumps `_rev`, and you've correctly included the locally-mutated flags (`isUploaded`, `resourceOffline`, `status`, `messages`). The risk is local write paths: for `getRecentResources`, if a completed download writes something like `resourceLocalAddress` without touching `resourceOffline` or `_rev`, that emission is dropped. Worth confirming for `MyLibrary` specifically, and worth a comment on each comparator so the next person adding a field knows to extend it.

**Nit:** the new tests use `mockk<org.ole.planet.myplanet.model.UserEntity>` fully qualified inline rather than importing `UserEntity` — #15698 cleaned up exactly this style in another test file.

**Verdict:** needs another round — swap `zip` for indexing, de-duplicate the comparator, and confirm the `MyLibrary` field list.

---

dogi:
**Re-review — Rating: 92/100** (was 76)

All three items addressed, and the third one properly rather than by assertion.

**1. The shared extension replaces the copy-paste and the `zip`:**

```kotlin
inline fun <T> Flow<List<T>>.distinctByContent(crossinline sameItem: (T, T) -> Boolean): Flow<List<T>> =
    distinctUntilChanged { old, new -> old.size == new.size && old.indices.all { i -> sameItem(old[i], new[i]) } }
```

No `Pair` allocation per comparison, which matters because these flows fire hardest during the sync storm the guard exists to absorb. And each call site now supplies only the per-entity field comparison — the part that actually differs.

**2. `FeedbackRepositoryImpl`'s duplicate comparator is gone.** Hoisting the manager/owner choice into `val flow = if (…) … else …` and applying one `distinctByContent` afterwards is cleaner than what I suggested; there's now a single comparator instead of the same five-field lambda on both branches.

**3. The `MyLibrary` field list was actually re-examined**, not just documented. `resourceLocalAddress` and `downloadedRev` are now in the comparator, which closes exactly the gap I raised — a completed download writing a local path without touching `resourceOffline` or `_rev` would previously have been suppressed and the recent-resources list would have gone stale. And each comparator carries a comment naming what it covers, so the next person adding a field knows to extend it.

`getPendingDownloads` correctly uses the bare `distinctUntilChanged()` — it's a `List<String>` with structural equality, so no comparator is needed there.

The tests are the strongest part: dedup coverage on all four flows, both `FeedbackRepositoryImpl` branches, and — most valuable — `getPersonalResources emits when local properties like title change`, which proves the comparator isn't over-suppressing. A dedup guard without that second test is how flows go silently stale.

**Nit:** `distinctByContent` lives in `FlowExtensions.kt` alongside `collectWhenStarted`/`collectLatestWhenStarted`, which are `Fragment` extensions for UI lifecycle collection. A repository-layer flow operator in that file is a slightly odd fit — `utils/FlowOperators.kt` or similar would separate "how the UI collects" from "how a flow dedupes".

**Nit — natural follow-up:** now that this exists, `VoicesRepositoryImpl`'s two hand-rolled `old.size == new.size && old.zip(new).all { … }` predicates are the last copies of the old shape. #15932 in this queue is editing exactly those two predicates right now, so it's worth coordinating: whichever lands second could convert them to `distinctByContent` and delete the scaffolding entirely.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #15861, opened 2026-08-19
chat: smoother history ai provider view modelling (fixes #15999)

### LINKED ISSUE #15999: Refactor AI provider fetch to ViewModel and remove duplicate repo injection
Tracking issue for PR #15861.

This change resolves the boundary erosion in the Chat feature by pointing `ChatHistoryFragment`'s AI provider fetch call to `ChatViewModel`, allowing us to delete the duplicate `ChatRepository` injected into the fragment. It also refactors the ViewModel's `fetchAiProviders` method into a non-suspending `viewModelScope` call that mutates local state flows, preventing fragments from needing to launch coroutines merely for data calls. Tests have been updated and run locally to confirm correctness.

---
*PR created automatically by Jules for task [8372757692291457348](https://jules.google.com/task/8372757692291457348) started by @dogi*

dogi:
This change resolves the boundary erosion in the Chat feature by pointing `ChatHistoryFragment`'s AI provider fetch call to `ChatViewModel`, allowing us to delete the duplicate `ChatRepository` injected into the fragment. It also refactors the ViewModel's `fetchAiProviders` method into a non-suspending `viewModelScope` call that mutates local state flows, preventing fragments from needing to launch coroutines merely for data calls. Tests have been updated and run locally to confirm correctness.

---
*PR created automatically by Jules for task [8372757692291457348](https://jules.google.com/task/8372757692291457348) started by @dogi*


Okuro3499 (approved):

dogi:
**Review — rating: 86/100**

This fixes a permanent stuck-state bug, which the title undersells. Worth spelling out:

The old code launched the fetch on `viewLifecycleOwner.lifecycleScope` but wrote the result into the **shared** ViewModel. If the view was destroyed mid-fetch — and `openCallFragment` replaces with `addToBackStack`, so navigating away does exactly that — the coroutine was cancelled before reaching `setAiProvidersLoading(false)`, leaving `_aiProvidersLoading = true` in shared state. And the guard is:

```kotlin
fun shouldFetchAiProviders(): Boolean =
    _aiProviders.value == null && !_aiProvidersLoading.value
```

So with providers still null and loading stuck true, `shouldFetchAiProviders()` returns false **forever** and the AI providers can never be fetched again for the life of that ViewModel. Moving the coroutine to `viewModelScope` means the fetch finishes and the flags settle regardless of the view — the right scope, since the result outlives the view. On a slow or offline connection, which is this app's normal condition, that's easy to hit.

Two other real improvements:

- **Layer violation removed.** `ChatHistoryFragment` injected `ChatRepository` and called it directly while its sibling went through the ViewModel.
- **Two drifted copies consolidated.** `ChatDetailFragment`'s version fell back to cached providers; `ChatHistoryFragment`'s didn't. The optional `cachedProviders` parameter preserves both behaviors rather than silently picking one. Adding the explicit `setAiProvidersError(false)` on the success path is equivalent to the old code (which set it false before launching) and clearer.

**Nit — the cached lookup is now eager.** `fetchAiProviders(serverUrl, getCachedProviderAvailability())` evaluates the cache on every call, including calls that bail out immediately at `shouldFetchAiProviders()`, and including the common case where the fetch succeeds and the cache is never used. A lambda parameter (`cachedProviders: () -> Map<String, Boolean>? = { null }`) keeps it lazy.

**Nit — missing tests on the branch that matters.** The updated test covers the happy path only. The cached-fallback branch (`providers == null` → cached non-null → error false, providers = cached) is the logic that just became centralized and testable, and a test asserting `aiProvidersLoading` settles to false is the regression guard for the bug above.

**Question:** both fragments write the same shared state, so is it deliberate that only `ChatDetailFragment` passes the cache? The asymmetry was there before, but it's now a bare argument difference between two call sites with nothing explaining it — worth either aligning them or leaving a comment.

**Verdict:** mergeable.

---

PR #15862, opened 2026-08-19
teams: smoother repository next leader querying (fixes #16000)

### LINKED ISSUE #16000: Batch the next-leader user lookup (N+1)
Tracking issue for PR #15862.

Batch the next-leader user lookup (N+1)

`TeamsRepositoryImpl.getNextLeaderCandidate` (`:1145-1147`) calls
`userRepository.getUserById(userId)` once per candidate member inside `mapNotNull`.

Replaced the `mapNotNull` loop calling `userRepository.getUserById(userId)` with a single call to `userRepository.getUsersByIds(memberUserIds)`.

Note: The required batch method `getUsersByAnyIds` (used by `getUsersByIds`) already exists in `UserDao` and `UserRepositoryImpl`. No additional method needed to be added to `UserDao`.

---
*PR created automatically by Jules for task [5884245311346002037](https://jules.google.com/task/5884245311346002037) started by @dogi*

## Summary by CodeRabbit

  * Improved leader candidate selection by retrieving member information in a single batch.
  * Reduced unnecessary repeated data requests while preserving existing successor-selection behavior.

dogi:
Batch the next-leader user lookup (N+1)

`TeamsRepositoryImpl.getNextLeaderCandidate` (`:1145-1147`) calls
`userRepository.getUserById(userId)` once per candidate member inside `mapNotNull`.

Replaced the `mapNotNull` loop calling `userRepository.getUserById(userId)` with a single call to `userRepository.getUsersByIds(memberUserIds)`.

Note: The required batch method `getUsersByAnyIds` (used by `getUsersByIds`) already exists in `UserDao` and `UserRepositoryImpl`. No additional method needed to be added to `UserDao`.

---
*PR created automatically by Jules for task [5884245311346002037](https://jules.google.com/task/5884245311346002037) started by @dogi*

## Summary by CodeRabbit

  * Improved leader candidate selection by retrieving member information in a single batch.
  * Reduced unnecessary repeated data requests while preserving existing successor-selection behavior.


dogi:
**Review — rating: 88/100**

Reviewed alongside #15936, which also targets `getNextLeaderCandidate`. **They are not duplicates** — this PR batches the *user lookup*, #15936 pushes the *member filter* into SQL. Different halves of the same function, and they compose: together both bottlenecks are gone. Whoever merges second gets a trivial textual conflict, nothing more.

On the change itself, two things could have broken it and neither does:

- **Order doesn't matter here.** `WHERE id IN (…)` returns rows in arbitrary order, not input order, so I checked how `users` is consumed. It's only used to build `userMap = users.associateBy { it.id }`; the actual selection runs over `members` via `maxByOrNull`, and the return is `successorMember?.userId?.let { userMap[it] }`. The list is a lookup table, never an ordered sequence, so the reordering is invisible.
- **The IN-clause parameter limit is already handled.** `getUsersByIds` chunks at 400 and `distinctBy { it.id }`s the result, so there's no SQLite bound-parameter ceiling to hit on a large team — the fix #15919 is retrofitting in `TagsRepositoryImpl` is already in place here.

Equivalence otherwise holds: both the old `mapNotNull { getUserById(it) }` and the new batch drop ids with no matching user row, and the extra dedup is harmless since `associateBy` collapses duplicates anyway.

Net effect: N sequential suspend queries (one per team member) become `ceil(N/400)`. That's the right fix for an N+1.

**Nit:** the `if (userIds.isNotEmpty()) … else emptyList()` guard is redundant — `getUsersByIds` already returns `emptyList()` on empty input as its first line. This reduces to:

```kotlin
val users = userRepository.getUsersByIds(members.mapNotNull { it.userId })
```

**Verdict:** mergeable — and worth landing together with #15936 rather than choosing between them.

---

Okuro3499 (approved):

PR #15863, opened 2026-08-19
all: smoother dictionary repository providing (fixes #16001)

### LINKED ISSUE #16001: Refactor DictionaryActivity to use DictionaryRepository
Tracking issue for PR #15863.

Extract Dictionary logic from Activity into a dedicated Repository to seal UI->DAO leaks.

---
*PR created automatically by Jules for task [9647196109042178604](https://jules.google.com/task/9647196109042178604) started by @dogi*

dogi:
Extract Dictionary logic from Activity into a dedicated Repository to seal UI->DAO leaks.

---
*PR created automatically by Jules for task [9647196109042178604](https://jules.google.com/task/9647196109042178604) started by @dogi*


dogi:
**Review — rating: 68/100**

The goal is right and worth doing — `DictionaryActivity` was the last place injecting a DAO straight into the UI layer, and `CLAUDE.md` calls that out explicitly. The `@Binds` wiring is correct, and the onCreate branch is faithfully preserved (`insertDictionaryData()` returns false exactly when the file is missing, which is what the old `FileUtils.checkFileExist` guard decided).

But the refactor moved the I/O into a repository without fixing its dispatcher discipline, which is the main reason to do this move at all.

**1. A full-dictionary JSON parse runs on the main thread.** In `insertDictionaryData()`, only the file *read* is inside `withContext(dispatcherProvider.io)`:

```kotlin
val data = withContext(dispatcherProvider.io) { FileUtils.getStringFromFile(…) }
JsonUtils.gson.fromJson(data, JsonArray::class.java)   // caller's dispatcher
…
val entities = jsonArray.map { … }                     // caller's dispatcher
```

Both the Gson parse of the whole file and the `map` that builds thousands of `DictionaryEntity` objects (eight `JsonUtils.getString` calls apiece) run on the caller's context — and the caller is `lifecycleScope.launch`, i.e. Main. `FileUtils.checkFileExist` at the top is on Main too. Only `insertAll` is on IO.

This is carried over verbatim from the old Activity code, so it isn't a regression — but a dictionary big enough to need its own download is a genuine ANR risk (the app ships an `ANRWatchdog`), and a repository is precisely where this codebase expects the whole operation wrapped. One `withContext(dispatcherProvider.io)` around the entire body fixes it.

**2. The `Boolean` conflates three outcomes.** `true` means "already populated", "inserted successfully", *and* "file existed but the JSON failed to parse" — the `json == null` path falls straight through to `return true`. So a corrupt dictionary file reports success, and the UI shows a count of 0 with search listeners wired up and no way to recover. Worth distinguishing that case.

**3. Two call sites now behave differently.** The missing-file branch moved into `onCreate`, but `loadDictionaryIfNeeded()` — still called from the broadcast receiver on download completion (line 55) — calls `insertDictionaryData()` and discards a `false` return, then sets the count and wires listeners as if it succeeded. Putting the branch inside the shared method would keep both paths consistent, and would also stop `onCreate` from re-implementing the three lines `loadDictionaryIfNeeded()` already does.

**Nit:** `withContext(dispatcherProvider.io)` around `count()` and `findByWord()` is redundant — Room `suspend` DAO methods already execute off the main thread via Room's own executors, as `CLAUDE.md` notes.

**Nit:** `import …repository.DictionaryRepository` is appended at the bottom after `utils.collectWhenStarted` and separated by a blank line; it belongs in alphabetical order with the other `repository` imports.

**Nit:** no test for a brand-new repository. Every other repository in `repository/` has one, and `insertDictionaryData`'s branches (missing file, already populated, parse failure) are exactly what a unit test should pin.

**Verdict:** needs another round — mainly getting the parse and mapping off the main thread while the code is in your hands.

---

Okuro3499 (dismissed):

dogi:
**Re-review — Rating: 78/100** (was 68)

Three of the four items are properly fixed:

- **The whole operation is on IO now.** `withContext(dispatcherProvider.io)` wraps the file read, the Gson parse *and* the `map` that builds the entities, so the ANR risk on a large dictionary is gone. That was the main reason to do this move.
- **Redundant wrappers removed** — `count()` and `findByWord()` call the DAO directly, which is correct since Room's `suspend` DAO methods already run off the main thread.
- **Tests added** covering all four branches: missing file, already populated, parse failure, and successful insert with a real JSON payload. Note the success test also pins the `"antonoym"` key spelling, which is a nice accident — that typo is load-bearing and now protected.

The `Boolean` ambiguity and the two-call-site inconsistency were both handled by folding the missing-file branch into `loadDictionaryIfNeeded()`, which is the right shape. But consolidating them created a new path worth one more round:

**A file that downloads but doesn't parse now re-downloads forever.** `insertDictionaryData()` returns `false` for both "file missing" and "parse failed" (the `json == null` and `catch` branches), and the Activity treats any `false` as "fetch it":

```kotlin
if (dictionaryRepository.insertDictionaryData()) { … } else {
    DownloadUtils.openDownloadService(this@DictionaryActivity, list, false)
}
```

The receiver at line 53 fires `loadDictionaryIfNeeded()` on `download.progress == 100`, so the cycle closes: download completes → parse fails → `false` → download again → completes → parse fails. Each lap also fires the "downloading started" toast. Before this PR the receiver path parsed, failed, and simply fell through to setting the count, so this loop is new. It needs a genuinely unparseable (or mis-pathed) file to trigger, but it's unbounded once triggered, and burning bandwidth in a loop is a bad failure mode for the deployments this app targets.

Distinguishing the outcomes fixes both this and the ambiguity I raised last time:

```kotlin
sealed interface DictionaryLoad {
    data object FileMissing : DictionaryLoad      // download
    data object AlreadyPopulated : DictionaryLoad // render
    data object Inserted : DictionaryLoad         // render
    data class Failed(val cause: Throwable?) : DictionaryLoad  // surface, don't re-download
}
```

**Nit:** `FileUtils.checkFileExist(...)` at the top of `insertDictionaryData` still runs on the caller's dispatcher (Main). It's one `File.exists()`, so it's not the ANR risk the parse was — but it's inside the function whose whole job is now "do the I/O off Main", so it may as well move inside the `withContext`.

**Nit:** `import org.ole.planet.myplanet.repository.DictionaryRepository` is still out of order — it now sits between `utils.DispatcherProvider` and `utils.DownloadUtils` rather than with the other `repository` imports.

**Nit:** the new tests carry leftover debugging — two `println` calls and a `try { … } catch (e: Throwable) { e.printStackTrace(); throw e }` wrapper in `json parsing fails` that only rethrows.

**Verdict:** needs another round — mainly so a bad file stops triggering an endless re-download.

---

dogi:
**Re-review — Rating: 72/100** (was 78)

The design ask is done properly — and then two things came in with it that have to go before this can merge.

**What's fixed:**

- **`DictionaryLoad` is implemented as suggested** — `FileMissing` / `AlreadyPopulated` / `Inserted` / `Failed(cause)`, and the Activity's `when` routes them correctly: only `FileMissing` triggers `openDownloadService`, and `Failed` toasts `dictionary_parsing_failed` without re-downloading. The endless download→parse-fail→download loop is closed, which was the blocker.
- **`checkFileExist` moved inside `withContext(dispatcherProvider.io)`**, so the whole function is off the main thread as intended.
- **Debug leftovers gone** from the tests (no `println`, no rethrow wrapper).
- Dropping the pre-check from `onCreate` in favour of one `loadDictionaryIfNeeded()` call is the right consolidation — the file-existence decision now lives in one place.

**1. CI is red — `myPlanet test (default, shard 1/2)` failed on this head** ([job 96535309842](https://github.com/open-learning-exchange/myplanet/actions/runs/32402974512/job/96535309842)); shard 2, build (default) and build (lite) all passed. This isn't a base problem: the same shard is green on the nine other PRs I walked in this batch off the same master, and master's own test run is green. The log tail I can retrieve only reaches the post-job block (`testDefaultDebugUnitTest` marked failed, "Cache cleanup was skipped due to a build failure"), so I can't name the failing test from here — but `DictionaryRepositoryImplTest` is the only test class this diff adds or rewrites, so that's where to look first.

One thing in it worth examining while you're there: `unmockkAll()` in `@After` tears down **every** MockK object mock in the JVM, not just this class's `mockkObject(FileUtils)`. `ChatApiServiceTest` in this same suite already carries a comment warning that `mockkObject` creates global singleton state and that parallel tests would be flaky because of it. `unmockkObject(FileUtils)` scopes the teardown to what this class actually mocked.

**2. `plan.md` is committed at the repo root.** It's a 32-line agent worklog naming internal tooling (`replace_with_git_merge_diff`, `read_file`), and step 7 is literally "Request review and delete plan" — which didn't happen. It needs deleting from the branch.

It also documents two things the diff didn't deliver. Step 3 says "Correctly alphabetize `org.ole.planet.myplanet.repository.DictionaryRepository` import" — the imports still land between `utils.Constants` and `utils.DispatcherProvider`, so that carried-over nit is still open. Step 6 says to run `./gradlew testDefaultDebugUnitTest`, which is the check that's now failing.

**Nit:** the four test names still describe the old Boolean API — `returns false if file does not exist`, `returns true if data is already populated`, `returns false if json parsing fails`, `returns true and inserts entities on success` — while the assertions check `DictionaryLoad` types. Renaming them to the states keeps the failure output readable.

**Nit:** `assertFalse` is imported but no longer used.

**Nit:** `DictionaryActivity.loadDictionaryCount()` is now a one-line pass-through to `dictionaryRepository.count()` with a single caller (the `collectWhenStarted` block). Worth inlining.

**Verdict:** needs another round — get shard 1 green and drop `plan.md`; the `DictionaryLoad` work itself is right.

---

dogi:
**Re-review — Rating: 78/100** (was 72)

CI half of the last verdict is cleared: `myPlanet test (default, shard 1/2)` is green now (job 96540013035). Both shard jobs were re-run under the same workflow run while the build jobs stayed at their original 18:25 timestamps — so the head didn't move and nothing in the code changed between the red run and the green one. That makes the earlier failure a flake rather than a real break, and `DictionaryRepositoryImplTest` is fine.

`plan.md` is still committed at the repo root, so this stays blocked on that alone. Deleting it from the branch is the whole remaining ask.

While it's open, the three nits from last round are also still standing: the `repository.DictionaryLoad`/`DictionaryRepository` imports still sit between `utils.Constants` and `utils.DispatcherProvider`; the four test names still describe the old Boolean API (`returns false if file does not exist` for a function that returns `DictionaryLoad.FileMissing`); and `assertFalse` is imported unused.

Given the flake, I'd also suggest scoping the teardown — `unmockkObject(FileUtils)` instead of `unmockkAll()`. It isn't the cause of this failure (nothing changed between the two runs), but a global unmock in `@After` is shared-state cleanup reaching outside its own class, and this suite already carries a written warning about `mockkObject` singleton state in `ChatApiServiceTest`.

**Verdict:** needs another round — drop `plan.md` and this is done.

---

dogi:

`plan.md` is gone from the branch, which was the whole remaining blocker. The three nits also landed: the `repository.*` imports now sit correctly between `model.Download` and `services.BroadcastService`, and the four test names describe the sealed-interface API (`returns FileMissing…`, `returns AlreadyPopulated…`) rather than the old Boolean. Shard 1 is green on the current head (job 96560885210), so the earlier red is behind us.

The end state is good: `insertDictionaryData()` does the existence check, the count check, the read, the parse and the entity mapping all inside one `withContext(dispatcherProvider.io)`, and `count()`/`findByWord()` correctly go straight to the DAO since Room's suspend methods dispatch themselves.

**Nit:** on `DictionaryLoad.Failed` the Activity toasts and returns without calling `setClickListener()`, so after a parse failure the search button is inert. The old code fell through to `setClickListener()` on that path (searching an empty table just reported "word not available"). Leaving the button dead is defensible, but it's a behaviour change worth being deliberate about.

**Nit:** `e.printStackTrace()` is redundant now that `Failed(cause)` carries the throwable to the caller.

**Verdict:** ready to merge.

---

Okuro3499 (approved):

PR #15864, opened 2026-08-19
courses: smoother resources payload adapting (fixes #16002)

### LINKED ISSUE #16002: perf: Add payload-scoped range invalidation for Courses and Resources adapters
Tracking issue for PR #15864.

This PR introduces payload-scoped range invalidations for `CoursesAdapter` and `ResourcesAdapter` to avoid unnecessary full row rebinding and Glide reloads when changing view mode or guest identity status. 

### Changes
- **Payload Constants:** Added `PAYLOAD_VIEW_MODE` and `PAYLOAD_IDENTITY`.
- **Targeted Updates:** 
  - `setViewMode` now passes `PAYLOAD_VIEW_MODE`.
  - `updateIdentity` now passes `PAYLOAD_IDENTITY`.
- **Partial Binding:** 
  - Overridden `onBindViewHolder(holder, position, payloads)` handles `PAYLOAD_IDENTITY` by updating only guest-dependent UI elements (e.g. checkbox visibility).
  - Explicitly prevents reloading covers via Glide on an identity-only update.
  - `PAYLOAD_VIEW_MODE` still correctly forces a full bind because it involves structural changes (Grid vs List).
- **Guest Cleanup:** `ResourcesAdapter` now correctly clears check listeners when switching to guest mode.

---
*PR created automatically by Jules for task [7460191229448487224](https://jules.google.com/task/7460191229448487224) started by @dogi*

dogi:
This PR introduces payload-scoped range invalidations for `CoursesAdapter` and `ResourcesAdapter` to avoid unnecessary full row rebinding and Glide reloads when changing view mode or guest identity status. 

### Changes
- **Payload Constants:** Added `PAYLOAD_VIEW_MODE` and `PAYLOAD_IDENTITY`.
- **Targeted Updates:** 
  - `setViewMode` now passes `PAYLOAD_VIEW_MODE`.
  - `updateIdentity` now passes `PAYLOAD_IDENTITY`.
- **Partial Binding:** 
  - Overridden `onBindViewHolder(holder, position, payloads)` handles `PAYLOAD_IDENTITY` by updating only guest-dependent UI elements (e.g. checkbox visibility).
  - Explicitly prevents reloading covers via Glide on an identity-only update.
  - `PAYLOAD_VIEW_MODE` still correctly forces a full bind because it involves structural changes (Grid vs List).
- **Guest Cleanup:** `ResourcesAdapter` now correctly clears check listeners when switching to guest mode.

---
*PR created automatically by Jules for task [7460191229448487224](https://jules.google.com/task/7460191229448487224) started by @dogi*


dogi:
**Review — rating: 84/100**

The risk with payload-scoped binds is stranding state that only the full bind used to refresh, so I traced every identity-dependent site in both adapters. Both are complete:

- **`CoursesAdapter`**: `isGuest` is read in exactly one place — `setupCheckbox` (line 322). Mapping `PAYLOAD_IDENTITY` onto the selection branch re-runs `updateVisibilityForMyCourse` *and* `setupCheckbox`, so nothing identity-dependent is missed.
- **`ResourcesAdapter`**: four sites depend on identity — checkbox visibility (`bindSelectionAndDownload`, line 363), the `isOwnResource` long-click gate (lines 388–390, using both `currentUserName` and `isGuest`), and the checkbox click listener (line 400). Calling `bindSelectionAndDownload` *and* `bindClicks` covers all four. That's why the extra `bindClicks` call is there and it's the right call.

**The `else { checkbox.setOnClickListener(null) }` is a real bug fix**, worth calling out separately: `if (!isGuest) { checkbox.setOnClickListener { … } }` had no else, so a ViewHolder recycled into a guest state kept its stale click listener. That becomes more reachable now that identity changes take the partial path, and clearing it is correct.

The win is worth having — the full bind does Glide cover loads and preview resolution per item, so skipping it on an identity flip is not a micro-optimization.

**One thing the description oversells:** `PAYLOAD_VIEW_MODE` delivers no performance change. Both the new branch and the old unpayloaded notify end at `super.onBindViewHolder(holder, position, payloads)`, i.e. a full bind. That's *correct* — a view-mode change flips `getItemViewType`, so a full rebind and a new ViewHolder are required — and checking it first is a good call, since a coalesced view-mode + selection payload then safely takes the superset path. But it's a documented no-op rather than an optimization, and the PR reads as though both payloads buy something.

**Nit:** the two adapters now express the same concept differently — Courses folds identity into the selection flag, Resources calls two bind functions. Same idea, two shapes, in two files that already drift; a short comment or a shared convention would help the next person.

**Nit:** no test. This is precisely the change where one missed identity-dependent field becomes a subtle stale-UI bug, and the coverage today is the green build plus manual reasoning.

**Verdict:** mergeable.

---

Okuro3499 (dismissed):
```
File: app/src/test/java/org/ole/planet/myplanet/ui/courses/CoursesAdapterTest.kt and a new ResourcesAdapterTest.kt.                                                              
Problem: No test exercises setViewMode/updateIdentity payload dispatch or verifies that onBindViewHolder(holder, position, payloads) correctly re-runs setupCheckbox/bindClicks  
on PAYLOAD_IDENTITY without a full rebind. This is exactly the kind of partial-bind logic that's easy to silently regress (e.g., a future payload combined incorrectly, or       
bindClicks guard forgotten).
```

dogi:
**Review — rating: 88/100**

Correct, and it carries a real bug fix that the title doesn't mention.

**The bug fix:** `setupCheckbox` returned early on both the `isGuest` and `!showCheckbox` paths after setting `visibility = GONE`, but left any previously-attached click listener in place. On a recycled holder that meant a hidden checkbox still carrying a live `setOnClickListener` from whatever row occupied that view before. Adding `checkbox.setOnClickListener(null)` to both early returns fixes it — and it matters *more* under this PR, because a partial bind no longer re-runs the full bind that used to overwrite the listener. Same fix on the `ResourcesAdapter` side via the new `else { checkbox.setOnClickListener(null) }`. Both new tests assert `hasOnClickListeners() == false`, so the fix is pinned rather than incidental.

**The payload scoping is right, including the part that deliberately doesn't optimize.** `PAYLOAD_IDENTITY` takes the partial path (re-run `updateVisibilityForMyCourse` + `setupCheckbox` only), so a guest-mode flip no longer re-binds titles and re-issues Glide loads for every row. `PAYLOAD_VIEW_MODE` falls through to `super.onBindViewHolder` — which is correct, since `getItemViewType` keys off `viewMode`, so a mode switch changes the view type and RecyclerView must rebuild the holder anyway.

I also checked the null handling: `partialHandled` is only set once `getItem(position)` returns non-null, so a null item now falls through to the full bind instead of the old bare `?: return`. That's a small improvement over master.

**Nit:** since `PAYLOAD_VIEW_MODE` always ends in a full rebind, its only real effect is suppressing the default change animation on `notifyItemRangeChanged`. That's worth a one-line comment — otherwise the next reader sees a payload constant that appears to do nothing and "simplifies" it away.

**Nit:** `ResourcesAdapterTest` declares `@Mock lateinit var mockContext: Context` and never uses it — `testContext` from `ApplicationProvider` is used throughout.

**Nit:** it also hand-rolls an anonymous `DispatcherProvider` with four `Dispatchers.Unconfined` fields, where the project already has `TestDispatcherProvider` for exactly this.

**Nit:** both test files write a lot of fully-qualified names inline — `androidx.test.core.app.ApplicationProvider`, `com.google.android.material.R.style.Theme_MaterialComponents`, `android.widget.LinearLayout`, `android.view.View.GONE`, `org.ole.planet.myplanet.utils.ListViewMode.LIST` — several of which have imports available in the same file. `import org.ole.planet.myplanet.model.MyLibrary` is also out of alphabetical order.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #15865, opened 2026-08-19
resources: smoother repository offline dao marking (fixes #16003)

### LINKED ISSUE #16003: Optimize markResourcesAsNotOffline to use single UPDATE statement
Tracking issue for PR #15865.

Replaced the inefficient read-modify-write pattern in `ResourcesRepositoryImpl.markResourcesAsNotOffline` with a single Room `@Query` `UPDATE` statement that uses an `IN (:ids)` clause.
Also filters by `resourceOffline = 1` to prevent unnecessary writes.

---
*PR created automatically by Jules for task [6765606281889409191](https://jules.google.com/task/6765606281889409191) started by @dogi*

dogi:
Replaced the inefficient read-modify-write pattern in `ResourcesRepositoryImpl.markResourcesAsNotOffline` with a single Room `@Query` `UPDATE` statement that uses an `IN (:ids)` clause.
Also filters by `resourceOffline = 1` to prevent unnecessary writes.

---
*PR created automatically by Jules for task [6765606281889409191](https://jules.google.com/task/6765606281889409191) started by @dogi*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling when resources are removed from offline storage, ensuring their offline status is updated consistently.
  * Streamlined bulk updates for multiple resources.

* **Chores**
  * Updated the Android app version to 0.65.81.


Okuro3499 (approved):

dogi:
**Review — rating: 88/100**

Correct and a bigger win than "one statement instead of two" suggests.

- **Same rows, same effect.** The old `getOfflineByResourceIds` selected on `resourceId IN (…) AND resourceOffline = 1` and the new `UPDATE` carries the identical predicate, so the targeted set is unchanged.
- **It stops rewriting whole rows.** `upsertAll` wrote every column of each matched row, which means running the `@TypeConverters` (Gson) round-trip per row just to flip one boolean. The `UPDATE` touches one column.
- **It removes a lost-update hazard.** The old read-modify-write took an in-memory copy and upserted all columns back, so any concurrent change to another column of those rows between the SELECT and the upsert was silently clobbered. A targeted `UPDATE` can't do that.
- **Atomic now** — one statement in one implicit transaction, rather than a SELECT followed by a separate write.

The thing that would have made this silently break the UI is Room invalidation, and it holds: Room's `InvalidationTracker` derives affected tables from `@Query` write statements too, so Flows observing `my_library` still emit after the UPDATE.

**Nit — unbounded `IN` clause, and this function has a caller that will exercise it.** `resourceIds.toList()` goes straight into `IN (:ids)` with no chunking, so a large set can exceed SQLite's bound-parameter limit (999 on older versions). `FreeSpaceWorker` calls this with every resource directory it cleared while freeing disk space — by nature the case where the list is biggest. The old code had exactly the same exposure so it's not a regression, but the codebase already has the pattern to copy: `UserRepositoryImpl.getUsersByIds` chunks at 400, and #15919 is retrofitting the same fix in `TagsRepositoryImpl`. Since you're already in this function, it's the natural moment.

**Nit:** nothing exercises the new SQL — `FreeSpaceWorkerTest` mocks the repository, so the query itself is untested. #15859 just landed the in-memory `Room.inMemoryDatabaseBuilder` pattern in `ExamDaoTest` if you want a cheap template (skip the `@Config(sdk = …)` pin; #15914 is removing those).

**Verdict:** mergeable.

---

PR #15867, opened 2026-08-20
login: smoother storage category view modelling (fixes #16004)

### LINKED ISSUE #16004: refactor: extract storage category detail logic to ViewModel
Tracking issue for PR #15867.

Extracted storage listing and deletion logic into `StorageCategoryViewModel`. The fragment only acts as a dumb UI component passing selection states up and submitting list payload states from the ViewModel. Concurrent duplicate deletions are strictly rejected.

---
*PR created automatically by Jules for task [8530915726999371506](https://jules.google.com/task/8530915726999371506) started by @dogi*

dogi:
Extracted storage listing and deletion logic into `StorageCategoryViewModel`. The fragment only acts as a dumb UI component passing selection states up and submitting list payload states from the ViewModel. Concurrent duplicate deletions are strictly rejected.

---
*PR created automatically by Jules for task [8530915726999371506](https://jules.google.com/task/8530915726999371506) started by @dogi*


Okuro3499 (approved):

dogi:
**Review — rating: 84/100**

Solid extraction, and two choices in it are better than the minimum:

- **`Channel` + `receiveAsFlow` for the dismiss signal** is the right call rather than a StateFlow — a one-shot event shouldn't re-fire when the collector restarts, and because `collectWhenStarted` uses `repeatOnLifecycle(STARTED)`, the buffered channel still delivers a completion that lands while the sheet is stopped.
- **Selections now survive rotation.** `items` was a fragment field before, so checkbox state was lost on configuration change; it's ViewModel state now. `hasLoaded` correctly stops a reload from wiping it.

Plus the repository injection is out of the fragment and `DispatcherProvider` is injected rather than hard-coded.

I went looking for a regression in the visibility state machine, since the old `loadResources` explicitly forced `resourceList`, `emptyText`, `actionButtons`, `selectAllRow` and `selectAllDivider` to GONE before launching and the new collector has no `isLoading` branch. It's fine in practice: all five are `visibility="gone"` in the layout, and `loadResources` flips `isLoading` synchronously before the collector is scheduled, so the first emission the collector sees is already the loading state. Worth knowing that it holds by virtue of XML defaults plus call ordering rather than because the state machine says so — an explicit `state.isLoading ->` branch would make it self-evident and would be needed if a reload path is ever added.

Two things I'd tighten:

**1. `isDeleting` never re-enables the buttons.** The collector has `if (state.isDeleting) { … isEnabled = false }` with no else, so nothing restores them. Mostly masked because a successful delete dismisses the sheet, but it's asymmetric — `isEnabled = !state.isDeleting` on both buttons is the same line count and always correct.

**2. A failed delete now sticks around longer than it used to.** There's no try/catch around `deleteOfflineResources`, so a throw leaves `isDeleting = true`, sends no event, and the sheet never dismisses with both buttons disabled. The old code had the same missing try/catch, but the difference is durability: the fragment-scoped coroutine died with the view, so reopening the sheet reset everything, whereas this state now lives in the ViewModel. Worth a `runCatching` that clears `isDeleting` and surfaces a failure.

**Nit:** no test for a new ViewModel whose logic is entirely pure and trivially testable — `toggleAllChecked`'s all-checked inversion, `toggleItemChecked`, and the `hasLoaded` guard are three quick assertions with a mocked repository.

**Nit:** `import javax.inject.Inject` sits at the bottom after the `org.ole.planet…` imports rather than in style-guide order (same as #15858).

**Verdict:** mergeable.

---

PR #15868, opened 2026-08-20
teams: smoother voices sorting (fixes #16207)

### LINKED ISSUE #16207: Optimize `VoicesFragment` news sorting into a single pass
Tracking issue for PR #15868.

This commit optimizes the `sortNews` and `setData` functions in `VoicesFragment`.
By converting `sortNews` to return a `List<News>`, it naturally filters out null elements once at the top of the processing block, allowing the `setData` consumers to reuse the processed list without incurring multiple redundant `.filterNotNull()` allocations and iterations.
The change preserves the `Trace.beginSection()` timing behavior and removes the unnecessary `toMutableList()` intermediate step.

---
*PR created automatically by Jules for task [9193551358648221452](https://jules.google.com/task/9193551358648221452) started by @dogi*

dogi:
This commit optimizes the `sortNews` and `setData` functions in `VoicesFragment`.
By converting `sortNews` to return a `List<News>`, it naturally filters out null elements once at the top of the processing block, allowing the `setData` consumers to reuse the processed list without incurring multiple redundant `.filterNotNull()` allocations and iterations.
The change preserves the `Trace.beginSection()` timing behavior and removes the unnecessary `toMutableList()` intermediate step.

---
*PR created automatically by Jules for task [9193551358648221452](https://jules.google.com/task/9193551358648221452) started by @dogi*


Okuro3499 (approved):

dogi:
**Review — rating: 90/100**

Clean, and the allocation count undersells it — the better change is the tightened return type.

Old path per refresh: `toMutableList()`, then `sortedWith()`, then `filterNotNull()` at **three** separate use sites (`setupVoicesAdapter`, `submitList`, `showNoData`) — five intermediate lists. New path: `filterNotNull()` then `sortedByDescending()`, and it sorts a shorter list since nulls are gone first. The `toMutableList()` was pure waste to begin with: `sortedWith` returns a new list from any `Iterable`, so the mutable copy was never needed.

Changing the return to `List<News>` is what makes the three call-site `filterNotNull()`s disappear rather than just relocating them. `setupVoicesAdapter` already required `List<News>`, so the fragment was filtering to satisfy a non-null boundary it could simply have honored at the source.

Ordering is provably unchanged: `sortDate` is `var sortDate: Long = 0` — non-null — so the old `news?.sortDate ?: 0L` was only ever guarding the null *element*, which `filterNotNull()` now removes. Kotlin's sorts are stable (TimSort), so filtering before rather than after sorting leaves the surviving items in the same relative order. `showNoData(…, sortedList.size, …)` reports the same count it did with the old `filterNotNull().size`, and `downloadResourcesForNews(List<News?>)` accepts the narrowed list fine.

The `Trace` section still wraps the work in try/finally, so the instrumentation is intact.

**Passing observation, not a request:** sorting the feed is arguably `NewsViewModel`'s job rather than the fragment's, and #15858/#15867 are moving that direction. Fine to leave here — the type tightening is the valuable part and doesn't depend on where the function lives.

No test, but `sortNews` is a private fragment method and the equivalence is provable from the types; I wouldn't hold the PR for it.

**Verdict:** mergeable.

---

PR #15870, opened 2026-08-20
sync: smoother transaction shared preferences managing (fixes #16005)

### LINKED ISSUE #16005: perf: switch SharedPreferences.commit() to apply()
Tracking issue for PR #15870.

Replaced blocking SharedPreferences.commit() calls with asynchronous
apply() to avoid synchronous disk I/O on the calling thread.

In SharedPrefManager.kt, collapsed clearPreferences() into a single
atomic `edit { ... }` block. In TransactionSyncManager.kt, replaced
the checkpoint removal commit() with apply() on the sync hot path.
Updated corresponding unit tests to verify apply().

---
*PR created automatically by Jules for task [12529758804219223291](https://jules.google.com/task/12529758804219223291) started by @dogi*

dogi:
Replaced blocking SharedPreferences.commit() calls with asynchronous
apply() to avoid synchronous disk I/O on the calling thread.

In SharedPrefManager.kt, collapsed clearPreferences() into a single
atomic `edit { ... }` block. In TransactionSyncManager.kt, replaced
the checkpoint removal commit() with apply() on the sync hot path.
Updated corresponding unit tests to verify apply().

---
*PR created automatically by Jules for task [12529758804219223291](https://jules.google.com/task/12529758804219223291) started by @dogi*


dogi:
**Review — rating: 82/100**

The `clearPreferences()` change is worth more than the "perf" framing — it fixes a real atomicity bug. The old code did:

```kotlin
editor.clear().apply()                 // write #1
for ((key, value) in tempStorage) { editor.putBoolean(key, value) }
editor.commit()                        // write #2, same editor
```

Two separate writes on a reused editor, with a genuine window in between where `FIRST_LAUNCH` and `MANUAL_CONFIG` were **absent** — any concurrent read or a process death in that gap would see defaults instead of the values being preserved. And `clearPreferences()` is exactly the kind of call (logout/reset) that runs when the app may be tearing down. Folding it into a single `pref.edit { clear(); … }` makes the clear and the restores one transaction, halves the writes, and stops depending on the subtle semantics of continuing to use an editor after `apply()`. Worth noting `putBoolean` in the same transaction survives the `clear()` regardless of call order, since `commitToMemory` applies the clear flag before the modifications — so the merged form is safe as written.

The test updates are right, including `just Runs` for the void `apply()` and dropping the now-unused `commit()` stub in `TransactionSyncManagerCheckpointTest`.

**Nit — the `TransactionSyncManager` change is the one I'd reconsider.** That `remove(checkpointKey)` runs on the sync's background thread, so `commit()` was never blocking the main thread and there's no perf win to collect. What it does cost is durability: `apply()` writes asynchronously, so a process kill right after a fully-completed sync can leave the checkpoint in place and make the next sync redo a table it already finished. That's redundant work rather than corruption, and it's self-correcting — but it's a trade with no upside on a background thread. `commit()` is the better default there; if consistency across the codebase is the goal, that's worth saying explicitly in the description.

Everything else here is a straight improvement.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #15871, opened 2026-08-20
all: smoother tts utils managing (fixes #16006)

### LINKED ISSUE #16006: perf: defer TTSManager TextToSpeech bind to first use
Tracking issue for PR #15871.

This branch implements the deferral of the `TextToSpeech` engine binding in `TTSManager`.

Previously, `TextToSpeech(context)` was constructed eagerly in the `@Singleton`'s `init` block, binding the system TTS service on the main thread anytime `TTSManager` was injected (e.g., when a resource viewer opens).

This update introduces a lazy `ensureTts(onInit: () -> Unit)` method, invoked only from `speak()`. The deferred `speak` action is passed as a callback so that it executes immediately after successful initialization. `stop()` now gracefully handles the uninitialized state, and getters do not trigger initialization.

Unit tests passed successfully.

---
*PR created automatically by Jules for task [287832783654215194](https://jules.google.com/task/287832783654215194) started by @dogi*

dogi:
This branch implements the deferral of the `TextToSpeech` engine binding in `TTSManager`.

Previously, `TextToSpeech(context)` was constructed eagerly in the `@Singleton`'s `init` block, binding the system TTS service on the main thread anytime `TTSManager` was injected (e.g., when a resource viewer opens).

This update introduces a lazy `ensureTts(onInit: () -> Unit)` method, invoked only from `speak()`. The deferred `speak` action is passed as a callback so that it executes immediately after successful initialization. `stop()` now gracefully handles the uninitialized state, and getters do not trigger initialization.

Unit tests passed successfully.

---
*PR created automatically by Jules for task [287832783654215194](https://jules.google.com/task/287832783654215194) started by @dogi*


dogi:
**Review — rating: 72/100**

The motivation is solid. `TTSManager` is `@Singleton` and injected in exactly one place — `ResourceViewerFragment` — so before this change, *opening any resource* constructed a `TextToSpeech` and bound the system TTS service, even though most users never press speak. Deferring that is a real saving, not a micro-optimization.

But the deferral introduces a silent-drop path at exactly the moment it matters most.

**Requests during the init window vanish.** Once construction has started but the callback hasn't fired, `tts != null` while `isInitialized == false`, so:

```kotlin
if (tts != null) {
    if (isInitialized) onInit()   // false → request discarded
    return
}
```

The *first* tap survives, because its lambda is the constructor's callback. Every tap during the engine's init latency (easily ~1s on a cold TTS service) does nothing at all. And this window is now the common case rather than a rarity: init used to begin when the fragment was injected, so by the time a user pressed speak the engine was usually ready; now init begins *on the press*. A user who taps twice — entirely natural when the first tap appears to do nothing — loses the second one.

The fix is small: hold the pending request and run it from the init callback.

```kotlin
private var pendingText: String? = null
// in speak(): if init is in flight, pendingText = text; return
// in the SUCCESS branch: pendingText?.let { tts?.speak(it, …) }; pendingText = null
```

**Init failure now latches permanently.** When `status != TextToSpeech.SUCCESS` the whole callback body is skipped, so `isInitialized` stays false while `tts` stays non-null — every later `speak()` takes the early return forever, with no retry and nothing surfaced to the user, holding a bound-but-dead engine for the app's lifetime. The old code ended up equally mute, but the deferred design makes the recovery obvious: null out `tts` on failure so the next `speak()` retries.

**Nit:** `isInitialized` should be `@Volatile`. It's written from the TTS init callback and read on the caller's thread, and in this version that read decides whether a request is dropped — a stale `false` silently loses an utterance. The race predates the PR, but it only became load-bearing here.

**Nit:** `if (tts != null) { tts?.stop(); … }` is doubly guarded; `tts?.let { it.stop(); _state.value = State.IDLE }` says the same thing once. Behavior is unchanged either way, since state already starts at `IDLE`.

**Passing observation:** there's no `shutdown()` anywhere, so the singleton holds the engine for the app lifetime once created. Out of scope here — and this PR makes "never created at all" the common case, which is the bigger win.

**Verdict:** needs another round — queue the pending utterance so first-use taps aren't dropped.

---

Okuro3499 (dismissed):
```
Problem: stop() only calls tts?.stop(); it never clears pendingText. Sequence: user taps "Read Aloud" → ensureTts creates the engine and sets pendingText, init is still in      
    flight → ResourceViewerFragment.onPause() (line 551) unconditionally calls ttsManager.stop() (since TTSManager is a singleton and isSpeaking is still false during the pending   
    window, this fires) → shortly after, the async onInit callback lands with SUCCESS, sees pendingText still set, and calls tts?.speak(...) anyway. The result: TTS audio starts    
    playing after the user paused/left the screen and explicitly asked it to stop.                                                                                                   
    Fix: Clear pendingText in stop(), e.g.:                                                                                                                                          
  fun stop() {                                                                                                                                                                       
      pendingText = null                                                                                                                                                             
      tts?.let {                                                                                                                                                                     
          it.stop()                                                                                                                                                                  
          _state.value = State.IDLE                                                                                                                                                  
      }                                                                                                                                                                              
  }   
```

dogi:
**Re-review — Rating: 79/100** (was 72)

All three items from last round are in: `pendingText` carries the request across the init window, `isInitialized` is `@Volatile`, and the failure branch nulls `tts` so the next `speak()` retries instead of latching mute forever. `stop()` also lost its double guard. The deferral itself remains a genuine win — `TTSManager` is a `@Singleton` injected only into `ResourceViewerFragment`, so this stops every resource open from binding the system TTS service.

Two small things in the new code, both a line or two:

**1. `stop()` doesn't cancel a pending utterance.** `pendingText` is only cleared inside the init callback, so:

```
tap speak  → engine construction starts, pendingText = "…"
tap stop   → tts?.let { … } — tts is non-null, stop() called on an engine that isn't speaking yet
init lands → pendingText?.let { tts?.speak(it, QUEUE_FLUSH, …) }   // speaks anyway
```

The user asked for silence and gets audio a beat later, with no way to prevent it. This is reachable precisely in the window the PR makes common — init now starts on the press rather than on injection, so "speak, change your mind, stop" is a normal sequence rather than a rare one. `stop()` should clear `pendingText`.

**2. The failure branch drops the engine without shutting it down.** `tts = null` discards the reference, but the `TextToSpeech` object has already bound the system TTS service; without `tts?.shutdown()` (or shutting down the instance the callback belongs to) that binding is never released, and since the next `speak()` now constructs a fresh one, repeated failures accumulate dead engines. Retrying is the right behavior — it just needs to release the old one first.

**Nit:** `pendingText` isn't `@Volatile` though `isInitialized` is. It's written on the caller's thread and read in the init callback, which is exactly the cross-thread pattern that justified the annotation on the other field.

**Nit:** `tts = TextToSpeech(context) { … }` assigns the field *after* the constructor returns, while the callback body reads `tts?.speak(...)`. Android dispatches `onInit` on the main thread, so a main-thread `speak()` is safe — the assignment completes first. A background-thread call is not: the callback could run with `tts` still null and drop the utterance it was queued to deliver. Worth documenting that `speak()` is main-thread-only, or capturing the instance in a local that the callback closes over.

**Verdict:** needs another round — clearing `pendingText` in `stop()` and calling `shutdown()` on the failed engine is about three lines, and then this is a clean win.

---

dogi:
**Re-review — Rating: 88/100** (was 79)

All four points are addressed, including both nits:

- **`stop()` clears `pendingText`** — so "speak, change your mind, stop" no longer produces audio a beat later.
- **The failure branch shuts the engine down** — `instance?.shutdown()` before `tts = null`, so repeated init failures no longer leak bound TTS service connections while still allowing the next `speak()` to retry.
- **`pendingText` is `@Volatile`**, matching `isInitialized`.
- **The callback closes over a local** (`var instance`) instead of reading the `tts` field, which was the fix for the assignment-ordering gap I flagged.

The deferral itself still reads well: `tts != null` short-circuits, an in-flight init parks the text in `pendingText`, and the init callback flushes it with `QUEUE_FLUSH` — matching the semantics `speak()` had before.

**Nit — the local-capture fix narrows the ordering window but doesn't close it, and it's still main-thread-only.** `instance` is a captured `var` (boxed in a non-volatile `Ref`), assigned after the constructor returns, and `tts = instance` runs after that. On the main thread this is airtight, because `onInit` is posted to the main looper and can't run until `ensureTts` returns. Off the main thread it isn't: the callback could see `instance == null`, skip `setOnUtteranceProgressListener`, clear `pendingText`, and then `tts = instance` publishes an engine with no progress listener — `speak()` works but `_state` never leaves `IDLE`, so `isSpeaking` is wrong forever. Worse in the failure branch: `tts = null` inside the callback can be overwritten by the trailing `tts = instance`, leaving a shut-down engine parked in the field and every later `speak()` silently doing nothing.

Both are unreachable today — `TTSManager` is injected only into `ResourceViewerFragment` and driven from UI callbacks — so this is a documentation ask, not a code one: a `// speak()/stop() must be called from the main thread` comment on the class, or an explicit main-thread assertion, records the invariant the design now depends on.

**Nit:** `stop()` moved `_state.value = State.IDLE` inside `tts?.let { }`, so with no engine the state isn't reset. `SPEAKING` can only be set by a progress listener on a live engine, so it can't currently be stranded — but the unconditional reset was the more obviously-correct form.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #15872, opened 2026-08-20
courses: smoother progress repository submissions mapping (fixes #16007)

### LINKED ISSUE #16007: Optimize submissionMap by indexing exam IDs once
Tracking issue for PR #15872.

Optimize submissionMap by replacing two linear scans inside a nested loop with a single map lookup. Preserves first-index behavior.

---
*PR created automatically by Jules for task [12201935214581396969](https://jules.google.com/task/12201935214581396969) started by @dogi*

dogi:
Optimize submissionMap by replacing two linear scans inside a nested loop with a single map lookup. Preserves first-index behavior.

---
*PR created automatically by Jules for task [12201935214581396969](https://jules.google.com/task/12201935214581396969) started by @dogi*


dogi:
**Review — rating: 90/100**

Correct, and it gets right the one detail that makes this refactor easy to break.

The old inner loop did **two** linear scans per answer — `examIds.contains(question.examId)` and then `examIds.indexOf(question.examId)` — so the cost was O(answers × examIds). Indexing once up front makes it O(examIds + answers).

**The subtle part is duplicate exam ids, and the `containsKey` guard is exactly why this is safe.** `indexOf` returns the *first* matching index, and the loop preserves that by only writing a key the first time it's seen. The tempting one-liner —

```kotlin
examIds.withIndex().associate { (i, id) -> id to i.toString() }
```

— keeps the **last** index instead, which would silently reassign mistakes to a different exam slot whenever `examIds` contains a repeat. The explicit loop is the right call here.

Equivalence otherwise holds cleanly: `examIndexMap` holds exactly the distinct values of `examIds`, so `examIndexMap[examId] != null` is the same predicate as `examIds.contains(examId)`, and the value is the same first index. `examId` is non-null (the `HashMap<String, String>` lookup wouldn't compile otherwise), so there's no null-key edge to worry about.

One secondary win the description doesn't claim: the old code called `.toString()` on the index **per answer**, allocating a String each time. Now it's one per distinct exam id.

**Nit:** `putIfAbsent(id, index.toString())` expresses "first wins" more directly than `if (!examIndexMap.containsKey(id))`, if you want it tighter.

No test added, but the equivalence is provable from the code — same membership predicate, same first-index semantics — so I wouldn't hold it for that.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #15874, opened 2026-08-20
teams: smoother repository dao querying (fixes #16009)

### LINKED ISSUE #16009: Refactor: Push teams-table Flow filters into SQL
Tracking issue for PR #15874.

Resolves issue 85 by pushing the Kotlin Flow filters in `queryTransactions` and `getReportsFlow` down into Room SQL via a new `observeByTeamIdAndDocType` query. This prevents unrelated updates in the `teams` table from triggering re-emissions for these specific doc types. Also applies `distinctUntilChanged` using an explicit ID/Rev comparator to handle Room entity instantiation correctly.

---
*PR created automatically by Jules for task [16176007988889566505](https://jules.google.com/task/16176007988889566505) started by @dogi*

dogi:
Resolves issue 85 by pushing the Kotlin Flow filters in `queryTransactions` and `getReportsFlow` down into Room SQL via a new `observeByTeamIdAndDocType` query. This prevents unrelated updates in the `teams` table from triggering re-emissions for these specific doc types. Also applies `distinctUntilChanged` using an explicit ID/Rev comparator to handle Room entity instantiation correctly.

---
*PR created automatically by Jules for task [16176007988889566505](https://jules.google.com/task/16176007988889566505) started by @dogi*


dogi:
**Review — rating: 86/100**

Well-targeted, and the two halves of the change reinforce each other in a way the title doesn't convey.

**The SQL push-down is a bigger win than "filter in SQL" sounds.** The `teams` table is docType-discriminated — it holds teams, memberships, transactions, reports and tasks in one table — so `observeAll()` was materializing *every* row, with the Gson `@TypeConverters` running per row, on every write to that table, just to keep the handful of report rows for one team. The new predicate is index-supported (`MyTeam` declares `Index("teamId")` and `Index("docType")`), so SQLite does the narrowing and Room only builds the matching objects.

**The `distinctUntilChanged` is load-bearing, and the explicit comparator is required.** Worth spelling out because it's easy to get wrong: Room's `InvalidationTracker` is table-level, so the query still re-runs on *any* `teams` write — the emission count doesn't drop, only the row count does. The comparator is what actually suppresses the downstream sort/map/UI-diff when the filtered slice is unchanged. And since `MyTeam` is an `open class` rather than a `data class`, `equals` is identity and a bare `distinctUntilChanged()` would have been completely inert (that's the trap #15869 fell into). Comparing `_id`/`_rev` explicitly is the right call.

`flowOn(dispatcherProvider.default)` is also correctly placed — the filter/sort/comparator all ran on the collector's thread before.

**One thing to confirm: `_rev` only moves when the server says so.** `_rev` is populated from the CouchDB document (`team._rev = JsonUtils.getString("_rev", doc)`), so the comparator assumes every meaningful mutation either changes the list size or bumps `_rev`. A local archive is fine — the size changes. But a local edit to a *displayed* field on an existing row (a transaction's `amount`, `description`, or `date`) that doesn't bump `_rev` would be suppressed, leaving stale values on screen — and for `date` also a stale sort order, since it's the sort key. If in-app finance edits write through without a `_rev` change, the comparator needs those fields in it; if every edit round-trips through sync first, this is fine as written.

**Nit:** the same four-line comparator is duplicated in both flows. A small private `Flow<List<MyTeam>>.distinctByIdAndRev()` extension would keep them from drifting — `TeamsRepositoryImpl` is already the largest file in the repo.

**Nit:** `getReportsFlow` is still `suspend` while only returning a cold `Flow` — the same needless suspend that #15869 removed from `getOfflineLogins`, propagated through `TeamsFinancesRepository` and `EnterprisesViewModel`. Not introduced here, but this is the natural moment.

**Nit:** no test. The SQL/Kotlin filter equivalence and the comparator are both cheap to pin in `TeamsRepositoryImplTest`, which already mocks `teamDao`.

**Verdict:** mergeable — please sanity-check the `_rev` question above.

---

Okuro3499 (approved):

PR #15876, opened 2026-08-20
all: smoother notifications view modelling (fixes #16011)

### LINKED ISSUE #16011: Refactor NotificationsViewModel loadNotifications to classify types once
Tracking issue for PR #15876.

Refactor NotificationsViewModel loadNotifications and buildGroupedList to avoid multiple list traversals and reallocations.

---
*PR created automatically by Jules for task [8040347339579517484](https://jules.google.com/task/8040347339579517484) started by @dogi*

dogi:
Refactor NotificationsViewModel loadNotifications and buildGroupedList to avoid multiple list traversals and reallocations.

---
*PR created automatically by Jules for task [8040347339579517484](https://jules.google.com/task/8040347339579517484) started by @dogi*


dogi:
**Review — rating: 87/100**

Behavior-preserving and correct. I checked all four derived lists against the originals: each old `filter { it.type.lowercase() == X && … }` becomes the corresponding group plus the non-type half of its predicate, and `groupBy` returns a `LinkedHashMap` with encounter order preserved inside each group — so `taskIds`, `taskTitles`, `joinRequestIds` and `joinRequestsWithoutRelatedId` come out with the same contents *and* the same order as before. That matters here because these feed batch repository lookups.

Being precise about where the win is: the four filter passes each called `.lowercase()` on every notification, so this drops 4N String allocations to N. It does allocate one list per distinct type present, including types nothing looks up, so it's not strictly fewer allocations overall — but `lowercase()` was the dominant cost and the result reads far better than four near-identical filters.

Hoisting `typeOrder` is the other real win, and it's more than cosmetic: `buildNotificationItems` runs on every list rebuild — including selection toggles — so that 7-element `listOf` was being allocated on every one.

**Nit worth doing before merge: `KNOWN_TYPES` and `TYPE_ORDER` are now the same seven strings in the same order, declared twice.** That's a live trap — add an eighth type to one and forget the other, and it either gets folded into the "Other" group or sorts to the end, with nothing failing loudly. One source of truth fixes it:

```kotlin
val TYPE_ORDER = listOf("join_request", "team_join", "task", "chat", "voice_reply", "resource", "storage")
val KNOWN_TYPES = TYPE_ORDER.toSet()
```

`toSet()` gives a `LinkedHashSet`, so lookup semantics at both `KNOWN_TYPES` call sites (lines 224 and 249) are unchanged.

**Nit:** `NotificationsViewModelTest` already exists, and the ordering invariant this PR now depends on (grouped types render in `TYPE_ORDER`, unknown types last) would be a cheap regression test — especially useful if you take the derivation above.

**Verdict:** mergeable.

---

Okuro3499 (approved):

Okuro3499 (approved):

PR #15877, opened 2026-08-20
courses: smoother layout handling (fixes #16012)

### LINKED ISSUE #16012: Optimization: stop rebuilding CoursesFragment LayoutManager on every toggle
Tracking issue for PR #15877.

Optimization to prevent rebuilding the LayoutManager in CoursesFragment on every toggle.
- Added cached fields for grid/list toggle buttons to avoid repeated `findViewById`.
- Updated `applyRecyclerLayoutManager` to reuse existing `LayoutManager` instance if the type matches the target mode, updating `spanCount` for GridLayoutManager. This prevents dropping the recycled view pool and forces fewer re-layouts.

---
*PR created automatically by Jules for task [9720656326509254263](https://jules.google.com/task/9720656326509254263) started by @dogi*

dogi:
Optimization to prevent rebuilding the LayoutManager in CoursesFragment on every toggle.
- Added cached fields for grid/list toggle buttons to avoid repeated `findViewById`.
- Updated `applyRecyclerLayoutManager` to reuse existing `LayoutManager` instance if the type matches the target mode, updating `spanCount` for GridLayoutManager. This prevents dropping the recycled view pool and forces fewer re-layouts.

---
*PR created automatically by Jules for task [9720656326509254263](https://jules.google.com/task/9720656326509254263) started by @dogi*


dogi:
**Review — rating: 78/100**

Reviewed back to back with #15884, which applies the same optimization to `ResourcesFragment`. The core change is right in both, but the view handling differs, and this one has a leak that #15884 avoids.

**The optimization itself is correct and worth having.** Assigning `recyclerView.layoutManager` tears down every ViewHolder, drops scroll position, and forces a full re-layout, so doing it on a grid→grid span change was pure waste. Reusing the existing manager and setting `spanCount` is the right move — and it works, because `GridLayoutManager.setSpanCount()` invalidates the span-index cache and calls `requestLayout()` itself, while early-returning when the count is unchanged, so a no-op toggle costs nothing.

You also got the subtle part right: `GridLayoutManager` **extends** `LinearLayoutManager`, so a bare `is LinearLayoutManager` check would have matched a grid manager and skipped the switch to list mode. The `!is GridLayoutManager` guard is what makes it correct.

**The problem: `toggleGridButton` and `toggleListButton` are `lateinit var` View references that are never cleared.** `onDestroyView` already exists at line 583 and does cleanup (`filterController.detach()`, `adapterCourses.setListener(null)`) — but nothing nulls these fields. Because `openCallFragment` adds fragments to the back stack, a `CoursesFragment` instance routinely outlives its view, and each of these fields then retains the whole destroyed view hierarchy through the button's parent chain for as long as the instance is on the stack.

`ResourcesFragment` in #15884 handles exactly this: nullable `var`s plus explicit clearing in `onDestroyView`. Same fix here — make them `ImageButton?`, use `?.` at the call sites, and null them alongside the existing cleanup.

Worth noting this isn't a new pattern in the file — `orderByDate` and `orderByTitle` are already un-nulled `lateinit` Buttons — but that's an argument for clearing all four while you're in `onDestroyView`, not for adding two more.

**Nit:** the list branch uses an empty `if` with an explanatory comment. #15884's inverted form says the same thing without the empty body, and consistency between the two fragments would be nice:

```kotlin
if (currentLayoutManager !is LinearLayoutManager || currentLayoutManager is GridLayoutManager) {
    recyclerView.layoutManager = LinearLayoutManager(requireContext())
}
```

**Verdict:** needs another round — clear the cached button references in `onDestroyView`, then this is straightforwardly mergeable.

---

Okuro3499 (approved):

dogi:
**Re-review — Rating: 90/100** (was 78)

The leak is fixed, and fixed more thoroughly than I asked: `toggleGridButton` and `toggleListButton` are nullable and cleared in `onDestroyView`, and you took the same treatment to `orderByDate`/`orderByTitle` — which is the right call, since those were the pre-existing un-nulled `lateinit` Views in the same file. All four are nulled next to the existing `adapterCourses.setListener(null)` cleanup, so a `CoursesFragment` sitting on the back stack no longer retains a destroyed view hierarchy through any of them.

Two things I checked rather than assumed:

- **Assignment happens before first use.** The two `findViewById` calls are added in `onViewCreated` *above* `additionalSetup()` / `setupViewModeToggle()`, and `setupViewModeToggle()` → `updateToggleUi()` is the first reader. So the cached fields are populated before any listener is attached or tint applied — no silently-skipped `?.` on first render.
- **`enableSortButtons` got simpler for free.** The `::orderByDate.isInitialized` guards became `?.`, which expresses the same "may not exist yet" condition without the reflection-flavored check.

The optimization itself still reads correctly, including the subtle part: the list branch now uses `!is LinearLayoutManager || is GridLayoutManager`, which is required because `GridLayoutManager` extends `LinearLayoutManager` — a bare `!is LinearLayoutManager` would skip the switch to list mode. Adopting #15884's inverted form also removed the empty `if` body, so the two fragments now read the same way.

**Verdict:** mergeable.

---

dogi:
**Re-review — Rating: 91/100** (was 90)

The layout-manager reuse is correct, including the inheritance trap it has to dodge. `GridLayoutManager` *extends* `LinearLayoutManager`, so a naive `if (current !is LinearLayoutManager)` would treat a grid as already-a-list and leave the grid in place on toggle; the second clause catches exactly that:

```kotlin
if (currentLayoutManager !is LinearLayoutManager || currentLayoutManager is GridLayoutManager) {
    recyclerView.layoutManager = LinearLayoutManager(requireContext())
}
```

And the grid branch mutates `spanCount` on the existing manager rather than replacing it, which is the actual win — `GridLayoutManager.setSpanCount` early-returns when the count is unchanged and otherwise invalidates the span cache and calls `requestLayout()`, so the re-layout still happens without discarding view holders and scroll state.

The view-reference work that came with it is a genuine improvement in its own right: `orderByDate`/`orderByTitle` move from `lateinit` to nullable (so `enableSortButtons` drops its `::isInitialized` guards), `toggle_grid`/`toggle_list` are resolved once in `onViewCreated` instead of `findViewById` on every `updateToggleUi` call, and all four are nulled in `onDestroyView` before `super`. That last part is what keeps a fragment from holding destroyed views.

**Nit:** the list-branch condition is correct but reads as a puzzle. `if (currentLayoutManager?.javaClass != LinearLayoutManager::class.java)` expresses "not exactly a plain LinearLayoutManager" directly, or a short comment noting that `GridLayoutManager` is a `LinearLayoutManager` subclass would save the next reader the derivation.

**Verdict:** mergeable.

---

PR #15842, opened 2026-08-19
sync: smoother upload shelfing (fixes #15983)

dogi:
Refactored `UploadToShelfService` to remove nested `appScope.launch` blocks in `uploadToShelf` and `uploadSingleUserToShelf`. Changed these methods to `private suspend fun`. This fixes broken structured concurrency and prevents double dispatching, allowing exceptions to correctly propagate to the outer coroutine's try/catch blocks.

---
*PR created automatically by Jules for task [210971732348277734](https://jules.google.com/task/210971732348277734) started by @dogi*

## Summary by CodeRabbit

* **Improvements**
  * Improved the reliability and consistency of uploading user data to Shelf.
  * Upload progress and completion results are now handled more predictably within the active upload process.

PR #15878, opened 2026-08-20
courses: smoother submissions repository exams starting (fixes #16141)

### LINKED ISSUE #16141: Refactor exam-start orchestration out of ExamTakingFragment
Tracking issue for PR #15878.

Refactored the exam-start orchestration logic (fetching, deleting, and creating pending submissions) out of `ExamTakingFragment` and into a new `startExamSession` method in `SubmissionsRepository`. Also moved the pending-fetch logic for `saveExamAnswer` directly into `SubmissionsRepositoryImpl` to reduce UI coupling to database transactions.

- Added `startExamSession` to `SubmissionsRepository` and `SubmissionsRepositoryImpl`.
- Updated `ExamTakingFragment` to utilize `startExamSession`.
- Modified `ExamAnswerData` to include a `userId` field to support internal fetching within the repository.
- Removed manual `getSubmissionsByParentId` logic from `updateAnsDb` within `ExamTakingFragment` and integrated the fallback natively within `SubmissionsRepositoryImpl.saveExamAnswer`.

---
*PR created automatically by Jules for task [8917027368947886249](https://jules.google.com/task/8917027368947886249) started by @dogi*

## Summary by CodeRabbit

- **New Features**
  - Improved exam session startup, including resuming pending sessions or restarting them when requested.
  - Added more reliable handling for surveys and team-based sessions.
  - Saved exam answers are now associated with the current user.

- **Bug Fixes**
  - Improved pending submission matching and handling of stale submissions.
  - Added retry handling and clearer error reporting when session creation fails.

dogi:
Refactored the exam-start orchestration logic (fetching, deleting, and creating pending submissions) out of `ExamTakingFragment` and into a new `startExamSession` method in `SubmissionsRepository`. Also moved the pending-fetch logic for `saveExamAnswer` directly into `SubmissionsRepositoryImpl` to reduce UI coupling to database transactions.

- Added `startExamSession` to `SubmissionsRepository` and `SubmissionsRepositoryImpl`.
- Updated `ExamTakingFragment` to utilize `startExamSession`.
- Modified `ExamAnswerData` to include a `userId` field to support internal fetching within the repository.
- Removed manual `getSubmissionsByParentId` logic from `updateAnsDb` within `ExamTakingFragment` and integrated the fallback natively within `SubmissionsRepositoryImpl.saveExamAnswer`.

---
*PR created automatically by Jules for task [8917027368947886249](https://jules.google.com/task/8917027368947886249) started by @dogi*

## Summary by CodeRabbit

- **New Features**
  - Improved exam session startup, including resuming pending sessions or restarting them when requested.
  - Added more reliable handling for surveys and team-based sessions.
  - Saved exam answers are now associated with the current user.

- **Bug Fixes**
  - Improved pending submission matching and handling of stale submissions.
  - Added retry handling and clearer error reporting when session creation fails.
  - Improved restoration of previously saved answers when resuming an exam.


dogi:
**Review — rating: 55/100**

The direction is right — collapsing the nested `viewLifecycleOwner.lifecycleScope.launch(io)` inside an already-suspending function and moving exam-start orchestration behind the repository are both real improvements (that nested-launch pattern is the same one #15842 removed from `UploadToShelfService`). But there are correctness problems in a path where the cost of being wrong is lost or misfiled learner submissions, so I don't think this can go in as-is.

**1. Survey submissions are now written with `type = "exam"`.** In the survey branch you build the request with the type hardcoded:

```kotlin
val request = CreateExamSubmissionRequest(user?.id, user?.dob, user?.gender, currentExam, "exam", null)
```

where the old code passed the fragment's actual `type`. `createExamSubmission` persists it directly — `this.type = type` — so survey submissions get stored as exams. That changes how they're classified downstream (`getExamSubmissionsByUser` filters `WHERE type = 'exam'`, and the survey listings filter the other way), so surveys would start showing up as exams and drop out of survey lists.

The root cause is the gate inside `startExamSession`: because it only creates when `request.type == "exam"`, the survey path has to misreport its type to reach the create. Better to make the intent an explicit parameter — e.g. `startExamSession(..., recreate: Boolean)` — and keep passing the real type through to the request.

**2. `sub` is now overwritten unconditionally, losing a deliberate guard.** The block you removed was:

```kotlin
if (sub == null) {
    val submissions = submissionsRepository.getSubmissionsByParentId(parentId, user?.id, "pending")
    sub = submissions.firstOrNull()
}
```

The `if (sub == null)` preserved an already-set submission. Now `sub = submissionsRepository.startExamSession(...)` runs for both types before the type check, so a pre-set `sub` (a resumed session) is replaced by the lookup result — and if the lookup finds nothing, `sub` becomes **null**. On the survey path this also means `startExamSession` is called twice: once at the top with the real type (lookup only, clobbering `sub`) and again in the else branch with `"exam"`.

**3. The retry loop can hand back a submission that no longer exists.** If `deleteExamSubmissions` succeeds and `createExamSubmission` then throws on all three attempts, `submission` still holds the pending row found before the loop — which the delete has since removed. That row's id then flows into `saveExamAnswer`. There's also no way for the caller to tell "no pending submission existed" from "three failed attempts", since both return `null`-ish results. Three immediate retries of a local Room write is also unlikely to change the outcome, and the original code didn't retry at all — if retrying is intentional it deserves a comment, and failure needs to be distinguishable.

**4. `parentId` is computed on every answer save.** In `saveExamAnswer`, `val parentId = question.examId?.let { examDao.getById(it) … }` is evaluated eagerly, before the elvis chain that usually doesn't need it. That's an extra `examDao.getById` on every answered question, where the old code only did the parent lookup in the fragment when `sub` was null. Move it inside the branch that consumes it.

**Nit:** no tests, and `SubmissionsRepositoryImplTest` already exists. `startExamSession` is new repository logic gating exam creation — the delete-then-create path, the retry-exhaustion path, and the survey (no-create) path are the three cases I'd want pinned before this lands.

Worth keeping: the `saveExamAnswer` fallback consolidation is a good idea — moving the "find the pending submission by parent" fallback out of the fragment and into the repository is the right home for it.

**Verdict:** needs another round — items 1 and 2 are behavioral regressions in submission handling.

---

Okuro3499 (dismissed):
```
Problem: For team exams, the old code (master) always called createExamSubmission unconditionally whenever sub == null || isTeam — guaranteeing a fresh, team-tagged submission    
(with teamObject/membershipDoc set) every time a team exam starts. The new code routes this through startExamSession(..., recreate = false), which first looks up any existing     
pending submission for parentId/userId and returns it if found — with no team awareness. Since the fragment's own top-of-function check (ExamTakingFragment.kt:104-109) already    
populates sub from that same lookup for non-"exam" types, this means: whenever isTeam == true and a pending submission already exists at that parentId/userId (e.g. from a prior   
individual attempt), the fragment will now silently reuse that non-team submission instead of creating a fresh, team-attributed one. This can misattribute a submission (missing   
team context) or bleed progress between an individual attempt and a team session. 
```

dogi:
**Re-review — Rating: 76/100** (was 55)

Both regressions are fixed, and fixed at the root rather than papered over:

- **Item 1 (survey submissions written as `type = "exam"`)** — gone. `startExamSession(..., recreate: Boolean)` makes the intent an explicit parameter, so both branches now pass the fragment's real `type` into `CreateExamSubmissionRequest`. That was the one that would have misfiled learner data.
- **Item 2 (`sub` clobbered, `startExamSession` called twice on the survey path)** — the top-of-function pre-fetch is now `if (type != "exam" && sub == null)`, so the `sub == null` guard is preserved for surveys and skipped entirely for exams (which always recreate). No double call.
- **Item 3 (retry loop returning a deleted submission)** — resolved. Exhausting the retries now throws `IllegalStateException` with the last cause, so a phantom row can't reach `saveExamAnswer`, and "no pending submission" is distinguishable from "three failures". Returning non-null `Submission` instead of a nullable is the right signature for that.
- **Item 4 (`parentId` computed on every answer save)** — now inside the `?: run { … }`, so the extra `examDao.getById` only happens on the fallback path.
- **Tests** — four cases covering recreate/no-recreate × found/created, plus retry exhaustion.

Two things from the new code, one of which I'd fix before merging:

**The exception this now throws is unhandled at both call sites.** The exam branch used to wrap its DB work in `try { … } catch (e: Exception) { e.printStackTrace() }`; this version removes that and calls `startExamSession` bare inside `viewLifecycleOwner.lifecycleScope.launch(dispatcherProvider.io)`. `initializeExamData()` doesn't wrap `initExam()` in a try either (the only try/catch in that region of the file is the one this PR deletes). So three failed writes now take the app down instead of leaving a broken exam screen. Throwing is the right design — it just needs a catch that tells the learner the exam couldn't be started, ideally with the retry budget surfaced rather than silent.

**A behavior change on the team-survey path worth confirming is deliberate.** The `else` branch is still gated on `sub == null || isTeam`, but the body changed meaning:

```kotlin
// before: always a NEW submission
sub = submissionsRepository.createExamSubmission(request)
// after: reuse a pending one if one exists, else create
sub = submissionsRepository.startExamSession(currentExam.id, parentId, user?.id, request, recreate = false)
```

When `sub == null` the two agree — the pre-fetch above already established there's no pending row, so `startExamSession` creates one. But with `isTeam` and `sub != null`, `sub` came from that same pending lookup, so `startExamSession` hands back the same row where the old code created a fresh submission. Reuse looks like the better behavior (it stops team surveys accumulating duplicate pending rows), so I'd keep it — but it's a silent change in submission identity, so say in the PR description that it's intended.

**Nit:** the retry-exhaustion test asserts inside a `catch`, so it passes silently if nothing is thrown. `assertThrows<IllegalStateException> { … }` (or a `fail()` after the call) makes the assertion real.

**Nit:** the retry comment explains *why* ("transient SQLite constraints during rapid operations") but three immediate attempts with no delay won't outlast a real constraint violation. If the failure mode you have in mind is a race, a small backoff would help; if it isn't, one attempt and a clear error may be honest enough.

**Verdict:** needs another round — catch the new exception at both call sites, then this is close.

---

dogi:

Both call sites handle the exception now, and the placement is better than the minimum I asked for.

The exam branch wraps the whole `startExamSession` + `withContext(main)` block, so a failure surfaces to the learner instead of taking the app down. The survey branch does the same — and importantly it moved `startExam(questions?.get(currentIndex))` and `updateNavButtons()` *inside* the try, so an exhausted retry no longer proceeds to render an exam against a null submission. That's the failure mode a narrower try/catch would have left open.

The test nit is fixed properly too: `assertThrows(IllegalStateException::class.java) { runBlocking { … } }` replaces the assert-inside-catch, so the retry-exhaustion test can no longer pass by throwing nothing. All four `startExamSession` cases are still covered.

Together with the earlier round this now handles the whole path: survey submissions keep their real `type`, `sub` isn't clobbered, a phantom deleted submission can't reach `saveExamAnswer`, and the failure is both distinguishable and visible.

**The new user-facing strings are hardcoded English.** `Utilities.toast(requireContext(), "Failed to start exam session. Please retry.")` appears twice as a literal, in an app that ships `ar`, `es`, `fr`, `ne` and `so`. Everything else in this file goes through `getString(...)` — this needs a `<string name="failed_to_start_exam_session">` and the five translations. It's the one thing I'd fix before merging, since it's a string a learner sees at the moment their exam won't start.

**Nit:** `Utilities.toast(requireContext(), …)` — `requireContext()` throws if the fragment has detached. The `viewLifecycleOwner.lifecycleScope` scoping means the continuation shouldn't resume after the view is gone, so this is defensible, but `Utilities.toast(activity, …)` is the nullable-safe form used elsewhere in this codebase (`ResourceDetailFragment`, `FeedbackFragment`) and costs nothing.

**Nit:** `import org.ole.planet.myplanet.utils.Utilities` is inserted between `com.google.gson.JsonObject` and `dagger.hilt.android.AndroidEntryPoint`, in the middle of the third-party import block.

**Nit (carried over):** the team-survey path still reuses a pending submission where the old `sub == null || isTeam` branch always created a fresh one. I think reuse is the better behavior — it stops team surveys accumulating duplicate pending rows — but it's a silent change in submission identity and still isn't mentioned in the description.

**Nit (carried over):** three immediate retries with no backoff, against a comment that attributes the failure to "transient SQLite constraints during rapid operations". If that's a race, a short delay would help; if it isn't, one attempt and a clear error is more honest.

**Verdict:** mergeable — please move those two toasts into string resources.

---

Okuro3499 (dismissed):
```
for team exams, master always called createExamSubmission unconditionally whenever sub == null || isTeam — guaranteeing a fresh, team-tagged submission                   
  (teamObject/membershipDoc set) every time a team exam starts. This branch now calls startExamSession(..., recreate = false) instead, which first looks up any existing pending     
  submission for parentId/userId and returns it if found, with no awareness of isTeam. Since the fragment's own top-of-function check (ExamTakingFragment.kt:104-109) already        
  populates sub from that exact same lookup for non-"exam" types, the sequence is:                                                                                                   
                                                                                                                                                                                     
  1. sub gets set from a prior pending submission (possibly a solo attempt) at the top of initializeExamData().                                                                      
  2. isTeam == true routes into the sub == null || isTeam branch anyway.                                                                                                             
  3. startExamSession(recreate = false) re-fetches and returns that same pending submission.                                                                                         
                                                                                                                                                                                     
  So a team exam session silently reuses a stale, non-team-tagged submission instead of creating a fresh, correctly-attributed one — a data integrity regression, and deterministic  
  (not just a rare race) whenever a pending submission already exists at that parentId/userId. 
```

dogi:
**Re-review — Rating: 88/100** (was 88)

Head is unchanged since my last round, so the assessment stands: survey submissions keep their real `type` via the explicit `recreate` parameter, the `sub == null` pre-fetch guard is preserved for surveys and skipped for exams, retry exhaustion throws `IllegalStateException` rather than returning a phantom deleted submission, `parentId` is only recomputed on the fallback path, and both call sites wrap the whole start-plus-render block so an exhausted retry can't render an exam against a null submission. Five `startExamSession` tests cover recreate × deleteStale × found/created plus the throw.

The one item I filed as a nit last round is still open, and I'd rather restate it plainly than let it disappear: `Utilities.toast(requireContext(), "Failed to start exam session. Please retry.")` appears twice as a bare English literal in a file where every other user-facing string goes through `getString(...)`, in an app shipping `ar`, `es`, `fr`, `ne` and `so`. It's a two-line fix — one `<string>` entry per locale file and two call sites — and it's the difference between an error path that works for every user and one that works for English speakers. I called it a nit and I'm not going to reclassify it as a blocker after the fact, but it is a defect that reaches users, not a style preference.

**Nit:** `import org.ole.planet.myplanet.utils.Utilities` was inserted between `com.google.gson.JsonObject` and `dagger.hilt.android.AndroidEntryPoint`, which breaks the file's import ordering — `org.ole.*` belongs after the `kotlin`/`kotlinx` block.

**Verdict:** ready, as before; please take the string extraction with it.

---

Okuro3499 (approved):

PR #15880, opened 2026-08-20
perf(CameraUtils): move JPEG byte copy off main thread

dogi:
Moves the `buffer.get(bytes)` call from the `ImageReader.OnImageAvailableListener`
(which executes on the main thread) into a coroutine launched on `dispatcherProvider.io`.
This prevents stalling the UI thread during image captures. Also ensures `image.close()`
happens safely within the same coroutine after the buffer is copied.

---
*PR created automatically by Jules for task [3569912595753770855](https://jules.google.com/task/3569912595753770855) started by @dogi*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved image capture reliability by moving image processing and saving to background execution.
  * Reduced the risk of UI slowdowns or freezes when capturing and saving photos.

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/utils/CameraUtils.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/utils/CameraUtils.kt
@@ -52,11 +52,11 @@ object CameraUtils {
         imageReader?.setOnImageAvailableListener({ reader ->
             val image = reader.acquireLatestImage()
             if (image != null) {
-                val buffer = image.planes[0].buffer
-                val bytes = ByteArray(buffer.capacity())
-                buffer.get(bytes)
-                image.close()
-                scope.launch {
+                scope.launch(dispatcherProvider.io) {
+                    val buffer = image.planes[0].buffer
+                    val bytes = ByteArray(buffer.capacity())
+                    buffer.get(bytes)
+                    image.close()
                     savePicture(bytes, callback, dispatcherProvider)
                 }
             }


dogi:
**Review — rating: 66/100**

The premise checks out — the listener really is on the main thread (`setOnImageAvailableListener({ … }, mainHandler)` at line 63), so the copy was main-thread work. But the fix trades a negligible cost for a resource-lifetime hazard, and there's a standard fix that avoids both.

**`image.close()` is now conditional on the coroutine running.** Moving it inside `scope.launch(dispatcherProvider.io)` means that if `scope` is cancelled between the listener firing and the coroutine being dispatched, the body never executes and the `Image` is never closed. That's a realistic window: a capture is often the last thing that happens before the caller navigates away and its scope dies. The old code closed synchronously in the listener, so cancellation couldn't strand it.

**And the reader has exactly one buffer.** Line 51 creates it with `maxImages = 1`:

```kotlin
imageReader = ImageReader.newInstance(IMAGE_WIDTH, IMAGE_HEIGHT, ImageFormat.JPEG, 1)
```

Returning from the listener with the image still open holds that single slot. Today you're saved by the capture being one-shot — `capturePhoto` calls `stopRepeating()`/`abortCaptures()` and issues a single `capture(...)`, and the repeating preview targets the SurfaceTexture rather than the reader — so a second frame shouldn't arrive. But the invariant "the only acquired image is closed before the listener returns" is what makes that safe, and this removes it. If anyone later adds a burst or points a repeating request at the reader, `acquireLatestImage()` throws `IllegalStateException: maxImages (1) has already been acquired` **on the main thread**, inside the listener.

**What's being saved is very small.** A 640×480 JPEG is tens of kilobytes; `ByteArray(capacity)` plus `buffer.get(bytes)` on that is microseconds. It's not a frame-drop source.

**The idiomatic fix is the handler, not the coroutine.** Register the listener on a background `HandlerThread` and keep acquire → copy → close synchronous:

```kotlin
private val cameraThread = HandlerThread("CameraUtils").apply { start() }
private val cameraHandler = Handler(cameraThread.looper)
// …
imageReader?.setOnImageAvailableListener({ reader -> … }, cameraHandler)
```

That gets the copy off the main thread *and* keeps the close prompt and unconditional. It would also move the two `setRepeatingRequest(..., mainHandler)` calls (lines 161 and 184) off the main thread, which are doing more main-thread camera work than the byte copy ever was.

**Two pre-existing issues while you're in this file**, not introduced here: line 51 reassigns `imageReader` without closing the previous instance, so repeated `capturePhoto()` calls leak readers; and `buffer.capacity()` should generally be `buffer.remaining()` for a JPEG plane.

**Verdict:** needs another round — move the listener to a background handler rather than deferring the close into a cancellable coroutine.

---

PR #15881, opened 2026-08-20
all: smoother feedback composer view modelling (fixes #16014)

### LINKED ISSUE #16014: Route feedback submission through FeedbackComposerViewModel
Tracking issue for PR #15881.

Extracted feedback submission logic from `FeedbackFragment` into a dedicated `FeedbackComposerViewModel`. The ViewModel now manages the user lookup and single submit call, exposing state via `StateFlow` (Idle, Submitting, Saved, Error). The UI collects this state to manage progress and dismissal behavior safely without holding a direct reference to repositories.

---
*PR created automatically by Jules for task [8651185155439874750](https://jules.google.com/task/8651185155439874750) started by @dogi*

dogi:
Extracted feedback submission logic from `FeedbackFragment` into a dedicated `FeedbackComposerViewModel`. The ViewModel now manages the user lookup and single submit call, exposing state via `StateFlow` (Idle, Submitting, Saved, Error). The UI collects this state to manage progress and dismissal behavior safely without holding a direct reference to repositories.

---
*PR created automatically by Jules for task [8651185155439874750](https://jules.google.com/task/8651185155439874750) started by @dogi*


dogi:
**Review — rating: 76/100**

More than a move — it fixes two things worth naming:

- **It adds error handling that didn't exist.** The old `createAndSaveFeedback` call had no try/catch, so a failure killed the coroutine and the toast/dismiss never ran. Errors now surface and the buttons come back.
- **It fixes a real submit-before-ready bug.** The old fragment fetched the user name in `onViewCreated` into `var user: String? = ""` and read that field at submit time. Submit before that coroutine finished and the feedback was saved with an **empty user name**. Resolving the name inside the submit path removes the race entirely.

Disabling both buttons during `Submitting` (double-submit prevention) and keeping the write in `viewModelScope` so it survives a configuration change are both good additions too.

**The problem: `Saved` is a one-shot effect stored in a `StateFlow`.** `collectLatestWhenStarted` uses `repeatOnLifecycle(STARTED)`, and `_uiState` never leaves `Saved`. So any stop/start or view recreation after a successful submit replays it — a second toast, a second `mListener?.onFeedbackSubmitted()`, and a second `dismiss()`. A duplicate listener callback likely means a duplicate list refresh downstream.

Your own #15867 has the right pattern for this — a `Channel` plus `receiveAsFlow` for the dismiss signal, precisely so a completion can't be redelivered. Either use that here, or reset to `Idle` after handling `Saved`.

**Nit — the error text isn't localizable.** `SubmitState.Error(e.message ?: "Unknown error")` puts a raw exception message in front of the user, and the fallback is a hardcoded English literal. This project ships five translations and `feedback_saved` is already a string resource; the failure path deserves the same.

**Nit:** three imports are now dead — `androidx.lifecycle.lifecycleScope`, `javax.inject.Inject`, and `kotlinx.coroutines.launch` were used only by the `@Inject` fields and the two `lifecycleScope.launch` blocks this PR removes. Kotlin only warns, so CI stays green.

**Nit:** `object Idle`/`Submitting`/`Saved` can be `data object` on Kotlin 2.4.

**Nit:** no test, and this package is one where that stands out — `FeedbackDetailViewModelTest` and `FeedbackListViewModelTest` both exist, so the composer is the odd one out. The success and error paths are two assertions with a mocked repository.

**Verdict:** needs another round — make `Saved` a one-shot event so submit effects can't replay.

---

Okuro3499 (approved):

dogi:
**Re-review — Rating: 89/100** (was 76)

The replay problem is fixed with the right primitive: `Channel(Channel.BUFFERED)` + `receiveAsFlow()`, so `Saved` is consumed once and cannot be redelivered by a STOPPED→STARTED cycle. No second toast, no second `onFeedbackSubmitted()`, no second `dismiss()`. Splitting the state in two — a `StateFlow<Boolean>` for `isSubmitting` (which *should* replay, since the buttons need re-syncing on view recreation) and a channel for the one-shot effects — is exactly the right division, and better than resetting a single state back to `Idle`.

It also has a property worth pointing out, because it's the reason a channel beats `SharedFlow` here: if the dialog is stopped when the write finishes, the event sits buffered and is delivered on restart rather than being dropped. So the submit result survives backgrounding without duplicating.

The error path is localized now — `R.string.error` is `Error: %1$s` and translated into all five shipped languages, so the wrapper text follows the user's locale.

Both prior nits about dead code are gone (`lifecycleScope`, `javax.inject.Inject`, `launch` all removed with the fields that used them), and the test covers both branches, asserting the `Error` payload rather than just its type.

The two things I called out originally as the real wins survive intact: resolving the user name inside the submit path (instead of reading a field a background coroutine may not have filled) removes the empty-user-name race, and `viewModelScope` keeps the write alive across a configuration change.

**Nit:** the fallback `event.message ?: "An error occurred"` is still a hardcoded English literal — the one part of the failure path that won't translate.

**Nit:** `object Saved` can be `data object` on Kotlin 2.4 (useful `toString`, and it reads as intentional rather than legacy).

**Verdict:** mergeable.

---

PR #15883, opened 2026-08-20
courses: smoother take view modelling (fixes #16015)

### LINKED ISSUE #16015: Move the course-rating eligibility decision into `TakeCourseViewModel`
Tracking issue for PR #15883.

Resolves issue where course-rating eligibility decision logic was improperly located in `TakeCourseFragment`. The logic has now been migrated to `TakeCourseViewModel` using a sealed interface (`RatingPromptDecision`) for the UI to consume, preventing the leakage of the `RatingsRepository` dependency and raw data classes into the UI layer. Tests were added to verify all prompt decision branches.

---
*PR created automatically by Jules for task [4889462478729103865](https://jules.google.com/task/4889462478729103865) started by @dogi*

dogi:
Resolves issue where course-rating eligibility decision logic was improperly located in `TakeCourseFragment`. The logic has now been migrated to `TakeCourseViewModel` using a sealed interface (`RatingPromptDecision`) for the UI to consume, preventing the leakage of the `RatingsRepository` dependency and raw data classes into the UI layer. Tests were added to verify all prompt decision branches.

---
*PR created automatically by Jules for task [4889462478729103865](https://jules.google.com/task/4889462478729103865) started by @dogi*

## Summary by CodeRabbit

- **New Features**
  - Course rating prompts now appear only when a course has not yet been rated.
  - Prompts are skipped when required information is unavailable or rating checks cannot be completed.

- **Bug Fixes**
  - Improved rating-prompt eligibility handling to prevent inappropriate or duplicate prompts.

- **Chores**
  - Updated the Android app release to version 0.65.92 (build 6592).


dogi:
**Review — rating: 89/100**

Right move — the fragment was injecting `RatingsRepository` solely to answer "has this user already rated?", which is a decision, not view plumbing. It's gone from the fragment now.

**There is one behavior change worth stating, and I think it's a fix.** Previously, when `cId` was valid but `userId` was null or empty, `hasRated` fell to `false` and the dialog was **shown**. The new `getRatingPromptDecision` returns `Skip` for a null/empty `userId`, so no prompt. Asking someone with no user id to submit a rating was never going to end well, so tightening it is correct — but it isn't mentioned in the description, and it's the kind of thing worth being deliberate about.

**What I especially like: you pinned the pre-existing failure default instead of quietly changing it.** `getRatingPromptDecision_whenRepositoryThrows_returnsShow` locks in that a lookup failure still results in a prompt. That's a genuine judgement call (risk re-prompting someone who already rated, vs. never prompting), it matches the old `catch { false }`, and now it's documented by a test rather than by an anonymous catch block.

Test coverage is thorough — all four null/empty permutations, the throw path, no-rating, `existingRating` set, and `userRating` set. That exercises both sides of the `userRating != null || existingRating != null` short-circuit, which is exactly where a move like this could silently invert.

**Nit — the guard is now split across two places.** The fragment checks `cId != null` while the ViewModel checks `courseId.isNullOrEmpty()`. The outcome is still correct (an empty `cId` yields `Skip`, so the dialog is blocked), but correctness now depends on reading both halves. `!cId.isNullOrEmpty()` at the call site would keep it self-evident.

**Nit:** `RatingPromptDecision` with exactly `Show`/`Skip` is a named Boolean. That's a defensible trade for readability at the call site, but if you keep it, `data object` is preferred on Kotlin 2.4 for sane `toString`/`equals`.

**Nit:** the three new test imports aren't in alphabetical order (`RatingEntry` < `RatingSummary` < `RatingsRepository`).

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #15884, opened 2026-08-20
resources: smoother layout handling (fixes #16016)

### LINKED ISSUE #16016: Cache toggle views and reuse LayoutManager in ResourcesFragment
Tracking issue for PR #15884.

This PR resolves performance issues related to repeated view lookups and unnecessary recreation of LayoutManager in `ResourcesFragment`.
- The `layoutViewToggle`, `toggleGridButton`, and `toggleListButton` views are now cached as fields in `onViewCreated` rather than fetched via `findViewById` during every property access, removing four view-tree walks during UI updates.
- The `RecyclerView.LayoutManager` applies the reuse pattern (similar to `CoursesFragment`), updating span count in `GridLayoutManager` dynamically rather than rebuilding it completely upon layout changes.
- Safe-calls are implemented where necessary to prevent NullPointerExceptions.
- Fragment view references are properly nulled out in `onDestroyView` to avoid memory leaks.

---
*PR created automatically by Jules for task [15763913515390850692](https://jules.google.com/task/15763913515390850692) started by @dogi*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resource list display stability when switching between list and grid views.
  * Preserved compatible layouts during view-mode changes for smoother UI updates.
  * Added safer handling when view controls are unavailable, reducing potential crashes.

dogi:
This PR resolves performance issues related to repeated view lookups and unnecessary recreation of LayoutManager in `ResourcesFragment`.
- The `layoutViewToggle`, `toggleGridButton`, and `toggleListButton` views are now cached as fields in `onViewCreated` rather than fetched via `findViewById` during every property access, removing four view-tree walks during UI updates.
- The `RecyclerView.LayoutManager` applies the reuse pattern (similar to `CoursesFragment`), updating span count in `GridLayoutManager` dynamically rather than rebuilding it completely upon layout changes.
- Safe-calls are implemented where necessary to prevent NullPointerExceptions.
- Fragment view references are properly nulled out in `onDestroyView` to avoid memory leaks.

---
*PR created automatically by Jules for task [15763913515390850692](https://jules.google.com/task/15763913515390850692) started by @dogi*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resource list display stability when switching between list and grid views.
  * Preserved compatible layouts during view-mode changes for smoother UI updates.
  * Added safer handling when view controls are unavailable, reducing potential crashes.


dogi:
**Review — rating: 89/100**

Reviewed back to back with #15877, which applies the same optimization to `CoursesFragment`. This is the version that gets the view lifecycle right.

**The LayoutManager reuse is correct.** Reassigning `recyclerView.layoutManager` destroys every ViewHolder, loses scroll position and forces a full re-layout, so skipping it for a grid→grid span change is a real win. It also takes effect properly: `GridLayoutManager.setSpanCount()` invalidates the span-index cache and calls `requestLayout()` internally, and early-returns when the value is unchanged, so redundant toggles cost nothing.

The condition handles the trap correctly — `GridLayoutManager` **extends** `LinearLayoutManager`, so a plain `is LinearLayoutManager` check would have matched a grid manager and silently skipped the switch to list mode. Your inverted form (`!is LinearLayoutManager || is GridLayoutManager`) is also cleaner than the equivalent empty-`if` in #15877; I've suggested that fragment adopt this shape.

**The view caching is done properly, which is the main thing separating this from #15877.** Nullable `var`s, `?.` at every call site, and all three fields cleared in `onDestroyView` alongside the existing `_binding = null`. That matters because these fragments sit on the back stack with destroyed views, so a retained button reference would pin the whole dead view hierarchy through its parent chain.

Two details I checked rather than assumed:

- **`layoutViewToggle`'s fallback is preserved.** The old computed property was `findViewById(R.id.layout_view_toggle) ?: (toggleGridButton?.parent as? View)`, and the cached assignment keeps both halves. Resolving it once per view creation instead of per access is safe — the hierarchy can't change within a single view lifecycle — and both use sites (lines 462 and 471) were already null-safe calls, so nullability semantics are unchanged.
- **Assignment order is fine.** The three fields are populated at the top of `onViewCreated`, before `setupViewModeToggle()` and `updateToggleUi()` run.

The `findViewById` removal is a genuine if small saving too — `updateToggleUi` runs on every toggle *and* every layout-manager application, and each call was doing two lookups.

**Nit:** no test, but this is fragment view plumbing with no `androidTest` source set — not reasonable to ask for here.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #15887, opened 2026-08-20
sync: smoother repository time logger providing (fixes #16209)

### LINKED ISSUE #16209: Refactor SyncTimeLogger to constructor injection
Tracking issue for PR #15887.

Refactors `SyncTimeLogger` from a global object utilizing Hilt service locators (`EntryPointAccessors`) to an `@Singleton` injected class, breaking its hard-coupling to `MainApplication` and improving testability across the sync layer.

---
*PR created automatically by Jules for task [10901609925359142385](https://jules.google.com/task/10901609925359142385) started by @dogi*

## Summary by CodeRabbit

- **New Features**
  - Added server reachability checks with URL fallback support and short-term result caching.
  - Improved synchronization timing summaries and persistence.

- **Improvements**
  - Standardized timing metrics for database operations across synchronization workflows.
  - Improved dependency handling for more reliable synchronization logging.

dogi:
Refactors `SyncTimeLogger` from a global object utilizing Hilt service locators (`EntryPointAccessors`) to an `@Singleton` injected class, breaking its hard-coupling to `MainApplication` and improving testability across the sync layer.

---
*PR created automatically by Jules for task [10901609925359142385](https://jules.google.com/task/10901609925359142385) started by @dogi*

## Summary by CodeRabbit

- **New Features**
  - Added server reachability checks with URL fallback support and short-term result caching.
  - Improved synchronization timing summaries and persistence.

- **Improvements**
  - Standardized timing metrics for database operations across synchronization workflows.
  - Improved dependency handling for more reliable synchronization logging.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6697
-        versionName = "0.66.97"
+        versionCode = 6698
+        versionName = "0.66.98"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/di/ServiceModule.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/di/ServiceModule.kt
@@ -87,8 +87,9 @@ object ServiceModule {
         progressRepository: ProgressRepository,
         surveysRepository: SurveysRepository,
         dispatcherProvider: DispatcherProvider,
-        userSessionManager: UserSessionManager
+        userSessionManager: UserSessionManager,
+        syncTimeLogger: org.ole.planet.myplanet.utils.SyncTimeLogger
     ): TransactionSyncManager {
-        return TransactionSyncManager(apiInterface, context, voicesRepository, chatRepository, feedbackRepository, sharedPrefManager, userRepository, userSyncRepository, activitiesRepository, teamsSyncRepository, notificationsRepository, tagsRepository, ratingsRepository, submissionsRepository, coursesRepository, communityRepository, healthRepository, progressRepository, surveysRepository, dispatcherProvider, userSessionManager)
+        return TransactionSyncManager(apiInterface, context, voicesRepository, chatRepository, feedbackRepository, sharedPrefManager, userRepository, userSyncRepository, activitiesRepository, teamsSyncRepository, notificationsRepository, tagsRepository, ratingsRepository, submissionsRepository, coursesRepository, communityRepository, healthRepository, progressRepository, surveysRepository, dispatcherProvider, userSessionManager, syncTimeLogger)
     }
 }
--- a/app/src/main/java/org/ole/planet/myplanet/di/ServiceModule.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/di/ServiceModule.kt
@@ -34,6 +34,7 @@ import org.ole.planet.myplanet.services.UploadToShelfService
 import org.ole.planet.myplanet.services.UserSessionManager
 import org.ole.planet.myplanet.services.sync.TransactionSyncManager
 import org.ole.planet.myplanet.utils.DispatcherProvider
+import org.ole.planet.myplanet.utils.SyncTimeLogger
 
 @Qualifier
 @Retention(AnnotationRetention.BINARY)
@@ -88,7 +89,7 @@ object ServiceModule {
         surveysRepository: SurveysRepository,
         dispatcherProvider: DispatcherProvider,
         userSessionManager: UserSessionManager,
-        syncTimeLogger: org.ole.planet.myplanet.utils.SyncTimeLogger
+        syncTimeLogger: SyncTimeLogger
     ): TransactionSyncManager {
         return TransactionSyncManager(apiInterface, context, voicesRepository, chatRepository, feedbackRepository, sharedPrefManager, userRepository, userSyncRepository, activitiesRepository, teamsSyncRepository, notificationsRepository, tagsRepository, ratingsRepository, submissionsRepository, coursesRepository, communityRepository, healthRepository, progressRepository, surveysRepository, dispatcherProvider, userSessionManager, syncTimeLogger)
     }
--- a/app/src/main/java/org/ole/planet/myplanet/repository/SyncRepositoryImpl.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/SyncRepositoryImpl.kt
@@ -41,7 +41,8 @@ class SyncRepositoryImpl @Inject constructor(
     private val coursesRepository: CoursesRepository,
     private val eventsRepository: EventsSyncWriter,
     private val teamsSyncRepository: TeamsSyncRepository,
-    private val transactionSyncManager: dagger.Lazy<TransactionSyncManager>
+    private val transactionSyncManager: dagger.Lazy<TransactionSyncManager>,
+    private val syncTimeLogger: SyncTimeLogger
 ) : SyncRepository {
     override fun uploadLoginData(): Flow<SyncUiState> {
         val workRequest = OneTimeWorkRequest.Builder(UserDataWorker::class.java)
@@ -129,7 +130,7 @@ class SyncRepositoryImpl @Inject constructor(
 
     private suspend fun processShelfDataOptimizedSync(shelfId: String?, shelfData: Constants.ShelfData, shelfDoc: JsonObject?): Int {
         var processedCount = 0
-        val logger = SyncTimeLogger
+        val logger = syncTimeLogger
 
         try {
             val array = getJsonArray(shelfData.key, shelfDoc)
--- a/app/src/main/java/org/ole/planet/myplanet/repository/SyncRepositoryImpl.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/SyncRepositoryImpl.kt
@@ -200,7 +200,7 @@ class SyncRepositoryImpl @Inject constructor(
                         "teams" -> processedCount += teamsSyncRepository.batchInsertMyTeams(documentsToProcess)
                     }
                     val realmDuration = SystemClock.elapsedRealtime() - realmStartTime
-                    logger.logRealmOperation("shelf_insert", shelfData.type, realmDuration, documentsToProcess.size)
+                    logger.logDbOperation("shelf_insert", shelfData.type, realmDuration, documentsToProcess.size)
                 }
 
                 val batchDuration = SystemClock.elapsedRealtime() - batchStartTime
--- a/app/src/main/java/org/ole/planet/myplanet/services/sync/SyncManager.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/services/sync/SyncManager.kt
@@ -70,7 +70,8 @@ class SyncManager @Inject constructor(
     private val timeProvider: TimeProvider,
     private val teamsRepository: TeamsRepository,
     private val userSyncRepository: UserSyncRepository,
-    private val syncRepository: SyncRepository
+    private val syncRepository: SyncRepository,
+    private val syncTimeLogger: SyncTimeLogger
 ) {
     private val timestampFormat = DateTimeFormatter.ofPattern("HH:mm:ss.SSS").withZone(ZoneId.systemDefault())
     private val isSyncing = AtomicBoolean(false)
@@ -133,7 +134,7 @@ class SyncManager @Inject constructor(
 
     private suspend fun startFullSync() {
         val syncStartTime = SystemClock.elapsedRealtime()
-        val logger = SyncTimeLogger
+        val logger = this.syncTimeLogger
         logger.startLogging()
         Log.d("SyncPerf", "═══════════════════════════════════════════════════════════════")
         Log.d("SyncPerf", "FULL SYNC STARTED at ${timestampFormat.format(Instant.now())}")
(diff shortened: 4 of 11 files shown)


dogi:
**Review — rating: 74/100**

The core refactor is right and mostly well executed. Two details I checked that matter:

- **`@Singleton` preserves the semantics.** `SyncTimeLogger` accumulates state in `processTimes`/`processItemCounts`, so a non-scoped binding would have fragmented timings across injection sites. Scoping it keeps the single-instance behavior the `object` had.
- **The test improvement is real.** `TransactionSyncManagerCheckpointTest` drops `mockkObject(SyncTimeLogger)`/`unmockkObject` in favour of a plain `mockk<SyncTimeLogger>()`. Replacing global object mocking with an injected double is exactly the payoff this refactor should buy.

Getting `EntryPointAccessors.fromApplication(MainApplication.context, …)` and `MainApplication.applicationScope` out of a utility class, and routing the log write through `DiagnosticsRepository` instead of `MainApplication.createLog`, are all improvements.

**But the PR smuggles in a hand-rolled reachability check, and that's a step backwards.** `MainApplication.isServerReachable(...)` was replaced with a new private `isServerReachable` built on raw `HttpURLConnection`:

- **The connection is never closed.** There's no `conn.disconnect()` on any path, success or failure, so every call leaks a connection — up to twice per sync summary.
- **It bypasses the app's HTTP stack.** Everything else goes through OkHttp/Retrofit with the configured client, interceptors (`RetryInterceptor`) and `network_security_config.xml`. A raw `HttpURLConnection` gets none of that.
- **It forces `http://`.** `if (!urlString.startsWith("http://") && !urlString.startsWith("https://")) "http://$urlString"` will downgrade a scheme-less https-only server to cleartext, which the network security config may then block — a different failure mode than the existing path.
- **It duplicates logic rather than moving it.** `MainApplication.isServerReachable` still exists for its other callers, so there are now two reachability implementations that can drift.

The fix is in the spirit of the PR: inject the existing reachability capability rather than reimplementing it, so this class keeps getting its collaborators from the graph instead of hand-building a network call.

**Nit:** `saveSummaryToRealm` and `logRealmOperation` still say "Realm", a store this project no longer uses at all. You're already editing both — good moment to rename to `saveSummaryToRoom` / `logDbOperation`.

**Nit:** `SyncTimeLoggerTest` now constructs the logger with six positional `mockk(relaxed = true)` arguments to test `extractProcessName`, which is pure string manipulation. That's brittle — a new constructor parameter breaks it confusingly. Keeping `extractProcessName` in a `companion object` would let the test stay dependency-free.

**Nit:** `java.net.URL`/`java.net.HttpURLConnection` are imported after `java.time.*`; `ServiceModule` uses a fully-qualified `org.ole.planet.myplanet.utils.SyncTimeLogger` parameter type instead of an import; and `val logger = this.syncTimeLogger` doesn't need the `this.` or the alias.

**Verdict:** needs another round — inject reachability instead of reimplementing it with a leaking `HttpURLConnection`.

---

Okuro3499 (dismissed):
```
Problem: HttpURLConnection is opened and connected but never disconnected — no finally block, unlike the MainApplication.getResponseCode() it was derived from.
```

dogi:
**Review — rating: 74/100**

The core refactor is one I'd very much like to see land. `SyncTimeLogger` was an `object` reaching into `EntryPointAccessors.fromApplication(MainApplication.context, …)` for its dependencies, and it's now a `@Singleton` class with `TimeProvider`, `@ApplicationScope` scope, `DispatcherProvider`, `SharedPrefManager`, `ServerUrlMapper` and `DiagnosticsRepository` injected. `MainApplication.applicationScope` and `MainApplication.createLog` both drop out. `@Singleton` is the right scope — the `ConcurrentHashMap` process state stays shared across `SyncManager`, `TransactionSyncManager` and `SyncRepositoryImpl`, so `startLogging`/`startProcess` still coordinate the way they did as an object.

**The test change is a bigger win than it looks.** `TransactionSyncManagerCheckpointTest` was doing `mockkObject(SyncTimeLogger)` plus `unmockkObject` in teardown — a global singleton mock shared with every other test in the JVM. It now takes a plain `mockk<SyncTimeLogger>(relaxed = true)` passed through the constructor. That removes exactly the class of shared-test-state hazard that `ChatApiServiceTest` carries a written warning about.

**But the PR adds a reachability abstraction and then doesn't use it.**

`ServerReachabilityProvider` is new here and is the right shape: `@Singleton`, injected `@StandardHttpClient OkHttpClient`, `HEAD` requests, a 30-second TTL cache, primary-then-alternative fallback via `ServerUrlMapper`, and `if (e is CancellationException) throw e` so it doesn't eat cancellation. Nothing in this diff injects it — not `SyncTimeLogger`, not anything else — so as shipped it's dead code.

Meanwhile `SyncTimeLogger` grew its *own* private replacement for `MainApplication.isServerReachable`:

```kotlin
private suspend fun isServerReachable(urlString: String): Boolean = withContext(dispatcherProvider.io) {
    …
    val conn = url.openConnection() as HttpURLConnection
    conn.requestMethod = "GET"
    …
} catch (e: Exception) { false }
```

That's the worse of the two on three counts: raw `HttpURLConnection` instead of the configured OkHttp client (so no interceptors, no timeouts config, no shared connection pool), `GET` instead of `HEAD` (downloads a body to answer a yes/no), and a bare `catch (e: Exception)` that **swallows `CancellationException`** — a cancelled sync gets reported as "server unreachable" and the surrounding code carries on, which is the bug the new provider explicitly guards against.

The repo doesn't need a fifth implementation of this. `MainApplication.isServerReachable` (8+ call sites), `SyncActivity.isServerReachable`, and `BellDashboardFragment.isServerReachable` already exist. Injecting `ServerReachabilityProvider` into `SyncTimeLogger` and deleting the private copy makes this PR strictly subtractive, and sets up the follow-up that collapses the other three.

**Nit:** `val logger = this.syncTimeLogger` in three places in `SyncManager` — the `this.` is redundant now that it's a constructor property.

**Nit:** `ServiceModule` declares `syncTimeLogger: org.ole.planet.myplanet.utils.SyncTimeLogger` fully qualified inline while every other parameter uses an import.

**Nit:** `java.net.URL` / `java.net.HttpURLConnection` are inserted after the `java.time.*` imports rather than before, and in reverse order relative to each other.

**Nit:** `SyncTimeLoggerTest` now constructs the logger with six positional `mockk(relaxed = true)` arguments to test `extractProcessName`, which touches none of them — any new constructor parameter breaks that line. If `extractProcessName` is pure, it belongs in a `companion object` (or as a top-level function) so the test needs no instance at all.

**Nit:** `TransactionSyncManagerCheckpointTest` keeps three `every { mockSyncTimeLogger.… } returns Unit` stubs on a mock already created with `relaxed = true`.

**Verdict:** needs another round — wire `SyncTimeLogger` to `ServerReachabilityProvider` and drop the private `HttpURLConnection` copy; the injection work itself is good.

---

dogi:
**Re-review — Rating: 70/100** (was 74)

The core refactor still holds up — `SyncTimeLogger` as an injected `@Singleton` instead of an `object` reaching through `EntryPointAccessors`, `MainApplication.applicationScope`/`createLog` gone, and the checkpoint test swapping `mockkObject(SyncTimeLogger)` for a constructor-passed `mockk` remains the best part of the PR.

**The one blocking item from last round is unchanged, and looking closer it's worse than "unused".** `ServerReachabilityProvider` is still added and still injected by nothing. Meanwhile `SyncTimeLogger` grew its own private `isServerReachable`, and that copy is a downgrade on four counts against the class sitting unused beside it:

```kotlin
conn.requestMethod = "GET"     // provider uses .head() — GET pulls a body to answer a yes/no
…
} catch (e: Exception) {
    false                       // provider rethrows CancellationException; this swallows it
}
```

- **It swallows `CancellationException`.** This runs inside `appScope.launch(dispatcherProvider.io)`, so when the scope is cancelled the call returns `false` instead of propagating, and execution continues into the block below that rewrites server-URL preferences via `serverUrlMapper.updateUrlPreferences`. Writing preferences during shutdown is not a good failure mode, and it's exactly the case `ServerReachabilityProvider` already handles with `if (e is CancellationException) throw e`.
- **No cache.** The provider has a 30-second TTL; `saveSummaryToRealm` calls this up to twice per summary with no memoisation.
- **`HttpURLConnection` instead of the injected `@StandardHttpClient` OkHttpClient**, so it bypasses the app's configured timeouts, interceptors and TLS setup.
- **`conn.disconnect()` is never called**, and there's no `use`-style cleanup.

Injecting `ServerReachabilityProvider` and deleting the private copy resolves all four at once and makes the new class earn its place.

**Nit:** the Realm→DB rename is half done. `logRealmOperation` keeps its name while its payload became `DbOperationLog`; the summary prints "DB OPERATION STATISTICS" but the locals are still `totalRealmOps`/`totalRealmTime`/`realmPercentage`; and `saveSummaryToRealm` now calls `diagnosticsRepository.saveLogToRoom`. Worth finishing in one pass rather than leaving both vocabularies in the file.

**Nit:** `ServiceModule` declares `syncTimeLogger: org.ole.planet.myplanet.utils.SyncTimeLogger` fully qualified while every other parameter in that provider is imported.

**Nit:** `SyncManager` uses `this.syncTimeLogger` in three places where the bare name works; and `SyncTimeLoggerTest` gained an `io.mockk.mockk` import it doesn't appear to use, since `extractProcessName` moved to a companion and the test calls it unchanged.

**Verdict:** needs changes — use `ServerReachabilityProvider` or drop it; either way the private `isServerReachable` must stop swallowing cancellation.

---

dogi:
**Re-review — Rating: 72/100** (was 70)

The Realm→DB rename is finished now — `totalDbOps`, `totalDbTime`, `totalDbItems` and `dbPercentage` replaced the mismatched locals, so the summary no longer mixes both vocabularies. That was one of the nits.

**The blocker is byte-for-byte unchanged for a second round.** `ServerReachabilityProvider` is still added and still injected by nothing, and `SyncTimeLogger.isServerReachable` still hand-rolls the same weaker duplicate: `HttpURLConnection` with `requestMethod = "GET"` instead of the injected `@StandardHttpClient` with `.head()`, no cache against the provider's 30-second TTL, no `disconnect()`, and `catch (e: Exception) { false }` — which swallows `CancellationException` inside `appScope.launch(dispatcherProvider.io)` and then falls through into the `serverUrlMapper.updateUrlPreferences` block that rewrites server-URL preferences.

Details are in my previous comment; nothing has changed about them. Either inject the provider and delete the private copy, or drop the provider from this PR and at minimum rethrow cancellation in the private one.

Also still standing: `ServiceModule` declares `syncTimeLogger: org.ole.planet.myplanet.utils.SyncTimeLogger` fully qualified while every sibling parameter is imported; `SyncManager` uses `this.syncTimeLogger` in three places where the bare name works; `logRealmOperation` keeps its old name while its payload is `DbOperationLog`; `saveSummaryToRealm` still says Realm while calling `diagnosticsRepository.saveLogToRoom`; the `java.net.URL`/`HttpURLConnection` imports sit after `java.time.format.DateTimeFormatter`; and `SyncTimeLoggerTest` has an `io.mockk.mockk` import it doesn't use.

The underlying refactor — object to injected `@Singleton`, `EntryPointAccessors` gone, and the checkpoint test taking a constructor `mockk` instead of `mockkObject` — is still worth landing. It's the reachability duplicate holding it up.

**Verdict:** needs changes — same single item as last round.

---

dogi:
**Re-review — Rating: 84/100** (was 72)

The standing blocker is resolved. `ServerReachabilityProvider` now has a real consumer — injected into `SyncTimeLogger` and called for both the primary and alternative probes — and the private `isServerReachable` that swallowed `CancellationException` inside `appScope.launch` is gone, replaced by the provider's `if (e is CancellationException) throw e`. I also checked the logging port against what it replaced: `diagnosticsRepository.saveLogToRoom("sync summary", summary, "${timeProvider.now()}")` is exactly what `MainApplication.createLog` did at `MainApplication.kt:137-141`, minus an `EntryPointAccessors` hop, so nothing is lost. The test change is a genuine improvement too — `mockk<SyncTimeLogger>(relaxed = true)` passed as a constructor arg retires the `mockkObject`/`unmockkObject(SyncTimeLogger)` pair, which is the whole point of making it injectable.

The one thing I'd want stated before this is finished: `MainApplication.isServerReachable`/`isPrimaryServerReachable` and their own `reachabilityCache` are still there with 15+ callers (`ChatDetailFragment`, `SyncActivity`, `ServerReachabilityWorker`, `TeamsRepositoryImpl`, `SubmissionsUploader`, `BellDashboardViewModel`, …). So the app now has two reachability implementations with two independent 30s caches — the same URL can be reachable in one and not the other, and the same probe runs twice. That's fine as a first step if migrating the rest is the plan; it's a drift hazard if it isn't. Worth saying which.

Separately, and explicitly **not** this PR's bug: the caller does `primaryAvailable = serverReachabilityProvider.isServerReachable(mapping.primaryUrl)`, but that method tries the primary *and* the alternative, so it returns `true` whenever the alternative is up — making the `!primaryAvailable && alternativeUrl != null && …` failover branch below it unreachable in exactly the case it exists for. Your port is faithful to `MainApplication.kt:178-204`, so this is pre-existing, and the same shape appears at `TeamsRepositoryImpl.kt:969`, `SubmissionsUploader.kt:38`, and `ServerUrlMapper.kt:98`. Mentioning it because `isPrimaryServerReachable` — which `MainApplicationTest` has a dedicated `never consults the alternative URL mapping` test for — is currently called by nothing in the new class, and switching that one call site would both fix the branch and give it a consumer.

**Nit:** `kotlinx.coroutines.withContext` is imported into `SyncTimeLogger` but unused (the new body only does `appScope.launch(dispatcherProvider.io)`); `REACHABILITY_CACHE_TTL_MS` is an instance `val` in constant casing and wants a `companion object`/`const`; and removing `val logger = SyncTimeLogger` left stray double blank lines in `startFullSync`, `resourceTransactionSync`, `myLibraryTransactionSync`, and above the new `companion object`. Also: `ServerReachabilityProvider` has no test of its own while its predecessor has three — porting them is mechanical.

**Verdict:** the blocker that held this for two rounds is genuinely fixed and the conversion is complete; shipping as-is, with the `MainApplication` migration as the named follow-up.

---

PR #15888, opened 2026-08-20
life: smoother personals repository dao querying (fixes #16019)

### LINKED ISSUE #16019: Perf: optimize updatePersonalAfterSync by collapsing to one statement
Tracking issue for PR #15888.

Perf: optimize updatePersonalAfterSync by collapsing to one statement

Currently, `PersonalsRepositoryImpl.updatePersonalAfterSync` does `findById(id)`, mutates three fields, then calls `personalDao.update(personal)`. This requires two round trips and does a full row rewrite per uploaded personal resource, inside a per-item upload loop.

This commit introduces a new targeted `@Query` method in `PersonalDao` (`updateUploadedStatus`) that directly updates `isUploaded`, `_id`, and `_rev` for the matching ID in a single SQL statement. `PersonalsRepositoryImpl` and the associated unit tests have been updated to use this new targeted query.

---
*PR created automatically by Jules for task [9170145368412520964](https://jules.google.com/task/9170145368412520964) started by @dogi*

dogi:
Perf: optimize updatePersonalAfterSync by collapsing to one statement

Currently, `PersonalsRepositoryImpl.updatePersonalAfterSync` does `findById(id)`, mutates three fields, then calls `personalDao.update(personal)`. This requires two round trips and does a full row rewrite per uploaded personal resource, inside a per-item upload loop.

This commit introduces a new targeted `@Query` method in `PersonalDao` (`updateUploadedStatus`) that directly updates `isUploaded`, `_id`, and `_rev` for the matching ID in a single SQL statement. `PersonalsRepositoryImpl` and the associated unit tests have been updated to use this new targeted query.

---
*PR created automatically by Jules for task [9170145368412520964](https://jules.google.com/task/9170145368412520964) started by @dogi*


dogi:
**Review — rating: 88/100**

Same shape as #15865 and correct for the same reasons, plus one extra thing worth checking here because the statement writes `_id`.

I verified it isn't touching a primary key: `Personal` declares `@PrimaryKey @JvmField var id: String`, with `_id` as a separate nullable CouchDB column, and the only index is `Index("userId")` — no unique index on `_id`, so `SET _id = :newId` can neither move the PK nor trip a uniqueness conflict. The `WHERE id = :id` predicate matches what `findById(id)` selected, so the same row is targeted.

The rest lines up:

- **Missing-row behavior is preserved.** `findById(...)?.let { … }` silently did nothing when absent; an `UPDATE` matching zero rows does the same.
- **It stops rewriting the whole row.** `personalDao.update(personal)` wrote every column, which means running the entity's converters just to set three fields. Now it's three columns.
- **Atomic.** The read-modify-write window is gone, so a concurrent change to another column of that row can't be clobbered by a stale in-memory copy.
- Room derives affected tables from `@Query` writes, so anything observing `my_personal` still refreshes.

The test rewrite is the right call too — asserting `coVerify { personalDao.updateUploadedStatus(...) }` instead of inspecting a mutated in-memory entity matches what the code now actually does.

**Nit:** `updateUploadedStatus` undersells the method — it also assigns the remote `_id` and `_rev`. Something like `markUploaded(id, newId, rev)` or `applyUploadResult` would tell the next reader that this is the post-upload reconciliation, not just a flag flip.

**Nit:** nothing exercises the SQL itself, since the DAO is mocked in both tests. Same note I left on #15865 — #15859's in-memory `Room.inMemoryDatabaseBuilder` pattern is a cheap template if you want one real round-trip.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #15889, opened 2026-08-20
sync: smoother url utils managing (fixes #16210)

### LINKED ISSUE #16210: perf: Hoist `UrlUtils.getUrl()` and `UrlUtils.header` in `SyncManager` out of loops
Tracking issue for PR #15889.

Extracted `UrlUtils.getUrl()` and `UrlUtils.header` into local variables in `SyncManager.kt` specifically for `resourceTransactionSync` and `getShelvesWithDataBatchOptimized`. This avoids redundant calls which rebuild the server base URL from `SharedPreferences` and recompute basic auth during repetitive background sync processes.

---
*PR created automatically by Jules for task [18415815748116795182](https://jules.google.com/task/18415815748116795182) started by @dogi*

dogi:
Extracted `UrlUtils.getUrl()` and `UrlUtils.header` into local variables in `SyncManager.kt` specifically for `resourceTransactionSync` and `getShelvesWithDataBatchOptimized`. This avoids redundant calls which rebuild the server base URL from `SharedPreferences` and recompute basic auth during repetitive background sync processes.

---
*PR created automatically by Jules for task [18415815748116795182](https://jules.google.com/task/18415815748116795182) started by @dogi*

## Summary by CodeRabbit

* **Performance**
  * Improved synchronization request handling by reusing connection details during resource and shelf data updates.
* **Bug Fixes**
  * Preserved existing synchronization behavior while streamlining resource counts, batch requests, and shelf discovery.


dogi:
**Review — rating: 90/100**

Correct and complete — master has exactly six `UrlUtils.` call sites in `SyncManager` (lines 279, 286, 306, 315, 322, 442) and all six are converted, so there's no half-hoisted path left behind in the `coroutineScope` body.

**Neither accessor is free, which is what makes this worth doing.** `UrlUtils.header` is a `get()` property that resolves the prefs manager, reads `getUrlUser()`/`getUrlPwd()`, and Base64-encodes a basic-auth header on every access; `getUrl()` goes through `dbUrl(spm())` to rebuild the CouchDB URL. Lines 306/315/322 sit **inside the batch loop**, so at batch size ~100 over a large resources table that's hundreds of redundant pref reads, string builds and Base64 encodes eliminated.

**There's also a correctness argument the description doesn't make.** Because both are pref-backed getters, the old code could in principle observe a URL change *mid-pagination* — if anything rewrote the server-URL preferences (as `ServerUrlMapper.updateUrlPreferences` does on an alternative-URL switch), later batches would hit a different server while `skip` offsets were still computed against the first one. Pinning one base URL for the whole paged pull removes that window, so hoisting makes the loop more consistent, not just faster.

**Nit:** the second hoist (line 442, the shelf `_all_docs` fetch) has a single use each and no loop, so those two locals are stylistic rather than a saving. Fine for symmetry — just worth knowing the perf claim rests entirely on the resource batch loop.

**Heads-up on merge order:** #15895 also edits `SyncManager`, and #15887 touches the same `logger` lines in `syncResources`. No logical conflict, but they'll collide textually.

Same optimization as #15927 in `UploadCoordinator` — both look right; I'm reviewing that one alongside #15891.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #15890, opened 2026-08-20
all: smoother search text change flowing (fixes #16020)

### LINKED ISSUE #16020: Optimize search text flows by adding distinctUntilChanged
Tracking issue for PR #15890.

Fixes redundant re-filtering when searching in SurveyFragment, TeamFragment, SubmissionsFragment, and ChatHistoryFragment by correctly chaining `distinctUntilChanged()` and `drop(1)`.

---
*PR created automatically by Jules for task [1911567779248763530](https://jules.google.com/task/1911567779248763530) started by @dogi*

dogi:
Fixes redundant re-filtering when searching in SurveyFragment, TeamFragment, SubmissionsFragment, and ChatHistoryFragment by correctly chaining `distinctUntilChanged()` and `drop(1)`.

---
*PR created automatically by Jules for task [1911567779248763530](https://jules.google.com/task/1911567779248763530) started by @dogi*


dogi:
**Review — rating: 89/100**

Both operators are correct, and both depend on `textChanges()`'s contract in ways worth spelling out — because the obvious failure modes are real and this implementation avoids them.

**`distinctUntilChanged()` works because the flow emits Strings, not the `Editable`.** `ViewExtensions.textChanges()` does `trySend(s?.toString())`. Had it forwarded the raw `CharSequence` from `onTextChanged`, that's the *same mutable `SpannableStringBuilder` instance* on every keystroke, and since it doesn't override `equals`, `distinctUntilChanged` would have compared identity, matched every time, and silently suppressed all search after the first character. Emitting a `String` means the comparison is real value equality.

**`.drop(1)` drops exactly the priming emission, not a keystroke.** `textChanges()` ends with `.onStart { emit(text?.toString()) }`, so there is always a synthetic first value on collection. Dropping it removes a redundant search on every screen entry, and it makes the three fragments consistent with `SubmissionsFragment`, which already had it.

Ordering is right too — `distinctUntilChanged` *after* `debounce` is what suppresses redundant downstream queries (a value that debounces to the same string as last time, e.g. typing a character and deleting it inside the window, or `onTextChanged` firing for IME composition without a net text change).

**One scenario worth a sanity check:** dropping the priming emission means a *pre-populated* field no longer re-applies its filter on collection. On rotation that's fine here — all four route to ViewModels (`by viewModels()` or the shared one), which survive configuration changes and retain the filter, so the results stay consistent with the restored text. Just flagging it as the assumption `.drop(1)` rests on.

**Not a duplicate of #15860**, though the titles suggest overlap: that PR guards *Room* flows and needs hand-written field comparators because the entities are `open class`es with identity equality; this one guards *search text* flows carrying Strings, where the default comparison is the correct one. Different targets, different mechanics.

**Nit:** only `TeamFragment` carries `@OptIn(FlowPreview::class)` for `debounce`, while the other three call it without — `debounce` is stable in coroutines 1.11, so that opt-in looks vestigial. Pre-existing, not yours.

**Verdict:** mergeable.

---

Okuro3499 (approved):
a

PR #15891, opened 2026-08-20
sync: smoother upload batch coordinating (fixes #16021)

### LINKED ISSUE #16021: Optimize lookups and filtering in UploadCoordinator batch processing
Tracking issue for PR #15891.

Optimize lookups and filtering in UploadCoordinator batch processing

Replaced O(N) list scans within `failedResults.mapNotNull` in batch processing functions with O(1) map lookups using `associateBy { it.localId }`. This removes O(n·m) operations and ensures index batch reconciliation by localId.

Replaced O(N^2) list containment check `succeeded.filter { it !in dbFailed }` with a HashSet of failed local ids to optimize batch upload filtering.

---
*PR created automatically by Jules for task [17036315595790696911](https://jules.google.com/task/17036315595790696911) started by @dogi*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved upload failure handling when database updates fail.
  * Prevented successfully uploaded items from being incorrectly marked as failed.
  * Added handling for cases where no database update failures are reported.

dogi:
Optimize lookups and filtering in UploadCoordinator batch processing

Replaced O(N) list scans within `failedResults.mapNotNull` in batch processing functions with O(1) map lookups using `associateBy { it.localId }`. This removes O(n·m) operations and ensures index batch reconciliation by localId.

Replaced O(N^2) list containment check `succeeded.filter { it !in dbFailed }` with a HashSet of failed local ids to optimize batch upload filtering.

---
*PR created automatically by Jules for task [17036315595790696911](https://jules.google.com/task/17036315595790696911) started by @dogi*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved upload failure handling when database updates fail.
  * Prevented successfully uploaded items from being incorrectly marked as failed.
  * Added handling for cases where no database update failures are reported.


dogi:
**Review — rating: 74/100**

Reviewed as a pair with #15927, which also edits `UploadCoordinator.kt`. **Not duplicates** — that PR hoists `UrlUtils.getUrl()` out of the per-item loops (lines 149–190, 373–405), this one replaces list scans with hash lookups (lines 66, 255, 324, 440). The line ranges don't overlap, so both can land.

The change is **correct** — no behavior difference. `UploadedItem` is a `data class`, so `it !in dbFailed` was structural equality, and `dbFailed` holds the very same instances pulled out of `succeeded` via `succeeded.find { … }`, so identity and structural comparison agreed. Keying on `localId` explicitly is arguably clearer than leaning on data-class equality, and I'd take it on readability grounds.

**But the performance argument doesn't hold at the actual batch size, and the common path gets slightly worse.**

`config.batchSize` defaults to `UploadConstants.BATCH_SIZE = 50`, so the O(n·m) scan being replaced is bounded at 50×50 — microseconds. Meanwhile:

```kotlin
val succeededMap = succeeded.associateBy { it.localId }
val failedLocally = failedResults.mapNotNull { succeededMap[it.localId] }
```

`failedResults` is empty whenever the local DB updates succeed, which is the normal case. So this now builds a ~50-entry `HashMap` unconditionally to serve **zero** lookups. Same for `dbFailedIds`: `dbFailed` is usually empty, and `filter { it !in emptyList }` was already O(1) per element, so the `.map{}.toHashSet()` adds two allocations to replace a scan that never ran.

Easy fix that keeps the intent and wins in both directions — bail before building the index:

```kotlin
if (failedResults.isEmpty()) return emptyList()
val succeededMap = succeeded.associateBy { it.localId }
```

and likewise guard `dbFailedIds` behind `dbFailed.isNotEmpty()` (or just keep the `filter` when it's empty).

**Nit:** the same `dbFailedIds` construction is duplicated at lines 66 and 324, and the same `succeededMap` pattern at 255 and 440. If you keep them, a small private helper would stop the two copies drifting.

**Verdict:** needs another round — guard the index construction so the empty-failure path (the normal one) doesn't pay for it.

---

Okuro3499 (approved):

Okuro3499 (approved):

dogi:

The empty-path guards are in at all four sites: `if (failedResults.isEmpty()) return emptyList()` before `associateBy` (lines 259 and 450), and `if (dbFailed.isEmpty())` before building `dbFailedIds` (lines 66 and 330). That was the whole objection — the normal case (local DB updates all succeed) no longer allocates a ~50-entry `HashMap` to serve zero lookups, and the hash lookup is still there for the case that actually has work to do.

Behavior is still equivalent. Keying on `localId` rather than leaning on `UploadedItem`'s data-class equality is the clearer expression of the intent, and `dbFailed` is populated from the same instances found in `succeeded`, so both comparisons agreed to begin with.

One boundary worth naming since it's a pattern that has bitten this codebase elsewhere: `associateBy` is last-wins where `find` was first-wins, so if a single batch ever contained two `UploadedItem`s sharing a `localId` but differing in `remoteId`/`remoteRev`, the two versions would pick different rows. Not reachable today — a batch is built from distinct DB rows — but it's the kind of thing that changes silently if the batch query ever grows a join.

**Nit (carried over, now slightly worse):** the `dbFailed.isEmpty()` block is duplicated verbatim at 66 and 330, and the `succeededMap` pattern at 259 and 450. Last round each was one line; now each is five, so the copies have more room to drift. A small private helper — `fun List<UploadedItem>.excluding(failed: List<UploadedItem>)` and `fun resolveFailedLocally(...)` — would collapse both pairs.

**Verdict:** mergeable.

---

PR #15892, opened 2026-08-20
courses: smoother exams answers caching (fixes #16022)

### LINKED ISSUE #16022: perf(exam): index questions before restoring saved answers
Tracking issue for PR #15892.

Optimizes the `populateCacheFromSavedAnswers` function in `ExamTakingFragment` to prevent O(N^2) latency when resuming surveys.

---
*PR created automatically by Jules for task [10149596108979946157](https://jules.google.com/task/10149596108979946157) started by @dogi*

dogi:
Optimizes the `populateCacheFromSavedAnswers` function in `ExamTakingFragment` to prevent O(N^2) latency when resuming surveys.

---
*PR created automatically by Jules for task [10149596108979946157](https://jules.google.com/task/10149596108979946157) started by @dogi*


dogi:
**Review — rating: 89/100**

Correct and reads better. Turns O(answers × questions) into O(questions + answers) by indexing once outside the loop, and the null handling is equivalent: `questions?.associateBy { … } ?: emptyMap()` means every lookup misses and every iteration hits `return@forEach`, exactly as `questions?.find { … }` returning null did before. `answer.questionId` is null-checked before the lookup, so the nullable key type is fine.

**One semantic difference worth knowing about:** `associateBy` keeps the **last** entry for a duplicate key, whereas `find` returned the **first**. Question ids should be unique within an exam so this shouldn't bite, but it's the same detail #15872 handled deliberately (it used a `containsKey` guard precisely to preserve `indexOf`'s first-wins semantics). Here the mapped value is the question itself rather than a positional index, so a collision would mean duplicate question rows — a data problem either way. Just flagging it rather than asking for a change.

**Nit:** the map is built even when `answers` is an empty list. One-time cost per exam resume rather than per item, so it doesn't matter — unlike #15891, where the same "build the index unconditionally" pattern lands on a hot path.

**Heads-up:** #15878 rewrites the exam-start block in this same file. It's in a different region (`onViewCreated` vs. line ~778), so no logical conflict — and that one's back with the author, so this should land first regardless.

**Verdict:** mergeable.

---

Okuro3499 (approved):

Okuro3499 (approved):

PR #15873, opened 2026-08-20
all: less repositories methods is more (fixes #16008)

dogi:
Removes 5 repository methods that have no production call sites, as well as the newly orphaned DAO methods and unit tests, to reduce fat interface bloat.

---
*PR created automatically by Jules for task [863377141462472712](https://jules.google.com/task/863377141462472712) started by @dogi*

PR #15893, opened 2026-08-20
courses: smoother submissions repository answers querying (fixes #16023)

### LINKED ISSUE #16023: Optimize answer lookup in SubmissionsRepositoryImpl
Tracking issue for PR #15893.

Optimized `SubmissionsRepositoryImpl.getSubmissionDetail` by indexing answers by question ID, avoiding quadratic lookup overhead for large surveys.

---
*PR created automatically by Jules for task [10954294467742674607](https://jules.google.com/task/10954294467742674607) started by @dogi*

dogi:
Optimized `SubmissionsRepositoryImpl.getSubmissionDetail` by indexing answers by question ID, avoiding quadratic lookup overhead for large surveys.

---
*PR created automatically by Jules for task [10954294467742674607](https://jules.google.com/task/10954294467742674607) started by @dogi*


dogi:
**Review — rating: 90/100**

Correct, and it handles the detail that makes this refactor easy to get wrong.

`find` returns the **first** match, and a naive `associateBy` would have kept the **last** — silently changing which answer is shown when a submission carries two entries for the same question. The `!answersLookup.containsKey(qId)` guard preserves first-wins exactly. That's the same care #15872 took and worth calling out, since two other PRs in this queue reached for `associateBy` without it.

The null handling is equivalent too. `question.id` must be non-null for `answersLookup[question.id]` to compile against a `MutableMap<String, Answer>`, and with a non-null key the old `find { it.questionId == question.id }` could never match an answer whose `questionId` was null — which is precisely what the `qId != null` filter excludes from the map. So the lookup set is identical.

Complexity goes from O(questions × answers) to O(questions + answers). Modest per call, but it compounds if this runs once per submission in a list view.

**Nit:** `putIfAbsent` says "first wins" in one line instead of three:

```kotlin
submission.answers?.forEach { answer ->
    answer.questionId?.let { answersLookup.putIfAbsent(it, answer) }
}
```

**Heads-up:** #15873 and #15878 also touch `SubmissionsRepositoryImpl` in different regions — no logical conflict, but whoever lands last may need a manual resolve.

**Verdict:** mergeable.

---

Okuro3499 (approved):

Issue #66, opened 2018-06-10
download file from http / https


leomaxi:
Download files from http / https address (couchdb attachment file) unto local drive on Android device.
Process must queue multiple downloads
Show download process 
Prompt / alert user when download is completed

onlyserenity:
@leomaxi I can use `AsyncTask `(Update UI and have multiple tasks in background) and use `DownloadManager `class to download and save it on the internal storage of the device. 
I just have two questions to plug what I wrote so far:

* Where should this be implemented? (Drawer, action bar, login activity... etc?
* Are the links to those urls are fixed or the user must enter them manually?

leomaxi:
Please implement this on the dashboard fragment activity (users will see their materials in my library and can click to download it offline )
The links to the online server files will be from the database and will be fixed.. After download, we will save the local file URL in a local database for access.

PR #15894, opened 2026-08-20
community: smoother home dialog handling (fixes #16024)

### LINKED ISSUE #16024: Fix memory leaks and layout performance in HomeCommunityDialogFragment
Tracking issue for PR #15894.

Fixes a memory leak and excessive layout calculations in `HomeCommunityDialogFragment` as described in issue #66.

Previously, `BottomSheetBehavior.BottomSheetCallback` and `TabLayoutMediator` instances were created anonymously and not cleared when the Fragment view was destroyed, causing memory leaks and duplicated callbacks across view recreations. Also, `bottomSheet.requestLayout()` was being called unconditionally during sliding operations.

This patch:
- Adds references to retain the callback and the mediator.
- Calls `.removeBottomSheetCallback()` and `.detach()` respectively in `onDestroyView`.
- Clears the ViewPager2 adapter in `onDestroyView`.
- Checks if the layout height actually needs to be changed before applying it and requesting layout in `onSlide` and `onStart`.

---
*PR created automatically by Jules for task [15620820084660906912](https://jules.google.com/task/15620820084660906912) started by @dogi*

dogi:
Fixes a memory leak and excessive layout calculations in `HomeCommunityDialogFragment` as described in issue #66.

Previously, `BottomSheetBehavior.BottomSheetCallback` and `TabLayoutMediator` instances were created anonymously and not cleared when the Fragment view was destroyed, causing memory leaks and duplicated callbacks across view recreations. Also, `bottomSheet.requestLayout()` was being called unconditionally during sliding operations.

This patch:
- Adds references to retain the callback and the mediator.
- Calls `.removeBottomSheetCallback()` and `.detach()` respectively in `onDestroyView`.
- Clears the ViewPager2 adapter in `onDestroyView`.
- Checks if the layout height actually needs to be changed before applying it and requesting layout in `onSlide` and `onStart`.

---
*PR created automatically by Jules for task [15620820084660906912](https://jules.google.com/task/15620820084660906912) started by @dogi*


dogi:
**Review — rating: 87/100**

Three of these are genuine fixes, and it's worth separating them from the defensive one.

**`binding.viewPager2.adapter = null` is the real leak fix.** `CommunityPagerAdapter` is `FragmentStateAdapter(fm)` holding a `FragmentActivity`, and `FragmentStateAdapter` registers a lifecycle observer plus an adapter data observer on attach. Leaving it wired to a destroyed `ViewPager2` keeps a chain from the dead view to the activity and its FragmentManager. This is the documented ViewPager2 leak pattern and clearing it is correct.

**`tabLayoutMediator?.detach()` is real too.** `attach()` installs an `OnPageChangeCallback` on the ViewPager2 and an adapter data observer; without `detach()` both stay registered, holding the TabLayout and the mediator's lambda (which closes over the binding).

**The `requestLayout()` guard is a real win.** `onSlide` fires every frame of a drag, and the old code set `layoutParams.height` and called `requestLayout()` unconditionally — a full measure/layout pass per frame. Since `newHeight` is a pixel int derived from a float offset, consecutive frames frequently compute the same value, so the equality check skips most of them. Same for the `WRAP_CONTENT` guard in the resize path.

**The callback removal is hygiene rather than a demonstrable leak** — `bottomSheetBehavior` belongs to the dialog's bottom sheet view and dies with the dialog, so the anonymous callback wasn't outliving anything. Still the right shape, and holding it in a field is the only way to remove it.

**One omission for a PR about leaks: `bottomSheetBehavior` itself is never nulled.** To be accurate, `BottomSheetBehavior` holds its view through a `WeakReference` (`viewRef`/`nestedScrollingChildRef`), so this isn't a view leak — but it's the obvious one-line completion next to the two fields you *do* clear.

**Nit:** the cleanup runs *after* `super.onDestroyView()`. It works — `_binding` isn't nulled until the end, and `detach()` only unregisters callbacks — but the convention is to release view-scoped resources before calling super, so a future reader doesn't have to reason about whether the hierarchy is still intact.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #15897, opened 2026-08-20
actions: smoother robolectric sdk prefetching (fixes #15935)

### LINKED ISSUE #15935: Fix Robolectric race condition by pre-fetching android-all jars
Parallel test forks race to download Robolectric's `android-all-instrumented` jars at runtime, causing one fork to read a half-written file and fail all subsequent Robolectric tests in that JVM.

Tracked by PR #15897.

dogi:
## Summary

Fixes a flaky test failure where parallel test forks race to download Robolectric's `android-all-instrumented` jars at runtime, causing one fork to read a half-written file and fail all subsequent Robolectric tests in that JVM.

## Changes

- **app/build.gradle**: Added Robolectric SDK jar pre-fetching infrastructure
  - New `robolectricSdkJars` map defining API level → jar version mappings (levels 26, 27, 28, 30, 31, 32, 33, 34, 36)
  - One Gradle configuration per API level to avoid version conflict resolution
  - New `fetchRobolectricSdks` Sync task that stages all jars into `build/robolectric-sdks` before tests run
  - Validation to ensure all pinned SDK levels have their jars staged
  - Test task configuration to use offline mode when `-ProbolectricOffline=true` is passed

- **.github/workflows/test.yml**: Updated CI test invocation
  - Added `-ProbolectricOffline=true` flag to test command
  - New post-test verification step that checks no jars were fetched at runtime (would indicate a race condition occurred)
  - Logs staged jar count and size to job summary for visibility

- **docs/TESTING.md**: Documented the Robolectric runtime behavior
  - Explained the race condition and why it occurs
  - Documented the offline mode solution and how to maintain `robolectricSdkJars`
  - Added guidance for adding new `@Config(sdk = [N])` annotations

- **CLAUDE.md**: Updated test infrastructure documentation
  - Noted the two-shard test split strategy in CI
  - Added reference to Robolectric SDK jar management

## Implementation Details

The solution works by:
1. Gradle resolves all `android-all-instrumented` jars upfront into a shared directory before any test fork starts
2. Robolectric is configured with `robolectric.offline=true` and `robolectric.dependency.dir` pointing to that directory
3. Test forks read pre-staged jars instead of competing to download them, eliminating the race condition
4. CI verifies no runtime fetches occurred; if jars appear in `~/.m2`, it means a new SDK level was added without updating `robolectricSdkJars`

The jar versions come from Robolectric's `DefaultSdkProvider` and must be updated whenever Robolectric is bumped.

https://claude.ai/code/session_0158c4tYJWvypSoGxrASAD4B

## Summary by CodeRabbit

* **Tests**
  * Improved Android test reliability by staging required Robolectric SDKs before tests run.
  * Added verification to detect unexpected runtime downloads and prevent corrupted test environments.
  * Updated test execution to support consistent offline operation across parallel test shards.

* **Documentation**
  * Documented the updated test workflow, artifact handling, and guidance for adding support for new Android API levels.


dogi:
**Review — rating: 84/100**

Right diagnosis, right mechanism, and the part most likely to be quietly wrong — the SDK level map — is provably complete. Both build jobs are green. The two test shards are the open question, below.

**The map is exactly right, which is worth stating because it's easy to get wrong.** Grepping the suite for every `sdk =` form, not just numeric literals, the levels in play are `26`, `32`, `33`, `34` as integers plus `Build.VERSION_CODES.O` (26), `O_MR1` (27), `P` (28, seven uses), `R` (30), `S` (31) and `UPSIDE_DOWN_CAKE` (34) as constants — and `36` as the default, since there's no `robolectric.properties` and Robolectric falls back to `targetSdk`. That's precisely the nine keys in `robolectricSdkJars`, with 29 and 35 correctly absent. Nothing pinned is unused and nothing used is unpinned; someone reading only the numeric literals would have "trimmed" four live levels out of that map.

The plumbing lines up end to end, too:

- Gradle stages Maven artifact filenames, so a jar lands as `android-all-instrumented-<version>.jar` — exactly the name Robolectric's offline resolver builds from artifactId + version. No translation layer to get wrong.
- `rootDir/build/robolectric-sdks` matches the workflow's `find build/robolectric-sdks` from the repo root.
- `settings.gradle` has no `mavenLocal()`, so nothing but Robolectric's own fetcher can write `~/.m2/repository/org/robolectric` — the guard's premise holds even though its message doesn't (below).
- The `dependsOn` only attaches under the flag, so local runs are untouched.
- And the shards getting this far is itself evidence: `fetchRobolectricSdks` throws unless all nine resolve and stage, so a fast failure would have happened in the first minute. All nine coordinates are real published artifacts. What the run still has to prove is that they're the versions *4.16.1* asks for.

**`12f8cf3` is the main thing I'd push back on.** That commit deletes every explanatory comment from the block, and two of the three rationales it removed exist nowhere else — `docs/TESTING.md` keeps the race, the offline mechanism, the map's location and the failure message, but not these:

- *One configuration per level, not one holding all nine.* They're the same module at nine different versions, so a single configuration conflict-resolves down to the newest jar and every test pinning an older level then fails. Without that sentence, nine configurations for nine coordinates of one module reads as obvious redundancy — and "simplifying" it produces a failure that looks nothing like its cause.
- *The directory sits outside `app/build` on purpose*, because the test workflow caches `app/build` and this is ~1.2 GB Gradle can hand over for free. Without that, `new File(rootDir, …)` in a module build file reads as a mistake.

Both are exactly the kind of comment that earns its keep: they explain a decision the code can't, and they pre-empt a plausible wrong edit. I'd restore those two even if the rest of the block stays bare.

**The CI guard never fails, and its message describes something that can't happen.** With `robolectric.offline=true` and `robolectric.dependency.dir` set, Robolectric uses the local resolver, which *throws* on a missing jar (`Path is not a file: …`, as `TESTING.md` correctly documents) — it does not fall back to Maven. So a level missing from the map leaves `~/.m2` empty and surfaces as a test failure; the branch that tells you to "add the SDK to `robolectricSdkJars`" can't be reached that way. What the check does genuinely catch is the offline flag *not taking effect* — a typo'd `-P`, a `Test` task created outside `withType(Test)`, someone dropping the flag from the workflow. That's the regression worth catching, because its only other symptom is the flake coming back. So keep the check, reword it to say that, and give it an `exit 1`: a `:warning:` line in a job summary on an otherwise-green run is a line nobody reads.

**Nits.**

- `configurations.create(name)` produces a legacy configuration that's consumable, resolvable and declarable all at once. On Gradle 9.7 the typed factory is right there — `configurations.resolvable("robolectricSdk$apiLevel")` — and it keeps this off the deprecation path.
- The staging assertion counts jars rather than checking names, so nine files with one wrong name passes and defers the failure to test time. Asserting each expected filename exists would catch a bad pin at staging, where the error can name the culprit. (A duplicated version string is already caught, since Sync would stage eight.)

**The thing I'd actually watch before merging.** This run's shards have been going 27 minutes against the ~3–4 minutes `TESTING.md` quotes, with both builds finished in five. If that's the one-time cost of pulling ~1.2 GB into a cold Gradle home, it amortises and this is a clear win. If it recurs, the trade is a flaky rerun for a permanently slow pipeline, and it's worth reconsidering — nine levels is the *correct* set, so the cost can't be trimmed by dropping entries; it would have to come from a longer-lived cache. Worth checking the Gradle-home cache size too: adding ~1.2 GB to the entry this workflow restores competes with the repo's Actions cache budget, and if it starts evicting, the sharding wins documented in `CLAUDE.md` erode silently.

**Verdict:** `ready`. The mechanism is sound and the map is right, so I'd merge this — but I'm not leaving it `merge`-labelled while its own test shards haven't reported after 27 minutes, since wall clock is the thing this change exists to protect. Confirm one green run, restore those two comments, and it's good to go.

---

dogi:
**Correction to my review above — the wall-clock caveat was wrong, and I'm withdrawing it.**

All four checks are green. The test shards finished at `11:37:14` and `11:37:34` against a `11:30:02` start — **7m11s and 7m32s**, not the 27+ minutes I reported. The check-runs API kept serving me `in_progress` well after they'd completed and I took that at face value instead of checking it against the start time. My mistake.

That reframes the only real reservation I had. ~7.5 minutes *including* a cold pull of ~1.2 GB is a sensible first-run cost, roughly double the ~3–4 minute steady state, and the download is exactly the part that amortises once the Gradle home is cached. So there's nothing to hold for here — no second confirming run needed, and the "reconsider the trade if it recurs" paragraph doesn't apply.

Everything else stands: the nine-level map is provably the right set, the offline plumbing lines up end to end, and green shards now confirm the pinned versions are the ones 4.16.1 actually asks for — the last thing the code alone couldn't settle.

Still worth doing before merge, both cheap:

- Restore the two comments `12f8cf3` removed — the one explaining why it's one configuration per level, and the one explaining why the directory sits outside `app/build`. Neither rationale exists anywhere now, and both pre-empt a plausible wrong edit.
- Reword the `~/.m2` guard to say what it actually catches (the offline flag not taking effect, not a missing SDK level) and give it an `exit 1`.

The Gradle-home cache size is still worth a glance after a few runs, since ~1.2 GB competes with the repo's Actions cache budget — but that's a thing to watch, not a reason to wait.

Rating unchanged at **84/100**, and with the shards green I'd merge it as soon as those comments are back.

---

Okuro3499 (dismissed):
```
Problem: The explanatory comments for robolectricSdkJars/the per-level configuration split/the robolectricSdkDir placement were deleted with no functional change and a          
    non-descriptive commit message ("Update build.gradle"). Without them, a future contributor bumping Robolectric or adding an SDK level has no explanation for why there are 9     
    separate configurations (merging them silently breaks older-SDK tests via dependency conflict resolution) or why the staging dir is outside app/build (it relies on Gradle's own 
    dependency cache via setup-gradle, not the actions/cache step, to avoid re-downloading). 
```

PR #15898, opened 2026-08-20
all: smoother layout change listening (fixes #16211)

### LINKED ISSUE #16211: Fix memory leak and layout thrashing in Fragments
Tracking issue for PR #15898.

Fix memory leak and redundant layout updates by correctly unregistering the layout change listener and coalescing posted span updates.

---
*PR created automatically by Jules for task [7415948109410022578](https://jules.google.com/task/7415948109410022578) started by @dogi*

dogi:
Fix memory leak and redundant layout updates by correctly unregistering the layout change listener and coalescing posted span updates.

---
*PR created automatically by Jules for task [7415948109410022578](https://jules.google.com/task/7415948109410022578) started by @dogi*


dogi:
**Review — rating: 86/100**

Three of these are genuine, and it's worth separating them from the defensive one — same as on #15894.

**`removeCallbacks(spanUpdateRunnable)` in `onDestroyView` is the real fix.** A posted `Runnable` sits in the view's message queue holding the fragment, and if it fires after teardown it reads `recyclerView.layoutManager` on a destroyed view. Cancelling pending posts closes that window — and it's only possible *because* the runnable is now a stable field rather than a fresh lambda per post.

**`removeCallbacks` before `post` is the "thrashing" fix.** The old listener queued a new runnable on every layout pass with a width delta, so a resize or animation could pile up many redundant span recalculations. Now at most one is pending.

**The `spanCount` guard is real.** `GridLayoutManager.spanCount =` triggers `requestLayout()` internally even when the value is unchanged, so comparing first genuinely skips full relayouts of the list.

**The listener removal itself is hygiene rather than a demonstrable leak** — the RecyclerView is part of the fragment's own view hierarchy, so it dies with the view either way. Still the right shape.

Adding the `::recyclerView.isInitialized` guard to `ResourcesFragment.onResume` also brings it in line with `CoursesFragment`, which already had it.

Nice detail: the cleanup goes at the *top* of `onDestroyView`, before super — which is the convention, and the opposite of what #15894 did. Worth making those two consistent.

**The duplication is my main concern.** The runnable, the layout-change listener, the `spanCount` guard, the `onResume` re-post and the `onDestroyView` teardown are now byte-identical in both fragments, and both extend `BaseRecyclerFragment` (which already owns `recyclerView`, judging by the `isInitialized` checks). This is exactly the shape that belongs in the base class — especially since #15877 and #15884 are *also* reworking LayoutManager handling in these same two files, so the grid-span logic is being touched from three directions at once. Hoisting it once would stop the two copies drifting.

**Heads-up on merge order:** #15877 and #15884 edit `setViewMode`/`updateToggleUi` and LayoutManager construction in these files. Expect textual conflicts, and re-check the `spanCount` guard still makes sense once a cached/reused LayoutManager is in play.

**Verdict:** mergeable.

---

ragilzakaria (approved):
**Rating:** 92/100

Comprehensive lifecycle and layout cleanup across `CoursesFragment` and `ResourcesFragment`:

- **Prevents Memory Leaks**: Replaces anonymous listener lambdas with dedicated `layoutChangeListener` and `spanUpdateRunnable` instances, unregistering listeners and clearing pending runnable callbacks inside `onDestroyView()`.
- **Throttles Rapid Layout Calls**: Coalesces layout updates via `recyclerView.removeCallbacks(spanUpdateRunnable)` before `recyclerView.post(spanUpdateRunnable)` to prevent layout thrashing on resize/animations.
- **Guards `spanCount` Mutation**: Adds `if (layoutManager.spanCount != currentSpan)` check before re-assigning span count, skipping redundant internal `requestLayout()` passes when the span has not changed.

**Housekeeping Note:**
- Update PR title to house style: `ui: smoother recycler layout listener cleanup (fixes #<issue_number>)`.

**Verdict:** Approved! Solid lifecycle fix and layout optimization.

Okuro3499 (approved):

PR #15900, opened 2026-08-20
resources: smoother collections view modelling (fixes #16028)

### LINKED ISSUE #16028: Refactor CollectionsFragment to use CollectionsViewModel
Tracking issue for PR #15900.

The UI fragment was inappropriately interacting with the repository directly and handling the business logic of parent/child tag maps inside the adapter initialization block.

This refactor separates concerns by:
1. Creating `CollectionsViewModel`, decorated with `@HiltViewModel`.
2. Extracting the `tagsRepository` query to the new ViewModel's `loadTags()` method.
3. Establishing a sealed `CollectionsState` (Idle, Loading, Success, Empty, Error) for safe state handling.
4. Suppressing duplicate loads by checking the active `dbType` in the ViewModel.
5. Emitting an immutable `Map<String, List<TagEntity>>` instead of the raw `HashMap`.
6. Consuming the state via `collectLatestWhenStarted` in the Fragment, safely pushing transformed lists to the `ResourcesTagsAdapter`.

---
*PR created automatically by Jules for task [15207535608124274232](https://jules.google.com/task/15207535608124274232) started by @dogi*

dogi:
The UI fragment was inappropriately interacting with the repository directly and handling the business logic of parent/child tag maps inside the adapter initialization block.

This refactor separates concerns by:
1. Creating `CollectionsViewModel`, decorated with `@HiltViewModel`.
2. Extracting the `tagsRepository` query to the new ViewModel's `loadTags()` method.
3. Establishing a sealed `CollectionsState` (Idle, Loading, Success, Empty, Error) for safe state handling.
4. Suppressing duplicate loads by checking the active `dbType` in the ViewModel.
5. Emitting an immutable `Map<String, List<TagEntity>>` instead of the raw `HashMap`.
6. Consuming the state via `collectLatestWhenStarted` in the Fragment, safely pushing transformed lists to the `ResourcesTagsAdapter`.

---
*PR created automatically by Jules for task [15207535608124274232](https://jules.google.com/task/15207535608124274232) started by @dogi*


dogi:
**Review — rating: 72/100**

Reviewed as a pair with #15917, which also edits `CollectionsFragment.kt`. **Not duplicates** — that PR optimizes `buildTagDataList`'s internals, this one changes where the tag data comes from. Complementary and both landable; #15917 is smaller and should go first.

The direction is right, and two details are genuinely good:

- **`lateinit` is properly eliminated.** `list` and `childMap` become initialized `emptyList()`/`emptyMap()`, which is what lets you delete the `!::list.isInitialized` guard from the search flow — that guard only existed to paper over the `lateinit`. Narrowing `childMap` from `HashMap` to `Map` is right too.
- **The `loadTags` re-entry guard** stops a reload wiping state on re-collection after a configuration change.

**But it ships three states the UI doesn't handle, and that's worse than not having them.** `Loading`, `Empty` and `Error` are all empty branches carrying `// Could show …` comments. The consequence on the error path is concrete: `btnOk` is only made visible in the `Success` branch, so a failed tag load leaves the dialog permanently blank with no OK button, no message, and no way to retry — a silent dead end.

That's also a behavior change worth being explicit about. Previously `setListAdapter` had no try/catch, so a repository failure propagated as an uncaught coroutine exception. Now `loadTags` catches it into `CollectionsState.Error` and the fragment discards it. Not crashing is an improvement; swallowing it into an unusable dialog is not. Either handle `Error`/`Empty` (a toast and a dismiss, or an inline message) or collapse the sealed class to the states you actually render — a placeholder comment in a `when` branch reads like the case is covered when it isn't.

**Nit:** `CollectionsViewModel` doesn't inject `DispatcherProvider`, unlike `StorageCategoryViewModel` in #15867. It works because the repository switches to IO internally, but the project convention is an injected provider so tests can substitute deterministic dispatchers.

**Nit:** no test, for a ViewModel whose logic is all cheaply testable — the `dbType`/state re-entry guard, `Empty` vs `Success` selection, and error mapping are three short assertions with a mocked `TagsRepository`.

**Nit:** `adapter` is still `lateinit`, but it's now assigned synchronously in `onViewCreated` before `setListeners()`, so the surviving `!::adapter.isInitialized` check in the search flow is dead.

**Nit:** `import javax.inject.Inject` sits after the `org.ole.planet…` imports rather than in style-guide order — same as #15858 and #15867.

**Verdict:** needs another round — render the `Error` and `Empty` states (or drop them) so a failed load isn't an unusable dialog.

---

ragilzakaria (dismissed):
**Decision:** 🛑 **Request Changes**
**Rating:** 74/100

Good architectural direction extracting `TagsRepository` from `CollectionsFragment` into a dedicated `CollectionsViewModel`, but a few items need resolution:

1. **Unhandled UI States (`Empty` / `Error`)**:
   - `CollectionsState.Loading`, `CollectionsState.Empty`, and `CollectionsState.Error` branches currently contain placeholder comments (`// Could show ...`).
   - Because `btnOk` is only made visible in `CollectionsState.Success`, an error or empty result leaves the dialog permanently blank with no feedback and no way for the user to proceed or dismiss cleanly.
   - Please handle the `Error` and `Empty` states in the UI (e.g. showing a toast/message or dismissing gracefully) or simplify the state model.

2. **Inject `DispatcherProvider`**:
   - Follow project conventions by injecting `DispatcherProvider` into `CollectionsViewModel` rather than hardcoding dispatchers, ensuring tests can supply deterministic dispatchers.

3. **Add Unit Tests**:
   - Please add `CollectionsViewModelTest.kt` to cover `loadTags` behavior across `Success`, `Empty`, `Error`, and the re-entry guard.

4. **Style Guide Import Order**:
   - Place `import javax.inject.Inject` before `org.ole.planet…` imports in `CollectionsViewModel.kt`.

**Verdict:** Request changes — please handle the error/empty UI states and add `CollectionsViewModelTest`.

dogi:

The empty branches are gone and the states now do something:

```kotlin
is CollectionsState.Empty -> { Utilities.toast(…, getString(R.string.no_data_available)); dismiss() }
is CollectionsState.Error -> { Utilities.toast(…, state.message); dismiss() }
is CollectionsState.Loading, CollectionsState.Idle -> { /* Ignore transient states … */ }
```

That closes the dead end I was worried about — a failed tag load no longer leaves a blank dialog with no OK button and no way out. Dismissing on both terminal states is the right call for a dialog, and the comment on `Loading`/`Idle` makes the omission deliberate rather than looking unfinished.

The other three points are addressed too: `DispatcherProvider` is injected, the adapter is created synchronously in `onViewCreated` (so the state collector only submits lists), and there's now a test class covering all four paths — success with a parent/child mapping, empty, error message propagation, and the `dbType` re-entry guard. Those are exactly the assertions worth having.

**Nit — the error text reaches the user raw.** `Error(e.message ?: "An error occurred")` puts a database exception string in a toast, and the fallback is a hardcoded English literal in an app that ships five translations. `R.string.error` (`Error: %1$s`) is already translated and used elsewhere for exactly this.

**Nit — two imports are now unused in the fragment.** `javax.inject.Inject` (line 12) and `org.ole.planet.myplanet.repository.TagsRepository` (line 25) had no other use than the `@Inject lateinit var tagsRepository` field this PR removes. Worth checking `kotlinx.coroutines.launch` too — `setListAdapter` was its only consumer, and what remains uses `launchIn`.

**Nit:** `if (list.isEmpty() && childMap.isEmpty())` — `childMap` is derived from `tagsWithChildren.entries`, so an empty `list` guarantees an empty `childMap`. The second clause can't independently fail.

**Nit:** `import javax.inject.Inject` in the ViewModel sits between `model.TagEntity` and `repository.TagsRepository`; and the `object` members of the sealed class could be `data object` on Kotlin 2.4.

**Nit:** the test hand-rolls an anonymous `DispatcherProvider` with five fields where the project's `TestDispatcherProvider` already exists.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #15902, opened 2026-08-20
login: smoother settings shared preference managing (fixes #16030)

### LINKED ISSUE #16030: Refactor beta_auto_download to write through SharedPrefManager
Tracking issue for PR #15902.

Removed direct SharedPreferences injection from SettingsActivity and instead created a setter in SharedPrefManager for the beta auto download feature, properly centralizing persistence responsibilities.

---
*PR created automatically by Jules for task [13458823096484357961](https://jules.google.com/task/13458823096484357961) started by @dogi*

dogi:
Removed direct SharedPreferences injection from SettingsActivity and instead created a setter in SharedPrefManager for the beta auto download feature, properly centralizing persistence responsibilities.

---
*PR created automatically by Jules for task [13458823096484357961](https://jules.google.com/task/13458823096484357961) started by @dogi*


dogi:
**Review — rating: 89/100**

Behavior-identical, and I checked the thing that could have made it otherwise: the old write went through the injected `@DefaultPreferences SharedPreferences`, while the existing `getBetaAutoDownload()` reads `PreferenceManager.getDefaultSharedPreferences(context)` directly. Had those been different stores, this refactor would have silently moved the write. They aren't — `SharedPreferencesModule.provideDefaultSharedPreferences` returns exactly `PreferenceManager.getDefaultSharedPreferences(context)`, so the setter and getter hit the same file both before and after.

The encapsulation is the right call: `"beta_auto_download"` was a magic string at three sites, the setter now sits next to its getter, and a raw `SharedPreferences` injection disappears from a UI class in favour of `SharedPrefManager` — which is the convention the rest of the codebase follows. Dropping the unused `defaultPref` field and its two imports is compiler-verified by the green build.

**Nit — finish the encapsulation with a constant.** `SharedPrefManager`'s companion object already holds every other key (`SAVED_USERS`, `MANUAL_CONFIG`, `FIRST_LAUNCH`, `VERSION_DETAIL`, …), so `"beta_auto_download"` being an inline literal in *two* methods is now the odd one out. A `private const val BETA_AUTO_DOWNLOAD = "beta_auto_download"` makes the pair impossible to drift.

**Nit:** both accessors call `PreferenceManager.getDefaultSharedPreferences(context)` on every invocation rather than sharing a lazy field. Consistent with the existing getter, so no worse than before — just noting it while you're in there.

**Heads-up:** #15870 also edits `SharedPrefManager` (`clearPreferences`, ~line 271). Different region, so it should merge cleanly.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #15903, opened 2026-08-20
resources: smoother repository dashboard view modelling (fixes #16031)

### LINKED ISSUE #16031: Dashboard: make the library load reactive like courses and teams
Tracking issue for PR #15903.

Add getMyLibraryFlow and observe it to keep library up to date

---
*PR created automatically by Jules for task [3956553458169050643](https://jules.google.com/task/3956553458169050643) started by @dogi*

dogi:
Add getMyLibraryFlow and observe it to keep library up to date

---
*PR created automatically by Jules for task [3956553458169050643](https://jules.google.com/task/3956553458169050643) started by @dogi*

## Summary by CodeRabbit

* **New Features**
  * Library content now updates automatically when changes are detected.
  * Improved handling for loading library content for specific users.

* **Bug Fixes**
  * Prevented unnecessary refreshes when library data has not changed.

* **Chores**
  * Updated the Android app version to 0.66.19.


dogi:
**Review — rating: 76/100**

The goal is right and most of the execution is good — the dashboard library card was stale until something re-triggered `loadUserContent`, and making it reactive like courses and teams is the correct fix. The new DAO method mirrors the existing query exactly and is correctly non-suspend for a `Flow`.

**Special credit for the `distinctUntilChanged` comparator.** Comparing `_id`/`_rev` pairwise is the right call here: `MyLibrary` is an `open class` with identity equality, so a bare `distinctUntilChanged()` would have been completely inert — exactly the trap #15869 fell into on the same kind of Room flow. This is the #15860 approach and it works.

**But `emptyFlow()` changes behavior on the null path, and that path looks intentional.**

`getMyLibrary` guards with `if (userId.isNullOrBlank()) return emptyList()`, so the old code emitted an empty list and **cleared** `uiState.library`. `getMyLibraryFlow` returns `emptyFlow()`, which completes without emitting — so the collector never runs and the library card **keeps whatever was already there**. Two independent signals say null is anticipated rather than theoretical: `loadUserContent(userId: String?)` is declared nullable, and the repository carries its own null guard.

On a shared offline device — which is myPlanet's core use case — that means a logout or user switch can leave the previous user's library on the dashboard. The fix is one token:

```kotlin
if (userId.isNullOrBlank()) return flowOf(emptyList())
```

That restores the clearing behavior exactly. Happy to be told the null path is unreachable in practice, but as written the guard reads like it should clear and doesn't.

**Worth knowing about the comparator's blind spot:** `_rev` only changes when CouchDB does. A purely local mutation that doesn't bump `_rev` — e.g. the `resourceOffline` flip in #15865's targeted `UPDATE` — won't be seen as a change, so the dashboard won't refresh for it. Fine for the title/count this card renders; just don't extend the comparator's use to anything that displays local-only flags.

**Nit:** `kotlinx.coroutines.flow.emptyFlow()` and `kotlinx.coroutines.flow.flowOf(...)` are written fully-qualified in the impl and the tests, while `Flow`/`flowOf` are already imported in both files.

**Verdict:** needs another round — one-token fix so a null/blank user clears the card instead of stranding the previous user's library.

---

dogi:

The one-token fix is in:

```kotlin
override fun getMyLibraryFlow(userId: String?): Flow<List<MyLibrary>> {
    if (userId.isNullOrBlank()) return flowOf(emptyList())
    return myLibraryDao.getForUserPatternFlow(userIdPattern(userId))
}
```

`flowOf(emptyList())` emits once, so the collector runs and clears `uiState.library` — matching what the old `getMyLibrary` guard did. A logout or user switch on a shared device can no longer strand the previous user's library on the dashboard, which was the whole objection.

The comparator was also rewritten from the `zip`-based form to an indexed loop with early returns, which drops the intermediate pair list per comparison while keeping the `_id`/`_rev` semantics. Still the right choice for a Room entity with identity equality — a bare `distinctUntilChanged()` here would be inert.

**Nit (carried over, now only in the tests):** the impl uses the imported `flowOf`, but `DashboardViewModelTest` still writes `kotlinx.coroutines.flow.flowOf(...)` fully qualified in two places while `flowOf` is already imported in that file and used unqualified for `getMyTeamsFlow`/`getMyCoursesFlow` right beside it.

**Nit:** `getMyLibraryFlow` is not a `suspend` function, so the test stubs should be `every { … }` rather than `coEvery { … }`. It compiles and passes either way, but `coEvery` on a plain function reads as though collection is being stubbed when it's the flow-returning call that is.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #15905, opened 2026-08-20
Refactor EnterprisesFinancesFragment to move transaction creation to ViewModel

dogi:
Refactored the `EnterprisesFinancesFragment` to follow architectural boundaries by moving the transaction creation logic into `EnterprisesFinancesViewModel`. 
- Added a `createTransaction` method and a `SharedFlow` (`transactionCreated`) in the ViewModel.
- The UI now delegates the transaction logic and observes the result stream to display toasts, rather than handling the coroutine scope directly in the dialog's positive button handler.
- Preserved a safe null-check on the captured date to avoid regressions.

---
*PR created automatically by Jules for task [810506911767809936](https://jules.google.com/task/810506911767809936) started by @dogi*

## Summary by CodeRabbit

* **New Features**
  * Added support for opening HTML resources using configured entry files, including nested paths.
  * Improved finance transaction entry with a scrollable dialog, date filters, sorting, totals, and balance warnings.
  * Added scrolling to course-progress content.
  * Added localized transaction date labels in Arabic, Spanish, French, Nepali, Somali, and English.

* **Bug Fixes**
  * Preserved existing security details when updates omit fields.
  * Prevented incomplete resource syncs from deleting valid local resources.
  * Improved handling of credentials in alternative server URLs.
  * Updated the app version to 0.65.55.


dogi:
**Review — rating: 75/100**

**The scope fix is the valuable part and the description undersells it.** The old code ran `createTransaction` in `viewLifecycleOwner.lifecycleScope`, so rotating or backgrounding the dialog mid-write **cancelled the write** — for a financial transaction, that's a real reliability bug, not just a layering wart. Moving it to `viewModelScope` means the write survives configuration changes. Moving `val capturedDate = date ?: return@setPositiveButton` out so no coroutine starts for a null date is a small bonus.

**But the result delivery has a hole in exactly the scenario you just fixed.** `MutableSharedFlow<Result<Unit>>()` is constructed with the defaults — `replay = 0`, no extra buffer — and the collector runs under `collectLatestWhenStarted`, i.e. `repeatOnLifecycle(STARTED)`, so it is cancelled while the fragment is stopped. Sequence:

1. User taps OK, rotates the device.
2. `viewModelScope` survives, the repository call completes.
3. `_transactionCreated.emit(result)` fires with **zero subscribers** → the value is dropped (replay 0 means nothing is retained).
4. The post-rotation collector subscribes to a flow that has nothing to give it.

The transaction succeeds and the user gets **no confirmation toast at all** — the very case the scope change was meant to protect. An error result vanishes the same way.

The fix already exists in this codebase, one PR over: #15867 uses `Channel(Channel.BUFFERED)` + `receiveAsFlow()` for precisely this "one-shot event that must survive a stopped collector" problem. A channel buffers the value and hands it to the next collector exactly once. `replay = 1` would also deliver it, but then the toast re-fires on every later re-collection, which is worse.

**Nit:** `FileUtils.readBytesFromUri(...)` still reads the whole image into a `ByteArray` on the main thread in the fragment before handing it over. That's pre-existing, but this refactor was well placed to fix it — passing the `Uri` to the ViewModel and reading bytes on `dispatcherProvider.io` would finish the job.

**Nit:** no test on `createTransaction`. The dropped-emission behavior above is exactly what a test asserting "result reaches the collector" would have caught.

**Verdict:** needs another round — switch the event to a `Channel`/`receiveAsFlow` so a rotation mid-write doesn't swallow the outcome.

---

dogi:
**Re-review — Rating: 75/100** (was 75)

Reviewed at hunk level — see the note on diff size below — and the event delivery is unchanged:

```kotlin
private val _transactionCreated = MutableSharedFlow<Result<Unit>>()
val transactionCreated: SharedFlow<Result<Unit>> = _transactionCreated.asSharedFlow()
…
_transactionCreated.emit(result)
```

Still `replay = 0` with no extra buffer, still emitted from `viewModelScope`, still collected under `collectLatestWhenStarted`. So the hole is exactly where it was: tap OK, rotate, the write completes in `viewModelScope`, `emit` finds zero subscribers, and with no replay and no buffer the value is discarded — `MutableSharedFlow.emit` returns immediately rather than parking the value. The transaction is written and the user is told nothing, on success or on failure. That's the scenario the scope change was made to protect, so it's worth closing before this lands.

The fix is two lines, and #15867 and #15881 in this same queue both already use it:

```kotlin
private val _transactionCreated = Channel<Result<Unit>>(Channel.BUFFERED)
val transactionCreated: Flow<Result<Unit>> = _transactionCreated.receiveAsFlow()
```

A channel holds the value for the next collector and delivers it exactly once, so a rotation mid-write defers the toast instead of losing it.

The scope fix itself is still the valuable part and still correct — `viewModelScope` means a rotation no longer cancels a financial write mid-flight.

**Separate problem: this PR can't be reviewed from its diff.** It is labelled `medium` and its own change is two files, but `base…head` comes to **38 files and ~900 added lines** — `app/build.gradle`, `AppDatabase.kt`, `SyncManager.kt`, `FileUtils.kt`, `TeamsRepositoryImpl.kt`, five `values-*/strings.xml`, six layouts. Those are master's commits, not yours: the base is pinned at `41d89f46` (Aug 19) while master is now at `373420b`, so master's own work — the `TeamsRepository` → `TeamsFinancesRepository`/`TeamsMembersRepository` split, and the `AddTransactionBinding` → `DialogAddTransactionBinding` rename with the new `ConcatAdapter`/`FinanceHeaderAdapter` header — shows up as though it were yours. Merging master in would collapse the diff to the two files you actually touched. As it stands a reviewer has to grep the diff to find your change, and the `TeamsFinancesRepository` constructor swap in the ViewModel is indistinguishable from your work.

I reviewed the `EnterprisesFinancesViewModel` and `EnterprisesFinancesFragment` hunks specifically and sampled the rest; if any of those 36 other files *is* intentional, say so and I'll look properly.

**Nit (carried over):** `FileUtils.readBytesFromUri(...)` still reads the whole image into a `ByteArray` on the main thread before handing it to the ViewModel. Passing the `Uri` through and reading on `dispatcherProvider.io` would finish the job this refactor started.

**Nit (carried over):** no test on `createTransaction`. A test asserting the result reaches a collector is what would have caught the dropped emission.

**Verdict:** needs another round — switch to `Channel`/`receiveAsFlow`, and merge master in so the diff shows your change.

---

PR #15866, opened 2026-08-20
teams: smoother notifications repository querying (fixes #16206)

dogi:
Extracted `TeamsInfoLookup` interface with four read-only methods (`getTeamLabelInfo`, `getJoinRequestInfo`, `getTeamNamesByIds`, `getJoinRequestsInfo`) from `TeamsRepository`.
Updated `TeamsRepository` to extend `TeamsInfoLookup`.
Updated `NotificationsRepositoryImpl` to inject `TeamsInfoLookup` instead of the full `TeamsRepository`.
Added `@Binds` in `RepositoryModule.kt` to bind `TeamsInfoLookup` to `TeamsRepositoryImpl`.
This narrows the dependency surface area injected into `NotificationsRepositoryImpl`.

---
*PR created automatically by Jules for task [5332752157606575293](https://jules.google.com/task/5332752157606575293) started by @dogi*

## Summary by CodeRabbit

* **Refactor**
  * Improved separation of team information lookup functionality.
  * Notifications continue retrieving team labels, names, and join request details through the updated repository structure.
  * Existing team-related behavior remains available through the reorganized interfaces.

PR #15850, opened 2026-08-19
teams: smoother voices repository editing (fixes #15991)

dogi:
Created a new `VoicesEditActions` interface with the required post editing and replying methods. Replaced all occurrences of `VoicesRepository` inside `VoicesAdapter` and `VoicesActions` with this new interface, and provided anonymous implementations of the interface in `VoicesFragment`, `TeamsVoicesFragment`, and `ReplyActivity` that proxy calls to the underlying `VoicesRepository`. Tested with existing tests.

---
*PR created automatically by Jules for task [3349045254508017818](https://jules.google.com/task/3349045254508017818) started by @dogi*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of voice post editing and replies across the app.
  * Maintained editing and reply functionality in voice feeds and reply screens.

* **Chores**
  * Updated the Android app version to 0.65.70 (build 6570).

PR #15906, opened 2026-08-20
enterprises: smoother teams repositories view modelling (fixes #16212)

### LINKED ISSUE #16212: Extract enterprise reports cluster out of TeamsRepositoryImpl
Tracking issue for PR #15906.

Extracted the 5 enterprise-specific reports methods from `TeamsRepositoryImpl` into a dedicated `EnterprisesRepository` to modularize the codebase.

---
*PR created automatically by Jules for task [14221665447251306634](https://jules.google.com/task/14221665447251306634) started by @dogi*

## Summary by CodeRabbit

* **New Features**
  * Added dedicated finance report management, including creating, editing, archiving, viewing, and CSV export.
  * Reports now display calculated financial details such as profit/loss and ending balances.

* **Bug Fixes**
  * Improved separation of enterprise report functionality for more reliable report handling.

* **Tests**
  * Added coverage to verify CSV exports include accurate dates and financial calculations.

dogi:
Extracted the 5 enterprise-specific reports methods from `TeamsRepositoryImpl` into a dedicated `EnterprisesRepository` to modularize the codebase.

---
*PR created automatically by Jules for task [14221665447251306634](https://jules.google.com/task/14221665447251306634) started by @dogi*

## Summary by CodeRabbit

* **New Features**
  * Added dedicated finance report management, including creating, editing, archiving, viewing, and CSV export.
  * Reports now display calculated financial details such as profit/loss and ending balances.

* **Bug Fixes**
  * Improved separation of enterprise report functionality for more reliable report handling.

* **Tests**
  * Added coverage to verify CSV exports include accurate dates and financial calculations.


dogi:
**Review — rating: 74/100**

I read this one in full rather than sampling — despite the `enormous` label it's ~180 moved lines plus DI wiring and one consumer update, which is tractable.

**The direction is right.** `TeamsRepositoryImpl` is the largest file in the repo and `CLAUDE.md` explicitly calls it out as a candidate for splitting by responsibility. The finance-report cluster is a coherent seam, and `EnterprisesViewModel` swaps its dependency outright rather than gaining one.

I diffed all five moved bodies against the originals. Only one differs: `exportReportsAsCsv` ends rows with `.append("\n")` where the original used `.append('\n')` — same output, one extra String, no behavioral change. Everything else is verbatim.

**But the extraction copies the plumbing instead of extracting it.**

`attachTeamImage` and `updateTeamEntityById` now exist **twice** — the originals stay in `TeamsRepositoryImpl` (other methods still need them) and identical private copies land in `EnterprisesRepositoryImpl`. Two copies of the same 10-line helpers, in a file whose whole purpose is to reduce the mess in the other one. They will drift. Either pull them into a shared helper both can use, or have `EnterprisesRepositoryImpl` depend on the narrow slice of `TeamsRepository` it needs (#15866 shows the interface-segregation pattern this codebase already uses).

**And a new file was the moment to drop what the old one had accumulated:**

- **`requireRoomEntity()` is a no-op** — `private fun MyTeam.requireRoomEntity(): MyTeam { return this }`. Copying an identity function into a fresh file is pure dead abstraction; `teamDao.upsert(reportEntry)` says the same thing.
- **`.map { it }`** at the end of `getReportsFlow`'s filter chain is a no-op copy.
- **`suspend fun getReportsFlow(...): Flow<...>`** — a function that only returns a cold flow has nothing to suspend on. #15869 just fixed exactly this for `getOfflineLogins`; a brand-new interface shouldn't be born with it.
- **`MainApplication.context`** in `attachTeamImage` is a static context grab inside a repository. Inherited, but a new class is the place to inject `@ApplicationContext` — same objection I raised on #15887.

**No test, and this is financial arithmetic.** `exportReportsAsCsv` computes `profitLoss = (sales + otherIncome) - (wages + otherExpenses)` and `endingBalance = profitLoss + beginningBalance`. The diff touches no test files at all, so these five methods were untested before and still are. A pure-function move is the cheapest, safest moment in the lifecycle to add a characterization test — and for report totals it's the kind of thing you want pinned before anyone refactors it again.

**Heads-up:** `getReportsFlow` streams `teamDao.observeAll()` and filters in Kotlin. #15874 is pushing teams-table Flow filters into SQL; worth checking whether that PR covers this query or whether moving it here puts it out of scope.

**Nit:** the `EnterprisesRepository` imports in `RepositoryModule.kt` are inserted after `EventsRepositoryImpl` — alphabetically Enterprises precedes Events.

**Verdict:** needs another round — share `attachTeamImage`/`updateTeamEntityById` instead of duplicating them, and drop the no-ops the move carried over.

---

dogi:
**Re-review — Rating: 82/100** (was 74)

Three of the four items are done, and the one I cared most about — the test — is done well.

- **Duplication is gone.** `attachTeamImage` and `updateTeamEntityById` now exist once, in `TeamsRepositoryImpl`, and `EnterprisesRepositoryImpl` calls them through an injected `TeamsRepository`. No drifting copies.
- **`.map { it }` dropped** from `getReportsFlow`.
- **`getReportsFlow` is no longer `suspend`** in the new interface — correct for a function that just returns a cold flow.
- **`exportReportsAsCsv` is now tested**, with the arithmetic spelled out (`70 - 25 = 45`, `45 + 100 = 145`) and the full CSV compared. That's exactly the characterization test worth having before anyone refactors report totals again, and it's the strongest part of this round.
- `MainApplication.context` stayed behind in `TeamsRepositoryImpl` rather than being copied into the new file, which is the better outcome — the new class is clean of it.

**The remaining objection is that the sharing widened the interface you're trying to shrink.** `attachTeamImage`, `updateTeamEntityById` and `requireRoomEntity` moved from `private` to public members of `TeamsRepository`, so every consumer of that interface now sees three internal plumbing methods. The goal was to reduce what `TeamsRepository` carries; this adds to it.

A narrow interface gives Enterprises what it needs without exposing the plumbing repo-wide — the pattern #15850 in this queue just used for `VoicesEditActions`:

```kotlin
interface TeamEntityWriter {
    suspend fun attachTeamImage(teamId: String, imageName: String, imageData: ByteArray)
    suspend fun updateTeamEntityById(id: String, updater: (MyTeam) -> Unit): Boolean
}
interface TeamsRepository : TeamEntityWriter { … }
```

`EnterprisesRepositoryImpl` then injects `TeamEntityWriter`, and nothing else changes.

**`requireRoomEntity` should be deleted, not promoted.** It's `fun MyTeam.requireRoomEntity(): MyTeam { return this }` — an identity function — and it's now an interface member invoked as `teamDao.upsert(with(teamsRepository) { reportEntry.requireRoomEntity() })`. Every implementer of `TeamsRepository` has to declare it, callers need a `with` scope, and it does nothing. `teamDao.upsert(reportEntry)` is the whole of it.

**Nit:** `EnterprisesViewModel.getReportsFlow` is still `suspend` even though the repository method no longer is — it now has nothing to suspend on.

**Nit (carried over):** `getReportsFlow` still filters `teamDao.observeAll()` in Kotlin, so every teams-table write re-emits and re-filters the whole table. Moved verbatim, so not a regression — worth checking against #15874, which is pushing these filters into SQL.

**Verdict:** mergeable — the blocking items are addressed. The narrow-interface split and deleting `requireRoomEntity` would be worth a follow-up.

---

Okuro3499 (changes requested):
```
app/src/main/java/org/ole/planet/myplanet/repository/TeamsRepository.kt / TeamsRepositoryImpl.kt, EnterprisesRepositoryImpl.kt: Remove attachTeamImage, updateTeamEntityById, and
    MyTeam.requireRoomEntity() from the TeamsRepository interface (revert them to private in TeamsRepositoryImpl). Reimplement the equivalent ~15 lines directly in                  
    EnterprisesRepositoryImpl using the TeamDao it already injects (drop the teamsRepository dependency entirely, or keep it only if a future enterprise-report operation genuinely  
    needs a TeamsRepository-level domain method). This keeps the new repository self-contained and avoids permanently widening TeamsRepository's public surface for an implementation
    detail.                                                                                                                                                                          
  - TeamsRepository.kt:6, TeamsRepositoryImpl.kt:37: Delete the now-unused import org.ole.planet.myplanet.model.FinanceReportParams from both files.                                 
```

dogi:
**Re-review — Rating: 84/100** (was 82)

The interface widening is undone — `TeamsRepository` now only *loses* the five report methods, and `attachTeamImage`/`updateTeamEntityById`/`requireRoomEntity` are private again. `getReportsFlow` stays non-suspend, `.map { it }` is still gone, and the `exportReportsAsCsv` characterization test is still there with the arithmetic spelled out. `EnterprisesRepositoryImpl` now depends on `TeamDao` directly rather than on `TeamsRepository`, which is a cleaner seam than either previous round.

I should be straight that I put you in a loop here: round 1 duplicated the two helpers and I objected; round 2 shared them by widening the interface and I objected to that; round 3 is back to duplication. That's on me for naming the problem twice without naming a third option. The third option is a small injected collaborator — something like a `TeamEntityWriter` holding `attachTeamImage` and the read-modify-write, `@Inject`ed into both impls. Neither interface widens and there's one copy.

That said, I'm not bouncing this a third time over it, because the duplication is smaller than I made it sound. I looked up `requireRoomEntity` and it is literally `private fun MyTeam.requireRoomEntity(): MyTeam { return this }` — an identity function left over from the Realm migration, called at 13 sites in `TeamsRepositoryImpl`. So dropping it in the new file is correct, not a divergence, and it reduces `updateTeamEntityById` to four trivial lines around `teamDao.getById` + `upsert`. Duplicating *that* is fine. The one genuinely worth sharing is `attachTeamImage`, because it carries file I/O and puts a `MainApplication.context` static back into the new file that round 2 had kept out.

**Nit:** `updateTeamEntityById` has a leftover `val model = entity` alias with no purpose now that `requireRoomEntity()` is gone; there's a stray double blank line above `attachTeamImage`; and `EnterprisesViewModel.getReportsFlow` is still declared `suspend` even though the repository's no longer is. Separately, deleting that no-op `requireRoomEntity` and its 13 call sites would be a clean small PR of its own.

**Verdict:** the seam and the interface are right and the test is the valuable part; ready, with `attachTeamImage` as the follow-up.

---

Okuro3499 (approved):

PR #15907, opened 2026-08-20
teams: smoother voices label view modelling (fixes #16033)

### LINKED ISSUE #16033: perf: reverse-index label map in VoicesViewModel
Tracking issue for PR #15907.

Optimizes `VoicesViewModel` by removing an O(n) linear scan over `Constants.LABELS` that occurs inside nested loops during news filtering and label collection, replacing it with an O(1) dictionary lookup.

---
*PR created automatically by Jules for task [7703412730919365557](https://jules.google.com/task/7703412730919365557) started by @dogi*

dogi:
Optimizes `VoicesViewModel` by removing an O(n) linear scan over `Constants.LABELS` that occurs inside nested loops during news filtering and label collection, replacing it with an O(1) dictionary lookup.

---
*PR created automatically by Jules for task [7703412730919365557](https://jules.google.com/task/7703412730919365557) started by @dogi*


dogi:
**Review — rating: 74/100**

Correct, but the optimization doesn't pay at the actual data size, and it's built twice.

**Correctness is fine.** `find` returns the first matching entry while `associate` keeps the last, so this would diverge if two `LABELS` keys shared a value. They don't — `Constants.LABELS` is `{"Offer" to "offer", "Help wanted" to "help", "Request for advice" to "advice"}`, three distinct values — and a miss still falls through to `VoicesLabelManager.formatLabelValue(label)` exactly as before.

**But `LABELS` has three entries.** The scan being replaced was at most 3 comparisons per label, and the replacement allocates a fresh 3-entry `HashMap` **per function call, in two separate functions**. For a news list with a couple of dozen labels that's roughly a wash; for a short list it's a small regression. The "reverse-index" framing suggests an N×M collapse, but M is 3.

**If this is worth doing, do it once.** `LABELS` is populated in `Constants`'s `init` and never changes afterwards, so the reverse index belongs next to it — computed once for the process rather than rebuilt on every label pass:

```kotlin
// Constants.kt
val LABEL_VALUE_TO_NAME: Map<String, String> by lazy { LABELS.entries.associate { it.value to it.key } }
```

Both call sites then just index into it, the duplication disappears, and the cost genuinely goes to zero instead of moving.

**Worth flagging while you're here:** `var LABELS = mutableMapOf<String, String>()` is a mutable global reassigned in `init`. That's the reason a cached reverse index feels risky at all. Making it a `val` with an immutable map would make the shared index obviously safe — a small change with a much better payoff than this one.

**Verdict:** needs another round — hoist a single reverse index into `Constants` instead of rebuilding a 3-entry map in two call sites.

---

dogi:
**Re-review — Rating: 92/100** (was 74)

Both parts done, including the one I raised as an aside:

```kotlin
val LABELS = mapOf("Offer" to "offer", "Help wanted" to "help", "Request for advice" to "advice")
val LABEL_VALUE_TO_NAME: Map<String, String> by lazy { LABELS.entries.associate { it.value to it.key } }
```

`LABELS` moves out of `init` into an immutable `val`, so the reverse index is computed once per process instead of a fresh 3-entry `HashMap` per call in two functions — the cost genuinely goes to zero rather than moving. And making `LABELS` a `val mapOf` is the better half of the change: a mutable global reassigned during `init` was the reason a cached index looked risky in the first place.

Two things the green build confirms for free: nothing else reassigned `Constants.LABELS` (it was `var`) and nothing mutated it through the `MutableMap` type, or the narrowing to `Map<String, String>` wouldn't compile. That's what made this safe to do.

Behavior is unchanged — the miss path still falls through to `VoicesLabelManager.formatLabelValue(label)`, and with three distinct values the `find`-vs-`associate` first/last-wins difference can't be observed.

**Nit:** with `LABELS` now a compile-time literal map, `by lazy` on `LABEL_VALUE_TO_NAME` buys nothing but a synchronized wrapper on first access — a plain `val` initialized in the same object would do. Harmless either way.

**Verdict:** mergeable.

---

Okuro3499 (approved):

Okuro3499 (approved):

PR #15908, opened 2026-08-20
all: smoother exam utils answering (fixes #16034)

### LINKED ISSUE #16034: Optimize ExamAnswerUtils.checkTextAnswer allocations
Tracking issue for PR #15908.

Optimize `ExamAnswerUtils.checkTextAnswer` allocations by pulling `Locale` and `ans.lowercase()` outside the loop. Added test coverage for multiple candidates.

---
*PR created automatically by Jules for task [17995620133751563826](https://jules.google.com/task/17995620133751563826) started by @dogi*

dogi:
Optimize `ExamAnswerUtils.checkTextAnswer` allocations by pulling `Locale` and `ans.lowercase()` outside the loop. Added test coverage for multiple candidates.

---
*PR created automatically by Jules for task [17995620133751563826](https://jules.google.com/task/17995620133751563826) started by @dogi*


dogi:
**Review — rating: 92/100**

Clean and unambiguously an improvement. For M correct choices the old code recomputed `ans.lowercase(...)` inside the loop, so it allocated M copies of the *answer* — typically the longer string — plus two `Locale.getDefault()` calls per iteration. Now that's one answer lowercase and M choice lowercases, with the locale resolved once. This runs per question while grading, so it's worth having.

Behavior is identical across the edges: `correctChoices == null` returned `null == true` → false, which the explicit `return false` matches; an empty list returned false both ways; and the predicate itself is unchanged.

**It also closes a latent inconsistency.** The old code called `Locale.getDefault()` separately for the answer and for each choice, so a `Locale.setDefault` on another thread mid-loop could lowercase the two sides under different locales. Unlikely, but hoisting makes the comparison self-consistent by construction.

**The added tests are the right ones**, and worth calling out — most PRs in this queue add none. A multi-choice question with mixed case exercises exactly the loop the hoisting touches, and `assertFalse("third choice is missing")` pins the negative case so the `any` can't silently become an `all` or a no-op.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #15899, opened 2026-08-20
teams: smoother voices item callback diffing (fixes #16027)

dogi:
Removed unnecessary memory allocations in `VoicesAdapter.kt` by directly comparing `labels` and `imageUrls` fields using Kotlin's native `==` operator, eliminating the redundant `toList()` copies during list diffing operations.

---
*PR created automatically by Jules for task [15335797016179556313](https://jules.google.com/task/15335797016179556313) started by @dogi*

PR #15910, opened 2026-08-20
teams: smoother voices reply counting (fixes #16036)

### LINKED ISSUE #16036: perf(voices): eliminate per-bind string allocations in VoicesAdapter
Tracking issue for PR #15910.

Addresses performance bottleneck in `VoicesAdapter` by eliminating per-bind string allocations.

- Added `<string name="reply_count_format" translatable="false">(%1$d)</string>` to `strings.xml`.
- Replaced `String.format` in `VoicesAdapter.kt` with `context.getString()`.
- Replaced `.lowercase(...).endsWith(".gif")` with `isGif()` helper using `ignoreCase = true`.

---
*PR created automatically by Jules for task [6461055045260487223](https://jules.google.com/task/6461055045260487223) started by @dogi*

dogi:
Addresses performance bottleneck in `VoicesAdapter` by eliminating per-bind string allocations.

- Added `<string name="reply_count_format" translatable="false">(%1$d)</string>` to `strings.xml`.
- Replaced `String.format` in `VoicesAdapter.kt` with `context.getString()`.
- Replaced `.lowercase(...).endsWith(".gif")` with `isGif()` helper using `ignoreCase = true`.

---
*PR created automatically by Jules for task [6461055045260487223](https://jules.google.com/task/6461055045260487223) started by @dogi*


dogi:
**Review — rating: 86/100**

The `isGif` change is worth more than the title claims, and the other hunk claims something it doesn't do.

**`isGif` is both an allocation win and a latent i18n bug fix.** `path.lowercase(Locale.getDefault())` allocated a full lowercased copy of the file path on every bind just to test a four-character suffix — `endsWith(".gif", ignoreCase = true)` allocates nothing. But the more interesting part: `lowercase(Locale.getDefault())` is locale-sensitive, and under a Turkish locale `'I'` lowercases to dotless `'ı'`, so `".GIF".lowercase(tr_TR)` becomes `".gıf"` and the check **fails**. A `.GIF` on a Turkish-locale device was being loaded down the non-GIF path — a static first frame instead of an animation. `ignoreCase = true` uses locale-independent case folding, so that's fixed. Worth leading with in the description; this app ships in six languages and the Turkish-I trap hides for years.

Extracting the helper also de-duplicates the two call sites, which is how the second one stays correct.

**The `reply_count_format` hunk isn't a perf change, though.** `context.getString(id, arg)` still runs `String.format` internally and still allocates a String, plus it adds a resource lookup — so no allocation is eliminated. What it genuinely changes is *which* locale formats the number: `String.format(Locale.getDefault(), …)` uses the **system** locale, while `getString` formats with the **configuration** locale. Since this app overrides the app language per user via `LocaleUtils.onAttach`, the config locale is the more correct one — the digits in `(%d)` now follow the app's language rather than the device's. Small correctness improvement, so keep it; just don't file it under allocations.

`translatable="false"` on a pure-punctuation format string is the right call — it stays out of the five locale files and keeps lint quiet.

**Nit:** `import java.util.Locale` is now unused. It had exactly three uses in this file (lines 703, 889, 982) and this PR removes all three. Kotlin only warns, so CI stays green.

**Heads-up:** #15899 touches this file's DiffUtil block (~lines 80–100) and #15850 its constructor. Different regions from yours (700, 883–990), so they should merge cleanly.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #15911, opened 2026-08-20
community: smoother leaders view modelling (fixes #16037)

### LINKED ISSUE #16037: perf: move LeadersViewModel init parsing off main thread
Tracking issue for PR #15911.

Fixes an issue where `LeadersViewModel` parses the initial JSON string synchronously on the main thread during `init`. It achieves this by injecting `DispatcherProvider` and offloading the logic in `loadLeaders` to `viewModelScope.launch(dispatcherProvider.default)`.

---
*PR created automatically by Jules for task [16122232876135416461](https://jules.google.com/task/16122232876135416461) started by @dogi*

dogi:
Fixes an issue where `LeadersViewModel` parses the initial JSON string synchronously on the main thread during `init`. It achieves this by injecting `DispatcherProvider` and offloading the logic in `loadLeaders` to `viewModelScope.launch(dispatcherProvider.default)`.

---
*PR created automatically by Jules for task [16122232876135416461](https://jules.google.com/task/16122232876135416461) started by @dogi*

## Summary by CodeRabbit

* **New Features**
  * Updated the app version to 0.66.13.

* **Bug Fixes**
  * Improved the community leaders screen to handle unavailable data safely.
  * Prevented UI updates when leader data is missing.
  * Improved asynchronous loading of community leaders for a smoother experience.


dogi:
**Review — rating: 74/100**

Right instinct — `UserEntity.parseLeadersJson` is a Gson parse running in the ViewModel's `init`, i.e. on the main thread during fragment creation — and injecting `DispatcherProvider` rather than hard-coding a dispatcher matches the project convention.

**But making it async introduces a guaranteed flash of "No data available".**

`_leaders` is a `StateFlow` initialised to `emptyList()`, and `LeadersFragment` collects it like this:

```kotlin
collectWhenStarted(viewModel.leaders) { leadersList ->
    if (leadersList.isEmpty()) {
        binding?.tvNodata?.let { it.text = getString(R.string.no_data_available) }
    } else { … }
}
```

Before this change, `loadLeaders()` completed synchronously inside `init`, so the collector's **first** emission was already the parsed list and the empty-state text never appeared for a populated community. Now the first emission is always `emptyList()`, so every visit to the Leaders screen renders "No data available" and then replaces it once the parse lands. Since `StateFlow` replays its current value on subscribe, this is deterministic rather than a race — it will happen every time.

That's a poor trade: a small main-thread hitch swapped for a visible wrong message on every screen entry.

The fix is to make "not loaded yet" distinguishable from "genuinely empty". Simplest version — a nullable initial value:

```kotlin
private val _leaders = MutableStateFlow<List<UserEntity>?>(null)
```

and have the fragment show the empty text only for a non-null empty list. A small sealed state (Loading / Success / Empty) works too, though after #15900 I'd rather see two states that are both handled than four where three are stubs.

**Nit:** `dispatcherProvider.default` is right for the Gson parse; the `getCommunityLeaders()` prefs read that precedes it is nominally I/O, though it's memory-cached after first load, so this is fine as-is.

**Nit:** no `LeadersViewModelTest` exists, and the async behavior is exactly what one would pin — that the flow eventually emits the parsed leaders, and what it emits before that.

**Verdict:** needs another round — distinguish the pre-load state so the empty message doesn't flash.

---

dogi:
**Re-review — Rating: 90/100** (was 74)

The pre-load state is distinguishable now, which was the whole objection:

```kotlin
private val _leaders = MutableStateFlow<List<UserEntity>?>(null)
…
collectWhenStarted(viewModel.leaders) { leadersList ->
    if (leadersList == null) return@collectWhenStarted
    …
}
```

`null` means "not parsed yet" and `emptyList()` means "genuinely no leaders", so the first replayed emission no longer renders "No data available". The guaranteed flash on every visit to the Leaders screen is gone, and the empty text still shows for a community that really has none.

The original win survives intact: `UserEntity.parseLeadersJson` is a Gson parse that no longer runs on the main thread during ViewModel construction, and `DispatcherProvider` is injected rather than `Dispatchers.Default` being hard-coded.

**Nit (carried over):** still no `LeadersViewModelTest`, and the async behavior is now more worth pinning than before — that the flow starts at `null`, then emits the parsed list, and emits `emptyList()` rather than `null` when the stored JSON is blank. Three assertions with a mocked `SharedPrefManager` and `TestDispatcherProvider`.

**Nit:** `dispatcherProvider.default` covers both the prefs read and the parse. The read is nominally I/O but memory-cached after first access, so this is fine as-is — just noting it's a deliberate choice rather than an oversight.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #15913, opened 2026-08-20
all: smoother server utils configuring (fixes #16039)

### LINKED ISSUE #16039: Hoist constants in ServerConfigUtils
Tracking issue for PR #15913.

Optimize ServerConfigUtils by moving the pinMap and local network regex to private vals.

---
*PR created automatically by Jules for task [4551991252337829076](https://jules.google.com/task/4551991252337829076) started by @dogi*

dogi:
Optimize ServerConfigUtils by moving the pinMap and local network regex to private vals.

---
*PR created automatically by Jules for task [4551991252337829076](https://jules.google.com/task/4551991252337829076) started by @dogi*


dogi:
**Review — rating: 85/100**

Both hoists are correct with no behavior change: `ServerConfigUtils` is an `object`, so the two `private val`s initialise once at class init, and both are built from compile-time constants (`BuildConfig`) and a literal pattern — nothing that could vary per call.

**The regex hoist is the half that clearly earns its keep.** `Regex("^172\\.(1[6-9]|2[0-9]|3[0-1])\\..*")` was compiling a pattern on every `isLocalNetworkUrl` call. Recompiling a regex per invocation is waste you never want, independent of how hot the caller is.

The `pinMap` hoist is the same idea — an 11-entry map was allocated per `getPinForUrl` call — though that function runs about once per server selection, so the saving is theoretical.

**Since this is tagged `close?`, let me address the two objections I'd expect:**

- **"Does hoisting the PINs into a static field worsen the exposure?"** No. `CLAUDE.md` flags that the `PLANET_*_PIN` values are baked into `BuildConfig` and recoverable from the APK, but they're already `static final` String constants resident for the app's lifetime. A map holding references to them doesn't extend their residency or add a new leak — the memory-footprint delta is one small map.
- **"Is this churn on a file that's about to be rewritten?"** This is the real question. `CLAUDE.md` lists remediation for these PINs — rotate them server-side, move the values out of tracked `gradle.properties`, and replace the build-time constants with a runtime auth mechanism. If that work is queued, restructuring `pinMap` now is effort that gets discarded with it. The regex hoist survives that rewrite; the `pinMap` half doesn't.

So my read: nothing here is wrong, and if you want to keep it, keep it. If the `close?` is about avoiding churn ahead of the PIN remediation, the narrow version — hoist only `localNetworkRegex` — captures the actual win and leaves the part that's about to be rewritten alone. Your call rather than mine; I'm labelling on correctness.

**Verdict:** mergeable as-is — but a maintainer may reasonably prefer to keep only the regex hoist.

---

Okuro3499 (approved):

PR #15886, opened 2026-08-20
all: less room models text utils is more (fixes #16018)

dogi:
Replaces `TextUtils.isEmpty()` with Kotlin's `isNullOrEmpty()` and removes `android.text.TextUtils` imports in `Answer.kt`, `HealthExamination.kt`, `StepExam.kt`, and `TeamTask.kt`. This allows these Room entity models to be tested without Robolectric by keeping them free of Android framework dependencies.

---
*PR created automatically by Jules for task [15951391960438368361](https://jules.google.com/task/15951391960438368361) started by @dogi*

PR #15916, opened 2026-08-20
enterprises: smoother finances view binding (fixes #16142)

### LINKED ISSUE #16142: perf: Stop per-bind string and drawable allocations in EnterprisesFinancesAdapter
Tracking issue for PR #15916.

Optimizes `EnterprisesFinancesAdapter.onBindViewHolder` by removing two major sources of allocation in the hot path:
1. Replaced `TextUtils.equals(item.type?.lowercase(), ...)` which created a string on every bind with Kotlin's `equals(..., ignoreCase = true)`.
2. Extracted the `GradientDrawable`/`LayerDrawable` construction into a lazy `private val` property to share the single instance across all even-row view holders, and added `null` background resetting for odd rows to properly support recycling.

---
*PR created automatically by Jules for task [2283347378023897791](https://jules.google.com/task/2283347378023897791) started by @dogi*

dogi:
Optimizes `EnterprisesFinancesAdapter.onBindViewHolder` by removing two major sources of allocation in the hot path:
1. Replaced `TextUtils.equals(item.type?.lowercase(), ...)` which created a string on every bind with Kotlin's `equals(..., ignoreCase = true)`.
2. Extracted the `GradientDrawable`/`LayerDrawable` construction into a lazy `private val` property to share the single instance across all even-row view holders, and added `null` background resetting for odd rows to properly support recycling.

---
*PR created automatically by Jules for task [2283347378023897791](https://jules.google.com/task/2283347378023897791) started by @dogi*


dogi:
**Review — rating: 85/100**

Two of these are genuine bug fixes that the "perf" title hides.

**1. The `equals(ignoreCase = true)` change fixes a Turkish-locale bug.** Beyond dropping a per-bind `lowercase()` allocation, `"DEBIT".lowercase(tr_TR)` produces `"debıt"` with a dotless ı, which doesn't equal `"debit"` — so on a Turkish-locale device every debit row was being rendered as a **credit**. `equals(other, ignoreCase = true)` is locale-independent. Same trap as #15910's `isGif`, and here it was mislabelling financial rows. Null semantics are preserved: `String?.equals(String, ignoreCase)` returns false for a null receiver, matching `TextUtils.equals(null, "debit")`.

**2. `else { layout.background = null }` fixes a recycling bug.** Odd positions previously got no assignment at all, so a recycled holder that had been an even row kept its striped background — inconsistent striping while scrolling. Clearing it is correct.

**On the hoisted Drawable — correct here, but worth knowing why.** Sharing one `Drawable` instance across many Views is a pattern to be careful with: a Drawable carries a single callback and one set of bounds. It's safe in this case because `View.drawBackground()` re-sets bounds immediately before each draw and this drawable is static and stateless. The constraint to keep in mind: don't later give it an alpha, a color filter, or state-dependent behavior, and don't extend the same trick to a stateful drawable — those *do* break when shared.

`.mutate()` is unnecessary, though. It decouples a **shared** `ConstantState`, which matters for drawables inflated from resources, not for one you just constructed with `GradientDrawable()`. It runs once so it costs nothing, but it signals a hazard it doesn't actually address — and the thing it looks like it's guarding against (per-view bounds) isn't what it does.

**Nit:** `import android.text.TextUtils` is now unused — line 44 was its only use in this file. You removed the `Locale` import but not this one. Incidentally, this also advances #15886's TextUtils cleanup by one file.

**Verdict:** mergeable.

---

Okuro3499 (dismissed):
```
File: EnterprisesFinancesAdapter.kt, updateBackgroundColor (line 89-95) and the alternateColor lazy property (lines 33-41)                                                       
    Problem: A single shared Drawable instance is assigned as background to multiple concurrently-visible LinearLayout rows of varying height. Bounds/callback state is per-instance,
    not per-view, so only the last-bound row gets correct bounds; others render the border stretched/misaligned to stale bounds.                                                     
    
```

dogi:
**Re-review — Rating: 88/100** (was 85)

The Drawable is per-`FinanceViewHolder` now instead of one instance shared adapter-wide, which resolves the caveat I raised last round properly rather than just working around it. A `Drawable` carries one callback and one set of bounds, so a single shared instance across many Views is only safe while it stays static and stateless; scoping it to the holder plus the `mutate()` means nobody has to remember that constraint if someone later adds an alpha or a color filter. Ten to twenty instances instead of one is still a rounding error next to the per-bind allocation this replaces.

Both bug fixes survive the rework intact: `item.type.equals("debit", ignoreCase = true)` still kills the Turkish-locale trap where `"DEBIT".lowercase(tr_TR)` produced a dotless `ı` and rendered every debit row as a credit, and `else { holder.binding.llayout.background = null }` still clears the striping on recycled odd-position holders.

**Nit:** the rework orphaned two imports that the diff doesn't remove — `android.text.TextUtils` (line 7) had exactly one use, the `TextUtils.equals` you replaced, and `android.widget.LinearLayout` (line 11) had exactly one, the old `updateBackgroundColor(layout: LinearLayout, …)` signature. `java.util.Locale` was correctly dropped, so these two look like an oversight rather than a decision. Also `by lazy` defaults to `LazyThreadSafetyMode.SYNCHRONIZED`, which allocates a lock per holder for a property only ever touched on the main thread — `lazy(LazyThreadSafetyMode.NONE)` fits a ViewHolder field better in a PR about allocations.

**Verdict:** ready — the Drawable scoping is the right answer to the last round's note.

---

Okuro3499 (approved):

PR #15896, opened 2026-08-20
login: smoother storage breakdown indexing (fixes #16026)

dogi:
optimize file extension lookup in StorageBreakdownFragment

---
*PR created automatically by Jules for task [2823278049278303599](https://jules.google.com/task/2823278049278303599) started by @dogi*

## Summary by CodeRabbit

* **New Features**
  * Improved storage usage scanning for faster and more consistent file categorization.
  * Unknown file types are now grouped under a fallback category.

* **Chores**
  * Updated the Android app version to 0.66.3.

PR #15917, opened 2026-08-20
resources: smoother collections tag data querying (fixes #16041)

### LINKED ISSUE #16041: Optimize `CollectionsFragment.buildTagDataList` with O(1) lookups
Tracking issue for PR #15917.

Replaced nested O(N) list traversals (`.any` and `.find`) in `CollectionsFragment.buildTagDataList` with local O(1) lookups (`HashSet` for selected IDs and `HashMap` for parent items) to eliminate performance degradation when expanding large collection trees. Duplicate-ID behavior and expansion states have been correctly preserved.

---
*PR created automatically by Jules for task [16388880518082293562](https://jules.google.com/task/16388880518082293562) started by @dogi*

dogi:
Replaced nested O(N) list traversals (`.any` and `.find`) in `CollectionsFragment.buildTagDataList` with local O(1) lookups (`HashSet` for selected IDs and `HashMap` for parent items) to eliminate performance degradation when expanding large collection trees. Duplicate-ID behavior and expansion states have been correctly preserved.

---
*PR created automatically by Jules for task [16388880518082293562](https://jules.google.com/task/16388880518082293562) started by @dogi*


dogi:
**Review — rating: 90/100**

Reviewed as a pair with #15900, which also edits `CollectionsFragment.kt`. **Not duplicates** — that PR changes *where the tag data comes from* (moves loading into a new `CollectionsViewModel`) and leaves `buildTagDataList` untouched; this one changes *how the list is built* from whatever data arrives. They're complementary: #15900 calls this function on every state emission, so your optimization makes its hot path cheaper. Hunks are adjacent but non-overlapping, so both can land.

**Exactly behavior-preserving.** I went in expecting a null-handling difference from `mapNotNull` — if `TagEntity.id` were nullable, `selectedItemsList.any { it.id == parentTag.id }` would have matched two null ids while a `mapNotNull`-built set never can. It isn't: `TagEntity.id` is `@PrimaryKey var id: String = ""`, non-null. So the set contains every selected id and the predicate is identical.

The `!parentMap.containsKey(it.tag.id)` guard also preserves `find`'s **first**-match semantics, where a plain `associateBy` would have kept the last — the same detail #15893 and #15896 got right.

**The win is on an interaction path, which is what makes it worth doing.** The old version scanned `currentTagDataList` once per parent *and* `selectedItemsList` once per parent and per expanded child — and `currentTagDataList` grows as collections expand, so it degrades exactly when a user is drilling into a large tag tree. `buildTagDataList` re-runs on every tag click, expand and selection change, so this is O(n²)-ish work on each tap, now linear.

**Nit:** `mapNotNull { it.id }` can be `map { it.id }` — `id` is non-null, so the null filter is dead (and may draw a "useless null check" warning).

**Nit:** `putIfAbsent(it.tag.id, it)` says "first wins" more directly than the `containsKey` check.

**Verdict:** mergeable — and worth landing before #15900, since it's the smaller of the two.

---

Okuro3499 (approved):

PR #15918, opened 2026-08-20
resources: smoother repository detail querying (fixes #16143)

### LINKED ISSUE #16143: Refactor ResourceDetailFragment user library logic
Tracking issue for PR #15918.

Refactors `ResourceDetailFragment` to collapse the two-step logic of fetching the user model and conditionally calling `updateUserLibrary` into a single, encapsulated `setUserLibrary` call on `ResourcesRepository`.

The repository implementation internally resolves the current user via its injected `UserRepository`, checks if the requested addition/removal is actually required, and returns a success boolean. The fragment UI has been updated at both the automatic download-completion check and the manual remove/add button listener to simply pass the `resourceId` and `isAdd` boolean to the repository, reacting to the return value to trigger toasts and state updates.

---
*PR created automatically by Jules for task [13383083943677432976](https://jules.google.com/task/13383083943677432976) started by @dogi*

dogi:
Refactors `ResourceDetailFragment` to collapse the two-step logic of fetching the user model and conditionally calling `updateUserLibrary` into a single, encapsulated `setUserLibrary` call on `ResourcesRepository`.

The repository implementation internally resolves the current user via its injected `UserRepository`, checks if the requested addition/removal is actually required, and returns a success boolean. The fragment UI has been updated at both the automatic download-completion check and the manual remove/add button listener to simply pass the `resourceId` and `isAdd` boolean to the repository, reacting to the return value to trigger toasts and state updates.

---
*PR created automatically by Jules for task [13383083943677432976](https://jules.google.com/task/13383083943677432976) started by @dogi*


dogi:
**Review — rating: 74/100**

The consolidation is worth doing and it quietly fixes a real bug, but the new return value doesn't mean what the fragment thinks it means.

**What's good, and worth keeping:**

- Both call sites lose their own `userRepository.getUserModel()` calls — the add path had *two* (one before the update, one after) — so the user lookup happens once, inside the repository.
- **The toast was inverted before.** Old code toasted "added to my library" when `library.userId?.contains(currentUserId) != true` *after* assigning the update result, so a successful add left the id present and the toast never fired; it only appeared when the add hadn't taken effect. The new code toasts when the add actually happened. That's a fix, not just a move — worth calling out in the PR body, because it's a visible behavior change reviewers will notice.
- Not toasting on the no-op path (`btnRemove` when the state already matches) is better than the old unconditional toast.

**The problem: `setUserLibrary` returns `true` without checking whether the write happened.**

```kotlin
val library = getLibraryItemByResourceId(resourceId) ?: getLibraryItemById(resourceId)
…
updateUserLibrary(resourceId, userId, add)   // MyLibrary? — result discarded
return true
```

`updateUserLibrary` only writes under `myLibraryDao.getByResourceId(resourceId)?.let { … }`. So when the item is findable by **id** but not by **resourceId** — precisely the case your `?: getLibraryItemById(resourceId)` fallback exists to cover — it records the activity via `markResourceAdded`, never touches the library row, and returns whatever the two lookups find. `setUserLibrary` reports `true` anyway, so the fragment toasts "added to my library", flips `btnRemove` to the close icon, and calls `setLibraryData()` for a change that was never persisted. Next time the screen loads, the resource isn't in the library.

Checking the post-state fixes it:

```kotlin
val updated = updateUserLibrary(resourceId, userId, add) ?: return false
return (updated.userId?.contains(userId) == true) == add
```

**Related: `false` now means two different things** — "already in that state, nothing to do" and (once the above is fixed) "the update didn't stick". The fragment treats both as "say nothing", which is right for the first and wrong for the second. Returning the updated `MyLibrary?` instead of a `Boolean` would solve both problems at once and remove the extra `fetchLibrary(id)` round-trip each call site now does — `updateUserLibrary` already hands back the fresh item.

**Nit:** no test for `setUserLibrary`, and it has four branches worth pinning (no user, already-in-state for add, already-in-state for remove, successful toggle). `ResourcesRepositoryImplTest` already exists.

**Nit:** the remove path's toast is assembled by string concatenation — `getString(R.string.resources) + " " + getString(...)` — which was pre-existing, but it won't read correctly in the RTL locale this app ships (`ar`). A single parameterized string resource would.

**Verdict:** needs another round — make the `Boolean` reflect whether the library row actually changed.

---

dogi:

You took the `MyLibrary?` option, which solves both problems at once:

```kotlin
val updated = updateUserLibrary(resourceId, userId, add) ?: return null
return if ((updated.userId?.contains(userId) == true) == add) updated else null
```

The post-state is verified against the request, so the case I was worried about — an item findable by `id` but not by `resourceId`, where `updateUserLibrary` records the activity but never writes the library row — now returns `null` instead of a confident `true`. And returning the item rather than a Boolean removes the extra `fetchLibrary(id)` round-trip from both call sites, and separates "already in that state" (returns the existing library) from "the write didn't stick" (returns null). `false` no longer means two things.

The four tests are the four branches I'd have asked for: no user, already-added no-op, already-removed no-op, and a successful toggle asserting both the `upsert` and the resulting membership.

**The RTL concatenation is properly fixed** — `getString(R.string.resources) + " " + …` is replaced by `format_added_to_mylibrary` / `format_removed_from_mylibrary` with a `%1$s` placeholder. That's the right shape.

**But the two new strings only exist in `values/strings.xml`.** `values-ar`, `values-es`, `values-fr`, `values-ne` and `values-so` don't have them, so all five shipped locales fall back to English for this toast — including Arabic, whose RTL layout was the reason for the change. `CLAUDE.md` asks for the translation in each `values-{lang}/strings.xml` alongside the default. Since `added_to_my_library` and `removed_from_mylibrary` are already translated, the new format strings can reuse those translations with the placeholder inserted.

**Nit:** the add path decides whether to toast by comparing sizes — `library.userId?.size != updated.userId?.size`. Since `setUserLibrary` now returns the *existing* item on a no-op and `null` on failure, the membership check is more direct than the size proxy: `updated.userId?.contains(userId) != library.userId?.contains(userId)`, or just have the repository tell you (a small result type, or a nullable-on-no-op variant).

**Nit:** the new tests write `org.ole.planet.myplanet.model.UserEntity()` fully qualified four times; the file already imports `MyLibrary` from that package.

**Verdict:** mergeable — please add the five translations.

---

Okuro3499 (dismissed):
```
app/src/main/res/values/strings.xml:581-582 — format_added_to_mylibrary and format_removed_from_mylibrary were added only to the default (English) strings file. Per this repo's 
    documented localization convention (CLAUDE.md → Localization), every new string must also be added to values-ar/, values-es/, values-fr/, values-ne/, values-so/strings.xml. None
    of the five language files were updated (verified via grep — zero matches). Fix: add translated (or at minimum English-fallback) entries for both keys to each                   
    values-{lang}/strings.xml
```

dogi:
**Re-review — Rating: 90/100** (was 88)

The localization gap is closed: `format_added_to_mylibrary` and `format_removed_from_mylibrary` now exist in all six string files, not just `values/`, so `ar`, `es`, `fr`, `ne` and `so` get real translations instead of falling back to English. That was the only thing outstanding.

The rest of this round tightens the toast logic further, correctly. The add path now only toasts and swaps `btnRemove` when the membership actually changed, and the remove path moved its toast and `setLibraryData()` inside the `updated != null` branch — previously it fired unconditionally, which was the same class of bug as the inverted toast this PR originally fixed. Setting `contentDescription` alongside the icon swap is a nice extra; the icon was changing meaning without the accessibility label following it.

**Nit:** `val changed = library.userId?.size != updated.userId?.size` infers membership change from list length. It works — `setUserLibrary`'s lookups go through the DAO, so `updated` is a freshly materialised instance and not the fragment's `library` — but it would break silently if the repository ever returned the same object the fragment already holds, since `changed` is computed after the mutation would have landed. A membership check would be direct, though it'd mean plumbing the user id back into the fragment, which is exactly what this PR removed. Fine as-is; worth a comment saying why size is the proxy.

**Verdict:** ready.

---

Okuro3499 (approved):

PR #15919, opened 2026-08-20
fix(repository): chunk IN-clause parameters in TagsRepositoryImpl

dogi:
Chunked `tagDao.getByIds` in `TagsRepositoryImpl.kt` at `allTagIds` and `tagIds` usage sites via `.chunked(900).flatMap {}`.

---
*PR created automatically by Jules for task [13115458719460808514](https://jules.google.com/task/13115458719460808514) started by @dogi*

## Summary by CodeRabbit

  * Improved tag searches and linked-tag retrieval when working with large sets of tags or resources.
  * Large requests are now processed in smaller batches to provide more reliable results.
  * Existing behavior for empty searches and returned results remains unchanged.

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/repository/TagsRepositoryImpl.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/TagsRepositoryImpl.kt
@@ -55,19 +55,22 @@ class TagsRepositoryImpl @Inject constructor(
     }
 
     override suspend fun getLinkIdsForTagNames(dbType: String, tagNames: List<String>): List<String> {
-        val matchingTagIds = tagDao.getByNames(tagNames).map { it.id }
+        val matchingTagIds = chunkedIn(tagNames) { tagDao.getByNames(it) }.map { it.id }
         if (matchingTagIds.isEmpty()) {
             return emptyList()
         }
-        return tagDao.getByDbAndTagIds(dbType, matchingTagIds).mapNotNull { it.linkId }
+        return chunkedIn(matchingTagIds) { tagDao.getByDbAndTagIds(dbType, it) }.mapNotNull { it.linkId }
     }
+
+    private suspend fun <T> chunkedIn(ids: List<String>, query: suspend (List<String>) -> List<T>): List<T> =
+        if (ids.size <= 900) query(ids) else ids.chunked(900).flatMap { query(it) }
 
     private suspend fun getLinkedTagsBulk(db: String, linkIds: List<String>): Map<String, List<TagEntity>> {
         if (linkIds.isEmpty()) {
             return emptyMap()
         }
 
-        val links = tagDao.getByDbAndLinkIds(db, linkIds)
+        val links = chunkedIn(linkIds) { tagDao.getByDbAndLinkIds(db, it) }
         if (links.isEmpty()) {
             return emptyMap()
         }
@@ -77,7 +80,7 @@ class TagsRepositoryImpl @Inject constructor(
             return emptyMap()
         }
 
-        val parentTagsById = allTagIds.chunked(900).flatMap { tagDao.getByIds(it) }.associateBy { it.id }
+        val parentTagsById = chunkedIn(allTagIds) { tagDao.getByIds(it) }.associateBy { it.id }
 
         val tagsByLinkId = mutableMapOf<String, MutableList<TagEntity>>()
         val tagsSetByLinkId = mutableMapOf<String, MutableSet<String>>()
@@ -107,7 +110,7 @@ class TagsRepositoryImpl @Inject constructor(
             return emptyList()
         }
 
-        val parents = tagIds.chunked(900).flatMap { tagDao.getByIds(it) }
+        val parents = chunkedIn(tagIds) { tagDao.getByIds(it) }
         if (parents.isEmpty()) {
             return emptyList()
         }
--- a/app/src/main/java/org/ole/planet/myplanet/repository/TagsRepositoryImpl.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/TagsRepositoryImpl.kt
@@ -9,6 +9,9 @@ import org.ole.planet.myplanet.utils.JsonUtils
 class TagsRepositoryImpl @Inject constructor(
     private val tagDao: TagDao,
 ) : TagsRepository {
+    companion object {
+        private const val MAX_IN_PARAMS = 900 // SQLITE_MAX_VARIABLE_NUMBER is 999 on API 26-30
+    }
 
     override suspend fun getTags(dbType: String?): List<TagEntity> {
         return tagDao.getParentTags(dbType)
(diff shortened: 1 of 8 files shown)


dogi:
**Review — rating: 72/100**

Right problem, right technique, but applied to the two call sites least likely to overflow while the one most likely to is left untouched two lines above.

**The problem is real.** `SQLITE_MAX_VARIABLE_NUMBER` is 999 on the SQLite that ships with API 26–30, and Room expands `IN (:ids)` to one bind parameter per element, so an over-length list throws `SQLiteException: too many SQL variables` rather than degrading. Chunking at 900 is a good margin — it leaves room for the other bound arguments in the same statement.

**But `getLinkedTagsBulk` overflows at line 70, not line 80.** Both are in the same function, and the unchunked one takes the caller's list:

```kotlin
val links = tagDao.getByDbAndLinkIds(db, linkIds)   // line 70 — caller-supplied, unchunked
…
val allTagIds = links.mapNotNull { it.tagId }.distinct()
val parentTagsById = allTagIds.chunked(900).flatMap { tagDao.getByIds(it) }  // line 80 — chunked
```

`linkIds` traces back to the full resource list: `ResourcesRepositoryImpl.getResourceTagsBulk(ids)` → `getTagsForResources(ids)` → `getLinkedTagsBulk("resources", ids)`, reached from `getResourceListModels`. That's every resource in the library — a Planet catalogue crosses 999 routinely. Meanwhile `allTagIds` is the count of *distinct tags* on those resources, drawn from a controlled vocabulary that's typically dozens. So on the device where this matters, execution throws at line 70 and never reaches the line this PR fixed. Same for `getTagsForCourses` via `CoursesRepositoryImpl:557`.

`getLinkIdsForTagNames` has the same shape twice — `tagDao.getByNames(tagNames)` and `tagDao.getByDbAndTagIds(dbType, matchingTagIds)` are both unchunked. Lower risk (both lists are tag-sized), but worth covering while you're here so the file has no remaining unbounded `IN`.

Since this now needs applying in five places, it's worth extracting rather than repeating:

```kotlin
private suspend fun <T> chunkedIn(ids: List<String>, query: suspend (List<String>) -> List<T>): List<T> =
    if (ids.size <= CHUNK) query(ids) else ids.chunked(CHUNK).flatMap { query(it) }
```

That also keeps the single-query fast path for the common case, which the current `chunked(900).flatMap { }` gives up — it allocates a list-of-lists and a second list even for three ids.

**Nit:** `getLinkedTags` (line 110) re-orders results into `tagIds` order via `tagIds.mapNotNull { parentsById[it] }`, so chunking is order-safe there. `getLinkedTagsBulk` builds a map keyed by link id, so it's order-insensitive too. Worth stating in the PR description — "chunking cannot change results here" is the claim a reviewer most wants confirmed, and it happens to be true in both.

**Nit:** no test. `room-testing` is already a dependency and #15846 in this queue just added an in-memory DAO test; inserting 1,000 tag links and asserting the bulk lookup returns them all would pin the behavior — and would currently fail at line 70.

**Verdict:** needs another round — chunk `getByDbAndLinkIds` too, ideally through one shared helper.

---

dogi:
**Re-review — Rating: 92/100** (was 72)

All five call sites are covered now, through the helper rather than five copies of `.chunked(900).flatMap { }`:

```kotlin
private suspend fun <T> chunkedIn(ids: List<String>, query: suspend (List<String>) -> List<T>): List<T> =
    if (ids.size <= 900) query(ids) else ids.chunked(900).flatMap { query(it) }
```

The `ids.size <= 900` fast path is the detail that makes this strictly better than the first version — a three-element lookup issues one query and allocates nothing extra, where `chunked(900).flatMap { }` unconditionally built a list-of-lists plus a result list.

Most importantly `getByDbAndLinkIds` (line 70) is wrapped, which was the actual overflow. That's the one taking the caller's full resource list via `getResourceListModels` → `getTagsForResources`, so on API 26–30 it was the call that threw `too many SQL variables` before execution ever reached the `getByIds` that the first version chunked. `getByNames` and `getByDbAndTagIds` in `getLinkIdsForTagNames` are covered too, so the file has no unbounded `IN` left.

**Nit:** `900` is repeated twice inside the helper. A `private const val MAX_IN_PARAMS = 900` (with a one-line comment about SQLITE_MAX_VARIABLE_NUMBER being 999 on API 26–30) would explain the number and keep the two uses in step.

**Nit:** `chunkedIn` sits between `getLinkIdsForTagNames` and `getLinkedTagsBulk` rather than with the other private helpers — easy to miss when scanning for it.

**Nit (carried over):** still no test. `room-testing` is a dependency and #15846 just added an in-memory `CourseDaoTest`; inserting 1,000 tag links and asserting the bulk lookup returns all of them would pin the behavior and would have failed against the previous revision.

**Verdict:** mergeable.

---

Okuro3499 (dismissed):
```
Problem: format_added_to_mylibrary and format_removed_from_mylibrary (added in app/src/main/res/values/strings.xml:581-582) have no translated counterparts, unlike every other  
    string pair added alongside them historically. This breaks the localization convention documented in CLAUDE.md and means non-English users will see English toast text where they
    previously saw a translated one (removed_from_mylibrary/added_to_my_library were translated).
```

dogi:
**Re-review — Rating: 78/100** (was 92)

All three nits from last round are addressed, and the Kotlin change is now in good shape: `MAX_IN_PARAMS = 900` is a named constant carrying the `SQLITE_MAX_VARIABLE_NUMBER is 999 on API 26-30` rationale, `chunkedIn` moved down with the other private helpers, and there's finally a test. The five call sites — including `getByDbAndLinkIds`, the one that was actually overflowing — all go through the helper, and the `ids.size <= MAX_IN_PARAMS` fast path is still there.

**But the diff now carries six `strings.xml` files that have nothing to do with chunking IN-clause parameters**, and they collide with an open PR. `format_added_to_mylibrary` and `format_removed_from_mylibrary` are #15918's strings — that PR is still open and `ready`, and it adds the same two keys to the same six files. Whichever lands second gets a conflict across all six, and the translations aren't the same text:

| | this PR | #15918 |
|---|---|---|
| `es` | `%1$s Eliminado de miBiblioteca.` | `%1$s eliminado de myLibrary` |
| `ar` | `%1$s تمت الإزالة من مكتبتي` | `%1$s تمت إزالتها من myLibrary` |
| `fr` | `%1$s Supprimé de ma bibliothèque` | `%1$s retiré de myLibrary` |

The versions here translate the product name — "miBiblioteca", "ma bibliothèque", "Maktabaddayda" — while the `added_to_my_library` string directly above them in every one of those files keeps "myLibrary" untranslated, as does this PR's own `format_added_to_mylibrary`. So within a single file the same brand name is rendered two different ways depending on which of the two adjacent strings fires. #15918's version is the consistent one, and it's the PR that actually uses these keys — nothing in `TagsRepositoryImpl` or its test references either.

Dropping the six resource hunks (`git checkout master -- app/src/main/res/values*/strings.xml`) is the whole fix; the Kotlin side needs nothing.

**Nit:** the new test would pass against the unchunked version too. `tagDao` is a MockK mock, so nothing enforces the 999-parameter limit — one call with 1000 ids and two calls with 900+100 both satisfy the `answers` block. It does pin the cross-chunk aggregation (a `map`-instead-of-`flatMap` bug would surface), which is worth having, but adding `coVerify(exactly = 2) { tagDao.getByDbAndLinkIds("resources", any()) }` is the one line that makes it fail if the chunking is ever removed.

**Verdict:** the chunking is ready to merge; the strings aren't this PR's to ship.

---

PR #15920, opened 2026-08-20
resources: smoother view modelling (fixes #16213)

### LINKED ISSUE #16213: Refactor: Move user lookup to ResourcesViewModel in ResourcesFragment
Tracking issue for PR #15920.

Moves the user lookup into ResourcesViewModel to adhere to architectural boundaries, as User identity is a ViewModel concern. Exposes it as a StateFlow and updates ResourcesFragment to collect this flow.

---
*PR created automatically by Jules for task [4441894186217322030](https://jules.google.com/task/4441894186217322030) started by @dogi*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resource updates when the active user changes.
  * Ensured opened-resource indicators refresh correctly for the current user.
  * Improved synchronization between user state, opened resources, and the resource list.

dogi:
Moves the user lookup into ResourcesViewModel to adhere to architectural boundaries, as User identity is a ViewModel concern. Exposes it as a StateFlow and updates ResourcesFragment to collect this flow.

---
*PR created automatically by Jules for task [4441894186217322030](https://jules.google.com/task/4441894186217322030) started by @dogi*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resource updates when the active user changes.
  * Ensured opened-resource indicators refresh correctly for the current user.
  * Improved synchronization between user state, opened resources, and the resource list.


dogi:
**Review — rating: 74/100**

Moving the user lookup behind the ViewModel is right, and the observation restructure is an improvement I'd keep: splitting `openedResourceIds` into its own `repeatOnLifecycle` block means it no longer waits on the user lookup to finish before subscribing, and the nesting gets a level shallower. I also checked that re-collecting `currentUser` on every STARTED transition can't stack observers — `observeOpenedResourceIds` early-returns when `userId == currentObservedUserId && observeOpenedResourcesJob?.isActive == true`, so the repeated calls are no-ops.

**The problem is the one line in `getAdapter()`:**

```kotlin
val user = viewModel.currentUser.value   // was: userRepository.getUserModel()
```

`_currentUser` starts `null` and is filled by an `init { viewModelScope.launch { … } }` whose body suspends on `userRepository.getUserModel()`. So `.value` is a snapshot with no ordering guarantee relative to `getAdapter()`, and `user` feeds two things that are wrong when it's null:

```kotlin
isGuest = user?.isGuest() == true,      // guest → false
currentUserName = user?.name,           // → null
```

A guest whose adapter is built early gets the non-guest affordances, and `currentUserName` being null affects the ownership/edit path. The `else` branch calls `updateIdentity(...)` on later invocations, so it self-heals on the next refresh — which makes the bug intermittent and silent rather than obvious, the worse failure mode.

In practice you'll often win the race, because `getAdapter()` suspends on `viewModel.getLibraryListModels(...)` one line earlier and that DB round-trip usually outlasts the init coroutine. That's exactly why this needs fixing rather than leaving: it will pass manual testing and then misbehave on a fast device or a warm cache.

`getAdapter()` is already `suspend`, so awaiting costs nothing:

```kotlin
val user = viewModel.currentUser.filterNotNull().first()
```

or expose a `suspend fun currentUser(): UserEntity?` on the ViewModel that memoizes, and keep the `StateFlow` purely for the observing path. Either removes the race outright.

**Nit:** the first `currentUser` emission is the `null` seed, so `checkList()` now runs once with `userModel == null` before the real user arrives. Harmless as far as I can tell, but `filterNotNull()` on that collector would make the intent explicit.

**Nit:** the test adds `mockk<org.ole.planet.myplanet.repository.UserRepository>(relaxed = true)` fully qualified inline rather than importing it, and nothing asserts `currentUser` ends up populated — which is the behavior this PR introduces.

**Verdict:** needs another round — await the user in `getAdapter()` instead of sampling it.

---

dogi:
**Re-review — Rating: 91/100** (was 74)

The race is closed, and the solution is better than the one I suggested:

```kotlin
private val userDeferred: Deferred<UserEntity?> = viewModelScope.async {
    userRepository.getUserModel().also { _currentUser.value = it }
}
suspend fun getCurrentUser(): UserEntity? = userDeferred.await()
```

`getAdapter()` now awaits that instead of sampling `.value`, so `isGuest` and `currentUserName` are always built from a resolved user. I'd proposed `currentUser.filterNotNull().first()` — which would have **hung forever** if `getUserModel()` legitimately returned null. `await()` on a `Deferred<UserEntity?>` returns the null and lets the adapter build with the same semantics as before. Good catch.

Doing the lookup once as a `Deferred` also means the repeated `getAdapter()` calls (initial, refresh, view-mode change) share one query rather than each triggering their own, which master didn't do either.

The two nits are addressed too: the collector is `currentUser.filterNotNull().collectLatest { … }`, so `checkList()` no longer runs once with a null user; and the test asserts both `currentUser.value` and `getCurrentUser()` after `advanceUntilIdle()`, with `UserRepository`/`UserEntity` imported rather than inlined.

I also re-checked that re-collecting on every STARTED transition still can't stack observers — `observeOpenedResourceIds` early-returns when `userId == currentObservedUserId && observeOpenedResourcesJob?.isActive == true`.

**Nit:** the ViewModel now exposes the same value two ways — `getCurrentUser()` for one-shot awaits and `currentUser` for the collector. Both have a real caller so this isn't redundancy, but a comment saying which to use when would save the next person guessing.

**Nit:** `viewModelScope.async { }` starts eagerly at construction, so if `getUserModel()` throws, the exception is held until `getCurrentUser()` awaits it and then surfaces inside `getAdapter()`. That matches master's behavior (which called the repository directly there), so it's not a regression — just worth knowing the failure lands in the same place.

**Verdict:** mergeable.

---

Okuro3499 (dismissed):
```
Permanently memoizes the user for the ViewModel's lifetime; no way to refresh if identity changes mid-session
```

dogi:
**Re-review — Rating: 62/100** (was 91)

The head no longer contains the fix I passed last round, and what replaced it has the exact failure I flagged before that.

Last round `ResourcesViewModel` had:

```kotlin
private val userDeferred: Deferred<UserEntity?> = viewModelScope.async {
    userRepository.getUserModel().also { _currentUser.value = it }
}
suspend fun getCurrentUser(): UserEntity? = userDeferred.await()
```

and `getAdapter()` called `getCurrentUser()`. The current head has neither. It has `fetchCurrentUser()` writing into a `MutableStateFlow<UserEntity?>(null)`, and `getAdapter()` doing:

```kotlin
val user = viewModel.currentUser.filterNotNull().first()
```

That's the approach I originally proposed and then withdrew, because **it never completes when `getUserModel()` returns null**. `_currentUser` starts null and `fetchCurrentUser` only ever assigns the lookup's result, so a null user leaves `filterNotNull().first()` suspended forever and `getAdapter()` never returns — the resources list simply never renders, with no error. The old code's `user?.isGuest()` / `user?.name` handling is direct evidence that null is a state this screen expects. `await()` on a `Deferred<UserEntity?>` returns the null and lets the adapter build with the pre-PR semantics; `first()` on a filtered flow cannot.

There's a second, smaller consequence: `fetchCurrentUser()` now sits inside `repeatOnLifecycle(STARTED)`, so nothing kicks the lookup off until the fragment reaches STARTED. If `getAdapter()` runs first it parks until then — harmless when a user exists, but it widens the window for the hang above.

The rest of the round is unchanged and still good: splitting `openedResourceIds` into its own `repeatOnLifecycle` block, and `currentUser.filterNotNull().collectLatest` so `checkList()` doesn't run against a null user. I re-checked that repeated `observeOpenedResourceIds` calls still can't stack observers.

**Verdict:** restore the `Deferred`/`await()` version from the previous round — it was correct and it's the only part that regressed.

---

dogi:
**Re-review — Rating: 91/100** (was 62)

The `Deferred` version is back and the hang is gone:

```kotlin
private val userDeferred: Deferred<UserEntity?> = viewModelScope.async {
    userRepository.getUserModel().also { _currentUser.value = it }
}
suspend fun getCurrentUser(): UserEntity? = userDeferred.await()
```

`getAdapter()` calls `getCurrentUser()`, so `await()` on a `Deferred<UserEntity?>` returns the null and lets the adapter build with the pre-PR semantics, instead of `filterNotNull().first()` parking forever on a legitimately-null user. Doing the lookup once as a `Deferred` also means the repeated `getAdapter()` calls (initial, refresh, view-mode change) share one query rather than each issuing their own — better than master.

The rest holds up as before: `currentUser.filterNotNull().collectLatest` keeps `checkList()` from running against a null user, `openedResourceIds` gets its own `repeatOnLifecycle` block so it no longer waits on the user lookup to subscribe, and `observeOpenedResourceIds` still early-returns on `userId == currentObservedUserId && observeOpenedResourcesJob?.isActive == true`, so re-collecting on every STARTED transition can't stack observers. Only `filterNotNull` was added to the fragment's imports this time and it's used, so nothing is left dangling.

Worth noting the test gets a subtlety right that's easy to get wrong: because `userDeferred` is an eager property initializer, `coEvery { userRepository.getUserModel() }` has to be stubbed *before* construction, which is why it re-creates the ViewModel rather than reusing the one from `setup()`. Asserting both `currentUser.value` and `getCurrentUser()` after `advanceUntilIdle()` covers both consumers.

**Verdict:** ready — this is the version I passed two rounds ago, restored intact.

---

Okuro3499 (approved):

PR #15922, opened 2026-08-20
all: smoother user repository markdown view modelling (fixes #16043)

### LINKED ISSUE #16043: Refactor MarkdownDialogFragment to use ViewModel for user sync check
Tracking issue for PR #15922.

This PR refactors `MarkdownDialogFragment` to stop injecting `UserRepository` directly, as per the reported issue.

**Changes:**
1. **Collapsed two-step check:** Added `hasActiveUserSyncAction()` to `UserRepository` and `UserRepositoryImpl`. This method fetches the active user ID and then checks for a user sync action internally, removing the need for the UI to handle the "fetch ID -> verify ID -> check action" flow.
2. **Introduced ViewModel:** Created `MarkdownDialogViewModel` to wrap the new repository call.
3. **Refactored Fragment:** Removed the `@Inject lateinit var userRepository: UserRepository` from `MarkdownDialogFragment` and replaced it with `private val viewModel: MarkdownDialogViewModel by viewModels()`. The fragment now cleanly delegates the check to the ViewModel.

This ensures proper separation of concerns by preventing the UI from handling identity plumbing logic directly. All tests have passed.

---
*PR created automatically by Jules for task [2124924328351432905](https://jules.google.com/task/2124924328351432905) started by @dogi*

dogi:
This PR refactors `MarkdownDialogFragment` to stop injecting `UserRepository` directly, as per the reported issue.

**Changes:**
1. **Collapsed two-step check:** Added `hasActiveUserSyncAction()` to `UserRepository` and `UserRepositoryImpl`. This method fetches the active user ID and then checks for a user sync action internally, removing the need for the UI to handle the "fetch ID -> verify ID -> check action" flow.
2. **Introduced ViewModel:** Created `MarkdownDialogViewModel` to wrap the new repository call.
3. **Refactored Fragment:** Removed the `@Inject lateinit var userRepository: UserRepository` from `MarkdownDialogFragment` and replaced it with `private val viewModel: MarkdownDialogViewModel by viewModels()`. The fragment now cleanly delegates the check to the ViewModel.

This ensures proper separation of concerns by preventing the UI from handling identity plumbing logic directly. All tests have passed.

---
*PR created automatically by Jules for task [2124924328351432905](https://jules.google.com/task/2124924328351432905) started by @dogi*


dogi:
**Review — rating: 84/100**

Behavior-preserving and correctly factored at the repository end. `hasActiveUserSyncAction()` is a faithful inline of what the fragment was doing — `getActiveUserIdSuspending()`, then `hasUserSyncAction(userId)` when non-empty, `false` otherwise — so the `isCompleted` condition is unchanged. Pulling that two-step dance behind one method is worth it on its own: the "empty user id means no sync action" rule now lives in one place instead of at each call site, and `UserRepository` leaves the fragment, which is the layering `CLAUDE.md` asks for.

**The ViewModel is pure indirection, though, and it's worth deciding whether you want it.** As written it holds no state, owns no scope, and caches nothing:

```kotlin
suspend fun hasActiveUserSyncAction(): Boolean = userRepository.hasActiveUserSyncAction()
```

The call still runs in the fragment's `viewLifecycleOwner.lifecycleScope.launch`, so it's still cancelled on view destruction and still re-runs on every view recreation — none of the usual reasons for a ViewModel apply. Two options, either defensible:

1. **Make it earn its place** — expose the answer as state the fragment collects, so a rotation doesn't re-query: `val hasSyncAction: StateFlow<Boolean> = flow { emit(userRepository.hasActiveUserSyncAction()) }.stateIn(viewModelScope, …)`.
2. **Drop it** and inject the repository result via the existing pattern — the repository method is the part that carries the value.

I'd take (1), since the button-visibility computation also depends on `courseStatus` and `voiceCount` and would read naturally as derived state.

**Nit:** no test. `hasActiveUserSyncAction` has a branch (empty vs non-empty active user id) and `UserRepositoryImplTest` already exists — two assertions with a stubbed `getActiveUserIdSuspending` would pin it. The ViewModel needs none while it stays a pass-through.

**Nit:** `hasUserSyncAction(userId)` stays on the interface alongside the new method, so there are now two ways to ask nearly the same question. Fine if other callers need the explicit-id form; worth a look at whether any still do.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #15923, opened 2026-08-20
teams: smoother surveys repository adopting (fixes #16044)

### LINKED ISSUE #16044: Refactor SurveysRepositoryImpl to remove direct TeamDao lookup
Tracking issue for PR #15923.

Replaces the usage of `TeamDao` in `SurveysRepositoryImpl` with `TeamsRepository`. This properly segregates the domain logic, ensuring the surveys repository only interacts with the Teams layer through its repository interface, rather than accessing its database tables directly. Handled circular dependency injection by leveraging `dagger.Lazy`.

---
*PR created automatically by Jules for task [8679187900879146791](https://jules.google.com/task/8679187900879146791) started by @dogi*

dogi:
Replaces the usage of `TeamDao` in `SurveysRepositoryImpl` with `TeamsRepository`. This properly segregates the domain logic, ensuring the surveys repository only interacts with the Teams layer through its repository interface, rather than accessing its database tables directly. Handled circular dependency injection by leveraging `dagger.Lazy`.

---
*PR created automatically by Jules for task [8679187900879146791](https://jules.google.com/task/8679187900879146791) started by @dogi*


dogi:
**Review — rating: 82/100**

Reusing `TeamsRepository.getTeamByIdOrTeamId` instead of re-implementing the id-or-teamId fallback is the right instinct — that two-step lookup is exactly the kind of thing that drifts when only one copy gets fixed. And `dagger.Lazy` earns its place here for a reason worth stating: `TeamsRepositoryImpl` pulls in ~20 dependencies (`UploadManager`, `ServerUrlMapper`, `Gson`, several DAOs), so deferring construction keeps survey-only code paths from building that whole graph for one name lookup.

**One semantic difference to be aware of — the elvis moved.**

```kotlin
// before: falls back when the NAME is null
teamDao.getById(teamId)?.name ?: teamDao.getByTeamId(teamId)?.name
// after: falls back when the TEAM is null
getTeamByIdOrTeamId(teamId)?.name    // → getTeamEntityByAnyId: getById(id) ?: getByTeamId(id)
```

So if `getById(teamId)` returns a row whose `name` is null, the old code would go on to try `getByTeamId` while the new code stops and yields null — which skips the whole `if (!teamName.isNullOrEmpty())` block and the team survey isn't created. I went looking for a row that could trigger it and don't think one exists in practice: `getById` matches `_id`, so for a team id it finds the team document (which carries the name), and the `getByTeamId` fallback would only find membership/link rows that have no name either. So both paths end at null anyway. Worth knowing it isn't a literal refactor, though — if a name-less team document is ever possible, this changes behavior silently.

**On the framing:** `CLAUDE.md` explicitly sanctions repositories injecting the DAOs they need, so `TeamDao` here wasn't a layering violation — the actual win is de-duplicating the fallback logic, not removing a smell. Worth saying that way in the PR body so the next reader doesn't go hunting for other "direct DAO lookups" to remove.

**Nit:** two imports are now unused — `org.ole.planet.myplanet.data.room.dao.TeamDao` in both `SurveysRepositoryImpl` (line 24) and `SurveysRepositoryImplTest` (line 27). Their only usages were the lines this PR replaced.

**Nit:** `dagger.Lazy<TeamsRepository>` is written fully qualified inline while other constructor params use imported types. Both styles exist in `repository/`, but within one file an import reads better.

**Nit:** the test swaps the mock but doesn't add a case for the new path — a stub returning a team with a null name, or no team at all, would pin the boundary above.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #15925, opened 2026-08-20
courses: smoother repository step data querying (fixes #16045)

### LINKED ISSUE #16045: Refactor step prerequisites to CoursesRepository
Tracking issue for PR #15925.

Addresses issue 64: Assemble step prerequisites in `CoursesRepository`, not `CourseStepFragment`.

Changes made:
- Added `StepPrerequisites` data class to `org.ole.planet.myplanet.model`.
- Declared `getStepPrerequisites` in `CoursesRepository` and implemented it in `CoursesRepositoryImpl`.
- Refactored `CourseStepFragment` to call the new repository method to determine prerequisites rather than assembling them piecewise in the UI layer.

---
*PR created automatically by Jules for task [15851345582427776481](https://jules.google.com/task/15851345582427776481) started by @dogi*

## Summary by CodeRabbit

* **Improvements**
  * Course steps now more accurately reflect enrollment status and available exams or surveys.
  * Exam and survey controls are displayed with clearer visibility and labeling based on course content.
  * Progress saving now behaves consistently for enrolled learners.

dogi:
Addresses issue 64: Assemble step prerequisites in `CoursesRepository`, not `CourseStepFragment`.

Changes made:
- Added `StepPrerequisites` data class to `org.ole.planet.myplanet.model`.
- Declared `getStepPrerequisites` in `CoursesRepository` and implemented it in `CoursesRepositoryImpl`.
- Refactored `CourseStepFragment` to call the new repository method to determine prerequisites rather than assembling them piecewise in the UI layer.

---
*PR created automatically by Jules for task [15851345582427776481](https://jules.google.com/task/15851345582427776481) started by @dogi*

## Summary by CodeRabbit

* **Improvements**
  * Course steps now more accurately reflect enrollment status and available exams or surveys.
  * Exam and survey controls are displayed with clearer visibility and labeling based on course content.
  * Progress saving now behaves consistently for enrolled learners.


dogi:
**Review — rating: 72/100**

The structural half is a real improvement: `hideTestIfNoQuestion` no longer opens a nested `viewLifecycleOwner.lifecycleScope.launch` from inside an already-suspending load (the same anti-pattern #15858 and #15878 removed elsewhere), and `setMenuVisibility` stops launching a coroutine just to answer a question the fragment can hold. Passing the two booleans in as parameters makes the function pure UI, which is what it should be.

**But the consolidation adds three redundant queries instead of removing any.** `loadData` already calls `getCourseStepData(stepId, userId)`, which does:

```kotlin
val stepExams  = examDao.getByStepIdAndType(stepId, "courses").map { it }
val stepSurvey = examDao.getByStepIdAndType(stepId, "surveys").map { it }
val userHasCourse = isMyCourse(userId, intermediate.step.courseId)
```

and `getStepPrerequisites` then re-runs all three — both `getByStepIdAndType` calls and `isMyCourse` — before adding the two `hasSubmission` lookups and the title. So each step load now issues three extra DB round-trips, and `isMyCourse` is evaluated twice per load with `data.userHasCourse` computed and then discarded in favour of `prerequisites.isMyCourse`.

Everything the new method needs is already in scope inside `getCourseStepData`. Folding it in there gives you one repository call and no duplicate queries:

```kotlin
data class CourseStepData(
    …,
    val userHasCourse: Boolean,
    val hasExam: Boolean,
    val hasSurvey: Boolean,
    val courseTitle: String?,
)
```

with `hasExam` computed from the `stepExams` it just fetched. That's strictly less work than master, whereas the current shape is more.

**Merge-order conflict — this collides with #15854, which I just marked ready.** That PR rewrites the same `setMenuVisibility` body, replacing `lifecycleScope.launch` with `val owner = view?.let { viewLifecycleOwner } ?: return; owner.lifecycleScope.launch { … }`; this PR deletes the coroutine entirely. Whichever lands second will conflict, and the resolution isn't mechanical: if this one lands after, the guard #15854 added becomes unnecessary (there's no `viewLifecycleOwner` access left), and this version is the better outcome. Worth coordinating rather than letting the merge decide.

**One behavior note:** `setMenuVisibility` now reads the cached `userHasCourse`, which is `false` until `loadData` finishes, where it previously queried live. If the hook fires before the load completes, the progress save is skipped rather than deferred. I believe this is unreachable in practice — `setMenuVisibility` is a ViewPager1 hook and `CoursesPagerAdapter` is a ViewPager2 `FragmentStateAdapter`, which drives child state via `setMaxLifecycle` and never calls it — but that's an argument for deleting the override, not for the cached read.

**Nit:** `org.ole.planet.myplanet.model.StepPrerequisites` is written fully qualified in three places (interface declaration, impl signature, constructor call) instead of imported.

**Nit:** no test. `getStepPrerequisites` has four branches (exams present/absent × surveys present/absent) and `CoursesRepositoryImplTest` already exists.

**Nit:** the `try/catch` left in `setMenuVisibility` now only wraps `launchSaveCourseProgress()`, which launches a job rather than doing the work — so it can no longer catch what it was written to catch.

**Verdict:** needs another round — fold the prerequisites into `getCourseStepData` so this stops re-querying what it already has.

---

dogi:
**Re-review — Rating: 89/100** (was 72)

Folded in exactly as hoped. `CourseStepData` carries `hasExam`/`hasSurvey`, and `getCourseStepData` computes them from the `stepExams`/`stepSurvey` it already fetched — so the two duplicate `examDao.getByStepIdAndType` calls and the second `isMyCourse` are gone, along with the `getStepPrerequisites` method and the `StepPrerequisites` class. One repository call per step load instead of two.

It also tidied the awkward part of the original: `CourseStepData(step, resources, stepExams, stepSurvey, false)` followed by `.copy(userHasCourse = …)` is now a single named-argument constructor call, which reads better and stops allocating an intermediate.

The structural win from the first round survives: `hideTestIfNoQuestion` is a pure UI function taking two booleans, with no `viewLifecycleOwner.lifecycleScope.launch` nested inside an already-suspending load, and `setMenuVisibility` no longer opens a coroutine to answer a question the fragment can hold.

Net against master this is now strictly better — same query count as before minus the nested launch, where the previous revision was three queries worse.

**Merge-order note (still open):** #15854 is now `ready` and rewrites this same `setMenuVisibility` body, replacing `lifecycleScope.launch` with a `view?.let { viewLifecycleOwner } ?: return` guard. This PR deletes the coroutine outright, which makes that guard unnecessary. Whichever lands second will conflict, and the resolution isn't mechanical — this version is the better end state, so merging this first and dropping #15854's hunk is the cleaner path. Worth coordinating rather than letting the merge decide.

**Nit (carried over):** `setMenuVisibility` reads the cached `userHasCourse`, which is `false` until `loadData` completes, where master queried live. I still believe it's unreachable — it's a ViewPager1 hook and `CoursesPagerAdapter` is a ViewPager2 `FragmentStateAdapter`, which drives child state via `setMaxLifecycle` and never calls it — which is an argument for deleting the override rather than caching for it.

**Nit (carried over):** the surviving `try/catch` in `setMenuVisibility` now wraps only `launchSaveCourseProgress()`, which launches a job rather than doing the work, so it can no longer catch what it was written to catch.

**Nit:** no test for the two new fields, and `CoursesRepositoryImplTest` exists. Four branches (exams present/absent × surveys present/absent) and the `hasSubmission` stub is the only mock needed.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #15926, opened 2026-08-20
teams: less survey resume override is more (fixes #16046)

### LINKED ISSUE #16046: Fix: SurveyFragment redundant onResume data reload
Tracking issue for PR #15926.

This removes the blind `onResume` reload of surveys in `SurveyFragment` since the data is correctly loaded in `onViewCreated` and updates are handled by the realtime synchronization via `RealtimeSyncHelper`.

---
*PR created automatically by Jules for task [2835915195452626089](https://jules.google.com/task/2835915195452626089) started by @dogi*

dogi:
This removes the blind `onResume` reload of surveys in `SurveyFragment` since the data is correctly loaded in `onViewCreated` and updates are handled by the realtime synchronization via `RealtimeSyncHelper`.

---
*PR created automatically by Jules for task [2835915195452626089](https://jules.google.com/task/2835915195452626089) started by @dogi*


dogi:
**Review — rating: 85/100**

The duplicate load is real. `onViewCreated` already calls `loadSurveys(isTeam, teamId, false)` at line 108, and `onResume` fires immediately after on every screen entry — so each open ran the whole pipeline twice: the surveys query, `getSurveyInfos`, and `getSurveyFormState`. Removing the second one is a straight win.

I checked what still keeps the list fresh, since `_surveys` is a plain `MutableStateFlow` fed by an imperative one-shot fetch (not a Room `Flow`), so nothing auto-refreshes:

- `onViewCreated` — initial load, and again after any view recreation.
- `onDataUpdated` (line 219) — the `RealtimeSyncMixin` callback, watching `"exams"`, reloads when sync brings new data.
- The two radio `setOnClickListener`s (160, 165) — reload on the team/adopt toggle.

**It also fixes a state mismatch after rotation, which is worth knowing.** The radios use `setOnClickListener`, so `showHideRadioButton()`'s programmatic `rbTeamSurvey.isChecked = true` does *not* fire them. So on view recreation while viewing adoptable surveys, the old sequence was: `onViewCreated` loads with `isTeamShareAllowed = false`, the radio UI resets to "Team survey", then `onResume` reloaded with the ViewModel's preserved `true` — leaving the radio saying "Team survey" while the list showed adoptable ones. Without `onResume` both reset together and the UI is self-consistent. The user loses their toggle on rotation either way; now at least the label matches the list.

**The one path to verify by hand:** taking a survey changes that survey's own submission state, which feeds `surveyInfos` — and a local submission write doesn't come through `RealtimeSyncManager` (`getWatchedTables()` is `["exams"]` and `shouldAutoRefresh` returns `false`). If the survey-taking screen replaces this fragment, the return trip re-runs `onViewCreated` and reloads; if it ever opens as a dialog or activity over it, the list is now stale where `onResume` would have caught it. Worth a manual pass: take a survey, press back, confirm the row's state updates.

**Nit:** if that manual check turns up stale state, the fix isn't to restore `onResume` — it's to add `"submissions"` to `getWatchedTables()`, so the refresh is tied to the data that changed rather than to screen visibility.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #15927, opened 2026-08-20
sync: smoother upload url utils coordinating (fixes #16047)

### LINKED ISSUE #16047: perf: hoist UrlUtils.getUrl() out of batch loops in UploadCoordinator
Tracking issue for PR #15927.

Moved `UrlUtils.getUrl()` to a `baseUrl` val outside of the `batch.forEach` loop in `UploadCoordinator.uploadBatch` and `UploadCoordinator.uploadBatchRoom`. This avoids unnecessary allocations and pref reads for every item in the batch.

---
*PR created automatically by Jules for task [17550553186129510648](https://jules.google.com/task/17550553186129510648) started by @dogi*

dogi:
Moved `UrlUtils.getUrl()` to a `baseUrl` val outside of the `batch.forEach` loop in `UploadCoordinator.uploadBatch` and `UploadCoordinator.uploadBatchRoom`. This avoids unnecessary allocations and pref reads for every item in the batch.

---
*PR created automatically by Jules for task [17550553186129510648](https://jules.google.com/task/17550553186129510648) started by @dogi*


dogi:
**Review — rating: 89/100**

Reviewed as a pair with #15891, which also edits `UploadCoordinator.kt`. **Not duplicates** — that PR swaps list scans for hash lookups (lines 66, 255, 324, 440), this one hoists `UrlUtils.getUrl()` out of the per-item upload loops (149–190, 373–405). Disjoint line ranges, so both can land.

**Complete and correct.** Master has exactly six `UrlUtils.getUrl()` call sites in this file — 160, 162, 190, 383, 385, 405 — and all six are converted, so there's no half-hoisted path left in either function.

The win is real, unlike some hoists: `UrlUtils.getUrl()` resolves to `dbUrl(spm())`, which reads SharedPreferences and rebuilds the credentialed CouchDB URL on every call. These sites sit inside `batch.forEach`, so it was 1–2 calls **per uploaded item** (the second on the 409 conflict path), at `UploadConstants.BATCH_SIZE = 50` per batch across every batch of every upload run. Hoisting to one call per batch is a genuine saving, and the string interpolation was allocating a fresh URL string each time regardless.

There's also a consistency argument worth noting, same as on #15889: because `getUrl()` is pref-backed, the old code could in principle pick up a changed server URL *mid-batch* — so some items in one batch would POST to a different host than others, with no indication in the results. Pinning one base URL per batch removes that window.

**Nit:** `val baseUrl = UrlUtils.getUrl()` is computed before the `batch.forEach`, so an empty batch does one needless pref read. Irrelevant in practice — just noting it since the caller chunks and could hand over an empty list.

Same optimization as #15889 in `SyncManager`; both look right to me.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #15930, opened 2026-08-20
chat: smoother api dispatcher providing (fixes #16050)

### LINKED ISSUE #16050: Refactor ChatApiService response read to run on IO thread
Tracking issue for PR #15930.

Fixes an issue where `ChatApiService.fetchAiProviders` blocked the main thread when reading the retrofit stream.

---
*PR created automatically by Jules for task [17375963452774153930](https://jules.google.com/task/17375963452774153930) started by @dogi*

dogi:
Fixes an issue where `ChatApiService.fetchAiProviders` blocked the main thread when reading the retrofit stream.

---
*PR created automatically by Jules for task [17375963452774153930](https://jules.google.com/task/17375963452774153930) started by @dogi*


dogi:
**Review — rating: 86/100**

Correct fix in the right place. `ResponseBody.string()` is a blocking read — for a non-buffered body it pulls the remaining bytes off the socket — so it belongs on IO rather than on whatever dispatcher the caller happens to be using. `fetchAiProviders` is a `suspend fun` with no `withContext` of its own, so before this it ran on the caller's dispatcher, and a caller on `Dispatchers.Main` would block the UI thread on network I/O.

Two things I checked:

- **It's complete for this file.** `fetchAiProviders` (line 33) is the only place a body is read manually; `sendChatRequest` hands back `Response<ChatResponse>` and lets Retrofit/Gson do the deserialization on OkHttp's own threads.
- **It matches the existing convention.** `ConfigurationsRepositoryImpl.fetchApkVersionString` already does `result.data.string()` inside `withContext(dispatcherProvider.io)`, so this is the established pattern rather than a new one.

Threading `DispatcherProvider` through the constructor instead of hard-coding `Dispatchers.IO` is the right call per `CLAUDE.md`, and updating the test to pass `TestDispatcherProvider(UnconfinedTestDispatcher())` keeps it deterministic — the assertions still run without needing an explicit advance.

**Nit:** the two new imports are inserted after `org.ole.planet.myplanet.model.ChatResponse`, which breaks the file's alphabetical ordering — `kotlinx.coroutines.withContext` belongs up with the other `kotlinx`/`okhttp3` entries and `DispatcherProvider` with the `org.ole...utils` group.

**Nit:** `withContext(io)` wraps only the `string()` call, so the null/blank check and the JSON parse below stay on the caller's dispatcher. `JsonUtils` parsing of a provider map is trivial, so this is fine — but wrapping the whole body would be simpler to reason about and would make the function's dispatcher independent of its callers.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #15931, opened 2026-08-20
Optimize flow collection in Voices ViewModels

dogi:
Optimize flow collection in Voices ViewModels

Switch inner Flow collects to use `collectLatest` and `distinctUntilChanged` on list identity (ids + `_rev`) before writing state in `VoicesViewModel.observeCommunityNews` and `TeamsVoicesViewModel.observeDiscussions`. This prevents every database emission from needlessly rebuilding labels and re-running downstream `combine` filters when the underlying data hasn't structurally changed or when previous passes are still running.

---
*PR created automatically by Jules for task [13812727840857075756](https://jules.google.com/task/13812727840857075756) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/ui/teams/voices/TeamsVoicesViewModel.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/teams/voices/TeamsVoicesViewModel.kt
@@ -10,6 +10,8 @@ import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.asStateFlow
+import kotlinx.coroutines.flow.collectLatest
+import kotlinx.coroutines.flow.distinctUntilChanged
 import kotlinx.coroutines.flow.receiveAsFlow
 import kotlinx.coroutines.launch
 import org.ole.planet.myplanet.model.MyLibrary
@@ -62,7 +64,7 @@ class TeamsVoicesViewModel @Inject constructor(
     fun observeDiscussions(teamId: String) {
         observeJob?.cancel()
         observeJob = viewModelScope.launch {
-            voicesRepository.getDiscussionsByTeamIdFlow(teamId).collect {
+            voicesRepository.getDiscussionsByTeamIdFlow(teamId).distinctUntilChanged { old, new -> old.map { it.id to it._rev } == new.map { it.id to it._rev } }.collectLatest {
                 _discussions.value = it
             }
         }
--- a/app/src/main/java/org/ole/planet/myplanet/ui/teams/voices/TeamsVoicesViewModel.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/teams/voices/TeamsVoicesViewModel.kt
@@ -10,8 +10,6 @@ import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.asStateFlow
-import kotlinx.coroutines.flow.collectLatest
-import kotlinx.coroutines.flow.distinctUntilChanged
 import kotlinx.coroutines.flow.receiveAsFlow
 import kotlinx.coroutines.launch
 import org.ole.planet.myplanet.model.MyLibrary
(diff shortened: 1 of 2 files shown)


dogi:
**Review — rating: 62/100**

Both flows this wraps **already have a `distinctUntilChanged`**, and a much stronger one, applied inside `VoicesRepositoryImpl`:

```kotlin
// getCommunityNews (line 157) and getDiscussionsByTeamIdFlow (line 187)
    old.size == new.size && old.zip(new).all { (o, n) ->
        o.id == n.id && o.time == n.time && o.labels?.toSet() == n.labels?.toSet() &&
        o.message == n.message && o.isEdited == n.isEdited &&
        o.imageUrls?.toList() == n.imageUrls?.toList() && o.images == n.images &&
        o.viewIn == n.viewIn && o.viewableBy == n.viewableBy &&
        o.viewableId == n.viewableId && o.sharedBy == n.sharedBy
    }
}
```

So the new ViewModel-level predicate can never suppress a duplicate the repository hasn't already suppressed. Everything reaching `observeDiscussions`/`observeCommunityNews` is, by construction, an emission the repository judged to be a *real* change.

**That makes the weaker key actively harmful, not just redundant.** `old.map { it.id to it._rev }` compares document identity and revision only. `_rev` comes from CouchDB and is refreshed on the upload round-trip, so a **local** write doesn't bump it. Edit your own message, or toggle a label: Room updates, the repository's predicate sees `message`/`labels` differ and emits — and the new ViewModel predicate sees the same `id to _rev` pair and swallows it. The list stops updating until an unrelated emission with a changed `_rev` comes through. The repository comparing eleven fields instead of two is the evidence that id+rev was already known to be insufficient here.

**The predicate also allocates more than the code it's optimizing.** Each check builds two `List<Pair<String, String?>>` — 2N `Pair`s plus two lists per emission — where the repository's version allocates nothing per element beyond the two small sets. If allocation reduction is the goal, this moves in the wrong direction; #15932 is filing down `toSet()` calls in the same file that this adds `map{}.to Pair` allocations in front of.

**On `collect` → `collectLatest`:** harmless but inert. Neither body suspends (`_discussions.value = it`; and in `VoicesViewModel`, `collectLabels` is a plain function), so there's never a slow collector for `collectLatest` to cancel — it just adds a child-coroutine-per-emission. I checked the two-assignment body in `observeCommunityNews` for a torn-state risk between `_baseNewsList` and `_labels`; there's no suspension point between them, so that's safe either way.

**What I'd do instead:** drop this and leave the deduplication where the field knowledge lives. If the real concern is `collectLabels(filtered)` recomputing on emissions that don't change labels, dedupe *that* — e.g. assign `_labels.value` only when the computed set differs — which targets the actual cost without weakening the list's freshness.

**Verdict:** needs another round — the upstream flows already dedupe on eleven fields; adding an `id`/`_rev` filter downstream can only suppress genuine local edits.

---

dogi:
**Re-review — Rating: 70/100** (was 62)

The harmful part is gone, which matters: the `id`/`_rev` `distinctUntilChanged` is removed from both ViewModels, the `TeamsVoicesViewModel` change is dropped entirely, and `collectLatest` reverts to `collect`. Local edits that don't bump `_rev` can no longer be swallowed. Thank you for taking the "dedupe the computation instead" suggestion rather than reworking the filter.

**But what's left is a no-op.** `_labels` is a `MutableStateFlow<List<String>>`, and `MutableStateFlow` conflates by `equals` — assigning a value equal to the current one does not emit. `List<String>` has structural equality, so:

```kotlin
val newLabels = collectLabels(filtered)
if (_labels.value != newLabels) { _labels.value = newLabels }
```

does exactly what `_labels.value = newLabels` already did. The guard reads as though `StateFlow` re-emits equal values, which it doesn't — so it's likely to mislead the next person more than it helps.

It also doesn't save the cost that actually exists. `collectLabels` is `suspend … withContext(dispatcherProvider.default)`, so every emission pays a dispatch plus the label scan — and `newLabels` is computed *before* the comparison, so the guard can't skip it.

Skipping the computation would need deduplication upstream of the call — and `getCommunityNews` already applies an eleven-field `distinctUntilChanged` in `VoicesRepositoryImpl`, so by the time an emission arrives here the news list genuinely changed. There's nothing left to skip.

My honest read is that this PR has no work left in it: dropping the guard leaves an empty diff. I'd close it and keep the finding — that the labels recompute is already minimal — in the issue trail. If you'd rather land something, the one real improvement in this area is memoizing `collectLabels` on the input list identity, but given the upstream guard I don't think it pays.

**Verdict:** needs another round — either drop the redundant guard (which empties the PR, so closing is cleaner) or replace it with something that skips the computation rather than the emission.

---

PR #15933, opened 2026-08-20
dashboard: smoother activities monthly counting (fixes #16052)

### LINKED ISSUE #16052: Optimize computeMonthlyCounts in ActivitiesFragment
Tracking issue for PR #15933.

Optimized `computeMonthlyCounts` in `ActivitiesFragment.kt` to prevent allocations of throwaway lists by iterating and populating a map in a single pass using `.fold()`.

---
*PR created automatically by Jules for task [1030355350000889084](https://jules.google.com/task/1030355350000889084) started by @dogi*

dogi:
Optimized `computeMonthlyCounts` in `ActivitiesFragment.kt` to prevent allocations of throwaway lists by iterating and populating a map in a single pass using `.fold()`.

---
*PR created automatically by Jules for task [1030355350000889084](https://jules.google.com/task/1030355350000889084) started by @dogi*


dogi:
**Review — rating: 87/100**

Reviewed alongside #15937, which also edits `ActivitiesFragment.kt`. **Not duplicates** — that one caches `DateFormatSymbols().months` for `getMonth` (lines 28 and 123–127), this one rewrites `computeMonthlyCounts` (lines 61–75). No overlapping hunks, no shared logic; either merge order works.

The rewrite is equivalent, and I can say that with more confidence than usual because `ActivitiesFragmentTest` already pins the exact cases a fold rewrite could break: `loginTime = null` is skipped, an out-of-range timestamp is dropped, two January logins collapse to a count of 2, March to 1, and an empty input returns an empty map. All four assertions pass on both shards, so the old chain and the new fold agree on the null path, the boundary filter, and the counting.

The allocation argument holds too: `mapNotNull → filter → map` built three intermediate lists plus a `Grouping` and its result map, where the fold now allocates one map. And `Calendar.getInstance()` was already hoisted in master, so the per-element work is just the `timeInMillis` set and a `get`.

**Nit — the scale is small enough to say so.** `logins` is one user's login activity over the displayed range, so this is hundreds of elements at most; the win is a handful of short-lived allocations per chart render, not a measurable frame-time change. Worth stating in the PR body so the next reader doesn't go looking for a hot loop.

**Nit:** `fold` with a mutable accumulator reads a little against the grain — `fold` implies a pure reduction, and the body mutates and returns the same instance. Two more idiomatic spellings of the same thing:

```kotlin
buildMap<Int, Int> { … }.toSortedMap()          // intent-revealing
acc.merge(month, 1, Int::plus)                   // instead of acc[month] = (acc[month] ?: 0) + 1
```

`merge` is available on minSdk 26. Either is a readability preference, not a correctness point.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #15934, opened 2026-08-20
all: smoother feedback caching (fixes #16053)

### LINKED ISSUE #16053: Perf: Cache drawables and colors in FeedbackAdapter
Tracking issue for PR #15934.

Optimized `FeedbackAdapter` to prevent redundant object allocations and resource lookups during `onBindViewHolder`.

- Cached `primaryColorStateList` and `greyColorStateList` as adapter properties, initialized lazily in `onCreateViewHolder`.
- Moved the instantiation of `bg_primary` background drawables to the `FeedbackViewHolder`'s `init` block, ensuring each view gets exactly one separate instance, preventing tint bleeding across recycled rows.
- Removed inline `ContextCompat.getColor`, `ContextCompat.getDrawable`, and `ColorStateList.valueOf` calls from `onBindViewHolder`.

---
*PR created automatically by Jules for task [15326800545379681182](https://jules.google.com/task/15326800545379681182) started by @dogi*

dogi:
Optimized `FeedbackAdapter` to prevent redundant object allocations and resource lookups during `onBindViewHolder`.

- Cached `primaryColorStateList` and `greyColorStateList` as adapter properties, initialized lazily in `onCreateViewHolder`.
- Moved the instantiation of `bg_primary` background drawables to the `FeedbackViewHolder`'s `init` block, ensuring each view gets exactly one separate instance, preventing tint bleeding across recycled rows.
- Removed inline `ContextCompat.getColor`, `ContextCompat.getDrawable`, and `ColorStateList.valueOf` calls from `onBindViewHolder`.

---
*PR created automatically by Jules for task [15326800545379681182](https://jules.google.com/task/15326800545379681182) started by @dogi*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved feedback display consistency by applying status and priority colors more reliably.

* **Performance**
  * Reduced repeated color and background processing while displaying feedback items.

* **Chores**
  * Updated the Android app version to 0.66.30.


dogi:
**Review — rating: 86/100**

Both moves are sound, and the split between them is the right one: the `bg_primary` background never varies by item, so it belongs in the holder's `init` (once per holder), while the tint does vary by `priority`/`status`, so it stays in `onBindViewHolder`. That drops two `ContextCompat.getDrawable` lookups and two `getColor` lookups plus two `ColorStateList` allocations from every bind, down to two drawable lookups per holder and two colour resolutions per adapter.

I checked the two things that usually break this kind of caching:

- **Nothing else touches those backgrounds**, so setting them once per holder can't be undone by a later rebind.
- **Theme changes can't serve stale colours** — a configuration change recreates the fragment and therefore the adapter, so the cached `ColorStateList`s are re-resolved. That's what makes caching on the adapter (rather than in a `companion object`) the correct scope.

**Nit — the cached fields are nullable and flow into an API that accepts null.** `ViewCompat.setBackgroundTintList(view, null)` silently *clears* the tint, so if a bind ever preceded `onCreateViewHolder` the rows would render untinted rather than crash. That can't happen today, but `lateinit` or initializing from a context you already hold would make the invariant enforced rather than incidental.

**Nit — `greyColorStateList` holds `md_amber_500`.** The old local was called `greyColor` too, so this is inherited, but the caching made it a field with a longer reach. Worth renaming to what it means (`secondaryColorStateList`, or `inactiveColorStateList`) while it's being touched.

**Nit — the background could leave the code entirely.** Since both branches set the same `R.drawable.bg_primary`, declaring `android:background="@drawable/bg_primary"` on `tvPriority`/`tvStatus` in `row_feedback.xml` removes the holder `init` altogether and gets the same result at inflation time.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #15936, opened 2026-08-20
teams: smoother repository leader candidate querying (fixes #16214)

### LINKED ISSUE #16214: optimise getNextLeaderCandidate query
Tracking issue for PR #15936.

This PR optimises the `getNextLeaderCandidate` query. Previously, `getNextLeaderCandidate` fetched all memberships for a team and then filtered out archived members, the current leader, and an optionally excluded user in Kotlin. 

This change introduces a purpose-specific DAO query, `getEligibleNextLeaderCandidates`, that applies these three predicates in SQL. By applying the filters in SQL, we ensure that archived, current-leader, and excluded rows never leave Room, thus reducing the input set and improving performance. Ranking behaviour is unaffected.

---
*PR created automatically by Jules for task [18045469404999572358](https://jules.google.com/task/18045469404999572358) started by @dogi*

dogi:
This PR optimises the `getNextLeaderCandidate` query. Previously, `getNextLeaderCandidate` fetched all memberships for a team and then filtered out archived members, the current leader, and an optionally excluded user in Kotlin. 

This change introduces a purpose-specific DAO query, `getEligibleNextLeaderCandidates`, that applies these three predicates in SQL. By applying the filters in SQL, we ensure that archived, current-leader, and excluded rows never leave Room, thus reducing the input set and improving performance. Ranking behaviour is unaffected.

---
*PR created automatically by Jules for task [18045469404999572358](https://jules.google.com/task/18045469404999572358) started by @dogi*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved selection of next team leader candidates.
  * Excludes current leaders, archived members, and optionally a specified member from consideration.
  * Ensures candidate filtering is applied consistently when selecting a successor.


dogi:
**Review — rating: 85/100**

Reviewed as a pair with #15862, which also touches `getNextLeaderCandidate`. **Not duplicates** — this PR pushes the *member filter* into SQL, #15862 batches the *user lookup* that follows. They fix different bottlenecks in the same function and compose cleanly; expect only a textual conflict for whichever lands second.

The part I most expected to be wrong is right: **NULL status is handled.** In Kotlin, `it.status != "archived"` is `true` when `status` is null, so those rows were kept. A naive `status != 'archived'` in SQL evaluates to NULL and would have silently dropped every membership with no status — and `CLAUDE.md` calls this trap out specifically. Writing `(status IS NULL OR status != 'archived')` preserves the original semantics exactly. `!it.isLeader` → `isLeader = 0` and the `(:excludeUserId IS NULL OR userId != :excludeUserId)` guard are both faithful too.

There is one deliberate-looking asymmetry worth recording: when `userId` is NULL and `excludeUserId` is non-null, Kotlin's `null != "x"` kept the row, while `userId != :excludeUserId` evaluates to NULL and drops it. That's **benign** — the very next step is `mapNotNull { it.userId }` (or `member.userId?.let { … }`), which discards those rows anyway, so nothing observable changes. Flagging it so nobody later "fixes" it into an `OR userId IS NULL` that reintroduces dead rows.

The win is real: instead of materializing every membership row for a team and filtering in Kotlin, only eligible candidates come back.

**Nit — this is where a Room test earns its keep.** The whole change is SQL NULL semantics, which is precisely what a mocked DAO can't verify. #15859 just added `ExamDaoTest` using `Room.inMemoryDatabaseBuilder`, and the same shape here — rows with null `status`, an archived row, a leader row, and a null-`userId` row against a non-null `excludeUserId` — would pin all four boundaries cheaply. (If you copy that test, skip the `@Config(sdk = [32])` pin; #15914 is removing those.)

**Nit:** the query is a single very long line. A multi-line raw string would make the four predicates readable at a glance.

**Verdict:** mergeable — and worth landing together with #15862 rather than choosing between them.

---

Okuro3499 (approved):

PR #15937, opened 2026-08-20
dashboard: smoother activities fragment date format caching (fixes #16054)

### LINKED ISSUE #16054: perf(ActivitiesFragment): cache DateFormatSymbols().months
Tracking issue for PR #15937.

The `getMonth` function in `ActivitiesFragment` was instantiating `DateFormatSymbols` on every call, which caused unnecessary overhead because it is invoked repeatedly during chart rendering. This change caches the resulting array of month names in a `private val` and updates `getMonth` (now `internal`) to return values from the cache.

---
*PR created automatically by Jules for task [1945060620464538889](https://jules.google.com/task/1945060620464538889) started by @dogi*

dogi:
The `getMonth` function in `ActivitiesFragment` was instantiating `DateFormatSymbols` on every call, which caused unnecessary overhead because it is invoked repeatedly during chart rendering. This change caches the resulting array of month names in a `private val` and updates `getMonth` (now `internal`) to return values from the cache.

---
*PR created automatically by Jules for task [1945060620464538889](https://jules.google.com/task/1945060620464538889) started by @dogi*


dogi:
**Review — rating: 88/100**

Reviewed alongside #15933, which also edits `ActivitiesFragment.kt`. **Not duplicates** — that one rewrites `computeMonthlyCounts` (lines 61–75), this one caches the month names for `getMonth` (lines 28, 123–127). Disjoint hunks and disjoint logic; either can merge first.

This is the better-motivated of the two. `DateFormatSymbols()` loads locale data on construction, and `.months` hands back a *clone* of the internal array on every access — so the old `getMonth` paid both costs per call, and `getMonth` is the chart's X-axis `ValueFormatter` (line 105), called for every label on every render. Hoisting it to one field turns 12+ constructions-plus-clones per render into one per fragment.

Caching per instance rather than in a `companion object` is the right call: month names are locale-dependent, and a language change recreates the fragment, so a per-instance field refreshes while a static one would go stale.

The `fun` → `internal fun` change is fine and consistent — `computeMonthlyCounts` next door is already `internal`, and `ActivitiesFragmentTest.getMonth_returnsValidMonthName` lives in the same module, so it still compiles and still passes. Worth a line in the PR body, since a visibility change reads as unrelated to a caching fix.

**Nit — consider `by lazy`.** A property initializer runs at fragment *construction*, before `onAttach`. `by lazy` would defer it to the first `getMonth` call, which means a fragment that's constructed but never rendered does no locale work, and the read happens well after the host activity has applied its configuration. Same caching benefit, later and safer initialization.

**Nit:** `DateFormatSymbols.getInstance()` is the recommended factory over the constructor — it can serve a cached instance and respects installed locale providers. Since you now call it once per fragment the difference is negligible, but it's the more conventional spelling.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #15938, opened 2026-08-20
sync: smoother realtime table flowing (fixes #16055)

### LINKED ISSUE #16055: Give RealtimeSyncManager a per-table scoped flow
Tracking issue for PR #15938.

Added a new `updatesFor(table)` method to `RealtimeSyncManager` to return a `Flow` scoped to specific tables. Replaced existing manual filters (`.filter { it.table == "tableName" }`) in `ChatViewModel`, `TeamViewModel`, and `TeamDetailFragment` with this new method. This centralizes the logic, simplifies the UI/ViewModel consumer implementations, and fixes broken unit tests in `ChatViewModelTest`.

---
*PR created automatically by Jules for task [17089590333680121761](https://jules.google.com/task/17089590333680121761) started by @dogi*

dogi:
Added a new `updatesFor(table)` method to `RealtimeSyncManager` to return a `Flow` scoped to specific tables. Replaced existing manual filters (`.filter { it.table == "tableName" }`) in `ChatViewModel`, `TeamViewModel`, and `TeamDetailFragment` with this new method. This centralizes the logic, simplifies the UI/ViewModel consumer implementations, and fixes broken unit tests in `ChatViewModelTest`.

---
*PR created automatically by Jules for task [17089590333680121761](https://jules.google.com/task/17089590333680121761) started by @dogi*


dogi:
**Review — rating: 85/100**

Good API addition — it puts the table-name filter next to the flow that knows about tables, instead of leaving every consumer to spell `it.table == "…"` by hand.

The one change here that could have gone wrong is narrowing `TeamViewModel.getTeamUpdateFlow()` from the unfiltered `dataUpdateFlow` to `updatesFor("teams")`, since a caller filtering for some *other* table would silently stop receiving anything. It has exactly one caller — `TeamDetailFragment:445` — and it was already filtering on `"teams"`, so moving the filter up is exactly equivalent. `ChatViewModel` likewise.

Worth being precise about what this does and doesn't buy: `filter` is applied per subscriber, so every collector still wakes on every table's emission and discards the misses. This is a readability and correctness-by-construction win, not a throughput one — worth saying so in the PR body rather than letting "scoped flow" imply the manager routes emissions.

**The main gap: two of the four inline table checks are left behind.**

```kotlin
// MyHealthFragment:251 — filters inside the collector body
collectWhenStarted(realtimeSyncManager.dataUpdateFlow) { update ->
    if (update.table == "health" && update.shouldRefreshUI) { refreshHealthData() }
}
// UserRepositoryImpl:82
realtimeSyncManager.dataUpdateFlow.filter { it.table == "achievements" && it.shouldRefreshUI }.map { }
```

Both are single-table and convert directly. Leaving them means the codebase now has two ways to express the same thing, which is the state a helper like this is meant to end. (`RealtimeSyncMixin` is correctly left alone — it filters against `getWatchedTables()` and adds `distinctUntilChanged` + `debounce`, so it genuinely wants the unfiltered stream.)

**Nit — `shouldRefreshUI` is the other half of the same boilerplate.** All four sites pair the table check with it. If most consumers only ever want refresh-worthy updates, a second helper (or `updatesFor(table, refreshOnly = true)`) would collapse the remaining `if (update.shouldRefreshUI)` bodies too.

**Nit — the test's stub is now order-sensitive in a way that can hide a mistake.** `realtimeSyncManager` is `mockk(relaxed = true)` with `every { updatesFor("chats") } returns dataUpdateFlow`; if the production string ever changed, the relaxed mock would hand back an empty flow and the test would pass while the feature was dead. `verify { realtimeSyncManager.updatesFor("chats") }` would pin the argument.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #15940, opened 2026-08-20
courses: smoother ratings repository user querying (fixes #16095)

### LINKED ISSUE #16095: refactor: remove UserRepository from RatingsRepositoryImpl
Tracking issue for PR #15940.

Remove `UserRepository` from `RatingsRepositoryImpl` to decouple session resolution from the ratings logic. Update `submitRating` to explicitly take `UserEntity` instead of `userId`. Clean up `RatingSummaryModel` and `getCourseRatingSummary` as they were redundant to `getRatingSummary`. Update consumers (`CourseDetailViewModel`, `RatingSummaryProvider`, `RatingsViewModel`) to pass the resolved session data down. Update tests accordingly.

---
*PR created automatically by Jules for task [11297625123470524748](https://jules.google.com/task/11297625123470524748) started by @dogi*

dogi:
Remove `UserRepository` from `RatingsRepositoryImpl` to decouple session resolution from the ratings logic. Update `submitRating` to explicitly take `UserEntity` instead of `userId`. Clean up `RatingSummaryModel` and `getCourseRatingSummary` as they were redundant to `getRatingSummary`. Update consumers (`CourseDetailViewModel`, `RatingSummaryProvider`, `RatingsViewModel`) to pass the resolved session data down. Update tests accordingly.

---
*PR created automatically by Jules for task [11297625123470524748](https://jules.google.com/task/11297625123470524748) started by @dogi*

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved course rating refresh behavior so rating summaries update more reliably.
  - Improved rating submission handling, including better support for user identity validation.
  - Preserved rating updates and new submissions while reducing failures caused by missing or blank user identifiers.

- **Quality Improvements**
  - Added coverage for successful rating submissions and updated rating-refresh scenarios.

- **Release**
  - Updated the app to version 0.66.62.


dogi:
**Review — rating: 85/100**

This is the right way to do this refactor, and it's genuinely subtractive rather than a shuffle. Reviewed alongside #15948, which removes `UserRepository` from `EventsRepositoryImpl` by a different mechanism — worth comparing the two.

**What makes it work: the caller already had the object.** `RatingsViewModel` holds the `UserEntity` (`_userState.value = user`), and was passing `user.id` down purely so the repository could look the *same user* back up:

```kotlin
private suspend fun findUserForRating(userId: String): UserEntity {
    val user = userRepository.getUserById(userId)
    return requireNotNull(user) { … }
}
```

Taking `user: UserEntity` deletes that round trip, the helper, and the dependency. Same for `getCourseRatingSummary` — `CourseDetailViewModel` already has `currentState.user`, so the repository method *and* the `RatingSummaryModel` wrapper both disappear. Two API surfaces removed, nothing added.

The test migration confirms it rather than papering over it: `mockUserLookup` is gone, tests pass a real `UserEntity`, and the two failure-mode tests collapse into the one that still exists (`submitRating throws when user id is blank`) — because "user not found" is no longer a thing the repository can experience.

**One thing to fix, and one to check:**

**`refreshRatings` now silently does nothing when the state has no user.**

```kotlin
val userId = user?.id?.takeIf { it.isNotBlank() } ?: user?._id
if (userId != null) {
    val summary = ratingSummaryProvider(courseId, userId)
    _uiState.value = currentState.copy(ratingSummary = summary)
}
```

Previously `getCourseRatingSummary` resolved the user itself and returned it, so the ViewModel populated `user` from the result even when the state didn't have one. Now a null `user` means no summary and no user — an `if` with no `else`, inside a `catch (e: Exception) { // Optionally handle error }`. It should be reachable only if `loadCourseDetail` left `user` null, but that's exactly the case the old code covered. Worth either falling back to a lookup or making the branch explicit so it can't be mistaken for "can't happen".

**`dispatcherProvider` is now an unused dependency.** Its only use in `RatingsRepositoryImpl` was the `withContext(dispatcherProvider.io)` inside the deleted `getCourseRatingSummary` (line 41 on master), so the constructor parameter and `import kotlinx.coroutines.withContext` are both dead now. `import …repository.UserRepository` too, and the same import in `RatingsRepositoryImplTest`.

Losing that `withContext(io)` is fine, incidentally — `getRatingSummary` only calls `ratingDao.getByTypeAndItem`, and Room runs `suspend` DAO queries off the main thread itself. Worth a line in the description so it doesn't read as a dropped safety net.

**Nit:** `RatingSummaryProvider` is now a one-line pass-through to `ratingsRepository.getRatingSummary("course", courseId, userId)` with a single caller. It was earning its place when it wrapped a course-specific repository method; now it's an indirection with the `"course"` literal as its only content.

**Verdict:** mergeable.

---

PR #15942, opened 2026-08-20
resources: smoother title view modelling (fixes #15941)

### LINKED ISSUE #15941: library: fix order by title
https://github.com/user-attachments/assets/720df73a-7566-4267-a32f-13f8c1dabcde

when we first click order by title, the list is ordered from z-a, not a-z. i think it is better if we can do it from a-z

ragilzakaria:
https://github.com/user-attachments/assets/6a7943d6-60c7-441c-9a1b-c4f5b2df62bb

fixes #15941 

## Summary by CodeRabbit

* **Bug Fixes**
  * Resource lists now sort by title in descending order by default.
  * Improved sorting behavior for titles and dates when switching between ascending and descending order.

* **Tests**
  * Added coverage to verify title and date sorting in both directions.

* **Release**
  * Updated the app version to 0.66.33.


dogi:
**Review — rating: 91/100**

One-character production change, and it's the right one — but the justification is better than the description gives.

`toggleTitleSortOrder` flips the flag *before* sorting (`isTitleAscending = !isTitleAscending; applyCurrentSort(list)`), so an initial `true` meant the first tap on "sort by title" produced **Z→A**. Starting at `false` makes the first tap A→Z, which is what anyone expects from a fresh title sort.

**The stronger argument: this makes Resources match Courses.** `CoursesViewModel` already has `private var isTitleAscending = false` (line 44) with the identical toggle-then-sort shape, so the two list screens have been disagreeing about what the first title-sort click does. Framing it as "align Resources with Courses" is more compelling than "smoother sorting" and tells the next reader why `false` is correct rather than arbitrary.

I checked the flag isn't read anywhere else — it's private and only touched by the toggle and `applyCurrentSort`, so there's no arrow icon or header state that could now disagree with the list order.

The two new tests are the right ones and they document the intended asymmetry: title's first toggle is ascending (Apple, Banana, Cherry) while date's first toggle is descending (newest first). Both are the conventional default for their type, and now that's pinned rather than folklore. The test-file tidy-up — replacing the fully-qualified `org.ole.planet.myplanet.model.ResourceItem`/`ResourceListModel`/`TagItem` with imports — is a welcome bonus.

**Nit:** `emptyList<TagItem>()` in `createResourceModel` doesn't need the explicit type argument; `emptyList()` infers from the parameter.

**Verdict:** mergeable.

---

Okuro3499 (approved):

dogi (approved):

PR #15943, opened 2026-08-20
Inject DispatcherProvider into BellDashboardViewModel

dogi:
Inject DispatcherProvider into BellDashboardViewModel to get string building/list mappings off main.

---
*PR created automatically by Jules for task [7813160799545953368](https://jules.google.com/task/7813160799545953368) started by @dogi*

## Summary by CodeRabbit

* **Removed Features**
  * Removed challenge prompts and related progress dialogs from the dashboard.
  * Removed challenge eligibility checks, progress calculations, and challenge-server configuration.
  * Removed the associated challenge dialog layout and localized text resources.
* **Tests**
  * Removed tests covering challenge-related progress calculations.


dogi:
**Review — rating: 62/100**

The premise is sound — `handleDueReminders` is reached from `viewModelScope.launch { surveysRepository.dueRemindersFlow().collect { … } }`, and `viewModelScope` is `Dispatchers.Main.immediate`, so this CPU work really was on the main thread. Injecting `DispatcherProvider` rather than hard-coding `Dispatchers.Default` is also the project convention.

**But the granularity makes it slower, not faster.** There are now four separate `withContext(dispatcherProvider.default)` blocks, and two of them are *inside the `for` loop*:

```kotlin
for (surveyIds in remindersToShow) {
    val surveyIdList = withContext(dispatcherProvider.default) { surveyIds.split(",").filter { it.isNotBlank() } }
    …
    val pendingSurveys = withContext(dispatcherProvider.default) { surveyIdList.mapNotNull { submissionsById[it] }.filter { … } }
```

Each `withContext` is a dispatch plus a continuation allocation — comfortably more expensive than `split(",")` on a short comma-separated id list or a `mapNotNull` over a handful of elements. So for N reminders this adds 2N+2 thread hops to do work measured in microseconds, and the loop now bounces default→main→default per iteration because the suspend repository calls in between resume elsewhere.

One block around the whole body is both simpler and strictly better:

```kotlin
private suspend fun handleDueReminders(remindersToShow: List<String>) = withContext(dispatcherProvider.default) {
    …
}
```

The repository calls inside already switch to IO internally, so hoisting doesn't pin DB work to `default`; it just means the string handling and map building happen once, off the main thread, without per-item hops. If you'd rather be surgical, wrap only `allSubmissions.associateBy { it.id }` — the one piece that scales with data size — and leave the `split`s inline.

**Nit:** `import kotlinx.coroutines.withContext` is inserted between `repository.TeamsRepository` and `repository.UserRepository`, splitting the `org.ole.planet…` group. It belongs with the other `kotlinx` imports.

**Nit:** the test passes `org.ole.planet.myplanet.utils.TestDispatcherProvider(testDispatcher)` fully qualified inline, twice, where an import would do. Good that it uses the shared `TestDispatcherProvider` rather than hand-rolling one, though.

**Nit:** no test covers the change itself. Since the point is "this work is off the main thread", a test asserting the reminder flow still emits correctly under a `StandardTestDispatcher` is about the only thing that would catch a regression here — and the existing two tests were only updated for the new constructor arg.

**Verdict:** needs another round — one `withContext` around the function body instead of four inside it.

---

dogi:
**Re-review — Rating: 64/100** (was 62)

The granularity complaint is fixed — one `withContext(dispatcherProvider.default)` around the whole block instead of four, none inside a loop. But the diff now targets `DashboardViewModel` rather than the `handleDueReminders` code I reviewed last round, and against this new target the premise doesn't hold.

**The hop guards work that never runs.** The block's only substantive statements are behind `shouldPrompt`:

```kotlin
val shouldPrompt = today.isAfter(LocalDate.of(2024, 11, 30)) &&
        today.isBefore(endDate) &&          // endDate = LocalDate.of(2025, 1, 16)
        serverUrl in validUrls
```

`today` is `LocalDate.now()`, so `isBefore(2025-01-16)` has been false since January 2025 — `shouldPrompt` is permanently false and the branch returns `null` after three cheap date comparisons. `getCourseStatusString` and `hasUserCompletedSync` never execute.

**What does still run is one suspend call that already dispatches itself.** `progressRepository.findProgressForCourse(...)` sits above the `if`, and repository methods do their own `withContext`. So the net effect is an extra dispatch on the way in and out, protecting nothing. For reference, the CPU cost you'd be moving even if the window were open is `JsonUtils.getInt` twice plus one `getString` — microseconds.

The genuinely useful change in this area is the opposite one: that whole `ChallengeDialogData` branch is dead code past 2025-01-16 and could be deleted, along with `validUrls`, `getCourseStatusString` and the `hasUserCompletedSync` call, if nobody is planning to reopen the window.

**Nit:** the PR title still says `BellDashboardViewModel`; the diff only touches `DashboardViewModel`.

**Nit:** `import kotlinx.coroutines.withContext` is placed before the `kotlinx.coroutines.flow.*` group — it sorts after them.

**Verdict:** needs changes — either drop the hop, or repoint the PR at the dead-code removal.

---

dogi:
**Re-review — Rating: 78/100** (was 64)

You took the better of the two options I offered, and the result is far more valuable than the dispatcher hop it replaces. The December-2024 challenge window closed on 2025-01-16, so `ChallengePrompter`, `MarkdownDialogFragment`, `dialog_campaign_challenge.xml`, `evaluateChallengeDialog` and `getCourseStatusString` were all unreachable — deleting them beats optimising them.

**The sweep stops short in four places, all in files this PR already touches.**

1. **`DashboardActivity` now computes two locals it doesn't use.** The call you removed was the only consumer of both:

```kotlin
val validUrls = ServerConfigUtils.getChallengeServerUrls()
val isGuest = user?.id?.startsWith("guest") == true
reportFullyDrawn()
```

With those gone, `ServerConfigUtils.getChallengeServerUrls()` has no callers left either.

2. **`DashboardViewModel` keeps the event channel with nothing on either end.** `ChallengeDialogData` (line 58) and `_challengeDialogEvent` / `challengeDialogEvent` (lines 107–108) survive, but the only emitter was inside `evaluateChallengeDialog` and the only collector was the `collectWhenStarted` block you deleted from `DashboardActivity`.

3. **`calculateIndividualProgress` and `calculateCommunityProgress`** were called only from `ChallengePrompter`. They now exist solely to satisfy `DashboardViewModelTest` — worth deleting along with those test cases rather than leaving public methods whose only caller is their own test.

4. **The resources go too**: `drawable/december_challenge` plus `community_earnings`, `your_earnings`, `congratulations`, `per_survey`, `share_opinion`, `remember_sync` and `daily_voices` in `values/strings.xml` and all five `values-{lang}` files.

None of this is harmful — Kotlin only warns on unused locals and declarations, and `minifyEnabled=false` means unused resources just ship — which is why CI is green. But a PR whose whole purpose is removing dead code shouldn't leave a dead `MutableSharedFlow` and two dead locals behind.

**Worth checking while you're in there:** `CustomClickableSpan` is *not* dead — `CourseStepFragment:154` still uses it, so leave it.

**Verdict:** needs changes — finish the sweep; the direction is exactly right and what's left is mechanical.

---

dogi:
**Re-review — Rating: 76/100** (was 78)

All four items from last round are done: the orphaned `validUrls`/`isGuest` locals are gone from `DashboardActivity`, `getChallengeServerUrls()` is out of `ServerConfigUtils`, the seven challenge strings are removed from all six locale files, and `calculateIndividualProgress`/`calculateCommunityProgress` went with their tests. The feature deletion itself is right and complete — `ChallengePrompter`, `MarkdownDialogFragment`, `dialog_campaign_challenge.xml`, `december_challenge.webp` and `evaluateChallengeDialog` are all unreachable after 2025-01-16, so removing them beats optimising them.

What's left is the next ring out — code that *this diff* makes dead, and that nothing in the toolchain will flag (Kotlin only warns on unused declarations, and `minifyEnabled=false` ships unused resources). Complete list, so this can close in one pass:

1. **`DashboardViewModel:58`** — `data class ChallengeDialogData`. Its only producer and consumer are both deleted here.
2. **`DashboardViewModel:107-108`** — `_challengeDialogEvent` / `challengeDialogEvent`. The `emit` was inside `evaluateChallengeDialog` and the collector was the `collectWhenStarted` block you removed from `DashboardActivity`, so this is a `MutableSharedFlow` with nothing on either end.
3. **`SharedPrefManager:242-243`** — `getHasShownCongrats()` / `setHasShownCongrats()` plus the `HAS_SHOWN_CONGRATS` key. `ChallengePrompter` was the only caller.
4. **`course_completed`, `course_in_progress`, `course_not_started`** — `getCourseStatusString` was their only consumer (no layout or menu references either), so that's 18 dead entries across the six locale files.
5. **Repository methods that lose their last caller:** `VoicesRepository.getCommunityVoiceDates` and `SubmissionsRepository.hasPendingSurvey` were used only by `evaluateChallengeDialog`, interface and impl both. `ProgressRepository.hasUserCompletedSync` likewise — note the `ActivitiesRepository` one behind it may still be reachable, so check that one rather than deleting through. Worth re-checking whether `DashboardViewModel` still needs `voicesRepository`, `submissionsRepository`, `coursesRepository` and `application` in its constructor once those go.
6. **Docs:** `CLAUDE.md:97` still lists `MarkdownDialogFragment` under `ui/components/` (count 5), `CLAUDE.md:313` still lists `ChallengePrompter`, and `docs/CODE_STYLE_GUIDE.md:768` says the hardcoded course ID is in 4 places including two files this PR deletes — it'll be 2 after this lands.

To be clear: this isn't a repeat of last round's list, it's the ring that only became visible once last round's items landed. `CustomClickableSpan` should still stay — `CourseStepFragment:154` uses it.

**Verdict:** the deletion is right; finish the sweep it opened up, then this is straightforward.

---

dogi:

The diff is byte-for-byte what I reviewed last round — CI re-ran, but no new commits touched the tree. All six leftovers from my [previous comment](https://github.com/open-learning-exchange/myplanet/pull/15943#issuecomment-5393274300) are still open: `ChallengeDialogData`, the `_challengeDialogEvent` flow with nothing on either end, `getHasShownCongrats`/`setHasShownCongrats`, the three `course_*` strings across six locale files, the repository methods that lose their last caller, and the three stale doc references.

Nothing to add — the feature deletion is still right, and the list is still the shortest path to finishing it.

**Verdict:** unchanged; the sweep this PR opened up still needs closing.

---

PR #15944, opened 2026-08-20
all: smoother fragment manager back stack listening (fixes #16056)

### LINKED ISSUE #16056: Fix: Make unpaired FragmentManager back-stack listeners disposable
Tracking issue for PR #15944.

Fix: Make unpaired FragmentManager back-stack listeners disposable

---
*PR created automatically by Jules for task [9938843279581325173](https://jules.google.com/task/9938843279581325173) started by @dogi*

dogi:
Fix: Make unpaired FragmentManager back-stack listeners disposable

---
*PR created automatically by Jules for task [9938843279581325173](https://jules.google.com/task/9938843279581325173) started by @dogi*


dogi:
**Review — rating: 82/100**

Correct and consistent, though the framing is stronger than the effect.

Both listeners are now stored in a field and removed in `onDestroy`, which matches what `DashboardElementActivity` already does for its own listener (`addOnBackStackChangedListener(this)` at line 42, `removeOnBackStackChangedListener(this)` at line 197). Those two were the only unpaired registrations left in the app, so after this every `addOnBackStackChangedListener` call site has a matching removal. That consistency is worth having on its own — the next person adding one will copy a complete pattern.

**On the leak claim:** a `FragmentManager` belongs to its host, so `supportFragmentManager`'s listener list dies with the Activity it's attached to. The lambda captures `this`, `result` and `binding`, but nothing outlives the Activity holding them, so this isn't a leak a heap dump would flag. The real value is the pairing discipline, not reclaimed memory — worth saying that way in the description so nobody goes looking for the leak later.

`PublicSurveyActivity` is the cleaner of the two: added in `onCreate`, removed in `onDestroy`, one call each way, and it sits right next to the existing `unregisterFragmentLifecycleCallbacks` so the symmetry is visible.

**Nit — the field pattern assumes single registration.** `backStackListener = FragmentManager.OnBackStackChangedListener { … }` overwrites the field, so if `setupNavigation()` were ever called twice, the first listener would stay registered with nothing referencing it and `onDestroy` would remove only the second. Not reachable today (`setupNavigation()` has exactly one call site, line 181), but the field makes it look safe to call again when it isn't. A `remove` before the `add`, or an `if (backStackListener == null)` guard, makes it actually idempotent.

**Nit:** `backStackListener?.let { supportFragmentManager.addOnBackStackChangedListener(it) }` immediately after assigning it non-null — the `?.let` is dead ceremony. Assigning to a local first (`val listener = FragmentManager.OnBackStackChangedListener { … }; backStackListener = listener; supportFragmentManager.addOnBackStackChangedListener(listener)`) reads plainly and drops the null dance at both ends.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #15948, opened 2026-08-20
teams: smoother events repository detail view modelling (fixes #16058)

### LINKED ISSUE #16058: Remove UserRepository from EventsRepositoryImpl
Tracking issue for PR #15948.

Refactor EventsRepositoryImpl to remove UserRepository dependency

Removed the `UserRepository` injection from `EventsRepositoryImpl`. Instead of `EventsRepositoryImpl` independently resolving the user context, it now requires the `EventsDetailViewModel` to pass the `allUsers` list and the `userId` directly into `getJoinedMembers` and `toggleAttendance`. Unit tests have been updated accordingly.

---
*PR created automatically by Jules for task [10018330304807200551](https://jules.google.com/task/10018330304807200551) started by @dogi*

## Summary by CodeRabbit

* **New Features**
  * Attendance can now be toggled for a specific user.
  * Users without a valid account ID are prevented from being added to event attendance.
* **Bug Fixes**
  * Improved event member lookup for more accurate results.
  * Prevented attendance updates when no active user is available.
* **Tests**
  * Expanded attendance coverage for joining, leaving, empty event IDs, and missing user IDs.

dogi:
Refactor EventsRepositoryImpl to remove UserRepository dependency

Removed the `UserRepository` injection from `EventsRepositoryImpl`. Instead of `EventsRepositoryImpl` independently resolving the user context, it now requires the `EventsDetailViewModel` to pass the `allUsers` list and the `userId` directly into `getJoinedMembers` and `toggleAttendance`. Unit tests have been updated accordingly.

---
*PR created automatically by Jules for task [10018330304807200551](https://jules.google.com/task/10018330304807200551) started by @dogi*

## Summary by CodeRabbit

* **New Features**
  * Attendance can now be toggled for a specific user.
  * Users without a valid account ID are prevented from being added to event attendance.
* **Bug Fixes**
  * Improved event member lookup for more accurate results.
  * Prevented attendance updates when no active user is available.
* **Tests**
  * Expanded attendance coverage for joining, leaving, empty event IDs, and missing user IDs.


dogi:
**Review — rating: 72/100**

Reviewed alongside #15940, which removes `UserRepository` from `RatingsRepositoryImpl`. Same goal, and that one is worth looking at because it takes the better route: the caller there already held a `UserEntity`, so passing the object deleted a lookup. Here the caller doesn't hold anything, so the dependency is removed by passing **the entire users table** through the API:

```kotlin
suspend fun getJoinedMembers(meetupId: String, allUsers: List<UserEntity>): List<UserEntity>
```

`EventsDetailViewModel` now calls `userRepository.getAllUsers()` and hands the result down for the repository to filter with `memberIdSet.contains(user.id) || user._id in memberIdSet`. The dependency edge is gone on paper, but the whole-table read didn't go away — it moved up a layer and became part of the contract, and every future caller has to know to fetch it.

**`CLAUDE.md` sanctions the simpler fix:** repositories inject the DAOs they need. `EventsRepositoryImpl` can take `UserDao` directly, keep `getJoinedMembers(meetupId)` as it was, and replace the whole-table scan with a query:

```kotlin
@Query("SELECT * FROM users WHERE id IN (:ids) OR _id IN (:ids)")
suspend fun getByIdsOrRemoteIds(ids: List<String>): List<UserEntity>
```

That removes the repository→repository edge *and* the full scan, instead of trading one for a leakier signature. (If you go that way, chunk the `IN` list — see #15919 in this queue for the 999-parameter limit.)

**Two smaller consequences of the current shape:**

**The load path got slower.** `val allUsers = userRepository.getAllUsers()` now runs *before* the two `async` blocks, so the users read is serialized ahead of the meetup read where it used to happen inside `membersDeferred` and overlap with it.

**`toggleAttendance` reads the table again.** The ViewModel calls `getAllUsers()` a second time right after the toggle, so joining or leaving a meetup is now two full user-table reads per tap.

**A behavior change worth confirming:** the ViewModel now early-returns on a blank current-user id, where the old repository path returned `getMeetupById(meetupId)`. So `_meetup.value` is no longer refreshed in that case — probably fine, but it also means the repository's own `if (userId.isBlank()) return getMeetupById(meetupId)` branch is now unreachable from this call site while still being tested (`toggleAttendance("meetup1", "")`). Two guards for one condition; pick one.

**Nit:** `import org.ole.planet.myplanet.repository.UserRepository` is now unused in both `EventsRepositoryImpl` (its only reference was the removed constructor parameter) and `EventsRepositoryImplTest`.

**Nit:** `.map { it }` at the end of `getJoinedMembers` is a no-op that predates this PR but sits in a block you're already rewriting.

**Credit where due:** the test genuinely got better — `EventsRepositoryImplTest` no longer needs a `UserRepository` mock and passes a plain list, which is the one real upside of the parameter approach.

**Verdict:** needs another round — inject `UserDao` and query by id instead of passing the user table through the signature.

---

dogi:
**Re-review — Rating: 58/100** (was 72)

The main design ask landed: the whole-table read is gone. `getJoinedMembers` now queries only the ids it needs via `userDao.getUsersByAnyIds`, and `toggleAttendance(meetupId, userId)` takes the id as a parameter instead of resolving it. That's the right shape, and it's better than the `allUsers: List<UserEntity>` contract from last round. But the chunk size is wrong in a way that crashes on older devices.

**`chunked(900)` overruns SQLite's bind-parameter limit.** The DAO query names the list twice:

```kotlin
@Query("SELECT * FROM users WHERE id IN (:userIds) OR _id IN (:userIds)")
suspend fun getUsersByAnyIds(userIds: List<String>): List<UserEntity>
```

Room expands each `IN (:userIds)` occurrence independently, so a chunk of N ids binds **2N** parameters. `SQLITE_MAX_VARIABLE_NUMBER` is 999 on API 26–30 (raised to 32766 only in API 31+), and `minSdk` is 26. A meetup with 500+ joined members therefore throws `SQLiteException: too many SQL variables` on Android 8–11 — the exact failure mode the chunking was added to prevent.

`UserRepositoryImpl.getUsersByIds` already solves this and is worth copying verbatim, including the `400`:

```kotlin
val userIdList = userIds.distinct()
userIdList.chunked(400).forEach { chunk -> result.addAll(userDao.getUsersByAnyIds(chunk)) }
return result.distinctBy { it.id }
```

Note it also `distinct()`s the input first, which your version skips — duplicate member ids inflate the bind count for no benefit.

**A thinking-out-loud comment got committed.** In `EventsRepositoryImplTest`:

```kotlin
assertEquals(3, result.size) // The distinctBy will filter out duplicates but mock returns 3 unique entities. Wait, we need to assert the chunking behavior. Let's just assert 3 returned items.
```

**And that rewrite dropped the only assertion that mattered.** The old test proved `getJoinedMembers` returns exactly the two users whose ids appear in the meetup list and excludes `user3`. The new one mocks `getUsersByAnyIds(any())` to return all three and then asserts three came back — it can no longer fail if the filtering breaks, because the filtering now lives in SQL that the mock replaces. The `id`-or-`_id` matching rule is the interesting behaviour here and it's currently untested; an in-memory Room test (`Room.inMemoryDatabaseBuilder`, per `docs/TESTING.md`) would actually exercise it.

**New race in `toggleAttendance`.** `EventsDetailViewModel` now reads `_user.value?.id ?: ""`, but `_user` is only populated inside `loadData`'s `viewModelScope.launch`. A tap that lands before that coroutine completes passes `""`, and the repository's `if (userId.isBlank()) return getMeetupById(meetupId)` returns the meetup untouched — so the join silently does nothing and the UI shows no change. The old code called `getUserModel()` at toggle time and always had a user. Gating the button on `user != null`, or awaiting the deferred as #15920 does, closes it.

**Verdict:** needs changes — the 900-element chunk is a crash on API 26–30; also drop the stray comment and restore a test that proves the id/`_id` filtering.

---

dogi:
**Re-review — Rating: 84/100** (was 58)

The crash is fixed. `chunked(400)` keeps `getUsersByAnyIds` at 800 bind parameters per chunk — under SQLite's 999 limit on API 26–30 even though the query names `:userIds` twice — and it now matches `UserRepositoryImpl.getUsersByIds`, so the two places that hit this DAO agree on the constant. The stray thinking-out-loud comment is gone, and the test mock no longer returns a third user it then asserted on, so `assertEquals(2, result.size)` describes real behaviour again.

The underlying change is the right one: querying by id instead of pulling the whole users table, with `distinctBy { it.id }` covering the case where a member matches on both `id` and `_id` across chunks.

**The `_user` race changed shape rather than going away.** `val currentUserId = _user.value?.id ?: return@launch` is cleaner than passing `""` down to be rejected — it skips a pointless DAO round trip — but the user-visible outcome is the same: a tap landing before `loadData`'s coroutine populates `_user` does nothing, with no feedback. Gating the join/leave button on `user != null`, or awaiting the user the way #15920 does with `viewModelScope.async { … }`, is what actually closes it. Narrow window, so not a blocker.

**Nit:** `getUsersByIds` also `distinct()`s its input before chunking. Duplicate member ids in `memberIds` inflate the bind count for nothing, and it's one call to match the sibling implementation exactly.

**Nit:** the filtering contract — a member id matching either `id` or `_id` — now lives entirely in SQL and the test mocks that SQL away, so nothing exercises it. An in-memory Room test (`Room.inMemoryDatabaseBuilder`, per `docs/TESTING.md`) would cover the one interesting behaviour here.

**Verdict:** ready to merge.

---

Okuro3499 (approved):

PR #15954, opened 2026-08-20
teams: smoother surveys view modelling (fixes #16062)

### LINKED ISSUE #16062: perf(SurveysViewModel): optimize case-insensitive sorting
Tracking issue for PR #15954.

Avoid O(n log n) throwaway string allocations when sorting Surveys by Title.

---
*PR created automatically by Jules for task [13165888133867346472](https://jules.google.com/task/13165888133867346472) started by @dogi*

dogi:
Avoid O(n log n) throwaway string allocations when sorting Surveys by Title.

---
*PR created automatically by Jules for task [13165888133867346472](https://jules.google.com/task/13165888133867346472) started by @dogi*


dogi:
**Review — rating: 88/100**

This fixes a real allocation problem, not a hypothetical one. `sortedBy { it.name?.lowercase(Locale.getDefault()) }` expands to `sortedWith(compareBy(selector))`, and `compareBy` runs its selector on **every comparison** — so sorting N surveys allocated on the order of `N log N` lowercased strings, not N. `compareBy(String.CASE_INSENSITIVE_ORDER) { it.name ?: "" }` folds case per character inside the comparator and allocates nothing.

Ordering is preserved where it matters. `sortedBy` with a nullable selector treats `null` as smallest via `compareValues`, and `it.name ?: ""` maps null to the empty string, which is also smallest under `CASE_INSENSITIVE_ORDER` — so null-named surveys still lead in `TITLE_ASC` and trail in `TITLE_DESC`. The only difference is that `null` and `""` now compare equal instead of `null` sorting strictly first, and since both sorts are stable that isn't observable.

It also quietly removes a locale dependency: `lowercase(Locale.getDefault())` is subject to the Turkish dotted/dotless-I mapping, so `"I"` folded differently under `tr` than elsewhere and could reorder rows by device locale. `CASE_INSENSITIVE_ORDER` is locale-independent. Dropping the now-unused `java.util.Locale` import is correct — `getSortDate` doesn't use it.

**Nit:** neither version is real collation. Both compare UTF-16 code units after case folding, so in Spanish and French accented characters still sort after `z` (`ñ`, `é` land past the alphabet). If the goal is correct alphabetical order in the shipped locales, `Collator.getInstance()` as the comparator is the change to make — and it would still avoid the per-comparison allocation.

**Verdict:** ready to merge.

---

Okuro3499 (approved):

PR #15956, opened 2026-08-20
all: smoother utilities toast handling (fixes #16064)

### LINKED ISSUE #16064: perf(Utilities): stop allocating a Handler per background toast
Tracking issue for PR #15956.

This change optimizes `Utilities.toast` to avoid allocating a new `Handler` on every background toast.

---
*PR created automatically by Jules for task [17487410153034421175](https://jules.google.com/task/17487410153034421175) started by @dogi*

dogi:
This change optimizes `Utilities.toast` to avoid allocating a new `Handler` on every background toast.

---
*PR created automatically by Jules for task [17487410153034421175](https://jules.google.com/task/17487410153034421175) started by @dogi*


dogi:
**Review — rating: 85/100**

Safe and correct. The cached `Handler` holds `Looper.getMainLooper()`, which is a process-lifetime object rather than a `Context`, so there's nothing here to leak — this is the standard idiom for a static main-thread poster. `Handler.post` is thread-safe, which matters because this branch is by definition reached from non-main threads, and the main-thread branch above it still bypasses the handler entirely.

The saving is one `Handler` per background toast, and `Utilities.toast` is called from the sync and upload paths where several can fire in sequence, so it isn't nothing — but a `Handler` is a small object and this won't show up in a trace. Worth taking as tidiness more than performance.

**Nit:** `by lazy` defaults to `SYNCHRONIZED`, so the first call pays a lock. For a field in an `object` initialised from arbitrary threads that's the right default, but a plain `private val mainHandler = Handler(Looper.getMainLooper())` would work too — `Utilities` is an `object`, so class init already gives you the once-only guarantee, and `Looper.getMainLooper()` is valid from any thread at any time.

**Verdict:** ready to merge.

---

Okuro3499 (approved):

PR #15953, opened 2026-08-20
all: smoother colors context caching (fixes #16061)

dogi:
Resolves an issue where `ContextCompat.getColor` was repeatedly called in `UserArrayAdapter` and `ServerAddressAdapter` during `onBindViewHolder` and `PAYLOAD_SELECTION`.

- Modifies the constructors of `UserArrayAdapter` and `ServerAddressAdapter` to accept a `Context` argument.
- Caches the frequently used `selectedColor` and `transparentColor` values as class-level properties.
- Updates caller sites in `TeamsTasksFragment` and `ServerDialogExtensions` to supply the necessary context.
- Skips redundant allocations on every scroll and selection.

---
*PR created automatically by Jules for task [9349214527347736200](https://jules.google.com/task/9349214527347736200) started by @dogi*

PR #15957, opened 2026-08-20
teams: smoother list caching (fixes #16065)

### LINKED ISSUE #16065: perf: Cache action strings and color in TeamsAdapter
Tracking issue for PR #15957.

Caches string and color resource lookups in `TeamsAdapter` to improve scroll performance.

---
*PR created automatically by Jules for task [11826598221091744750](https://jules.google.com/task/11826598221091744750) started by @dogi*

dogi:
Caches string and color resource lookups in `TeamsAdapter` to improve scroll performance.

---
*PR created automatically by Jules for task [11826598221091744750](https://jules.google.com/task/11826598221091744750) started by @dogi*


dogi:
**Review — rating: 82/100**

The saving is genuine: `bindJoinLeave` ran up to one `getString` plus, on the pending-request branch, a `ContextCompat.getColor` on every bind, and a teams list scrolls. Hoisting all five into fields removes that from the bind path, and all four `context.getString` uses were converted — nothing is left reading the now-deleted `val context = root.context`.

Correct in practice, too: `onCreateViewHolder` necessarily runs before any `onBindViewHolder` for a given adapter, so the fields are always populated by the time `bindJoinLeave` reads them.

**The nullable-sentinel shape is the weak part.** Five fields typed `String?`/`Int` plus an `if (actionEditString == null)` guard encode "initialised" in the data itself, which means the compiler can't tell you if the invariant ever breaks — it just interpolates, and `contentDescription = "$actionEditString ${team.name}"` would read *"null Team Name"* to a screen reader, while `pendingColor = 0` is a fully transparent tint. Neither is reachable today, but neither fails loudly if it becomes reachable.

#15953, in this same round, uses the shape that avoids all of it — take the `Context` in the constructor and declare the values as non-null `by lazy(LazyThreadSafetyMode.NONE)`:

```kotlin
class TeamsAdapter(private val context: Context, …) {
    private val actionEdit by lazy(LazyThreadSafetyMode.NONE) { context.getString(R.string.edit) }
    private val pendingColor by lazy(LazyThreadSafetyMode.NONE) {
        ContextCompat.getColor(context, R.color.pending_request_indicator)
    }
```

Same caching, non-null types, no sentinel, no dependency on callback ordering. Worth converging the two PRs on one pattern since they'll be read together.

**Nit:** like the other adapter-caching PRs in flight, the cached strings go stale if this adapter instance outlives a locale change. Activity recreation normally rebuilds it, so this is consistent with #15934/#15916/#15953 rather than a new problem.

**Verdict:** ready to merge; converging on #15953's constructor-`Context` shape would be better.

---

Okuro3499 (approved):

PR #15958, opened 2026-08-20
community: smoother tab view modelling (fixes #16066)

### LINKED ISSUE #16066: Refactor CommunityTabFragment to use MVVM
Tracking issue for PR #15958.

Extracts data fetching logic to a ViewModel and exposes state via StateFlow.

---
*PR created automatically by Jules for task [11076550096000981439](https://jules.google.com/task/11076550096000981439) started by @dogi*

dogi:
Extracts data fetching logic to a ViewModel and exposes state via StateFlow.

---
*PR created automatically by Jules for task [11076550096000981439](https://jules.google.com/task/11076550096000981439) started by @dogi*

## Summary by CodeRabbit

* **New Features**
  * Improved community tab setup by loading community, parent, planet, and user information together.
  * Community headers now display the appropriate community or planet name and planet type.

* **Bug Fixes**
  * Improved reliability of community tab data loading and configuration.

* **Chores**
  * Updated the app version from 0.66.42 to 0.66.43.


dogi:
**Review — rating: 68/100**

The extraction itself is clean — `CommunityTabState` bundles the four values coherently, the three injected dependencies move out of the Fragment where they belong, and the Fragment is left with just view wiring. But the collection operator changed the execution contract in a way the old code didn't have.

**The pager is rebuilt on every foreground.** `collectWhenStarted` is `repeatOnLifecycle(Lifecycle.State.STARTED) { flow.collect(collector) }`, so the collector is cancelled on STOP and restarted on START — and restarting a `StateFlow` collection immediately replays the current value. The old code was `viewLifecycleOwner.lifecycleScope.launch { … }` in `onViewCreated`, which ran exactly once per view. So every time the user backgrounds and returns while this fragment's view is alive, the block runs again and:

```kotlin
binding.viewPager2.adapter = CommunityPagerAdapter(requireActivity(), …)   // pager reset to page 0
TabLayoutMediator(binding.tabLayout, binding.viewPager2) { … }.attach()    // second mediator attached
```

Assigning a new adapter destroys and recreates the pager's fragments, so the user loses their tab position and any scroll state on resume. And each new `TabLayoutMediator` registers its own page-change callback and adapter observer on the same `TabLayout` — the previous one is never `detach()`ed, so the callbacks accumulate for as long as the view lives.

Easiest fix is to make the one-shot nature explicit, since `state` is set once and never updated:

```kotlin
    val state = viewModel.state.filterNotNull().first()
    …
}
```

That also lets you drop the `if (state == null) return@collectWhenStarted` guard and the nullable `CommunityTabState?`. If you'd rather keep `collectWhenStarted`, guard the setup on `binding.viewPager2.adapter == null` and hold the mediator in a field so it can be detached.

**Nit:** `import javax.inject.Inject` sits at the bottom of the new ViewModel's import list, after the `org.ole.planet.myplanet.*` group; it belongs between `dagger.*` and `kotlinx.*`.

**Verdict:** needs changes — the resume path rebuilds the pager and leaks mediators.

---

dogi:

Fixed exactly at the root. `viewModel.state.filterNotNull().first()` inside `viewLifecycleOwner.lifecycleScope.launch` suspends once, resumes when the state lands, and the coroutine completes — so the pager adapter is assigned once per view and exactly one `TabLayoutMediator` is ever attached. The `repeatOnLifecycle(STARTED)` re-entry that rebuilt both on every foreground is gone.

It also cleaned up on the way: `filterNotNull()` makes the nullable `CommunityTabState?` an implementation detail instead of something the collector has to guard, so the `if (state == null) return@collectWhenStarted` early-return disappeared with it.

The extraction itself is unchanged and still good — three dependencies out of the Fragment, one coherent state object, and the Fragment left doing only view wiring.

**Nit:** `import javax.inject.Inject` moved but is still out of order — it now sits between `kotlinx.coroutines.launch` and the `org.ole.planet.myplanet.*` group, and belongs above `kotlinx.*` (`dagger` < `javax` < `kotlinx` < `org`).

**Verdict:** ready to merge.

---

Okuro3499 (approved):

PR #15960, opened 2026-08-20
sync: smoother realtime tables watching (fixes #16068)

### LINKED ISSUE #16068: perf: Snapshot watched tables as Set in RealtimeSyncHelper
Tracking issue for PR #15960.

Optimizes `RealtimeSyncHelper` by caching the watched tables list into a `Set` snapshot once per subscription setup, rather than executing the configuration lookup and scanning an O(N) list linearly for every emitted `TableDataUpdate`.

---
*PR created automatically by Jules for task [15254959190788201797](https://jules.google.com/task/15254959190788201797) started by @dogi*

dogi:
Optimizes `RealtimeSyncHelper` by caching the watched tables list into a `Set` snapshot once per subscription setup, rather than executing the configuration lookup and scanning an O(N) list linearly for every emitted `TableDataUpdate`.

---
*PR created automatically by Jules for task [15254959190788201797](https://jules.google.com/task/15254959190788201797) started by @dogi*


dogi:
**Review — rating: 87/100**

Correct, and I checked the thing that would have made it unsafe: every `RealtimeSyncMixin` implementation returns a fresh constant — `listOf("feedback")`, `listOf("courses")`, `listOf("resources")`, `listOf("exams")` — so nothing computes its watched tables dynamically and hoisting the snapshot out of the filter can't miss a later change.

Worth being precise about where the saving actually comes from, though. The `List`→`Set` half does almost nothing: every implementation returns exactly one element, so `contains` was already a single comparison. The real win is that `mixin.getWatchedTables()` was being **called** on every emission, and each of those calls allocates a new `listOf(...)`. Now it happens once per `setupRealtimeSync()`. That matters because `dataUpdateFlow` fans every table's update out to every subscribed fragment, so each fragment was allocating a list per event for events it then discarded.

**Nit:** this quietly makes `getWatchedTables()` a contract that must be constant per setup call. Nothing enforces it, and a future mixin that returns a user-dependent table list would silently watch a stale set. A line on the interface saying the result is snapshotted at setup would keep the next implementer out of that trap.

**Verdict:** ready to merge.

---

Okuro3499 (approved):

Okuro3499 (approved):

PR #15961, opened 2026-08-20
teams: smoother voices label managing (fixes #16069)

### LINKED ISSUE #16069: Optimize `VoicesLabelManager.getLabel` via reverse-indexed map and single ChipCloud instance
Tracking issue for PR #15961.

Optimize the `VoicesLabelManager.getLabel` function and `ChipCloud` rendering by replacing O(N) lookups with an O(1) reverse-indexed map lookups inside a single `showChips` function call and reusing one `ChipCloud` instance.

---
*PR created automatically by Jules for task [6514613379406387600](https://jules.google.com/task/6514613379406387600) started by @dogi*

dogi:
Optimize the `VoicesLabelManager.getLabel` function and `ChipCloud` rendering by replacing O(N) lookups with an O(1) reverse-indexed map lookups inside a single `showChips` function call and reusing one `ChipCloud` instance.

---
*PR created automatically by Jules for task [6514613379406387600](https://jules.google.com/task/6514613379406387600) started by @dogi*


dogi:
**Review — rating: 66/100**

Two changes are bundled here and they pull in opposite directions.

**The single-ChipCloud change is a genuine fix — keep it.** The old loop constructed a new `ChipCloud(context, binding.fbChips, chipConfig)` *per label*, each one attached to the same container and each registering its own `setDeleteListener`. Building one and adding all chips to it is both cheaper and more correct.

**The reverse-index change is a pessimization.** `Constants.LABELS` has three entries:

```kotlin
LABELS = mutableMapOf(
    "Offer" to "offer",
    "Help wanted" to "help",
    "Request for advice" to "advice"
)
```

The old `getLabel` walked those three keys with no allocation. The new code runs

```kotlin
val reverseLabels = Constants.LABELS.entries.associate { it.value to it.key }
```

on **every `showChips` call** — that is, on every voice row bind — allocating a `LinkedHashMap` plus its entry nodes to make O(1) lookups on a three-element map. Constructing the map costs more than the scan it replaces, and it now happens even for rows with no labels at all, since it sits above the `if (labels.isNotEmpty())` guard.

Hoisting it fixes both problems at once. `LABELS` is populated in `Constants`' `init` and never written afterwards, so a lazy field is safe:

```kotlin
companion object {
    private val reverseLabels by lazy { Constants.LABELS.entries.associate { it.value to it.key } }
    …
}
```

Then `showChips` is a pure lookup, and moving the `val reverseLabels` line below the `isNotEmpty()` guard stops empty rows paying anything.

**Nit:** with the map hoisted, `reverseLabels[it] ?: formatLabelValue(it)` appears twice — once for the display label, once inside the delete listener's `firstOrNull`. That's the old `getLabel` body; keeping a private `getLabel` that reads the hoisted map would restore the single definition without reintroducing the scan.

**Verdict:** needs changes — hoist the reverse map out of the per-bind path; the ChipCloud consolidation is good as-is.

---

dogi:
**Re-review — Rating: 89/100** (was 66)

The reverse map is out of the bind path. `private val reverseLabels by lazy { … }` in the companion object means it's built once per process instead of once per `showChips` call, which was the whole problem — a three-entry `LinkedHashMap` allocated for every voice row, including rows with no labels at all.

Hoisting it is safe here, which is worth stating since `Constants.LABELS` is a `var`: it's populated in `Constants`' `init` block and never written afterwards, and `by lazy` doesn't evaluate until the first `getLabel` call, long after class initialisation.

Restoring `getLabel` as a private method reading the hoisted map is better than the inlined `reverseLabels[it] ?: formatLabelValue(it)` from the previous round — the fallback rule now has one definition again, and both the chip label and the delete listener's `firstOrNull` go through it.

The single-`ChipCloud` change, which was the good half all along, is untouched: one instance for the whole row instead of one per label, with a single delete listener.

**Nit:** `val labels = voice.labels ?: emptyList()` is hoisted above the `isNotEmpty()` guard, so a label-less row still allocates `emptyList()` — which is a shared singleton in Kotlin, so this costs nothing. Mentioning it only because it's the shape that caused the original problem; here it's fine.

**Verdict:** ready to merge.

---

Okuro3499 (approved):

PR #15962, opened 2026-08-20
teams: smoother members repository view modelling (fixes #16219)

### LINKED ISSUE #16219: Refactor: decouple UI from TeamsSyncRepository
Tracking issue for PR #15962.

Decoupled UI from `TeamsSyncRepository`'s narrow interface by adding a semantic `recordTeamActivity()` pass-through to `TeamsRepository`, delegating the call internally, and updating the three specified fragment call sites in `TeamDetailFragment` and `TeamResourcesFragment`. Field injections in base fragments and ViewModels remain untouched per requirements.

---
*PR created automatically by Jules for task [12510097022340534144](https://jules.google.com/task/12510097022340534144) started by @dogi*

## Summary by CodeRabbit

* **Enhancements**
  * Team join requests now record activity more consistently.
  * Adding or removing team resources now records the related activity through the team activity workflow.

dogi:
Decoupled UI from `TeamsSyncRepository`'s narrow interface by adding a semantic `recordTeamActivity()` pass-through to `TeamsRepository`, delegating the call internally, and updating the three specified fragment call sites in `TeamDetailFragment` and `TeamResourcesFragment`. Field injections in base fragments and ViewModels remain untouched per requirements.

---
*PR created automatically by Jules for task [12510097022340534144](https://jules.google.com/task/12510097022340534144) started by @dogi*

## Summary by CodeRabbit

* **Enhancements**
  * Team join requests now record activity more consistently.
  * Adding or removing team resources now records the related activity through the team activity workflow.


dogi:
**Review — rating: 74/100**

The direction is right: `syncTeamActivities()` is a sync-layer verb, and having fragments call it through `TeamsSyncRepository` meant the UI named an interface that exists for the sync managers. Exposing `recordTeamActivity()` on `TeamsRepository` gives the UI an intent-shaped name against the interface it already depends on, and all three call sites are converted correctly.

**But the injection it was meant to remove is still there.** `teamsSyncRepository` is declared in `BaseTeamFragment`:

```kotlin
// BaseTeamFragment.kt:32
lateinit var teamsSyncRepository: TeamsSyncRepository
```

and those three call sites — `TeamDetailFragment:279`, `TeamResourcesFragment:123` and `:191` — are its **only** consumers in the codebase. After this PR the field has zero users, but it and its import stay, so every team fragment still has `TeamsSyncRepository` injected into it. The decoupling is one line short of real; deleting the field and the import is what makes the title true.

**Scope note:** `TeamViewModel:130,141` and `RequestsViewModel:61` also call `teamsSyncRepository.syncTeamActivities()`. Leaving them is a defensible split if "UI" means fragments here, but they're the same layer by most readings, and they'd need the same treatment before `TeamsSyncRepository` is genuinely out of `ui/`. Worth saying which you intend in the PR body so the follow-up is obvious.

**Nit:** `recordTeamActivity()` delegating to `syncTeamActivities()` inside the same class leaves two public entry points to one behaviour on `TeamsRepositoryImpl`. That's fine as an interface-segregation seam — different name per interface, one implementation — but it only reads that way if the UI-facing side is the *only* way fragments can reach it, which loops back to deleting the base-class field.

**Verdict:** needs changes — remove the now-unused `teamsSyncRepository` field from `BaseTeamFragment`.

---

dogi:

The last mile is done — `BaseTeamFragment` drops both the `@Inject lateinit var teamsSyncRepository` field and the `TeamsSyncRepository` import, so no team fragment carries that dependency any more. That's what makes the title true; previously every fragment still had it injected with nothing reading it.

The rest was already right and is unchanged: `recordTeamActivity()` on `TeamsRepository` gives the UI an intent-shaped verb against the interface it already depends on, and `TeamsRepositoryImpl` implementing both interfaces means the sync layer keeps `syncTeamActivities()` without the UI ever naming it.

**Nit:** `TeamViewModel:130,141` and `RequestsViewModel:61` still call `teamsSyncRepository.syncTeamActivities()` directly. Fragments are clean now, so `ui/` isn't fully free of the sync interface — worth a line in the PR body saying whether that's deliberate scope or a follow-up, so the next person doesn't have to re-derive it.

**Verdict:** ready to merge.

---

Okuro3499 (dismissed):
```
app/src/main/java/org/ole/planet/myplanet/repository/TeamsRepositoryImpl.kt:965 
    (recordTeamActivity()) — no unit test covers the new method;                    
    TeamsRepositoryImplTest only tests syncTeamActivities() directly. Add a test    
    asserting recordTeamActivity() delegates to syncTeamActivities().               
  - Either finish the migration for TeamViewModel.kt and RequestsViewModel.kt in    
    this PR, or split the work: keep this PR to the TeamsRepository interface       
    addition + a tracked follow-up issue for the two ViewModels, and update the     
    commit/PR description so it doesn't claim the sync interface was "removed from  
    the UI layout completely" when two UI-layer classes still depend on it directly
```

dogi:
**Re-review — Rating: 90/100** (was 88)

Unchanged from the round I passed, plus a test that's worth having: `recordTeamActivity delegates to syncTeamActivities` asserts the three uploads (`uploadResource(null)`, `uploadTeams()`, `uploadTeamActivities()`) actually fire through the new entry point, so the delegation can't be quietly broken later. That's a real behavioural assertion rather than a "calls the other method" mock check.

**Nit:** `unmockkObject(MainApplication.Companion)` is the last statement of the test body rather than in `@After` or a `finally`, so a failed assertion leaves the object mock installed for whatever runs next in that class — and since the tests are sharded by class, that's a confusing failure to chase. The neighbouring `syncTeamActivities uses dispatcherProvider io` test has the same shape, so this is matching local style rather than introducing the problem; hoisting both into `@After` would be a small tidy. The fully-qualified `io.mockk.mockkObject` / `org.ole.planet.myplanet.MainApplication.Companion` inline is likewise consistent with its neighbour but reads better as imports.

Standing note from last round, unchanged: `TeamViewModel:130,141` and `RequestsViewModel:61` still call `teamsSyncRepository.syncTeamActivities()` directly, so `ui/` isn't fully free of the sync interface even though every fragment now is. Still worth a line in the PR body saying whether that's deliberate scope.

**Verdict:** ready to merge.

---

Okuro3499 (dismissed):
```
NOT FIXED. TeamViewModel.kt:33,130,141 and RequestsViewModel.kt:27,61 still inject TeamsSyncRepository directly
  and call syncTeamActivities() — unchanged from before. The new commit's message still asserts "This removes the sync-manager interface from the UI layout completely" while its    
  own next line says "Follow-up required for TeamViewModel and RequestsViewModel" — so the overstated claim persists in the commit history even though it's demonstrably false (those
  two ViewModels are UI-layer classes still depending on the sync interface).
```

dogi:
**Review — rating: 74/100**

The goal is right — UI classes shouldn't reach for a sync-facing interface — but three things in the execution work against it.

The biggest is `RequestsViewModel`, which moves *the wrong way*. It was injecting `TeamsMembersRepository` (14 methods) and now injects `TeamsRepository` (44 methods), so a PR about narrowing the UI's dependency surface triples it for that class. The widening isn't incidental either: it's forced by declaring `recordTeamActivity()` on `TeamsRepository` when the only caller in that file also needs `respondToMemberRequest` from `TeamsMembersRepository`. Declaring `recordTeamActivity()` on `TeamsMembersRepository` instead would let that injection stay narrow and cost nothing elsewhere, since `TeamsRepository` already extends it.

Second, `recordTeamActivity()` is a pure alias — its whole body is `syncTeamActivities()`, and `TeamsSyncRepository.syncTeamActivities()` still exists on the same implementation. The result is one operation reachable under two names from the same object, with nothing in either signature telling the next maintainer which is the intended entry point. Worth either a KDoc line saying "UI-facing alias; sync layer calls `syncTeamActivities` directly", or renaming so the split reads as deliberate.

**Nit:** both edited test files keep their now-unused imports — `TeamsSyncRepository` at `TeamViewModelTest.kt:22` and at `RequestsViewModelTest.kt:19` — since only the fields were deleted. Kotlin just warns, so CI stays green. (The stray indentation fix in `RequestsViewModelTest` is a nice touch.)

**Verdict:** the direction is correct and the call-site swaps are faithful, but `RequestsViewModel` ends up more coupled than it started; move `recordTeamActivity()` down to `TeamsMembersRepository` and this is ready.

---

dogi:

All three points are addressed, and the first one properly. `recordTeamActivity()` now lives on `TeamsMembersRepository`, so `RequestsViewModel` goes back to injecting the 14-method interface instead of the 44-method one — the PR no longer works against its own stated goal. The KDoc ("UI-facing alias to record a team activity. The internal sync layer calls `syncTeamActivities()` directly.") answers exactly the question the two-names-one-operation split raised, so the next maintainer knows which entry point is theirs. And both test files now drop the `TeamsSyncRepository` import along with the field, with `RequestsViewModelTest` switching cleanly to `TeamsMembersRepository`.

The declaration site works because `TeamsRepository` extends `TeamsMembersRepository`, so every other call site — `TeamDetailFragment`, `TeamViewModel`, `TeamResourcesFragment` — resolves it unchanged.

**Nit, and I'll own that I proposed this shape:** "record team activity" on a *members*-scoped interface is a slight naming stretch, since `TeamResourcesFragment` calls it after adding and removing resource links, which has nothing to do with membership. It's the right trade against re-widening `RequestsViewModel`, but if this grows a second activity-recording method later, a small `TeamsActivityRepository` would be the cleaner home.

**Verdict:** the coupling regression is gone, the alias is documented, and no imports left behind; ready.

---

Okuro3499 (approved):

PR #15963, opened 2026-08-20
life: smoother personals resources opening (fixes #16070)

### LINKED ISSUE #16070: Refactor openResource extension extraction
Tracking issue for PR #15963.

Replaced regex split with `substringAfterLast` to extract file extension more efficiently, reducing memory allocation.

---
*PR created automatically by Jules for task [5365792263465186297](https://jules.google.com/task/5365792263465186297) started by @dogi*

dogi:
Replaced regex split with `substringAfterLast` to extract file extension more efficiently, reducing memory allocation.

---
*PR created automatically by Jules for task [5365792263465186297](https://jules.google.com/task/5365792263465186297) started by @dogi*


dogi:
**Review — rating: 84/100**

This is a crash fix as well as a simplification, which the description undersells. The old expression

```kotlin
val arr = path?.split("\\.".toRegex())?.dropLastWhile { it.isEmpty() }?.toTypedArray()
when (arr?.get(arr.size - 1)) {
```

throws `ArrayIndexOutOfBoundsException` whenever `dropLastWhile` empties the array — `path = ""` splits to `[""]` and drops to `[]`, giving `arr.size - 1 == -1`, and a path of only dots does the same. `substringAfterLast('.', "")` returns `""` on both and falls through the `when` harmlessly. It also drops a `Regex` compile per call, on the click path.

I checked the no-delimiter case too: the old code returned the whole path (single-element array), the new one returns `""`. Neither matches any branch, so the outcome is the same.

**`.lowercase()` needs a Locale.** Adding case-insensitivity is a real improvement — `.PDF` and `.JPG` open now where they previously fell through silently — but `lowercase()` with no argument uses `Locale.getDefault()`, which follows the *device* locale rather than the app's. On a Turkish-locale device `"GIF".lowercase()` is `"gıf"` with a dotless ı, which does not equal `"gif"`, so a `.GIF` file silently fails to open for exactly the users the change was supposed to help. `lowercase(Locale.ROOT)` costs one argument and makes the folding locale-independent:

```kotlin
val extension = path?.substringAfterLast('.', "")?.lowercase(Locale.ROOT)
```

`"gif"` is the only branch in this `when` with an `i` in it, so that's the single reachable case today — but the trap applies to any extension added later (`tif`, `avi`, `midi`).

**Verdict:** ready to merge once `lowercase` takes `Locale.ROOT`.

---

Okuro3499 (approved):

PR #15964, opened 2026-08-20
all: smoother notifications view modelling (fixes #16071)

### LINKED ISSUE #16071: perf(Notifications): optimize markAsRead to a single pass
Tracking issue for PR #15964.

Optimized `markAsRead` to perform a single pass over the notification list to improve performance when updating individual read statuses.

---
*PR created automatically by Jules for task [15639009190469940771](https://jules.google.com/task/15639009190469940771) started by @dogi*

dogi:
Optimized `markAsRead` to perform a single pass over the notification list to improve performance when updating individual read statuses.

---
*PR created automatically by Jules for task [15639009190469940771](https://jules.google.com/task/15639009190469940771) started by @dogi*


dogi:
**Review — rating: 86/100**

Semantics are preserved across all four cases, which is what matters for a rewrite like this. Not-found → every element maps to itself; found-and-read → same; found-and-unread with `currentFilter == "unread"` → `null` drops it, matching the old `filter { it.id != notificationId }`; found-and-unread otherwise → `copy(isRead = true)`, matching `markAsRead(notificationId)`. And `wasUnread` is set on exactly the same condition, so the `_unreadCount` decrement below is unaffected.

The pass count really does halve: the old path ran `find` and then a second traversal in either `filter` or `markAsRead`.

**Pre-existing, but this PR is standing right next to it:** `wasUnread` is a `var` captured inside `MutableStateFlow.update { }`, which is a compare-and-set retry loop — the lambda can run more than once under contention. If a retry happens after the first attempt set `wasUnread = true`, the flag is never reset, so a concurrent update that already marked the notification read could still let the `_unreadCount` decrement through. Returning the flag out of the update instead of capturing it (e.g. compute the new list and the flag together, or use `updateAndGet` and derive `wasUnread` by comparing) would close it. Not introduced here, but this is the moment where fixing it is nearly free.

**Nit:** the nested `if/else` inside `mapNotNull` is three levels deep for what is really a three-way decision. A `when` on `notif.id != notificationId` / `notif.isRead` / `currentFilter` would read flatter.

**Nit:** `mapNotNull` always builds a new list, where the old code returned `currentList` unchanged in the no-op case. `StateFlow` conflates by `equals` and `List.equals` is elementwise, so no spurious emission — just one throwaway allocation on the miss path.

**Verdict:** ready to merge.

---

Okuro3499 (approved):

PR #15965, opened 2026-08-20
courses: smoother base exams regex splitting (fixes #16072)

### LINKED ISSUE #16072: Refactor regex split to substringBefore in BaseExamFragment
Tracking issue for PR #15965.

In `BaseExamFragment.kt`, the parsing logic for extracting `id` out of `parentId` was relying on a regex split which built two intermediate collections just to take everything before the first `@`.

```kotlin
id = if (sub?.parentId?.contains("@") == true) {
    sub?.parentId?.split("@".toRegex())?.dropLastWhile { it.isEmpty() }?.toTypedArray()?.get(0)
} else {
    sub?.parentId
}
```

This logic was simplified directly to:

```kotlin
id = sub?.parentId?.substringBefore("@")
```

The zero-allocation `substringBefore` gracefully handles absent delimiters by returning the original string, perfectly solving the problem and safely optimizing performance.

Tested via `./gradlew :app:testDefaultDebugUnitTest --tests "*BaseExamFragment*"`.

---
*PR created automatically by Jules for task [16320207046569813459](https://jules.google.com/task/16320207046569813459) started by @dogi*

dogi:
In `BaseExamFragment.kt`, the parsing logic for extracting `id` out of `parentId` was relying on a regex split which built two intermediate collections just to take everything before the first `@`.

```kotlin
id = if (sub?.parentId?.contains("@") == true) {
    sub?.parentId?.split("@".toRegex())?.dropLastWhile { it.isEmpty() }?.toTypedArray()?.get(0)
} else {
    sub?.parentId
}
```

This logic was simplified directly to:

```kotlin
id = sub?.parentId?.substringBefore("@")
```

The zero-allocation `substringBefore` gracefully handles absent delimiters by returning the original string, perfectly solving the problem and safely optimizing performance.

Tested via `./gradlew :app:testDefaultDebugUnitTest --tests "*BaseExamFragment*"`.

---
*PR created automatically by Jules for task [16320207046569813459](https://jules.google.com/task/16320207046569813459) started by @dogi*


dogi:
**Review — rating: 91/100**

Exactly equivalent on every input, and it removes a crash the old form could reach. I walked the cases:

- `parentId` contains `@` → old took element 0 of the split; `substringBefore("@")` is the same prefix.
- `parentId` has no `@` → old returned `sub?.parentId` via the `else` branch; `substringBefore` returns the receiver unchanged when the delimiter is absent. Same value, and the `contains("@")` test that selected between them is now unnecessary.
- `parentId` is `null` → both yield `null`.
- `parentId` is `"@"` → the old code throws. `"@".split(Regex("@"))` gives `["", ""]`, `dropLastWhile { it.isEmpty() }` strips both, and `.get(0)` on the resulting empty array is an `ArrayIndexOutOfBoundsException`. The new form returns `""`.

That last case is a real, if narrow, hazard for a `parentId` coming off a sync payload, and it disappears for free. The `"@".toRegex()` compile per call goes with it — this runs on exam resume.

Collapsing the `if/else` into one expression also removes the double `sub?.parentId` evaluation, which is the kind of thing that goes wrong later when someone makes `sub` a computed property.

**Nit:** `substringBefore('@')` (the `Char` overload) skips the `String.indexOf(String)` path for a one-character delimiter. Immaterial here, but it's the more precise call.

**Verdict:** ready to merge.

---

Okuro3499 (approved):

PR #15966, opened 2026-08-20
courses: smoother submissions exams view modelling (fixes #16073)

### LINKED ISSUE #16073: Optimize exam query filtering in SubmissionViewModel
Tracking issue for PR #15966.

Optimizes query filtering in `SubmissionViewModel` to avoid allocating an intermediate `Map`. Matching `examId`s are now projected directly into a `Set`.

---
*PR created automatically by Jules for task [5384337330227160131](https://jules.google.com/task/5384337330227160131) started by @dogi*

dogi:
Optimizes query filtering in `SubmissionViewModel` to avoid allocating an intermediate `Map`. Matching `examId`s are now projected directly into a `Set`.

---
*PR created automatically by Jules for task [5384337330227160131](https://jules.google.com/task/5384337330227160131) started by @dogi*


dogi:
**Review — rating: 86/100**

Correct and a real saving. `Map.filter { }` builds a whole new `LinkedHashMap` — keys *and* values — and the old code then threw the values away by taking `.keys`. `mapNotNullTo(HashSet())` builds only the id set, so the intermediate map and its entry nodes disappear. The resulting set is identical, and the downstream `examIds.contains(it.parentId)` was already an O(1) hash lookup either way, so nothing changes for the caller.

This runs on every keystroke in the submissions search field, over the full exam map, so trimming the per-query allocation is worth having rather than theoretical.

**Nit:** `mapNotNullTo(HashSet())` reads as a bit of a puzzle at the call site — the destination-collection overloads are the least familiar part of the stdlib collection API. `buildSet { examMap.forEach { (id, exam) -> if (…) add(id) } }` says the same thing more plainly if you'd rather optimise for the next reader; either way the allocation profile is the same.

**Nit:** the predicate `exam.name?.contains(query, ignoreCase = true) == true` does a case-insensitive scan per exam per keystroke. If the exam map ever gets large this is the part that will show up, not the set construction — a lowercased-name cache alongside the map would be the next step.

**Verdict:** ready to merge.

---

Okuro3499 (approved):

PR #55, opened 2018-06-03
Adds sync settings on login activity (fixes #49)

onlyserenity:
* Added Sync settings on login activity

PR #15967, opened 2026-08-20
all: smoother pager list submitting (fixes #16074)

### LINKED ISSUE #16074: perf: remove redundant copies in pager adapters
Tracking issue for PR #15967.

This PR addresses performance issue #55 by removing redundant list copies in the pager adapters.

Changes:
- **`CoursesPagerAdapter`**: Replaced the allocating `listOf(null) + steps` strategy with an inline, allocation-free `AbstractList` wrapper that accesses the underlying lists directly without copying. Replaced `mutableListOf` with a direct assignment `steps = newSteps`.
- **`TeamPagerAdapter`**: Removed redundant `.toList()` copying on the already immutable `newPages` list.

---
*PR created automatically by Jules for task [1484881925745271730](https://jules.google.com/task/1484881925745271730) started by @dogi*

dogi:
This PR addresses performance issue #55 by removing redundant list copies in the pager adapters.

Changes:
- **`CoursesPagerAdapter`**: Replaced the allocating `listOf(null) + steps` strategy with an inline, allocation-free `AbstractList` wrapper that accesses the underlying lists directly without copying. Replaced `mutableListOf` with a direct assignment `steps = newSteps`.
- **`TeamPagerAdapter`**: Removed redundant `.toList()` copying on the already immutable `newPages` list.

---
*PR created automatically by Jules for task [1484881925745271730](https://jules.google.com/task/1484881925745271730) started by @dogi*


dogi:
**Review — rating: 72/100**

Two independent changes with quite different value.

**`TeamPagerAdapter` is fine.** Dropping `.toList()` from `pages = newPages.toList()` removes a defensive copy that the class wasn't making anywhere else anyway — the constructor already takes `private var pages: List<TeamPageConfig>` and holds the caller's reference directly. So this makes `updatePages` consistent with construction rather than introducing new aliasing, and `TeamDetailFragment` passes a freshly built `pageConfigs`.

**`CoursesPagerAdapter` I'd revert.** The two anonymous `AbstractList` views exist to avoid `listOf(null) + steps`:

```kotlin
val oldItems = object : kotlin.collections.AbstractList<String?>() {
    override val size get() = steps.size + 1
    override fun get(index: Int) = if (index == 0) null else steps[index - 1]
}
```

What that buys is two small `ArrayList` allocations, once per `submitList` call — and `submitList` runs when a course's steps load, not per frame or per bind, over a list of maybe five to twenty step ids. What it costs is fourteen lines of machinery in place of two, a fully-qualified `kotlin.collections.AbstractList` written inline twice, and a subtlety that isn't obvious on reading: `oldItems` reads the `steps` **field**, not a snapshot, so it silently changes meaning after `steps = newSteps` two lines below. That's harmless today because `DiffUtil.DiffResult` holds the computed operations rather than the lists, but it's a trap sitting in code that only exists to save an allocation nobody will measure.

`steps` becoming `var steps: List<String>` assigned from `newSteps` is fine on its own and worth keeping — it pairs with the early-return `if (newSteps == steps) return` guard above it.

**Verdict:** needs changes — keep the `TeamPagerAdapter` half and the `var steps` change, drop the `AbstractList` views.

---

dogi:
**Re-review — Rating: 87/100** (was 72)

Both asks landed. The two anonymous `AbstractList` views are gone and `listOf(null) + steps` is back, so the diff logic reads plainly again and nothing depends on a view that silently changes meaning when `steps` is reassigned two lines later.

What's kept is the part that was worth keeping:

- `private var steps: List<String> = emptyList()` with `steps = newSteps` instead of `clear()` + `addAll()`. That drops a full copy per `submitList` and pairs naturally with the `if (newSteps == steps) return` guard above it.
- `TeamPagerAdapter`'s `pages = newPages` without the `.toList()`. That copy was already inconsistent with the constructor, which takes `private var pages: List<TeamPageConfig>` and holds the caller's reference directly — so this makes the two paths agree rather than introducing new aliasing, and `TeamDetailFragment` passes a freshly built `pageConfigs`.

Net result is two small, honest allocation removals with no readability cost, which is what the PR title claimed in the first place.

**Nit:** both adapters now alias caller-supplied lists in `submitList`/`updatePages`. That's fine for every current caller, but if either ever gets handed a `MutableList` the caller keeps mutating, the `newSteps == steps` guard would start comparing a list to itself. Declaring the parameters as `List<…>` already signals read-only intent — no change needed, just worth knowing.

**Verdict:** ready to merge.

---

Okuro3499 (approved):

PR #15968, opened 2026-08-20
login: smoother server dialog extensions url handling (fixes #16075)

### LINKED ISSUE #16075: refactor(SyncActivity): optimize URL stripping in refreshServerList
Tracking issue for PR #15968.

Update `refreshServerList` in `ServerDialogExtensions.kt` to apply the `httpsPrefixRegex` URL stripping exactly once per server candidate instead of multiple times across three separate passes. Pre-compute a list mapping each element to its stripped URL, and use this list to deduce the `pinnedEntry`, filter the deduplicated list, and compute `pinnedIndex` without evaluating the regular expression again.

---
*PR created automatically by Jules for task [1616971581101598763](https://jules.google.com/task/1616971581101598763) started by @dogi*

dogi:
Update `refreshServerList` in `ServerDialogExtensions.kt` to apply the `httpsPrefixRegex` URL stripping exactly once per server candidate instead of multiple times across three separate passes. Pre-compute a list mapping each element to its stripped URL, and use this list to deduce the `pinnedEntry`, filter the deduplicated list, and compute `pinnedIndex` without evaluating the regular expression again.

---
*PR created automatically by Jules for task [1616971581101598763](https://jules.google.com/task/1616971581101598763) started by @dogi*


dogi:
**Review — rating: 86/100**

The precompute is worth having and the rewrite is behaviour-preserving, which took a bit of checking because the selection index moved.

`urlWithoutProtocol` was being recomputed per element in up to three separate passes — the `find`, the `filter`, and then `indexOfFirst` in the `submitList` callback. Pairing each server with its stripped URL once cuts that to one regex replace per entry.

On equivalence: the `finalList` condition folds `pinnedServer != null` into the outer `if` rather than nesting, which lands on the same list in all four combinations. The index is the part that changed shape — the old code searched `finalList` unconditionally, the new one returns a literal `0` when reordering happened and otherwise searches `candidates`. Both are right: in the reordered branch the pinned entry is at index 0 by construction, and in the fall-through branch `finalList === filteredList` while `candidates` is a `map` of it, so positions line up.

**Nit worth acting on:** the three-part condition `showAdditionalServers && urlWithoutProtocol.isNotEmpty() && pinnedEntry != null` is now written twice — once to build the list, once inside the async `submitList` callback to decide the index. Those two must agree or the wrong row gets selected, and the callback fires after a diff so a mismatch wouldn't be obvious. Hoisting `val isPinnedFirst = …` and using it in both places removes the chance of drift.

**Nit:** `candidates.filter { … }.map { it.first }` builds two lists; `mapNotNull { if (it.second != urlWithoutProtocol) it.first else null }` does it in one.

**Verdict:** ready to merge.

---

Okuro3499 (approved):

PR #15969, opened 2026-08-20
life: smoother health examination blood pressure handling (fixes #16076)

### LINKED ISSUE #16076: Fix: split blood pressure on a literal, not a regex
Tracking issue for PR #15969.

Fixes performance issue in HealthExaminationActivity where a regex was compiled on every keystroke in the blood pressure field.

---
*PR created automatically by Jules for task [2427466481444453723](https://jules.google.com/task/2427466481444453723) started by @dogi*

dogi:
Fixes performance issue in HealthExaminationActivity where a regex was compiled on every keystroke in the blood pressure field.

---
*PR created automatically by Jules for task [2427466481444453723](https://jules.google.com/task/2427466481444453723) started by @dogi*


dogi:
**Review — rating: 86/100**

Correct and free. `"/"` has no regex metacharacters, so `split("/".toRegex())` and `split("/")` produce identical output — the only difference is that the second doesn't compile a `Pattern`, and this sits on the submit path where it ran on every validation press.

Worth noting what this one does *not* have: two sibling PRs in the queue (#15963, #15965) replaced the same `split(regex).dropLastWhile { it.isEmpty() }` idiom in places where the result was indexed directly, and both were latent `ArrayIndexOutOfBoundsException`s on empty input. Here the very next line is

```kotlin
if (sysDia.size > 2 || sysDia.isEmpty()) { … error … }
```

so an empty `etBloodpressure` produces an empty array, hits the guard and shows `blood_pressure_should_be_systolic_diastolic` rather than crashing. The guard was already correct; this change is purely the regex removal.

**Nit:** the PR is titled "Fix:" but nothing was broken — `perf:` or `refactor:` matches what changed, and it matters a little for the changelog.

**Nit:** `.dropLastWhile { it.isEmpty() }.toTypedArray()` is still doing more than needed. `split` returns a `List` that indexes and reports `size` perfectly well, so `toTypedArray()` is a copy for nothing; and with the size guard right below, `dropLastWhile` only affects whether trailing empty segments are counted — `"120/"` currently becomes one element rather than two, which is worth being deliberate about since it decides whether that input is accepted.

**Verdict:** ready to merge.

---

Okuro3499 (approved):

PR #15971, opened 2026-08-20
all: smoother notifications repository allocations querying (fixes #16078)

### LINKED ISSUE #16078: perf(NotificationsRepositoryImpl): avoid intermediate Set allocations
Tracking issue for PR #15971.

Avoid intermediate Set allocations in NotificationsRepositoryImpl

---
*PR created automatically by Jules for task [3718161227988690998](https://jules.google.com/task/3718161227988690998) started by @dogi*

dogi:
Avoid intermediate Set allocations in NotificationsRepositoryImpl

---
*PR created automatically by Jules for task [3718161227988690998](https://jules.google.com/task/3718161227988690998) started by @dogi*


dogi:
**Review — rating: 87/100**

Correct, and the one thing that could have made it not correct checks out. Dropping the eager `.toSet()` means `markAsRead`/`deleteByIds` now receive whatever `getByIds` returned rather than a deduplicated copy — but `AppNotification.id` is the `@PrimaryKey` and the query is `SELECT * FROM notifications WHERE id IN (:ids)`, so at most one row comes back per id and the set could never have removed anything. Same rows in, same rows out, with two intermediate `Set`s and their `toList()` copies gone from each method.

Keeping `Set<String>` as the return type means callers are unaffected; the conversion just moved to the boundary where it's actually part of the contract.

**Adjacent, not introduced here:** neither `getByIds` nor `markAsRead` chunks its `IN (:ids)` list, so a caller passing more than 999 ids throws `too many SQL variables` on API 26–30. `markAllUnreadAsRead` is the plausible route to a large set. Both queries name the parameter once, so the safe chunk is 999 rather than the 400 that `getUsersByAnyIds` needs for its double bind — worth a follow-up given #15948 and #15919 hit the same ceiling this week.

**Verdict:** ready to merge.

---

Okuro3499 (approved):

PR #15972, opened 2026-08-20
sync: smoother upload repository querying (fixes #16079)

### LINKED ISSUE #16079: perf(UploadRepositoryImpl): remove redundant collection copies
Tracking issue for PR #15972.

In `UploadRepositoryImpl.kt`, two collection copies were being performed unnecessarily:
1. `examDao.getPendingAdoptedSurveys().map { it } as List<T>` - `.map { it }` allocates a new identical list purely to feed the unchecked cast.
2. `examDao.upsertAll(updated.mapNotNull { it })` - `updated` is `mutableListOf<StepExam>()` and only non-null exams are added, making the filter a no-op copy.

Both redundant copies were removed.

---
*PR created automatically by Jules for task [5312487176588387602](https://jules.google.com/task/5312487176588387602) started by @dogi*

dogi:
In `UploadRepositoryImpl.kt`, two collection copies were being performed unnecessarily:
1. `examDao.getPendingAdoptedSurveys().map { it } as List<T>` - `.map { it }` allocates a new identical list purely to feed the unchecked cast.
2. `examDao.upsertAll(updated.mapNotNull { it })` - `updated` is `mutableListOf<StepExam>()` and only non-null exams are added, making the filter a no-op copy.

Both redundant copies were removed.

---
*PR created automatically by Jules for task [5312487176588387602](https://jules.google.com/task/5312487176588387602) started by @dogi*


dogi:
**Review — rating: 89/100**

Both removals are identity operations, so this is pure subtraction with no behaviour to reason about.

`examDao.getPendingAdoptedSurveys().map { it }` copied a list into an identical list before the cast — the `.map { it }` couldn't change element type or content, it just allocated a second `ArrayList`. And `updated.mapNotNull { it }` can only filter if the element type is nullable; since removing it compiles, `updated` is already `List<Exam>`, which makes that one a copy too. Green CI is the proof for the second, since a nullable element type would have failed the `upsertAll` signature.

Small, but these are on the upload path where the lists can be large, and copies like this are exactly the kind of thing that survives for years because they look like they're doing something.

**Nit:** the real awkwardness in `queryPending` is the `@Suppress("UNCHECKED_CAST")` plus `as List<T>` on every branch — the `.map { it }` was probably added at some point to make a cast look less bare. Out of scope here, but if `UploadQueryContract<T>` carried the element type the casts could go too.

**Verdict:** ready to merge.

---

Okuro3499 (approved):

PR #15973, opened 2026-08-20
all: smoother selection utils membership checking (fixes #16080)

### LINKED ISSUE #16080: perf: drop redundant membership scan in SelectionUtils.handleCheck
Tracking issue for PR #15973.

Removing an item from a `MutableList` via `remove()` gracefully returns `false` if the item is absent. Checking `contains()` prior to removing causes two sequential O(N) linear scans when unchecking a present item.

This commits assigns `list[i]` to a local variable `item` and modifies the uncheck branch to call `selectedItems.remove(item)` directly, optimizing the path while maintaining the original behavior and absent-item safety. Tests have been executed and pass successfully.

---
*PR created automatically by Jules for task [17500028005013566072](https://jules.google.com/task/17500028005013566072) started by @dogi*

dogi:
Removing an item from a `MutableList` via `remove()` gracefully returns `false` if the item is absent. Checking `contains()` prior to removing causes two sequential O(N) linear scans when unchecking a present item.

This commits assigns `list[i]` to a local variable `item` and modifies the uncheck branch to call `selectedItems.remove(item)` directly, optimizing the path while maintaining the original behavior and absent-item safety. Tests have been executed and pass successfully.

---
*PR created automatically by Jules for task [17500028005013566072](https://jules.google.com/task/17500028005013566072) started by @dogi*


dogi:
**Review — rating: 88/100**

Both halves are right. `MutableList.remove(element)` already returns `false` and leaves the list alone when the element isn't present, so the `contains` guard bought nothing and cost a full O(n) scan before `remove` did its own — this turns two traversals into one on the miss path and one on the hit path either way. Hoisting `list[i]` into `item` also drops it from three evaluations to one.

No behaviour changes at the edges: `list[i]` was already evaluated on both branches, so moving it above the `if` doesn't change when an out-of-range `i` throws, and removing an absent element remains a no-op.

**Nit:** `MutableList<T?>` / `List<T?>` means the selection list can hold nulls, and `remove(null)` would then drop the first null rather than the intended item. Pre-existing signature, and callers presumably never pass nulls — but since the function is now doing an unguarded `remove`, non-null type parameters would make that guarantee explicit rather than incidental.

**Verdict:** ready to merge.

---

Okuro3499 (approved):

PR #15975, opened 2026-08-20
sync: smoother retry interceptor testing (fixes #16082)

### LINKED ISSUE #16082: fix: RetryInterceptor cancellation delay
Tracking issue for PR #15975.

Fixes an issue where `RetryInterceptor.backoff` burned a full 250ms sleep slice when a call was cancelled during `proceed()`. The `isCanceled()` check has been hoisted to the top of `backoff()`, and the `while` loop has been slightly restructured to verify cancellation after each sleep block. Added a specific test for cancelling during the sleep slice.

---
*PR created automatically by Jules for task [870720486204443947](https://jules.google.com/task/870720486204443947) started by @dogi*

dogi:
Fixes an issue where `RetryInterceptor.backoff` burned a full 250ms sleep slice when a call was cancelled during `proceed()`. The `isCanceled()` check has been hoisted to the top of `backoff()`, and the `while` loop has been slightly restructured to verify cancellation after each sleep block. Added a specific test for cancelling during the sleep slice.

---
*PR created automatically by Jules for task [870720486204443947](https://jules.google.com/task/870720486204443947) started by @dogi*


dogi:
**Review — rating: 64/100**

I traced both versions statement by statement and I don't think the production change alters behaviour. Writing out the check/sleep sequence for each:

- **Before:** the check sits at the *top* of the loop, so the order is `check → remaining → sleep → check → remaining → sleep → … → check → remaining≤0 → return`.
- **After:** the check moved to entry plus *after* each sleep, giving `check → remaining → sleep → check → remaining → sleep → … → check → remaining≤0 → return`.

The old top-of-loop check *is* the after-sleep check for the previous iteration and *is* the entry check for the first one, so the two sequences are the same points in time. The final iteration matches too: old checks cancellation then returns on `remaining <= 0`; new checks after the last sleep, loops, then returns.

Moving the entry check outside the `try` doesn't change anything either — the only `catch` is for `InterruptedException`, so the `IOException` propagated identically from inside it.

If you're seeing a real cancellation delay in practice, the cause is more likely `MAX_BACKOFF_SLICE_MS` — cancellation can only be observed at slice boundaries, so the worst-case latency is one slice regardless of where the check sits. Shrinking the slice, or waiting on a `CountDownLatch`/monitor that cancellation can signal instead of `Thread.sleep`, is what would actually tighten it. Happy to be corrected if there's a trace showing otherwise.

**The new test is worth keeping, but it can't distinguish the two versions** — with `initialDelay = 5_000L` and cancellation flipped at 300ms, the old code also throws at the next slice boundary. It's genuine new coverage for cancellation-during-backoff, which didn't exist; it just isn't evidence for this diff.

Three things to fix in it regardless:

- **`cancelled` is a plain `var` written from a second thread and read from the interceptor thread** with no `@Volatile` or synchronization. That's a data race, and the read is legally hoistable. `AtomicBoolean` costs nothing here.
- **The thread is never joined**, so it outlives a failing assertion.
- **It burns ~500ms of real wall clock** and depends on timing on a loaded CI runner. `timeProvider` is already injected and already used for the deadline — driving cancellation through a fake clock rather than `Thread.sleep(300)` would make it deterministic and instant.

**Verdict:** needs changes — either show what the restructure fixes, or reframe this as the test-coverage PR it actually is (and fix the race).

---

dogi:

Right call. The production change is withdrawn entirely and the PR is now test-only, which is what the trace supported — the restructured check/sleep sequence was identical to the original.

The rewritten test fixes all three problems:

- **No second thread**, so the data race on a non-`@Volatile` `var` is gone. Driving cancellation from `every { isCanceled() } answers { callCount++; callCount > 1 }` keeps everything on one thread.
- **No unjoined thread** to outlive a failed assertion.
- **`initialDelay = 10L`** instead of 5,000, so the test costs milliseconds rather than half a second of real wall clock.

And it does test the thing it claims: with cancellation reported on the second `isCanceled()` call, the throw happens at the top of the second loop iteration — after a sleep has occurred — so `testCallCancelledDuringBackoffSleep` is an accurate name for what it exercises.

**Nit worth knowing about:** the test is now coupled to the *number* of `isCanceled()` invocations rather than to a state change. If anyone later adds or moves a cancellation check in `backoff` — which is precisely what the reverted diff did — `callCount > 1` starts firing at a different point and the test keeps passing while testing something else. Gating on a flag that the first sleep sets, or advancing the injected `timeProvider`, would pin the intent instead of the call count.

**Nit:** the title is still `fix: RetryInterceptor cancellation delay` for a diff that changes only a test. `test:` would match, and it matters for the changelog.

**Nit:** there's overlap with the existing `Backoff should not have slept for the full delay` case above it. This one adds the exception type and message, which the older one doesn't assert, so it earns its place — just noting they're neighbours.

**Verdict:** ready to merge.

---

Okuro3499 (approved):

PR #15974, opened 2026-08-20
courses: smoother progress activities repositories context querying (fixes #16081)

dogi:
Wrap the heavy repository aggregations in `withContext(io)`.

Three methods did list grouping/mapping on the caller's dispatcher:
`ProgressRepositoryImpl.getCourseProgress`
`ActivitiesRepositoryImpl.getMostOpenedResource`
`CoursesRepositoryImpl.getCourseById`/`getCourseSteps`.

Wrapped each body in `withContext(dispatcherProvider.io)`, injecting the provider into
`ActivitiesRepositoryImpl`. Updated tests to mock the `DispatcherProvider` with unconfined dispatchers.

---
*PR created automatically by Jules for task [2060952662471295631](https://jules.google.com/task/2060952662471295631) started by @dogi*

PR #15977, opened 2026-08-20
chat: smoother repository search view modelling (fixes #16084)

### LINKED ISSUE #16084: Refactor chat search logic from ViewModel to Repository
Tracking issue for PR #15977.

Refactor chat search logic from ViewModel to Repository

Moves the search implementation, including precomputed chat indices and caching, out of `ChatViewModel` into `ChatRepositoryImpl`. Introduces a domain-level `ChatSearchMode` enum to avoid leaking presentation layer search toggles to the repository. The ViewModel now safely delegates queries directly.

---
*PR created automatically by Jules for task [2876165767112124184](https://jules.google.com/task/2876165767112124184) started by @dogi*

dogi:
Refactor chat search logic from ViewModel to Repository

Moves the search implementation, including precomputed chat indices and caching, out of `ChatViewModel` into `ChatRepositoryImpl`. Introduces a domain-level `ChatSearchMode` enum to avoid leaking presentation layer search toggles to the repository. The ViewModel now safely delegates queries directly.

---
*PR created automatically by Jules for task [2876165767112124184](https://jules.google.com/task/2876165767112124184) started by @dogi*


dogi:
**Review — rating: 58/100**

Moving search out of the ViewModel is defensible layering, and the matching logic itself was transplanted faithfully — the four-bucket ranking, the `break` on first hit per chat, and the `ChatSearchMode` enum replacing the two-boolean `isFullSearch`/`isQuestion` signature are all improvements. But the way state moved with it turns a correctly-scoped cache into process-wide mutable state.

**`ChatRepositoryImpl` is `@Singleton` and now holds per-user search state.**

```kotlin
@Singleton
class ChatRepositoryImpl @Inject constructor(…) {
    private var allChats: List<ChatHistory> = emptyList()
    private var precomputedChats: List<PrecomputedChat> = emptyList()
```

Three consequences:

- **`searchChats` silently depends on a prior `getChatHistoryForUser` call.** Nothing in the interface says so; call it first and you get `emptyList()` rather than an error. Today one call path happens to order them correctly, which is exactly the kind of invariant that breaks when a second caller appears.
- **The state is shared across users.** In the ViewModel these fields were scoped to a ViewModel instance and died with it. On a singleton they persist across user switches, so a search can return the previous user's chats until the next `getChatHistoryForUser`.
- **The two fields are written without synchronization and read from `Dispatchers.Default`.** They're assigned sequentially with no `@Volatile` or lock, so a concurrent `searchChats` can observe `allChats` updated while `precomputedChats` is still the old value.

Passing the precomputed list in — `searchChats(chats: List<ChatHistory>, query: String, mode: ChatSearchMode)` — or keeping the precompute in the ViewModel and moving only the pure matching, gets the layering win without any of this.

**`allChats` in the repository is write-only.** It's assigned in `getChatHistoryForUser` and read only by the `buildPrecomputedChats(allChats)` on the next line; the ViewModel keeps its own `allChats` for the empty-query reset. The repository field can go.

**A stray reasoning transcript got committed in the test**, and the test around it is broken in two ways:

```kotlin
fun `searchChats with empty query returns empty when filtered list logic is applied`() = runTest {
    …
    // Given searchByTitle splits by " " and filterNot isEmpty(), an empty string leads to empty queryParts
    // Wait, actually, the repository search logic returns all if empty? No, searchByTitle returns …
    // Actually, let's just assert the result size.
    assertTrue(result.isNotEmpty())
```

The name says "returns empty" and the assertion says `isNotEmpty()`. And `assertTrue(result.isNotEmpty())` doesn't pin the behaviour either way — the interesting fact is that an empty query makes `normalizedQueryParts` empty, so `.all { }` is vacuously true and *every* chat matches. Assert that (`assertEquals(2, result.size)`) and the test earns its place. This is the second stray thinking-out-loud comment this week (#15948 had one) — worth a pre-push skim.

**Both test files hand-mock `DispatcherProvider` with `Dispatchers.Unconfined`** instead of `TestDispatcherProvider`, same as #15974. `Unconfined` runs eagerly on the calling thread, which is precisely the determinism a `TestDispatcher` exists to provide.

**Nit:** `import kotlinx.coroutines.withContext` is inserted between `services.SharedPrefManager` and `services.sync.ServerUrlMapper` in `ChatRepositoryImpl`, and `import kotlinx.coroutines.Dispatchers` lands after the `org.ole` block in `ChatRepositoryImplTest`.

**Verdict:** needs changes — make the search stateless rather than caching per-user data on a singleton.

---

dogi:
**Re-review — Rating: 80/100** (was 58)

The singleton state is gone. `searchChats(query, mode, chats)` takes the list as a parameter, `allChats`/`precomputedChats` are no longer fields on the `@Singleton`, and `buildPrecomputedChats` is called inside the search. That removes all three problems at once: no hidden ordering contract, no cache surviving a user switch, and nothing to race on.

`getChatHistoryForUser` now sorts and returns without storing, which puts the sort in the repository where it belongs while leaving lifetime to the caller. The stray reasoning comment is gone, and the third test now asserts `assertEquals(2, result.size)` — pinning the actually-interesting behaviour, that an empty query makes `normalizedQueryParts` empty so `.all { }` is vacuously true and everything matches. Both test files use `TestDispatcherProvider(StandardTestDispatcher())`.

**One thing the statelessness cost you, worth a follow-up:** `buildPrecomputedChats` now runs on every `searchChats` call, and `ChatViewModel.searchChats` fires per keystroke (`searchJob?.cancel()` then relaunch, no debounce). So every keystroke re-runs `Utilities.normalizeText` over the title, every query and every response of every chat. Before this PR the precompute happened once per history load and was reused across keystrokes. Cancellation limits the damage, but for a user with long histories this is real work on the typing path.

The fix keeps the statelessness: have `getChatHistoryForUser` return the precomputed form (or expose `buildPrecomputedChats` and let the ViewModel hold the result and pass it in), so the precompute follows the data's lifetime rather than the query's. Given the rest of this queue is micro-optimisations, it'd be odd to leave a per-keystroke recompute in.

**Nit:** the third test's name still says `searchChats with empty query returns empty when filtered list logic is applied` while it asserts 2 results. The assertion is right now; the name still says the opposite.

**Nit:** `import kotlinx.coroutines.withContext` is wedged between `model.Data` and `model.News` in `ChatRepositoryImpl`; `kotlinx.coroutines.Dispatchers` sits between `model.News` and `services.SharedPrefManager` in `ChatRepositoryImplTest` (and looks unused now); and both test files construct `kotlinx.coroutines.test.StandardTestDispatcher()` fully qualified rather than importing it.

**Verdict:** ready to merge; please open a follow-up for the per-search precompute.

---

Okuro3499 (approved):

PR #15978, opened 2026-08-20
all: smoother code style guide indexing (fixes #16085)

### LINKED ISSUE #16085: docs: forbid @Index on json lists queried with wildcards
Tracking issue for PR #15978.

Documents the limitation of `@Index` on JSON list fields, as proposed by the issue "28 — Index the userId shelf columns (needs a different fix than proposed)". The index would trigger a database version bump (resulting in dropped data on sync) while providing no performance benefit for `LIKE '%"id"%'` queries due to leading wildcards. This implements the "Keep the finding, discard the remedy" instruction by explicitly forbidding this pattern in `CODE_STYLE_GUIDE.md`.

---
*PR created automatically by Jules for task [14831741354601771006](https://jules.google.com/task/14831741354601771006) started by @dogi*

dogi:
Documents the limitation of `@Index` on JSON list fields, as proposed by the issue "28 — Index the userId shelf columns (needs a different fix than proposed)". The index would trigger a database version bump (resulting in dropped data on sync) while providing no performance benefit for `LIKE '%"id"%'` queries due to leading wildcards. This implements the "Keep the finding, discard the remedy" instruction by explicitly forbidding this pattern in `CODE_STYLE_GUIDE.md`.

---
*PR created automatically by Jules for task [14831741354601771006](https://jules.google.com/task/14831741354601771006) started by @dogi*


dogi:
**Review — rating: 88/100**

Both halves of the rule are accurate, and the second one is the half that matters.

The index claim is right: SQLite can sometimes use a B-tree for `LIKE 'prefix%'`, but never for a leading wildcard, so `LIKE '%"id"%'` over a JSON-list column is a full scan with or without the index.

The cost claim is the part worth writing down. Per `CLAUDE.md`, `RoomModule` builds with `fallbackToDestructiveMigration(true)` and there are no hand-written migrations, so any schema change — including one that adds an index and changes nothing else — forces a version bump, drops the database, and re-pulls from CouchDB. That discards unsynced local writes on every device that hasn't synced since. Trading pending user data for an index that does nothing is a bad deal, and stating it as a rule is the right way to stop someone rediscovering it.

Placement fits: it lands in the "real patterns that have caused bugs in this codebase before" section next to the `CancellationException` and main-thread rules, which is the same register.

**Nit:** it would land harder with the specific column that prompted it. "like `userId`" is doing a lot of work — naming the actual entity and query (`Meetup.userId`, or whichever it was) turns the rule into something a reader can go look at.

**Nit:** the paragraph is one long sentence chain. Splitting the mechanism from the cost — index does nothing / bump wipes unsynced data — would make the second half harder to skim past, since that's the expensive one.

**CI note:** both build jobs are green; the two test shards were still running when I looked. The diff touches only `docs/CODE_STYLE_GUIDE.md`, so there's no Kotlin input to `testDefaultDebugUnitTest` and no way for it to fail on this change.

**Verdict:** ready to merge.

---

Okuro3499 (approved):

PR #15979, opened 2026-08-21
all: smoother stable id utils generating (fixes #16144)

### LINKED ISSUE #16144: perf: implement stable IDs on list adapters
Tracking issue for PR #15979.

Implemented stable ID generation on the app's hottest `ListAdapter`s to improve `RecyclerView` track animations. Avoided `hashCode()` and implemented a custom `StableIdGenerator` that calculates FNV-1a 64-bit hashes for reliable, collision-resistant numeric IDs. Enabled on `CoursesAdapter`, `ResourcesAdapter`, `SurveysAdapter`, `TeamsAdapter`, and `VoicesAdapter`.

---
*PR created automatically by Jules for task [1453544108449837608](https://jules.google.com/task/1453544108449837608) started by @dogi*

dogi:
Implemented stable ID generation on the app's hottest `ListAdapter`s to improve `RecyclerView` track animations. Avoided `hashCode()` and implemented a custom `StableIdGenerator` that calculates FNV-1a 64-bit hashes for reliable, collision-resistant numeric IDs. Enabled on `CoursesAdapter`, `ResourcesAdapter`, `SurveysAdapter`, `TeamsAdapter`, and `VoicesAdapter`.

---
*PR created automatically by Jules for task [1453544108449837608](https://jules.google.com/task/1453544108449837608) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6756
-        versionName = "0.67.56"
+        versionCode = 6757
+        versionName = "0.67.57"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/ui/courses/CoursesAdapter.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/courses/CoursesAdapter.kt
@@ -33,6 +33,7 @@ import org.ole.planet.myplanet.utils.CourseSubjectClassifier
 import org.ole.planet.myplanet.utils.DiffUtils
 import org.ole.planet.myplanet.utils.ListViewMode
 import org.ole.planet.myplanet.utils.SelectionUtils
+import org.ole.planet.myplanet.utils.StableIdGenerator
 import org.ole.planet.myplanet.utils.UrlUtils
 
 class CoursesAdapter(
@@ -124,6 +125,13 @@ class CoursesAdapter(
         if (context is OnHomeItemClickListener) {
             homeItemClickListener = context
         }
+        setHasStableIds(true)
+    }
+
+    override fun getItemId(position: Int): Long {
+        val item = getItem(position)
+        val id = StableIdGenerator.generateStringId(item.courseId)
+        return if (id != RecyclerView.NO_ID) id else StableIdGenerator.generateFallbackId(item)
     }
 
     fun setViewMode(mode: ListViewMode, onChanged: (() -> Unit)? = null) {
--- a/app/src/main/java/org/ole/planet/myplanet/ui/courses/CoursesAdapter.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/courses/CoursesAdapter.kt
@@ -32,6 +32,7 @@ import org.ole.planet.myplanet.utils.CourseSubjectClassifier
 import org.ole.planet.myplanet.utils.DiffUtils
 import org.ole.planet.myplanet.utils.ListViewMode
 import org.ole.planet.myplanet.utils.SelectionUtils
+import org.ole.planet.myplanet.utils.StableIdGenerator
 import org.ole.planet.myplanet.utils.UrlUtils
 
 class CoursesAdapter(
@@ -125,7 +126,9 @@ class CoursesAdapter(
     }
 
     override fun getItemId(position: Int): Long {
-        return org.ole.planet.myplanet.utils.StableIdGenerator.generateStringId(getItem(position).courseId)
+        val item = getItem(position)
+        val id = StableIdGenerator.generateStringId(item.courseId)
+        return if (id != RecyclerView.NO_ID) id else StableIdGenerator.generateFallbackId(item)
     }
 
     fun setViewMode(mode: ListViewMode, onChanged: (() -> Unit)? = null) {
--- a/app/src/main/java/org/ole/planet/myplanet/ui/courses/CoursesAdapter.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/courses/CoursesAdapter.kt
@@ -121,6 +121,11 @@ class CoursesAdapter(
         if (context is OnHomeItemClickListener) {
             homeItemClickListener = context
         }
+        setHasStableIds(true)
+    }
+
+    override fun getItemId(position: Int): Long {
+        return org.ole.planet.myplanet.utils.StableIdGenerator.generateStringId(getItem(position).courseId)
     }
 
     fun setViewMode(mode: ListViewMode, onChanged: (() -> Unit)? = null) {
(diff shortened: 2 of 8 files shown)


dogi:
**Review — rating: 66/100**

The FNV-1a implementation is correct — I checked the constants. `-3750763034362895579` is `0xcbf29ce484222325` (the 64-bit offset basis) as a signed long, `1099511628211` is `0x100000001b3` (the prime), and the xor-then-multiply order is FNV-1a rather than FNV-1. But the adapters it's wired into have a collision that stable IDs don't tolerate.

**`RecyclerView.NO_ID` is returned for a nullable key, and multiple items can hit it.**

```kotlin
if (key.isNullOrEmpty()) return RecyclerView.NO_ID
```

`TeamsAdapter.getItemId` keys on `getItem(position)._id`, and `TeamDetails._id` is declared `val _id: String?` — that's the CouchDB document id, which is null for any team created locally and not yet synced. So every unsynced team reports `-1`, and they all report the *same* `-1`.

That matters specifically because of `setHasStableIds(true)`. With stable IDs, `Recycler.getScrapOrCachedViewForId()` matches a scrapped holder by id, and `tryGetViewHolderForPositionByDeadline` only rebinds when the holder is unbound, invalid, or flagged for update. A holder scrapped from one `-1` item can therefore be handed to a different `-1` item without rebinding — two unsynced teams rendering each other's content during a scroll or an update. `VoicesAdapter` has the same exposure: its own `areItemsTheSame` guards `oldItem.id.isNotEmpty()`, which is direct evidence that empty ids occur in that list.

**Stable IDs are also redundant here.** All five of these are `ListAdapter`s, so `AsyncListDiffer` already establishes item identity through `areItemsTheSame` — `TeamsAdapter` on `_id`, `VoicesAdapter` on `id`. Adding `getItemId` gives RecyclerView a second, independent identity channel. Where the two agree it buys nothing; where they disagree or collide on `NO_ID` it produces exactly the mis-binding above. AndroidX's guidance is that `DiffUtil`-backed adapters don't need stable IDs, and the PR doesn't show a problem this solves — no jank trace, no before/after.

If there's a specific symptom driving this (lost scroll position, animations restarting), it'd be worth naming, because the fix for that is usually in the diff callback rather than a parallel id scheme.

**Nit:** `org.ole.planet.myplanet.utils.StableIdGenerator.generateStringId(...)` is written fully qualified in all five adapters. One import each.

**Nit:** the KDoc says "FNV-1a 64-bit hash algorithm", but it hashes `key[i].code` — UTF-16 code units, not bytes. Fine for ASCII CouchDB ids, just not what canonical FNV-1a computes, and worth saying so since someone may later compare hashes against another implementation.

**Nit:** `testGenerateStringId_stableAndCollisionResistant` asserts one pair of strings doesn't collide. That tests determinism (`id1 == id3`) well; it doesn't test collision resistance. Splitting the two names would be more honest.

**Verdict:** needs changes — `NO_ID` can't be shared between items under `setHasStableIds(true)`; and the case for stable IDs on `ListAdapter` needs making first.

---

dogi:
**Re-review — Rating: 80/100** (was 66)

The collision is closed. `generateFallbackId` hands every keyless item its own descending id from a `WeakHashMap`, so the case I was worried about — several `-1` items letting `getScrapOrCachedViewForId()` hand one item's holder to another without rebinding — can't happen any more. `TeamDetails._id` being null for unsynced teams no longer maps them all onto `RecyclerView.NO_ID`.

The thing to know about the fix is that it's stable **per instance**, not per logical item, and the KDoc says so honestly. That's the right guarantee for the two adapters whose items are value-like (`TeamDetails` and `SurveyInfo` are data classes, so structurally equal items hash to the same fallback across emissions), but `VoicesAdapter` holds `News` and `SurveysAdapter` holds `StepExam` — both `open class` Room entities with no `equals` override. Room materialises a fresh instance on every emission, so a keyless `News` gets a *new* fallback id each time the flow ticks, and RecyclerView can't match the scrap it just put away. It's not a rendering bug — it's the stable-ID benefit quietly not applying, in the adapter that most needs it. `VoicesAdapter.areItemsTheSame` guarding `oldItem.id.isNotEmpty()` is the evidence that empty ids actually occur there. Deriving the fallback from a content key (say `"${item.time}|${item.message}"`) rather than instance identity would make it hold for those two as well.

Worth noting the mirror-image edge on the data-class side: two `TeamDetails` that are structurally equal *and* both have a null `_id` share one fallback id, which is the original failure mode in miniature. Indistinguishable rows, so low stakes, but it's the reason instance identity and content identity can't both be right here.

I re-checked the FNV-1a constants and they're still correct (`0xcbf29ce484222325`, `0x100000001b3`, xor-then-multiply), and the `"Aa"`/`"BB"` test is a good choice since those collide under `String.hashCode()`.

**Nit:** `nextFallbackId` is a mutable counter on a global `object` shared by every adapter — fine while `getItemId` is main-thread-only, but worth an `AtomicLong` or a per-adapter generator if that ever stops being true. The test file doesn't cover the case that matters most: two `equals` instances resolving to the same fallback id.

**Verdict:** the mis-binding is fixed; shipping, with the per-instance caveat as the follow-up for `VoicesAdapter` and `SurveysAdapter`.

---

Okuro3499 (approved):

dogi:
**Heads up: this branch was rebuilt, so existing checkouts of it are stale.**

All four checks were red on `8e71c89` with a Kotlin compilation error. The merge that brought master in was clean at the text level — the diff against the master side was exactly the intended change — but it carried a semantic conflict. `SurveysAdapter` was written against the old master, where the adapter's item type had an `id`; master had since moved it to `ListAdapter<SurveyRow, …>`, and `SurveyRow(exam, surveyInfo, formState)` has no `id` of its own. Git merged both sides without a conflict marker and the result didn't compile, which took down both builds and both test shards.

Rather than stack a fix on top, the branch is now current master plus the change as one commit. `SurveysAdapter.getItemId` keys off `item.exam.id`. The other four adapters check out unchanged against today's master: `Course.courseId`, `ResourceItem.id`, `TeamDetails._id`, `News.id` all resolve. CI is green on the rebuilt head and the PR is mergeable.

On the re-review nits, in a follow-up commit:

- `@Synchronized` on `generateFallbackId` rather than an `AtomicLong`. The counter isn't the only unguarded state — `WeakHashMap` isn't thread-safe either, so making just the counter atomic would cover the smaller half of the problem.
- KDoc corrected. It claimed the ID was stable per *object instance*; `WeakHashMap` keys by `equals`, which is precisely the distinction the review turns on, and the doc now says which item types get which guarantee.
- Tests for both halves: equal keys share a fallback ID, identity-equality keys don't.

The per-instance caveat itself is left as the follow-up you called it. Worth noting it still applies to `SurveysAdapter` after the master merge: `SurveyRow` is a data class, but its generated `equals` delegates to `StepExam`, which is an open Room entity with no `equals` override — so a keyless `SurveyRow` still gets a fresh fallback ID per emission, same as `VoicesAdapter` with `News`. Deriving the fallback from a content key would fix both together, and that's a change to how each adapter picks its key rather than something local to `StableIdGenerator`.

---

PR #15981, opened 2026-08-21
enterprises: smoother finances transaction view modelling (fixes #16145)

### LINKED ISSUE #16145: Refactor EnterprisesFinancesFragment transaction creation to ViewModel
Tracking issue for PR #15981.

Refactored `EnterprisesFinancesFragment` by moving the transaction creation logic into `EnterprisesFinancesViewModel`.

1. Added `createTransaction` method in `EnterprisesFinancesViewModel` to handle the database write side.
2. Exposed a `transactionCreated: SharedFlow<Result<Unit>>` from the ViewModel to signal UI events.
3. Updated `EnterprisesFinancesFragment` to call the ViewModel and collect the `transactionCreated` flow instead of directly invoking the repository from its click listener.

---
*PR created automatically by Jules for task [8838762243590058953](https://jules.google.com/task/8838762243590058953) started by @dogi*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved enterprise transaction creation handling.
  * Added clear success notifications after transactions are created.
  * Displays localized error messages, with a fallback message when needed.
  * Improved lifecycle-aware processing to help prevent transaction results from being lost.

dogi:
Refactored `EnterprisesFinancesFragment` by moving the transaction creation logic into `EnterprisesFinancesViewModel`.

1. Added `createTransaction` method in `EnterprisesFinancesViewModel` to handle the database write side.
2. Exposed a `transactionCreated: SharedFlow<Result<Unit>>` from the ViewModel to signal UI events.
3. Updated `EnterprisesFinancesFragment` to call the ViewModel and collect the `transactionCreated` flow instead of directly invoking the repository from its click listener.

---
*PR created automatically by Jules for task [8838762243590058953](https://jules.google.com/task/8838762243590058953) started by @dogi*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved enterprise transaction creation handling.
  * Added clear success notifications after transactions are created.
  * Displays localized error messages, with a fallback message when needed.
  * Improved lifecycle-aware processing to help prevent transaction results from being lost.


dogi:
**Review — rating: 76/100**

The layering move is right, and it fixes something real that the description undersells: `createTransaction` ran in `viewLifecycleOwner.lifecycleScope`, so backgrounding or rotating mid-submit cancelled the whole thing — including the image upload. On `viewModelScope` it now completes. Hoisting `val capturedDate = date ?: return@setPositiveButton` out of the coroutine is clearer than the old `return@launch` too.

**But the event flow can't deliver when the fragment isn't listening.**

```kotlin
private val _transactionCreated = MutableSharedFlow<Result<Unit>>()
```

The no-argument constructor gives `replay = 0`, `extraBufferCapacity = 0`, `onBufferOverflow = SUSPEND`. So `emit` suspends until a subscriber actually receives the value — and `collectLatestWhenStarted` unsubscribes on STOP. That's precisely the case the `viewModelScope` change was made for: the user submits, backgrounds the app while the image uploads, the result arrives with no collector, and `emit` parks the coroutine until they come back. Then the toast fires, minutes later and out of context.

`DashboardViewModel` in this repo already solves it the same way you'd want to here:

```kotlin
private val _challengeDialogEvent = MutableSharedFlow<ChallengeDialogData>(extraBufferCapacity = 1)
```

With `extraBufferCapacity = 1` the emit completes immediately and the buffered event is delivered on the next collection. A `Channel(Channel.BUFFERED).receiveAsFlow()` gives the same one-shot semantics. Either is a one-line change.

**Nit:** `collectLatestWhenStarted` on a discrete event flow means a second result can cancel the first collector body mid-flight. Harmless for a toast, but `collectWhenStarted` is the better fit for events you want each of.

**Nit:** `Result<Unit>` crossing a flow boundary works, but a small sealed type (`TransactionResult.Success` / `.Failed(message)`) would let the ViewModel decide the message and keep `getString` lookups out of the collector — worth considering if more outcomes get added.

**Verdict:** needs changes — give the event flow a buffer so the result isn't held hostage to the fragment's lifecycle.

---

dogi:

First, a correction to my last review, since you changed code on the strength of it. I said the bare `MutableSharedFlow<Result<Unit>>()` would make `emit` suspend until a collector exists. That's wrong: with `replay = 0` and no subscribers, `SharedFlowImpl.tryEmitLocked` takes the `nCollectors == 0` fast path and returns `true`, so `emit` completes immediately and the value is dropped. The rendezvous suspension only happens when a subscriber *is* attached and hasn't picked the value up yet. So `extraBufferCapacity = 1` does remove that suspension — worth having, and it matches `FeedbackDetailViewModel._events`, which is the closest analogue in the repo — but it does not rescue a toast emitted while the fragment is stopped. That case is still a silent drop, before and after. Guaranteed delivery would need `replay = 1` plus consumption tracking, which for a toast isn't worth the machinery; I'd leave it as-is and accept the drop rather than replay a stale toast on every return to STARTED.

None of that touches the actual merit here, which holds up: moving `createTransaction` into `viewModelScope` means backgrounding the screen mid-upload no longer cancels the write, and that's a real fix. `collectWhenStarted` (not `collectLatest`) is the right choice for a discrete event, and hoisting `capturedDate` to `return@setPositiveButton` keeps the same semantics as the old `return@launch`.

**Nit:** after this diff, `androidx.lifecycle.lifecycleScope` and `kotlinx.coroutines.launch` are both unused in the fragment — line 308 was their only consumer, and line 338's `pickImageLauncher.launch(...)` is an `ActivityResultLauncher` method, not the coroutine builder. Kotlin only warns on unused imports so CI stays green either way. Two follow-ups worth more than the nit: `FileUtils.readBytesFromUri` still reads the whole image into a `ByteArray` on the main thread in the click handler, which is the next obvious thing to push into the ViewModel now that the call is there; and `EnterprisesFinancesViewModelTest` already exists with one test, so a `createTransaction` case asserting the emitted `Result` is nearly free.

**Verdict:** sound change, and the buffer parameter is right even though my stated reason for it wasn't.

---

Okuro3499 (dismissed):
```
File: app/src/test/java/org/ole/planet/myplanet/ui/enterprises/EnterprisesFinancesViewModelTest.kt                                                                               
    Problem: No test covers createTransaction() or transactionCreated — the only method under test is the pre-existing getTeamTransactions. 
```

dogi:
**Re-review — Rating: 88/100** (was 84)

The two `createTransaction` tests I asked for are here and they're built correctly for a replay-0 `SharedFlow`: the collector is started on `backgroundScope` with an `UnconfinedTestDispatcher` *before* `createTransaction` runs, so the emission can't be missed the way a naïve `first()` after the call would miss it. Success and failure are both covered, and `assertEquals(Result.failure<Unit>(error), results[0])` holds because it's the same exception instance on both sides.

Production code is unchanged from the round I reviewed, and still right: `viewModelScope` keeps the write alive across backgrounding, `collectWhenStarted` is the correct collector for a discrete event, and `extraBufferCapacity = 1` matches `FeedbackDetailViewModel._events`.

**Nit:** the two follow-ups from last round are both still open. `androidx.lifecycle.lifecycleScope` and `kotlinx.coroutines.launch` are unused in the fragment now that line 308 is gone (line 338's `pickImageLauncher.launch(...)` is an `ActivityResultLauncher` method, not the coroutine builder). And `FileUtils.readBytesFromUri` still reads the whole image into a `ByteArray` on the main thread inside the click handler — the natural next step now that there's a ViewModel call to move it into. In the tests, `job.cancel()` is redundant on a `backgroundScope` job, and `testDispatcher.scheduler.advanceUntilIdle()` can just be `advanceUntilIdle()` inside `runTest`.

**Verdict:** ready.

---

Okuro3499 (approved):

PR #15932, opened 2026-08-20
teams: smoother voices repository change flowing (fixes #16051)

dogi:
Removed unnecessary `.toSet()` and `.toList()` allocations in the `distinctUntilChanged` predicates for `getCommunityNews` and `getDiscussionsByTeamIdFlow` in `VoicesRepositoryImpl`, replacing them with direct `==` structural equality checks to improve performance and reduce memory churn.

---
*PR created automatically by Jules for task [12947697254878949128](https://jules.google.com/task/12947697254878949128) started by @dogi*

PR #15982, opened 2026-08-21
dashboard: smoother activities repository offline logins flowing (fixes #16215)

### LINKED ISSUE #16215: Optimize getOfflineLogins flow and run chart computation off main thread
Tracking issue for PR #15982.

This commit removes the unnecessary `suspend` keyword from the `getOfflineLogins` flow function in the Activities repository layer and adds `distinctUntilChanged` to avoid redundant emissions. In the UI layer, it flattens a nested coroutine launch in `ActivitiesFragment` by employing `repeatOnLifecycle` natively and shifts the heavy chart computation (`computeMonthlyCounts`) off the main thread onto the default dispatcher. Test files were adjusted accordingly to account for the removal of `suspend`.

---
*PR created automatically by Jules for task [4108740194802974110](https://jules.google.com/task/4108740194802974110) started by @dogi*

dogi:
This commit removes the unnecessary `suspend` keyword from the `getOfflineLogins` flow function in the Activities repository layer and adds `distinctUntilChanged` to avoid redundant emissions. In the UI layer, it flattens a nested coroutine launch in `ActivitiesFragment` by employing `repeatOnLifecycle` natively and shifts the heavy chart computation (`computeMonthlyCounts`) off the main thread onto the default dispatcher. Test files were adjusted accordingly to account for the removal of `suspend`.

---
*PR created automatically by Jules for task [4108740194802974110](https://jules.google.com/task/4108740194802974110) started by @dogi*


dogi:
**Review — rating: 68/100**

Two of the three changes are good. The third is the headline one and it does nothing.

**`suspend fun getOfflineLogins` → `fun` is correct.** A function that only hands back a `Flow` has nothing to suspend for, Room's `Flow`-returning DAO methods are non-suspend by requirement, and the test's `coEvery` → `every` follows properly.

**Moving `computeMonthlyCounts` to `withContext(dispatcherProvider.default)` is the right dispatcher** for a `mapNotNull`/`filter`/`map`/group chain, and it's genuinely off the collector's Main thread now.

**`.distinctUntilChanged()` on this flow is inert.**

```kotlin
return offlineActivityDao.observeByUserNameAndType(...).distinctUntilChanged()
```

`OfflineActivity` is `open class OfflineActivity` with no `equals` override, so it inherits identity equality. The bare `distinctUntilChanged()` compares consecutive `List<OfflineActivity>` values with `equals`, which for a list means element-by-element `equals` — and Room materialises fresh instances on every emission. Two structurally identical query results therefore compare unequal, and nothing is ever filtered. This is the trap `docs/CODE_STYLE_GUIDE.md` covers: on Room `@Entity` open classes the predicate form is required. Something like

```kotlin
    old.size == new.size && old.zip(new).all { (a, b) -> a.id == b.id && a.loginTime == b.loginTime }
}
```

would actually dedup — and since the chart only consumes `loginTime`, comparing just the times is enough. As written the flow re-renders the chart on every Room invalidation exactly as before.

**The test re-introduces the dispatcher anti-pattern that #15974 is removing in this same batch.**

```kotlin
fragment.dispatcherProvider = object : DispatcherProvider {
    override val main = Dispatchers.Unconfined
    …
```

`TestDispatcherProvider` exists at `app/src/test/java/org/ole/planet/myplanet/utils/TestDispatcherProvider.kt` for this, and `Unconfined` gives up the ordering determinism a `TestDispatcher` provides. `#15974` just converted two files away from exactly this shape; worth landing on the same side.

**Nit:** the fragment now hand-rolls `viewLifecycleOwner.repeatOnLifecycle(STARTED) { flow.collectLatest { … } }`, which is the body of the `collectLatestWhenStarted` helper it stopped importing. Avoiding the nested `launch` is a fair reason, but `collectLatestWhenStarted(activitiesRepository.getOfflineLogins(userName)) { … }` still works from inside the outer coroutine and keeps one implementation of the pattern.

**Verdict:** needs changes — the `distinctUntilChanged` needs the predicate form to do anything, and the test should use `TestDispatcherProvider`.

---

dogi:

Both blockers are fixed, and the predicate is the right one. `OfflineActivity` is an `open class` with `@PrimaryKey var id: String` and `loginTime: Long?` and no `equals` override, which is why the bare `distinctUntilChanged()` was inert; comparing `id` and `loginTime` pairwise now actually filters, and since `computeMonthlyCounts` reads only `loginTime`, that pair is a superset of what the consumer can observe. Polarity is right too — the lambda returns `true` for "equivalent, drop it". Dropping `suspend` from `getOfflineLogins` is correct for a Room `Flow` builder and source-compatible for the one production caller, and `withContext(dispatcherProvider.default)` is the right dispatcher for a CPU-bound group-and-count rather than `io`. The test now uses `TestDispatcherProvider(UnconfinedTestDispatcher())`, matching #15974, and `coEvery` → `every` on the non-suspend DAO function is a correct tightening.

One heads-up rather than a criticism: `old.size == new.size && old.zip(new).all { … }` is exactly the shape already in `VoicesRepositoryImpl` at lines 157 and 189, so this matches house style as it stands on master — but #15932 ("drop allocations in `VoicesRepositoryImpl` distinctUntilChanged predicates") is open and `ready` and is reworking those same predicates to stop allocating the intermediate `Pair` list. Whichever lands second should pick up the other's shape so the two don't diverge.

**Nit:** the dedup itself is untested — `ActivitiesRepositoryImplTest` still collects a single emission, so nothing pins the behaviour this PR exists for. Emitting two equal lists and asserting one collection would catch a future regression back to the identity-`equals` version, which is the whole failure mode here.

**Verdict:** correct fix to the right problem; ready.

---

Okuro3499 (approved):

dogi:
Conflict resolved — head is now `2f6ebb8`, `mergeable_state: clean`, all six checks green (build default + lite, both test shards).

Two things in the description no longer match the diff, because master moved underneath this branch:

- **The `repeatOnLifecycle` flattening is gone.** master since moved `ActivitiesFragment` onto `ActivitiesViewModel` and rewrote `computeMonthlyCounts` as a single-pass `fold`. I kept master's shape and re-applied only this branch's contribution on top, so the fragment change is now just `computeMonthlyCounts` becoming `suspend` + `withContext(dispatcherProvider.default)`.
- **`distinctUntilChanged` → `distinctByContent`.** #15932 landed the non-allocating helper in `utils/FlowExtensions.kt` while this PR sat, which is the convergence the re-review asked for. The hand-rolled `old.zip(new)` predicate is replaced by `.distinctByContent { a, b -> a.id == b.id && a.loginTime == b.loginTime }` — same semantics, no intermediate `Pair` list.

Also added the regression test the re-review asked for: three emissions where the middle one is structurally identical to the first, asserting the flow collapses them to two. That pins the actual failure mode — `OfflineActivity` has no `equals` override, so a bare `distinctUntilChanged()` would silently go inert again.

One caveat worth stating: I could not run the suite locally (no Android SDK in this container — `testDefaultDebugUnitTest` fails at configuration time), so CI above is the first execution of these tests. I did check the new imports against `.agents/skills/kotlin-importing`; the ones I added sort correctly. It flags two pre-existing misorderings that came in from master (`DispatcherProvider` after `JsonUtils` in `ActivitiesRepositoryImpl`, `StandardTestDispatcher`/`TestScope` in `ActivitiesRepositoryImplTest`) — left alone rather than widen this PR.

The `conflict` label is stale now; the drainer adds it and nothing removes it. I left the labels untouched since re-queueing is your call.

---

PR #16094, opened 2026-08-24
courses: smoother view recycler gliding (fixes #16092)

### LINKED ISSUE #16092: courses: app crashes with IllegalArgumentException (Glide load for destroyed activity) on screen rotation
https://github.com/user-attachments/assets/c7730d5c-c5e0-4565-8696-5feeb0d31030

When navigating to the **Courses** screen and rotating the device (e.g. switching between portrait and landscape), the app crashes immediately and force closes.

### Steps to Reproduce
1. Open myPlanet and log in.
2. Navigate to **Courses** (or **My Courses**).
3. Rotate the device from Portrait to Landscape (or trigger an orientation change).
4. Notice the app crashes and exits to the home screen.

### Expected Behavior
The app should handle configuration/orientation changes smoothly and recreate the Courses screen without crashing.

ragilzakaria:
https://github.com/user-attachments/assets/d6647f74-f596-48f3-bb17-e984adac2d52

fixes #16092

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved course image cleanup when course views are recycled.
  - Added safer handling when the associated activity has been destroyed.
  - Ensured cleanup works consistently in both grid and list course views.

- **Tests**
  - Added coverage for recycling grid and list views after activity destruction.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6704
-        versionName = "0.67.4"
+        versionCode = 6705
+        versionName = "0.67.5"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/ui/courses/CoursesAdapter.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/courses/CoursesAdapter.kt
@@ -1,5 +1,6 @@
 package org.ole.planet.myplanet.ui.courses
 
+import android.app.Activity
 import android.content.Context
 import android.graphics.drawable.GradientDrawable
 import android.os.Bundle
@@ -247,13 +248,13 @@ class CoursesAdapter(
 
     override fun onViewRecycled(holder: RecyclerView.ViewHolder) {
         super.onViewRecycled(holder)
-        when (holder) {
-            is GridViewHolder -> {
-                Glide.with(context).clear(holder.binding.ivCover)
-            }
-            is ListViewHolder -> {
-                Glide.with(context).clear(holder.binding.ivCover)
-            }
+        val targetView = when (holder) {
+            is GridViewHolder -> holder.binding.ivCover
+            is ListViewHolder -> holder.binding.ivCover
+            else -> null
+        }
+        targetView?.let { view ->
+            Glide.with(context.applicationContext).clear(view)
         }
     }
 
--- a/app/src/test/java/org/ole/planet/myplanet/ui/courses/CoursesAdapterTest.kt
+++ b/app/src/test/java/org/ole/planet/myplanet/ui/courses/CoursesAdapterTest.kt
@@ -2,16 +2,23 @@ package org.ole.planet.myplanet.ui.courses
 
 import android.app.Application
 import android.content.Context
+import android.widget.LinearLayout
+import androidx.appcompat.app.AppCompatActivity
 import androidx.recyclerview.widget.RecyclerView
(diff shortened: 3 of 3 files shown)


dogi:
**Review — rating: 72/100**

The core of the fix is right: `Glide.with(activity)` throws `You cannot start a load for a destroyed activity`, and switching to `Glide.with(context.applicationContext)` sidesteps that because the application `RequestManager` is never destroyed. Clearing still works from there — `RequestManager.clear(View)` wraps the view in a `ClearTarget` whose `getRequest()` reads the view's tag, so `untrackOrDelegate` finds and clears the request even though it was started by a different manager.

**But the diff ships three overlapping defenses for one problem, and one of them points the wrong way.** Once you're on `applicationContext`, the destroyed-activity check can't fire — so the `isDestroyed || isFinishing` early return isn't protecting the Glide call, it's *skipping* the clear precisely when teardown makes releasing most worthwhile. And `context as? Activity` misses the case that matters most: if `context` is a `ContextThemeWrapper` around the Activity (which is what an adapter often holds), the cast yields null and the guard is silently bypassed anyway. The two behaviours — "skip on destroy" and "never skip because the cast failed" — can't both be intended.

The `try { … } catch (e: Exception) { }` is the third layer. With `applicationContext` there's no documented throw left to catch, and an empty catch with only a comment means any future breakage here is invisible. Deleting the guard and the catch leaves the whole method as `targetView?.let { Glide.with(context.applicationContext).clear(it) }`, which is the fix, and reads as such.

**The test can't fail.** It builds a Grid holder and a List holder against a live `ApplicationProvider` context and calls `onViewRecycled` with no assertions — it passes identically against master, since the crash needs a *destroyed* Activity context. Robolectric can produce one (`Robolectric.buildActivity(...).setup().destroy()`), and driving `onViewRecycled` with that is the case worth pinning.

**Nit:** the new test reaches for `androidx.test.core.app.ApplicationProvider`, `org.ole.planet.myplanet.utils.ListViewMode` and `android.widget.LinearLayout` fully qualified inline rather than importing them.

**Verdict:** keep `applicationContext`, drop the guard and the catch, and give the test a destroyed activity.

---

dogi:
**Re-review — Rating: 85/100** (was 72)

All three findings are addressed, and by deletion rather than by adding a fourth guard — which is the right instinct. The empty `catch` is gone, the `isDestroyed` check that skipped the clear at exactly the moment it was needed is gone, and what remains is the one thing that actually fixes it: `Glide.with(context.applicationContext)`. That's correct because `Glide.with(Activity)` is what throws on a destroyed activity, while the application-scoped `RequestManager` never does — and `clear(View)` still cancels the right request, since it wraps the view in a `ClearTarget` whose `getRequest()` reads the view's tag rather than any manager-held state. Collapsing the two identical `when` branches into a `targetView` lookup is a genuine simplification on top.

**The test is now a real regression test**, which was my main objection last round. Building a destroyed `AppCompatActivity` and calling `onViewRecycled` through it would hit `IllegalArgumentException: You cannot start a load for a destroyed activity` against master, so it fails there and passes here. The import cleanup in that file (FQN'd `ArgumentMatchers` and `LinearLayout` turned into real imports) is a nice bonus.

**Nit — one thing came in that isn't used.** This PR adds `import android.app.Activity`, but `Activity` appears nowhere in `CoursesAdapter.kt`, neither on master nor in the new code. It's left over from the `isDestroyed` version that got removed. Kotlin only warns, so CI stays green.

**Nit:** the new test has no assertions — its contract is "this must not throw", which is legitimate but reads as unfinished. One comment line saying so would stop someone later "fixing" it by adding a meaningless assert.

**Verdict:** the right minimal fix, three defences reduced to one, and a test that genuinely fails on master; ready once the stray `Activity` import goes.

---

Okuro3499 (approved):

dogi (approved):

PR #16096, opened 2026-08-24
life: smoother repository layout view modelling (fixes #16089)

### LINKED ISSUE #16089: mylife page is empty
it should show list of mylife items
https://github.com/user-attachments/assets/75d091ed-d4b4-46e2-bace-faa2b85f502f

ragilzakaria:
https://github.com/user-attachments/assets/44457d8d-b9fe-4c2a-b8fa-7c6702742ecf

fixes #16089

## Summary by CodeRabbit

- **Improvements**
  - Improved the Life list layout and loading behavior.
  - Life items now consistently appear in ascending custom order.
  - Reordering items updates their order immediately and persists reliably.
  - Improved handling of unassigned and signed-in user items to prevent lists from mixing.
  - Default Life items are added when no items are available.

- **Bug Fixes**
  - Fixed visibility and ordering updates not refreshing the displayed list correctly.
  - Improved fallback handling when account information is unavailable.


dogi:
**Review — rating: 76/100**

The layout fix is the real find here and the PR body undersells it. `fragment_life.xml`'s root is a `<LinearLayout>` with **no `android:orientation`**, which defaults to *horizontal* — so the `FrameLayout` holding the RecyclerView was sized `0dp` wide with `layout_weight="1"` (filling horizontally) and `wrap_content` tall. Adding `orientation="vertical"` and switching the child to `match_parent`/`match_parent` is the correct fix for the sizing bug. Fixing the stale `tools:context=".ui.mylife.LifeFragment"` → `.ui.life.LifeFragment` while you were in there is a nice catch too.

The `"--"` fallback is also better than it looks: it isn't a magic string invented here, it's the house sentinel already used at `BaseRecyclerFragment:95`, `BaseDashboardFragment:293/348/369` and `TeamCoursesFragment:46`, all as `getUserId().ifEmpty { "--" }`. This version chains prefs → user model → `"--"`, which is strictly better and still lands where the rest of the app lands.

**The observer move introduces a race that the guard then hides.** Previously `collectWhenStarted(viewModel.myLifeList)` lived *inside* `getAdapter()`, so by construction the adapter existed before any emission could arrive — the `isObserverAttached` boolean was ugly but the ordering was safe. Now the collector runs in `onViewCreated`, and `BaseRecyclerFragment:76-81` calls `getAdapter()` from inside a coroutine launched by `super.onViewCreated(...)`. So `lifeAdapter` may still be uninitialized when the first list lands, and `if (::lifeAdapter.isInitialized)` silently drops it. Since `myLifeList` is a StateFlow, nothing re-emits within that STARTED period — the user sees an empty list until they background and return, at which point `repeatOnLifecycle` re-subscribes and replays.

In practice the adapter path is pure construction and `loadMyLifeList()` goes through `withContext(io)` and the DB, so the adapter almost always wins the race — but "almost always" is what makes this the kind of bug that shows up once on a slow device and is never reproduced. Constructing `lifeAdapter` eagerly instead of `lateinit`, or collecting after the adapter is assigned, removes both the race and the guard.

The five new `LifeViewModelTest` cases are good — they cover the seed path, the `"--"` fallback by name, and use `TestDispatcherProvider` rather than a hand-rolled `Unconfined` provider.

**Verdict:** layout and sentinel changes are right; the collector needs to be ordered after the adapter rather than guarded against it.

---

dogi:
**Re-review — Rating: 79/100** (was 76)

Both of my previous findings are fixed, and the second one properly. `android:orientation="vertical"` is on the root, and the `FrameLayout` moving from `0dp` + `layout_weight="1"` to `match_parent`/`match_parent` is the correct companion change now that the orientation is explicit. Correcting `tools:context` from the long-dead `.ui.mylife.LifeFragment` to `.ui.life.LifeFragment` is a nice catch nobody asked for.

The race is gone at the root rather than papered over. Deleting `isObserverAttached` and moving `collectWhenStarted` + `loadMyLifeList()` out of the suspend `getAdapter()` and into the lifecycle method means the observer is attached at a deterministic point on every view creation, instead of depending on when the base class happens to call `getAdapter()` and on a flag that survived view destruction. Extracting `initAdapter()` keeps `getAdapter()` honest and is idempotent, so the double call is harmless. Nulling `itemTouchHelper` in `onDestroyView` is a real fix too — it was holding the destroyed `RecyclerView`.

**What holds this back is `?: "--"`.** That's a magic sentinel introduced with no comment, and it changes which partition the Life data lives in. Previously a session with no user resolved to `userId = null`, so `getMyLifeByUserId(null)` queried the NULL rows and `seedMyLifeIfEmpty(null, …)` seeded them. Now the same session queries and seeds under the literal string `"--"` — so any device that already has null-seeded defaults will find nothing, re-seed under `"--"`, and leave the original rows orphaned in the table. The new test enshrines it as "default dash user" rather than questioning it.

I can see why it's here — a user-less session probably showed an empty Life screen, and that's squarely "list loading", so it's in scope. But it needs three things before it ships: a named constant instead of an inline `"--"`, one comment sentence saying why `null` wasn't sufficient, and a decision on whether the pre-existing null rows should be migrated or simply don't exist in practice. If the answer to the last one is "no real device has them", say so and this is done.

**Heads-up, not a problem today:** `MyLife.defaultItems(context, userId)` still matches master, so nothing conflicts right now — but #16175 (currently `ready`) changes that signature to `(userId, resolveLabel)`. Whichever lands second needs a one-line touch-up here.

**Verdict:** layout and race both properly fixed and the five new ViewModel tests are solid — everything here is ready-grade except the undocumented `"--"` sentinel, which is a small ask away from done.

---

dogi:
**Re-review — Rating: 68/100** (was 76)

The layout fix is real and correctly diagnosed: the root `LinearLayout` had no `android:orientation`, so it defaulted to horizontal, and the single `FrameLayout` child at `layout_width="0dp" layout_weight="1" layout_height="wrap_content"` got full width but content-sized height — which is why the `match_parent` RecyclerView inside never filled the fragment and `tv_message` had nothing to centre in. `orientation="vertical"` plus `match_parent`/`match_parent` is the right correction, and the stale `tools:context=".ui.mylife.LifeFragment"` → `.ui.life.LifeFragment` catches a package rename the file missed.

Moving the collection out of `getAdapter()` into `onViewCreated` is also correct, and it retires `isObserverAttached` for the right reason rather than by accident: `BaseRecyclerFragment` calls `getAdapter()` from three places (`onViewCreated`, the refresh path, and `setAdapter`), so the flag was guarding against repeat invocation. `onViewCreated` runs exactly once per view, and `Fragment.collectWhenStarted` scopes to `viewLifecycleOwner`, so the collector dies at `onDestroyView` with no double-subscription possible. Ordering is safe too — `initAdapter()` runs synchronously before the base's `lifecycleScope.launch` body resumes, so `lifeAdapter` is always initialized before either the collector lambda or the base's `getAdapter()` touches it, and `recyclerView` is already assigned in the base's `onCreateView`.

**Blocking: `itemTouchHelper = null` in `onDestroyView` kills drag-to-reorder on the second view.** The helper is only ever constructed inside `initAdapter()`'s `if (!::lifeAdapter.isInitialized)` block, but `lifeAdapter` is a `lateinit` field on the *fragment instance*, which outlives the view. `openCallFragment` reaches `DashboardElementActivity.openCallFragment` with `addToBackStack = true`, so navigating away from My Life and pressing back recreates the view on the same instance: `initAdapter()` sees the adapter already initialized, skips the block, `itemTouchHelper?.attachToRecyclerView(recyclerView)` no-ops on null, and `onStartDrag`'s `itemTouchHelper?.startDrag(it)` no-ops too. Reorder is silently dead until the process restarts. On master the helper was never nulled, so it survived alongside the adapter and re-attached fine (`attachToRecyclerView` handles swapping RecyclerViews). Either construct the helper outside the `isInitialized` guard, or keep the field and release the view reference with `itemTouchHelper?.attachToRecyclerView(null)` in `onDestroyView`.

**Nit:** the new tests set up the main dispatcher by hand with `Dispatchers.setMain`/`resetMain` rather than the repo's `MainDispatcherRule`, though they do use `TestDispatcherProvider`. Also worth noting the tests exercise `LifeViewModel`, which this PR doesn't change — the behaviour that *did* change (adapter/helper lifecycle, layout) is the untested part, and it's where the regression above sits. They're still worth having: routing the seed assertion through `seedMyLifeIfEmpty(userId, any())` keeps them independent of `MyLife.defaultItems`' signature, which #16175 is currently changing.

**Verdict:** solid layout and lifecycle work with one two-line regression to undo; fix the `itemTouchHelper` nulling and this is ready.

---

dogi:

Both points from last round are fixed, and the `itemTouchHelper` one took the better of the two options I offered. `attachToRecyclerView(null)` is the documented detach path: with a non-null `mRecyclerView` it runs `destroyCallbacks()`, clears the field, and skips `setupCallbacks()`, so the RecyclerView reference is released at `onDestroyView` without discarding the helper. Because the helper survives, the next `initAdapter()` re-attaches it to the fresh RecyclerView even though the `isInitialized` guard skips the construction block — so drag-to-reorder keeps working after a back-stack return, which is exactly what nulling the field broke. Nulling would have needed the constructor moved out of the guard as well; this way is one line and leaves the guard meaningful.

The tests now go through `MainDispatcherRule(testDispatcher)` with the hand-rolled `setMain`/`resetMain` and the `@After` gone, which is the shared infra `docs/TESTING.md` points at.

Everything I checked last time still holds: the layout diagnosis (root `LinearLayout` defaulting to horizontal, so the weighted `FrameLayout` got full width but `wrap_content` height), and moving the collection to `onViewCreated` where it runs once per view under `viewLifecycleOwner`, which is what `isObserverAttached` was really guarding against given `BaseRecyclerFragment` calls `getAdapter()` from three sites. Re-running `loadMyLifeList()` on each view creation is a redundant fetch on a back-stack return, but the StateFlow replays its current value to the new collector first, so the list is never empty in the meantime and the reload just picks up any changes.

**Verdict:** clean fix for the regression, tests aligned with repo conventions — ready.

---

Okuro3499 (changes requested):
sorting the list is breaking the list as it shows a different order from the expected one
https://github.com/user-attachments/assets/1f57a630-19c5-42cf-964f-279a5fee18dd

dogi:
## Review: fix the empty My Life page

**Rating: 86/100** — this fixes a genuine data-leak in the DAO predicate plus a layout defect, and it ships the one test that fails on `master`. The main risk is that it now overlaps six other open PRs.

### The DAO predicate was wrong in both directions

```sql
-- before
WHERE (:userId IS NULL OR userId IS NULL OR userId = :userId)
```

- `:userId` **null** → the first disjunct is true, so it matched **every row in the table**, for every user.
- `:userId` **non-null** → matched that user's rows **plus every row with a null `userId`**. Guest-owned items leaked into every signed-in user's My Life list.

```sql
-- after
WHERE (:userId IS NULL AND (userId IS NULL OR userId = '' OR userId = '--'))
   OR (:userId IS NOT NULL AND userId = :userId)
```

That's a proper partition: the guest bucket and each user's bucket are now disjoint. Applied consistently to `countByUserId`, which matters because `seedMyLifeIfEmpty` gates on that count — under the old predicate a guest's count was the whole table, so seeding never ran for a fresh user.

### The `'--'` arm is load-bearing, not invented

I checked, because a magic sentinel in SQL is worth verifying. `"--"` is an established guest marker in this codebase:

- `BaseRecyclerFragment:95` — `sharedPrefManager.getUserId().ifEmpty { "--" }`
- `BaseDashboardFragment:292, :347, :368`
- `TeamCoursesFragment:46`

`LifeFragment` extends `BaseRecyclerFragment`, so rows written for a guest genuinely can carry `userId = "--"`. Including it in the guest branch matches real data rather than a hypothetical. `normalizeUserId` collapsing `null`/`""`/`"--"` to a single `null` on the Kotlin side is the right complement — one concept, one representation.

### The layout was probably the visible symptom

`fragment_life.xml`'s root had **no `android:orientation`**, and `LinearLayout` defaults to *horizontal*. Its only child was `layout_width="0dp"` + `layout_weight="1"` + `layout_height="wrap_content"` — a combination that only makes sense in a horizontal parent, and which leaves the RecyclerView's height unconstrained. Setting `orientation="vertical"` and giving the `FrameLayout` `match_parent` on both axes is the correct fix, and the stale `tools:context=".ui.mylife.LifeFragment"` → `.ui.life.LifeFragment` is a good catch alongside it.

### Observer lifecycle

Moving `collectWhenStarted(viewModel.myLifeList)` and `loadMyLifeList()` out of the suspend `getAdapter()` and into `onViewCreated`, and deleting the `isObserverAttached` latch, is the right shape. The latch was reset in `onDestroyView` but the collection was set up inside a suspend function the base class calls on its own schedule — a combination that can easily leave a recreated view with no collector. `itemTouchHelper?.attachToRecyclerView(null)` on teardown is a good addition too.

### Tests

`getByUserId_doesNotMixDifferentUsersOrNullUsers` is the one that earns its place — on `master`, `getByUserId("user1")` returns two rows (the user's and the guest's) and the assertion of `1` fails. That's a real regression guard for the leak, driven against real Room rather than a mock. The `LifeViewModelTest` additions cover the userId resolution fallbacks, including the empty-prefs → `UserRepository` → null chain.

### Nits

- **Three of the four added `.sortedBy { it.weight }` are redundant.** `getByUserId` already ends `ORDER BY weight ASC`, and `distinctBy` preserves encounter order — so the sort in `getMyLifeByUserId`, and the two in `getMyLifeForDashboard` that follow it, re-sort an already-sorted list. Only the one applied to the SharedPreferences-cached list is doing work (JSON order isn't guaranteed). Worth dropping the other three so the DAO stays the single authority on ordering.
- **`onItemMoveFinished` rebuilds `MyLife` by enumerating six fields.** Any field not listed is silently dropped from the objects handed to `reorderCallback` and `submitList`. It's safe today — `updateMyLifeListOrder` only reads `_id`, and the adapter only renders `title`/`imageId`/`isVisible` — but it's a copy that will quietly lose data the next time a column is added. A `copy(weight = index)` on a data class, or reassigning `weight` on the existing instances, avoids that.

### Coordination — this is the contended one

Six open PRs touch the same files, and this is the deepest change of them:

| File | Also touched by |
|---|---|
| `MyLifeDao` | #16501 (`change`) — adds `getVisibleByUserId` with the **old** predicate copied |
| `LifeRepositoryImpl` | #16501, #16578 (`ready`), #16437, #16368 (`merge`) |
| `LifeAdapter` | #16508 (`change`), #16572 (`ready`) |
| `LifeViewModel` | #16590 (still in `review`) |

Two are worth calling out specifically:

- **#16501** adds `getVisibleByUserId` by copying the *broken* predicate. If it merges after this, the leak returns through the new query. Whichever lands second must adopt this PR's predicate.
- **#16508** removes the `.toList()` in `onItemMoveFinished`; this PR rewrites that same block to build fresh instances, which resolves the aliasing concern I raised there more cleanly. #16508 becomes unnecessary if this lands.

Given the overlap, I'd sequence this one first and rebase the others onto it.

**Verdict:** mergeable.

---

Okuro3499 (approved):

dogi (approved):

PR #16100, opened 2026-08-24
resources: less apply filter button is more (fixes #16091)

### LINKED ISSUE #16091: filter has duplicate buttons
resources filter button has a X and close button which have a similar function remove the close button
Image

ragilzakaria:
Root Cause & Fix

- Duplicate Close Button: In the Resources Category Filter dialog (fragment_library_filter.xml), there was both a top-right X close icon (iv_close) and a bottom "Close" button (btn_apply_filter), both performing dialog dismissal (dismiss()).
- Fix: Removed the redundant bottom btn_apply_filter button from  fragment_library_filter.xml and its click listener from ResourcesFilterFragment.kt, keeping the top-right X close button

fixes #16091

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6669
-        versionName = "0.66.69"
+        versionCode = 6670
+        versionName = "0.66.70"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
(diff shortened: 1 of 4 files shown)


dogi:
**Review — rating: 84/100**

Removing a second dismiss affordance labelled `@string/close` from a dialog that already has `iv_close` is straightforwardly right — one control, one meaning. The `setOnClickListener { dismiss() }` goes with it, so no orphaned binding reference is left behind, and the XML indentation fix on the two closing tags is a genuine tidy rather than churn.

The test is better-aimed than it first looks. You can't assert `btn_apply_filter` is gone — the id no longer exists, so the reference wouldn't compile — which means the useful assertion is the one it makes: the dialog still inflates after the removal and still has a working close affordance. That's exactly the regression this change could cause.

**Nit:** `org.junit.Assert.assertNull` is imported but never used — the body only calls `assertNotNull` twice.

**Nit:** `@Config(sdk = [32], application = Application::class)` pins this new class to its own Robolectric SDK, which spawns a sandbox and classloader that no other test in the suite shares. There's no stated reason for 32 specifically, and if the suite default works, dropping the `sdk` pin lets this class reuse the common sandbox. Worth mentioning because #16105 in this same batch is removing bespoke `sdk` values from `DownloadServiceTest` for exactly that cost — it'd be a shame to add one back in the same afternoon.

**Verdict:** ready.

---

Okuro3499 (approved):

dogi (approved):

PR #16101, opened 2026-08-24
all: consistent status bar (fixes #16099)

### LINKED ISSUE #16099: inconsistent status bar
update top of the app to be consistent all through the app some pages are white others are 
Image
Image

Okuro3499:
fixes #16099
Screenshot_20260824_162951
Screenshot_20260824_163000

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved edge-to-edge display across the app for more consistent status-bar and navigation-bar icon contrast, including dark mode.
  * Updated login screen backgrounds and scrolling layout for more consistent visual rendering.
  * Improved dashboard and toolbar appearance, including sizing and navigation-bar colors.
* **Chores**
  * Adjusted the application version for this release.


dogi:
**Review — rating: 78/100**

Centralises status-bar appearance: adds `setupEdgeToEdgeWithStatusBarToolbar` so a custom toolbar absorbs the status-bar inset into its own top padding, forces white icons in night mode, moves 15 of 16 call sites to `lightStatusBar = false`, and repaints the dashboard bars `?attr/colorPrimary` → `@color/mainColor`.

I checked the parts the diff can't settle on its own, and the execution is careful:

- **Idempotency is right.** `toolbarBasePadding*` is captured *outside* the listener, so repeated inset dispatches (rotation, IME) don't accumulate padding — the classic bug in this pattern, avoided.
- **The `app_bar_bell.xml` height change is load-bearing.** Switching `layout_height` from a fixed `?android:attr/actionBarSize` to `wrap_content` + `minHeight` is what lets the added top padding grow the bar instead of cropping its content. Easy to miss; it's here.
- **The padding target is the correct view.** `my_toolbar` is the first child of the dashboard's vertical `LinearLayout`, but `UITheme()` sets it `View.GONE` (DashboardActivity.kt:835) and `bellToolbar` VISIBLE, so the bell toolbar really is the topmost visible view — and the root's top padding is deliberately 0 so the toolbar spans the inset itself.
- **The night-mode guard is not dead code.** `OnboardingActivity.kt:66` is the one call site left at the `lightStatusBar = true` default, and its root is `@color/secondary_bg` (white in light, `colorPrimaryDark` in night) — so it genuinely needs dark icons in light mode, and `!isNightMode` is what keeps them white in dark. Correctly scoped rather than blanket-applied.
- **The login layout leaks nothing.** Dropping `secondary_bg` off the `NestedScrollView` is safe because both columns are `match_parent` height with opaque backgrounds (`@color/light_dark` weight 6, `@color/mainColor` weight 4); the only exposed root colour is the intended status-bar strip.
- **Colour semantics check out.** Light mode `mainColor` == `colorPrimary` == `md_blue_700` (no visual change); night mode `colorPrimary` is `#18191a` while `mainColor` stays `md_blue_700`, so the dashboard bar deliberately stays blue in dark mode — which is exactly what makes white icons correct everywhere.
- `setPadding`-overwrite and returning `CONSUMED` match the existing `setupEdgeToEdge` / `setupEdgeToEdgeWithKeyboard` in the same file, so those are house convention, not new regressions.
- CI: build (default + lite) and both test shards green.

What holds it under the line: `EdgeToEdgeUtilsTest` was touched only to mock `activity.resources.configuration` so the existing tests keep compiling — no assertion covers either piece of new behaviour. Two cheap ones would close it, in a util that already has a test file:

1. `lightStatusBar = true` + `UI_MODE_NIGHT_YES` ⇒ `isAppearanceLightStatusBars` is `false` (the OnboardingActivity path, the only reason the branch exists).
2. `setupEdgeToEdgeWithStatusBarToolbar` sets toolbar top padding to base + inset, and dispatching insets **twice** leaves it unchanged — that pins the idempotency contract above, which is the thing a future refactor is most likely to break by moving the capture inside the lambda.

**Nit:** the KDoc describes the third parameter as a Toolbar "standing in for the native ActionBar", but LoginActivity passes `binding.scrollContent`, a `NestedScrollView`. The parameter is really "the view that absorbs the status-bar inset" — worth renaming or rewording so the login call site doesn't read as a misuse.

**Verdict:** needs another round — correctness is verified and the approach is sound, but the new branch and new function ship untested. Add the two assertions and this is a straightforward `ready`.

---

J-S-webskas (changes requested):
In dark mode, the status bar crops the logo, icons, and the three dots on the top bar. Especially after the screen rotates.


https://github.com/user-attachments/assets/22bca676-ae5f-4554-94cf-ef98e4438568

dogi:
## Review — rating 40/100 → `change`

The two smallest changes here are the clearly right ones: adding `ViewCompat.requestApplyInsets(rootView)` after installing each listener fixes the real case where insets were already dispatched before the listener existed, and the new `EdgeToEdgeUtilsTest` case is a genuine unit test of an insets listener, captured via `slot<OnApplyWindowInsetsListener>` and invoked directly — a nice pattern for code that's normally untestable off-device. CI green on all five checks.

But the new helper has a hard-coded assumption that one of its two call sites violates, and I think it breaks the login screen.

### `LoginActivity` passes a `NestedScrollView` where the helper expects a `Toolbar`

```kotlin
EdgeToEdgeUtils.setupEdgeToEdgeWithStatusBarToolbar(
    this,
    binding.root,
    binding.scrollContent,   // androidx.core.widget.NestedScrollView
    ...
)
```

The helper's own KDoc says "activities whose top bar is a custom Toolbar", and its height logic depends on that:

```kotlin
val toolbarBaseHeight = toolbar.minimumHeight
...
val desiredHeight = toolbarBaseHeight + insets.top + toolbarBasePaddingBottom
toolbar.layoutParams?.let { params ->
    if (params.height != desiredHeight) {
        params.height = desiredHeight
        toolbar.layoutParams = params
    }
}
```

For `app_bar_bell` that works out, because this PR gives it `android:minHeight="?android:attr/actionBarSize"`. For `scrollContent` it does not. That view has `layout_height="match_parent"` and no `android:minHeight`, so `View.getMinimumHeight()` returns **0** (it reads `mMinHeight` only — background intrinsic height goes through `getSuggestedMinimumHeight()`, which isn't used here). That makes `desiredHeight` equal to `insets.top` alone, roughly the status-bar height, and since `MATCH_PARENT` is `-1` the guard always fires:

**the login form's scroll container gets its height pinned to about one status bar.** In `activity_login.xml` it's constrained top-and-bottom to the parent, so it would render as a centred sliver.

I've traced this from the diff rather than from a device, so if there's something that stops it — an override later in `onCreate`, or `scrollContent` not being the view I think it is — I'd want to see it, because the rest of the PR is fine and this is the only thing standing in its way. As written the reasoning looks airtight to me, and the new test can't catch it: it passes a mock with `minimumHeight = 140`, i.e. the Toolbar case.

Worth noting the height pinning may not be needed even for the real Toolbar. `Toolbar.onMeasure` sizes a `wrap_content` toolbar to `max(content + paddingTop + paddingBottom, suggestedMinimumHeight)`, so the `?android:attr/actionBarSize` minHeight this PR adds should already let it grow to absorb `insets.top` as padding. If that holds, dropping the `layoutParams` block entirely fixes the login screen and simplifies the helper.

### The dark app bar turns blue, and that isn't mentioned

Two background changes swap a theme attribute for a fixed colour:

- `app_bar_bell.xml`: `?attr/colorPrimary` → `@color/mainColor`
- `activity_dashboard.xml` `top_bar_navigation`: `@color/colorPrimary` → `@color/mainColor`

`mainColor` is `md_blue_700` in **both** `values/colors.xml` and `values-night/colors.xml`, whereas `?attr/colorPrimary` resolves through `values-night/styles.xml` to `@color/colorPrimary` = `#18191a`. So in dark mode the dashboard toolbar and top navigation go from near-black to blue.

That may well be the intent — you're forcing white status-bar icons everywhere, and blue carries them better than some backgrounds — but it's a dark-theme visual change riding along in an insets PR, and it's the kind of thing that gets noticed after release rather than in review. Worth stating in the description either way.

### The night-mode guard serves exactly one call site

```kotlin
controller.isAppearanceLightStatusBars = !isNightMode && lightStatusBar
```

Fifteen of the sixteen `setupEdgeToEdge*` call sites in the app now pass `lightStatusBar = false` explicitly, which short-circuits this regardless of `isNightMode`. The only one left on the default is `OnboardingActivity.kt:66`, so the whole `Configuration.UI_MODE_NIGHT_MASK` read exists to serve that single screen. Either give Onboarding the same explicit `false` and drop the branch, or keep the night check and stop passing `false` at fifteen call sites — right now it's both mechanisms doing the same job, and the next person to add an activity gets whichever one they happen to notice.

Fix the login-screen height and I'd look at this again; the rest is close.

---

J-S-webskas (changes requested):
https://github.com/user-attachments/assets/7dbd5272-967a-4ab2-af5e-c1a1643358d5

I ran the app with main and the login shows. But not with this one.

PR #16103, opened 2026-08-24
life: smoother personals repository querying (fixes #16148)

### LINKED ISSUE #16148: fix(personals): collapse double write in updatePersonalResource
Tracking issue for PR #16103.

Collapsed the double write in `PersonalsRepositoryImpl.updatePersonalResource` to prevent non-idempotent updaters from running twice. Tests have been updated.

---
*PR created automatically by Jules for task [9705152882305891709](https://jules.google.com/task/9705152882305891709) started by @dogi*

dogi:
Collapsed the double write in `PersonalsRepositoryImpl.updatePersonalResource` to prevent non-idempotent updaters from running twice. Tests have been updated.

---
*PR created automatically by Jules for task [9705152882305891709](https://jules.google.com/task/9705152882305891709) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6670
-        versionName = "0.66.70"
+        versionCode = 6671
+        versionName = "0.66.71"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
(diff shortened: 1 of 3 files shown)


dogi:
**Review — rating: 90/100**

Correct collapse, and the `?:` is the right operator rather than two independent `let` blocks. `Personal` has `@PrimaryKey var id` and a separate nullable `_id`, so for the usual row where `_id == id` both lookups returned the *same* record — the old code ran the updater twice on one object and issued two `update()` calls for a single logical edit. The only caller is `PersonalsFragment:125`, whose updater sets `description` and `title`, so this was a wasted write rather than corruption; a non-idempotent updater (a counter, an append) would have been silently doubled.

The behaviour difference worth naming explicitly: if a row existed with `_id == x` *and* a different row with `id == x`, master updated both and this updates only the `_id` match. That needs one string to be simultaneously a CouchDB document id and another record's local primary key, so it isn't reachable in practice — and "find the record by either key, update it once" is clearly what the function meant.

The tests are the part I'd point at: splitting into an `_id`-hit case and an `id`-fallback case is the right shape, and `coVerify(exactly = 0) { personalDao.findById(any()) }` in the first is the assertion that actually pins the short-circuit — without it a rewrite back to two lookups would still pass.

**Verdict:** ready.

---

Okuro3499 (approved):

PR #16106, opened 2026-08-24
all: smoother crash log store handling (fixes #16150)

### LINKED ISSUE #16150: Fix CrashLogStore cap counting malformed files
Tracking issue for PR #16106.

CrashLogStore `save` previously aborted if there were 20 or more items in the log directory, regardless of whether they were valid log files or not. This meant stray/malformed files could silently suppress real crash reports.

This extracts `isValidLogFile` to share the same file filtering requirements between `save` and `loadPendingLogs`, ignoring malformed logs during cap enforcement. Added test case for verification.

---
*PR created automatically by Jules for task [7081659680194256275](https://jules.google.com/task/7081659680194256275) started by @dogi*

dogi:
CrashLogStore `save` previously aborted if there were 20 or more items in the log directory, regardless of whether they were valid log files or not. This meant stray/malformed files could silently suppress real crash reports.

This extracts `isValidLogFile` to share the same file filtering requirements between `save` and `loadPendingLogs`, ignoring malformed logs during cap enforcement. Added test case for verification.

---
*PR created automatically by Jules for task [7081659680194256275](https://jules.google.com/task/7081659680194256275) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6674
-        versionName = "0.66.74"
+        versionCode = 6675
+        versionName = "0.66.75"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/utils/CrashLogStore.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/utils/CrashLogStore.kt
@@ -18,11 +18,19 @@ object CrashLogStore {
 
     private fun dir(context: Context): File = File(context.filesDir, DIR_NAME)
 
+    private fun isValidLogFile(file: File): Boolean {
+        if (!file.isFile || !file.name.endsWith(FILE_EXTENSION)) return false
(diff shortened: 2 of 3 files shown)


dogi:
**Review — rating: 92/100**

Real bug, cleanly fixed. `save()` counted `logDir.listFiles()?.size` — *every* entry in the directory — against `MAX_PENDING_FILES`, while `loadPendingLogs()` only ever returned entries matching the `<epoch>_<type>.log` shape. So a directory holding junk could refuse to persist genuine crash reports while reporting fewer than 20 pending, and the thing being dropped is a crash log, which by definition nobody is watching at the time.

I checked the extraction preserves `loadPendingLogs` exactly: master filtered `isFile && name.endsWith(FILE_EXTENSION)` and then rejected `separator <= 0` and a non-numeric prefix inside the `mapNotNull`; `isValidLogFile` is those same four conditions, so removing the two inline `return@mapNotNull null` guards is a pure move, not a loosening. Both call sites now share one definition of "is this one of ours", which is what made them disagree in the first place.

The test earns its place — 19 valid logs plus `not-a-log.txt` and `noseparator.log`, then asserting the save succeeds. Against master that's 21 entries, the cap trips, `save` returns null and `assertNotNull` fails. That's a test that would have caught this, rather than one that merely describes the new code.

**Verdict:** ready.

---

Okuro3499 (approved):

PR #16107, opened 2026-08-24
courses: smoother submissions repository exporting (fixes #16151)

### LINKED ISSUE #16151: refactor: drop redundant identity map in PDF exporter
Tracking issue for PR #16107.

Removed the identity map operation on the question list in both the single-PDF and multi-PDF export paths. `questionDao.getByExamId` already returns `List<ExamQuestion>`, so the map operation merely allocated a duplicate list unnecessarily. Tests have been successfully verified to confirm no logic was altered.

---
*PR created automatically by Jules for task [1356566584618091516](https://jules.google.com/task/1356566584618091516) started by @dogi*

dogi:
Removed the identity map operation on the question list in both the single-PDF and multi-PDF export paths. `questionDao.getByExamId` already returns `List<ExamQuestion>`, so the map operation merely allocated a duplicate list unnecessarily. Tests have been successfully verified to confirm no logic was altered.

---
*PR created automatically by Jules for task [1356566584618091516](https://jules.google.com/task/1356566584618091516) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6672
-        versionName = "0.66.72"
+        versionCode = 6673
+        versionName = "0.66.73"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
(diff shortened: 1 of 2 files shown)


dogi:
**Review — rating: 88/100**

`.map { question -> question }` is the identity function — it allocates a second `ArrayList` and copies every element to produce a list equal to the one it was given. Dropping it in both places changes nothing observable: `questionDao.getByExamId(it)` already returns `List<…>`, the result is bound to a `val` and only read, and `examId?.let { … }.orEmpty()` still resolves to the same type, so the `orEmpty()` on the nullable `let` result is unaffected.

Worth noting these were on the PDF export paths, which run over every question in an exam for every submission in the batch export — so the copies were per-exam-per-export rather than one-off.

Same shape as the `.map { it }` that came out of `getReportsFlow` in #15906, so this is the second sighting; a grep for `map {` bodies that return their own parameter would probably turn up the rest in one pass if there are more.

No test needed — there's no behaviour to pin, and the compiler proves the types line up.

**Verdict:** ready.

---

Okuro3499 (approved):

PR #16109, opened 2026-08-24
sync: smoother retry repository dao querying (fixes #16153)

### LINKED ISSUE #16153: Refactor RetryRepositoryImpl to use targeted @Query updates
Tracking issue for PR #16109.

Replaced read-modify-write data access patterns in RetryRepositoryImpl with direct, targeted `UPDATE` SQL queries for the `markInProgress` and `markCompleted` operations, avoiding extra round trips and full-row overwrites. Updated associated test cases to reflect the new targeted mock dependencies.

---
*PR created automatically by Jules for task [7840274181939526463](https://jules.google.com/task/7840274181939526463) started by @dogi*

dogi:
Replaced read-modify-write data access patterns in RetryRepositoryImpl with direct, targeted `UPDATE` SQL queries for the `markInProgress` and `markCompleted` operations, avoiding extra round trips and full-row overwrites. Updated associated test cases to reflect the new targeted mock dependencies.

---
*PR created automatically by Jules for task [7840274181939526463](https://jules.google.com/task/7840274181939526463) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6675
-        versionName = "0.66.75"
+        versionCode = 6676
+        versionName = "0.66.76"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/data/room/dao/RetryDao.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/data/room/dao/RetryDao.kt
@@ -48,4 +48,10 @@ interface RetryDao {
 
     @Query("UPDATE retry_operation SET status = 'pending', nextRetryTime = :retryTime WHERE status = 'in_progress'")
     suspend fun recoverStuck(retryTime: Long)
+
+    @Query("UPDATE retry_operation SET status = 'in_progress' WHERE id = :id")
+    suspend fun markInProgress(id: String): Int
+
+    @Query("UPDATE retry_operation SET status = 'completed', lastAttemptTime = :timestamp WHERE id = :id")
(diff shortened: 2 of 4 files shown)


dogi:
**Review — rating: 84/100**

Right change. `findById` → mutate → `update(entity)` is a read-modify-write across two statements that rewrites every column to change one; `UPDATE retry_operation SET status = … WHERE id = :id` is a single atomic statement, and it closes the window where a concurrent writer's changes get clobbered by the stale entity being written back. The not-found case behaves the same — master's `?.let` skipped the write, the query matches zero rows — so nothing regresses there.

The `Int` row count the DAO now returns is discarded by both repository methods, which keeps the `Unit` signature master had. Worth considering surfacing it: `markCompleted` on an id that no longer exists is currently indistinguishable from success, and the count is already plumbed to tell you.

**The status literals are the thing to watch.** `'in_progress'` and `'completed'` are hardcoded in the SQL while `RetryOperation` defines `STATUS_IN_PROGRESS = "in_progress"` and `STATUS_COMPLETED = "completed"` — they agree today, and Room can't interpolate Kotlin constants so there's no way to reference them directly. `recoverStuck` already had the same coupling, so this follows the file's convention rather than introducing it, but it now spans three queries. A test asserting the four `STATUS_*` constants equal their literal strings is a couple of lines and would fail loudly if someone renames one.

**Nit:** `markInProgress does nothing when operation not found` and `markCompleted does nothing when operation not found` no longer assert what their names say — after the rewrite both stub the DAO to return `0` and then `coVerify` the same call the passing test already verifies, so neither can distinguish found from not-found. That's inherent to moving the logic into SQL with a mocked DAO: the assertion has to become a `room-testing` in-memory one that inserts a row, calls the DAO, and reads the status back. Room's compile-time query validation covers the column and table names, so the literals above are the only part left unverified.

**Verdict:** ready.

---

Okuro3499 (approved):

PR #16110, opened 2026-08-24
sync: smoother login user managing (fixes #16154)

### LINKED ISSUE #16154: Fix substring matching in LoginSyncManager role check
Tracking issue for PR #16110.

Fixed a bug where `LoginSyncManager.isManager` was substring matching the entire stringified `roles` array against "manager". This incorrectly authorized roles containing "manager" as a substring (e.g., "managerial") and had locale-specific issues (e.g., Turkish dotted-I). The check now properly iterates through the elements of the `roles` array and performs an exact string comparison (ignoring case) only on primitive elements. Added unit tests for these edge cases.

---
*PR created automatically by Jules for task [10990895988571039204](https://jules.google.com/task/10990895988571039204) started by @dogi*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved manager-role recognition with case-insensitive exact matching.
  * Prevented similarly named or unrelated roles from being treated as manager access.
  * Preserved manager access for administrators.
  * Added handling for missing or empty role information during login synchronization.

* **Tests**
  * Added coverage for manager, administrator, non-manager, and missing-role login scenarios.

dogi:
Fixed a bug where `LoginSyncManager.isManager` was substring matching the entire stringified `roles` array against "manager". This incorrectly authorized roles containing "manager" as a substring (e.g., "managerial") and had locale-specific issues (e.g., Turkish dotted-I). The check now properly iterates through the elements of the `roles` array and performs an exact string comparison (ignoring case) only on primitive elements. Added unit tests for these edge cases.

---
*PR created automatically by Jules for task [10990895988571039204](https://jules.google.com/task/10990895988571039204) started by @dogi*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved manager-role recognition with case-insensitive exact matching.
  * Prevented similarly named or unrelated roles from being treated as manager access.
  * Preserved manager access for administrators.
  * Added handling for missing or empty role information during login synchronization.

* **Tests**
  * Added coverage for manager, administrator, non-manager, and missing-role login scenarios.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6706
-        versionName = "0.67.6"
+        versionCode = 6707
+        versionName = "0.67.7"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/model/UserEntity.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/model/UserEntity.kt
@@ -167,12 +167,12 @@ open class UserEntity(
     }
 
     fun isManager(): Boolean {
-        val hasManagerRole = rolesList?.any { it.contains("manager", ignoreCase = true) } == true
+        val hasManagerRole = rolesList?.any { it.equals("manager", ignoreCase = true) } == true
         return hasManagerRole || userAdmin ?: false
     }
 
     fun isLeader(): Boolean {
-        return rolesList?.any { it.contains("leader", ignoreCase = true) } == true
+        return rolesList?.any { it.equals("leader", ignoreCase = true) } == true
     }
 
     fun isGuest(): Boolean {
--- a/app/src/main/java/org/ole/planet/myplanet/services/sync/LoginSyncManager.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/services/sync/LoginSyncManager.kt
@@ -176,7 +176,12 @@ class LoginSyncManager @Inject constructor(
 
     private fun isManager(jsonDoc: JsonObject?): Boolean {
         val roles = jsonDoc?.get("roles")?.asJsonArray
-        val isManager = roles.toString().lowercase(Locale.getDefault()).contains("manager")
+        var isManager = false
+        roles?.forEach { role ->
+            if (role.isJsonPrimitive && role.asString.equals("manager", ignoreCase = true)) {
+                isManager = true
+            }
+        }
         return jsonDoc?.get("isUserAdmin")?.asBoolean == true || isManager
     }
 }
--- a/app/src/main/java/org/ole/planet/myplanet/services/sync/LoginSyncManager.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/services/sync/LoginSyncManager.kt
@@ -6,7 +6,6 @@ import dagger.hilt.android.qualifiers.ApplicationContext
 import java.net.ConnectException
 import java.net.SocketTimeoutException
 import java.net.UnknownHostException
-import java.util.Locale
 import javax.inject.Inject
 import javax.inject.Singleton
 import kotlinx.coroutines.CoroutineScope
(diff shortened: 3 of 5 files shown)


dogi:
**Review — rating: 88/100**

This is a bigger fix than "substring matching" suggests, because of what `isManager` gates. `roles.toString().lowercase(...).contains("manager")` stringified the whole JSON array and searched it, so any role merely *containing* the word matched — and a match short-circuits the verification branch, meaning a user whose role was `managerial` (or `submanager`, or `not-manager`) got `onSyncComplete()` instead of "user verification in progress". Exact matching closes that. The `managerial` test is the one that fails against master, which is exactly the test I'd want to see for this.

The rest of the new code is right for the right reasons: `role.isJsonPrimitive` before `asString` avoids the `UnsupportedOperationException` a non-primitive element would throw, `roles?.forEach` handles the missing-property case, and `String.equals(ignoreCase = true)` is locale-independent — which the old `lowercase(Locale.getDefault())` was not. Four tests covering capital `Manager`, the substring false positive, admin-without-role, and absent `roles` is the full matrix.

**`UserEntity.kt:173` still has the bug this PR fixes.** `val hasManagerRole = rolesList?.any { it.contains("manager", ignoreCase = true) } == true` is the same substring test, and line 178 does it for `leader`. After this merges, the app has two different answers to "does this user have the manager role" — one exact, one substring — and they'll disagree for exactly the inputs this PR was written about. Worth deciding which is intended and making both match; if `contains` is deliberate there (to catch compound roles), then this PR's exact match may be too strict and the reasoning belongs in a comment. Note #16114 in this same batch edits lines 183-184 of that file without touching 173.

**Nit:** `import java.util.Locale` is now unused — line 179 was its only consumer.

**Nit:** `var isManager` + `forEach` with no early exit reads awkwardly next to the `any {}` the codebase already uses for exactly this in `UserEntity`. `val isManager = roles?.any { it.isJsonPrimitive && it.asString.equals("manager", ignoreCase = true) } == true` is one line and short-circuits.

**Verdict:** ready — real fix in an auth path, pinned by a test that would have caught it.

---

Okuro3499 (dismissed):
```
- LoginSyncManager.kt:9 — import java.util.Locale is now unused (the old .lowercase(Locale.getDefault()) call was removed but the import wasn't). Remove it.                       
  - LoginSyncManager.kt vs UserEntity.kt:170 — Given the domain doc's explicit instruction to keep these two role checks in sync when changing manager-detection logic, either (a)   
    also change UserEntity.isManager()/isLeader() to exact-match in this PR for consistency, or (b) confirm in the PR description that production role strings are always exact      
    single tokens (as "guest" and "leader"'s sibling checks imply) so the divergence is inert, and file a follow-up to fix UserEntity.kt. Leaving it silently inconsistent risks a   
    real behavioral split between login-time and runtime manager checks.
```

dogi:
**Re-review — Rating: 91/100** (was 88)

This now closes the split-brain I flagged last round: `UserEntity.isManager()` and `isLeader()` move to `equals` alongside `LoginSyncManager`, so the app no longer has two disagreeing answers to "does this user hold the manager role". The unused `java.util.Locale` import is gone, and the `managerial` test pins the actual regression.

I checked the two flipped assertions against the domain model rather than taking them at face value, since a test changed from `assertTrue` to `assertFalse` is usually the tell for a test bent to fit the code. `docs/DOMAIN_MODEL.md:44` defines a manager as a user whose `rolesList` **contains `"manager"`**, and the summary table at `:314` says the same; the role vocabulary is exact strings, which `MyHealthFragment:230`'s `rolesList?.contains("health")` (list containment, not substring) independently corroborates. So `project_manager` was never meant to match, and the flips are correct.

Worth stating plainly for whoever merges this: `isManager()` gates 11 real call sites — feedback scope in `FeedbackRepositoryImpl:73`, upload behaviour at `UploadManager:93` and `:381`, the `isAdmin` extra at `DashboardActivity:690`, enterprise management, notification creation. This is a permission-narrowing change, so any deployment whose CouchDB emits a compound role such as `project_manager` will see those users lose manager rights on upgrade. `isLeader()` is safe by contrast — it has **zero** call sites, which `DOMAIN_MODEL.md:56` also records, so that flip cannot affect production.

**Verdict:** the cross-file inconsistency is resolved and the new semantics are the documented ones; ready, with the compound-role caveat worth a glance at real server data before it ships.

---

Okuro3499 (approved):

PR #16112, opened 2026-08-24
actions: smoother workflow test caching (fixes #16156)

### LINKED ISSUE #16156: build(ci): optimize build cache by excluding test reports
Tracking issue for PR #16112.

Optimized the CI test workflow by preventing stale test reports from being cached.

- Updated the `actions/cache` paths in `.github/workflows/test.yml` to exclude `app/build/test-results/**` and `app/build/reports/**`.
- Removed the manual step that was previously used to delete these redundant directories after cache restoration.

---
*PR created automatically by Jules for task [8971901367436479351](https://jules.google.com/task/8971901367436479351) started by @dogi*

dogi:
Optimized the CI test workflow by preventing stale test reports from being cached.

- Updated the `actions/cache` paths in `.github/workflows/test.yml` to exclude `app/build/test-results/**` and `app/build/reports/**`.
- Removed the manual step that was previously used to delete these redundant directories after cache restoration.

---
*PR created automatically by Jules for task [8971901367436479351](https://jules.google.com/task/8971901367436479351) started by @dogi*

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -48,6 +48,10 @@ jobs:
             build-dir-test-${{ matrix.build }}-shard${{ matrix.shard }}-
             build-dir-test-${{ matrix.build }}-
 
+      - name: drop stale test results from restored build dir
+        shell: bash
+        run: rm -rf app/build/test-results app/build/reports/tests
+
       - name: run unit tests
         shell: bash
         env:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -39,16 +39,14 @@ jobs:
         with:
           path: |
             app/build
+            !app/build/test-results/**
+            !app/build/reports/**
             .gradle/*
             !.gradle/configuration-cache
           key: build-dir-test-${{ matrix.build }}-shard${{ matrix.shard }}-${{ github.sha }}
           restore-keys: |
             build-dir-test-${{ matrix.build }}-shard${{ matrix.shard }}-
             build-dir-test-${{ matrix.build }}-
-
-      - name: drop stale test results from restored build dir
-        shell: bash
-        run: rm -rf app/build/test-results app/build/reports/tests
 
       - name: run unit tests
         shell: bash
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6707
-        versionName = "0.67.7"
+        versionCode = 6708
+        versionName = "0.67.8"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }


dogi:
**Review — rating: 82/100**

The reasoning is sound and the mechanism is proven in this very file: `actions/cache` resolves its `path` through `@actions/glob`, which supports negation, and the existing `!.gradle/configuration-cache` entry shows it working here already. Ordering is right too — negations have to follow the include they subtract from, and `app/build` precedes both exclusions. Test results and reports are pure outputs regenerated every run, so there's nothing to lose by not caching them, and for 174 test classes that's a few MB off every save and restore.

Removing the `drop stale test results from restored build dir` step follows logically: `rm -rf app/build/test-results app/build/reports/tests` only existed to undo the caching this PR stops doing.

**The sequencing is what I'd change.** Both halves land together, so if the exclusion silently doesn't apply, the `rm -rf` that was compensating is gone at the same moment — and the failure is quiet: stale results restored from cache, then uploaded as this run's artifact. Green CI doesn't rule that out, since a run passes either way. Landing the exclusion first, confirming from a run's cache size (or a `ls app/build/test-results` after restore) that it took, and deleting the `rm -rf` in a follow-up costs one extra PR and removes the guesswork. If you'd rather keep it as one change, a line in the PR body naming the observed cache-size drop would do the same job.

**Heads-up:** #16111 in this batch also edits `test.yml` — it adds `timeout-minutes` and narrows the upload `path` to just `app/build/test-results/`. Different hunks, so probably a clean merge, but that PR changes what gets *uploaded* while this changes what gets *restored*, and the stale-results failure mode above is exactly where they meet. Whichever lands second is worth a re-read of the combined file.

**Verdict:** ready.

---

Okuro3499 (dismissed):
```
Problem: actions/cache's exclusion patterns (!path) only apply when saving the archive — they determine what gets written into the tarball, not what gets extracted on restore. A
    tarball already saved before this change (under build-dir-test-default-shard1-<old-sha> etc.) still contains test-results/reports. Because the exact cache key always misses     
    (it's keyed on github.sha), essentially every run — including the very first ones after this merges — falls through restore-keys to the most recent matching prefix cache. Until 
    a fresh cache is saved without those paths for each build×shard combination (i.e., after this PR has run once per shard), every restore will still extract the old, stale        
    test-results/reports directories, unfiltered — exactly the contamination the removed rm -rf step existed to prevent (it was added in 32cab5381 specifically for this). The new   
    mechanism and the old one solve overlapping-but-different problems (future cache bloat vs. current-run staleness); dropping the old one leaves a real gap for at least one CI    
    generation per shard/flavor.
```

dogi:
**Re-review — Rating: 87/100** (was 82)

Narrowed to exactly the half I suggested landing first — the cache exclusion without the `rm -rf` removal, so the safety net stays in place until the exclusion is proven in a real run. That's the right sequencing for a change whose failure mode is silent (stale results restored, then published as this run's).

The mechanics check out. `actions/cache` resolves `path` through `@actions/glob` and honours `!` negation, and negations must follow the include they subtract from — both `!app/build/test-results/**` and `!app/build/reports/**` sit after `app/build`, matching the `.gradle/*` / `!.gradle/configuration-cache` precedent two lines down.

The thing that makes this safe rather than a regression is worth spelling out: dropping `test-results` from the Actions cache does not force the test task to re-run from scratch, because the fast path in this repo comes from the Gradle build cache inside the cached Gradle home, not from `app/build`. A `FROM-CACHE` test task re-materialises its own outputs, so the no-change run stays cheap while stale results can no longer survive into a new run.

**Verdict:** correct globs, correct ordering, and scoped to the reversible half; ready.

---

Okuro3499 (approved):

PR #16113, opened 2026-08-24
teams: smoother surveys repository querying (fixes #16157)

### LINKED ISSUE #16157: Refactor: Remove redundant identity maps in SurveysRepositoryImpl
Tracking issue for PR #16113.

Removed redundant identity `.map { it }` calls in `SurveysRepositoryImpl`.

These `.map { it }` calls did nothing but create copies of existing lists or add unnecessary no-op wrappers in Sequence chains. Removing them streamlines the code and avoids unnecessary list allocations. The return values (e.g. from DAO queries or local lists like `entities.asReversed()`) can be safely returned directly.

---
*PR created automatically by Jules for task [11145326985975867422](https://jules.google.com/task/11145326985975867422) started by @dogi*

dogi:
Removed redundant identity `.map { it }` calls in `SurveysRepositoryImpl`.

These `.map { it }` calls did nothing but create copies of existing lists or add unnecessary no-op wrappers in Sequence chains. Removing them streamlines the code and avoids unnecessary list allocations. The return values (e.g. from DAO queries or local lists like `entities.asReversed()`) can be safely returned directly.

---
*PR created automatically by Jules for task [11145326985975867422](https://jules.google.com/task/11145326985975867422) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6677
-        versionName = "0.66.77"
+        versionCode = 6678
+        versionName = "0.66.78"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/repository/SurveysRepositoryImpl.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/SurveysRepositoryImpl.kt
@@ -61,7 +61,7 @@ class SurveysRepositoryImpl @Inject constructor(
     }
 
     override suspend fun getExamQuestions(examId: String): List<ExamQuestion> {
-        return questionDao.getByExamId(examId).map { it }
+        return questionDao.getByExamId(examId)
     }
 
     override suspend fun adoptSurvey(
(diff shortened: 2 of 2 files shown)


dogi:
**Review — rating: 86/100**

Five of the six removals are exactly what the title says — `.map { it }` allocating a second list (or, in the `asSequence()` chains, wrapping every element in a `TransformingSequence` with an identity transform) to produce something equal to its input. Types line up unchanged in each case, so there's nothing to test.

**The sixth one deserved the thought it evidently got.** In `getSurveys`:

```kotlin
return (if (ascending) entities else entities.asReversed()).map { it }
```

`asReversed()` returns a *view* over `entities`, not a copy — so the `.map { it }` was doing double duty as a defensive snapshot, and dropping it means callers now get a live view. That's safe here for a specific reason: `entities` is a function-local `val` produced by `sortedBy`, which returns a fresh list that nothing else holds a reference to and nothing mutates after this line. So the view can never observe a change. Worth noting because it's the one removal that's an argument rather than a tautology — if `entities` ever becomes a field or gets mutated downstream, that `.map { it }` was the thing keeping it honest.

Same family as #16107 in yesterday's batch, so the sweep is continuing; between the two, `getExamQuestions` here and the PDF exporter's `questionDao.getByExamId(it).map { question -> question }` were the same expression in two files.

**Verdict:** ready.

---

Okuro3499 (approved):

PR #16114, opened 2026-08-24
all: smoother guest user role handling (fixes #16158)

### LINKED ISSUE #16158: Optimize guest role checks and use standard Base64 in UserEntity
Tracking issue for PR #16114.

Optimizes string comparisons in `isGuest()` method to remove unnecessary allocations and safe calls.
Replaces the Android-specific `Base64` usage with standard Java `Base64` to reduce framework dependency in `UserEntity`.
Updates corresponding unit tests.

---
*PR created automatically by Jules for task [14360611546774808612](https://jules.google.com/task/14360611546774808612) started by @dogi*

dogi:
Optimizes string comparisons in `isGuest()` method to remove unnecessary allocations and safe calls.
Replaces the Android-specific `Base64` usage with standard Java `Base64` to reduce framework dependency in `UserEntity`.
Updates corresponding unit tests.

---
*PR created automatically by Jules for task [14360611546774808612](https://jules.google.com/task/14360611546774808612) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6678
-        versionName = "0.66.78"
+        versionCode = 6679
+        versionName = "0.66.79"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/model/UserEntity.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/model/UserEntity.kt
@@ -1,6 +1,5 @@
 package org.ole.planet.myplanet.model
 
-import android.util.Base64
 import androidx.core.net.toUri
 import androidx.room.Entity
 import androidx.room.Index
@@ -120,7 +119,7 @@ open class UserEntity(
 
             inputStream?.use {
                 val bytes = it.readBytes()
-                Base64.encodeToString(bytes, Base64.NO_WRAP)
+                java.util.Base64.getEncoder().encodeToString(bytes)
             }
         } catch (e: Exception) {
             e.printStackTrace()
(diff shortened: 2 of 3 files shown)


dogi:
**Review — rating: 88/100**

Both halves check out, and the Base64 swap is worth more than the title implies.

`android.util.Base64.encodeToString(bytes, NO_WRAP)` and `java.util.Base64.getEncoder().encodeToString(bytes)` produce byte-identical output — both are the RFC 4648 basic alphabet, no line wrapping, with padding. `java.util.Base64` needs API 26 and `minSdk = 26`, so it's exactly at the floor and safe. The payoff is in the test: because it's a JVM class rather than an Android one, the test could drop `every { Base64.encodeToString(any(), NO_WRAP) } returns "encoded_base64_string"` and assert the *actual* encoding of the input bytes. The old test only proved the mock echoed its stub — it would have passed against an implementation that encoded nothing at all.

The `isGuest()` change is a straight allocation win: `it?.lowercase() == "guest"` built a new String per role per call, `it.equals("guest", ignoreCase = true)` compares in place, and null receivers still yield false so the null-safety is unchanged. To be precise about what it *isn't* — the bare `lowercase()` was using the default locale, but neither "guest" nor "learner" contains an `I`, so there was no Turkish-locale bug here, unlike the `isGif`/`"debit"` cases elsewhere in the repo. This is purely fewer allocations plus a clearer expression.

**Nit:** the test still has `import android.util.Base64` and `mockkStatic(Base64::class)` in `setup()`, both now with nothing using them — this PR removed their last consumers. And `java.util.Base64` is written fully qualified at both the production and test call sites where an import would read better now that the `android.util` one is gone.

**Nit:** while you're in this file, `isManager`-style checks at lines 173 and 178 still use `contains("manager", ignoreCase = true)` / `contains("leader", …)`. #16110 in this same batch replaces exactly that substring test with an exact match in `LoginSyncManager`, so these two will disagree once both land.

**Verdict:** ready.

---

Okuro3499 (approved):

PR #16115, opened 2026-08-24
Fix: suppress DEPRECATION warning in MainApplication

dogi:
Fix: suppress DEPRECATION warning in MainApplication

The release build scan flagged `ComponentCallbacks2.TRIM_MEMORY_RUNNING_LOW` as deprecated in API 35. Since the behavior (evict the PDF thumbnail cache) should remain, a `@Suppress("DEPRECATION")` annotation was added to ignore this warning, matching the codebase's existing pattern.

---
*PR created automatically by Jules for task [5772974590066889583](https://jules.google.com/task/5772974590066889583) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/MainApplication.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/MainApplication.kt
@@ -525,6 +525,7 @@ class MainApplication : Application(), WorkManagerConfiguration.Provider {
 
     override fun onTrimMemory(level: Int) {
         super.onTrimMemory(level)
+        @Suppress("DEPRECATION")
         if (level >= ComponentCallbacks2.TRIM_MEMORY_RUNNING_LOW) {
             PdfThumbnailLoader.evictAll()
         }


dogi:
**Review — rating: 70/100**

The annotation does what the title says and the placement is good — scoping `@Suppress("DEPRECATION")` to the `if` rather than the function or the file keeps it from hiding anything else that gets deprecated in here later.

**But it silences a question rather than answering it.** `TRIM_MEMORY_RUNNING_LOW` is deprecated because the platform stopped delivering the `TRIM_MEMORY_RUNNING_*` levels — on API 34+ `onTrimMemory` is essentially only called with `TRIM_MEMORY_COMPLETE`. So on current Android the condition `level >= TRIM_MEMORY_RUNNING_LOW` may simply never be reached, and `PdfThumbnailLoader.evictAll()` never runs. That's the thing worth knowing, and after this change the compiler stops asking.

Two ways out, either fine:

- If the eviction is still wanted, keep the callback (it's live for the pre-34 install base, which `minSdk = 26` makes a real share) and add a path that works on modern Android — reacting to `TRIM_MEMORY_COMPLETE`, or just bounding the thumbnail cache so it doesn't need a memory callback at all.
- If it's genuinely fine as-is, put the reason in a comment on the same line as the suppression. A bare `@Suppress` is indistinguishable six months later from one that was added without looking, and this one *was* analysed — that analysis is the valuable part and it currently lives only in this PR.

I'd take the comment as sufficient; I'm not asking for the modern path in this PR.

**Verdict:** one line of justification (or a confirmation that eviction still fires on 34+) and this is ready.

---

PR #16117, opened 2026-08-24
all: smoother feedback caching (fixes #16160)

### LINKED ISSUE #16160: Optimize FeedbackAdapter by caching strings, dates, and listeners
Tracking issue for PR #16117.

Optimize `FeedbackAdapter` to prevent re-resolving string resources, formatting dates twice, and allocating a fresh click listener lambda on every row bind. This brings it in line with the optimizations already present for the color state lists in `onCreateViewHolder`.

---
*PR created automatically by Jules for task [14209360592373769036](https://jules.google.com/task/14209360592373769036) started by @dogi*

dogi:
Optimize `FeedbackAdapter` to prevent re-resolving string resources, formatting dates twice, and allocating a fresh click listener lambda on every row bind. This brings it in line with the optimizations already present for the color state lists in `onCreateViewHolder`.

---
*PR created automatically by Jules for task [14209360592373769036](https://jules.google.com/task/14209360592373769036) started by @dogi*


dogi:
**Review — rating: 87/100**

Three changes, and the listener move is the one that matters most. Assigning `setOnClickListener` in `onBindViewHolder` allocated a fresh lambda on every bind — every scroll, every diff — where doing it once in the holder's `init` allocates one per holder, so a dozen for the whole list. Making `FeedbackViewHolder` an `inner class` to reach `getItem` is the right price for that; the implicit adapter reference costs nothing since the adapter outlives its holders.

The position handling is correct in the details that usually go wrong: `bindingAdapterPosition` (not `absoluteAdapterPosition`) is the position in this adapter's own list, which is what `ListAdapter.getItem` expects, and the `!= RecyclerView.NO_POSITION` guard covers a click landing on a detached holder.

Caching the three `getString` calls next to the existing colour caching is consistent — same `onCreateViewHolder` guard, same lifecycle, and the adapter is recreated on a configuration change so a locale switch re-reads them. Computing `getFormattedDate(feedback.openTime)` once instead of twice per bind is the simplest win of the three; it was being called for the content description and again for `tvOpenDate`.

Worth knowing about the click rewrite: resolving the item at click time rather than capturing it in bind opens a narrow window during a diff where `getItem(position)` reflects the new list while the row still shows the old one, so a click landing mid-animation can open a neighbour. It's the documented `ListAdapter` trade-off and the position-based form is the mainstream recommendation, so I'd keep it — just don't be surprised if a very fast tap during a refresh ever opens the wrong feedback.

**Nit:** `rowFeedbackBinding.root.contentDescription = feedback.title` inside the click handler looks vestigial — it sets an accessibility label at the moment the click has already been delivered, and it overwrites nothing useful since the descriptive string in `onBind` goes on `feedbackCardView`. Carried over unchanged from master, but this is the PR that touches it.

**Verdict:** ready.

---

Okuro3499 (approved):

PR #16116, opened 2026-08-24
sync: smoother photo url utils uploading (fixes #16159)

dogi:
Hoists the `UrlUtils.getUrl()` call out of the `PhotoUploader` per-photo `forEach` loop. This avoids unnecessary IO/SP lookups since the base URL is static per session.

---
*PR created automatically by Jules for task [7932366732563818232](https://jules.google.com/task/7932366732563818232) started by @dogi*

PR #16118, opened 2026-08-24
sync: smoother download auth header working (fixes #16161)

### LINKED ISSUE #16161: fix: hoist auth header in DownloadWorker
Tracking issue for PR #16118.

Hoist the auth header out of the per-file loop in DownloadWorker and pass it down to downloadFile so the Basic auth is not re-encoded for every file.

---
*PR created automatically by Jules for task [10326784632570692069](https://jules.google.com/task/10326784632570692069) started by @dogi*

dogi:
Hoist the auth header out of the per-file loop in DownloadWorker and pass it down to downloadFile so the Basic auth is not re-encoded for every file.

---
*PR created automatically by Jules for task [10326784632570692069](https://jules.google.com/task/10326784632570692069) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6681
-        versionName = "0.66.81"
+        versionCode = 6682
+        versionName = "0.66.82"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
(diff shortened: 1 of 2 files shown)


dogi:
**Review — rating: 86/100**

Same shape as #16116 in this batch and correct for the same reason: `UrlUtils.header` builds a Basic auth string from preferences on every read, and it was being read once per file inside `urls.forEachIndexed`. Threading it in as a parameter rather than capturing it makes the dependency explicit at `downloadFile`'s signature, which is the better of the two ways to do this.

The semantics change slightly and in the right direction: every file in a batch now uses the credentials as of loop start. Previously a mid-batch credential change (a re-login, a server switch) would apply from the next file onward, so a single download run could span two auth contexts — which is more likely to produce a confusing half-failed batch than to rescue one. Worth a line in the PR body since "pure hoist" undersells it.

No test, which is fine here — there's no observable behaviour to pin beyond the call count, and asserting "read the header once" would be testing the implementation rather than the outcome.

**Verdict:** ready.

---

Okuro3499 (approved):

PR #16121, opened 2026-08-24
community: smoother voices view modelling (fixes #16165)

### LINKED ISSUE #16165: fix: Drop map widening cast in VoicesViewModel.observeCommunityNews
Tracking issue for PR #16121.

Dropped the unnecessary list allocation map in `VoicesViewModel.observeCommunityNews` when casting from `List<News>` to `List<News?>`. Instead of allocating a new list, simply assign it statically with the explicit generic type.

---
*PR created automatically by Jules for task [11786779846300412122](https://jules.google.com/task/11786779846300412122) started by @dogi*

dogi:
Dropped the unnecessary list allocation map in `VoicesViewModel.observeCommunityNews` when casting from `List<News>` to `List<News?>`. Instead of allocating a new list, simply assign it statically with the explicit generic type.

---
*PR created automatically by Jules for task [11786779846300412122](https://jules.google.com/task/11786779846300412122) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6684
-        versionName = "0.66.84"
+        versionCode = 6685
+        versionName = "0.66.85"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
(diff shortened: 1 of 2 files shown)


dogi:
**Review — rating: 86/100**

The cast was never doing any work: Kotlin's `List<out E>` is covariant, so a `List<News>` already *is* a `List<News?>` and the explicit type on the local is enough. What the old line actually cost was a full `ArrayList` copy on every emission of `getCommunityNews`, which for a community feed is once per sync tick.

I checked the two consumers before signing off on sharing the repository's list instance instead of a private copy: `collectLabels(filtered)` and `filterNews` both only read (`list.forEach` populating a local map, then filtering into new lists), so nothing mutates the emitted list. `_baseNewsList` is a `MutableStateFlow<List<News?>>` and conflates structurally, and since the old copy was shallow the element identities were the same either way — so the conflation behaviour doesn't shift.

**Verdict:** genuinely equivalent and removes a per-emission copy; ready.

---

Okuro3499 (approved):

PR #16123, opened 2026-08-24
sync: smoother download service buffering (fixes #16167)

### LINKED ISSUE #16167: Optimize DownloadService download loop math and buffering
Tracking issue for PR #16123.

Optimize DownloadService download loop math and buffering

---
*PR created automatically by Jules for task [13660804626659621832](https://jules.google.com/task/13660804626659621832) started by @dogi*

dogi:
Optimize DownloadService download loop math and buffering

---
*PR created automatically by Jules for task [13660804626659621832](https://jules.google.com/task/13660804626659621832) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6686
-        versionName = "0.66.86"
+        versionCode = 6687
+        versionName = "0.66.87"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/services/DownloadService.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/services/DownloadService.kt
@@ -69,7 +69,7 @@ class DownloadService : Service() {
     @Inject
     lateinit var sharedPrefManager: SharedPrefManager
 
-    private var data = ByteArray(1024 * 4)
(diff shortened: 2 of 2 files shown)


dogi:
**Review — rating: 88/100**

The hoist is safe because `current` had exactly one consumer: I traced it through `copyStreamWithProgress` and it was only ever read inside the `now - lastNotificationUpdateTime >= NOTIFICATION_UPDATE_INTERVAL_MS` branch. `download.progress` / `currentFileProgress` are still assigned unconditionally in the `fileSize > 0` branch, so the intent payload is unchanged — only the throttled notification path moved. That takes a `Long`-to-`Double` division plus a `roundToInt` off the hot loop, which at the old 4 KiB buffer ran once per 4 KiB of every download.

`(total / 1024.0).roundToInt().toDouble()` followed by `.toInt()` was also a pointless round-trip through `Double`; `download.currentFileSize` is an `Int`, so the new form is both cheaper and more honest. Quadrupling the buffer to 16 KiB cuts read/write syscalls by 4× and matches what `BufferedInputStream` wants to hand over.

**Nit:** `data` is a single mutable `ByteArray` field on the `Service`, so it is shared across whatever downloads that service instance runs and is retained for the service's lifetime — now 16 KiB rather than 4 KiB. Both the sharing and the retention are pre-existing, not introduced here, but the PR touches that exact declaration so it's the natural moment to make it a local of `copyStreamWithProgress`.

**Verdict:** correct hoist and a well-chosen buffer size; ready.

---

Okuro3499 (approved):

PR #16125, opened 2026-08-24
life: smoother health examination handling (fixes #16169)

### LINKED ISSUE #16169: Fix trailing comma in health examination condition list
Tracking issue for PR #16125.

Fix trailing comma in health examination condition list

Replaced the manual StringBuilder loop that accumulated "$key, " per
true condition with `filter { ... }.joinToString(", ")`. This prevents
the dialog text from ending with a dangling ", " separator, fixing a
visible cosmetic defect.

---
*PR created automatically by Jules for task [2210295721415340900](https://jules.google.com/task/2210295721415340900) started by @dogi*

dogi:
Fix trailing comma in health examination condition list

Replaced the manual StringBuilder loop that accumulated "$key, " per
true condition with `filter { ... }.joinToString(", ")`. This prevents
the dialog text from ending with a dangling ", " separator, fixing a
visible cosmetic defect.

---
*PR created automatically by Jules for task [2210295721415340900](https://jules.google.com/task/2210295721415340900) started by @dogi*


dogi:
**Review — rating: 88/100**

This is a real user-visible fix, not just an idiom swap: the old loop appended `"$key, "` for every true condition including the last, so the rendered text always ended in a dangling `", "` — "Diabetes, Asthma, " instead of "Diabetes, Asthma". `joinToString(", ")` puts the separator between elements only, which is exactly the intent.

I checked the two edge cases where a rewrite like this usually changes behaviour and neither shifts. With no true conditions, the old code produced `""` from an empty `StringBuilder` and `joinToString` produces `""` from an empty list. And if `conditionsMap[key].asBoolean` throws on a non-boolean value, the old code threw mid-loop *before* the post-loop assignment to `conditionsText`, while the new code throws during `filter` — also before the assignment. Either way `conditionsText` keeps its prior value and the `catch` swallows it, so the failure mode is identical.

**Verdict:** correct fix for the trailing separator with no behavioural drift on the empty or throwing paths; ready.

---

Okuro3499 (approved):

PR #16126, opened 2026-08-24
teams: smoother voices label managing (fixes #16170)

### LINKED ISSUE #16170: Collapse double map lookup in VoicesLabelManager
Tracking issue for PR #16126.

Collapse the double map lookup in `VoicesLabelManager`'s delete listener.

The `when` block contained `Constants.LABELS.containsKey(labelText) -> Constants.LABELS[labelText]`.
This was changed to directly use `Constants.LABELS[labelText]` and fall back to searching the list of labels if it returns null, avoiding the double lookup.

---
*PR created automatically by Jules for task [5912152280780596553](https://jules.google.com/task/5912152280780596553) started by @dogi*

dogi:
Collapse the double map lookup in `VoicesLabelManager`'s delete listener.

The `when` block contained `Constants.LABELS.containsKey(labelText) -> Constants.LABELS[labelText]`.
This was changed to directly use `Constants.LABELS[labelText]` and fall back to searching the list of labels if it returns null, avoiding the double lookup.

---
*PR created automatically by Jules for task [5912152280780596553](https://jules.google.com/task/5912152280780596553) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6689
-        versionName = "0.66.89"
+        versionCode = 6690
+        versionName = "0.66.90"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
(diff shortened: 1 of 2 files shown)


dogi:
**Review — rating: 86/100**

The collapse is exactly equivalent here, and the reason is worth stating because it wouldn't hold for every map: `Constants.LABELS` is declared `mutableMapOf<String, String>()`, so its value type is non-nullable and `LABELS[labelText]` can only return `null` when the key is *absent*. That makes `containsKey(k) -> LABELS[k]` and `LABELS[k] ?: fallback` indistinguishable. If the value type were `String?` the rewrite would silently change behaviour, since a key present with a null value would newly fall through to the `labels.firstOrNull` branch.

Beyond saving one hash lookup on the common path, it also drops one read of a mutable global — `Constants.LABELS` is a `var` reassigned during init, so the old form read it twice with a window in between.

**Verdict:** equivalent given the non-null value type, and marginally more robust against the mutable global; ready.

---

Okuro3499 (approved):

PR #16132, opened 2026-08-25
login: smoother auth session updating (fixes #16222)

### LINKED ISSUE #16222: Fix resource leak in AuthSessionUpdater.sendPost
Tracking issue for PR #16132.

Fixes a resource leak in `AuthSessionUpdater.sendPost` by properly closing the `DataOutputStream` and ensuring `HttpURLConnection.disconnect()` is called on exceptions. Also replaces `printStackTrace()` with `Log.w`.

---
*PR created automatically by Jules for task [1382469628494991791](https://jules.google.com/task/1382469628494991791) started by @dogi*

dogi:
Fixes a resource leak in `AuthSessionUpdater.sendPost` by properly closing the `DataOutputStream` and ensuring `HttpURLConnection.disconnect()` is called on exceptions. Also replaces `printStackTrace()` with `Log.w`.

---
*PR created automatically by Jules for task [1382469628494991791](https://jules.google.com/task/1382469628494991791) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6710
-        versionName = "0.67.10"
+        versionCode = 6711
+        versionName = "0.67.11"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/data/auth/AuthSessionUpdater.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/data/auth/AuthSessionUpdater.kt
@@ -2,6 +2,7 @@ package org.ole.planet.myplanet.data.auth
 
 import dagger.assisted.Assisted
 import dagger.assisted.AssistedFactory
+import android.util.Log
 import dagger.assisted.AssistedInject
 import java.io.DataOutputStream
 import java.net.HttpURLConnection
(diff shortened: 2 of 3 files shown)


dogi:
**Review — rating: 85/100**

The leak is real and the fix is the right shape. On master `conn.disconnect()` sat on the success path only, so anything that threw between `openConnection()` and it — a socket timeout during `conn.outputStream`, a malformed body from `getJsonObject()`, an exception out of `callback.setAuthSession` — left the connection to be reclaimed by finalization. Moving it into `finally` closes all of those, and `DataOutputStream(...).use { }` does the same for the stream, which previously leaked on any throw between `writeBytes` and `close()`.

Ordering is preserved where it matters: `callback.setAuthSession(conn.headerFields)` still runs inside the `try`, before `disconnect()`, so the headers are read from a live connection.

**Nit:** the new imports break the file's alphabetical ordering — `android.util.Log` lands between `dagger.assisted.AssistedFactory` and `dagger.assisted.AssistedInject` in the main file, and in the test `io.mockk.mockkStatic` sits after the JUnit block with `android.util.Log` trailing the whole list. Also worth a comment while you're in here: the method is `sendPost` but sets `requestMethod = "GET"` with `doOutput = true`, which makes `HttpURLConnection` silently switch to POST. That's pre-existing, not yours, but it will confuse the next reader.

**Verdict:** correct leak fix with the header read still inside the live-connection window; ready once the imports are sorted.

---

Okuro3499 (approved):

PR #16133, opened 2026-08-25
teams: smoother base voices members selecting (fixes #16223)

### LINKED ISSUE #16223: Fix Voices member navigation view lifecycle crash
Tracking issue for PR #16133.

Fixed a potential crash in `BaseVoicesFragment.kt` by replacing `lifecycleScope` with `viewLifecycleOwner.lifecycleScope` for `onMemberSelected` navigation and ensuring fragment attachment is checked after a suspend function returns.

---
*PR created automatically by Jules for task [5927909977432968346](https://jules.google.com/task/5927909977432968346) started by @dogi*

dogi:
Fixed a potential crash in `BaseVoicesFragment.kt` by replacing `lifecycleScope` with `viewLifecycleOwner.lifecycleScope` for `onMemberSelected` navigation and ensuring fragment attachment is checked after a suspend function returns.

---
*PR created automatically by Jules for task [5927909977432968346](https://jules.google.com/task/5927909977432968346) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6711
-        versionName = "0.67.11"
+        versionCode = 6712
+        versionName = "0.67.12"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/base/BaseVoicesFragment.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/base/BaseVoicesFragment.kt
@@ -102,8 +102,9 @@ abstract class BaseVoicesFragment : BaseContainerFragment(), OnNewsItemClickList
     override fun onMemberSelected(userModel: UserEntity?) {
         if (!isAdded) return
 
-        lifecycleScope.launch {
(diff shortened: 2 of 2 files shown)


dogi:
**Review — rating: 88/100**

This is the correct fix for the crash, and the distinction is the whole point: a Fragment's `lifecycleScope` is tied to the *Fragment*, which outlives its view, so a coroutine started there survives `onDestroyView` and can reach `requireActivity().supportFragmentManager` with a detached fragment — `IllegalStateException`. `viewLifecycleOwner.lifecycleScope` is cancelled at `onDestroyView`, which closes the window that `showMemberDetails`' suspension opens.

The added `if (!isAdded) return@launch` after the suspend point is not redundant on top of that, which is worth saying since it looks like belt-and-braces: cancellation fires at `onDestroyView`, but `isAdded` goes false during a remove transaction while the view still exists, so the two guards cover different halves of the teardown.

I did check the one hazard this swap can introduce — `viewLifecycleOwner` throws if the view is null, and the `isAdded` guard at the top doesn't prevent that for a fragment sitting on the back stack. Both call sites are `VoicesAdapter.kt:648` and `:652`, which are click handlers on a live view, so it can't fire without a view in practice.

**Verdict:** correct scope for the coroutine and both guards pull their weight; ready.

---

Okuro3499 (approved):

PR #16135, opened 2026-08-25
courses: smoother inline resources caching (fixes #16225)

### LINKED ISSUE #16225: Fix: purge the `InlineResourceAdapter` preview cache in one pass
Tracking issue for PR #16135.

Optimizes removing stale previews from `InlineResourceAdapter`'s `textCache` by using `removeAll` instead of `filter {}.forEach {}`, avoiding intermediate allocations.

---
*PR created automatically by Jules for task [13242371361612207922](https://jules.google.com/task/13242371361612207922) started by @dogi*

dogi:
Optimizes removing stale previews from `InlineResourceAdapter`'s `textCache` by using `removeAll` instead of `filter {}.forEach {}`, avoiding intermediate allocations.

---
*PR created automatically by Jules for task [13242371361612207922](https://jules.google.com/task/13242371361612207922) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6713
-        versionName = "0.67.13"
+        versionCode = 6714
+        versionName = "0.67.14"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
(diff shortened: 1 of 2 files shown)


dogi:
**Review — rating: 88/100**

Correct, and the reason it's safe is worth checking rather than assuming, since `keys.removeAll { }` on a map view is exactly where a `ConcurrentModificationException` would hide. Kotlin's `MutableIterable.removeAll(predicate)` goes through `filterInPlace`, which walks a `MutableIterator` and calls `iterator.remove()` — the sanctioned way to mutate during iteration — so it's CME-free, unlike a naive `forEach { map.remove(it) }` over the live key set.

The other thing I checked is what `textCache` actually is, because `Collections.synchronizedMap` would make this unsafe (its `keySet` isn't synchronized for iteration). It's `private val textCache = mutableMapOf<String, String>()` at line 54 — a plain `LinkedHashMap` touched only from adapter callbacks on the main thread, so there's no concurrency to worry about.

Net effect is one pass instead of two plus no intermediate `ArrayList` of matching keys, which for a prefix purge over a preview cache is the whole point.

**Verdict:** correct in-place removal with no CME exposure; ready.

---

Okuro3499 (approved):

PR #16137, opened 2026-08-25
all: smoother notifications load view modelling (fixes #16227)

### LINKED ISSUE #16227: Fix unnecessary allocations in `loadNotifications`
Tracking issue for PR #16137.

Replaced the `groupBy { it.type.lowercase() }` operation with a single-pass loop using case-insensitive string comparisons.
This prevents allocating a lowercase string for every notification payload and eliminates the allocation of the multimap, while still preserving the insertion order within the two extracted lists.
Added a test to verify correct extraction using mixed-case types while ignoring other types.

---
*PR created automatically by Jules for task [10710371114674323822](https://jules.google.com/task/10710371114674323822) started by @dogi*

dogi:
Replaced the `groupBy { it.type.lowercase() }` operation with a single-pass loop using case-insensitive string comparisons.
This prevents allocating a lowercase string for every notification payload and eliminates the allocation of the multimap, while still preserving the insertion order within the two extracted lists.
Added a test to verify correct extraction using mixed-case types while ignoring other types.

---
*PR created automatically by Jules for task [10710371114674323822](https://jules.google.com/task/10710371114674323822) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6716
-        versionName = "0.67.16"
+        versionCode = 6717
+        versionName = "0.67.17"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/ui/notifications/NotificationsViewModel.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/notifications/NotificationsViewModel.kt
@@ -59,9 +59,15 @@ class NotificationsViewModel @Inject constructor(
         viewModelScope.launch {
             val payloadNotifications = notificationsRepository.getNotifications(userId, filter, isAdmin)
 
-            val byLoweredType = payloadNotifications.groupBy { it.type.lowercase() }
-            val taskNotifications = byLoweredType["task"] ?: emptyList()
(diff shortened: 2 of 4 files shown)


dogi:
**Review — rating: 89/100**

This is filed as an allocation fix but it also closes a real correctness hole, which is worth recording because nothing in the diff or the tests will show it. `it.type.lowercase()` with no `Locale` argument uses the *device* locale, and `"JOIN_REQUEST"` contains an `I` — under a Turkish locale it lowercases to `"joın_request"` with a dotless ı, which never matches the `"join_request"` key. So on a Turkish device, join-request notifications silently lost their related-ID enrichment. `String.equals(other, ignoreCase = true)` compares per-character without consulting a locale, so the new form is immune. (`"TASK"` has no `I`, so only the join-request branch was affected.) CI runs under `en_US`, so the new mixed-case test wouldn't have caught this either — the fix is real regardless.

The allocation win is genuine too: `groupBy` materialised a `LinkedHashMap` plus a `MutableList` for *every* distinct type in the payload — `resource`, and whatever else the server sends — when only two buckets are ever read. The single pass builds exactly the two lists needed.

Order is preserved either way (`groupBy` keeps encounter order within each group, and the loop appends in order), and the new test pins that with `coVerify { repository.getTaskTeamNamesByTaskIds(listOf("rel1", "rel3")) }` — an ordered list, so a reordering regression would fail it.

**Verdict:** correct, cheaper, and incidentally fixes a Turkish-locale mismatch on the join-request path; ready.

---

Okuro3499 (approved):

PR #16138, opened 2026-08-25
perf(courses): avoid allocating map key list on every bind

dogi:
Refactored `showStepMistakes` in `CoursesProgressAdapter` to directly iterate over `stepMistake` entries rather than extracting `keys.toList()`, eliminating avoidable map allocation and secondary `stepMistake[stepKey]` lookups per step during `onBindViewHolder`.
Updated `CoursesProgressAdapterTest` to explicitly verify text values inside layout children, ensuring stale data is properly overwritten upon rebinding.

---
*PR created automatically by Jules for task [4642710095273456428](https://jules.google.com/task/4642710095273456428) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/ui/courses/CoursesProgressAdapter.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/courses/CoursesProgressAdapter.kt
@@ -43,11 +43,10 @@ class CoursesProgressAdapter(private val context: Context) : ListAdapter<Courses
 
         if (stepMistake != null && stepMistake.isNotEmpty()) {
             binding.llHeader.visibility = View.VISIBLE
-            val keys = stepMistake.keys.toList()
             val textColor = ContextCompat.getColor(context, R.color.daynight_textColor)
 
             val currentChildCount = binding.llProgress.childCount
-            val requiredChildCount = keys.size
+            val requiredChildCount = stepMistake.size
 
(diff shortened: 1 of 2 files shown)


dogi:
**Review — rating: 83/100**

**Collision with #16139** — same file, same two hunks, same net behaviour. Both drop the `keys.toList()` copy per bind and both take the mistake count straight off the map entry instead of re-looking it up, and both add the same positional assertions to `CoursesProgressAdapterTest`. I verified the shared premise for both: `keys.toList()` snapshots the map's own iteration order, so iterating directly gives the identical sequence, and `stepMistake` is a Gson-deserialised `Map<String, Int>` (a `LinkedTreeMap`), so that order is the JSON key order and deterministic.

Your main-code form is the marginally better of the two — `for ((stepKey, mistakeCount) in stepMistake)` reads more naturally than `forEach` with an external counter. The difference that decides it is the test: this PR writes `android.widget.TextView` fully-qualified at six call sites, where #16139 adds `import android.widget.TextView` and keeps the assertions readable.

**Recommend closing in favour of #16139**, ideally after porting the `for ((k, v) in …)` loop form over to it, since that half is better here. Either way `entries.forEachIndexed` would drop the manual `i` counter both PRs keep.

**Verdict:** correct work, but duplicated by #16139, which wins on test readability; recommending close, not a comment on the change itself.

---

PR #16139, opened 2026-08-25
courses: smoother progress steps handling (fixes #16228)

### LINKED ISSUE #16228: Refactor map iterations in `CoursesProgressAdapter`
Tracking issue for PR #16139.

Avoid copying `stepMistake.keys.toList()` per recycle and iterating the map with key lookups. Instead, iterate `stepMistake.forEach` natively without extracting the key list to avoid memory and performance bottlenecks on UI re-binding. Tested view recycles.

---
*PR created automatically by Jules for task [5138935647577084094](https://jules.google.com/task/5138935647577084094) started by @dogi*

dogi:
Avoid copying `stepMistake.keys.toList()` per recycle and iterating the map with key lookups. Instead, iterate `stepMistake.forEach` natively without extracting the key list to avoid memory and performance bottlenecks on UI re-binding. Tested view recycles.

---
*PR created automatically by Jules for task [5138935647577084094](https://jules.google.com/task/5138935647577084094) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6717
-        versionName = "0.67.17"
+        versionCode = 6718
+        versionName = "0.67.18"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/ui/courses/CoursesProgressAdapter.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/courses/CoursesProgressAdapter.kt
@@ -43,11 +43,10 @@ class CoursesProgressAdapter(private val context: Context) : ListAdapter<Courses
 
         if (stepMistake != null && stepMistake.isNotEmpty()) {
             binding.llHeader.visibility = View.VISIBLE
-            val keys = stepMistake.keys.toList()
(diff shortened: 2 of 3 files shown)


dogi:
**Review — rating: 85/100**

Correct, and the second win is the one the title doesn't mention: dropping `stepMistake[stepKey]` removes a hash lookup per row on top of removing the `keys.toList()` copy per bind, because the entry already carries the value.

Iteration order is preserved, which is what this change hinges on. `keys.toList()` snapshots the map's own iteration order, so walking the map directly yields exactly the same sequence — and `stepMistake` is `Map<String, Int>` deserialised by Gson, which produces a `LinkedTreeMap`, so the order is the JSON key order and is deterministic. That also makes the new positional assertions in the test safe rather than brittle.

Those assertions are the best part of the PR: the previous test only checked `childCount`, so a bug that paired step 1's label with step 3's mistake count would have passed. Asserting both `TextView`s per row pins the actual step→count mapping.

**Collision:** #16138 is the same change to the same lines — I've marked that one `duplicate` and recommended closing it in favour of this PR, because it fully-qualifies `android.widget.TextView` six times in the test where this one adds the import.

**Nit:** the manual `var i = 0` / `i++` alongside `forEach` mixes imperative and functional styles; `stepMistake.entries.forEachIndexed { i, (stepKey, mistakes) -> … }` drops the counter entirely.

**Verdict:** correct, cheaper, and the new assertions cover a real pairing bug the old test couldn't see; ready.

---

Okuro3499 (approved):

PR #16140, opened 2026-08-25
Refactor AddHealthActivity to fully utilize view binding

dogi:
finish `AddHealthActivity`'s migration to its own view binding

---
*PR created automatically by Jules for task [10156740009156534314](https://jules.google.com/task/10156740009156534314) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/ui/health/AddHealthActivity.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/health/AddHealthActivity.kt
@@ -3,11 +3,9 @@ package org.ole.planet.myplanet.ui.health
 import android.app.DatePickerDialog
 import android.os.Bundle
 import android.view.MenuItem
-import android.view.View
 import android.widget.ArrayAdapter
-import android.widget.ImageView
-import android.widget.Spinner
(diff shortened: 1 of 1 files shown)


dogi:
**Review — rating: 79/100**

**Collision with #16178** — the same four `findViewById` call sites in `AddHealthActivity` replaced by the same binding accessors. The substitutions themselves are correct in both.

This PR has the better idea and the worse execution. `binding.progressBar.isVisible = isLoading` is genuinely nicer than the `if (isLoading) View.VISIBLE else View.GONE` ternary #16178 keeps. But it also removes `import android.view.View` while `View.OnClickListener` is still used four lines further down, and compensates by writing `android.view.View.OnClickListener` fully-qualified inline — an import was deleted that was still needed, and the FQN is the scar. The new `androidx.core.view.isVisible` import also lands above `androidx.appcompat.app.AppCompatActivity`, breaking the block's alphabetical order.

**Recommend closing in favour of #16178** and moving the `isVisible` change over to it — that's the half worth keeping, and it costs one correctly-placed import line.

**Verdict:** duplicated by #16178, which lands the same substitutions without deleting a live import; recommending close.

---

PR #16164, opened 2026-08-25
sync: smoother upload coordinating (fixes #16229)

### LINKED ISSUE #16229: Refactor: Extract upload DB reconciliation logic in UploadCoordinator
Tracking issue for PR #16164.

Extract duplicate DB reconciliation logic into a shared helper

This removes the duplicated logic between the `upload` and `uploadRoom` methods in `UploadCoordinator.kt` by extracting it into a single private helper function `reconcileDbFailures()`. This prevents code drift while ensuring exception text, retry policies, and ID ordering remain completely preserved.

---
*PR created automatically by Jules for task [16051839481271271282](https://jules.google.com/task/16051839481271271282) started by @dogi*

dogi:
Extract duplicate DB reconciliation logic into a shared helper

This removes the duplicated logic between the `upload` and `uploadRoom` methods in `UploadCoordinator.kt` by extracting it into a single private helper function `reconcileDbFailures()`. This prevents code drift while ensuring exception text, retry policies, and ID ordering remain completely preserved.

---
*PR created automatically by Jules for task [16051839481271271282](https://jules.google.com/task/16051839481271271282) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6718
-        versionName = "0.67.18"
+        versionCode = 6719
+        versionName = "0.67.19"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/services/upload/UploadCoordinator.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/services/upload/UploadCoordinator.kt
@@ -57,22 +57,9 @@ class UploadCoordinator @Inject constructor(
                 var dbFailedErrors = emptyList<UploadError>()
                 if (succeeded.isNotEmpty()) {
                     val dbFailed = updateDatabaseBatch(succeeded, config)
-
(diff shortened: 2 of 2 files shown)


dogi:
**Review — rating: 85/100**

The two blocks really were the same logic, and I checked they were equivalent before agreeing they collapse. The first wrote its `UploadError` with named arguments and the second positionally; the first bound `actuallySucceeded` to a local before adding, the second inlined the `filter` — but both produce the same error list and the same surviving-success set, including the `dbFailed.isEmpty()` fast path that skips building the `HashSet`. Extracting them loses nothing.

Unifying them is worth more than the ~13 lines saved: `updateDatabaseBatch` and `updateDatabaseBatchRoom` are two paths through the same reconciliation, and having them drift — one forgetting to exclude DB-failed IDs from `allSucceeded`, say — would silently mark unsaved uploads as done. One implementation makes that impossible.

**Nit:** `reconcileDbFailures` is inserted with no blank line before it and two after, so it butts against the preceding `}`. And now that the logic has a clean signature — two lists in, a pair out, no I/O — it's directly unit-testable; a test for the partial-failure case (some IDs in `dbFailed`, some not) would lock in the behaviour that matters and there's currently none.

**Verdict:** faithful extraction that removes a real drift risk between the two upload paths; ready.

---

Okuro3499 (approved):

PR #16174, opened 2026-08-25
Refactor MyLife to remove Context dependency

dogi:
This change removes the `android.content.Context` dependency from the `MyLife` data model, shifting the responsibility of resolving string resources to the caller. This aligns the data model better with pure Kotlin architecture by removing Android framework imports and presentation work from the entity class.

---
*PR created automatically by Jules for task [16317228685270544540](https://jules.google.com/task/16317228685270544540) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/model/MyLife.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/model/MyLife.kt
@@ -1,6 +1,5 @@
 package org.ole.planet.myplanet.model
 
-import android.content.Context
 import androidx.room.Entity
 import androidx.room.Ignore
 import androidx.room.Index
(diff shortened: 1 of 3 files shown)


dogi:
**Review — rating: 83/100**

**Collision with #16175** — identical signature change (`defaultItems(context, userId)` → `defaultItems(userId, resolveLabel: (Int) -> String)`) and identical call-site updates in `DashboardPluginFragment` and `LifeViewModel`. The seam is correct in both: `defaultItems` only used the `Context` for `getString`, so passing that one capability removes the Android dependency from the model, and `context::getString` resolves to the single-argument overload.

The difference is the body. This PR keeps the explicit seven-element `listOf(MyLife(…))`, where #16175 extracts a `defaultItemPairs` table and maps over it — putting the image-id-to-string-resource mapping in one readable place, which matters here because `seedMyLifeIfEmpty` derives each item's `weight` from its list position.

**Recommend closing in favour of #16175.** Both would benefit from annotating the parameter `(@StringRes Int) -> String`.

**Verdict:** correct, but duplicated by #16175, which factors the item table out; recommending close.

---

PR #16175, opened 2026-08-25
life: smoother dashboard plugin view modelling (fixes #16231)

### LINKED ISSUE #16231: Refactor MyLife to remove Context dependency
Tracking issue for PR #16175.

This commit modifies the `MyLife.defaultItems` function to resolve string resources via a provided resolver function `(Int) -> String` rather than depending directly on an Android `Context`. This strictly separates the Android framework presentation layer from the data model.

---
*PR created automatically by Jules for task [3189559195291767378](https://jules.google.com/task/3189559195291767378) started by @dogi*

dogi:
This commit modifies the `MyLife.defaultItems` function to resolve string resources via a provided resolver function `(Int) -> String` rather than depending directly on an Android `Context`. This strictly separates the Android framework presentation layer from the data model.

---
*PR created automatically by Jules for task [3189559195291767378](https://jules.google.com/task/3189559195291767378) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6720
-        versionName = "0.67.20"
+        versionCode = 6721
+        versionName = "0.67.21"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/model/MyLife.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/model/MyLife.kt
@@ -1,6 +1,5 @@
 package org.ole.planet.myplanet.model
 
-import android.content.Context
 import androidx.room.Entity
 import androidx.room.Ignore
 import androidx.room.Index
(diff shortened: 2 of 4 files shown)


dogi:
**Review — rating: 86/100**

Taking `(Int) -> String` instead of a `Context` is the right seam — `defaultItems` only ever used the context to call `getString`, so the lambda hands over exactly that capability and `MyLife` becomes plain-JVM. Both call sites pass `context::getString` / `requireContext()::getString`, which binds the single-argument overload rather than the `(Int, vararg Any)` one.

The `defaultItemPairs` table is what makes this the better of the two versions: the image-id-to-string-resource mapping is now a single seven-line list that can be read, reordered, or extended without touching construction logic, and adding an entry is one line instead of one line plus matching argument plumbing. It also puts the seed order in one obvious place, which matters here because `LifeRepositoryImpl.seedMyLifeIfEmpty` assigns `weight` from list position.

**Collision:** #16174 makes the same signature change with the same call sites but keeps the explicit seven-element `listOf(MyLife(...))`. I've marked it `duplicate` in favour of this PR.

**Nit:** `resolveLabel: (Int) -> String` should be `resolveLabel: (@StringRes Int) -> String` — without it nothing stops a caller passing an arbitrary int, and the annotation is free.

**Verdict:** correct seam and the pairs table is a genuine readability gain over the alternative; ready.

---

Okuro3499 (approved):

PR #16176, opened 2026-08-25
Refactor: Hoist onboarding dot drawables

dogi:
Refactor `OnboardingActivity` to avoid resolving `R.drawable.selected_item_dot` and `R.drawable.non_selected_item_dot` repeatedly. Instead of calling `ContextCompat.getDrawable()` on every dot for every page change (and during initialization), these are now resolved exactly once via `by lazy` properties and reused, reducing unnecessary allocations on the onboarding screen.

---
*PR created automatically by Jules for task [16977051722285940779](https://jules.google.com/task/16977051722285940779) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/ui/onboarding/OnboardingActivity.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/onboarding/OnboardingActivity.kt
@@ -43,6 +43,9 @@ class OnboardingActivity : AppCompatActivity() {
     lateinit var dispatcherProvider: DispatcherProvider
     private var pageChangeListener: ViewPager.OnPageChangeListener? = null
 
+    private val selectedDot by lazy { ContextCompat.getDrawable(this, R.drawable.selected_item_dot) }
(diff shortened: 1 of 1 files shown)


dogi:
**Review — rating: 83/100**

**Collision with #16177** — the same two `by lazy` drawable properties replacing the same four `ContextCompat.getDrawable` calls. The only difference between the two PRs is declaration placement: this one puts them after `pageChangeListener` at the end of the property list, #16177 puts them directly after `dots`.

The change is right in both, and I verified the assumption it rests on: sharing one `Drawable` instance across several `ImageView`s is normally risky, because each `setImageDrawable` reassigns the drawable's callback and each view writes its bounds. Here both `selected_item_dot.xml` and `non_selected_item_dot.xml` are plain `<shape android:shape="oval">` with a fixed 8dp size and a solid colour — stateless and identically sized — so it's safe. If either ever gains a state list or runtime tint, this needs `.mutate()` per dot.

**Recommend closing in favour of #16177**, purely on the grouping — the dot drawables read better next to `dots` than separated from it by the whole `@Inject` block.

**Verdict:** correct, but duplicated by #16177; recommending close on placement alone.

---

PR #16177, opened 2026-08-25
login: smoother onboarding (fixes #16232)

### LINKED ISSUE #16232: Hoist onboarding dot drawables
Tracking issue for PR #16177.

Replaced repeated calls to `ContextCompat.getDrawable` in `OnboardingActivity` with `by lazy` properties to cache the drawables.

---
*PR created automatically by Jules for task [7969056712951555634](https://jules.google.com/task/7969056712951555634) started by @dogi*

dogi:
Replaced repeated calls to `ContextCompat.getDrawable` in `OnboardingActivity` with `by lazy` properties to cache the drawables.

---
*PR created automatically by Jules for task [7969056712951555634](https://jules.google.com/task/7969056712951555634) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6721
-        versionName = "0.67.21"
+        versionCode = 6722
+        versionName = "0.67.22"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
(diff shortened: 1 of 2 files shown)


dogi:
**Review — rating: 85/100**

Correct hoist. `ContextCompat.getDrawable` was being called on every page change *and* once per dot during setup — for a 5-page onboarding that's a resource lookup and inflation per swipe, for two drawables that never change. `by lazy` also keeps them off the `onCreate` path entirely if the user never reaches the dots.

I checked the one thing that makes drawable hoisting risky, which is sharing a single `Drawable` instance across several `ImageView`s: each `setImageDrawable` calls `setCallback` on the drawable, so the last view assigned owns invalidation, and views also write the drawable's bounds. Both `selected_item_dot.xml` and `non_selected_item_dot.xml` are plain `<shape android:shape="oval">` with a fixed 8dp size and a solid colour — stateless, self-invalidating never, identical bounds — so sharing is safe here. Worth knowing that if either ever gains a state list, animation, or a runtime tint, this needs `.mutate()` per dot.

**Collision:** #16176 is the same change, differing only in where the two properties are declared. I've marked it `duplicate` in favour of this PR, which puts them immediately after `dots` — the state they exist to serve — rather than below the `@Inject` block at the far end of the property list.

**Verdict:** correct hoist, and safe given both drawables are stateless shapes; ready.

---

Okuro3499 (approved):

PR #16179, opened 2026-08-25
sync: smoother retry queue working (fixes #16234)

### LINKED ISSUE #16234: Hoist base URL and auth header out of RetryQueueWorker's per-operation path
Tracking issue for PR #16179.

Hoist `baseUrl` and `authHeader` in `RetryQueueWorker` out of the inner loop to prevent them from being evaluated per-operation, saving potentially up to 50 redundant prefs reads and base64 encodings per batch. Test coverage has been updated and passes.

---
*PR created automatically by Jules for task [18065150421663874000](https://jules.google.com/task/18065150421663874000) started by @dogi*

dogi:
Hoist `baseUrl` and `authHeader` in `RetryQueueWorker` out of the inner loop to prevent them from being evaluated per-operation, saving potentially up to 50 redundant prefs reads and base64 encodings per batch. Test coverage has been updated and passes.

---
*PR created automatically by Jules for task [18065150421663874000](https://jules.google.com/task/18065150421663874000) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6723
-        versionName = "0.67.23"
+        versionCode = 6724
+        versionName = "0.67.24"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/services/retry/RetryQueueWorker.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/services/retry/RetryQueueWorker.kt
@@ -114,6 +114,9 @@ class RetryQueueWorker @AssistedInject constructor(
 
             Log.i(TAG, "RETRY_QUEUE: Processing ${pendingOperations.size} pending operations")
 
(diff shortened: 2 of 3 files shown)


dogi:
**Review — rating: 86/100**

Correct hoist, and the semantic change is worth naming because it isn't neutral — it's an improvement here specifically. `UrlUtils.getUrl()` and `UrlUtils.header` were re-read per operation, so a server-URL or credential change landing mid-drain would have split one batch across two servers, with earlier operations posted to the old host and later ones to the new. Snapshotting once per `doWork()` gives the whole run one consistent target, which for a retry queue is what you want: the operations were queued against a particular server.

Threading them as parameters rather than as fields is the right call too — `processOperation` and `processOperationInternal` stay pure with respect to that state, so a future test can drive them directly.

I checked the test-hygiene risk that `mockkObject` on a Kotlin `object` usually carries, since a leaked static mock poisons every later test in the same fork: `@After fun tearDown() { unmockkAll() }` at line 78 already covers it.

**Nit:** the new stubs write `org.ole.planet.myplanet.utils.UrlUtils` fully-qualified three times; the file could just import `UrlUtils`.

**Verdict:** correct hoist that also makes a batch's server target consistent, with teardown already handling the static mock; ready.

---

Okuro3499 (approved):

PR #16180, opened 2026-08-25
Refactor: Deduplicate showZoomableImage implementation

dogi:
- Centralized `showZoomableImage` in `ImageViewerUtils` using ViewBinding.
- Replaced custom layout lookup with `DialogZoomableImageBinding`.
- Added support for tracking `activeDialog` to prevent multiple zooms simultaneously opening.
- Dynamically detect and handle HTTP(S) paths versus local file paths within the same method.
- Removed duplicate `showZoomableImage` logic and `currentZoomDialog` from `MarkdownUtils`, routing `CustomImageSpan.onClick` straight to the unified util.
- Pruned redundant imports.

---
*PR created automatically by Jules for task [1287135684813282195](https://jules.google.com/task/1287135684813282195) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/utils/ImageViewerUtils.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/utils/ImageViewerUtils.kt
@@ -4,36 +4,53 @@ import android.app.Dialog
 import android.content.Context
 import android.graphics.Color
 import android.view.LayoutInflater
-import android.widget.ImageView
 import androidx.core.graphics.drawable.toDrawable
 import com.bumptech.glide.Glide
 import com.bumptech.glide.load.engine.DiskCacheStrategy
-import com.github.chrisbanes.photoview.PhotoView
 import java.io.File
 import java.util.Locale
 import org.ole.planet.myplanet.R
+import org.ole.planet.myplanet.databinding.DialogZoomableImageBinding
 
 object ImageViewerUtils {
-    fun showZoomableImage(context: Context, imagePath: String) {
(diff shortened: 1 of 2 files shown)


dogi:
**Review — rating: 79/100**

**Collision with #16181** — the same consolidation of `MarkdownUtils`' private `showZoomableImage` into `ImageViewerUtils`, the same view-binding migration, the same `activeDialog` tracking. Two differences decide it, both against this version.

`val isHttp = imagePath.startsWith("http", ignoreCase = true)` is looser than it needs to be — it matches any path merely beginning with those four characters, where #16181's `startsWith("http://") || startsWith("https://")` matches only actual URLs. A local file whose name begins with "http" would be handed to Glide as a remote URL.

And this PR keeps `import org.ole.planet.myplanet.R` in `MarkdownUtils`, which becomes unused once `showZoomableImage` moves out — every remaining `R.` reference in that file lived inside the deleted function, and the one on line 174 is `Layout.Alignment`, not a resource. #16181 removes it.

**Recommend closing in favour of #16181.** The two findings I raised there apply equally here and are worth carrying over: `VoicesAdapter.kt:975` holds a *third* copy of this function that neither PR folds in, and the relocated `imagePath.lowercase(Locale.getDefault()).endsWith(".gif")` is locale-sensitive — `"IMG.GIF"` lowercases to `"ımg.gıf"` in Turkish and misses the branch, which `endsWith(".gif", ignoreCase = true)` fixes.

**Verdict:** correct in outline but duplicated by #16181, which has the tighter URL predicate and no leftover import; recommending close.

---

PR #67, opened 2018-06-11
Insert data from couchdb to realm db

leomaxi:

PR #16181, opened 2026-08-25
all: smoother image viewer utils handling (fixes #16235)

### LINKED ISSUE #16235: Refactor: Deduplicate showZoomableImage to ImageViewerUtils
Tracking issue for PR #16181.

Fixes #67 by combining the two `showZoomableImage` functions from `MarkdownUtils` and `ImageViewerUtils` into a single, unified method within `ImageViewerUtils`. 

- Migrates the unified dialog to use `DialogZoomableImageBinding` generated from the layout.
- Tracks `activeDialog` to prevent multiple dialogs from overlapping, cleaning it up cleanly `setOnDismissListener`.
- Supports direct Glide string loads for web URLs (Markdown) and local File-based loads (legacy ImageViewer) cleanly.
- Cleans up duplicate imports across both files.

---
*PR created automatically by Jules for task [13927619082286574875](https://jules.google.com/task/13927619082286574875) started by @dogi*

dogi:
Fixes #67 by combining the two `showZoomableImage` functions from `MarkdownUtils` and `ImageViewerUtils` into a single, unified method within `ImageViewerUtils`. 

- Migrates the unified dialog to use `DialogZoomableImageBinding` generated from the layout.
- Tracks `activeDialog` to prevent multiple dialogs from overlapping, cleaning it up cleanly `setOnDismissListener`.
- Supports direct Glide string loads for web URLs (Markdown) and local File-based loads (legacy ImageViewer) cleanly.
- Cleans up duplicate imports across both files.

---
*PR created automatically by Jules for task [13927619082286574875](https://jules.google.com/task/13927619082286574875) started by @dogi*


dogi:
**Review — rating: 82/100**

Collapsing `MarkdownUtils`' private copy into `ImageViewerUtils` is right, and the URL branch is what makes one function able to serve both callers — the old `ImageViewerUtils` always wrapped the path in `File(…)`, which is wrong for the remote URLs markdown images carry. Your `startsWith("http://") || startsWith("https://")` is the correct predicate.

Three things I found that this PR doesn't cover:

**There is a third copy.** `VoicesAdapter.kt:975` has its own `private fun showZoomableImage(context, imageUrl)` with the same layout, the same Glide chain, and the same close-listener, serving four call sites in that file. It differs only in using `ContextCompat.getColor(context, R.color.md_black_1000).toDrawable()` instead of `Color.BLACK.toDrawable()` and calling the shared `isGif(imageUrl)` helper. Folding that one in too is what would actually finish the dedup.

**A Turkish-locale bug rides along in the moved code.** `imagePath.lowercase(Locale.getDefault()).endsWith(".gif")` uses the device locale, and `'I'` lowercases to dotless `'ı'` in Turkish — so `"IMG_1234.GIF"` becomes `"ımg_1234.gıf"` and the `.gif` test fails, loading an animated GIF as a static frame. `endsWith(".gif", ignoreCase = true)` is locale-independent and is the one-word fix while you're already editing that line.

**The static dialog reference is newly introduced here.** `MarkdownUtils` already had `currentZoomDialog`, but `ImageViewerUtils` had none — so `activeDialog` puts an `object`-scoped reference to a `Dialog` (and through it, an Activity) on a path used by `EnterprisesFinancesAdapter:72` and `EnterprisesReportsAdapter:108` that previously had no such retention. `setOnDismissListener` covers normal dismissal but not an Activity destroyed with the dialog still showing.

**Collision:** #16180 is the same refactor. I've marked it `duplicate` in favour of this PR — its `startsWith("http")` also matches any path merely beginning with those four characters, and it leaves `import org.ole.planet.myplanet.R` behind in `MarkdownUtils` where nothing uses it any more.

**Verdict:** the right consolidation with the correct URL predicate, but it stops at two of three copies; ready.

---

Okuro3499 (approved):

PR #16182, opened 2026-08-25
Fix O(N^2) text wrapping performance in submission PDF exporter

dogi:
This commit fixes the text wrapping logic in `SubmissionsRepositoryExporter` that was rebuilding strings for every word, causing an O(N^2) time complexity. It implements a linear O(N) strategy using a `StringBuilder` and a running `lineWidth` while preserving the previous behavior.

---
*PR created automatically by Jules for task [10289388842848476961](https://jules.google.com/task/10289388842848476961) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/repository/SubmissionsRepositoryExporter.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/SubmissionsRepositoryExporter.kt
@@ -17,6 +17,7 @@ import org.ole.planet.myplanet.data.room.dao.QuestionDao
 import org.ole.planet.myplanet.data.room.dao.SubmissionDao
 import org.ole.planet.myplanet.model.Answer
 import org.ole.planet.myplanet.model.MembershipDoc
+import org.json.JSONObject
 import org.ole.planet.myplanet.model.Submission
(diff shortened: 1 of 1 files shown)


dogi:
**Review — rating: 79/100**

**Collision with #16183** — the same linear rewrite of `drawMultilineText`, same `lineWidth` accumulator, same net behaviour. The fix is right in both, and I verified the premise it rests on: summing `measureText(" $word")` incrementally matches measuring the whole line because the split point is always a space, and Arabic — which this app ships a locale for — does not join letters across whitespace, so no word's shaping depends on its predecessor.

Two things decide it against this version. `import org.json.JSONObject` lands in the middle of the `org.ole.planet.myplanet.model` import block, between `MembershipDoc` and `Submission`, where #16183 puts it in correct alphabetical position after `javax.inject.Inject`. And this PR removes only one of the two `e.printStackTrace()` calls — the second export function at line ~238 has an identical swallowed trace that #16183 also cleans up.

**Recommend closing in favour of #16183.** The nit I left there applies here too: `java.lang.StringBuilder()` should just be `StringBuilder()`, since Kotlin default-imports the typealias.

**Verdict:** correct fix, duplicated by #16183, which places the import properly and finishes the `printStackTrace` cleanup; recommending close.

---

PR #16183, opened 2026-08-25
courses: smoother submissions repository pdf exporting (fixes #16236)

### LINKED ISSUE #16236: Fix O(N^2) performance issue in PDF text wrapping
Tracking issue for PR #16183.

Fix O(N^2) performance issue in PDF text wrapping

Optimized `drawMultilineText` in `SubmissionsRepositoryExporter` to avoid measuring the entire accumulated string repeatedly for every word. It now accumulates text in a `StringBuilder`, keeps a running `lineWidth`, and only measures the newly added word (and space).

Also cleaned up by removing a noisy `printStackTrace` inside a catch block that deliberately returned null, and moved `JSONObject` fully-qualified instantiation to a top-level import.

---
*PR created automatically by Jules for task [3997099282363232018](https://jules.google.com/task/3997099282363232018) started by @dogi*

dogi:
Fix O(N^2) performance issue in PDF text wrapping

Optimized `drawMultilineText` in `SubmissionsRepositoryExporter` to avoid measuring the entire accumulated string repeatedly for every word. It now accumulates text in a `StringBuilder`, keeps a running `lineWidth`, and only measures the newly added word (and space).

Also cleaned up by removing a noisy `printStackTrace` inside a catch block that deliberately returned null, and moved `JSONObject` fully-qualified instantiation to a top-level import.

---
*PR created automatically by Jules for task [3997099282363232018](https://jules.google.com/task/3997099282363232018) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6725
-        versionName = "0.67.25"
+        versionCode = 6726
+        versionName = "0.67.26"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/repository/SubmissionsRepositoryExporter.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/SubmissionsRepositoryExporter.kt
@@ -11,6 +11,7 @@ import java.text.SimpleDateFormat
 import java.util.Date
 import java.util.Locale
 import javax.inject.Inject
+import org.json.JSONObject
 import org.ole.planet.myplanet.data.room.dao.AnswerDao
 import org.ole.planet.myplanet.data.room.dao.ExamDao
 import org.ole.planet.myplanet.data.room.dao.QuestionDao
@@ -120,7 +121,6 @@ internal class SubmissionsRepositoryExporter @Inject constructor(
                 document.close()
             }
         } catch (e: Exception) {
-                e.printStackTrace()
                 null
             }
     }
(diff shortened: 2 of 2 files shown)


dogi:
**Review — rating: 86/100**

The O(N²) is real and doubly so: `"$line $word"` rebuilt the whole accumulated line on every word *and* `paint.measureText` re-measured all of it, so a 200-word answer did ~20,000 character measurements and allocated 200 progressively longer strings. Tracking `lineWidth` incrementally and measuring only the increment makes it linear.

The correctness question this turns on is whether `measureText(" $word")` summed incrementally equals `measureText("$line $word")` measured whole — normally it wouldn't, because shaping and kerning can cross a boundary. It's safe here because the split point is always a space: this app ships an Arabic locale, and Arabic letter joining does not cross whitespace, so no word's shaping depends on the previous one. Each word is still measured intact, so intra-word shaping is untouched. Bidi doesn't affect it either, since advance widths are independent of visual ordering.

Behaviour at the edges is preserved: a single word wider than `maxWidth` still gets its own overflowing line rather than looping forever, and the trailing flush is unchanged.

**Collision:** #16182 makes the same fix. I've marked it `duplicate` in favour of this PR, which additionally removes the *second* `e.printStackTrace()` (the one in the other export function's catch, which #16182 leaves), and places `import org.json.JSONObject` in correct alphabetical position rather than in the middle of the `org.ole.planet.myplanet.model` block.

**Nit:** `java.lang.StringBuilder()` should just be `StringBuilder()` — Kotlin default-imports the typealias. Also both versions call `paint.measureText(word)` a second time on the wrap path after already measuring `" $word"`; `lineWidth = wordWidth - spaceWidth` with a hoisted space measurement would avoid it, though at one extra call per wrapped line it hardly matters.

**Verdict:** correct linear rewrite, safe across the locales this app ships, and it cleans up both swallowed stack traces; ready.

---

Okuro3499 (approved):

PR #16184, opened 2026-08-25
Refactor finance type checks to eliminate redundant string allocations

dogi:
Replaced `TextUtils.equals` and `lowercase()` combinations with `equals(..., ignoreCase = true)` in both `EnterprisesFinancesAdapter` and `EnterprisesFinancesFragment` to avoid redundant string allocations. Also removed unused imports.

---
*PR created automatically by Jules for task [15016540432361080664](https://jules.google.com/task/15016540432361080664) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/ui/enterprises/EnterprisesFinancesAdapter.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/enterprises/EnterprisesFinancesAdapter.kt
@@ -4,7 +4,6 @@ import android.content.Context
 import android.graphics.drawable.Drawable
 import android.graphics.drawable.GradientDrawable
 import android.graphics.drawable.LayerDrawable
-import android.text.TextUtils
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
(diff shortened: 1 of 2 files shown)


dogi:
**Review — rating: 80/100**

**Collision with #16185** — the same two comparison rewrites in `EnterprisesFinancesAdapter` and `EnterprisesFinancesFragment`. Both are correct, and both matter more than the titles suggest: `lowercase(Locale.getDefault())` uses Turkish casing on a Turkish device, where `'I'` becomes dotless `'ı'`, so `"DEBIT"` lowercased is `"debıt"` and `"CREDIT"` is `"credıt"` — neither matches, and every debit was displayed and totalled as a credit. `equals(…, ignoreCase = true)` is locale-independent, so this is a real money-on-screen fix.

The difference is the import churn. This PR moves `import java.util.Locale` in `EnterprisesFinancesFragment` — but `Locale` is still needed there at line 47 for `dateFormatter`, so nothing is gained, and the new position (after `java.time.ZoneId`, before `java.time.format.DateTimeFormatter`, with a blank line inserted) is out of alphabetical order where the original was correct. #16185 makes the identical functional change and leaves the import block alone.

**Recommend closing in favour of #16185.**

**Verdict:** correct and worth landing, but duplicated by #16185 without the gratuitous import move; recommending close.

---

PR #16187, opened 2026-08-25
login: smoother url utils handling (fixes #16237)

### LINKED ISSUE #16237: Refactor `String.format` to Kotlin string templates
Tracking issue for PR #16187.

Replaced plain string concatenations via `String.format` with standard Kotlin string templates for better readability and performance in `UrlUtils.kt` and `LoginSyncManager.kt` as suggested in issue 62.

---
*PR created automatically by Jules for task [5601409954831551306](https://jules.google.com/task/5601409954831551306) started by @dogi*

dogi:
Replaced plain string concatenations via `String.format` with standard Kotlin string templates for better readability and performance in `UrlUtils.kt` and `LoginSyncManager.kt` as suggested in issue 62.

---
*PR created automatically by Jules for task [5601409954831551306](https://jules.google.com/task/5601409954831551306) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6726
-        versionName = "0.67.26"
+        versionCode = 6727
+        versionName = "0.67.27"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/services/sync/LoginSyncManager.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/services/sync/LoginSyncManager.kt
@@ -51,7 +51,7 @@ class LoginSyncManager @Inject constructor(
             }
 
             val userUrl = try {
-                String.format("%s/_users/%s", UrlUtils.getUrl(), "org.couchdb.user:$userName")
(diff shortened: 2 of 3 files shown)


dogi:
**Review — rating: 84/100**

Both substitutions are exact — all four placeholders are `%s` with String arguments, so the rendered output is identical, and the nested `"org.couchdb.user:$userName"` interpolation flattens cleanly into the outer template.

The real gain is removing a locale dependency: `String.format` without an explicit `Locale` argument uses `Locale.getDefault()`, which is why Android Lint flags it. It happens to be harmless for `%s`, but it means these call sites carried a locale-sensitive formatter for no reason, and the health-access URL in particular is one you don't want reformatted by a device's locale rules. String templates have no such coupling.

I checked the `try`/`catch` around `userUrl` doesn't become dead: `String.format` could throw `IllegalFormatException` and a template can't, but `UrlUtils.getUrl()` is still inside the block and can still fail, so the catch keeps its purpose.

**Nit:** 37 `String.format` calls remain in `app/src/main/java`. Most pass an explicit `Locale.US` or `Locale.getDefault()` and are fine, but a handful have the same implicit-locale shape this PR is fixing — `TeamFragment.kt:83`, `EventsDetailFragment.kt:202`, `TakeCourseFragment.kt:193`. Those are the ones worth the same treatment.

**Verdict:** exact substitutions that drop an unnecessary locale dependency; ready.

---

Okuro3499 (approved):

PR #16188, opened 2026-08-25
resources: smoother collections view modelling (fixes #16238)

### LINKED ISSUE #16238: Optimize collection child map building in CollectionsViewModel
Tracking issue for PR #16188.

Optimized the construction of `childMap` in `CollectionsViewModel.kt` to avoid intermediate list allocations by replacing `.filter().associate()` with a single-pass `for` loop populating a `LinkedHashMap`. Verified via `CollectionsViewModelTest`.

---
*PR created automatically by Jules for task [12686441124302417797](https://jules.google.com/task/12686441124302417797) started by @dogi*

dogi:
Optimized the construction of `childMap` in `CollectionsViewModel.kt` to avoid intermediate list allocations by replacing `.filter().associate()` with a single-pass `for` loop populating a `LinkedHashMap`. Verified via `CollectionsViewModelTest`.

---
*PR created automatically by Jules for task [12686441124302417797](https://jules.google.com/task/12686441124302417797) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6727
-        versionName = "0.67.27"
+        versionCode = 6728
+        versionName = "0.67.28"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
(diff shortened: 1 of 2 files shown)


dogi:
**Review — rating: 85/100**

The chain really was doing three passes' worth of work for one result: `entries` materialises a set view, `filter` builds an intermediate `ArrayList` of entries, and `associate` allocates a `Pair` per surviving entry before inserting. The explicit loop does one pass with no intermediate list and no pairs.

Semantics are preserved exactly, which is the part worth confirming rather than assuming. `associate` accumulates into a `LinkedHashMap` and lets later duplicates overwrite earlier ones — declaring `LinkedHashMap` explicitly and assigning with `childMap[key] = value` gives the same iteration order and the same last-write-wins behaviour, so downstream code that depends on tag ordering sees no change.

**Nit:** `key.id ?: ""` means two tags with a null `id` collapse onto the same `""` key and the second silently wins. That's pre-existing and unchanged here, but the explicit loop makes it easy to see now — worth either skipping null-id tags or asserting they can't occur.

**Verdict:** faithful rewrite with the same ordering and overwrite semantics, one pass instead of three; ready.

---

Okuro3499 (approved):

PR #16189, opened 2026-08-25
actions: smoother workflow test timing (fixes #16239)

### LINKED ISSUE #16239: fix: safely parse malformed JUnit timing in CI diagnostics
Tracking issue for PR #16189.

Safely parses the `time` attributes in JUnit XML files to handle `NaN`, `inf`, and malformed strings gracefully by defaulting to `0.0` and emitting a warning to stderr, preventing the `test_timing_summary.py` CI step from crashing.

---
*PR created automatically by Jules for task [9958875739411169087](https://jules.google.com/task/9958875739411169087) started by @dogi*

dogi:
Safely parses the `time` attributes in JUnit XML files to handle `NaN`, `inf`, and malformed strings gracefully by defaulting to `0.0` and emitting a warning to stderr, preventing the `test_timing_summary.py` CI step from crashing.

---
*PR created automatically by Jules for task [9958875739411169087](https://jules.google.com/task/9958875739411169087) started by @dogi*

### DIFF
--- a/.github/scripts/test_timing_summary.py
+++ b/.github/scripts/test_timing_summary.py
@@ -14,11 +14,26 @@ has grown past what one CI job should carry gets noticed.
 """
 import argparse
 import glob
+import math
 import os
 import sys
 import xml.etree.ElementTree as ET
 
 TOP_N = 15
+
+
+def _parse_time(time_str: str | None, path: str) -> float:
+    try:
+        if time_str is None:
+            return 0.0
+        val = float(time_str)
+        if not math.isfinite(val) or val < 0.0:
+            print(f"Warning: malformed time '{time_str}' in {os.path.basename(path)}, defaulting to 0.0", file=sys.stderr)
+            return 0.0
+        return val
+    except ValueError:
+        print(f"Warning: malformed time '{time_str}' in {os.path.basename(path)}, defaulting to 0.0", file=sys.stderr)
+        return 0.0
 
 
 def main() -> int:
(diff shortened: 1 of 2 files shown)


dogi:
**Review — rating: 87/100**

This prevents an actual job failure, which is worth stating because "diagnostics script" makes it sound cosmetic. The step at `test.yml:81` declares `shell: bash`, so GitHub runs it with `-eo pipefail` — meaning `python3 … | tee` propagates the Python exit code rather than `tee`'s. Combined with `if: always()`, a single malformed `time` attribute in one JUnit XML would raise `ValueError`, fail the summarise step, and turn an otherwise-green test run red. Degrading to `0.0` with a warning is the right call for a reporting script.

The guard is more thorough than the title suggests, too: `float()` happily accepts `"nan"` and `"inf"`, and either would poison the `total` and wreck the sort ordering (`NaN` comparisons are never true, so `sort(reverse=True)` would produce garbage). `math.isfinite` plus the `< 0.0` check catches both. Warnings going to `stderr` keeps them out of the piped `GITHUB_STEP_SUMMARY` content.

**Nit:** the old `float(root.get("time") or 0)` treated an empty `time=""` as `0` silently via the `or` short-circuit; `_parse_time("")` now takes the `ValueError` path and prints a warning for what is arguably a legitimate case. Returning `0.0` early for a falsy `time_str` would keep the log clean.

**Verdict:** correct hardening of a step that can currently fail a green build; ready.

---

Okuro3499 (approved):

PR #16190, opened 2026-08-25
actions: bump `actions/upload-artifact` to 7 (fixes #16240)

### LINKED ISSUE #16240: ci: bump upload-artifact to v7 in test workflow
Tracking issue for PR #16190.

ci: bump upload-artifact to v7 in test workflow

Updates the `actions/upload-artifact` step in `.github/workflows/test.yml` from `v4` to `v7` to match the version used in `release.yml`. This addresses Dependabot's suggested update while retaining existing configuration values.

---
*PR created automatically by Jules for task [13871562942563447458](https://jules.google.com/task/13871562942563447458) started by @dogi*

## Summary by CodeRabbit

* **Chores**
  * Updated the test report artifact upload process to use a newer version of the upload tool.

dogi:
ci: bump upload-artifact to v7 in test workflow

Updates the `actions/upload-artifact` step in `.github/workflows/test.yml` from `v4` to `v7` to match the version used in `release.yml`. This addresses Dependabot's suggested update while retaining existing configuration values.

---
*PR created automatically by Jules for task [13871562942563447458](https://jules.google.com/task/13871562942563447458) started by @dogi*

## Summary by CodeRabbit

* **Chores**
  * Updated the test report artifact upload process to use a newer version of the upload tool.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -96,6 +96,7 @@ jobs:
         with:
           name: test-reports-${{ matrix.build }}-shard-${{ matrix.shard }}
           overwrite: true
+          if-no-files-found: error
           path: |
             app/build/reports/tests/
             app/build/test-results/
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -92,7 +92,7 @@ jobs:
 
       - name: upload test reports
         if: failure()
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v7
         with:
           name: test-reports-${{ matrix.build }}-shard-${{ matrix.shard }}
           overwrite: true
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6729
-        versionName = "0.67.29"
+        versionCode = 6730
+        versionName = "0.67.30"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }


dogi:
**Review — rating: 68/100**

The thing to know before merging this is that **the green checks on this PR provide no evidence the change works.** The step being edited is guarded by `if: failure()`, and all four checks passed — so `actions/upload-artifact@v7` was never resolved or executed in any run on this branch. If the tag doesn't exist, the first time anyone finds out is on a *real* test failure, when the run that was already red also loses its reports. That's the worst moment to discover it.

I can't verify the tag from this session — GitHub API access here is scoped to this repository, so I have no way to list `actions/upload-artifact`'s releases and I'd rather say so than assert a version from memory. Two things I'd want in the PR description before this lands: a link to the `v7` release, and a note on what changed across `v5` and `v6`, since this skips two majors at once and `upload-artifact`'s major bumps have historically carried breaking changes to artifact semantics rather than being routine.

There's a cheap way to get real evidence: temporarily flip the guard to `if: always()` on a scratch branch, confirm the upload step resolves and produces an artifact, then restore `if: failure()`. One run settles it.

**Nit:** while you're in this step, `if-no-files-found` is unset and defaults to `warn`, so a missing report directory passes silently — worth setting explicitly given the step only runs when something has already gone wrong.

**Verdict:** plausible but unverified, and structurally unverifiable by this PR's own CI; needs evidence the tag resolves and a note on the skipped majors.

---

dogi:
**Re-review — Rating: 89/100** (was 68)

The PR description settles what I couldn't check from my previous session, and I've now verified it inside the repo: **`.github/workflows/release.yml:128` already uses `actions/upload-artifact@v7` on master.** The tag exists, it's already in use here, and `release.yml` runs on every master push — so this isn't an unverifiable version jump, it's `test.yml` catching up with what the release pipeline has been running all along. That also disposes of my concern about skipping two majors: the semantics of v5–v7 are already proven against this repo's artifacts.

`if-no-files-found: error` addresses the silent-default point directly. One tradeoff worth knowing rather than guessing at: this step is guarded by `if: failure()`, so it fires precisely when something has already gone wrong — including a *build* failure before any test runs, where `app/build/reports/tests/` and `app/build/test-results/` won't exist. Under `error` that turns into a second red step on an already-red run. That's arguably the right signal (a failed test job should have reports), but if the noise proves annoying, `warn` is the version that stays quiet in that case.

The `if: failure()` guard still means this PR's own green checks never executed the changed step — but with `release.yml` as the existing proof, that no longer matters.

**Verdict:** verified against `release.yml` on master, and the missing-files default is now explicit; ready.

---

Okuro3499 (approved):

PR #16193, opened 2026-08-25
perf(utils): Simplify version-string parsing in compareVersions

dogi:
Fixes issue 45. Optimizes `VersionUtils.compareVersions` to reduce list and string allocations.

---
*PR created automatically by Jules for task [9759987720868590022](https://jules.google.com/task/9759987720868590022) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/utils/VersionUtils.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/utils/VersionUtils.kt
@@ -6,7 +6,19 @@ import android.os.Build
 import android.provider.Settings
 import androidx.core.content.pm.PackageInfoCompat.getLongVersionCode
 
+import java.util.Collections
+import java.util.LinkedHashMap
+
 object VersionUtils {
+
+    private val versionCache = Collections.synchronizedMap(
+        object : LinkedHashMap<String, List<Int>>(16, 0.75f, true) {
+            override fun removeEldestEntry(eldest: MutableMap.MutableEntry<String, List<Int>>): Boolean {
+                return size > 50
+            }
+        }
+    )
+
     fun getVersionCode(context: Context): Int {
         try {
             val pInfo = context.packageManager.getPackageInfo(context.packageName, 0)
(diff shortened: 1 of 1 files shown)


dogi:
**Review — rating: 66/100**

I traced the rewrite against the original and I believe it's correct, including the parts that look most fragile. The `i1 <= e1` bound is right (after the last segment `i1` becomes `dot + 1 == e1 + 1`), the `-lite` suffix is still stripped from `version1` only, latching `diff` at the first difference and continuing matches the old early return, and the `count1.compareTo(count2)` tiebreak reproduces the old `parts1.size.compareTo(parts2.size)` because the padded side is excluded from the comparison by the `hasPart1 && hasPart2` guard. The malformed cases (`""`, `"v"`, `"1.2."`, a `-lite` suffix on `version2`) throw `NumberFormatException` in both versions. `VersionUtilsTest` has 21 `compareVersions` assertions and they pass, which is real corroboration on top of my reading.

The problem is the trade, not the correctness. **`substring` allocates a `String` per segment**, so the rewrite doesn't remove the per-part allocations — it only removes two `ArrayList`s and the `map` intermediate. That's a handful of short-lived objects, on a function called when checking for an app update: once per launch, not per frame. In exchange, four self-evident lines become thirty-five of manual index arithmetic with two cursors, two counters, two end offsets, and a latched result — the shape of code where the *next* edit introduces an off-by-one that the current tests may not catch.

If there's a profile showing this on a hot path I've missed, say so in the description and I'll withdraw this. Otherwise the honest options are to revert to the readable form, or — if allocation genuinely matters — keep the readable parse and cache the parsed result, which would beat both versions.

**Verdict:** correct as far as I can trace and as far as the existing suite exercises, but it buys a negligible allocation saving on a cold path with a large maintainability cost; needs a justification or a revert.

---

dogi:
**Re-review — Rating: 72/100** (was 66)

Reverting the index arithmetic is the important half, and it fully resolves my main objection. `compareVersions` is back to the original four readable lines, so the off-by-one surface I was worried about is gone entirely — the parse logic is now byte-for-byte what master has and what `VersionUtilsTest`'s 21 assertions were written against.

The cache itself is implemented correctly. `getOrPut` on a `Collections.synchronizedMap` isn't atomic, but a double compute is harmless here because the parse is deterministic and idempotent; the `accessOrder = true` reordering that `get()` performs is protected by the synchronized wrapper; and if the compute lambda throws on malformed input (`"1.2.3-lite"` passed as `version2`, say) `getOrPut` never reaches its `put`, so a bad key can't poison the map. Same shape as #16127, which I passed.

**But now I can measure whether the cache earns its keep, and it doesn't.** The call chain is exactly one deep at every hop: `compareVersions` has a single caller, `isVersionAllowed` (`VersionUtils.kt:40`); that has a single caller, `ConfigurationsRepositoryImpl:300`, inside `checkVersion`; and `checkVersion` is invoked from `SyncActivity:223`, `LoginActivity:183` and `AutoSyncWorker:70` — each **behind an `apiInterface.checkVersion` network round-trip**. So this runs a handful of times per session, each time already gated on HTTP. The cache saves two small `ArrayList`s per invocation and costs a synchronized map, an anonymous `LinkedHashMap` subclass, two imports and nine lines of declaration. Against master's two lines, that's strictly more code for a saving no profiler will ever show.

My honest recommendation is to **close this** rather than iterate again. The valuable outcome — not landing the index-arithmetic rewrite — is already achieved by master being untouched.

**Nit:** `import java.util.LinkedHashMap` is redundant, since Kotlin default-imports the typealias — the same nit #16127 just removed, reintroduced here. There's also a stray blank line separating the two new imports from the block above.

**Verdict:** correct and no longer fragile, but the measurement now shows the cache can't pay for itself on a path that runs a few times per session behind a network call; closing is the cleanest resolution.

---

PR #16198, opened 2026-08-25
resources: smoother collections parent tagging (fixes #16246)

### LINKED ISSUE #16246: fix: drop the extra list copy in CollectionsFragment.onParentTagClicked
Tracking issue for PR #16198.

Removed the extra list copy in `CollectionsFragment.onParentTagClicked` by removing `.toList()` when submitting the newly created mutable list to the adapter.

---
*PR created automatically by Jules for task [9849415319212219547](https://jules.google.com/task/9849415319212219547) started by @dogi*

dogi:
Removed the extra list copy in `CollectionsFragment.onParentTagClicked` by removing `.toList()` when submitting the newly created mutable list to the adapter.

---
*PR created automatically by Jules for task [9849415319212219547](https://jules.google.com/task/9849415319212219547) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6734
-        versionName = "0.67.34"
+        versionCode = 6735
+        versionName = "0.67.35"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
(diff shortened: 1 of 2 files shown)


dogi:
**Review — rating: 87/100**

Safe, and it's a consistency fix as much as an allocation one: of the five `submitList` calls in this file, line 156 was the **only** one wrapping the field in `toList()` — lines 75, 119, 166 and 176 already pass `currentTagDataList` directly. So this removes an outlier rather than lowering a guard.

The reason it's safe is worth stating, because `ListAdapter.submitList` retains the reference it's handed and diffs the *next* list against it — so passing a `MutableList` that later gets mutated in place is a genuine way to corrupt diffing. That can't happen here: every one of the five sites **reassigns** `currentTagDataList = buildTagDataList(…).toMutableList()` immediately before submitting, so the adapter never holds a list this fragment goes on to modify. The only other access, line 127, is a read-only `forEach`.

That does mean the safety now rests entirely on that convention at all five sites rather than four. A single future `currentTagDataList.add(…)` between submissions would break diffing everywhere, not just here — declaring the field `List<TagData>` instead of `MutableList<TagData>` would make that impossible to write by accident.

**Verdict:** removes the file's lone inconsistent copy, and reassignment-not-mutation makes it sound; ready.

---

Okuro3499 (approved):

PR #16202, opened 2026-08-25
actions: smoother gradle configuring (fixes #16250)

### LINKED ISSUE #16250: Document Jetifier requirement and remove obsolete newDsl flag
Tracking issue for PR #16202.

## Summary
Updated Gradle configuration to clarify the necessity of AndroidX Jetifier and removed an obsolete build flag that is no longer supported by the Android Gradle Plugin.

## Changes
- **Added explanatory comment** for `android.enableJetifier=true`: Documents that this flag is load-bearing because transitive dependencies (material-dialogs, Circular-Progress-View, materialprogressbar) drag in pre-AndroidX support library artifacts (com.android.support:*:28.0.0), and pre-AndroidX AAR files in app/libs/ require Jetifier to prevent duplicate class errors and manifest merger failures.
- **Removed `android.newDsl=false`**: This flag is no longer recognized by Android Gradle Plugin 9.3.1 and serves no purpose in the current build configuration.

## Implementation Details
The comment clarifies the architectural constraint that prevents removal of Jetifier despite the app being fully AndroidX-based (Kotlin 100%, Room 2.8.4, etc.). This prevents future maintainers from accidentally removing the flag and breaking the build.

https://claude.ai/code/session_018Peg1CDT9Aoe2AktWY5s5B

dogi:
## Summary
Updated Gradle configuration to clarify the necessity of AndroidX Jetifier and removed an obsolete build flag that is no longer supported by the Android Gradle Plugin.

## Changes
- **Added explanatory comment** for `android.enableJetifier=true`: Documents that this flag is load-bearing because transitive dependencies (material-dialogs, Circular-Progress-View, materialprogressbar) drag in pre-AndroidX support library artifacts (com.android.support:*:28.0.0), and pre-AndroidX AAR files in app/libs/ require Jetifier to prevent duplicate class errors and manifest merger failures.
- **Removed `android.newDsl=false`**: This flag is no longer recognized by Android Gradle Plugin 9.3.1 and serves no purpose in the current build configuration.

## Implementation Details
The comment clarifies the architectural constraint that prevents removal of Jetifier despite the app being fully AndroidX-based (Kotlin 100%, Room 2.8.4, etc.). This prevents future maintainers from accidentally removing the flag and breaking the build.

https://claude.ai/code/session_018Peg1CDT9Aoe2AktWY5s5B

## Summary by CodeRabbit

- **Chores**
  - Updated Android build configuration by removing an obsolete setting.
  - Added documentation clarifying configuration requirements for compatibility with existing build dependencies.
  - No changes to application features or user-facing behavior.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6738
-        versionName = "0.67.38"
+        versionCode = 6739
+        versionName = "0.67.39"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/gradle.properties
+++ b/gradle.properties
@@ -22,9 +22,7 @@ org.gradle.configureondemand=false
 org.gradle.vfs.watch=true
 org.gradle.warning.mode=all
 android.useAndroidX=true
-# Load-bearing: material-dialogs, Circular-Progress-View and materialprogressbar
-# drag in com.android.support:*:28.0.0 transitively, and app/libs/*.aar are
-# pre-AndroidX. Removing this fails checkDuplicateClasses + manifest merger.
+# below can only be removed with material-dialogs, Circular-Progress-View, materialprogressbar and app/libs/*.aar
 android.enableJetifier=true
 android.nonTransitiveRClass=false
 android.nonFinalResIds=true
--- a/gradle.properties
+++ b/gradle.properties
@@ -22,10 +22,12 @@ org.gradle.configureondemand=false
 org.gradle.vfs.watch=true
 org.gradle.warning.mode=all
 android.useAndroidX=true
+# Load-bearing: material-dialogs, Circular-Progress-View and materialprogressbar
+# drag in com.android.support:*:28.0.0 transitively, and app/libs/*.aar are
+# pre-AndroidX. Removing this fails checkDuplicateClasses + manifest merger.
 android.enableJetifier=true
 android.nonTransitiveRClass=false
 android.nonFinalResIds=true
-android.newDsl=false
 
 PLANET_LEARNING_URL=planet.learning.ole.org
 PLANET_LEARNING_PIN=1983


dogi:
One correction to the description, plus the verification evidence.

### `android.newDsl` is recognized by AGP 9.3.1

The summary says the flag "is no longer recognized by Android Gradle Plugin 9.3.1 and serves no purpose." That isn't right, and the difference matters for review. From AGP 9.3.1's own `BooleanOption` table:

```
USE_NEW_DSL      android.newDsl
  FeatureStage.SoftlyEnforced(VERSION_10_0), default = true

ENABLE_JETIFIER  android.enableJetifier
  ApiStage.Deprecated(VERSION_10_0), default = false
```

So `android.newDsl=false` was recognized and honored — it actively pinned the project to the legacy DSL, and warned while doing so. Deleting it is therefore a real (benign) behavior change, not a no-op: the build flips to the new DSL default. I verified `app/build.gradle`'s legacy syntax still configures cleanly under it — `lintOptions`, `packagingOptions`, `flavorDimensions "mode"`, and the space-call setters (`testInstrumentationRunner "…"`, `doNotStrip "…"`) are all still accepted, with no deprecation warning left behind.

Both flags are removed in AGP 10.0, so this is worth knowing before that upgrade rather than treating the line as dead text.

### Why Jetifier had to stay

The originating ticket asked for both lines deleted, on the basis that no `com.android.support` dependency exists in `gradle/libs.versions.toml` or `app/build.gradle`. That grep is accurate but cannot see two things, and deleting the flag fails the build outright:

```
:app:checkDefaultDebugDuplicateClasses FAILED
  Duplicate class android.support.v4.app.INotificationSideChannel found in
  androidx.core:core:1.19.0 and com.android.support:support-compat:28.0.0
  (~28 duplicates)
:app:processDefaultDebugMainManifest FAILED  (manifest merger)
```

1. **Transitive Maven modules.** `dependencyInsight` on `defaultDebugRuntimeClasspath` shows `com.android.support:appcompat-v7:28.0.0` arriving via `material-dialogs:commons/core:0.9.6.0`, `Circular-Progress-View:0.1.2`, and `materialprogressbar:1.4.2`, pulling `recyclerview-v7`, `support-fragment`, `support-core-ui`, `animated-vector-drawable` with it. The aliases are `libs.material.dialogs.commons` etc., so no text search of the catalog finds `com.android.support`.
2. **`app/libs/*.aar`**, wired in by `implementation fileTree(...)`. `flexbox-1.0.0.aar` carries 95 `android.support` refs and zero androidx; `ChipCloud-3.0.5.aar` carries 6. `FlexboxLayout` calls `ViewCompat`; ChipCloud's `ConfigHelper`/`ToggleChip` call `ContextCompat`, `DrawableCompat`, `RoundedBitmapDrawable` — all reachable from `BaseDashboardFragment`, `ResourcesFragment`, `HealthExaminationActivity`, and the achievement screens.

Jetifier currently rewrites 662 artifacts, so it is doing real work rather than dead work on every clean build.

### Verification

Run locally on AGP 9.3.1 with SDK platform 37.0 / build-tools 37.0.0:

- `assembleDefaultDebug` — passes, APK produced.
- `testDefaultDebugUnitTest` — 1286 tests, **1 failure**: `DispatcherProviderIntegrationTest.testServiceUsesDispatcher`, with `kotlinx.coroutines.test.UncaughtExceptionsBeforeTest`.

That failure is **pre-existing and unrelated to this diff**. It passes when run in isolation, and a control run of the full suite on the unmodified `gradle.properties` fails identically (same class, 1286 tests / 1 failure). It is cross-test pollution inside a shared test JVM — stray `Dispatchers.IO` exceptions and an `IllegalStateException: The component was not created. Check that you have added the HiltAndroidRule` leaking from a sibling test in the same fork. Since CI shards into two jobs, the fork grouping differs there and it may or may not reproduce; worth filing separately either way.

### Follow-up

The `android.enableJetifier` deprecation warning the ticket wanted silenced is real, and AGP 10.0 removes the flag. Clearing it means replacing the support-library dependencies above — `material-dialogs` has an AndroidX 3.x line, and the two local AARs need modern equivalents (Material `ChipGroup` for ChipCloud). That is a dependency migration rather than a properties edit, so it is deliberately not in this PR.

---

dogi:
**Review — rating: 82/100**

Documenting a load-bearing flag is the right move and this one earns it. `android.enableJetifier=true` rewrites every dependency on every build, so it's a permanent tax and an obvious target for someone tidying `gradle.properties` — exactly the kind of removal that looks safe, passes review, and then fails the build in a way nobody connects back. A comment at the flag is where that warning belongs, and `CLAUDE.md` says nothing about it.

The core claim holds up. `app/libs/` contains `ChipCloud-3.0.5.aar` and `flexbox-1.0.0.aar`, both genuinely pre-AndroidX, and `material-dialogs:commons` and `Circular-Progress-View` are both declared in `libs.versions.toml` (lines 100 and 107). Naming `checkDuplicateClasses` and the manifest merger as the failure modes is right — those are precisely what breaks when support and AndroidX classes coexist.

**Nit — one of the three named libraries can't be found from the comment.** There's no `materialprogressbar` anywhere in `libs.versions.toml` or `app/build.gradle`; the only progress dependency is `circular-progress-view`. If it arrives transitively (MaterialDrawer has historically pulled `me.zhanghai.android.materialprogressbar`, and it's declared here with `transitive = true`), say so — "via materialdrawer" makes the note verifiable. If it's left over from an earlier iteration, drop it. As written, the first person to check will grep, find nothing, and start doubting the rest of the comment, which defeats its whole purpose. While you're there, naming `ChipCloud-3.0.5.aar` and `flexbox-1.0.0.aar` explicitly rather than `app/libs/*.aar` would tell a future reader exactly which two files need replacing before Jetifier can go.

On `android.newDsl=false`: it appears nowhere else in the repo, and both flavors build green without it. AGP hard-fails on *removed* options rather than warning, so if it were one of those master would already be red — meaning this is either still recognised or silently ignored, and dropping it is inert either way. Confirming against a `--warning-mode all` log diff would make that airtight, but nothing here suggests otherwise.

**Verdict:** genuinely useful documentation of a flag that would otherwise get "cleaned up", plus an inert deletion; ready once the third library is corrected or attributed.

---

Okuro3499 (approved):

PR #16203, opened 2026-08-25
refactor: walk version parts with a cursor in compareVersions

dogi:
## Summary

A readability/allocation refactor of `VersionUtils.compareVersions`, with **no behaviour change**. The tests that were here have moved to [#16217](https://github.com/open-learning-exchange/myplanet/pull/16217), which stands on its own and passes against master unmodified — this PR is now the implementation only, rebased on current master.

### Correcting the previous description

The auto-generated title and body claimed this "validates format" and makes the function "stricter about format violations". That was wrong, and I should have corrected it when the PR was opened. Master's `split(".").map { it.toInt() }` already parsed every part eagerly and already threw on every malformed one. This change **preserves** that; it adds no validation and no new format check.

## What actually changes

`removeSuffix`/`removePrefix`/`split`/`map { it.toInt() }` becomes a small private `VersionParts` helper that walks the dot-separated parts with a cursor. What that buys:

- Two intermediate `ArrayList`s and their boxed `Int`s go away.
- The per-part `substring` **stays** — it's what keeps `toInt()`'s exact semantics (unicode digits, `+` sign, overflow), so this is not an allocation-free parse.

Cost: roughly 30 lines against master's 8, plus one invariant a reader has to hold (`cursor <= end` means a part is still pending; an empty string still has one).

## Why the value is modest

`compareVersions` → `isVersionAllowed` → `ConfigurationsRepositoryImpl:300`, reached from three sites, each behind an `apiInterface.checkVersion` round-trip. So this saves a handful of small objects a few times per session, already gated on HTTP. It is a readability and tidiness argument, not a performance one — reviewers should weigh it on that basis, and closing this in favour of [#16217](https://github.com/open-learning-exchange/myplanet/pull/16217) alone is a reasonable call.

## Behaviour preserved

- Same signature; `v` prefix stripped from both arguments, `-lite` from the first only.
- First differing part decides; when the shared parts match, the shorter version sorts lower. The "one side exhausted" branch *is* that size comparison, which is why latching the first non-zero result reproduces master exactly.
- Both strings are still consumed to the end after the order is decided, so a malformed part anywhere still throws `NumberFormatException`.

## Verification

`testDefaultDebugUnitTest` can't run in this container (no Android SDK). Instead I compiled master's implementation and this one side by side and ran a differential test over 39×39 = 1521 input pairs — valid versions, `v`/`-lite` variants, empty strings, double and trailing dots, `2147483648` overflow, whitespace, `+`/`-` signs, Arabic-Indic digits. **Zero mismatches** in result sign or exception type. The only difference is which malformed part is named in the exception message when *both* strings are malformed, since parsing now interleaves. CI runs the real suite.

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/utils/VersionUtils.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/utils/VersionUtils.kt
@@ -7,6 +7,9 @@ import android.provider.Settings
 import androidx.core.content.pm.PackageInfoCompat.getLongVersionCode
 
 object VersionUtils {
+    private const val VERSION_PREFIX = "v"
+    private const val LITE_SUFFIX = "-lite"
+
     fun getVersionCode(context: Context): Int {
         try {
             val pInfo = context.packageManager.getPackageInfo(context.packageName, 0)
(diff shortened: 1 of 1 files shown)


dogi:
**Review — rating: 78/100**

This is a much better version of the idea than #16193 was, and the craft shows. `VersionParts` encapsulates the cursor walk instead of scattering four index variables through the comparison, the `hasNext` doc comment states the non-obvious invariant ("a cursor past `end` means the last part has been consumed; an empty string still has one"), and the `dot in cursor until end` guard is the piece #16193 was missing — it's what makes `"1.0.0-lite"` in the *second* argument still fail rather than silently parsing across the suffix.

I traced the equivalence and it holds, including the parts that usually break. The latch-on-first-difference plus ±1-when-one-side-is-exhausted reproduces master's `first difference, else size comparison` exactly, because the exhausted-side branch *is* the size comparison. The deliberate "keep consuming both cursors after the result is decided" preserves master's eager `.map { it.toInt() }`, so a malformed part anywhere still throws — and you tested precisely that, twice. I also checked `stop < cursor` is unreachable: any string that both starts with `"v"` and ends with `"-lite"` is at least 6 characters, so `end >= 1 = cursor`, and `substring` can't throw `StringIndexOutOfBounds` in place of `NumberFormatException`.

**But the five tests are the deliverable here, and they don't need the rewrite.** Walk them against master's implementation: `"1.x"` and `"1.0.x"` both die in `.map { it.toInt() }`; `""` splits to `[""]` and dies the same way; `"1.0.0-lite"` as `version2` splits to `["1","0","0-lite"]` and dies on the last part; and the `insufficient_version_parts` assertions hold via `parts1.size.compareTo(parts2.size)`. All five pass on master unchanged. That makes them excellent characterization tests — they're what proves the refactor is safe — but it also means they can land on their own, today, and lock this behaviour in regardless of what happens to the implementation.

**The title claims something the diff doesn't do.** "parse all parts and validate format" reads as new validation, but master already parsed every part eagerly and threw on malformed ones; this preserves that rather than adding it, and no new format check exists. What the rewrite actually buys is dropping two intermediate `ArrayList`s and their boxed `Int`s — `next()` still calls `substring`, so the per-part String allocations remain. On the path I traced last round (`compareVersions` → `isVersionAllowed` → `ConfigurationsRepositoryImpl:300` → `checkVersion`, called from three sites each behind an `apiInterface.checkVersion` round-trip) that's a handful of small objects a few times per session, already gated on HTTP, for roughly thirty lines against master's two.

**My ask is small:** split the tests into their own PR and merge that — they're a genuine gain and I'd sign off immediately. Then either drop the implementation change, or keep it and re-title it honestly as a readability refactor, with a sentence on why the no-intermediate-lists form is worth the extra surface here.

**Verdict:** correct, well-documented and properly characterized — but the tests carry all the value and pass without it, and the title promises validation the diff doesn't add.

---

dogi:
Closing in favour of [#16217](https://github.com/open-learning-exchange/myplanet/pull/16217), which carries the four characterization tests and passes against master unmodified.

The review was right on both counts. The tests were the deliverable and they didn't need the rewrite; and the original title claimed validation this diff never added — master's eager `map { it.toInt() }` already threw on every malformed part, so this preserved that behaviour rather than introducing it.

On the refactor itself: it is equivalent (verified over 1521 input pairs, zero mismatches in result sign or exception type) and CI was green, but what it actually buys is two intermediate lists and their boxed `Int`s, on a path that runs a few times per session behind an `apiInterface.checkVersion` round-trip. The per-part `substring` stays either way. That is not worth ~30 lines and a cursor invariant against master's 8, so it goes.

The behaviour it was written to protect is now pinned down by [#16217](https://github.com/open-learning-exchange/myplanet/pull/16217) regardless.

---

PR #16217, opened 2026-08-25
all: smoother version utils testing (fixes #16251)

### LINKED ISSUE #16251: test: characterize VersionUtils.compareVersions edge cases
Tracking issue for PR #16217.

## Summary

Adds four characterization tests for `VersionUtils.compareVersions` and refreshes one stale comment. **No production code changes** — every assertion passes against the current implementation, unmodified.

Split out of [#16203](https://github.com/open-learning-exchange/myplanet/pull/16203) on review feedback: the tests are valuable on their own and shouldn't be gated on whether that refactor lands.

## What is pinned down

| Case | Behaviour |
|---|---|
| `compareVersions("2.0", "1.x")` | Throws `NumberFormatException`. Both strings are parsed in full, so a malformed part still throws even though the leading parts already decide the order. |
| `compareVersions("1.0", "1.0.x")` | Throws. A malformed trailing part of the *longer* version is reached, past the parts that are actually compared. |
| `compareVersions("", "1.0.0")` | Throws. An empty string splits to `[""]`, not to an empty list. |
| `compareVersions("1.0.0", "1.0.0-lite")` | Throws. `-lite` is stripped from the first argument only, so in the second it is part of a numeric segment (`"0-lite"`). Callers pass the server's `minapk` there, which never carries the flavor suffix. |

The asymmetric `-lite` handling is pre-existing behaviour and this only documents it; it isn't an endorsement, and the test comment says why the asymmetry is harmless for the current caller.

## Also

The comment on `compareVersions_should_not_throw_on_insufficient_version_parts` referenced `kotlin.math.min(parts1.size, parts2.size)` — an implementation detail rather than the contract. Reworded to describe the behaviour: only the shared parts are compared, and the shorter version sorts lower when they match.

## Verification

`testDefaultDebugUnitTest` can't run in this container (no Android SDK), so I compiled the current `compareVersions` verbatim with the Kotlin compiler and executed all four new cases plus the two reworded assertions against it directly — all pass. CI runs the real suite on this PR.

---
_Generated by [Claude Code](https://claude.ai/code/session_01UUarreQQN1rGLdRkh9mCQe)_

dogi:
## Summary

Adds four characterization tests for `VersionUtils.compareVersions` and refreshes one stale comment. **No production code changes** — every assertion passes against the current implementation, unmodified.

Split out of [#16203](https://github.com/open-learning-exchange/myplanet/pull/16203) on review feedback: the tests are valuable on their own and shouldn't be gated on whether that refactor lands.

## What is pinned down

| Case | Behaviour |
|---|---|
| `compareVersions("2.0", "1.x")` | Throws `NumberFormatException`. Both strings are parsed in full, so a malformed part still throws even though the leading parts already decide the order. |
| `compareVersions("1.0", "1.0.x")` | Throws. A malformed trailing part of the *longer* version is reached, past the parts that are actually compared. |
| `compareVersions("", "1.0.0")` | Throws. An empty string splits to `[""]`, not to an empty list. |
| `compareVersions("1.0.0", "1.0.0-lite")` | Throws. `-lite` is stripped from the first argument only, so in the second it is part of a numeric segment (`"0-lite"`). Callers pass the server's `minapk` there, which never carries the flavor suffix. |

The asymmetric `-lite` handling is pre-existing behaviour and this only documents it; it isn't an endorsement, and the test comment says why the asymmetry is harmless for the current caller.

## Also

The comment on `compareVersions_should_not_throw_on_insufficient_version_parts` referenced `kotlin.math.min(parts1.size, parts2.size)` — an implementation detail rather than the contract. Reworded to describe the behaviour: only the shared parts are compared, and the shorter version sorts lower when they match.

## Verification

`testDefaultDebugUnitTest` can't run in this container (no Android SDK), so I compiled the current `compareVersions` verbatim with the Kotlin compiler and executed all four new cases plus the two reworded assertions against it directly — all pass. CI runs the real suite on this PR.

---
_Generated by [Claude Code](https://claude.ai/code/session_01UUarreQQN1rGLdRkh9mCQe)_


dogi:
**Review — rating: 90/100**

This is the split I hoped for on #16203, and separating it proves the point I could only argue analytically there: the diff touches one test file and nothing under `src/main`, so all four green checks are these tests passing against **unmodified master**. That's the empirical confirmation that they characterize existing behaviour rather than describing a change.

The first test is the one that earns the PR on its own. `compareVersions("2.0", "1.x")` throws today only because master parses eagerly — `.map { it.toInt() }` materialises both lists before a single comparison happens. Any future rewrite that walks the parts lazily and returns as soon as the order is decided would silently stop throwing here, and nothing in the suite would notice. Pinning it means the next attempt at that refactor (#16203 included) gets reviewed against a real guard rail instead of against reasoning. The comment says exactly that, in behavioural terms.

The lite-suffix test documents something that otherwise reads as a bug. `removeSuffix("-lite")` applying to `version1` but not `version2` looks like an oversight until you know the second argument is the server's `minapk`, which never carries a flavour suffix — and the comment records that rather than leaving the next reader to reverse-engineer it.

Rewriting the comment on `compareVersions_should_not_throw_on_insufficient_version_parts` is a real improvement too, not just churn: it moves from describing the implementation ("uses `kotlin.math.min(parts1.size, parts2.size)`") to describing the contract ("the shorter version sorts lower"). The old wording would have gone stale the moment anyone touched the function; the new one survives. The assertions are untouched. All four new tests also use `@Test(expected = NumberFormatException::class)`, matching the idiom already at line 51.

**Worth knowing, not a defect here:** I checked where the throw lands. `ConfigurationsRepositoryImpl.checkConfigurationUrl` wraps the call in a broad `catch (e: Exception) { e.printStackTrace(); UrlCheckResult.Failure(currentUrl) }`, so a malformed server `minapk` doesn't crash — it degrades to a generic URL-check failure with only a stack trace in the log, indistinguishable from the server being unreachable. That's pre-existing, and these tests are precisely where a future decision to surface it better would get recorded.

**Verdict:** tests-only, verified green against untouched production code, and the eager-parsing guard makes any later rewrite of this function safe to evaluate; ready.

---

Okuro3499 (approved):

PR #16254, opened 2026-08-25
life: smoother health examining (fixes #1939)

### LINKED ISSUE #1939: Prevent health record with empty spaces
Health examination prevent  spaces

Okuro3499:
fixes #1939

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection of completed health examination information, including fields containing whitespace.
  * Removed an unused dependency without changing the user-facing workflow.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6743
-        versionName = "0.67.43"
+        versionCode = 6744
+        versionName = "0.67.44"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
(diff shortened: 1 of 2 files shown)


dogi:
**Review — rating: 84/100**

Worth being explicit that this is not the like-for-like swap the title suggests: `!TextUtils.isEmpty(s)` is `s.isNotEmpty()`, so moving to `isNotBlank()` also reclassifies whitespace-only fields as "no info". That's the right call — a form where every box holds a space genuinely has nothing in it — but it changes stored data, not just style. `hasInfo` feeds `examination?.isHasInfo`, which `HealthExamination.serialize` writes out as the `hasInfo` property, so the flag the Planet server sees for such a record flips from `true` to `false`. Nothing in the app reads `isHasInfo` back, so the blast radius is server-side consumers only.

Dropping the `android.text.TextUtils` import cleans the file completely — CI compiling with the import gone proves `hasInfo` was its only use here, which is the sort of thing worth confirming rather than assuming when an import disappears. That's 1 of the 37 files still importing it, so #1939 has a way to go, but slicing it per file is the right shape.

**Nit:** `"${binding.etAllergies.text}"` builds a string just to test it, and if `getText()` were ever null the template yields the literal `"null"`, which `isNotBlank()` reports as true — the same latent wart the old code had. `Editable` is a `CharSequence`, so `binding.etAllergies.text.isNotBlank()` works directly, drops nine string allocations, and removes that edge entirely.

**Verdict:** correct tightening with a real (and desirable) behaviour change behind it — ready.

---

dogi (approved):

PR #16255, opened 2026-08-25
life: smoother achievements editing (fixes #16253)

### LINKED ISSUE #16253: back button not working
back button not working in achievements
https://github.com/user-attachments/assets/af360420-6f2a-415c-b51a-e2d4f1aa8f53

ragilzakaria:
https://github.com/user-attachments/assets/1516a7b2-587f-437d-9598-a410d8142535

fixes #16253

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6744
-        versionName = "0.67.44"
+        versionCode = 6745
+        versionName = "0.67.45"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/ui/user/EditAchievementFragment.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/user/EditAchievementFragment.kt
@@ -113,6 +113,9 @@ class EditAchievementFragment : BaseContainerFragment(), DatePickerDialog.OnDate
     }
 
     private fun setListeners() {
+        binding.toolbar.setNavigationOnClickListener {
+            FragmentNavigator.popBackStack(parentFragmentManager)
+        }
         binding.btnUpdate.setOnClickListener {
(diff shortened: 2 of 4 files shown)


dogi:
**Review — rating: 85/100**

I traced the pop path rather than trusting that a back arrow wired to `popBackStack` must work, because `FragmentNavigator.popBackStack` no-ops when `backStackEntryCount == 0` — a dead arrow would stay dead, just differently. It's fine: `AchievementFragment:60` opens this via `openCallFragment(EditAchievementFragment())`, which reaches `DashboardElementActivity.openCallFragment` and lands on a `replaceFragment(..., addToBackStack = true)` branch (the tag is `"EditAchievementFragment"`, not `""`, so it takes the non-counter path — both branches set it anyway). The fragment is added to the activity's manager, so `parentFragmentManager` is the right one and the entry is really there to pop. The arrow now does what the system back button already did.

Reusing the existing `@string/btn_back` for `navigationContentDescription` is the right instinct — a raw `app:navigationIcon` with no description gives TalkBack nothing to announce, and picking an existing string sidesteps adding a new entry to six locale files for a two-word label. The `tools:context` correction from `.ui.userprofile.` to `.ui.user.` catches the same kind of stale package reference as the one in #16096.

**The test covers the half that didn't break, though.** It inflates the binding and asserts the navigation icon and content description are present — that's the accessibility addition — but nothing exercises the click listener, which is the actual bug. I'm not asking for it: verifying the pop would need a launched fragment, and there's no `androidTest` source set or `FragmentScenario` precedent in this repo to copy. Worth knowing that a future refactor deleting `setNavigationOnClickListener` would leave this test green.

**Nit:** `@Config(manifest = Config.NONE)` is a no-op under the AGP-driven Robolectric setup here, and `testToolbarNavigationIconAndContentDescription` uses the `test`-prefix style where the rest of the suite uses backtick sentences (`` `loadMyLifeList emits items from repository when available` ``).

**Verdict:** small fix for a genuinely dead control, with an accessibility improvement that costs no new strings — ready.

---

Okuro3499 (approved):

dogi (approved):

dogi (approved):

PR #16256, opened 2026-08-25
resources: smoother viewer text truncating (fixes #673)

### LINKED ISSUE #673: CSV Viewer: App is crashed while rendering large csv file.
App is not able to render the big csv file. I have uploaded the csv file ~4mb. Sample csv is attached here.

[https://we.tl/t-q3N5vMQSvt](https://we.tl/t-q3N5vMQSvt)

Okuro3499:
fixes #673
Read text files on the IO dispatcher and cap display at 500K characters to prevent UI freezes. Shows a toast when content is truncated. Adds translated strings for all supported languages.

## Summary by CodeRabbit

* **New Features**
  * Added support for displaying large text files by showing the first 500,000 characters.
  * Added a notification when file content has been truncated.
  * Preserved appropriate rendering for Markdown and plain-text files.

* **Localization**
  * Added truncation messages in Arabic, Spanish, French, Nepali, Somali, and English.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6747
-        versionName = "0.67.47"
+        versionCode = 6748
+        versionName = "0.67.48"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/ui/viewer/ResourceViewerFragment.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/viewer/ResourceViewerFragment.kt
@@ -52,6 +52,7 @@ import java.io.File
 import java.util.regex.Pattern
 import javax.inject.Inject
 import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
 import org.ole.planet.myplanet.R
 import org.ole.planet.myplanet.callback.OnAudioRecordListener
 import org.ole.planet.myplanet.data.auth.AuthSessionUpdater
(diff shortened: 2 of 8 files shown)


dogi:
**Review — rating: 84/100**

I pulled #673 to check the fix actually targets the reported failure, since "large file" can mean two very different crashes. It's a ~4 MB CSV that crashes *while rendering* — so roughly 4 million characters handed to a `TextView`, and the crash is the layout pass, not memory. Capping at 500,000 puts that repro 8× under the old load, which is the right lever for that bug. The `ResourceType.CSV` branch routes through `setupTextViewer` alongside TEXT and MARKDOWN, so the reported case is genuinely covered.

The threading half is a real bonus the title doesn't advertise: `file.readText()` was a synchronous disk read plus full UTF-16 decode on the main thread for every text resource, and it's now behind `withContext(dispatcherProvider.io)`. `setupViewer()` was already `suspend`, so making `setupTextViewer` suspend needed no call-site churn. The `isAdded` guard after the suspend point is placed correctly, and it's safe for a reason worth noting: `textFileTitle`/`textContent` are resolved from `binding.root` *before* the suspend, so nothing dereferences `binding` after resumption and a detached-view write can't NPE. All six locale files carry the new string in the same commit, which is the part that most often gets half-done.

**The memory ceiling is unchanged, though.** `readText()` still materialises the entire file as a String before `substring` runs, so the truncation is a display cap, not a read cap. Harmless for a 4 MB CSV — that's ~8 MB of UTF-16 and no device will blink — but it means a genuinely huge resource still fails the same way it does today, and the PR's own framing ("truncate large text files") implies otherwise. A bounded read is a few lines: fill a `CharArray(MAX_TEXT_VIEWER_CHARS + 1)` from `file.reader().use { … }` and treat a full buffer as the truncated case, which also gives you the truncation flag for free.

**Nit:** for `ResourceType.MARKDOWN` the expensive work moved rather than disappeared — `MarkdownUtils.setMarkdownText` calls `markwon.setMarkdown` synchronously, and by then we're back on the main dispatcher, so up to 500K characters of commonmark parse on the UI thread. One limit shared between markdown and plain text is generous for markdown, which costs far more per character. Also `substring(0, 500_000)` can land between a surrogate pair (lone surrogate → replacement glyph) or mid code-fence, so the tail can render as raw syntax.

**Verdict:** fixes the crash #673 actually reports, moves a main-thread file read off the UI thread on the way, and localises properly — ready; the bounded read is worth a follow-up.

---

dogi (approved):

PR #16257, opened 2026-08-25
resources: smoother filtering (fixes #16102)

### LINKED ISSUE #16102: download filter
like the label "All", "In progress"... that is in courses create a new label in mylibrary to filter downloaded and undownloaded resources
Image

J-S-webskas:
https://github.com/user-attachments/assets/9c9e111f-6757-481f-a8e5-ee73da693458

## Summary by CodeRabbit

* **New Features**
  * Added library filters for viewing all, downloaded, or not-downloaded resources.
  * Added selectable, horizontally scrollable filter chips below the library search controls.
  * Filter selections are retained when applying or clearing other library filters.
* **Localization**
  * Added download-status filter labels in Arabic, Spanish, French, Nepali, Somali, and English.
* **Usability**
  * Download status is accurately determined for locally available and offline resources.


dogi:
**Review — rating: 62/100**

**Blocking: the filter does nothing on a Nepali device.** `selectedDownloadFilter` is assigned from `R.array.download_filter` items, but `applyFilterModels` matches it against `R.string.downloaded` / `R.string.not_downloaded` — separate resources that only work if their text happens to be byte-identical. I compared all six locales:

| locale | array item 2 vs `downloaded` | array item 3 vs `not_downloaded` |
|---|---|---|
| default / ar / es / fr / so | match | match |
| **ne** | `डाउनलोड गरियो` vs `डाउनलोड गरिएको` | `डाउनलोड गरिएको छैन` vs `डाउनलोड नगरिएको` |

Neither Nepali item matches, so both chips fall through to `else -> true`: the chip highlights, the list never filters, and nothing looks broken enough to report. The coupling is already visible in this diff — `not_downloaded` was recapitalised from "Not downloaded" to "Not Downloaded" purely so the English comparison would succeed, which is the fix working backwards. Key the selection off the item's index (or a small enum), never its display text, and the locale question disappears along with the need to keep two resource sets in sync.

**Deleting `hasActiveFilters()` regresses existing behaviour.** `clearTags.visibility` is now unconditionally `VISIBLE`, so the clear-filters control shows even with nothing to clear — it used to appear only when a search term, tag, subject, level, language, or medium was set. The change the new chip needed was one more clause in that function (`selectedDownloadFilter.isNotEmpty()`), not its removal.

**`getLocallyOfflineIds()` allocates per resource.** It's `locallyOfflineIds.toSet()` — a fresh defensive copy — and it's called inside the `applyFilterModels` predicate, so a library of N resources copies the set N times on every filter pass, on a screen whose whole purpose is large offline libraries. Hoist it above the filter. While you're there: that call has no `::adapterLibrary.isInitialized` guard, and this file guards that exact property in ten other places, so the codebase already treats it as reachable before initialisation.

The parts that are easy to forget are done right: the chip row is added to all three layout variants (default, `layout-land`, `layout-sw600dp`), the array is translated in all six locales, `selectedDownloadFilter` survives view recreation while `setupDownloadFilterChips` re-renders selection from it, and `lastDownloadFilter` is correctly threaded into the change-detection short-circuit so a chip tap isn't swallowed as "nothing changed".

**Nit:** `binding.root.findViewById<LinearLayout>(R.id.chip_filter_row)` with null fallbacks, when the ID is present in all three variants and view binding would generate a non-null field. And a unit test over the `applyFilterModels` predicate across the three states would have caught the Nepali case the moment it was written.

**Verdict:** the feature is wanted and the wiring is mostly sound, but comparing against localized text silently disables it in one shipped language — fix that, restore `hasActiveFilters()`, and hoist the set copy.

---

Okuro3499 (approved):

Okuro3499 (dismissed):
```
1. Download filter is silently broken in Nepali — root cause is a fragile comparison design                                                                                        
                                                                                                                                                                                     
  ResourcesFragment.kt:836-839 matches the tapped chip against translated string resources:                                                                                          
  when (selectedDownloadFilter) {                                                                                                                                                    
      getString(R.string.downloaded) -> isDownloaded                                                                                                                                 
      getString(R.string.not_downloaded) -> !isDownloaded                                                                                                                            
      else -> true                                                                                                                                                                   
  }                                                                                                                                                                                  
  selectedDownloadFilter is set from the raw text of a <string-array name="download_filter"> item (ResourcesFragment.kt:908), which is a set of hardcoded literal strings, translated
  independently from R.string.downloaded / R.string.not_downloaded.                                                                                                                  
                                                                                                                                                                                     
  In values-ne/strings.xml these two now diverge:                                                                                                                                    
  - download_filter items: "डाउनलोड गरियो" / "डाउनलोड गरिएको छैन"                                                                                                                             oaded / not_downloaded strings: "डाउनलोड गरिएको" / "डाउनलोड नगरिएको"
  - downloaded / not_downloaded strings: "डाउनलोड गरिएको" / "डाउनलोड नगरिएको"                                                                                                        
        a Nepali device, tapping "Downloaded" or "Not Downloaded" never matches either branch, falls to else -> true, and the filter does nothing — silently. (ar/es/fr/so happen to 
  match today, but nothing enforces that; a future retranslation of either resource independently breaks it again in any locale.) The fix is to compare against a stable key         
  (index/enum), not localized display text.                                                                                                                                          
                                                                                                                                                                                     
  2. clearTags button visibility regression                                                                                                                                          
                                                                                                                                                                                     
  The old hasActiveFilters() gate was deleted rather than extended for the new download filter (ResourcesFragment.kt:473-497, diff removed lines). clearTags.visibility is now       
  unconditionally VISIBLE whenever there's any library data — previously it only showed when a search/tag/subject/level/language/medium filter was actually active. Now the "clear   
  filters" button is always shown even with zero filters applied.
```

dogi:
**Re-review — Rating: 82/100** (was 62)

All three findings are fixed, and the first one is fixed by construction rather than by patching: with `selectedDownloadFilterIndex: Int` and `when (…) { 1 -> isDownloaded; 2 -> !isDownloaded; else -> true }`, there's no string comparison anywhere in the path, so the Nepali case that silently disabled the filter can't come back no matter how anyone edits the array or the strings later. Tagging chips with the index and comparing `(chip.tag as? Int) == selected` keeps the render path consistent with that.

`hasActiveFilters()` is back and now includes `selectedDownloadFilterIndex != 0`, so the clear-filters button reflects the new chip too — better than where it started. And `getLocallyOfflineIds()` is hoisted above `models.filter { }` with the `::adapterLibrary.isInitialized` guard attached, which closes both the per-item set copy and the unguarded property access in one line.

**One thing the rewrite dropped:** the old `hasActiveFilters()` read `::searchTags.isInitialized && searchTags.isNotEmpty()`; the new one is just `searchTags.isNotEmpty()`. `searchTags` is `lateinit` (line 83), and this file still guards it the old way at line 384 (`if (::searchTags.isInitialized) searchTags else emptyList()`) — so the codebase evidently considers an early read reachable. I can't prove `checkList()` runs before `initArrays()` on any path, so I'm not blocking on it, but restoring the two words costs nothing and removes the question. (The `etSearch.text?.isNotBlank() == true` simplification is exactly equivalent to the old `.toString().trim().isNotEmpty()` — that part's fine.)

**Nits:** `clearTagsButton()` does `val chipRow = binding.chipFilterRow; if (chipRow != null)`, but the ID is in all three layout variants so view binding generates a non-null field — `setupDownloadFilterChips()` calls `binding.chipFilterRow.removeAllViews()` with no check, confirming it. That comparison is always true and will warn. `scroll_chip_filter` is still reached via `binding.root.findViewById` in `checkList` while its sibling uses binding. `options.indices.forEach { label -> … chip.text = options[label] }` names an index `label`. And the `not_downloaded` recapitalisation is a leftover from the string-matching approach — harmless now, and arguably tidier next to the array, just no longer load-bearing.

**On CI:** all four checks were still `in_progress` at review time, so this is on the diff alone. Worth a glance once they land, since the last round's failure mode wasn't something CI would have caught either.

**Verdict:** the locale bug is gone structurally, the clear-button regression is undone and improved, and the filter loop no longer allocates per item — ready; restore the `isInitialized` guard when convenient.

---

Okuro3499 (approved):

dogi (approved):

PR #16258, opened 2026-08-25
resources: smoother grid dark mode searching (fixes #16129)

### LINKED ISSUE #16129: Dark mode: search bar text should be white
Image

J-S-webskas:
fixes #16129
Recording_20260825_140757

## Summary by CodeRabbit

* **Style**
  * Improved search field hint text visibility in night mode.
  * Preserved the existing text color.


dogi:
**Review — rating: 90/100**

This lands exactly on the repo's existing convention rather than inventing a fix: `android:textColorHint="@color/hint_color"` already appears in 38 layout files, including `layout_search.xml:44` — the direct sibling of this one, with the identical attribute and value. `layout_search_pill.xml` was the straggler, and now the two search bars match.

Two things I checked because they'd have made a one-liner insufficient. There's no `layout-night/layout_search_pill.xml` (that folder holds only `activity_login.xml` and `app_bar_bell.xml`), so the default layout serves both modes and there's no night variant left behind still showing the old hint. And `hint_color` is `#808080` in *both* `values/colors.xml` and `values-night/colors.xml` — deliberately identical, a mid-grey that reads against white and against the dark background, so a single non-qualified value is the right call here rather than a missing night override.

Worth noting the fix is broader than the `resources:` prefix suggests: `layout_search_pill.xml` is `<include>`d by both `fragment_my_library.xml` and `fragment_my_course.xml` across all three variants (default, `layout-land`, `layout-sw600dp`), so the Courses search hint gets fixed too.

**Nit, and really a follow-up rather than a change here:** nothing in `values*/` sets `android:textColorHint` on any theme — that's *why* 38 layouts each carry the attribute by hand, and why a new layout starts out broken by default. One `<item name="android:textColorHint">@color/hint_color</item>` in the app theme would fix this class of bug for every remaining straggler at once and let the per-layout copies be deleted over time.

**Verdict:** correct one-line fix that matches its sibling layout exactly, with no night variant missed — ready.

---

Okuro3499 (changes requested):
"search" hint not visible in dark mode
Image

dogi:
**Re-review — Rating: 90/100** (was 90)

The new hunk is a better fix than the one I reviewed, and it's worth putting numbers on. Against the night palette — `secondary_bg` = `colorPrimaryDark` = `#242526`, `card_bg` = `colorPrimaryLight` = `#3a3b3c` — the old `#808080` hint gave roughly **3.9:1** on `secondary_bg` and **2.8:1** on `card_bg`, both under the 4.5:1 WCAG AA threshold for normal text, and the second under even 3:1. `#BDBDBD` moves those to about **8.2:1** and **6.0:1**. So my earlier read that the identical light/dark value was "deliberately mode-neutral" was too generous: it was legible on white and marginal-to-failing on the dark surfaces.

Two things worth knowing before this lands, neither a defect:

**It reaches further than the search bar.** `hint_color` is referenced by 38 layout files, so every hint in the app gets brighter in dark mode — which is the right fix, and better than the per-layout attribute this PR started as. More specifically, `values-night/colors.xml:16` aliases `text_highlight_color` to `@color/hint_color`, and that's used on three `TextView`s in `row_survey.xml`. Survey row highlight text therefore changes in dark mode too. Brighter is very likely an improvement for a highlight, but it's a Surveys-screen change sitting under a `resources:` title.

**Light mode is untouched and still marginal.** `hint_color` stays `#808080` in `values/colors.xml`, which is about **3.9:1** on white — the same failing ratio the dark side just moved off. Out of scope for #16129, and probably its own issue, but the asymmetry is now baked in deliberately rather than by accident.

The original one-liner still stands on its own: `layout_search_pill.xml` was the straggler next to `layout_search.xml:44`, which already carried the identical attribute and value.

**Nit, carried over:** no theme in `values*/` sets `android:textColorHint`, which is why 38 layouts set it by hand. That's still the single change that would retire this whole class.

**Verdict:** the palette fix addresses the real cause rather than one layout, with measurable contrast gains — ready.

---

Okuro3499 (approved):

dogi (approved):

PR #16259, opened 2026-08-25
login: smoother settings text capitalizing (fixes #16130)

### LINKED ISSUE #16130: String: capitalize Text Size
Image

J-S-webskas:
Recording_20260825_141419

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6748
-        versionName = "0.67.48"
+        versionCode = 6749
+        versionName = "0.67.49"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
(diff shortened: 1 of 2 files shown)


dogi:
**Review — rating: 88/100**

The one use site is `res/xml/pref.xml:5` as an `android:title` on the preference — a standalone settings label, never composed into a sentence, so Title Case is unambiguously right and there's no call site where the new capitalization would read oddly. It also settles an inconsistency sitting one line away: `select_text_size` was already "Select Text Size", and the neighbouring theme strings are "Follow System", "Light", "Dark", "Small", "Medium" — "text size" was the only lowercase entry in the block.

**Leaving the other five locales untouched is correct here, not an omission.** `values-es` has "tamaño del texto" and `values-fr` "taille du texte"; Romance languages use sentence case for UI labels, and Arabic and Nepali have no letter case at all. Title-casing those to match English would be the actual bug. This is genuinely an English-only change.

**Worth a follow-up issue rather than widening this PR:** the same region of `values/strings.xml` still has `manual`, `show_less` ("show less"), `show_more`, and `new_chat` starting lowercase, which look like the same class of label. Not a blanket sweep though — `hour`/`hours` are likely composed into phrases, `server_address_content_description` reads as a spoken sentence, and `switching_off_manual_configuration_to_clear_data` is a full sentence needing only its first letter. A sweep that ignored those distinctions would do damage.

**Verdict:** closes #16130 exactly as scoped, with the right judgement on which locales to leave alone — ready.

---

Okuro3499 (approved):

dogi (approved):

PR #16261, opened 2026-08-26
actions: smoother coderabbit reviewing (fixes #16287)

### LINKED ISSUE #16287: Coderabbit label gated reviews
Tracking issue for PR #16261.

## Summary by CodeRabbit

* **Configuration**
  * Code reviews are now opt-in using a `review` label.
  * Incremental reviews and draft pull request reviews are enabled.
  * Reviews use a more assertive profile.
  * Automated pre-merge checks and documentation/unit-test finishing actions are disabled.
  * Simplification assistance remains enabled.
  * Added and updated guidance for application source, test, and build configuration files.

---
https://claude.ai/code/session_01TPnmR8Na2vb3CBjm4LGJr9

dogi:
## Summary by CodeRabbit

* **Configuration**
  * Code reviews are now opt-in using a `review` label.
  * Incremental reviews and draft pull request reviews are enabled.
  * Reviews use a more assertive profile.
  * Automated pre-merge checks and documentation/unit-test finishing actions are disabled.
  * Simplification assistance remains enabled.
  * Added and updated guidance for application source, test, and build configuration files.

---
https://claude.ai/code/session_01TPnmR8Na2vb3CBjm4LGJr9


Okuro3499 (approved):

dogi:
**Review — rating: 76/100**

The `pre_merge_checks` correction is the best thing in here and deserves calling out: the old file had `docstring_coverage: { mode: "off" }`, and the comment now records that the real key is `docstrings` and that unknown keys under that block validate silently. That means the perpetual KDoc warning the old config thought it had suppressed was never actually suppressed — a silent-no-op class of bug that's hard to notice and easy to repeat.

**`poem: false` was dropped and not replaced.** The old file set it; the new file has no `poem` key anywhere, and it defaults on — so CodeRabbit's closing haiku comes back on every review. That contradicts the header this same commit writes ("ceremony is off and substance is up") and the `tone_instructions` right below it ("No compliments, no restating the diff"), and it sits directly under a comment reading "Each defaults on; none is why anyone labels a PR" — which is exactly the list `poem` belongs in. `high_level_summary` and `collapse_walkthrough` were also explicitly set before and are now absent; worth confirming their defaults land where you want, since `collapse_walkthrough` defaulting off would make threads longer, not shorter.

**Deleting the whole `tools:` block moves gitleaks onto an implicit default, and this repo has stakes there.** CLAUDE.md documents that `gradle.properties` is tracked in git and holds real `PLANET_*_PIN` CouchDB credentials baked into `BuildConfig`. Secret scanning is the one entry in that deleted block with live consequences here, so it's the one I'd keep explicit rather than inherit. Relatedly, the new `tone_instructions` says "Skip whatever detekt and the Gradle build already report" in the same commit that removes `tools.detekt.enabled: true` — and detekt isn't in this build at all (nothing in any `.gradle`, `.toml`, `.kts`, or workflow references it), so that clause is leaning entirely on CodeRabbit's own bundled detekt still defaulting on.

On the core mechanism — whether `auto_review.enabled: false` plus `labels: ["review"]` gates rather than disables — I can't check CodeRabbit's schema from here, so I won't assert either way. But there's supporting evidence inside the diff: `review_status: false  # else a "Review skipped" notice on every unlabelled PR` only makes sense if CodeRabbit is still evaluating unlabelled PRs and declining them, which is label-filter behaviour, not off. That reads like something observed rather than assumed. This PR carries `review` itself, so the first labelled PR after it lands is the test.

**Worth knowing, not a defect:** `review` now does double duty — it's this repo's human triage queue label *and* the CodeRabbit trigger. So every PR entering the queue spends one of the 1–10 reviews an hour the header is budgeting for, and the `change` → author pushes → re-add `review` cycle re-triggers a full review each round. That's probably what you want for a re-review, but label churn is now review spend, and the header doesn't say so.

**Verdict:** good rewrite with a real catch in it, but restore `poem: false` and keep gitleaks explicit — two lines, and the config stops contradicting its own stated goal.

---

dogi:

Both points are addressed, and the more interesting one is a default I missed entirely: `auto_pause_after_reviewed_commits: 0`. Left at its default of 5, CodeRabbit would pause the incremental reviews `auto_incremental_review: true` had just switched on — silently, and specifically on long-lived PRs, which are the ones where opt-in review matters most. The old `AGENT_SPELLBOOK` row corroborates it, since it already described the auto-pause as observed behaviour ("auto-pauses after 5 reviewed commits"), so the config and the field notes now agree instead of quietly disagreeing.

`gitleaks` is pinned with the reasoning attached — `gradle.properties` tracked, real `PLANET_*_PIN` CouchDB credentials — while the other tools are left to defaults with a comment saying that's deliberate. Pinning the one entry with real stakes and inheriting the rest is the right split, and better than restoring the whole block.

On `poem`: it's pinned now, which settles it. Your comment says it already defaults off, and you have the live signal here that I don't — so I'll take that over my read that it defaults on. The pin is the right move either way, and your justification for it is the stronger argument: a decoration arriving because an upstream default flipped would undercut this file's own claim without anyone noticing.

The docs are the part that turns this from a config tweak into a change someone can follow later. The `AGENT_SPELLBOOK` row now describes the gating, drafts included, auto-pause off, and the grey "Review skipped" status on unlabelled PRs; `CODE_STYLE_GUIDE.md` → "PR Reviews" no longer claims CodeRabbit auto-reviews every non-draft push and records the `@coderabbitai review` escape hatch. That escape hatch also answers most of what I flagged as unverifiable last round — an unlabelled PR still being *given* a "Review skipped" status is label-filter behaviour, not off, and it now reads like something observed rather than assumed.

**On CI:** the head shows zero check runs, which is `paths-ignore` (`docs/**`, `**/*.md`) correctly skipping a docs-only final push, not a failure — and it's moot here, since the diff contains no compiled code for build or test to verify.

**Nit:** `tone_instructions` still says "Skip whatever detekt and the Gradle build already report", and detekt isn't part of this Gradle build at all — the clause only makes sense against CodeRabbit's own bundled detekt, which the `tools` comment now confirms is on by default. Worth a word tweak so the next reader doesn't go hunting for a detekt task in `app/build.gradle`.

**Verdict:** the gating is now documented where people will find it, the one security-relevant tool is pinned rather than inherited, and the auto-pause default that would have defeated the feature is switched off — ready.

---

PR #16275, opened 2026-08-26
login: smoother form submitting (fixes #16267)

### LINKED ISSUE #16267: double-tap race on Sign In
```
LoginActivity.kt:280 — double-tap race on Sign In. The click handler re-enables the button and dismisses the progress dialog immediately, while the actual login/sync coroutine 
it launched keeps running unawaited — a second tap starts a second concurrent login racing on shared session state.
````

Okuro3499:
fixes #16267
Move the coroutine scope to the call site and make submitForm a suspend function that directly calls AuthUtils.login.


J-S-webskas (approved):

dogi:
**Review — rating: 86/100**

The diff only wraps one of the three `submitForm` call sites, which is worth checking rather than assuming CI covers it — a `suspend` function called from a non-coroutine context is a compile error, so green tells you the other two compile, not that they behave the same. They do, and they come out better:

- **229** — `lifecycleScope.launch { delay(500); submitForm(...) }`; `submitForm` is the last statement in the block.
- **280** — `lifecycleScope.launch { … if (user == null || !user.isArchived) submitForm(...) else { … } }`; nothing follows the `if`/`else` inside the launch.
- **605** — newly wrapped in its own `lifecycleScope.launch`, which is exactly what `submitForm` used to do internally.

Both existing sites were already inside `lifecycleScope.launch`, so the old code was spawning a second coroutine on the *same scope* from inside the first. Removing that nesting is the actual win here, not just relocating the `launch`. Cancellation is unchanged — every path still hangs off `lifecycleScope` and dies at `onDestroy` — and because nothing follows `submitForm` in either enclosing block, the shift from fire-and-forget to awaited has no observable effect on sequencing. That's the property that makes this safe, and it's worth stating explicitly, since the same refactor with a statement after the call would have changed when it ran.

Making it `suspend` is also the more honest signature: `AuthUtils.login` is suspending work, and the caller now decides the scope instead of the function silently picking one.

**Nit:** the `delay(500)` before the auto-submit at 229 is a magic wait and pre-existing — untouched here, but now sitting one line from the call it gates, which makes it more conspicuous than it was.

**Verdict:** behaviour-preserving, verified at all three call sites, and it drops two redundant nested launches — ready.

---

dogi (approved):

PR #16276, opened 2026-08-26
resources: smoother filter initializing (fixes #16268)

### LINKED ISSUE #16268: crash on fragment recreation
```
ResourcesFilterFragment.kt:121 — crash on fragment recreation. filterListener?.getSelectedFilter()?.get("levels") as MutableSet<String> unchecked-casts a null chain when the  
fragment is recreated (config change/process restart) before setListener() is called again → NPE/crash.   
```

Okuro3499:
fixes #16268
Replace direct unchecked casts to MutableSet<String> with safe toMutableSet() calls and fallback to existing values, preventing potential ClassCastException.


dogi:
**Review — rating: 88/100**

The crash is real and reachable. `filterListener?.getSelectedFilter()?.get("levels") as MutableSet<String>` looks null-safe because of the `?.` chain, but the chain only produces `null` — and `null as MutableSet<String>` is a cast to a non-null type, which throws. `filterListener` is assigned only in `setListener` (line 41), so a `DialogFragment` recreated by the system with the filter sheet open has it null and takes down the screen on the first of those four lines. The `?:` fallback now keeps whatever the field already held.

The deeper point is that the cast was lying about the interface, not just about nullability: `OnFilterListener.getSelectedFilter()` returns `Map<String, Set<String>>` — read-only `Set`. The cast only ever succeeded because `ResourcesFragment` happens to store `HashSet`s, so it was leaning on a runtime type the contract doesn't promise. `.toMutableSet()` honours the declared type instead of betting on the implementation.

**That copy also removes shared mutable state, and I checked the round trip survives it.** Under the old aliasing, `selectedLvls` *was* `ResourcesFragment.levels` — the same object — so `addToList` at line 146 mutated the host fragment's live filter set in place, and line 149 then handed that same object back through `filter(...)`. Now the dialog mutates its own copy, and `ResourcesFragment.filter()` assigns all four arguments to its fields (lines 674-677) before re-running the query, so the end state is identical with no object shared between dialog and host. Collapsing four `getSelectedFilter()` calls into one is a free bonus — each built a fresh `HashMap` (lines 690-696).

**Nit:** `?: selectedLvls` assigns the field to itself as the fallback, which reads oddly even though preserving the prior value is the right behaviour. `?.let { selectedLvls = it.toMutableSet() }` says the same thing without the self-reference.

**Verdict:** fixes a reachable crash, drops accidental shared mutable state between the dialog and its host, and starts honouring the declared interface type — ready.

---

dogi (approved):

PR #16280, opened 2026-08-26
all: smoother room dao nullable querying (fixes #16278)

### LINKED ISSUE #16278: Systemic = vs IS bug on nullable Room query params
```
CourseProgressDao.kt (10,13,14,16,19), SubmissionDao.kt (15,23-25,28,30,37,39), NotificationDao.kt:12, RemovedLogDao.kt    
(12,15,25), TeamLogDao.kt:21, TeamTaskDao.kt:11. WHERE col = :param never matches when :param is null (SQL semantics), so course progress, submissions, notifications-read,     
tombstones, last-visit, and open-tasks queries silently return empty/no-op instead of erroring. Other DAOs in the same codebase already use IS correctly, so this is            
inconsistency, not a design choice. 
```

Okuro3499:
fixes #16278
Replace = with IS in Room @Query predicates where parameters are nullable (String?), ensuring NULL values match NULL rows correctly in SQL.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of records associated with missing or optional user, course, team, parent, and step identifiers.
  - Fixed database lookups, counts, updates, and deletions so entries with empty identifiers are matched correctly.
  - Improved retrieval of pending submissions, open tasks, notifications, team activity, and course progress when optional values are unavailable.


dogi:
**Review — rating: 88/100**

The edits are per-parameter rather than a find-and-replace, which is what makes this safe. `RemovedLogDao` is the clearest evidence: two adjacent queries both take a `type` argument, and the one declared `type: String?` becomes `IS` while the one declared `type: String` stays `=`. Same discipline in `NotificationDao` (`userId IS :userId AND type = :type`) and `CourseProgressDao`, where `stepNum = :stepNum` is correctly left alone as a non-null `Int`.

Two null-handling idioms are also correctly left untouched, and they'd have been easy to sweep up by mistake. `SubmissionDao.getByParentUserAndStatus` keeps `(:status IS NULL OR status = :status)` — that means "don't filter when the argument is null", which is a different question from "match rows whose status is NULL", and converting it would have changed results. `MyLifeDao` already solves the same problem explicitly with `(:userId IS NULL OR userId IS NULL OR userId = :userId)`.

I scanned every DAO for remaining nullable-param `= :param` predicates to check completeness. Against the file-and-line inventory in #16278 this PR is complete — all six files, all the listed sites. The scan did turn up four more that the issue's own inventory missed, none of them regressions here:

- `SubmissionDao.observePendingSurveys(userId: String?)` — in this very file, one line below `countPendingSurveys(userId: String)` whose `=` is correctly kept
- `AnswerDao.getBySubmissionAndQuestion(…, questionId: String?)`
- `PersonalDao.countByTitle(…, userId: String?)`
- `TagDao.getParentTags(db: String?, …)`

(The `markUploaded` hits in `SubmissionDao`, `TeamTaskDao` and `HealthExaminationDao` are `SET` assignments, not predicates — `IS` would be a syntax error there.)

**One thing worth confirming rather than a defect:** for `SELECT`s this change is strictly a fix, but four of the edits are on `DELETE`/`UPDATE` statements — `RemovedLogDao.deleteByTypeUserAndDoc`, `SubmissionDao.deleteByParentAndUser`, `deletePendingSurveyOrphans`, and `NotificationDao.markSummaryAsRead`. There, a null argument used to make the statement a guaranteed no-op; now it matches and mutates the NULL-column rows. That's presumably the intent — `deletePendingSurveyOrphans` deleting rows with no user is close to the definition of an orphan — but it's the one place where a null that previously slipped by harmlessly now removes data, so it's worth being deliberate about.

**Verdict:** correct, precisely scoped, matches the `IS`-for-nullable rule in CLAUDE.md, and closes #16278 as written — ready; the four sites above are worth folding into the issue for a follow-up.

---

dogi (approved):

PR #16313, opened 2026-08-27
actions: smoother workflow automerging (fixes #16301)

### LINKED ISSUE #16301: Serialise the automerge drain and stop it cloning every blob in history
`automerge.yml` has no `concurrency:` block (unlike `playstore.yml:53`). Two dispatch clicks put two jobs on the same queue, both bumping the version and pushing to protected master — the second claims a versionCode the first already took. Its checkout also uses `fetch-depth: 0` with no blob filter, downloading every blob of every commit when the script only reads a handful of files.

**Ask:**
- Add a `concurrency:` group (matching `playstore.yml`'s pattern) so two manual dispatches of the drain can't race on the same version bump.
- Add a blob filter to the checkout step (e.g. `filter: blob:none`) since the script only reads a handful of files, not full history content.

**Files:** `.github/workflows/automerge.yml`

**Verification:** no `concurrency` key present; `checkout@v7` with `fetch-depth: 0` at lines 95-97.

dogi:
Add concurrency group 'myPlanet-automerge' to serialize automerge workflow runs and prevent version bump races. Add 'filter: blob:none' to checkout step to prevent downloading all historical blobs.

Fixes #16301

---
*PR created automatically by Jules for task [5279313950849794510](https://jules.google.com/task/5279313950849794510) started by @dogi*

### DIFF
--- a/.github/workflows/automerge.yml
+++ b/.github/workflows/automerge.yml
@@ -82,7 +82,7 @@ permissions:
   actions: write
 
 concurrency:
-  group: myPlanet-automerge
+  group: automerge
   cancel-in-progress: false
 
 env:
--- a/.github/workflows/automerge.yml
+++ b/.github/workflows/automerge.yml
@@ -81,6 +81,10 @@ permissions:
   statuses: read
   actions: write
 
+concurrency:
+  group: myPlanet-automerge
+  cancel-in-progress: false
+
 env:
   DEFAULT_BASE: master
   GRADLE_FILE: app/build.gradle
@@ -95,6 +99,7 @@ jobs:
         uses: actions/checkout@v7
         with:
           fetch-depth: 0
+          filter: blob:none
           token: ${{ secrets.AUTOMERGE_TOKEN || github.token }}
 
       - name: stage helper scripts outside the work tree
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6768
-        versionName = "0.67.68"
+        versionCode = 6769
+        versionName = "0.67.69"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }


Okuro3499 (approved):

dogi:
**Review — rating: 88/100**

Two changes to `automerge.yml`: a concurrency group and a blobless checkout filter. Both are correct, and I traced each against what `automerge.sh` actually does.

**The concurrency group is right here** — and notably for a reason that doesn't hold in #16311, which adds the same construct to `release.yml`. `cancel-in-progress: false` still lets a *pending* run be superseded when a third enters the group. In `release.yml` that silently drops a shippable release. Here it costs nothing: `automerge.yml` is `workflow_dispatch`-only and the drain re-derives its work list from the `automerge` labels at run time, so a superseded duplicate dispatch loses no work — the surviving drain picks up every labelled PR anyway. Serializing is also genuinely load-bearing: two concurrent drains would each merge base, bump the version in `app/build.gradle`, and push to master, which is a guaranteed collision and possibly duplicate version codes. Using a constant group string rather than keying on ref is correct, since the drain always targets master regardless of dispatch inputs.

**`filter: blob:none` suits this checkout precisely.** The drain needs full *commit* history — hence `fetch-depth: 0`, for merge-base computation — but only a tiny fraction of blobs. I checked every git operation in the script and none walks blobs across history: `git diff --name-only` (line 352) resolves from trees; `git diff --unified=0 -- "$GRADLE_FILE"` (358) and `git show "origin/$BASE:$GRADLE_FILE"` (483) touch a single file; the `git merge` calls (491, 507) need blobs only for files that differ between base and head, a bounded set per PR. So the lazy-fetch round trips stay proportional to real work while the upfront clone skips the whole blob history — which on this repo (181 layouts plus drawables) is most of the download. Push is unaffected: the remote already has every base and head object, so negotiation sends only the new merge commit and bumped `build.gradle` blob.

`filter` is supported on `actions/checkout@v7` and composes with `fetch-depth: 0` without conflict.

CI: build (default + lite) and both test shards green. No tests expected — this is workflow YAML, and the repo has no harness for it.

**Verdict:** mergeable.

---

dogi:

---

PR #16315, opened 2026-08-27
all: smoother network monitor working (fixes #16299)

### LINKED ISSUE #16299: Let WorkManager wait for the network instead of parking NetworkMonitorWorker on it
`NetworkMonitorWorker.start` enqueues a `OneTimeWorkRequest` with **no constraints**, and `doWork` then blocks on `NetworkUtils.isNetworkConnectedFlow.first { it }`. On an offline device the worker holds an execution slot and its wake lock until the 10-minute limit, is stopped without a `Result`, and reschedules — a loop for as long as the device is offline.

**Ask:** add a `NetworkType.CONNECTED` constraint to the work request instead of blocking inside `doWork` — the pattern `MainApplication.scheduleAutoSyncWork` already uses.

**Files:** `services/NetworkMonitorWorker.kt`, `(new) test/services/NetworkMonitorWorkerTest.kt`

**Verification:** `start()` currently builds the request with only `.addTag()`; `doWork()` awaits `isNetworkConnectedFlow.first{it}`.

dogi:
Add NetworkType.CONNECTED constraint to NetworkMonitorWorker start request to let WorkManager wait for network connectivity rather than blocking inside doWork. Added corresponding unit tests in NetworkMonitorWorkerTest.

Fixes #16299

---
*PR created automatically by Jules for task [3520844021269107969](https://jules.google.com/task/3520844021269107969) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6769
-        versionName = "0.67.69"
+        versionCode = 6770
+        versionName = "0.67.70"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/services/NetworkMonitorWorker.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/services/NetworkMonitorWorker.kt
@@ -2,17 +2,17 @@ package org.ole.planet.myplanet.services
 
 import android.content.Context
 import androidx.hilt.work.HiltWorker
+import androidx.work.Constraints
 import androidx.work.CoroutineWorker
 import androidx.work.Data
 import androidx.work.ExistingWorkPolicy
+import androidx.work.NetworkType
 import androidx.work.OneTimeWorkRequestBuilder
 import androidx.work.WorkManager
 import androidx.work.WorkerParameters
 import dagger.assisted.Assisted
 import dagger.assisted.AssistedInject
 import java.util.concurrent.TimeUnit
-import kotlinx.coroutines.flow.first
-import org.ole.planet.myplanet.utils.NetworkUtils
 
 @HiltWorker
 class NetworkMonitorWorker @AssistedInject constructor(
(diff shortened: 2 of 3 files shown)


Okuro3499 (approved):

dogi:
**Review — rating: 89/100**

Replaces an in-`doWork` suspension on `NetworkUtils.isNetworkConnectedFlow.first { it }` with a `NetworkType.CONNECTED` constraint on the work request, and adds a test file for a worker that had none. This is the right direction, and it fixes more than it advertises.

What I verified:

- **The two paths are semantically equivalent at the entry point.** `isNetworkConnectedFlow` maps `_currentNetwork` through `isConnected()` off `ConnectivityManager`, which is what `NetworkType.CONNECTED` expresses declaratively — same "any connected network" predicate, moved from a runtime wait to a scheduling precondition.
- **It removes a real failure mode, not just a style wart.** `start()` is called once, from `MainApplication.kt:488`, at app startup. Under the old shape the worker was enqueued unconstrained, began executing immediately, and then parked inside `doWork` waiting for connectivity — against WorkManager's ~10-minute execution ceiling. A device that started the app offline and stayed offline past that window got its worker stopped mid-wait, and a suspended `CoroutineWorker` doesn't survive process death, so the reachability check was then only re-armed at the next launch. With the constraint, the request sits ENQUEUED, is persisted by WorkManager, and fires when connectivity actually returns — including after the process has been killed. The reconnection case is strictly more reliable now.
- **No dead code left behind.** `isNetworkConnectedFlow` still has three production consumers (`MainApplication.kt:430`, `SyncActivity.kt:656`, `BellDashboardViewModel.kt:53`), so dropping this call site doesn't strand the flow.
- **The test asserts the actual claim.** `start_enqueuesUniqueWorkWithConnectedNetworkConstraint` captures the request and checks `workSpec.constraints.requiredNetworkType` directly, rather than just asserting something was enqueued. The other two cover the reachability hand-off (including the `network_reconnection_trigger` input) and the retry-on-throw path. There's also an implicit guard worth noting: if the `first { it }` wait had been left in, `doWork_schedules…` would hang instead of passing, so the suite fails loudly if someone reintroduces it.
- CI: build (default + lite) and both test shards green.

**Nit:** the test mocks `WorkManagerImpl`, an internal implementation class, where the `WorkManager` API type would do — `getInstance` is stubbed to return it either way, and typing the field to the public class keeps the test off WorkManager internals. Not worth a round-trip on its own; fold it in if you touch the file.

**Verdict:** mergeable.

---

dogi:
**Promoted to `merge`** — Okuro3499 approved, my review stands at 89/100, no changes since.

---

PR #16320, opened 2026-08-27
teams: smoother repository member querying (fixes #16300)

### LINKED ISSUE #16300: Batch the per-member user lookups in TeamsRepositoryImpl (N+1)
`getJoinedMembers` (line 960) and `getRequestedMembers` (line 1051) both end in `.mapNotNull { userRepository.getUserById(it) }` — one query per member. `UserRepository.getUsersByIds` already exists and chunks into 400-id `IN` queries.

**Ask:** replace both per-member lookup loops with a single `getUsersByIds` call. **`getUsersByIds` may not return input order** — preserve the existing ordering and dedup semantics of both call sites when switching.

**Files:** `repository/TeamsRepositoryImpl.kt`

**Verification:** both call sites confirmed at the cited lines; `getUsersByIds` exists at `UserRepository.kt:21`.

dogi:
Batched per-member user lookups in getJoinedMembers and getRequestedMembers in TeamsRepositoryImpl to avoid N+1 queries while preserving original ordering and deduplication semantics.

Fixes #16300

---
*PR created automatically by Jules for task [16832460847586196363](https://jules.google.com/task/16832460847586196363) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6770
-        versionName = "0.67.70"
+        versionCode = 6771
+        versionName = "0.67.71"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/repository/TeamsRepositoryImpl.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/TeamsRepositoryImpl.kt
@@ -961,7 +961,9 @@ class TeamsRepositoryImpl @Inject constructor(
             .filter { !it.isDeletePending } // Filter so only the not pending members get query
             .mapNotNull { it.userId }
             .distinct()
-        return teamMembers.mapNotNull { userRepository.getUserById(it) }
+        if (teamMembers.isEmpty()) return emptyList()
+        val userMap = userRepository.getUsersByIds(teamMembers).associateBy { it.id }
+        return teamMembers.mapNotNull { userMap[it] }
     }
 
     override suspend fun getJoinedMembersWithVisitInfo(teamId: String): List<JoinedMemberData> {
@@ -1050,7 +1052,9 @@ class TeamsRepositoryImpl @Inject constructor(
         val requestedMemberIds = teamDao.getByTeamIdAndDocType(teamId, "request")
             .mapNotNull { it.userId }
             .distinct()
-        return requestedMemberIds.mapNotNull { userRepository.getUserById(it) }
+        if (requestedMemberIds.isEmpty()) return emptyList()
+        val userMap = userRepository.getUsersByIds(requestedMemberIds).associateBy { it.id }
+        return requestedMemberIds.mapNotNull { userMap[it] }
     }
 
     override suspend fun isTeamNameExists(name: String, type: String, excludeTeamId: String?): Boolean {
--- a/app/src/main/java/org/ole/planet/myplanet/repository/TeamsRepositoryImpl.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/TeamsRepositoryImpl.kt
@@ -962,7 +962,12 @@ class TeamsRepositoryImpl @Inject constructor(
             .mapNotNull { it.userId }
             .distinct()
         if (teamMembers.isEmpty()) return emptyList()
-        val userMap = userRepository.getUsersByIds(teamMembers).associateBy { it.id }
+        val users = userRepository.getUsersByIds(teamMembers)
+        val userMap = HashMap<String, UserEntity>(users.size * 2)
+        users.forEach { user ->
+            userMap[user.id] = user
+            user._id?.let { userMap[it] = user }
+        }
         return teamMembers.mapNotNull { userMap[it] }
     }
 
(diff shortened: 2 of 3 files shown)


Okuro3499 (dismissed):
```
1. TeamsRepositoryImpl.kt:965 and :1056 — associateBy { it.id } loses the id-OR-_id fallback that getUserById/getUsersByAnyIds provide at the DB layer. Fix by keying the map on   
     both fields so lookups by either value resolve to the same row, e.g.:                                                                                                           
  val userMap = HashMap<String, UserEntity>()                                                                                                                                        
  userRepository.getUsersByIds(teamMembers).forEach { u ->                                                                                                                           
      userMap[u.id] = u                                                                                                                                                              
      u._id?.let { userMap.putIfAbsent(it, u) }                                                                                                                                      
  }                                                                                                                                                                                  
  return teamMembers.mapNotNull { userMap[it] }                                                                                                                                      
     Apply identically at both call sites. (This gap is pre-existing in the other associateBy { it.id } call sites listed above too, but this diff is what makes it live for two     
     additional, user-facing team-membership screens, so it should be closed here rather than inherited silently.)
```

dogi:
**Review — rating: 64/100**

Collapsing the per-member `getUserById` N+1 into one `getUsersByIds` is the right call, and the order-preserving re-map is the right shape for it. But the re-keying drops members, and the new tests are constructed so they can't see it.

**The bug.** The two lookup paths don't resolve users the same way:

```sql
-- UserDao.getById          (what the old code called, per member)
SELECT * FROM users WHERE id = :id OR _id = :id LIMIT 1
-- UserDao.getUsersByAnyIds (what getUsersByIds calls, batched)
SELECT * FROM users WHERE id IN (:userIds) OR _id IN (:userIds)
```

Both match on **either** column — that part is faithfully batched. The loss happens after the query:

```kotlin
val userMap = userRepository.getUsersByIds(teamMembers).associateBy { it.id }
return teamMembers.mapNotNull { userMap[it] }
```

`associateBy { it.id }` keys the map by the `id` column *only*. A user that was matched via its `_id` gets filed under its `id`, which is a different string — so the lookup by the original `MyTeam.userId` misses and `mapNotNull` silently discards them.

Concretely, for a row with `id = "org.couchdb.user:alice"`, `_id = "abc123"` and a membership row carrying `userId = "abc123"`:
- old: `getById("abc123")` matches on `_id` → member appears
- new: the batch query matches on `_id`, then the entry is keyed `"org.couchdb.user:alice"`, so `userMap["abc123"]` is `null` → **member disappears**

This isn't hypothetical. `UserEntity` declares `id: String` as the primary key and `_id: String?` as a *separately indexed* column, and the parse path at `UserEntity.kt:197` sets `id` from `_id` when present but falls back to `"org.couchdb.user:${name}"` when it isn't — so divergent pairs are routine. The `OR _id = :id` in `getById` exists precisely because callers pass either form; if the two always agreed, that clause would be dead.

Effect: team members vanish from the members list and join requests vanish from the requested list, with no error — `getJoinedMembers` and `getRequestedMembers` are both affected.

**Fix** — key the map by both columns:

```kotlin
if (teamMembers.isEmpty()) return emptyList()
val users = userRepository.getUsersByIds(teamMembers)
val userMap = HashMap<String, UserEntity>(users.size * 2)
users.forEach { user ->
    userMap[user.id] = user
    user._id?.let { userMap[it] = user }
}
return teamMembers.mapNotNull { userMap[it] }
```

**The tests need to grow a case.** `getJoinedMembers uses getUsersByIds and preserves order` builds every user with `id = "userN"` and every member with `userId = "userN"` — only the `id`-matches-`id` path. Returning the users out of order to prove ordering is a nice touch, and the `coVerify(exactly = 0) { getUserById(any()) }` guard against regressing to N+1 is good, but neither test can fail on the bug above. Add one member whose `userId` matches a user's `_id` while that user's `id` differs, and assert it survives — that pins the actual contract `getById` had.

Worth crediting: `getUsersByIds` already chunks at 400 and `distinctBy`s, so the SQLite variable ceiling on large teams is handled, and the `isEmpty()` early return avoids an empty `IN ()`.

**Verdict:** needs another round — the batching is worth doing and the surrounding work is solid, but as written it silently drops members whose membership rows reference `_id`.

---

dogi:
## Review: 88/100 — mergeable

A real N+1 removal that gets the subtle part right. The trap here is this codebase's `id`/`_id` conflation, and both halves of it are handled.

**The query side matches the old semantics exactly.** `getUserById` resolves to `UserDao.getById`:

```sql
SELECT * FROM users WHERE id = :id OR _id = :id LIMIT 1
```

and the replacement `getUsersByIds` resolves to `UserDao.getUsersByAnyIds`:

```sql
SELECT * FROM users WHERE id IN (:userIds) OR _id IN (:userIds)
```

Both columns, in both versions — so the batch returns exactly the set the N single queries would have. This is the thing that would have quietly broken the feature: a member row whose `userId` holds a `_id` value would vanish from the team list if the batch query only matched `id`, and no test with tidy fixtures would notice.

**The lookup side reproduces it too.** Keying `userMap` on both `user.id` and `user._id` is what makes the in-memory join equivalent to `getById`'s `OR`, and iterating the original `teamMembers` list (not the query result) preserves order and multiplicity. `getUsersByIds` already `.chunked(400)`s, so the SQLite variable limit is handled upstream.

**Tests are the good kind.** `user1` deliberately has `id = "org.couchdb.user:alice"` with `_id = "user1"` while the membership row references `"user1"` — that's precisely the `_id`-match path. Returning the users shuffled (`listOf(user3, user1, user2)`) proves ordering comes from the id list rather than the query result, and `coVerify(exactly = 0) { userRepository.getUserById(any()) }` pins the N+1 removal itself. Both fail on `master`. The `isDeletePending` member is correctly excluded before the ids are collected, so the filter still runs first.

**Nits, none blocking**

- The 7-line map-building block is duplicated verbatim across `getJoinedMembers` and `getRequestedMembers`. `TeamsRepositoryImpl` is already the largest file in the tree (~1437 lines, called out in CLAUDE.md as a split candidate) — a private `usersByIdsPreservingOrder(ids: List<String>)` would collapse both and give the next caller somewhere obvious to go.
- `HashMap(users.size * 2)` is an initial *capacity*, and the map takes up to `2 × users.size` entries, so at load factor 0.75 it still resizes once. Harmless; if the sizing is meant to avoid that, it wants `/ 0.75`.
- Where a `_id` on one row collides with an `id` on another, the map's last-write-wins can pick a different row than `LIMIT 1` would have. Both are arbitrary in that (already broken) data state — worth knowing, not worth guarding.

**Impact:** a 50-member team goes from 50 round trips to one, on a screen users open constantly.

**CI** — build (default + lite), both test shards, labels green.

**Verdict:** mergeable.

---

PR #16322, opened 2026-08-27
all: less services submissions hilt injections is more (fixes #16307)

### LINKED ISSUE #16307: Delete four dead Hilt injections from services and one fragment
Four classes ask Hilt for dependencies they never read:
- `SyncManager:70` injects `TeamsRepository` (making the sync orchestrator formally depend on the whole teams domain)
- `UploadCoordinator:28` injects `@ApplicationContext Context`
- `DownloadService:58` field-injects `DispatcherProvider`
- `SubmissionsFragment:34` field-injects `UserSessionManager`

**Ask:** remove all four dead injections. `SyncManagerTest`'s constructor call needs to drop the `TeamsRepository` argument accordingly.

**Files:** `services/sync/SyncManager.kt`, `services/upload/UploadCoordinator.kt`, `services/DownloadService.kt`, `ui/submissions/SubmissionsFragment.kt`, `test/.../SyncManagerTest.kt`

**Verification:** all four confirmed unused at the cited lines.

dogi:
Removed four unused Hilt injections across SyncManager, UploadCoordinator, DownloadService, and SubmissionsFragment, and updated SyncManagerTest to match constructor signature changes.

Fixes #16307

---
*PR created automatically by Jules for task [11810403442557964676](https://jules.google.com/task/11810403442557964676) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6772
-        versionName = "0.67.72"
+        versionCode = 6773
+        versionName = "0.67.73"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/services/DownloadService.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/services/DownloadService.kt
@@ -55,9 +55,6 @@ import org.ole.planet.myplanet.utils.UrlUtils.header
 @AndroidEntryPoint
 class DownloadService : Service() {
     @Inject
-    lateinit var dispatcherProvider: DispatcherProvider
-
-    @Inject
     lateinit var downloadRepository: DownloadRepository
 
     @Inject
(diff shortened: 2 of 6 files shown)


Okuro3499 (approved):

dogi:
**Review — rating: 86/100**

Removes four unused injections: `DownloadService.dispatcherProvider`, `SyncManager.teamsRepository`, `UploadCoordinator`'s `@ApplicationContext context`, and `SubmissionsFragment.userSessionManager`.

The correctness argument here runs the right way round: an unused `private val` compiles fine, so a green build *before* the change would prove nothing — but these were removed and both flavors plus both test shards still compile, which does prove nothing referenced them. I also checked the one change that could go wrong silently: `SyncManager`'s constructor is positional, and dropping `teamsRepository` shifts every argument after it. `SyncManagerTest` drops the matching entry so the trailing `mockk(), mockk(), mockk(relaxed = true)` still line up with `userSyncRepository`, `syncRepository`, `syncTimeLogger`. That's correct, and it's the kind of thing that would otherwise mis-bind mocks without failing to compile.

**Nit:** `DownloadService.kt` keeps `import org.ole.planet.myplanet.utils.DispatcherProvider` (line 47) after the field is gone — it's the only other reference in the file, so it's now unused. Kotlin only warns, but `SubmissionsFragment` in this same diff drops both of its now-dead imports, so this looks like an oversight rather than a choice; worth matching given the repo runs a `kotlin-importing` skill for exactly this.

**Verdict:** mergeable — the leftover import is cosmetic and doesn't need another round on its own, though it's a one-line fix if you're touching the branch anyway.

---

dogi:
**Promoted to `merge`** — Okuro3499 approved, my review stands at 86/100, no changes since.

---

PR #16323, opened 2026-08-27
sync: smoother url utils auth header caching (fixes #16303)

### LINKED ISSUE #16303: Memoize the Basic-auth header instead of rebuilding it on every request
`UrlUtils.header` is a getter that on **every** access reads two SharedPreferences strings, allocates a `ByteArray` and Base64-encodes it. It has ~33 call sites including one per outgoing API call and one per list-item bind in `CoursesAdapter`/`CourseDetailFragment` (Glide `LazyHeaders` per cover image). Credentials change only on login/server switch.

**Ask:** cache the computed header, invalidating it on credential change (login/server switch), and clear the cache in the existing `resetForTesting()` hook so tests stay isolated.

**Files:** `utils/UrlUtils.kt`

**Verification:** confirmed at `UrlUtils.kt:29-38`.

dogi:
Memoize Basic-auth header in UrlUtils and invalidate on credential change

Fixes #16303

---
*PR created automatically by Jules for task [1413503547621873635](https://jules.google.com/task/1413503547621873635) started by @dogi*

## Summary by CodeRabbit

- **Bug Fixes**
  - Updated server URL and credential changes now immediately apply to authentication requests.
  - Prevented stale authorization headers from being reused after connection settings are changed.
- **Performance**
  - Reused computed authentication headers when credentials remain unchanged, reducing unnecessary recalculation.
- **Tests**
  - Added coverage verifying header reuse, invalidation, and regeneration after preference updates.


Okuro3499 (changes requested):
```
File: app/src/main/java/org/ole/planet/myplanet/services/sync/ServerUrlMapper.kt, function updateUrlPreferences (lines ~82-84)                                                   
    Problem: Writes url_user/url_pwd directly to a raw SharedPreferences.Editor, bypassing SharedPrefManager.setUrlUser/setUrlPwd, so UrlUtils.invalidateHeaderCache() is never      
    called when the app fails over to an alternative/clone server.                                                                                                                   
    Fix: Either route this write through SharedPrefManager.setUrlUser/setUrlPwd (preferred — keeps a single choke point for credential writes), or call                              
    UrlUtils.invalidateHeaderCache() explicitly right after editor.apply {} here. Given multiple callers (ServerReachabilityWorker, ChatRepositoryImpl, ConfigurationsRepositoryImpl,
    TeamsRepositoryImpl, SubmissionsUploader), fixing it centrally in ServerUrlMapper is safer than patching each call site.                                                         
    Add a regression test exercising ServerUrlMapper.updateUrlPreferences (or updateServerIfNecessary) followed by UrlUtils.header, asserting the header reflects the new credentials
    — the current UrlUtilsTest additions don't cover this path. 
```

dogi:
**Review — rating: 88/100**

Memoizes `UrlUtils.header` behind double-checked locking and invalidates on credential change. The memoization is textbook — `@Volatile` field, null-check outside the lock, re-check inside, immutable `String` so publication is safe.

The thing that actually decides whether a cache like this is correct is whether *every* mutation path invalidates, and here that's provable rather than hopeful: `URL_USER` and `URL_PWD` are `private const` inside `SharedPrefManager` (lines 38–39), so nothing outside the class can write them. That leaves exactly three writers, and all three are covered — `setUrlUser`, `setUrlPwd`, and `clearPreferences`, whose `pref.edit { clear() }` wipes both keys and now ends with `invalidateHeaderCache()`. A stale header here would mean silent 401s across the whole sync path, so it's worth stating that the coverage is closed by construction, not by inspection.

Ordering is right too, and it's easy to get backwards: both setters write the pref *before* invalidating. Since `edit { }` updates the in-memory map synchronously, a reader racing the invalidation recomputes from the new value rather than re-caching the old one.

Tests cover both the memoization (one `getUrlUser`/`getUrlPwd` call across two reads) and that invalidation actually yields a different header, and `@Before` calls `resetForTesting()` before `init()`, so the new tests don't depend on execution order despite `UrlUtils` being an object with static state — worth noting given CI shards by class.

**Nits:** `init()` is now the one public entry point that swaps the credential *source* without clearing the derived cache. It's harmless today — `MainApplication` calls it exactly once at startup (via the `UrlUtils.init` static import, which is why it doesn't grep as `UrlUtils.init(`) — but adding `cachedHeader = null` there closes the footgun for free. Also, the new tests call `io.mockk.verify` fully qualified rather than importing it.

**Verdict:** mergeable.

---

dogi:
**Correction to my review above — @Okuro3499 is right, and my central claim was wrong.**

I argued the invalidation coverage was "closed by construction" because `URL_USER`/`URL_PWD` are `private const` in `SharedPrefManager`, so only its three setters could write those keys. That reasoning had a hole: I grepped for the *constant identifiers*, which only matches `SharedPrefManager`. The keys are also written as **string literals** elsewhere:

```kotlin
// ServerUrlMapper.kt:82-90
editor.apply {
    putString("url_user", urlUser)
    putString("url_pwd", urlPwd)
    …
    apply()
}
```

That's a raw `SharedPreferences.Editor` write that never goes through `setUrlUser`/`setUrlPwd`, so `invalidateHeaderCache()` is never called. A repo-wide grep for `"url_user"|"url_pwd"` confirms `ServerUrlMapper.kt:83-84` is the only such site — but it's the one that matters most, because it runs exactly when the app **fails over to an alternative/clone server**. Credentials change, the cache doesn't, and `UrlUtils.header` keeps serving the previous server's Basic-auth header — the silent-401s-across-the-sync-path failure I described as impossible, on the one code path where it's most likely to bite.

Before this PR the getter recomputed every call, so the raw write was harmless. The memoization is what turns it into a bug, which makes it this PR's to fix.

@Okuro3499's preferred fix is the right one: route the write through `SharedPrefManager.setUrlUser`/`setUrlPwd` so credential writes keep a single choke point, rather than adding an `invalidateHeaderCache()` call after this `editor.apply {}` — the latter works but leaves the next raw writer free to reintroduce the same bug. The suggested regression test is also the right shape: drive `updateUrlPreferences` (or `updateServerIfNecessary`), then read `UrlUtils.header` and assert it reflects the new credentials. Nothing in the current `UrlUtilsTest` additions covers that path, which is why my review and the tests both passed it.

I'm moving this from `ready` to `change` accordingly. Revised rating: **68/100** (was 88) — the memoization and its invalidation-on-the-known-setters are still well built; the gap is the unguarded write path.

---

dogi:
## Review: 68/100 — needs a change

The invalidation coverage is genuinely thorough — I went looking for a missed writer and there isn't one. But the memoization has a lost-update race on the credential path, and the thing it's optimizing is close to free.

### The invalidation set is complete ✅

I enumerated every writer of `url_user` / `url_pwd` in `main`:

| writer | covered? |
|---|---|
| `SharedPrefManager.setUrlUser` / `setUrlPwd` (:168, :171) | ✅ this PR |
| └ callers: `ProcessUserDataActivity:146-147`, `ServerConfigUtils:115-116` | ✅ via the setters |
| `ServerUrlMapper.kt:83-84` — `putString("url_user"/"url_pwd")`, the **only** direct-key write that bypasses the setters | ✅ this PR (:93) |
| `SharedPrefManager` clear-preferences path (:294) | ✅ this PR |

That direct write in `ServerUrlMapper` is exactly the one a careless version of this change would have missed, and it's handled. Credit where due.

### The race ⚠️

`invalidateHeaderCache()` sets `cachedHeader = null` **outside** the monitor that `header` computes under. So:

1. **T1** (a request thread) reads `cachedHeader` → `null`, enters `synchronized(this)`, calls `spm.getUrlUser()` / `getUrlPwd()` → reads the **old** credentials.
2. **T2** (sync thread) runs `ServerUrlMapper.updateUrlPreferences`: commits the **new** credentials, then calls `invalidateHeaderCache()` → sets `cachedHeader = null`. It doesn't take the monitor, so it doesn't wait for T1.
3. **T1** finishes its block and writes `cachedHeader = <header built from the old credentials>`.

The cache is now poisoned with pre-switch credentials and **nothing clears it** until the next credential change or process death. Every request signs with the wrong password → sticky 401s.

This isn't hypothetical for this codebase: the window needs an in-flight `UrlUtils.header` read concurrent with a credential write, and `updateUrlPreferences` fires **during a sync**, which is precisely when many threads are reading `header` to sign requests. The symptom — intermittent auth failures against the alternative server, long after the switch — is about as hard to diagnose as bugs get.

Note that simply wrapping `invalidateHeaderCache()` in `synchronized(this)` **narrows but doesn't close** it: T1 can read old creds, then acquire the monitor after T2 has released it, and still write stale. Closing it needs the pref read and a generation snapshot under the same monitor:

```kotlin
@Volatile private var cachedHeader: String? = null
@Volatile private var generation = 0

fun invalidateHeaderCache() = synchronized(this) { generation++; cachedHeader = null }

val header: String
    get() {
        cachedHeader?.let { return it }
        val gen: Int; val user: String; val pwd: String
        synchronized(this) { gen = generation; user = spm().getUrlUser(); pwd = spm().getUrlPwd() }
        val computed = basicAuthHeader(user, pwd)
        synchronized(this) { if (generation == gen) cachedHeader = computed }
        return computed
    }
```

### Is the cache worth it at all?

What's being memoized:

```kotlin
fun basicAuthHeader(username: String, password: String): String {
    val credentials = "$username:$password".toByteArray()
    return "Basic ${Base64.encodeToString(credentials, Base64.NO_WRAP)}"
}
```

Two `SharedPreferences.getString` calls — in-memory `HashMap` lookups, since `SharedPreferences` holds the whole file in memory after load — plus a Base64 encode of ~40 bytes. Sub-microsecond. And `UrlUtils.header` is read **once per request**, immediately before a Retrofit call or a Glide load, so the saving is invisible against an HTTP round trip. The closest thing to a hot path is `CoursesAdapter:299` building a `GlideUrl` per bound row, and even there an image decode dwarfs it by orders of magnitude.

So the trade as written is: a process-wide mutable credential cache, a correctness-critical invalidation contract that every future writer of those two prefs must remember, and a lost-update race — in exchange for something unmeasurable. If there's a profile showing this mattering, it belongs in the description; if there isn't, dropping the cache is the stronger change.

### Tests

Both new tests are real regression tests (call-count assertions prove memoization and invalidation, and would fail on `master`), though they're white-box by necessity. Two nits: `assert(firstHeader != thirdHeader)` should be `assertNotEquals` — Kotlin's `assert` is a no-op unless assertions are enabled, and relying on Gradle's default `enableAssertions` is fragile; and `io.mockk.verify` is called fully-qualified inline rather than imported.

Neither test can catch the race, which is the point above: the bug is invisible to single-threaded tests.

**Verdict:** either close the race with a generation counter, or drop the memoization and keep just the (correct, useful) audit of credential writers. As it stands the cache trades a real intermittent auth failure for an unmeasurable saving.

---

dogi:
@openhands taking this over from Jules. Please read this whole comment before starting — the state is unusual.

## Why the handover

This branch is seven commits deep from Jules and is stuck on infrastructure, not on code. **Since commit 4 (`2b9267f`), the `push` event on this branch has stopped creating workflow runs.** Commits 1–3 each got `labels` + `build` + `test`; commits 4–7 got only `labels`, which survives because `labels.yml` is `pull_request_target` and does not depend on the push event. The `build`/`test` runs were never *created* — not queued, not cancelled — so there is nothing to re-run, and re-pushing has failed four times. I manually dispatched `test.yml` (it accepts `workflow_dispatch`); `build.yml` does not, so it still has no run on the current head.

**Please report in your reply whether your push produces `build` and `test` runs.** A push from a different actor is the one variable we have not tried, and that fact is worth more than the diff.

## What the review asked (68/100, 2026-08-28 13:44)

Two items. The invalidation coverage was found complete and correct — do not redo that part.

1. **A lost-update race.** `invalidateHeaderCache()` set `cachedHeader = null` outside the monitor that `header` computes under, so a request thread could read old credentials, then write them into the cache *after* a sync thread invalidated it — poisoning the cache with pre-switch credentials until process death. Symptom: sticky 401s against the alternative server, long after failover. The reviewer's suggested shape:

```kotlin
@Volatile private var cachedHeader: String? = null
@Volatile private var generation = 0

fun invalidateHeaderCache() = synchronized(this) { generation++; cachedHeader = null }

val header: String
    get() {
        cachedHeader?.let { return it }
        val gen: Int; val user: String; val pwd: String
        synchronized(this) { gen = generation; user = spm().getUrlUser(); pwd = spm().getUrlPwd() }
        val computed = basicAuthHeader(user, pwd)
        synchronized(this) { if (generation == gen) cachedHeader = computed }
        return computed
    }
```

Note the reviewer's warning: merely wrapping `invalidateHeaderCache()` in `synchronized` narrows the window but does not close it.

2. **Whether the cache is worth having at all.** What is memoized is two in-memory `SharedPreferences.getString` calls plus a Base64 encode of ~40 bytes, read once per request immediately before an HTTP round trip. The reviewer's position: if there is no profile showing this matters, dropping the memoization entirely — and keeping only the (correct, valuable) audit of credential writers — is the stronger change.

Jules's later commits claim to have added generation tracking. **Verify that against the current diff rather than trusting the commit message**, and either finish item 1 correctly or make the case for item 2. Also two test nits: `assert(firstHeader != thirdHeader)` should be `assertNotEquals` (Kotlin's `assert` is a no-op unless assertions are enabled), and `io.mockk.verify` is called fully-qualified inline instead of imported.

If you conclude no change is warranted, push nothing and say why.

---

olevim:

olevim:
I reviewed @dogi's two reviews (88/100 and 68/100), @Okuro3499's CHANGES_REQUESTED review, and the linked issue #16303. Here's the state and what I changed.

## Already correct in the current head (verified against the diff, not commit messages)

**Race fix (dogi, item 1).** `UrlUtils` already implements the generation-counter pattern exactly as suggested: `invalidateHeaderCache()` is `synchronized(this) { generation++; cachedHeader = null }`, and the `header` getter snapshots `generation` inside the monitor while reading prefs, then only stores `cachedHeader` if `generation == currentGen`. This *closes* the race rather than narrowing it — a writer that invalidates between T1's pref-read and T1's cache-write bumps the generation, so the stale value is dropped, not stored. ✅ no change needed.

**Okuro3499's instruction.** The direct `editor.apply { putString("url_user"…); putString("url_pwd"…) }` write in `ServerUrlMapper.updateUrlPreferences` is followed by `UrlUtils.invalidateHeaderCache()`, and `ServerUrlMapperTest.updateUrlPreferences invalidates UrlUtils cached header` is a regression test driving that path then asserting the new credentials. ✅ addressed via the explicit-invalidate option dogi endorsed.

**Test nits.** `assertNotEquals(firstHeader, thirdHeader)` is used (not `assert`), and `verify` is imported and called unqualified. ✅ already fixed.

## The one change I made — `f5434b4`

dogi's carried-forward nit (from the 88/100 review): `init()` swapped the credential *source* (`spmInstance`) without clearing the derived cache. So value-changes invalidated but source-swaps did not — an incomplete invalidation contract and a latent footgun if `init` is ever called again after startup. Closed it by wrapping the swap in `synchronized(this)`, bumping `generation`, and nulling `cachedHeader` — mirroring `resetForTesting()`/`invalidateHeaderCache()`. Minimal, same lock the getter uses, no behavior change for the single startup call.

```kotlin
fun init(sharedPrefManager: SharedPrefManager) {
    synchronized(this) {
        generation++
        spmInstance = sharedPrefManager
        cachedHeader = null
    }
}
```

## Why I kept the cache (dogi item 2)

dogi raised "drop the memoization entirely" as the alternative. The linked issue #16303 — authored by dogi — explicitly asks for the cache (`UrlUtils.header` has ~33 call sites including per list-item Glide binds in `CoursesAdapter`/`CourseDetailFragment`). With the race now correctly closed, the cache carries no correctness risk, so dropping it would contradict the issue's stated ask. Keeping it is the call that matches the issue.

## CI — the workflow state dogi asked about

**Yes, this push produced `build` and `test` runs.** Commits 4–7 (Jules) got only `labels`; this push from a different actor got `labels` + `build` (33205349069) + `test` (33205349075). Both are green — test both shards passed, build default+lite passed. That confirms the different-actor-push hypothesis: it was the one untried variable.

_This comment was created by an AI agent (OpenHands) on behalf of @olevim._

dogi:
## Review — rating 85/100 (was 68)

The lost-update race is properly fixed, and with the right pattern rather than a wider lock.

### The generation guard closes it

```kotlin
synchronized(this) {
    currentGen = generation
    user = spm.getUrlUser(); pwd = spm.getUrlPwd()
}
val computed = basicAuthHeader(user, pwd)
synchronized(this) {
    if (generation == currentGen) { cachedHeader = computed }
}
```

Walking the race I flagged last time:

1. Thread A reads credentials at `generation == N` and leaves the monitor.
2. Thread B calls `setUrlPwd` → `invalidateHeaderCache()` → `generation = N+1`, cache nulled.
3. Thread A computes a header from the **old** credentials, re-enters the monitor, sees `N+1 != N`, and **declines to publish**.
4. The next reader finds `cachedHeader == null` and recomputes from the new credentials.

The stale header still reaches thread A's own caller — unavoidable, since it read those credentials before the switch — but it can no longer poison the cache for everyone else, which was the sticky-401 failure mode. `@Volatile` on `cachedHeader` makes the unsynchronized fast path safe, and the double-check inside the monitor is correct.

### All write paths are covered

I checked this rather than assuming, since a missed writer would silently reintroduce the bug:

```
SharedPrefManager.kt:168  setUrlUser   → invalidates  (callers: ProcessUserDataActivity:146, ServerConfigUtils:115)
SharedPrefManager.kt:171  setUrlPwd    → invalidates  (callers: ProcessUserDataActivity:147, ServerConfigUtils:116)
ServerUrlMapper.kt:83-84  raw putString("url_user"/"url_pwd") → invalidates explicitly
SharedPrefManager.clearPreferences    → invalidates
```

Those are the only writers of `url_user`/`url_pwd` in `app/src/main`. Catching the `ServerUrlMapper` path is the easy one to miss — it bypasses the setters entirely with a raw editor, which is precisely why it needed its own call.

The tests are discriminating: `verify(exactly = 1)` then `exactly = 2` after invalidation genuinely fails without the cache and without the invalidation respectively, and the `ServerUrlMapper` test covers the raw-editor path end to end.

### One test worth adding

Nothing exercises the generation guard itself — all three tests are single-threaded, so `if (generation == currentGen)` could be deleted and they would all still pass. You can make it deterministic without threads, since `synchronized` is reentrant on the same thread:

```kotlin
every { mockSpm.getUrlPwd() } answers {
    UrlUtils.invalidateHeaderCache()   // a credential switch lands mid-read
    "pass1"
}
val stale = UrlUtils.header            // computed from pass1, must not be cached

every { mockSpm.getUrlUser() } returns "user2"
every { mockSpm.getUrlPwd() } returns "pass2"
assertNotEquals(stale, UrlUtils.header)
```

That fails if the guard is removed, and it is the only part of this change that is genuinely subtle.

Moving to `ready`.

---

Okuro3499 (approved):

PR #16324, opened 2026-08-27
sync: smoother realtime flow buffer managing (fixes #16306)

### LINKED ISSUE #16306: Raise RealtimeSyncManager's buffer so burst table updates aren't dropped
`MutableSharedFlow<TableDataUpdate>(extraBufferCapacity = 1)` with `tryEmit` and no replay: a sync notifying many tables in a burst silently drops updates for slow or momentarily-absent collectors, and the UI live-refresh helpers collect this flow.

**Ask:** raise `extraBufferCapacity` (and/or add bounded replay) so a burst of table-update notifications during sync doesn't silently drop updates for collectors that are momentarily behind or not yet subscribed.

**Files:** `services/sync/RealtimeSyncManager.kt`, `test/.../RealtimeSyncManagerTest.kt`

**Verification:** confirmed verbatim — the whole class is 25 lines and matches this description exactly.

dogi:
Raised `extraBufferCapacity` to 64 on `RealtimeSyncManager._dataUpdateFlow` and set `onBufferOverflow = BufferOverflow.DROP_OLDEST`, while keeping `replay = 0`.

`notifyTableUpdated` uses `tryEmit` on a flow with a single buffer slot and the default `SUSPEND` overflow, so a sync burst against a busy main thread silently discards the newest "this table changed, re-query" signals (the buffer kept stale pending signals and threw away the freshest). The larger capacity absorbs bursts for already-active collectors, and `DROP_OLDEST` ensures that under sustained pressure `tryEmit` never fails and subscribers see the most recent updates rather than the oldest.

Replay is intentionally left at 0: this is a change-notification flow consumed via `collectWhenStarted` (`repeatOnLifecycle(STARTED)`), so any replay would re-fire stale refreshes at every resubscriber on each foreground/navigation transition. A per-consumer high-water-mark would be the right tool if the goal were catching updates while a screen is stopped; a blanket replay cache is not.

Updated `RealtimeSyncManagerTest`:
- `testBurstTableUpdatesNotDropped` now collects on a `StandardTestDispatcher` (parked while emissions happen) so it genuinely exercises the buffer — it fails on the old `extraBufferCapacity = 1` config and passes at 64.
- `testNewSubscriberDoesNotReceiveHistoricalUpdatesWhenReplayIsZero` guards that replay stays 0.

Fixes #16306

---
*PR created automatically by Jules for task [15597840311090374455](https://jules.google.com/task/15597840311090374455) started by @dogi*


Okuro3499 (changes requested):
```
File: app/src/main/java/org/ole/planet/myplanet/services/sync/RealtimeSyncManager.kt:15-18                                                                                       
    Problem: replay = 16 causes redundant, unbounded-in-time re-delivery of historical TableDataUpdate events to every fragment/ViewModel that resubscribes via                      
    repeatOnLifecycle(STARTED), triggering spurious loadDataAsync()/RecyclerView refreshes on ordinary navigation and app-resume, not just on genuine new data.                      
    Fix: Drop replay back to 0 (or keep it at 0) and rely solely on the larger extraBufferCapacity = 64 to absorb bursts for already-active collectors. If the goal is genuinely to  
    avoid missing updates while a screen is stopped, that needs to be solved with a per-consumer high-water-mark (e.g., a last-seen version/timestamp Room already tracks per table) 
    rather than a shared replay cache — not a blanket SharedFlow replay on a process-lifetime singleton.
```

dogi:
**Review — rating: 84/100**

Two changes in one line: `extraBufferCapacity` 1 → 64, and `replay` 0 → 16. The first is a straightforward fix; the second is a real behaviour change, so I traced it through every consumer.

**The buffer bump is justified.** `notifyTableUpdated` calls `tryEmit` and discards the boolean, so with a single buffer slot any emission that arrives while a subscriber is still processing the previous one is silently dropped — exactly the burst pattern a table-by-table sync produces. 64 slots makes that effectively impossible.

**Replay is the part that needed checking**, because `collectWhenStarted` is `repeatOnLifecycle(STARTED)`, so every consumer re-subscribes on each foreground transition and will now immediately receive up to 16 historical updates. I looked at all five consumers and the blast radius is contained by construction: `TeamViewModel` and `ChatViewModel` go through `updatesFor(table)`; `MyHealthFragment` guards on `update.table == "health" && shouldRefreshUI`; `UserRepositoryImpl.achievementUpdates` filters to `"achievements"`; and `RealtimeSyncMixin` — the heaviest consumer — filters to its watched tables, then `distinctUntilChanged` on (table, counts), then `debounce(300)`. So a 16-event replay collapses to at most one refresh per matching table, and usually zero.

That leaves the intended win: a sync that lands while a screen is stopped now refreshes it on resume instead of leaving stale data on screen. The cost is one redundant re-query per resume when a matching event sits in the replay cache, which for these screens is roughly what you'd want on resume anyway.

**Nits:**
1. `testBurstTableUpdatesNotDropped` doesn't test what its name claims. The collector runs on `UnconfinedTestDispatcher`, so it resumes inside each `tryEmit` and drains the buffer synchronously — occupancy never exceeds one slot, and the test passes against the *old* `extraBufferCapacity = 1` too. To actually pin the buffer you need a collector that doesn't drain eagerly (a suspending collector, or a `StandardTestDispatcher` with delayed `runCurrent`). `testNewSubscriberReceivesReplayedUpdates` is the real regression test here — it emits before subscribing and genuinely fails at `replay = 0`.
2. With 37 entities, a full sync can touch more than 16 tables while the UI is stopped, so the oldest updates get evicted and those screens miss their catch-up. Fine as a heuristic, worth knowing it isn't a guarantee.
3. `tryEmit`'s return value is still ignored, so a drop stays invisible. A `Log` on `false` would turn the next occurrence into evidence rather than a mystery.

**Verdict:** mergeable — no correctness problem, and the replay semantics hold up against every current consumer.

---

dogi:
**Moving `ready` → `change`** — @Okuro3499 requested changes on the `replay = 16` half.

We agree on the mechanics; we differed on whether the cost is acceptable, and that call is theirs. My review already quantified it: `collectWhenStarted` is `repeatOnLifecycle(STARTED)`, so every consumer re-subscribes on each foreground transition and replays up to 16 historical events, and `distinctUntilChanged` state resets per subscription so it doesn't suppress across resumes. I judged that tolerable because each consumer filters by table and `RealtimeSyncMixin` debounces; Okuro3499 judges the spurious refreshes on ordinary navigation too high a price. That's a reasonable line to draw, and they own the UX.

Their alternative is also better engineering for the actual goal. Replay is a blunt instrument for "don't miss updates while a screen is stopped": it re-delivers to *every* resubscriber regardless of what each has already seen, and — as I noted — a full sync touching more than 16 tables evicts the oldest anyway, so it isn't even a guarantee. A per-consumer high-water-mark keyed on something Room already tracks per table gives each collector exactly the updates it missed, with no redundant delivery and no cache-size cliff.

**The `extraBufferCapacity` 1 → 64 half stands on its own** and is worth keeping: `notifyTableUpdated` uses `tryEmit` and discards the boolean, so a burst against a single buffer slot drops updates silently. Splitting that out would land immediately.

Revised rating: **72/100** (was 84).

---

dogi:
## Review: 76/100 — needs a change

The one-line buffer raise fixes a real dropped-signal bug. Three things need fixing around it, and one of them is that the headline test doesn't test the change.

### The bug being fixed is real ✅

`notifyTableUpdated` uses `tryEmit` on a `MutableSharedFlow(replay = 0, extraBufferCapacity = 1)` with the default `SUSPEND` overflow — so `tryEmit` **returns false and discards the value** whenever the single buffer slot is occupied. The emitter is the sync path (background thread, one `notifyTableUpdated` per table as each finishes); the consumers collect on the **main** dispatcher — `MyHealthFragment:252` via `collectWhenStarted`, `RealtimeSyncMixin:31`, `TeamViewModel:41`, `ChatViewModel:68`, `UserRepositoryImpl:82`. A sync burst against a busy main thread overruns one slot immediately, and the dropped signals are exactly the "this table changed, re-query" notifications a screen needs. 64 is sensible headroom for a full-table sync.

It's worth noting `updatesFor(table)` filters **downstream** of the shared flow, so every subscriber receives every table's update and discards most of them — which means one slow subscriber causes drops for all of them. That amplifies the value of the fix.

### 1. The drop policy is still backwards ⚠️

With `SUSPEND` + `tryEmit`, the value discarded on overflow is the **new** one — the buffer keeps stale pending signals and throws away the freshest. For a "something changed, go refresh" signal that's precisely inverted. Since these are conflatable notifications, pair the capacity with an explicit policy:

```kotlin
private val _dataUpdateFlow = MutableSharedFlow<TableDataUpdate>(
    extraBufferCapacity = 64,
    onBufferOverflow = BufferOverflow.DROP_OLDEST
)
```

`tryEmit` then never fails, and under sustained pressure subscribers see the most recent updates rather than the oldest. Raising the buffer alone makes the wrong-drop rarer; this makes it not matter.

### 2. `testBurstTableUpdatesNotDropped` passes on `master` ⚠️

The collector is launched on an `UnconfinedTestDispatcher`, so it resumes **inline** on the emitting call: each `tryEmit` hands the value straight to the collector, which appends to `results` and re-suspends before `tryEmit` returns. The buffer never holds more than one element, so all 30 arrive **with `extraBufferCapacity = 1`** — the test can't distinguish 1 from 64 and is green on `master`.

To actually exercise the change the collector must not drain synchronously:

```kotlin
val results = mutableListOf<TableDataUpdate>()
val job = launch(StandardTestDispatcher(testScheduler)) {
    manager.dataUpdateFlow.collect { results.add(it) }
}
runCurrent()                              // let the subscriber register
updates.forEach { manager.notifyTableUpdated(it) }   // emitted while collector is parked
advanceUntilIdle()                        // now let it drain
assertEquals(30, results.size)            // fails on master: only ~1-2 survive
```

That's the test that earns the change.

### 3. The title no longer matches the diff ⚠️

The PR is titled *"Raise RealtimeSyncManager flow buffer **and add replay capacity**"*, but no replay is added — and `testNewSubscriberDoesNotReceiveHistoricalUpdatesWhenReplayIsZero` explicitly asserts replay stays 0. Dropping the replay half looks like the right call (replay on a change-notification flow would re-fire stale refreshes at every new subscriber, and `collectWhenStarted` re-subscribes on **every** foreground transition — so replay would trigger a spurious re-query each time a fragment resumes). But the title becomes the squash-merge commit message, so it needs updating to match. The replay-is-zero test is a good guard to keep.

**Verdict:** add `DROP_OLDEST`, rewrite the burst test so it fails on `master`, and fix the title. The capacity change itself is right.

---

dogi:
@openhands taking this over from Jules. Please read this whole comment before starting — the state is unusual.

## Why the handover

This branch is six commits deep from Jules and is stuck on infrastructure, not on code. **Since commit 4, the `push` event on this branch has stopped creating workflow runs.** The early commits each got `labels` + `build` + `test`; the later ones got only `labels`, which survives because `labels.yml` is `pull_request_target` and does not depend on the push event. The `build`/`test` runs were never *created* — not queued, not cancelled — so there is nothing to re-run, and re-pushing has failed repeatedly. I manually dispatched `test.yml` (it accepts `workflow_dispatch`); `build.yml` does not, so it still has no run on the current head.

**Please report in your reply whether your push produces `build` and `test` runs.** A push from a different actor is the one variable we have not tried, and that fact is worth more than the diff.

## What the review asked (76/100, 2026-08-28 13:45)

The core change — raising `extraBufferCapacity` to 64 on `RealtimeSyncManager._dataUpdateFlow` — was found correct and valuable. `notifyTableUpdated` uses `tryEmit` on a flow with one buffer slot and `SUSPEND` overflow, so a sync burst against a busy main thread silently discards "this table changed, re-query" signals. Do not undo that.

Three follow-ups; verify each against the **current** diff, since Jules has pushed several times since the review:

1. **The drop policy is inverted.** With `SUSPEND` + `tryEmit`, the value discarded on overflow is the *newest* one — the buffer keeps stale pending signals and throws away the freshest, which is backwards for a change notification. Pair the capacity with `onBufferOverflow = BufferOverflow.DROP_OLDEST` so `tryEmit` never fails and subscribers see the most recent updates.

2. **`testBurstTableUpdatesNotDropped` passes on `master`.** It collects on an `UnconfinedTestDispatcher`, which resumes inline on each emitting call, so the buffer never holds more than one element and the test cannot distinguish capacity 1 from 64. The version that earns the change:

```kotlin
val results = mutableListOf<TableDataUpdate>()
val job = launch(StandardTestDispatcher(testScheduler)) {
    manager.dataUpdateFlow.collect { results.add(it) }
}
runCurrent()                                        // let the subscriber register
updates.forEach { manager.notifyTableUpdated(it) }  // emitted while the collector is parked
advanceUntilIdle()                                  // now let it drain
assertEquals(30, results.size)                      // fails on master: only ~1-2 survive
```

3. **The title contradicts the diff.** It still reads *"Raise RealtimeSyncManager flow buffer and add replay capacity"*, but no replay is added, and `testNewSubscriberDoesNotReceiveHistoricalUpdatesWhenReplayIsZero` explicitly asserts replay stays 0. Dropping the replay half was the right call — replay on a change-notification flow would re-fire stale refreshes at every new subscriber, and `collectWhenStarted` re-subscribes on every foreground transition. But this title becomes the squash-merge commit message, so please retitle to describe the actual change. Keep the replay-is-zero test as a guard.

If you conclude no change is warranted, push nothing and say why.

---

olevim:

olevim:
Reviewed the current head against @dogi's 76/100 follow-ups (2026-08-28 13:45) and @Okuro3499's changes-requested review.

**No code change warranted — pushed nothing.** All three code/test items are already satisfied in the current head (`09d89bb`):

1. **Drop policy** — `onBufferOverflow = BufferOverflow.DROP_OLDEST` is present alongside `extraBufferCapacity = 64`, so `tryEmit` never fails and subscribers see the most recent updates rather than the stale ones the old `SUSPEND` policy kept. ✅
2. **Burst test** — `testBurstTableUpdatesNotDropped` now collects on a `StandardTestDispatcher` with `runCurrent()`/`advanceUntilIdle()`, so the collector is parked while the 30 emissions happen and the buffer is genuinely exercised. It fails on the old `extraBufferCapacity = 1` config (only ~1–2 survive) and passes at 64. ✅
3. **Replay** — per @Okuro3499's request, replay stays at 0 (dropped the `replay = 16` half). `testNewSubscriberDoesNotReceiveHistoricalUpdatesWhenReplayIsZero` guards that. ✅

The one remaining item was metadata, not code: the title/body still claimed "add replay capacity of 16" while the diff adds none. I've re-titled to **"Raise RealtimeSyncManager flow buffer and drop oldest on overflow (fixes #16306)"** and rewritten the body to describe the actual change (capacity 64 + `DROP_OLDEST`, replay held at 0) — so the squash-merge commit message will match the diff.

A note on the earlier nit about logging `tryEmit`'s return value: with `DROP_OLDEST` + a 64-slot buffer, `tryEmit` always returns `true`, so a `Log` on `false` would be dead code — not added.

**Re: @dogi's CI question** — since no commit was pushed there is no new `push` event from me, so no new `build`/`test` runs to report. The existing manually-dispatched `test.yml` run on this head remains the standing result; this title/body edit touches PR metadata only and triggers no workflows.

_This comment was created by an AI agent (OpenHands) on behalf of @olevim._

dogi:
## Review — rating 87/100 (was 76)

All three points from the last pass are addressed, and the description is now one of the better ones in this queue — it explains the mechanism rather than restating the diff.

**`DROP_OLDEST` adopted.** The important consequence is one the body states correctly: with `SUSPEND` and a full buffer, `tryEmit` returns `false` and discards the *newest* signal while keeping stale pending ones — exactly backwards for a "this table changed, re-query" notification. Under `DROP_OLDEST`, `tryEmit` never fails and the freshest update always survives. 64 slots comfortably covers a full-sync burst across ~37 entities.

**The test is now discriminating.** Collecting on a `StandardTestDispatcher` with `runCurrent()` subscribes the collector without letting it consume, so the 30 emissions genuinely pile up against the buffer. On master's `extraBufferCapacity = 1` the second `tryEmit` onward returns `false` and the assertion fails; at 64 all 30 arrive in order. That is the test I was asking for.

**The title now matches the diff** — "Raise RealtimeSyncManager flow buffer and drop oldest on overflow" is exactly what the change does.

The reasoning for keeping `replay = 0` is right and worth having written down, since it is the non-obvious half:

> any replay would re-fire stale refreshes at every resubscriber on each foreground/navigation transition

That is the same `repeatOnLifecycle(STARTED)` re-subscription behaviour that made `replay = 1` a trap elsewhere in this codebase. `testNewSubscriberDoesNotReceiveHistoricalUpdatesWhenReplayIsZero` passes on master too, but as a guard against a future well-meaning replay it earns its place.

Nit: that second test's `launch(UnconfinedTestDispatcher())` omits `testScheduler`, so the collector runs on an unrelated scheduler. It works because `Unconfined` starts eagerly, but passing `testScheduler` would make it deliberate rather than incidental.

`mergeable_state` is `blocked` pending @Okuro3499's review, not a conflict.

Moving to `ready`.

---

Okuro3499 (approved):

PR #16331, opened 2026-08-27
actions: smoother dependabot handling (fixes #16302)

### LINKED ISSUE #16302: Group dependabot updates so one bump is not four CI runners
`.github/dependabot.yml` has two daily ecosystems with limits 10 and 15 and **no** `groups:` key, so every dependency bump opens its own PR, each firing `build.yml` (2-leg matrix) and `test.yml` (2-shard matrix) = four runners per bump, re-fired on every rebase, and each has to be drained through automerge one at a time.

**Ask:** add a `groups:` key to both ecosystems so compatible updates land in one PR instead of one-per-dependency.

**Files:** `.github/dependabot.yml`

**Verification:** no `groups` key present today; `open-pull-requests-limit` is 10 and 15 for the two ecosystems.

dogi:
Group Dependabot updates in .github/dependabot.yml for both github-actions and gradle ecosystems using wildcard patterns so updates land in consolidated PRs rather than opening one PR per dependency.

Fixes #16302

---
*PR created automatically by Jules for task [15076691839306247072](https://jules.google.com/task/15076691839306247072) started by @dogi*

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,6 +8,10 @@ updates:
     commit-message:
       prefix: "actions:"
     open-pull-requests-limit: 10
+    groups:
+      actions:
+        patterns:
+          - "*"
   - package-ecosystem: "gradle"
     directories:
       - "/"
@@ -17,3 +21,7 @@ updates:
     commit-message:
       prefix: "all:"
     open-pull-requests-limit: 15
+    groups:
+      dependencies:
+        patterns:
+          - "*"


Okuro3499 (approved):

dogi:
**Review — rating: 83/100**

Adds a catch-all `groups` block to both ecosystems so Dependabot bundles updates into one PR each instead of up to 10 actions PRs and 15 Gradle PRs per run. The schema is right — `groups` is valid under each `updates` entry, `patterns: ["*"]` matches everything, and it composes fine with the plural `directories:` on the Gradle entry. Given the open queue is currently 143 PRs, cutting a potential 25 daily down to 2 is the single highest-leverage thing in this batch.

**The trade-off worth naming:** `patterns: ["*"]` is all-or-nothing. Today a bad update fails its own PR and the other 14 still merge; after this, one broken dependency turns the whole group red and *no* Gradle update lands until someone works out which member broke it — and bisecting a grouped bump means editing the group PR by hand, since Dependabot won't split it for you. On a repo whose merge path is an automerge drain, that's a stall with no obvious owner.

The usual way to keep most of the noise win without that exposure is to group by risk rather than by everything, leaving majors to arrive alone:

```yaml
groups:
  dependencies:
    update-types: ["minor", "patch"]
```

Majors are where breakage actually lives, and they're a small share of the volume — so this still collapses the large majority of PRs while keeping the risky ones individually revertible.

Also now vestigial: `open-pull-requests-limit: 10` / `15` can no longer bind once grouping is on. Harmless, just no longer meaningful.

**Verdict:** mergeable as-is — the grouping is correct and the queue relief is real. I'd suggest the `update-types` split before merging, but it's a judgement call on your tolerance for a blocked dependency lane, not a defect.

---

dogi:
**Promoted to `merge`** — Okuro3499 approved, my review stands at 83/100, no changes since.

---

PR #16332, opened 2026-08-27
resources: smoother repository dao querying (fixes #16304)

### LINKED ISSUE #16304: Push MyLibrary.needToUpdate() from Kotlin into SQL
`needToUpdate() = !resourceOffline || (resourceLocalAddress != null && _rev != downloadedRev)` is evaluated in memory at five sites in `ResourcesRepositoryImpl`: `getLibraryListForUser` (194), `countLibrariesNeedingUpdate` (215 — a full row hydration to produce one `Int` for the dashboard badge), `getAllLibrariesToSync` (431), `getDownloadSuggestionList` (464, 470). All are expressible in SQLite with `IS NOT` for the null-safe rev compare.

**Ask:** push the predicate into `MyLibraryDao` queries. Also note: `getAllLibrariesToSync`'s `getSyncable()` + filter misses offline rows with a stale `_rev` — worth fixing as part of the same pass since it shares the same predicate.

**Files:** `data/room/dao/MyLibraryDao.kt`, `repository/ResourcesRepositoryImpl.kt`, `test/.../ResourcesRepositoryImplTest.kt`

**Verification:** all five call sites confirmed at the cited lines.

dogi:
Pushed the needToUpdate() predicate into MyLibraryDao SQL queries and updated ResourcesRepositoryImpl call sites.

Fixes #16304

---
*PR created automatically by Jules for task [584544609886955195](https://jules.google.com/task/584544609886955195) started by @dogi*


Okuro3499 (approved):

dogi:
**Review — rating: 87/100**

Pushes `MyLibrary.needToUpdate()` into `MyLibraryDao` SQL and turns one count into `COUNT(*)` instead of materialising rows to call `.count {}`. I checked the translation field by field against `needToUpdate()` (MyLibrary.kt:149):

- `resourceOffline` is a non-null `Boolean`, so `resourceOffline = 0` is an exact match for `!resourceOffline`.
- `resourceLocalAddress: String?` → `IS NOT NULL`, exact.
- Kotlin's `&&` binding tighter than `||` matches the explicit parens in the SQL.
- **`_rev IS NOT downloadedRev` is the part that had to be right, and is.** Both columns are nullable, and SQL `!=` yields NULL (falsy) when either side is NULL — which would have silently dropped exactly the rows where `downloadedRev IS NULL` and `_rev` is set, i.e. resources synced but never downloaded. `IS NOT` reproduces Kotlin `!=` across all four null combinations, and matches the `IS`-not-`=` convention in CLAUDE.md.

The tests are the right shape for a pushed-down predicate: `ResourcesRepositoryLibrarySyncTest` exercises the real DAO against real rows and covers all four boundaries — online, offline-with-stale-rev, offline-current, offline-without-local-address. That tests the SQL rather than a mock of it, which is the only way this class of change can actually be verified.

**One behaviour change worth stating plainly, because the diff reads like a pure refactor and this part isn't.** For the user-pattern call sites the push-down is exactly equivalent. `getSyncable()` is not: the old `getSyncable().filter { it.needToUpdate() }` was a **no-op filter**, since every `resourceOffline = 0` row satisfies `needToUpdate()` on the first clause — so the old effective result was just `resourceOffline = 0`. The new query is a strict superset that also returns offline rows with a stale rev, and it feeds `downloadAllFiles()` at MainApplication.kt:447 and SyncActivity.kt:572. So this widens automatic downloading.

I think that's a fix rather than a regression — the vacuous filter shows the predicate was always the intent — but I checked the obvious way it could go wrong on an offline-first app, and it holds: the sync parse path sets `downloadedRev = _rev` whenever the file is present on disk (MyLibrary.kt:262–265), so rows arriving from a sync or rebuilt after a destructive migration come back already aligned. No mass re-download on first run after this ships; the added rows are the narrow set where `_rev` moved without passing through that branch.

Worth a line in the PR description so the widened download scope is on the record.

**Verdict:** mergeable.

---

dogi:
**Promoted to `merge`** — Okuro3499 approved, my review stands at 87/100, no changes since.

---

PR #16333, opened 2026-08-27
all: smoother network module connection pooling (fixes #16305)

### LINKED ISSUE #16305: Size the OkHttp connection pool to the dispatcher's per-host limit
`NetworkModule.buildOkHttpClient` raises `Dispatcher.maxRequestsPerHost` to 20 but leaves `ConnectionPool` at OkHttp's default of 5 idle connections. Every sync and batch download targets a single CouchDB host, so 20 requests run concurrently and at most 5 connections survive for reuse — the other 15 pay a fresh TCP+TLS handshake, which dominates on the slow links this app targets.

**Ask:** size `ConnectionPool` to match (or reasonably cover) the 20 max-requests-per-host limit.

**Files:** `di/NetworkModule.kt`

**Verification:** `MAX_REQUESTS_PER_HOST=20` at `NetworkModule.kt:69`, `.dispatcher(dispatcher)` at `:76` with no `.connectionPool()` call today.

dogi:
Sized the OkHttp ConnectionPool in NetworkModule to match MAX_REQUESTS_PER_HOST (20) with a 5-minute keep-alive duration, ensuring up to 20 idle connections can be reused across concurrent CouchDB requests without incurring unnecessary TCP+TLS handshake overhead. Added a unit test in NetworkModuleTest to verify client dispatcher and connection pool configuration.

Fixes #16305

---
*PR created automatically by Jules for task [18404190358324513234](https://jules.google.com/task/18404190358324513234) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6773
-        versionName = "0.67.73"
+        versionCode = 6774
+        versionName = "0.67.74"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/di/NetworkModule.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/di/NetworkModule.kt
@@ -14,6 +14,7 @@ import java.util.concurrent.TimeUnit
 import javax.inject.Qualifier
 import javax.inject.Singleton
 import javax.net.SocketFactory
+import okhttp3.ConnectionPool
 import okhttp3.Dispatcher
 import okhttp3.OkHttpClient
 import org.ole.planet.myplanet.data.api.ApiInterface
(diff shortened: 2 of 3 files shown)


Okuro3499 (approved):

dogi:
**Review — rating: 82/100**

Sets an explicit `ConnectionPool(MAX_REQUESTS_PER_HOST, 5, TimeUnit.MINUTES)` so the idle-connection cap matches the dispatcher's `maxRequestsPerHost` instead of OkHttp's default 5.

The reasoning holds up. `maxIdleConnections` doesn't cap concurrency — OkHttp opens as many connections as the dispatcher allows — it caps how many stay *warm* for reuse. So the pre-existing config could run 20 concurrent requests to the host, then discard 15 of those connections the moment they went idle, forcing fresh TCP + TLS handshakes on the next batch. During a table-by-table sync that's exactly the wrong shape. Keeping the 5-minute keep-alive means only the idle cap changes, which is the minimal edit. The cost is ~20 idle sockets held for five minutes; idle TCP with no traffic doesn't hold the radio awake, so on the low-end devices this app targets that's a fair trade for skipping repeated handshakes.

One nuance for the record: OkHttp's `maxIdleConnections` is a pool-wide total, not per-host. This client also reaches the alternative/clone URL and the AI endpoints, so the warm set per host will in practice be under 20 — still strictly better than the default 5, just not a guaranteed 20 per host.

**Nit — the new test doesn't test the change.** `assertNotNull(okHttpClient.connectionPool)` passes identically before and after this PR, since `OkHttpClient` always has a pool; and `assertEquals(20, dispatcher.maxRequestsPerHost)` pins pre-existing behaviour. That's not the author's fault — OkHttp 5 doesn't expose `maxIdleConnections` on the public `ConnectionPool` type (it lives on the internal `RealConnectionPool`), so the configured value genuinely isn't assertable through the public API. I'd either drop the vacuous assertion or leave a one-line comment saying the pool size isn't publicly observable, so a later reader doesn't mistake it for coverage.

**Verdict:** mergeable.

---

dogi:
**Promoted to `merge`** — Okuro3499 approved, my review stands at 82/100, no changes since.

---

PR #16321, opened 2026-08-27
all: less hilt entry point accessors is more (fixes #16298)

dogi:
This change deletes six unused Hilt entry-point accessors (`apiInterface()`, `syncManager()`, `uploadManager()`, `uploadToShelfService()`, `userSessionManager()`, and `resourcesRepository()`) from `ServiceDependenciesEntryPoint.kt` and `CoreDependenciesEntryPoint.kt`, along with their unused imports.

Fixes #16298

---
*PR created automatically by Jules for task [13969260321779749803](https://jules.google.com/task/13969260321779749803) started by @dogi*

PR #16335, opened 2026-08-27
all: smoother url utils credential handling (fixes #16309)

### LINKED ISSUE #16309: Give the duplicated getUserInfo credential parser one platform-free home in UrlUtils
The same `uri.userInfo` splitter exists twice — `ProcessUserDataActivity.getUserInfo` (:239, compiling a `Regex` to split on one literal char) and the private `ServerUrlMapper.getUserInfo` (:111). Worse, `ServerConfigUtils.kt:99` — a plain util — reaches into an **Activity companion object** to call it, so a util transitively depends on an `AppCompatActivity`.

**Ask:** consolidate both copies into one platform-free helper in `UrlUtils` (returning e.g. a `Pair<String, String>` for username/password), have `ProcessUserDataActivity` and `ServerUrlMapper` delegate to it, and remove `ServerConfigUtils`'s dependency on the Activity companion object.

**Files:** `utils/UrlUtils.kt`, `ui/sync/ProcessUserDataActivity.kt`, `services/sync/ServerUrlMapper.kt`, `utils/ServerConfigUtils.kt`, `test/.../UrlUtilsTest.kt`

**Verification:** all three copies confirmed at the cited locations.

dogi:
Consolidated the duplicated `getUserInfo` URI user credential parsing logic into a single platform-free helper method `UrlUtils.getUserInfo`. Refactored `ProcessUserDataActivity`, `ServerUrlMapper`, and `ServerConfigUtils` to delegate to `UrlUtils.getUserInfo`, removing `ServerConfigUtils`'s dependency on `ProcessUserDataActivity`. Added comprehensive unit tests in `UrlUtilsTest`.

Fixes #16309

---
*PR created automatically by Jules for task [4490425251343478589](https://jules.google.com/task/4490425251343478589) started by @dogi*


dogi:
**Review — rating: 87/100**

Collapses three copies of the userinfo parser into one `UrlUtils.getUserInfo(String?)`. I diffed the two originals against the new one and the consolidation is behaviour-preserving: `ServerUrlMapper` used `split(":")` and `ProcessUserDataActivity` used `split(":".toRegex())`, but `:` isn't a regex metacharacter so both produce identical lists, and the `dropLastWhile { it.isEmpty() }` + `size > 1` guard is carried over verbatim. Taking `uri.userInfo` as the parameter instead of a `Uri` is the right seam — it makes the function pure and testable without an Android type.

The nicer structural win is one the description undersells: `ServerConfigUtils` (a `utils` object) no longer imports `ProcessUserDataActivity` (a UI activity) just to borrow a string parser. That dependency pointed the wrong way through the layers, and it's gone.

Six tests covering null, empty, `admin:secret`, no-colon, `admin:`, and multiple-colons is the right coverage for this — those are precisely the inputs where the three implementations could have silently diverged. Two behaviours they now pin are worth knowing about, both inherited rather than introduced:

- `"admin:"` yields `("", "")`, not `("admin", "")` — a username with an empty password is discarded entirely.
- `"admin:secret:extra"` yields `("admin", "secret")` — a password containing `:` is silently truncated at the first colon, which would present as an unexplained auth failure. Now that there's exactly one parser, that's a one-line fix (`split(":", limit = 2)`) if you ever want it, rather than three.

**Nit:** the consolidation leaves `ProcessUserDataActivity.getUserInfo` behind as a companion wrapper, but this PR rewires its last two callers — its own line 127 and `ServerConfigUtils` line 99 — to call `UrlUtils` directly, so it has zero callers afterwards. Deleting it finishes the job, and matches what #16321 and #16322 are doing elsewhere in this same queue.

**Verdict:** mergeable.

---

Okuro3499 (approved):

dogi:
**Promoted to `merge`** — Okuro3499 approved, my review stands at 87/100, no changes since.

---

PR #16339, opened 2026-08-27
chat: smoother history view modelling (fixes #16319)

### LINKED ISSUE #16319: Drop the duplicate chat-history load on screen open
`ChatHistoryFragment` calls `refreshChatHistory()` from `onViewCreated` (:98) and again from `onResume` (:151). `onResume` always follows `onViewCreated`, so opening the AI chat screen runs two full loads back to back — chat history, shared news messages, share targets, viewIn ids, plus three SharedPreferences reads each. A third trigger already exists for real changes (`refreshChatSignal` at :224).

**Ask:** remove the redundant call so the screen loads once on open, not twice — keep the real-change trigger (`refreshChatSignal`) intact.

**Files:** `ui/chat/ChatHistoryFragment.kt`

**Verification:** all three call sites confirmed at lines 98, 151, and 224.

dogi:
## Description

`ChatHistoryFragment` called `refreshChatHistory()` from both `onViewCreated` (:98) and `onResume` (:151). Since `onResume` always follows `onViewCreated`, opening the AI chat screen ran two full loads back to back — chat history, shared news messages, share targets, viewIn ids, plus three `SharedPreferences` reads each. A third trigger already exists for real changes (`refreshChatSignal` at :224).

This removes the redundant `onResume` override so the screen loads once on open instead of twice, while keeping the real-change trigger (`refreshChatSignal`) intact.

## Files

- `ui/chat/ChatHistoryFragment.kt` — removed the `onResume` override that duplicated the load, and dropped the explicit `refreshChatHistory()` call from `onViewCreated` (the initial load is now driven by the seeded signal — see below)
- `ui/chat/ChatViewModel.kt` — gave `_refreshChatSignal` a `replay = 1` so a sync update that fires while the fragment is stopped is buffered and delivered on the next foreground transition, **and** seeds it with `tryEmit(Unit)` in `init` so a cold ViewModel still triggers one initial load when the fragment subscribes (replacing the old explicit `onViewCreated` load)
- `ui/chat/ChatViewModelTest.kt` — behavioural tests: a missed chats update is still received by a subscriber that attaches after the emit; a cold ViewModel seeds an initial-load value

## Verification

- Confirmed the three call sites at lines 98, 151, and 224 before the change.
- The removed `onResume` override is covered by the existing refresh paths, so no
  behaviour is lost:
  - initial load — `setupRealtimeSync()` collects `refreshChatSignal`; the seeded replay value delivers one load on START (was: an explicit `onViewCreated` call at line 98, which alongside the replayed signal would have been a double load)
  - a chat created/edited in the detail screen — `DashboardActivity.refreshChatHistory()`
    (`DashboardActivity.kt:710`) resolves the history fragment and pushes a reload
  - sync-driven change — `refreshChatSignal` (collected in `setupRealtimeSync()`) re-emits on a `chats` `TableDataUpdate`
- Per review (72/100), `replay = 1` alone re-created the very per-resume reload this PR removes: `collectWhenStarted` is `repeatOnLifecycle(STARTED)`, which re-subscribes on every START, and a re-subscription to a `replay = 1` SharedFlow re-delivers the cached value — so `refreshChatHistory()` ran on every foreground transition, and on first open it ran twice (the explicit `onViewCreated` call plus the replayed signal). Seeding the signal in `init` and dropping the explicit `onViewCreated` load makes the replayed signal the sole load trigger: one load per foreground transition instead of two, while `replay = 1` still buffers a chats update that arrives while the fragment is stopped.
- The more idiomatic end state (folding "chats changed" into the existing `screenData` StateFlow and deleting the signal) is left out of scope here — `screenData` is pull-based, driven by fragment-supplied `userId`/`parentCode`/`communityName`, so that refactor would expand well beyond #16319.
- Per earlier review, the reflection-based regression test that asserted on `onResume` method metadata was dropped rather than kept as a misleading guard; the tests are behavioural (exercise the emit -> collect path).

---

_This PR was created by an AI agent (OpenHands) on behalf of the issue author._

Fixes #16319

@dogi can click here to [continue refining the PR](https://app.all-hands.dev/conversations/43187538-e9b7-41d6-bb7c-9897ec3fea3e)


dogi:
**Review — rating: 76/100**

**The production change is right, and I verified the refresh paths that have to cover for it.** Removing the `onResume` override is safe because every case it was serving is already handled:

- Initial load — `onViewCreated` calls `refreshChatHistory()` (line 98), and `onResume` always follows it, so the override was a guaranteed duplicate full load on first open.
- A chat created or edited in the detail screen — `ChatDetailFragment.kt:645` calls `(activity as DashboardActivity).refreshChatHistory()`, which resolves the history fragment and calls into it (`DashboardActivity.kt:724–727`). That's an explicit push on real change, which is strictly better than reloading on every resume and hoping.
- Sync-driven change — `ChatViewModel` collects `realtimeSyncManager.updatesFor("chats")`.

So the resume-time reload was pure waste: it re-ran a full history load every time the user came back from the detail screen, unlocked the phone, or switched apps. Good catch, and the `fixes #16319` framing matches.

**The test file is what holds this back.** Both tests assert on reflection metadata rather than behaviour:

```kotlin
assertFailsWith<NoSuchMethodException> {
    ChatHistoryFragment::class.java.getDeclaredMethod("onResume")
}
```

This pins the shape of the diff, not the contract. It passes whether or not chat history actually loads — the fragment could be entirely broken and this stays green. Worse, it encodes "this class may never override `onResume`" as a permanent rule: the next person with a legitimate reason to override it (re-registering a listener, resuming a player) gets a failure whose message points at duplicate loading, which won't be their problem. That's a landmine with a misleading label. The second test is tautological — `refreshChatHistory` is public and called from `DashboardActivity`, so the compiler already guarantees it exists.

Robolectric is already in the stack. A test that actually earns its place here launches the fragment and asserts `loadChatHistoryScreenData` is invoked **once** across `onViewCreated` → `onResume`, then once more after an external `refreshChatHistory()` call. That pins the real behaviour and stays true regardless of which lifecycle methods the class happens to override.

Either write that, or drop the file — the one-line deletion is self-evidently correct and doesn't need a guard.

**Verdict:** needs another round — production change is ready to merge as-is; only the test file needs to change.

---

Okuro3499 (dismissed):
```
 File: app/src/test/java/org/ole/planet/myplanet/ui/chat/ChatHistoryFragmentRefreshTest.kt                                                                                        
    Problem: The test only reflects on the class to assert onResume isn't declared (getDeclaredMethod("onResume") throwing NoSuchMethodException) and that refreshChatHistory exists.
    It doesn't exercise the fragment's lifecycle or verify loadChatHistoryScreenData/refreshChatHistory is actually called once instead of twice — it only encodes "this specific    
    override is absent from the source," which is redundant with reading the diff and would pass even if the duplicate-load bug were reintroduced through a different mechanism      
    (e.g., a call added on onStart, or a second call inside onViewCreated itself). It gives false confidence rather than regression protection. 
```

dogi:
> **Review — rating: 76/100**
> 
> **The production change is right, and I verified the refresh paths that have to cover for it.** Removing the `onResume` override is safe because every case it was serving is already handled:
> 
> * Initial load — `onViewCreated` calls `refreshChatHistory()` (line 98), and `onResume` always follows it, so the override was a guaranteed duplicate full load on first open.
> * A chat created or edited in the detail screen — `ChatDetailFragment.kt:645` calls `(activity as DashboardActivity).refreshChatHistory()`, which resolves the history fragment and calls into it (`DashboardActivity.kt:724–727`). That's an explicit push on real change, which is strictly better than reloading on every resume and hoping.
> * Sync-driven change — `ChatViewModel` collects `realtimeSyncManager.updatesFor("chats")`.
> 
> So the resume-time reload was pure waste: it re-ran a full history load every time the user came back from the detail screen, unlocked the phone, or switched apps. Good catch, and the `fixes #16319` framing matches.
> 
> **The test file is what holds this back.** Both tests assert on reflection metadata rather than behaviour:
> 
> ```kotlin
> assertFailsWith<NoSuchMethodException> {
>     ChatHistoryFragment::class.java.getDeclaredMethod("onResume")
> }
> ```
> 
> This pins the shape of the diff, not the contract. It passes whether or not chat history actually loads — the fragment could be entirely broken and this stays green. Worse, it encodes "this class may never override `onResume`" as a permanent rule: the next person with a legitimate reason to override it (re-registering a listener, resuming a player) gets a failure whose message points at duplicate loading, which won't be their problem. That's a landmine with a misleading label. The second test is tautological — `refreshChatHistory` is public and called from `DashboardActivity`, so the compiler already guarantees it exists.
> 
> Robolectric is already in the stack. A test that actually earns its place here launches the fragment and asserts `loadChatHistoryScreenData` is invoked **once** across `onViewCreated` → `onResume`, then once more after an external `refreshChatHistory()` call. That pins the real behaviour and stays true regardless of which lifecycle methods the class happens to override.
> 
> Either write that, or drop the file — the one-line deletion is self-evidently correct and doesn't need a guard.
> 
> **Verdict:** needs another round — production change is ready to merge as-is; only the test file needs to change.
> 
> _Generated by [Claude Code](https://claude.ai/code)_

and 

 File: app/src/test/java/org/ole/planet/myplanet/ui/chat/ChatHistoryFragmentRefreshTest.kt                                                                                        
    Problem: The test only reflects on the class to assert onResume isn't declared (getDeclaredMethod("onResume") throwing NoSuchMethodException) and that refreshChatHistory exists.
    It doesn't exercise the fragment's lifecycle or verify loadChatHistoryScreenData/refreshChatHistory is actually called once instead of twice — it only encodes "this specific    
    override is absent from the source," which is redundant with reading the diff and would pass even if the duplicate-load bug were reintroduced through a different mechanism      
    (e.g., a call added on onStart, or a second call inside onViewCreated itself). It gives false confidence rather than regression protection. 
    
@openhands

dogi:
@openhands-agent follow up on review and @Okuro3499 s improvement instructions

dogi:
@openhands follow up on review and @Okuro3499 s improvement instructions

dogi:
Followed up on the review and @Okuro3499's notes.

**What changed:** dropped `ChatHistoryFragmentRefreshTest.kt` (commit `3797555`) and updated the PR description to match.

**Why the test went away rather than being rewritten behaviourally:** both reviews flagged the same real problem — the file asserted on reflection metadata (`getDeclaredMethod("onResume")` throwing) instead of behaviour, so it stayed green whether or not history actually loaded and would trip a future, legitimate `onResume` override with a misleading "duplicate load" message. That's the false-confidence landmine called out.

The behavioural alternative — launching `ChatHistoryFragment` under Robolectric+Hilt and counting `loadChatHistoryScreenData` across `onViewCreated → onResume` — is the right idea in principle, but this fragment isn't cheap to launch: it's `@AndroidEntryPoint` (so it needs the full Hilt test component to inject `SharedPrefManager`), and its layout is a `SlidingPaneLayout` master-detail that also instantiates `ChatDetailFragment`. No `@AndroidEntryPoint` fragment is launched in any existing test in the repo, so that's new, heavy infra to stand up for a one-line deletion. Per @dogi's review ("the one-line deletion is self-evidently correct and doesn't need a guard … either write that, or drop the file"), the proportionate call was to drop it.

**The production change stands on its own**, verified against the three refresh paths the `onResume` override used to redundantly serve:
- initial open — `onViewCreated → refreshChatHistory()` (`ChatHistoryFragment.kt:98`)
- a chat created/edited in the detail screen — `DashboardActivity.refreshChatHistory()` (`DashboardActivity.kt:710`) resolves the history fragment and pushes a reload
- sync-driven change — `refreshChatSignal` (`ChatHistoryFragment.kt:218`) collected in `setupRealtimeSync()`

So removing the `onResume` override loses no behaviour; it only stops the guaranteed duplicate full load on every screen open. Ready for another look.

_This comment was created by an AI agent (OpenHands) on behalf of @dogi._

dogi:
## Review: 65/100 — needs a change

Deleting the blanket reload-on-every-resume is the right instinct, and most of the paths that replace it hold up. But one common path doesn't, and this five-line deletion is what makes it visible.

### What still refreshes, after the deletion

| trigger | covered? |
|---|---|
| first load / view recreation | ✅ `refreshChatHistory()` at `:98` in `onViewCreated` |
| navigating away and back | ✅ `FragmentNavigator.replaceFragment(..., addToBackStack = true)` destroys the view, so `onViewCreated` and `:98` re-run on pop |
| a chat created or edited in the detail pane | ✅ `ChatDetailFragment:645` → `DashboardActivity:724` → `fragment.refreshChatHistory()` |
| a **sync** that pulls new chats | ⚠️ see below |

The saving is real: `loadChatHistoryScreenData` re-fetches chat history *plus* news messages *plus* share targets on every foreground transition — every returned phone call, every dismissed dialog, every close of the sliding detail pane. And the adapter is built once in `observeScreenData()` and updated via `updateChatHistory`, so this was pure redundant fetching, not UI thrash.

### The sync path drops the signal while the fragment is stopped ⚠️

`ChatViewModel:64` declares:

```kotlin
private val _refreshChatSignal = MutableSharedFlow<Unit>()
```

All defaults — `replay = 0`, `extraBufferCapacity = 0`. The only subscriber is `setupRealtimeSync()`'s `collectWhenStarted(sharedViewModel.refreshChatSignal)`, and `collectWhenStarted` is `repeatOnLifecycle(Lifecycle.State.STARTED)`, so **it unsubscribes whenever the fragment drops below STARTED**. A `MutableSharedFlow` with no subscribers and `replay = 0` discards the value on `emit` and returns immediately — it doesn't buffer it and doesn't park the emitter.

So:

1. User is on the chat history screen; the app goes to background (view intact, fragment stopped, collector unsubscribed).
2. `AutoSyncWorker` pulls new chats → `realtimeSyncManager.updatesFor("chats")` emits → `ChatViewModel:71` calls `_refreshChatSignal.emit(Unit)` → **no subscriber, replay 0 → silently dropped**.
3. User returns. The view was never destroyed, so `:98` doesn't re-run, and the signal is gone.

On `master`, `onResume` reloaded and the new chats appeared. With this PR, **the list stays stale until the user navigates away and back**. Backgrounding-then-returning while a sync runs is not an edge case — it's the app's normal usage pattern.

**Fix, one line in `ChatViewModel`:**

```kotlin
private val _refreshChatSignal = MutableSharedFlow<Unit>(replay = 1)
```

`replay = 1` means a signal emitted while stopped is delivered when `collectWhenStarted` re-subscribes on the next foreground transition — exactly one refresh on return, which is what `onResume` was doing, only now conditional on something having actually changed. (`extraBufferCapacity = 1` with `DROP_OLDEST` works too; `replay = 1` is simpler and re-subscribe-friendly.)

### Related: this depends on #16324

The signal chain runs through `RealtimeSyncManager._dataUpdateFlow`, which on `master` is `MutableSharedFlow(extraBufferCapacity = 1)` fed by `tryEmit` — so under a sync burst it **drops** updates outright, including `"chats"`, before `ChatViewModel` ever sees them. #16324 raises that buffer to 64 and is the other half of making this deletion safe. Right now `onResume` is masking both defects at once.

Suggested order: **#16324 first**, then the `replay = 1` fix above, then this. Merged alone, this PR converts two latent notification-loss bugs into a visible stale-list bug.

### Also

No test accompanies the deletion. A `ChatViewModel` test asserting that a `"chats"` update with `shouldRefreshUI = true` emitted with no active subscriber is still received by a later subscriber would pin the `replay = 1` fix and fail on `master`.

**Verdict:** add `replay = 1` to `_refreshChatSignal` and land behind #16324; the deletion itself is right.

---

dogi:
@openhands follow up on my review and/or @Okuro3499 s improvement

---

dogi:
Followed up on the latest review (@dogi 65/100).

## The issue identified
Removing `onResume` left the sync-driven refresh dependent on `refreshChatSignal`, which was `MutableSharedFlow<Unit>()` — `replay = 0`, no buffer. `ChatHistoryFragment.setupRealtimeSync()` collects it via `collectWhenStarted` (`repeatOnLifecycle(STARTED)`), so while the fragment is stopped the collector is unsubscribed. A `chats` sync update firing in the background (`AutoSyncWorker → RealtimeSyncManager → ChatViewModel init → _refreshChatSignal.emit`) then had no subscriber to land on. On return the view was never destroyed, so `onViewCreated`/`:98` didn't re-run, and the list stayed stale until the user navigated away and back — exactly the case `onResume` was masking.

## The fix (commit ${COMMIT:0:7})
```kotlin
private val _refreshChatSignal = MutableSharedFlow<Unit>(replay = 1)
```
`replay = 1` buffers one missed update so it's delivered when `collectWhenStarted` re-subscribes on the next foreground transition — one refresh on return, conditional on a real change, rather than the unconditional reload-on-every-resume this PR removed. This is the one-line fix suggested in the review.

## Test
Added a behavioural `ChatViewModel` test (`refreshChatSignal delivers a missed update to a subscriber that attaches after the emit`) that emits a `chats` update with `shouldRefreshUI = true` while no `refreshChatSignal` subscriber is active, then asserts a subscriber attaching afterwards still receives it. It exercises the real emit → collect path (not reflection metadata), which is what the earlier reviews asked for, and fails on the old `replay = 0` flow.

## Note on #16324
As the review flagged, the upstream `RealtimeSyncManager._dataUpdateFlow` is `MutableSharedFlow(extraBufferCapacity = 1)` fed by `tryEmit`, so under a sync burst it can drop `"chats"` updates before `ChatViewModel` ever sees them. That's a separate, pre-existing defect addressed by #16324 (raises the buffer to 64 + adds replay). This PR's `replay = 1` fix covers the fragment-stopped stale-list regression this deletion would otherwise expose; #16324 covers the burst-drop case. I've left the dependency noted in the PR description rather than pulling #16324's change in here, to keep this PR scoped to #16319 — happy to reorder if you'd rather they land together.

Could not run the tests locally (no JDK in this environment); CI will confirm.

---
_This comment was created by an AI agent (OpenHands) on behalf of @dogi._

dogi:
## Review — rating 72/100 (was 65)

`replay = 1` does close the correctness hole I flagged: an emit with no subscriber now lands in the replay cache instead of being dropped, so a `chats` update arriving while the fragment is stopped is no longer lost forever. The new test is genuinely discriminating — on `replay = 0` the collector attaches after the emit, `signals` stays empty, and `assertEquals(1, signals.size)` fails. That is the right test for this fix.

The problem is that `replay = 1` also quietly re-creates the very `onResume` refresh this PR removes.

### The removed refresh comes back through the replay cache

`collectWhenStarted` is `repeatOnLifecycle(STARTED)` (`utils/FlowExtensions.kt:14-20`):

```kotlin
return viewLifecycleOwner.lifecycleScope.launch {
    viewLifecycleOwner.repeatOnLifecycle(Lifecycle.State.STARTED) {
        flow.collect(collector)
    }
}
```

`repeatOnLifecycle` cancels the inner `collect` on STOP and **re-subscribes on every START**. A re-subscription to a `replay = 1` SharedFlow immediately re-delivers the cached value, so `refreshChatHistory()` runs on every foreground transition — the same unconditional per-resume reload the deleted `onResume` was doing. Since `AutoSyncWorker` makes a `chats` `TableDataUpdate` a routine occurrence, the cache is populated in steady state and the two behaviours coincide.

It is actually a double load on first creation, because `onViewCreated` already refreshes directly:

```kotlin
setupRealtimeSync()      // subscribes; replayed signal -> refreshChatHistory()
checkAiProvidersIfNeeded()
...
refreshChatHistory()     // and again, explicitly
```

Master did the same two loads (`onViewCreated` + `onResume`), so this is not a regression — but it means the PR's headline change nets out to roughly nothing, and the mechanism is invisible in the diff. The next person trying to remove per-resume reloads will read `replay = 1` as an innocuous buffering tweak and not realise it *is* the resume refresh.

This is the replay-as-event antipattern: the signal is now state ("data changed at some point") while the collector still treats it as an event ("data just changed"), so it fires once per subscription rather than once per change.

### Suggested shape

Make `replay = 1` earn its place by letting the replayed signal *be* the initial load, so there is exactly one load per foreground transition:

1. Seed the cache in the ViewModel's `init`, before the `updatesFor("chats")` collector, so the signal is always populated even on a cold ViewModel that has never seen a sync:
   ```kotlin
   private val _refreshChatSignal = MutableSharedFlow<Unit>(replay = 1)
   init {
       _refreshChatSignal.tryEmit(Unit)   // initial load
       viewModelScope.launch { ... }
   }
   ```
2. Drop the explicit `refreshChatHistory()` from `onViewCreated` — the replayed signal covers it.

That gives one reload per start instead of two, keeps updates-while-stopped from being dropped, and makes the intent legible. Alternatively, fold "chats changed" into the existing `screenData` StateFlow and delete the signal entirely, which is the more idiomatic end state given the ViewModel already exposes `screenData`/`filteredChats` as StateFlows.

Nit: the test inlines `org.ole.planet.myplanet.model.TableDataUpdate` fully-qualified rather than importing it.

Moving to `change`.

---

olevim:

olevim:
Followed up on the latest review (72/100) and pushed commit 45a77ee.

## The problem the review flagged
`replay = 1` closed the missed-update hole, but it quietly re-created the very `onResume` reload this PR removes. `collectWhenStarted` is `repeatOnLifecycle(STARTED)`, which cancels the inner collect on STOP and re-subscribes on every START; a re-subscription to a `replay = 1` SharedFlow immediately re-delivers the cached value, so `refreshChatHistory()` ran on every foreground transition. On first open it ran **twice** — the explicit `onViewCreated` call (`:98`) plus the replayed signal delivered to the `setupRealtimeSync` collector — so the headline change ("loads once on open, not twice") netted out to roughly nothing, via a mechanism invisible in the diff. That is the replay-as-event antipattern: the signal became state ("data changed at some point") while the collector still treated it as an event ("data just changed").

## The fix (Option A from the review)
Let the replayed signal **be** the initial load, so there is exactly one load per foreground transition:

1. `ui/chat/ChatViewModel.kt` — seed the replay cache in `init` before the `updatesFor("chats")` collector:
   ```kotlin
   _refreshChatSignal.tryEmit(Unit)   // initial load
   ```
   so a cold ViewModel that has never seen a chats sync still has a value buffered.
2. `ui/chat/ChatHistoryFragment.kt` — drop the explicit `refreshChatHistory()` from `onViewCreated`. The fragment's `setupRealtimeSync()` subscription replays the seed on START and drives the one-and-only load.

Net: one load per foreground transition instead of two (the issue #16319 ask), `replay = 1` still buffers a chats update that arrives while the fragment is stopped, and the intent is now legible — the signal *is* the load trigger.

On a subsequent resume (view not recreated) the replayed value still drives a single reload, which is parity with master's `onResume` (a single reload) minus the duplicate-on-open — within the scope of #16319, which asked to drop the duplicate load on screen open, not to eliminate per-resume reloads entirely.

## Why not Option B (delete the signal, fold into `screenData`)
The review offered this as the more idiomatic end state, but `screenData` is pull-based — `loadChatHistoryScreenData(userId, parentCode, communityName)` is called with values the fragment reads from `sharedPrefManager`, which the ViewModel doesn't own. Folding the sync signal into `screenData` would require the ViewModel to source those params itself, which is a larger refactor well beyond #16319. Kept out of scope here; happy to do it as a follow-up if you'd like.

## Tests (`ui/chat/ChatViewModelTest.kt`)
- Added `refreshChatSignal seeds an initial load value for a cold ViewModel` — a freshly constructed ViewModel with no chats sync still delivers one value to a subscribing collector (locks in the seeded-initial-load behaviour; would fail without the `tryEmit(Unit)` seed).
- Updated `refreshChatSignal emits when RealtimeSyncManager emits chats update` to drain the seeded initial value before asserting on the sync-driven signal (the seed now occupies the replay cache on subscribe).
- `refreshChatSignal delivers a missed update to a subscriber that attaches after the emit` still passes: the buffered chats update replaces the seeded value in the `replay = 1` cache, so a subscriber attaching afterwards receives exactly one.
- Per the nit, `TableDataUpdate` is now imported instead of inlined fully-qualified.

Could not run tests locally (no JDK in this environment); CI will confirm.

---
_This comment was created by an AI agent (OpenHands) on behalf of @dogi._

dogi:
## Review — rating 86/100 (was 72)

This is exactly the shape I was hoping for, and the comments make the non-obvious part legible — which was my main complaint last round.

### The seed makes `replay = 1` coherent

```kotlin
init {
    // Seed the replay cache so a cold ViewModel still triggers one initial load when the
    // fragment subscribes — the replayed value replaces the explicit onViewCreated load.
    _refreshChatSignal.tryEmit(Unit)
```

`tryEmit` always succeeds on a `replay = 1` flow (the value goes to the replay cache regardless of subscribers), so the cache is non-empty from construction onward. That closes the gap in my earlier suggestion: a ViewModel that has never seen a `chats` sync still drives the first load.

The net effect is now one `loadChatHistoryScreenData` per foreground transition, down from two on first creation (`onViewCreated` + `onResume`) and two on every resume once a sync signal existed. The per-resume refresh is still there — it has to be, it is the safety net — but it happens once and through a single mechanism instead of two overlapping ones.

I checked the other callers: `ChatDetailFragment:645` → `DashboardActivity.refreshChatHistory()` → `fragment.refreshChatHistory()` is the post-send refresh path, and it is untouched by removing the `onViewCreated` call. Nothing else depended on the eager load.

### The tests are discriminating and the existing one was correctly adapted

- `seeds an initial load value for a cold ViewModel` collects on a fresh ViewModel and asserts one signal. On master (`replay = 0`, no seed) it gets zero and fails.
- `delivers a missed update to a subscriber that attaches after the emit` still fails on `replay = 0`. The `assertEquals(1, ...)` is right rather than lucky: the replay cache holds one value, so the sync emit *replaces* the seed rather than queueing behind it, and the comment says so.
- `emits when RealtimeSyncManager emits chats update` now drains and clears the seed before asserting, which is the correct adaptation rather than a loosened assertion.

Cleaning up the inline `org.ole.planet.myplanet.model.TableDataUpdate` into a real import while you were in there is appreciated.

Moving to `ready`.

---

Okuro3499 (approved):

PR #16340, opened 2026-08-27
all: smoother file utils checking (fixes #16312)

### LINKED ISSUE #16312: Read storage stats once per FileUtils helper, and stop creating directories during an existence check
`totalAvailableMemoryRatio` and `availableOverTotalMemoryFormattedString` each call both public wrappers, and each wrapper calls `getStorageStats` — so each helper runs the whole probe twice (two `getSystemService` lookups plus `StorageStatsManager` binder round trips) on the login screen, settings, storage breakdown and `TaskNotificationWorker`. Separately, `checkFileExist` resolves through `getSDPathFromUrl` -> `createFilePath`, which calls `mkdirs()` — a pure existence question writes directories, and it is called from list binding.

**Ask:**
- Collapse the double `getStorageStats` probe in both helpers to a single call.
- Stop `checkFileExist` from creating directories as a side effect of an existence check — resolve the path without calling `mkdirs()`.
- Do not add a time-based stats cache (a behavior change beyond scope here) — keep this a pure de-duplication.

**Files:** `utils/FileUtils.kt`

**Verification:** double-probe confirmed at `FileUtils.kt:343-357`; the `mkdirs()`-on-existence-check is the higher-value half of this fix.

dogi:
Optimized storage stats fetching in FileUtils by querying StorageStatsManager once per helper method call. Removed automatic directory creation from path resolution and file existence checks.

Fixes #16312

---
*PR created automatically by Jules for task [4940451286417140931](https://jules.google.com/task/4940451286417140931) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6774
-        versionName = "0.67.74"
+        versionCode = 6775
+        versionName = "0.67.75"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/services/DownloadService.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/services/DownloadService.kt
@@ -363,6 +363,7 @@ class DownloadService : Service() {
     private fun downloadFile(body: ResponseBody, url: String) {
         val fileSize = body.contentLength()
         val finalFile = FileUtils.getSDPathFromUrl(this@DownloadService, url)
+        finalFile.parentFile?.mkdirs()
         val tempFile = File(finalFile.parentFile, "${finalFile.name}.tmp")
         tempFile.delete()
         outputFile = finalFile
--- a/app/src/main/java/org/ole/planet/myplanet/services/DownloadWorker.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/services/DownloadWorker.kt
@@ -112,6 +112,7 @@ class DownloadWorker @AssistedInject constructor(
     private suspend fun downloadFileBody(body: ResponseBody, url: String, index: Int, total: Int) {
         val fileSize = body.contentLength()
         val outputFile: File = FileUtils.getSDPathFromUrl(context, url)
+        outputFile.parentFile?.mkdirs()
         var totalBytes: Long = 0
         var lastUpdateTime = 0L
 
--- a/app/src/main/java/org/ole/planet/myplanet/utils/FileUtils.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/utils/FileUtils.kt
@@ -48,39 +48,21 @@ object FileUtils {
         return File(externalFilesDir, "ole/$libraryId/$address")
     }
 
-    private fun createFilePath(context: Context, folder: String, filename: String): File {
+    private fun resolveFilePath(context: Context, folder: String, filename: String): File {
         val baseDirectory = File(getExternalFilesDir(context), folder)
 
-        if (filename.contains("/")) {
+        return if (filename.contains("/")) {
             val subDirPath = filename.substring(0, filename.lastIndexOf('/'))
             val fullDir = File(baseDirectory, subDirPath)
-
-            try {
-                if (!fullDir.exists() && !fullDir.mkdirs()) {
-                    throw IOException("Failed to create directory: ${fullDir.absolutePath}")
-                }
-            } catch (e: IOException) {
-                e.printStackTrace()
-                throw RuntimeException("Failed to create directory: ${fullDir.absolutePath}", e)
-            }
-
             val actualFilename = filename.substring(filename.lastIndexOf('/') + 1)
-            return File(fullDir, actualFilename)
+            File(fullDir, actualFilename)
         } else {
-            try {
-                if (!baseDirectory.exists() && !baseDirectory.mkdirs()) {
-                    throw IOException("Failed to create directory: ${baseDirectory.absolutePath}")
-                }
-            } catch (e: IOException) {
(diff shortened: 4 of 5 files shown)


dogi:
**Review — rating: 62/100**

Two independent changes. The storage-probe half is a clean win; the `mkdirs` half breaks resource downloading.

**The memory probes are good.** `totalAvailableMemoryRatio` and `availableOverTotalMemoryFormattedString` each called `totalMemoryCapacity()` *and* `totalAvailableMemory()`, and both of those call `getStorageStats()` — so each was doing two `StorageStatsManager` round trips. The destructuring is correct (`getStorageStats` returns `Pair(total, available)`, matching `.first`/`.second` at lines 343–345), and beyond halving the probes it makes the pair *consistent*: two separate reads could straddle a storage change and yield a ratio over 100% or a mismatched `available/total` string. Worth keeping as-is.

**The `mkdirs` removal is a regression.** Making path resolution pure is the right instinct — a `checkFileExist` probe should not create directories — but the deletion lands in the shared helper that **writers** also depend on, and nothing else fills the gap. `FileUtils.kt:59` and `:71` were the only `mkdirs` on this path; neither download writer has its own:

- `DownloadService.downloadFile` (line 365) resolves `finalFile`, builds `tempFile` in `finalFile.parentFile`, then opens `FileOutputStream(tempFile)` — which throws `FileNotFoundException` when the parent doesn't exist.
- `DownloadWorker.downloadFileBody` (line 114) resolves `outputFile` then calls `outputFile.sink()`, which is `FileOutputStream` underneath — same failure.

The target is `getExternalFilesDir()/ole/<libraryId>/…`, so the directory is missing for any library ID not already downloaded — i.e. every first download on a fresh install. CI won't catch it: there's no `androidTest` source set, and no unit test drives these writers.

**Fix** — keep the resolver pure and let the writers own creation:

```kotlin
// DownloadService.downloadFile
val finalFile = FileUtils.getSDPathFromUrl(this@DownloadService, url)
finalFile.parentFile?.mkdirs()

// DownloadWorker.downloadFileBody
val outputFile: File = FileUtils.getSDPathFromUrl(context, url)
outputFile.parentFile?.mkdirs()
```

A `FileUtils.getSDPathFromUrlForWrite()` that resolves then creates would work equally well and keeps the asymmetry explicit at the call site.

Also worth saying: the code being deleted was genuinely bad beyond the side effect — `e.printStackTrace()` and an `IOException` rewrapped as `RuntimeException`. Dropping that is an improvement; it's only the `mkdirs` itself that was load-bearing.

The new assertion in `checkFileExist_returnsFalseWhenFileDoesNotExist` correctly pins the read path. The gap is that no test covers the write path, which is exactly where the behaviour moved.

**Verdict:** needs another round — split the two halves or add the two `mkdirs` lines; the storage-probe change is mergeable as it stands.

---

dogi:
## Review: 86/100 — mergeable

Removing a filesystem side effect from what reads as a pure path resolver is the right call, and the writer coverage is complete. I checked every caller rather than trusting the two the diff touches.

### Every `getSDPathFromUrl` caller, and what it does

| caller | writes? | needs `mkdirs`? |
|---|---|---|
| `DownloadService:365` | ✅ writes `.tmp` then final file | ✅ added |
| `DownloadWorker:114` | ✅ writes the output file | ✅ added |
| `ConfigurationsRepositoryImpl:232` | read-only — `f.exists()`, then `Sha256Utils().getCheckSumFromFile(f)` | no |
| `DictionaryRepositoryImpl:42` | read-only — `getStringFromFile(...)` | no |
| `FileUtils:129` (`checkFileExist`) | read-only | no |

So exactly two writers, and both got the `parentFile?.mkdirs()`. In `DownloadService` it's correctly placed **before** `tempFile = File(finalFile.parentFile, "...tmp")`, which is the line that would have needed the directory first.

### The bug being fixed is user-visible

On `master`, `checkFileExist` → `getSDPathFromUrl` → `createFilePath` **created the directory tree** just to answer "does this file exist?" — always returning `false` right after, leaving an empty directory behind. Every resource the app has ever checked but never downloaded left litter under external storage. Given the app ships a storage-breakdown screen (`StorageBreakdownFragment`), that's junk a user could actually see and wonder about. `ConfigurationsRepositoryImpl`'s checksum check had the same side effect.

### The storage-stats dedup is correct

`getStorageStats` returns `Pair(totalBytes, availableBytes)` (:373), so `val (total, available) = getStorageStats(context)` maps to `(first, second)` — matching the old `totalMemoryCapacity` = `.first` / `totalAvailableMemory` = `.second`. Both rewritten functions now make one `StorageStatsManager` round trip instead of two, and — worth noting — they now read **a consistent snapshot**: previously `totalAvailableMemoryRatio` could divide an `available` from one instant by a `total` from another. Harmless in practice since `total` is stable, but strictly better.

### The test is a real regression test

Asserting the parent directory does **not** exist after `checkFileExist` fails on `master`, where the resolver creates it. Nice and direct.

**Nit:** it depends on nothing else in the class having already materialized `ole/123/resources/`. A per-test unique id in the URL would make it order-independent.

### One thing lost

The old code threw `RuntimeException("Failed to create directory: ${dir.absolutePath}")` when `mkdirs()` returned false. The new call sites do `parentFile?.mkdirs()` and discard the result, so a genuine failure (permissions, full volume) surfaces later as a bare `FileNotFoundException` from the stream open, without naming the directory. Not a behavior regression — the download still fails either way — but it's a diagnosability step backwards, and slightly against the grain of #16595's push for tagged, attributable logging. A cheap improvement at both sites:

```kotlin
val parent = outputFile.parentFile
if (parent != null && !parent.exists() && !parent.mkdirs()) {
    Log.e("DownloadWorker", "Failed to create directory: ${parent.absolutePath}")
}
```

**Contention:** #16441 also touches `FileUtils.getSDPathFromUrl`'s path parsing. Different hunks, but worth a re-check on whichever lands second.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16345, opened 2026-08-27
courses: smoother resources inline scoping (fixes #16310)

### LINKED ISSUE #16310: Move InlineResourceAdapter's preview file stats off the main thread
`adapterScope = CoroutineScope(SupervisorJob() + dispatcherProvider.main)` (:56, :61) and every preview path then does blocking filesystem stats on it: `file.exists()` at 203, 217, 227, 242, 254, 269 plus `getCacheKey` (:283) which adds `lastModified()` and `length()`. Lines 203/217 are invoked straight from `onBindViewHolder`, so a screenful of downloaded resources does several `stat()` syscalls per bind during a fling.

**Ask:** move the blocking file-stat calls off the main dispatcher (e.g. onto `dispatcherProvider.io`) while keeping UI updates on main.

**Files:** `ui/courses/InlineResourceAdapter.kt`

**Verification:** all line numbers confirmed exactly as cited.

dogi:
Move file.exists(), lastModified(), and length() calls in InlineResourceAdapter off the main dispatcher onto dispatcherProvider.io while keeping UI updates on main.

Fixes #16310

---
*PR created automatically by Jules for task [12617410181746357837](https://jules.google.com/task/12617410181746357837) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6775
-        versionName = "0.67.75"
+        versionCode = 6776
+        versionName = "0.67.76"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/ui/courses/InlineResourceAdapter.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/courses/InlineResourceAdapter.kt
@@ -5,6 +5,7 @@ import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
 import android.widget.ImageView
+import androidx.annotation.VisibleForTesting
 import androidx.recyclerview.widget.ListAdapter
 import androidx.recyclerview.widget.RecyclerView
 import com.bumptech.glide.Glide
@@ -64,7 +65,8 @@ class InlineResourceAdapter(
     }
 
     class ViewHolder(val binding: ItemInlineResourceBinding) : RecyclerView.ViewHolder(binding.root) {
-        private var previewJob: Job? = null
+        @get:VisibleForTesting
+        internal var previewJob: Job? = null
 
         fun cancelPreviousPreviews() {
             previewJob?.cancel()
--- a/app/src/main/java/org/ole/planet/myplanet/ui/courses/InlineResourceAdapter.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/courses/InlineResourceAdapter.kt
@@ -16,6 +16,7 @@ import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.cancel
 import kotlinx.coroutines.isActive
 import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
 import org.ole.planet.myplanet.R
 import org.ole.planet.myplanet.databinding.ItemInlineResourceBinding
 import org.ole.planet.myplanet.model.MyLibrary
@@ -176,20 +177,16 @@ class InlineResourceAdapter(
                 "ole/${resource.id}/${resource.resourceLocalAddress}"
             )
 
-            when {
-                mimeType?.startsWith("image") == true -> showImagePreview(binding, context, resourceFile)
-                mimeType?.startsWith("video") == true -> showVideoPreview(binding, context, resourceFile)
-                else -> {
-                    holder.setPreviewJob(adapterScope.launch {
-                        when {
-                            mimeType?.contains("pdf") == true -> showPdfPreview(holder, resourceFile)
-                            mimeType?.startsWith("audio") == true -> showAudioPreview(holder, resourceFile)
-                            mimeType?.contains("csv") == true || resource.resourceLocalAddress?.endsWith(".csv") == true -> showCsvPreview(holder, resourceFile)
-                            mimeType?.startsWith("text") == true || resource.resourceLocalAddress?.endsWith(".txt") == true || resource.resourceLocalAddress?.endsWith(".md") == true -> showTextPreview(holder, resourceFile)
-                        }
-                    })
+            holder.setPreviewJob(adapterScope.launch {
+                when {
+                    mimeType?.startsWith("image") == true -> showImagePreview(binding, context, resourceFile)
+                    mimeType?.startsWith("video") == true -> showVideoPreview(binding, context, resourceFile)
+                    mimeType?.contains("pdf") == true -> showPdfPreview(holder, resourceFile)
+                    mimeType?.startsWith("audio") == true -> showAudioPreview(holder, resourceFile)
+                    mimeType?.contains("csv") == true || resource.resourceLocalAddress?.endsWith(".csv") == true -> showCsvPreview(holder, resourceFile)
+                    mimeType?.startsWith("text") == true || resource.resourceLocalAddress?.endsWith(".txt") == true || resource.resourceLocalAddress?.endsWith(".md") == true -> showTextPreview(holder, resourceFile)
                 }
-            }
+            })
         } else {
             binding.pbDownload.visibility = View.VISIBLE
         }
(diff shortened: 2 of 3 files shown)


dogi:
**Review — rating: 80/100**

Moves `file.exists()` / `lastModified()` / `length()` off the main thread in `onBindViewHolder`. The threading is right, and I checked the thing that would have made this dangerous: `adapterScope` is built on `dispatcherProvider.main` (line 56), so the `launch` bodies stay on Main and only the `withContext(dispatcherProvider.io)` blocks hop off. Glide calls and visibility writes remain on the main thread — had the scope been Default or IO, folding the image/video branches into it would have moved view mutation off Main and crashed.

Three real improvements here, one of which the title doesn't claim:

- Disk stats leave the main thread — the actual point, and they were being done per-bind during scroll.
- **Image and video previews become cancellable.** They previously ran synchronously outside `setPreviewJob`, so a fast scroll could apply a Glide load to a holder already rebound to another item. Now every branch goes through `setPreviewJob`, which cancels the prior job (line 73–75), and `onViewRecycled` clears it. That's a latent recycling bug fixed in passing.
- `getFileCacheKeyIfExist` collapses what was `exists()` + `lastModified()` + `length()` — three separate stat calls, two of them on Main — into a single IO hop. Good consolidation, and the `?: return` preserves the old early-exit semantics exactly, including keeping the audio container's `visibility = VISIBLE` before the check.

**Nit — the tests promise more than they check.** `test bind text resource calls previewLoader on IO dispatcher` asserts only `tvResourceTitle.text`; `previewLoader` is a relaxed mock that's never verified, and nothing observes a dispatcher. `onViewRecycled cancels previous preview job` has no assertions at all and would pass with cancellation deleted. They aren't worthless — driving create/bind/recycle under Robolectric would catch an off-main-thread view mutation or an NPE in the bind path — but the names imply coverage that isn't there, which is how a future reader ends up trusting an uncovered path. Either rename them to what they are (smoke tests over the bind path) or add the assertions that match: `coVerify` the `previewLoader` call, and check `holder`'s job is cancelled after `onViewRecycled`.

One minor behavioural note: `showVideoPreview` now sets `videoThumbnailContainer.visibility` inside the coroutine, so it lands a dispatch later than before. Worth a glance that the bind path resets container visibility up front, otherwise a recycled holder can show the previous item's preview state for a frame.

**Verdict:** mergeable — production change is correct; the test naming is worth tidying but doesn't need another round.

---

Okuro3499 (dismissed):
```
File: app/src/test/java/org/ole/planet/myplanet/ui/courses/InlineResourceAdapterTest.kt                                                                                          
    Problem: Both dispatcherProvider.main and dispatcherProvider.io are wired to the same UnconfinedTestDispatcher instance (TestDispatcherProvider at                               
    app/src/test/java/org/ole/planet/myplanet/utils/TestDispatcherProvider.kt:9), so the test has no way to distinguish "ran on IO" from "ran on main" — yet the first test is named 
    test bind text resource calls previewLoader on IO dispatcher and asserts only holder.binding.tvResourceTitle.text, which is unrelated to dispatcher usage or even to             
    previewLoader being invoked at all. The second test, onViewRecycled cancels previous preview job, has zero assertions — it would pass even if                                    
    cancelPreviousPreviews()/setPreviewJob() were entirely broken
```

dogi:
**Moving `ready` → `change`** — @Okuro3499 requested changes on the tests, and on reflection they're right to treat this as blocking where I logged it as a nit.

We found the same thing: the first test asserts only `tvResourceTitle.text` and never verifies `previewLoader` was called, and the second has no assertions at all. Okuro3499 adds the detail that settles it — `TestDispatcherProvider` wires `main` and `io` to the **same** dispatcher instance, so a test named `…calls previewLoader on IO dispatcher` is structurally incapable of distinguishing IO from main. That's not a weak assertion; it's an unverifiable claim, and it's the exact claim the PR is titled after.

I passed this at 80 on the strength of the production change, which I still think is correct: I verified `adapterScope` is built on `dispatcherProvider.main` (line 56), so Glide and visibility writes stay on the main thread while only the stat calls hop to IO, and folding the image/video branches into `setPreviewJob` fixes a real recycling bug where a fast scroll could apply a Glide load to an already-rebound holder. None of that changes. But shipping a dispatcher-offload change whose tests can't observe dispatchers means the next regression here is silent, so the higher bar is the right one.

To make the tests real: give `main` and `io` distinct dispatchers so the two are distinguishable, `coVerify` the `previewLoader` invocation, and assert the holder's job is cancelled after `onViewRecycled`.

Revised rating: **74/100** (was 80).

---

dogi:
## Review: 82/100 — mergeable

Correct offload, and the two things that usually go wrong when moving work off the bind path both check out.

### Thread safety of the widened coroutine ✅

Folding image/video into the same `adapterScope.launch` as pdf/audio/csv/text means `Glide.with(...)` and every `View.visibility` write now happen inside a coroutine. That's safe here because the scope is

```kotlin
private var adapterScope = CoroutineScope(SupervisorJob() + dispatcherProvider.main)
```

so the launch body runs on **main** and only the `withContext(dispatcherProvider.io)` blocks hop off it, returning to main for the UI writes. If the scope had been `Default` this would have been a crash on every image row — worth stating explicitly since the diff alone doesn't show it.

### No stale-preview flash from recycling ✅

Making image/video async was my other concern: on `master` they were applied **synchronously** during bind, so a recycled holder was corrected in the same frame. It's fine because `updateStatusAndPreview` resets all six preview views to `GONE` synchronously *before* dispatching:

```kotlin
binding.ivResourcePreview.visibility = View.GONE
binding.videoThumbnailContainer.visibility = View.GONE
binding.tvTextPreview.visibility = View.GONE
binding.audioPreviewContainer.visibility = View.GONE
binding.pbDownload.visibility = View.GONE
binding.ivStatus.visibility = View.GONE
```

so a recycled row shows nothing, then the correct preview appears — never the previous row's.

Also a genuine side benefit: image/video rows now get a `previewJob`, so `onViewRecycled` cancels them too. On `master` only the text-ish branches were cancellable.

`showVideoPreview` and `showAudioPreview` still set their container `VISIBLE` **before** the existence check, exactly as on `master` — behavior preserved, not accidentally changed. And `getFileCacheKeyIfExist` folds `exists()` + `lastModified()` + `length()` into a single IO hop rather than three separate stats.

### The biggest main-thread stat is still on main ⚠️ (follow-up, not a blocker)

Two lines above the code this PR fixes:

```kotlin
val isDownloaded = resource.isResourceOffline() ||
    FileUtils.checkFileExist(context, UrlUtils.getUrl(resource))
```

`checkFileExist` does a synchronous `File.exists()` **on the main thread, on every bind of every row** — and unlike the preview stats (which only ran for already-downloaded rows) this one runs for every row that isn't flagged offline. So the function still does main-thread I/O per bind; the PR offloaded the smaller half.

It's worse than it looks on `master`: `checkFileExist` → `getSDPathFromUrl` currently **creates directories** as a side effect (that's what #16340 fixes), so today this line does a `mkdirs` per row bind on the main thread. #16340 removes the mkdirs; hoisting this check into the same `adapterScope.launch` would finish the job this PR starts.

### Tests

The three-phase dispatcher advance is a real regression test — `coVerify(exactly = 0) { previewLoader.getTextPreview(any()) }` after advancing **main only** fails on `master`, where `file.exists()` runs inline on main and the coroutine sails straight into `getTextPreview`. That's the right way to prove a dispatcher hop. The `onViewRecycled` test is a good companion.

**Two nits on testability plumbing:**

- The test reaches `externalFilesDir` by reflection (`getDeclaredField(...).isAccessible = true`). That breaks silently at runtime if the field is ever renamed. Since `previewJob` was already widened to `internal` for this same test class, widening `externalFilesDir` the same way would make it compile-checked and consistent.
- `@get:VisibleForTesting internal var previewJob` annotates only the getter while the setter is equally exposed. The repo's existing idiom (e.g. `ChatViewModel:59`) is `@VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)` on the property.

**Verdict:** mergeable. Worth a follow-up issue for the `checkFileExist` call, ideally sequenced after #16340.

---

PR #16353, opened 2026-08-27
all: smoother configurations repository sha256 utils checking (fixes #16352)

### LINKED ISSUE #16352: Remove android.util.Log from Sha256Utils
`getCheckSumFromFile` is pure `MessageDigest` over a `File` except for one `Log.w` on the failure path (:32) — the class's only `android` import, blocking an otherwise platform-free checksum helper.

**Ask:** remove the `Log.w` call and the `android.util.Log` import, keeping the helper platform-free (propagate/handle the failure some other way if the caller needs to know about it — check call sites first).

**Files:** `utils/Sha256Utils.kt`, `test/.../Sha256UtilsTest.kt`

**Verification:** confirmed `android.util.Log` at :3 is the only android import in the file.

dogi:
## Summary

`Sha256Utils.getCheckSumFromFile` was pure `MessageDigest` over a `File` except for a single `Log.w` on the failure path — the class's only `android` import, blocking an otherwise platform-free checksum helper.

This removes the `android.util.Log` dependency from the helper **and** closes a pre-existing fail-open on the integrity path that surfaced while doing so. The helper now returns `null` on failure instead of an empty string, so the sole caller no longer treats an unverifiable file as passing.

## Changes

- `utils/Sha256Utils.kt` — removed the `android.util.Log` import and the `Log.w(...)` call; `getCheckSumFromFile` now returns `String?` (`null` on failure) instead of an empty string. The helper stays platform-free.
- `repository/ConfigurationsRepositoryImpl.kt` — `checkCheckSum` now treats `null` as "cannot verify" → `false` (fail-closed), closing the `checksum.contains("")` hole that previously returned `true` for an unverifiable file. The lost `Log.w` diagnostic is re-surfaced here, where an Android dependency is already present.
- `test/.../Sha256UtilsTest.kt` — dropped the `mockkStatic(Log::class)` scaffolding (the helper no longer touches `Log`); failure-path tests now assert `null`.
- `test/.../ConfigurationsRepositoryImplTest.kt` — added `checkCheckSum returns false when checksum cannot be computed` covering the new fail-closed behavior.

## Test plan

- [x] `Sha256UtilsTest` updated to assert `null` on the failure paths (no `Log` mock needed).
- [x] `ConfigurationsRepositoryImplTest` added a fail-closed case for the `null` checksum.
- [ ] `./gradlew testDefaultDebugUnitTest` passes in CI.

_Note: the test suite could not be executed locally — no JDK/Gradle toolchain in this environment. CI will confirm._

fixes #16352

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6778
-        versionName = "0.67.78"
+        versionCode = 6779
+        versionName = "0.67.79"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/repository/ConfigurationsRepositoryImpl.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/ConfigurationsRepositoryImpl.kt
@@ -1,6 +1,7 @@
 package org.ole.planet.myplanet.repository
 
 import android.content.Context
+import android.util.Log
 import androidx.core.content.edit
 import androidx.core.net.toUri
 import com.google.gson.Gson
@@ -53,6 +54,10 @@ class ConfigurationsRepositoryImpl @Inject constructor(
     @PlainGson private val gson: Gson
 ) : ConfigurationsRepository {
     private val serverAvailabilityCache = ConcurrentHashMap<String, Pair<Boolean, Long>>()
+
+    companion object {
+        private const val TAG = "ConfigurationsRepository"
+    }
 
     override suspend fun checkHealth(): String {
         return try {
@@ -233,6 +238,10 @@ class ConfigurationsRepositoryImpl @Inject constructor(
                     if (f.exists()) {
                         val sha256 = withContext(dispatcherProvider.io) {
                             Sha256Utils().getCheckSumFromFile(f)
+                        }
+                        if (sha256 == null) {
+                            Log.w(TAG, "Could not compute checksum for $path")
+                            return false
                         }
                         return checksum.contains(sha256)
                     }
--- a/app/src/main/java/org/ole/planet/myplanet/utils/Sha256Utils.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/utils/Sha256Utils.kt
@@ -9,11 +9,6 @@ class Sha256Utils {
         private val HEX_CHARS = "0123456789abcdef".toCharArray()
     }
 
-    /**
-     * Returns the SHA-512 hex digest of [file], or `null` when the file cannot be read
-     * or the digest cannot be computed. The class is named Sha256Utils for historical
-     * reasons but computes a SHA-512 hash.
-     */
     fun getCheckSumFromFile(file: File): String? {
         return try {
             val digest = MessageDigest.getInstance("SHA-512")
--- a/app/src/main/java/org/ole/planet/myplanet/utils/Sha256Utils.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/utils/Sha256Utils.kt
@@ -9,7 +9,12 @@ class Sha256Utils {
         private val HEX_CHARS = "0123456789abcdef".toCharArray()
     }
 
-    fun getCheckSumFromFile(file: File): String {
+    /**
+     * Returns the SHA-512 hex digest of [file], or `null` when the file cannot be read
+     * or the digest cannot be computed. The class is named Sha256Utils for historical
+     * reasons but computes a SHA-512 hash.
+     */
+    fun getCheckSumFromFile(file: File): String? {
         return try {
             val digest = MessageDigest.getInstance("SHA-512")
             FileInputStream(file).use { fis ->
@@ -27,8 +32,8 @@ class Sha256Utils {
                 result[i * 2 + 1] = HEX_CHARS[v and 0x0F]
             }
             String(result)
-        } catch (e: Exception) {
-            ""
+        } catch (_: Exception) {
+            null
         }
     }
 }
--- a/app/src/main/java/org/ole/planet/myplanet/utils/Sha256Utils.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/utils/Sha256Utils.kt
@@ -1,6 +1,5 @@
 package org.ole.planet.myplanet.utils
 
-import android.util.Log
 import java.io.File
 import java.io.FileInputStream
 import java.security.MessageDigest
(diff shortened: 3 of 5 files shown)


dogi:
**Review — rating: 68/100**

The goal is right: `Sha256Utils` is otherwise pure JVM, and dropping `android.util.Log` lets `Sha256UtilsTest` shed its `mockkStatic(Log::class)` scaffolding entirely. That's a genuine simplification, and the test diff is the proof of it.

**But the deleted line was the only signal on a path that silently reports success.** Following the one caller — `ConfigurationsRepositoryImpl.checkCheckSum` (line 226) — the failure return flows into:

```kotlin
val sha256 = Sha256Utils().getCheckSumFromFile(f)   // "" when the digest throws
return checksum.contains(sha256)                     // contains("") is ALWAYS true
```

`String.contains("")` returns `true` (empty-string `indexOf` is 0), so a checksum that *could not be computed* makes `checkCheckSum` return `true` — the file is reported as passing integrity verification. Every other failure path in that function returns `false`; this is the one that fails open.

To be clear about ownership: that bug is pre-existing, not introduced here. What this PR changes is that it was previously *visible* — a `Log.w` with the file path and the exception — and now it is completely silent. On a file-integrity check, deleting the only evidence that verification fell over is the wrong trade even though the platform-decoupling motive is sound.

Both goals are reachable together. The cleanest version also fixes the fail-open:

```kotlin
fun getCheckSumFromFile(file: File): String? = try {
    // …
    String(result)
} catch (e: Exception) {
    null
}
```

and at the call site treat `null` as "cannot verify" → `false`, logging there where an Android dependency is already fine. That keeps `Sha256Utils` platform-free, keeps the diagnostic, and closes the `contains("")` hole in the same edit. If you'd rather not change the signature, routing the warning through the `DiagnosticsRepository` this repo already has — and is extending in #16362 and #16388 — preserves observability without the `android.util.Log` import.

**Verdict:** needs another round — keep the platform decoupling, but don't leave a silent fail-open on the integrity path. Returning `null` is a two-line change that makes this strictly better than where it started.

---

dogi:

---

dogi:
Pushed **2e6c262bff5fb21acb2a53f5a7e8dd491b0af307** addressing the review: keeping the platform decoupling while closing the fail-open on the integrity path.

**What changed**

- `Sha256Utils.getCheckSumFromFile` now returns `String?` (`null` on failure) instead of an empty string. The helper stays platform-free — no `android.util.Log` import.
- `ConfigurationsRepositoryImpl.checkCheckSum` now treats `null` as "cannot verify" → `false` (fail-closed), and re-surfaces the lost diagnostic via `Log.w` at the call site, where an Android dependency is already present. This closes the `checksum.contains("")` hole that previously reported an unverifiable file as passing integrity.
- `Sha256UtilsTest` updated to assert `null` on the failure paths.
- Added a `ConfigurationsRepositoryImplTest` case (`checkCheckSum returns false when checksum cannot be computed`) covering the new fail-closed behavior.

Net effect: the diagnostic that the original `Log.w` provided is preserved (moved to the caller), `Sha256Utils` remains pure JVM, and the integrity check no longer fails open.

_Note: as before, the test suite could not be executed locally (no JDK/Gradle toolchain in this environment); CI will confirm._

dogi:
## Review: 88/100 — mergeable, and the title badly undersells it

This is filed as "smoother sha256 utils platform freeing". It is actually a **checksum-verification bypass fix on the APK self-install path**, and that should be the headline.

### The bug on `master`

```kotlin
fun getCheckSumFromFile(file: File): String {
    return try { … } catch (e: Exception) { Log.w(…); "" }   // ← empty string on any failure
}
```

and its sole caller:

```kotlin
val sha256 = Sha256Utils().getCheckSumFromFile(f)
return checksum.contains(sha256)      // checksum.contains("") == true, always
```

`String.contains("")` is **unconditionally true**. So whenever the digest can't be computed — unreadable file, truncated file, `SecurityException`, anything — `checkCheckSum` reported **"the file matches the expected checksum."**

That verdict is not cosmetic. `DialogUtils.startDownloadUpdate` (:178) is the app-update flow:

```kotlin
val checksumMatch = configurationsRepository.checkCheckSum(path)
if (checksumMatch) {
    Utilities.toast(context, context.getString(R.string.apk_already_exists))
    FileUtils.installApk(context, path)     // ← hands the APK to the package installer
} else {
    DownloadService.startService(context, "$urls", false)   // re-download
}
```

So on `master` a cached APK whose integrity check *failed to run* was passed straight to `installApk` with an "already verified" toast, on the flavor that holds `REQUEST_INSTALL_PACKAGES`. Returning `null` and mapping it to `false` turns that into a re-download — fail-closed instead of fail-open. That's the right direction and it's worth saying plainly in the PR description and commit message, because a reviewer skimming "platform freeing" would never look for it.

**Blast radius of the signature change is nil:** `getCheckSumFromFile` has exactly one caller in `main` (`ConfigurationsRepositoryImpl:235`), so `String` → `String?` can't silently degrade anything else.

### The new test is a real regression test

Stubbing `getCheckSumFromFile` to `null` and asserting `checkCheckSum` is `false` fails on `master`, where the same scenario yields `""` → `contains` → `true`. Exactly the case that matters.

The `Sha256UtilsTest` cleanup is a genuine simplification — dropping `mockkStatic(Log::class)` plus the `@Before`/`@After` pair leaves a pure JUnit test with no Android framework dependency. That's the "platform freeing" in the title, and it's a nice side effect of moving the logging out.

### The one thing to fix: the exception is now discarded ⚠️

```kotlin
} catch (_: Exception) {
    null
}
```

The binding is dropped, and the caller's replacement log carries no throwable:

```kotlin
Log.w(TAG, "Could not compute checksum for $path")
```

On `master` you at least got `Log.w("Sha256Utils", "Failed to get checksum for file: …", e)` with the cause. For the code path that decides whether to install an APK, "could not compute checksum" without knowing whether it was `FileNotFoundException`, `SecurityException` or `NoSuchAlgorithmException` is a real loss — and it runs against the direction #16595 just set for this repo. Keep the cause:

```kotlin
} catch (e: Exception) {
    Log.w("Sha256Utils", "Failed to compute checksum for ${file.absolutePath}", e)
    null
}
```

That keeps the test's Log-free purity intact if you'd rather log at the call site instead — just pass the throwable through either way.

### Smaller notes

- **`contains` is still the comparison.** `checksum.contains(sha256)` is a substring match against the server's response body rather than an equality check on a parsed digest. With a real 128-char digest that's unlikely to false-positive, but it's a loose test for something gating an install. Not introduced here; worth an issue.
- The new test does its `unmockkStatic` / `unmockkObject` / `unmockkConstructor` at the **end of the test body**, so a failing assertion skips all three and leaks the mocks into sibling tests. The adjacent test has the same shape, so it's pre-existing — but `@After` would make the whole file more robust.
- `assertNotEquals("", checksum)` is now redundant beside `assertNotNull` and the 128-length check.
- Fully-qualified `io.mockk.*`, `android.util.Log::class` and `java.io.File` inline instead of imports.

**Verdict:** mergeable. Please restore the exception in the log, and retitle around the security fix rather than the test cleanup — this deserves to be findable in the history.

---

Okuro3499 (approved):

PR #16347, opened 2026-08-27
all: less notifications repository task team name is more (fixes #16338)

dogi:
## Summary

`NotificationsRepository.getTaskTeamName(taskTitle)` and its `NotificationsRepositoryImpl` override had no callers in `app/src/main` or `app/src/test` — `NotificationsViewModel` moved to the batched `getTaskTeamNamesByTaskTitles` / `getTaskTeamNamesByTaskIds`. The method kept a one-row-per-title query pattern alive in an otherwise fully batched interface.

This deletes the unused method and its implementation.

## Changes

- `repository/NotificationsRepository.kt`: removed `suspend fun getTaskTeamName(taskTitle: String): String?` from the interface.
- `repository/NotificationsRepositoryImpl.kt`: removed the matching `override`.

## Verification

- `grep -rn "getTaskTeamName\b" app/src/main app/src/test` now finds zero matches (previously only the declaration and the override).
- `./gradlew testDefaultDebugUnitTest --tests "...NotificationsRepositoryImplTest" --rerun-tasks` → BUILD SUCCESSFUL, 24 tests, 0 failures, 0 errors.
- `./gradlew testDefaultDebugUnitTest --tests "...NotificationsViewModelTest"` → BUILD SUCCESSFUL, 19 tests, 0 failures, 0 errors.

Fixes #16338

---

_This PR was created by an AI agent (OpenHands) on behalf of the user._

@dogi can click here to [continue refining the PR](https://app.all-hands.dev/conversations/bb46d836-d2d4-497a-914b-c9481bc88693)

PR #16360, opened 2026-08-27
all: less notifications repository exam dao is more (fixes #16318)

### LINKED ISSUE #16318: Drop the dead ExamDao injection from NotificationsRepositoryImpl
`NotificationsRepositoryImpl` declares `private val examDao: ExamDao` (:32) and imports it (:10) but never references it — a dead cross-feature DAO edge from notifications into the exam domain.

**Ask:** delete the unused import and constructor parameter, and update the three test-file references (constructor call sites/mocks) that currently pass an `ExamDao` to this repository.

**Files:** `repository/NotificationsRepositoryImpl.kt`, `test/.../NotificationsRepositoryImplTest.kt`

**Verification:** confirmed the only two occurrences in the production file are the import and the constructor parameter — no real usage.

dogi:
Removed unused ExamDao dependency injection from NotificationsRepositoryImpl and updated NotificationsRepositoryImplTest to reflect the simplified constructor.

Fixes #16318

---
*PR created automatically by Jules for task [8476041357477425582](https://jules.google.com/task/8476041357477425582) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6779
-        versionName = "0.67.79"
+        versionCode = 6780
+        versionName = "0.67.80"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/repository/NotificationsRepositoryImpl.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/NotificationsRepositoryImpl.kt
@@ -7,7 +7,6 @@ import java.util.Calendar
 import java.util.Date
 import java.util.UUID
 import javax.inject.Inject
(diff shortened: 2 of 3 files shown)


Okuro3499 (approved):

dogi:
**Review — rating: 89/100**

Drops an unused `ExamDao` from `NotificationsRepositoryImpl`. Its only occurrence in the file was the constructor parameter itself (line 32) — never referenced in a single method body — and since removing it still compiles across both flavors and both test shards, that's proof rather than inference.

Two details that make this lower-risk than the average constructor edit: `examDao` was the **last** positional parameter, so nothing after it shifts and no other argument can silently re-bind to the wrong mock; and the test is updated in the same diff, so the Hilt graph and the hand-constructed test instance stay in agreement.

Worth noting for sequencing: #16347 also edits `NotificationsRepositoryImpl.kt`, but in the method bodies (~line 249) rather than the imports and constructor, so the two land in disjoint hunks and shouldn't conflict.

**Verdict:** mergeable.

---

PR #16361, opened 2026-08-27
all: smoother notifications repository dao querying (fixes #16325)

### LINKED ISSUE #16325: Replace full-row notification loads with id-only projections
`markNotificationsAsRead` (:116) does `notificationDao.getByIds(...).map { it.id }`, `deleteNotifications` (:406) does the same, and `markAllUnreadAsRead` (:124) loads every unread row via `getNotifications(userId,"unread",false)` just to build an id set — full `AppNotification` rows (message text, dates, sync flags) materialized to learn which ids exist.

**Ask:** replace these three full-row loads with id-only projections (a new DAO query returning just ids where needed).

**Files:** `data/room/dao/NotificationDao.kt`, `repository/NotificationsRepositoryImpl.kt`, `test/.../NotificationsRepositoryImplTest.kt`

**Verification:** all three sites confirmed at the cited lines.

dogi:
Replaced full-row notification loads with id-only projections in NotificationDao and NotificationsRepositoryImpl for markNotificationsAsRead, markAllUnreadAsRead, and deleteNotifications. Updated unit tests in NotificationsRepositoryImplTest to verify id-only projection queries.

Fixes #16325

---
*PR created automatically by Jules for task [4091770736283420069](https://jules.google.com/task/4091770736283420069) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6780
-        versionName = "0.67.80"
+        versionCode = 6781
+        versionName = "0.67.81"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/data/room/dao/NotificationDao.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/data/room/dao/NotificationDao.kt
@@ -36,6 +36,12 @@ interface NotificationDao {
     @Query("SELECT * FROM notifications WHERE id IN (:ids)")
     suspend fun getByIds(ids: List<String>): List<AppNotification>
 
+    @Query("SELECT id FROM notifications WHERE id IN (:ids)")
+    suspend fun getIdsByIds(ids: List<String>): List<String>
+
+    @Query("SELECT id FROM notifications WHERE userId = :userId AND isRead = 0")
+    suspend fun getUnreadIds(userId: String): List<String>
+
     @Query("UPDATE notifications SET isRead = 1, createdAt = :createdAt, needsSync = CASE WHEN isFromServer = 1 THEN 1 ELSE needsSync END WHERE id IN (:ids)")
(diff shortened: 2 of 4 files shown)


dogi:
**Review — rating: 88/100**

Three call sites stop loading full rows just to pull `.id` off them. Two are exact projections — `getIdsByIds` is `getByIds(...).map { it.id }` done in SQL, same set, same order semantics. Room validates `@Query` at compile time, so green KSP also confirms the column and the `List<String>` scalar mapping.

**The third one does more than the title claims, and in the right direction.** `markAllUnreadAsRead` previously read its ID set from `getNotifications(userId, "unread", false)`, whose predicate is:

```sql
WHERE (userId = :userId OR (:isAdmin = 1 AND userId = 'SYSTEM'))
  AND message != 'INVALID' AND message != ''
  AND (… :filter = 'unread' AND isRead = 0)
```

but the write it then performs is `markAllUnreadAsRead(userId, …)`:

```sql
UPDATE notifications SET isRead = 1, … WHERE userId = :userId AND isRead = 0
```

Those didn't agree. The UPDATE marks *every* unread row for the user, including ones with `message = 'INVALID'` or empty — while the returned set excluded exactly those. So the function was reporting a narrower change than it actually made. The new `getUnreadIds` (`WHERE userId = :userId AND isRead = 0`) matches the UPDATE's predicate character for character, so the returned set now describes precisely what was modified.

I'd flag that in the PR description: it's a behaviour change, not just a projection. The returned set is now a superset that can include IDs for rows the notification list never displays. That's the correct direction — a read-back should mirror its write — and harmless for a caller that only uses the IDs to reconcile state, but it deserves to be on the record rather than discovered later.

The performance argument is strongest on this same call site: it was materialising every unread notification, message bodies included, to keep only the primary key.

Tests cover both empty-set short circuits and both happy paths, asserting the DAO interactions rather than just the return value. Appropriate for delegation logic; the new SQL itself leans on Room's compile-time validation rather than a Room instrumentation test, which is a fair trade for queries this simple.

**Verdict:** mergeable.

---

Okuro3499 (approved):

dogi:

---

PR #16362, opened 2026-08-27
all: smoother diagnostics repository log building (fixes #16342)

### LINKED ISSUE #16342: Reuse buildApkLog in the batch diagnostics path
`DiagnosticsRepositoryImpl` builds the same `ApkLog` fields twice — once in `buildApkLog` used by `saveLogToRoom`, and again as an inline `ApkLog().apply {}` inside `saveLogsToRoom` with its own precomputed `versionName`/`parentCode`/`planetCode`. The two have already drifted (`model?.let { userId = it.id }` vs `modelId?.let { userId = it }`).

**Ask:** have `saveLogsToRoom` reuse `buildApkLog` instead of duplicating the field-construction logic, resolving the drift between the two (decide which behavior is correct and keep just that one).

**Files:** `repository/DiagnosticsRepositoryImpl.kt`

**Verification:** confirmed the duplication at :40-59 and :72-94.

dogi:
## Summary

`DiagnosticsRepositoryImpl` built the same `ApkLog` fields twice: once in `buildApkLog` (used by `saveLogToRoom`) and again as an inline `ApkLog().apply {}` inside `saveLogsToRoom`, with precomputed `versionName`/`parentCode`/`planetCode`. The two had already drifted — `buildApkLog` used `modelId?.let { userId = it }` while the inline path used `model?.let { userId = it.id }`.

This makes `saveLogsToRoom` reuse `buildApkLog`, so both save paths share a single field-construction path and the drift is resolved. The resolved behavior keeps `userId = model?.id` (the value both expressions reduce to), now expressed through `buildApkLog`'s `modelId` parameter.

## Files changed

- `repository/DiagnosticsRepositoryImpl.kt` — `saveLogsToRoom` now maps each `PendingLog` through `buildApkLog` instead of constructing `ApkLog` inline.
- `app/src/test/.../repository/DiagnosticsRepositoryImplTest.kt` — new test covering both `saveLogToRoom` and `saveLogsToRoom`, asserting the shared field-construction (type, error, time, page, parentCode, createdOn, version, userId, generated id) plus the no-user and exception edge cases.

## Verification

```
./gradlew testDefaultDebugUnitTest --tests "org.ole.planet.myplanet.repository.DiagnosticsRepositoryImplTest"
```

Result: `tests="7" skipped="0" failures="0" errors="0"` — all passing.

Closes #16342.

---

@dogi can click here to [continue refining the PR](https://app.all-hands.dev/conversations/3e029501-b764-46f5-8d40-28529d3710d1)

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6781
-        versionName = "0.67.81"
+        versionCode = 6782
+        versionName = "0.67.82"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/repository/DiagnosticsRepositoryImpl.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/DiagnosticsRepositoryImpl.kt
@@ -27,7 +27,9 @@ class DiagnosticsRepositoryImpl @Inject constructor(
     }
 
     private fun buildApkLog(
-        spm: SharedPrefManager,
+        parentCode: String,
+        planetCode: String,
+        versionName: String?,
         modelId: String?,
         time: String,
         type: String,
@@ -35,12 +37,12 @@ class DiagnosticsRepositoryImpl @Inject constructor(
     ): ApkLog {
         return ApkLog().apply {
             id = "${UUID.randomUUID()}"
-            parentCode = spm.getParentCode()
-            createdOn = spm.getPlanetCode()
+            this.parentCode = parentCode
+            createdOn = planetCode
             modelId?.let { userId = it }
             this.time = time
             page = ""
-            version = VersionUtils.getVersionName(context)
+            version = versionName
             this.type = type
             if (error.isNotEmpty()) {
                 this.error = error
@@ -51,7 +53,15 @@ class DiagnosticsRepositoryImpl @Inject constructor(
     override suspend fun saveLogToRoom(type: String, error: String, time: String): Boolean {
         return try {
             val model = userSessionManager.getUserModel()
-            val log = buildApkLog(sharedPrefManager, model?.id, time, type, error)
+            val log = buildApkLog(
+                sharedPrefManager.getParentCode(),
+                sharedPrefManager.getPlanetCode(),
+                VersionUtils.getVersionName(context),
+                model?.id,
+                time,
+                type,
+                error,
+            )
             apkLogDao.insert(log)
             true
         } catch (e: Exception) {
@@ -64,8 +74,11 @@ class DiagnosticsRepositoryImpl @Inject constructor(
         if (pendingLogs.isEmpty()) return true
         return try {
             val model = userSessionManager.getUserModel()
+            val parentCode = sharedPrefManager.getParentCode()
+            val planetCode = sharedPrefManager.getPlanetCode()
+            val versionName = VersionUtils.getVersionName(context)
             val logsToInsert = pendingLogs.map { pending ->
-                buildApkLog(sharedPrefManager, model?.id, pending.time, pending.type, pending.error)
+                buildApkLog(parentCode, planetCode, versionName, model?.id, pending.time, pending.type, pending.error)
             }
             apkLogDao.insertAll(logsToInsert)
             true
--- a/app/src/main/java/org/ole/planet/myplanet/repository/DiagnosticsRepositoryImpl.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/DiagnosticsRepositoryImpl.kt
@@ -64,24 +64,8 @@ class DiagnosticsRepositoryImpl @Inject constructor(
         if (pendingLogs.isEmpty()) return true
         return try {
             val model = userSessionManager.getUserModel()
-            val versionName = VersionUtils.getVersionName(context)
-            val parentCode = sharedPrefManager.getParentCode()
-            val planetCode = sharedPrefManager.getPlanetCode()
-
             val logsToInsert = pendingLogs.map { pending ->
-                ApkLog().apply {
-                    id = "${UUID.randomUUID()}"
-                    this.parentCode = parentCode
-                    this.createdOn = planetCode
-                    model?.let { userId = it.id }
-                    this.time = pending.time
-                    page = ""
-                    version = versionName
-                    this.type = pending.type
-                    if (pending.error.isNotEmpty()) {
-                        this.error = pending.error
-                    }
-                }
+                buildApkLog(sharedPrefManager, model?.id, pending.time, pending.type, pending.error)
             }
             apkLogDao.insertAll(logsToInsert)
             true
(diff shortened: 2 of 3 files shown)


dogi:
**Review — rating: 72/100**

I checked `buildApkLog` field by field against the block it replaces, and the dedup is faithful — `id` (fresh UUID per call), `parentCode`, `createdOn`, the `modelId?.let { userId = it }` null guard, `time`, `page = ""`, `version`, `type`, and the `if (error.isNotEmpty())` guard all match exactly. Collapsing two copies of that constructor into one is worth doing.

**But it makes the batch path slower, which is the path the title says it smooths.** The old code hoisted three lookups above the `map`:

```kotlin
val versionName = VersionUtils.getVersionName(context)   // once
val parentCode  = sharedPrefManager.getParentCode()      // once
val planetCode  = sharedPrefManager.getPlanetCode()      // once
val logsToInsert = pendingLogs.map { … }                 // N iterations, reusing all three
```

`buildApkLog` does all three itself (lines 38, 39, 43), and it's now called once per element — so N pending logs means N× `VersionUtils.getVersionName(context)`, which is a `PackageManager.getPackageInfo` call, i.e. IPC to the system server, plus N× each preference read. The hoisting in the original wasn't incidental; this refactor quietly undoes it. Flushing pending crash/ANR logs typically happens at startup, which is the worst moment to add per-item IPC.

**Fix** — keep the single construction site and the hoisting, by passing the resolved values in:

```kotlin
private fun buildApkLog(
    parentCode: String, planetCode: String, versionName: String,
    modelId: String?, time: String, type: String, error: String
): ApkLog

// batch: resolve the three once, then map
// single: resolve inline at the one call site
```

**The tests are otherwise strong** — both paths, null user, exception paths, empty list, every field asserted, and `assertFalse(first.id == second.id)` specifically pinning that each log gets its own UUID, which is the exact thing a careless hoist would break. Worth noting they *can't* catch the issue above: `VersionUtils` is mocked with `mockkObject`, so N calls look the same as one. Adding `verify(exactly = 1) { VersionUtils.getVersionName(any()) }` to the batch test would pin the hoisting the same way the UUID assertion pins per-log identity.

**Verdict:** needs another round — the dedup is right, it just shouldn't cost the batch path its hoisted lookups.

---

dogi:

---

dogi:
Thanks @dogi — addressed in d5f3384.

The dedup stayed, but `buildApkLog` no longer resolves the three context lookups itself. Its signature now takes the already-resolved `parentCode`, `planetCode`, and `versionName`:

```kotlin
private fun buildApkLog(
    parentCode: String, planetCode: String, versionName: String?,
    modelId: String?, time: String, type: String, error: String
): ApkLog
```

- **Batch path (`saveLogsToRoom`)** hoists the three lookups above the `map` again — `VersionUtils.getVersionName(context)` (the PackageManager IPC), `getParentCode()`, and `getPlanetCode()` each run once for the whole batch, then are passed into `buildApkLog` per element. The startup-time flush keeps its hoisting.
- **Single path (`saveLogToRoom`)** resolves the same three inline at its one call site and passes them in.

The drift the original PR fixed is preserved (still `userId = model?.id` via the `modelId` param), and both save paths still share one field-construction site.

The batch test now pins the hoisting the way the UUID assertion pins per-log identity:

```kotlin
verify(exactly = 1) { VersionUtils.getVersionName(any()) }
verify(exactly = 1) { sharedPrefManager.getParentCode() }
verify(exactly = 1) { sharedPrefManager.getPlanetCode() }
```

As you noted, the existing mocks couldn't catch the per-item regression on their own; these would have failed under the old `buildApkLog`-per-call version (it called each lookup N times).

Verification:
```
./gradlew testDefaultDebugUnitTest --tests "org.ole.planet.myplanet.repository.DiagnosticsRepositoryImplTest"
```
`tests="7" skipped="0" failures="0" errors="0"` — all passing with the new hoisting assertions.

I didn't find a separate set of improvement instructions from @Okuro3499 on the PR or the linked issue #16342 (only the coderabbit rate-limit notice and your two comments), so I worked from your review. If there are further instructions I missed, point me at them.

_This comment was created by an AI agent (OpenHands) on behalf of the maintainer._

dogi:
## Review: 86/100 — mergeable

Good dedup, and the accompanying test file is the more valuable half: `DiagnosticsRepositoryImpl` had **zero** coverage before this.

**The two constructions really were identical.** I diffed the deleted inline block against `buildApkLog` field by field — id, parentCode, createdOn, time, `page = ""`, version, type, and the `if (error.isNotEmpty())` guard all line up. The one place they could have diverged is the user id:

- old: `model?.let { userId = it.id }`
- new: `modelId = model?.id`, then `modelId?.let { userId = it }`

Equivalent, because `UserEntity.id` is a non-null `String` — so `model?.id` is null exactly when `model` is null. Worth checking rather than assuming, since a nullable `id` would have changed behavior for a logged-in user with a null id.

The real win is that the two paths can no longer drift: adding a field to `ApkLog` used to mean remembering to touch both the single and batch constructors, and forgetting the batch one would have silently dropped the field from every crash log flushed at startup.

**Changing the parameter from `spm` to explicit values is what makes the sharing possible** — `buildApkLog` can't take a `SharedPrefManager` and still be callable from a hoisted batch loop.

### One correction to the test's comment

```kotlin
// The three context lookups are hoisted above the map, so each runs once
// for the whole batch regardless of size (VersionUtils.getVersionName is
// a PackageManager IPC — the batch path flushes pending logs at startup).
verify(exactly = 1) { VersionUtils.getVersionName(any()) }
```

The hoisting isn't new — `master` already had `versionName`/`parentCode`/`planetCode` computed above the `map`; this diff only reorders those three lines. So those three `verify(exactly = 1)` calls pass on `master` too. They're fine as guards (they'd catch a future refactor that pushed the lookups back into the loop), but the comment reads as if the PR introduced the optimization. Worth rewording so the next reader doesn't go looking for a perf win that isn't in this diff.

More generally: **every** test here passes on `master`, and for a pure refactor that's exactly right — they're characterization tests pinning the behavior that must not change. Good practice, not a gap.

### Nit worth acting on: five same-typed positional parameters

```kotlin
private fun buildApkLog(
    parentCode: String, planetCode: String, versionName: String?,
    modelId: String?, time: String, type: String, error: String,
)
```

Seven parameters, five of them `String`/`String?` in a row, called positionally at both sites. Transposing `parentCode` and `planetCode` compiles cleanly and silently mislabels every diagnostic log. Named arguments at the two call sites would cost one line each and remove the whole class of mistake. (Credit where due: the new tests *would* catch it — they assert `parentCode` and `createdOn` to distinct values — which is a good argument for the tests, not against the nit.)

### Smaller test notes

- `assertFalse(first.id == second.id)` reads better as `assertNotEquals(first.id, second.id)`.
- `unmockkObject` is imported but unused (`unmockkAll()` in `@After` covers it).
- `context = mockk()` without `relaxed` only works because `VersionUtils` is mocked so nothing dereferences it — fine, but it'll fail confusingly if a future field needs the real context.

The error-path tests (`returns false on exception`, plus `coVerify(exactly = 0)` on the insert) are the ones I'd most want here, and they're present for both methods.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16365, opened 2026-08-27
all: smoother version utils android id caching (fixes #16317)

### LINKED ISSUE #16317: Resolve the device android id once instead of on every serialization
`VersionUtils.getAndroidId` queries `Settings.Secure` through the content resolver on every call and `NetworkUtils.getUniqueIdentifier` repeats the same query independently. Both are called per serialized document (`SearchActivity`, `UserEntity`, `MyPlanet` x2, `ActivitiesRepositoryImpl`), so a 50-item upload batch issues 50+ content-resolver round trips for a process-lifetime constant. In the same file, `getCustomDeviceName` re-resolves `coreEntryPoint.sharedPrefManager()` (:189) although the object already memoizes it at :34.

**Ask:** cache the Android ID (a process-lifetime constant) instead of re-querying the content resolver per call, and reuse the already-memoized `sharedPrefManager()` in `getCustomDeviceName` instead of re-resolving it. Keep the existing null-propagation behavior when the ID is unavailable — don't cache a null incorrectly across calls that could later succeed.

**Files:** `utils/VersionUtils.kt`, `utils/NetworkUtils.kt`, `test/.../NetworkUtilsTest.kt`

**Verification:** duplicate `sharedPrefManager()` resolution confirmed at `NetworkUtils.kt:34` vs `:189`.

dogi:
Cache process-lifetime Android ID in VersionUtils to prevent content resolver round trips per serialized document, delegate getUniqueIdentifier to VersionUtils.getAndroidId, reuse memoized sharedPrefManager in NetworkUtils.getCustomDeviceName, and add comprehensive unit tests.

Fixes #16317

---
*PR created automatically by Jules for task [12205646375015651849](https://jules.google.com/task/12205646375015651849) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6783
-        versionName = "0.67.83"
+        versionCode = 6784
+        versionName = "0.67.84"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/utils/NetworkUtils.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/utils/NetworkUtils.kt
@@ -170,7 +170,7 @@ object NetworkUtils {
     }
 
     fun getUniqueIdentifier(): String {
-        val androidId = Settings.Secure.getString(context.contentResolver, Settings.Secure.ANDROID_ID)
+        val androidId = VersionUtils.getAndroidId(context)
         val buildId = Build.ID
         return androidId + "_" + buildId
     }
@@ -186,8 +186,7 @@ object NetworkUtils {
     }
 
     fun getCustomDeviceName(context: Context): String {
-        val spm = coreEntryPoint.sharedPrefManager()
-        return spm.getCustomDeviceName()
+        return sharedPrefManager.getCustomDeviceName()
     }
 
     fun extractProtocol(url: String): String? {
--- a/app/src/main/java/org/ole/planet/myplanet/utils/VersionUtils.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/utils/VersionUtils.kt
@@ -4,9 +4,13 @@ import android.content.Context
 import android.content.pm.PackageManager
 import android.os.Build
 import android.provider.Settings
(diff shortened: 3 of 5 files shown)


dogi:
**Review — rating: 84/100**

Two independent wins, both real:

**The memoized `SharedPrefManager` matters more than the diff suggests.** `NetworkUtils` already had `private val sharedPrefManager: SharedPrefManager by lazy { coreEntryPoint.sharedPrefManager() }` (line 33), and `getCustomDeviceName` was bypassing it to do a fresh entry-point lookup on every call. That function has ~18 call sites, most of them in serializers — `MyLibrary`, `Rating`, `NewsLog`, `UserEntity`, `MyPlanet`, plus `UploadManager` and `UploadConfigs` — so it runs once per uploaded document. Routing it through the existing lazy is a straightforward hot-path fix.

**The ANDROID_ID cache is correctly built.** `@Volatile`, and deliberately *not* caching null so a transient failure can be retried — which the second test pins by seeding null, then a real value, and asserting the real one comes through. The first test proves caching by mutating `Settings.Secure` between calls and asserting the stale value is returned. Both are genuine regression tests, not smoke tests. And `resetAndroidIdCacheForTesting()` is added to both `NetworkUtilsTest` and `VersionUtilsTest` setup — necessary hygiene for object-level state given CI shards by class.

Two nits, neither blocking:

1. **`getCustomDeviceName(context: Context)` no longer uses `context`.** Kotlin doesn't warn on unused parameters, so the misleading signature will just sit there implying the context matters. Dropping it is mechanical across the 18 call sites; if you'd rather not churn them, a one-line comment saying it's retained for call-site compatibility would stop the next reader wondering.

2. **The `String` → `String?` change renders a literal `"null"` at two sites rather than failing.** Most callers pass the result to `JsonObject.addProperty`, which handles null fine. But `MyPlanet.kt:32` builds a document `_id` as `getAndroidId(context) + "@" + …`, and `ActivitiesRepositoryImpl.kt:418` interpolates it into a URL path — both would yield `null@…` instead of the old non-null contract's failure. ANDROID_ID is effectively never null on a real device, so this is theoretical, but a bad `_id` persists on the server where a crash wouldn't. If you want the nullable contract (and the tests say you do), an explicit `?: ""` or a guard at those two sites closes it.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16367, opened 2026-08-27
enterprises: smoother repository context injecting (fixes #16343)

### LINKED ISSUE #16343: Inject Context into EnterprisesRepositoryImpl instead of reaching for MainApplication.context
`EnterprisesRepositoryImpl` imports `MainApplication` (:11) and uses `MainApplication.context` inside `attachTeamImage` (:128) although the repository is otherwise context-free — a data-layer-to-app-singleton leak.

**Ask:** inject `@ApplicationContext Context` through the constructor (matching the pattern other repositories use) instead of reaching for the `MainApplication` singleton.

**Files:** `repository/EnterprisesRepositoryImpl.kt`, `test/.../EnterprisesRepositoryImplTest.kt`

**Verification:** confirmed both lines.

dogi:
## Summary

`EnterprisesRepositoryImpl` imported `MainApplication` and reached for the `MainApplication.context` singleton inside `attachTeamImage`, even though the repository is otherwise context-free — a data-layer-to-app-singleton leak.

This injects `@ApplicationContext Context` through the constructor (matching the pattern other repositories such as `ResourcesRepositoryImpl` and `TeamsRepositoryImpl` use) and uses the injected context instead of the `MainApplication` singleton.

## Changes

- `repository/EnterprisesRepositoryImpl.kt`
  - Added `@param:ApplicationContext private val context: Context` constructor parameter (Hilt auto-provides `@ApplicationContext`).
  - Removed the `MainApplication` import; `attachTeamImage` now passes the injected `context` to `MyTeam.getAttachmentFile(...)`.
- `test/.../EnterprisesRepositoryImplTest.kt`
  - Constructed the repository with a mock `Context`.
  - Added `addReport with image writes attachment using injected context`, which mocks `FileUtils.getOlePath(context)` to a temp dir and asserts the attachment bytes are written through the injected context (verifying the context flows end-to-end instead of reaching the singleton).

## Verification

- `./gradlew testDefaultDebugUnitTest --tests "org.ole.planet.myplanet.repository.EnterprisesRepositoryImplTest"` — both tests pass (`tests=2, failures=0, errors=0`).
- Confirmed no remaining `MainApplication` references in `EnterprisesRepositoryImpl.kt`.

Fixes #16343.

---

@dogi can click here to [continue refining the PR](https://app.all-hands.dev/conversations/d29e87b4-456f-4a04-a0eb-04640fc3f6da)


dogi:
**Review — rating: 88/100**

Replaces a reach for the global `MainApplication.context` with a proper `@ApplicationContext` injection. Same instance at runtime, so no behaviour change — the value is that `EnterprisesRepositoryImpl` no longer depends on application-wide static state, and the import is gone rather than merely unused.

**The test is the strongest part and does real work.** Mocking `FileUtils.getOlePath(context)` to redirect into a temp tree, then asserting `logo.png` actually exists with the exact bytes, proves the injected context threads all the way through `MyTeam.getAttachmentFile(context, …)` to the write. That's a behavioural test of the injection, not an assertion that the constructor accepts an argument — which is precisely what makes this change verifiable rather than merely plausible.

`@param:ApplicationContext` is explicit about the use-site target where Kotlin would have picked `param` anyway; being explicit is the better habit, especially with use-site defaults shifting in newer Kotlin. Adding `context` as the *first* constructor parameter shifts everything after it, but Hilt binds by type rather than position and the only hand-written construction site (the test) is updated in the same diff — a green build across both flavors confirms there's no other manual instantiation.

**Sequencing note:** #16363 also edits `EnterprisesRepositoryImpl.kt` and `EnterprisesRepositoryImplTest.kt`, and the two overlap. In the test both rewrite the same block — the `mockk()` → `mockk(relaxed = true)` fields, the `TestDispatcherProvider(UnconfinedTestDispatcher())` swap, and the `EnterprisesRepositoryImpl(...)` call — and both touch the import block in the main file. Whichever lands second will need a rebase. Nothing wrong with either PR; just don't queue them back to back and expect a clean automerge.

**Nit:** the temp directory from `Files.createTempDirectory` is never cleaned up. `TemporaryFolder` as a `@get:Rule` handles that automatically and is already used elsewhere in the suite.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16368, opened 2026-08-27
life: smoother repository view modelling (fixes #16328)

### LINKED ISSUE #16328: Move the my-life seed-if-empty sequence out of LifeViewModel into LifeRepository
`LifeViewModel.loadMyLifeList` orchestrates a data-layer sequence from the UI: read `getMyLifeByUserId`, and if empty call `seedMyLifeIfEmpty` then read the whole table again — three repository round trips per screen open, two of them full table scans, plus a race window the repository's own `seedMutex` cannot close from outside. `LifeRepositoryImpl` already owns this exact pattern internally for the dashboard (`getMyLifeForDashboard`).

**Ask:** move the seed-if-empty orchestration into `LifeRepository`/`LifeRepositoryImpl` so `LifeViewModel` makes one call, matching the dashboard path. Note the `isVisible`/cache differences from `getMyLifeForDashboard` — this screen's semantics aren't identical, so don't just delegate to that method. `ui/life/LifeViewModelTest.kt` may be owned by another in-flight PR (#16096) — check for conflicts before finalizing.

**Files:** `repository/LifeRepository.kt`, `repository/LifeRepositoryImpl.kt`, `ui/life/LifeViewModel.kt`, `test/.../LifeRepositoryImplTest.kt`

**Verification:** confirmed at `LifeViewModel.kt:32-45`.

dogi:
Refactored `LifeViewModel.loadMyLifeList` to delegate seed-if-empty data layer orchestration to `LifeRepositoryImpl.getMyLifeByUserId`, reducing multiple repository round trips and race windows to a single call.

Fixes #16328

---
*PR created automatically by Jules for task [311082300425698428](https://jules.google.com/task/311082300425698428) started by @dogi*


dogi:
**Review — rating: 85/100**

Good move, and it quietly fixes something the title doesn't mention: **`ensureLatest` was a no-op**. The old implementation took the parameter and never referenced it in the body, so every `ensureLatest = true` call site was documentation for behaviour that didn't exist. Replacing it with `defaultItems`, which the body actually uses, removes a parameter that was actively misleading.

The layering choice is the right one. `MyLife.defaultItems(userId, context::getString)` stays in the ViewModel and the repository receives an already-built list, so localization remains in the UI layer and `LifeRepositoryImpl` doesn't acquire a `Context` dependency it would otherwise need. Seed-if-empty is genuinely a data-lifecycle concern, so it belongs behind the repository rather than being sequenced by the caller.

Backward compatibility is handled precisely by the `|| defaultItems.isEmpty()` clause: every existing call site now passes nothing, hits that branch, and returns exactly what it returned before — including `getMyLifeForDashboard`, which keeps its own separate `seedMyLifeIfEmpty(effectiveUserId, seedBase)` path rather than double-seeding. The read → seed → re-read sequence matches the ViewModel code it replaces one for one.

Tests cover both new branches, and `returnsMany listOf(emptyList(), listOf(seededItem))` is the right way to model the re-read — it would fail if the second fetch were dropped.

**Two small things:**

1. **The defaults are now built eagerly.** The ViewModel previously called `MyLife.defaultItems(userId, context::getString)` only inside the `isEmpty()` branch — i.e. essentially first launch. Now it's constructed on every `loadMyLife`, running a `getString` per default item whether or not anything gets seeded. Making the parameter `defaultItems: () -> List<MyLife> = { emptyList() }` and invoking it only at the seeding branch keeps the call site just as clean and restores the laziness.
2. `LifeRepositoryTest.getMyLifeByUserId(userId)` asserts an empty result for the no-defaults path but doesn't assert that nothing was seeded. A `coVerify(exactly = 0) { myLifeDao.insertAll(any()) }` there would pin the backward-compatibility property that the whole change rests on.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16370, opened 2026-08-27
community: smoother home dialog flowing (fixes #16327)

### LINKED ISSUE #16327: Make HomeCommunityDialogFragment consume CommunityTabViewModel instead of injecting repositories
`HomeCommunityDialogFragment` field-injects `SharedPrefManager` and `ConfigurationsRepository` (:28-31) and rebuilds in `initCommunityTab` exactly the three values `CommunityTabViewModel` already publishes as `CommunityTabState` — while its sibling `CommunityTabFragment`, over the identical layout, consumes the ViewModel properly.

**Ask:** have `HomeCommunityDialogFragment` consume `CommunityTabViewModel`'s state instead of re-deriving it from injected repositories. **Careful:** this dialog's pager id is `communityName@parentCode`, not the tab fragment's `planetCode@parentCode` — don't silently switch which id gets used.

**Files:** `ui/community/HomeCommunityDialogFragment.kt`

**Verification:** confirmed the duplicated injection and rebuild.

dogi:
Refactored HomeCommunityDialogFragment to consume state published by CommunityTabViewModel instead of field injecting SharedPrefManager and ConfigurationsRepository to re-derive state. Maintained communityName@parentCode as the adapter ID as requested.

Fixes #16327

---
*PR created automatically by Jules for task [15641830446624082505](https://jules.google.com/task/15641830446624082505) started by @dogi*


Okuro3499 (approved):

dogi:
**Review — rating: 87/100**

Drops two direct dependencies (`SharedPrefManager`, `ConfigurationsRepository`) from the fragment in favour of the ViewModel that already computes this state. I checked `CommunityTabViewModel` composes the values from the identical sources — `sharedPrefManager.getParentCode()`, `getCommunityName()`, `configurationsRepository.getPlanetType()` (lines 34–36) — so the rendered community name, parent code and planet type are unchanged. This is real dedup rather than indirection: the fragment was re-deriving state the VM already produces.

The coroutine handling is correct where it easily could not have been. `filterNotNull().first()` is a one-shot await rather than an ongoing collection, so the absence of `repeatOnLifecycle` is right rather than an oversight, and `viewLifecycleOwner.lifecycleScope` cancels it on view destroy — meaning the `binding` accesses after the suspension point can't run against a torn-down view. Keeping `binding.appBar.setBackgroundColor(...)` outside the coroutine is also correct, since it needs no state.

Worth knowing rather than fixing: adapter setup moves from synchronous to post-suspension, so in principle the sheet could render with empty tabs first. In practice `viewModelScope` dispatches on `Main.immediate` and the VM's `init` does only in-memory prefs reads plus a non-suspending `getPlanetType()`, so the state is almost certainly set before the fragment subscribes. If `getPlanetType()` ever becomes genuinely async, that window becomes visible and would want a loading state.

**Verdict:** mergeable.

---

PR #16381, opened 2026-08-27
all: smoother notification list item html caching (fixes #16334)

### LINKED ISSUE #16334: Move HTML parsing out of the NotificationListItem model into the adapter
`NotificationListItem.Item.parsedText` is a lazy `Html.fromHtml(...)` inside `model/` (:18-19) — the sealed class's only `android.text` dependency. Its sole consumer is `NotificationsAdapter.ItemViewHolder.bind` (:113), so the parse belongs at bind time, not baked into the model.

**Ask:** move the `Html.fromHtml` parsing from `NotificationListItem` into `NotificationsAdapter`'s bind, keeping the model platform-free.

**Files:** `model/NotificationListItem.kt`, `ui/notifications/NotificationsAdapter.kt`

**Verification:** confirmed both files and that `parsedText` has exactly one consumer.

dogi:
Moved Html.fromHtml parsing from NotificationListItem model into NotificationsAdapter.ItemViewHolder.bind, leaving the model platform-free.

Fixes #16334

---
*PR created automatically by Jules for task [17284592860637804382](https://jules.google.com/task/17284592860637804382) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6786
-        versionName = "0.67.86"
+        versionCode = 6787
+        versionName = "0.67.87"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/model/NotificationListItem.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/model/NotificationListItem.kt
@@ -1,7 +1,5 @@
 package org.ole.planet.myplanet.model
 
-import android.text.Html
-
 sealed class NotificationListItem {
     data class Header(
         val type: String,
@@ -14,9 +12,5 @@ sealed class NotificationListItem {
         val notification: Notification,
         val isSelected: Boolean = false,
         val isSelectionMode: Boolean = false
-    ) : NotificationListItem() {
-        val parsedText: CharSequence by lazy(LazyThreadSafetyMode.NONE) {
-            Html.fromHtml(notification.formattedText.toString(), Html.FROM_HTML_MODE_LEGACY)
-        }
-    }
+    ) : NotificationListItem()
 }
--- a/app/src/main/java/org/ole/planet/myplanet/ui/notifications/NotificationsAdapter.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/notifications/NotificationsAdapter.kt
@@ -1,5 +1,6 @@
 package org.ole.planet.myplanet.ui.notifications
 
+import android.text.Html
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
@@ -110,7 +111,10 @@ class NotificationsAdapter(
 
         fun bind(item: NotificationListItem.Item) {
             val notification = item.notification
-            binding.title.text = item.parsedText
+            binding.title.text = Html.fromHtml(
+                notification.formattedText.toString(),
+                Html.FROM_HTML_MODE_LEGACY
+            )
             binding.timestamp.text = formatRelativeTime(notification.createdAt)
             binding.root.alpha = if (notification.isRead) 0.6f else 1.0f
 
--- a/app/src/main/java/org/ole/planet/myplanet/ui/notifications/NotificationsAdapter.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/notifications/NotificationsAdapter.kt
@@ -38,6 +38,7 @@ class NotificationsAdapter(
 
     private var dateFormatter: DateTimeFormatter? = null
     private var lastLocale: Locale? = null
+    private val parsedHtmlCache = LruCache<String, CharSequence>(100)
 
     private fun getDateFormatter(): DateTimeFormatter {
         val currentLocale = Locale.getDefault()
@@ -54,7 +55,6 @@ class NotificationsAdapter(
     companion object {
         private const val VIEW_TYPE_HEADER = 0
         private const val VIEW_TYPE_ITEM = 1
-        private val parsedHtmlCache = LruCache<String, CharSequence>(100)
     }
 
     override fun getItemViewType(position: Int): Int = when (getItem(position)) {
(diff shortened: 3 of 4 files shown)


Okuro3499 (approved):

dogi:
**Review — rating: 70/100**

The architectural instinct is right: `android.text.Html` has no business in a `model/` class, and moving the parse into the adapter makes `NotificationListItem` a pure data type. Worth doing.

**But it drops the memoization, and that's a real regression.** The old property was:

```kotlin
val parsedText: CharSequence by lazy(LazyThreadSafetyMode.NONE) {
    Html.fromHtml(notification.formattedText.toString(), Html.FROM_HTML_MODE_LEGACY)
}
```

— parsed **once per item**, cached for the item's lifetime. The new code calls `Html.fromHtml` inside `bind()`, so it now runs on **every bind**: every time a row scrolls into view, every rebind, every list update. `Html.fromHtml` runs a full HTML parse and builds a `Spanned` with span objects — it's one of the more expensive things you can put on a bind path, and it's now on the main thread during scroll on a list that can be long.

That also cuts against the direction of the surrounding queue — #16345 moves work *off* the bind path, and #16422 computes a display name once per row rather than per bind. This is the same pattern in reverse.

**The two goals aren't in tension.** Keep the parse in the adapter (where the platform dependency belongs) and memoize it there, e.g. an `LruCache<String, CharSequence>` keyed on `notification.id`, or a small map keyed on the id with the raw text as a validity check. That keeps the model clean *and* keeps the parse off the scroll path. Parsing once in the ViewModel/mapper when the list is built also works, though it puts a `Spanned` back into the item.

One thing that isn't a problem, for the record: the `by lazy` property was never part of the `data class`'s generated `equals`/`hashCode`, so removing it doesn't affect DiffUtil behaviour.

**Verdict:** needs another round — right move, but re-add caching in the adapter before this lands.

---

dogi:
## Review: 70/100 — needs a change

Getting `android.text.Html` out of a class in `model/` is the right move. The replacement cache is keyed on the wrong thing, and it turns a self-correcting per-instance `lazy` into a sticky process-wide one.

### The architectural half is good ✅

`NotificationListItem` sits in `model/` and was importing `android.text.Html` — a data holder reaching into the Android framework, which also meant anything unit-testing it needed Robolectric. Parsing is a presentation concern and belongs in the adapter. No argument there.

### The cache key is `notification.id`, but the value depends on `notification.formattedText` ⚠️

```kotlin
var titleText = parsedHtmlCache.get(notification.id)
if (titleText == null) {
    titleText = Html.fromHtml(notification.formattedText.toString(), …)
    parsedHtmlCache.put(notification.id, titleText)
}
```

`formattedText` is **not** a fixed function of the id. `NotificationsViewModel.formatNotification(notification, taskTeamNames, joinRequestDetails)` derives it from enrichment maps that are filled in asynchronously — `taskTeamNames` starts as `getTaskTeamNamesByTaskIds(taskIds)` and is then *augmented* by `getTaskTeamNamesByTaskTitles(taskTitles)`, and the task branch renders through `formatTaskNotification(title, date, relatedId, taskTeamNames)`.

So the same notification id legitimately produces different text across loads: before the team is resolved it renders one way, after it renders with the team name. With an id key, **the first parse wins permanently** and the enriched text never reaches the screen (until LRU eviction at 100 entries — which on a short list is never).

This is a regression specifically because of what it replaced. `parsedText` was `by lazy` **on the `Item` data class instance**, and `buildGroupedList` constructs fresh `Item`s on every rebuild — so a changed `formattedText` produced a new instance and a fresh parse. Per-instance laziness self-invalidates; an id-keyed static cache does not.

**Fix:** key on the content, not the identity.

```kotlin
val key = notification.formattedText.toString()
var titleText = parsedHtmlCache.get(key)
if (titleText == null) {
    titleText = Html.fromHtml(key, Html.FROM_HTML_MODE_LEGACY)
    parsedHtmlCache.put(key, titleText)
}
```

That keeps every bit of the performance win — repeated binds of unchanged text still hit — while making a changed message miss and re-parse, which is exactly what you want.

### Secondary: the cache is process-wide and never cleared

`parsedHtmlCache` lives in the `companion object`, so it's shared across every `NotificationsAdapter` instance and survives for the life of the process — including across logout and user switch. Ids differ between users so there's no wrong-text-for-wrong-user risk, but the previous user's notification text stays resident in memory after sign-out, and nothing in the logout path clears it. An instance field on the adapter would scope it to the screen and still capture the win (the case being optimized is re-binding during scroll and regroup, both within one adapter's lifetime).

### Worth knowing about the win itself

The change is a genuine improvement over `master` beyond the layering: because `submitList` rebuilds `Item`s on every regroup, `master` re-parsed the whole visible set each time the list was regrouped, not just on first bind. A correctly-keyed cache removes that. So this is worth doing — it just needs the right key.

### No test

A test that binds the same notification id twice with two different `formattedText` values, asserting the second render reflects the second text, would fail on this branch and pass after the key fix. Worth adding given the bug is invisible to every other kind of test.

**CI** — build (default + lite), both shards, labels green. (Green isn't evidence here — nothing exercises the cache.)

**Verdict:** switch the key to the formatted text (and consider scoping the cache to the adapter instance), then this is a merge.

---

dogi:
## Review — rating 85/100 (was 70)

The staleness bug is fixed properly. The cache is now keyed on `rawText` itself rather than the notification id, so the key fully determines the value and an edited notification can no longer render its previous text. That is the right fix rather than a patch over the symptom, and the new test pins it — same id `"1"`, different `formattedText`, asserting the title actually changes.

### The caching change is also better motivated than the title suggests

Worth noting for the record, because it is the strongest argument for this PR and it is not in the description. The old `parsedText` was a `by lazy` on `NotificationListItem.Item`:

```kotlin
val parsedText: CharSequence by lazy(LazyThreadSafetyMode.NONE) {
    Html.fromHtml(notification.formattedText.toString(), Html.FROM_HTML_MODE_LEGACY)
}
```

That memoises per *object instance*, and `Item` instances are recreated on every list emission from the ViewModel, so in practice the lazy was thrown away on each `submitList` and every rebuild re-parsed every visible row. An adapter-level LRU keyed on content survives list rebuilds, so the caching now actually does something. The diff reads like a refactor; it is really a fix for a cache that never hit.

Removing `android.text.Html` from `model/` is the correct direction independently — HTML parsing is a presentation concern and does not belong on a data class.

I checked `Notification.formattedText` is a non-null `CharSequence`, so the `.toString()` is safe and unchanged from master. `LruCache` is internally synchronised and `bind` is main-thread only, so no concurrency concern. 100 entries of short notification strings is negligible memory.

### Non-blocking nits

- Prefer `androidx.collection.LruCache` over `android.util.LruCache`. Same API, but it is not a framework class, so tests touching it do not need Robolectric. This is the same suggestion I left on #16385 — worth landing consistently across both.
- If #16385 also lands, the two content-keyed parse caches are close enough to be worth a shared helper rather than two hand-rolled get/put blocks.

Moving to `ready`.

---

PR #16383, opened 2026-08-27
sync: smoother retry interceptor backoff time providing (fixes #16330)

### LINKED ISSUE #16330: Make RetryInterceptor's backoff testable and stop sleeping wall-clock seconds in CI
`RetryInterceptor.backoff` uses `Thread.sleep` against the real clock (:88) while `TimeProvider` only drives the deadline (:78, :84). `RetryInterceptorTest` sets `initialDelay=10L` but still sleeps against real time; on master the class costs 12.3s with `testSuccessAfterIOException` alone at 10.2s.

**Ask:** route the actual sleep through the injected `TimeProvider`/dispatcher abstraction (matching how the deadline already uses it) so tests can use a virtual/test clock instead of sleeping real wall-clock seconds.

**Files:** `data/api/RetryInterceptor.kt`, `test/.../RetryInterceptorTest.kt`

**Verification:** the `Thread.sleep`/`TimeProvider` split confirmed in the source.

dogi:
Routed RetryInterceptor's backoff sleep through TimeProvider instead of calling Thread.sleep directly against real wall-clock time. Added sleep method to TimeProvider interface with implementations in SystemTimeProvider (Thread.sleep) and TestTimeProvider (advances virtual time). Updated RetryInterceptorTest to use TestTimeProvider so unit tests run instantly without sleeping real seconds.

Fixes #16330

---
*PR created automatically by Jules for task [3308411598340943660](https://jules.google.com/task/3308411598340943660) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6788
-        versionName = "0.67.88"
+        versionCode = 6789
+        versionName = "0.67.89"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/data/api/RetryInterceptor.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/data/api/RetryInterceptor.kt
@@ -85,7 +85,7 @@ class RetryInterceptor @Inject constructor(
                 if (remaining <= 0) {
                     return
                 }
-                Thread.sleep(minOf(remaining, MAX_BACKOFF_SLICE_MS))
+                timeProvider.sleep(minOf(remaining, MAX_BACKOFF_SLICE_MS))
             }
         } catch (e: InterruptedException) {
             Thread.currentThread().interrupt()
--- a/app/src/main/java/org/ole/planet/myplanet/utils/TimeProvider.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/utils/TimeProvider.kt
@@ -10,8 +10,14 @@ package org.ole.planet.myplanet.utils
  */
 interface TimeProvider {
     fun now(): Long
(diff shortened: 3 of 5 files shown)


dogi:
**Review — rating: 84/100**

Routing the backoff sleep through `TimeProvider` is the right seam, and `TestTimeProvider.sleep` advancing `currentTime` rather than no-op'ing is what makes it actually work: the retry loop recomputes `remaining` from `timeProvider.now()`, so virtual time progresses exactly as wall-clock would and the loop converges on the same number of slices. A `sleep` that simply returned would have made the loop's termination condition depend on real elapsed time and turned this into a flaky or infinite test.

**The best judgment call here is the one the diff makes quietly:** the interrupt test builds its own `RetryInterceptor(broadcastService, SystemTimeProvider())` instead of using the shared virtual one. You can't test `InterruptedException` against a clock that never blocks, so keeping exactly one genuinely-blocking test for the interrupt path while virtualizing the rest is correct — and it's the sort of thing that's easy to get wrong by virtualizing everything and silently losing the interrupt coverage.

**Two nits:**

1. `SystemTimeProvider.sleep` overrides the interface default with a byte-identical body, so one of the two is redundant. More substantively, a *defaulted* `sleep` means any future `TimeProvider` implementation silently inherits real blocking — for a seam whose whole purpose is to be substitutable in tests, making it abstract would force each implementor to decide. With only two implementors today that's a cheap change.
2. The PR delivers testability but doesn't yet spend it: no test asserts the backoff schedule. Now that `currentTime` accumulates the slept millis, `assertEquals(expectedTotal, (timeProvider as TestTimeProvider).currentTime)` after a retry sequence would pin the actual backoff behaviour — which is the thing this change exists to make observable, and is currently still unverified.

Worth noting `Thread.sleep` inside an OkHttp interceptor is fine and pre-existing — interceptors run on dispatcher threads, so blocking there is expected.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16386, opened 2026-08-27
all: smoother storage category view model testing (fixes #16376)

### LINKED ISSUE #16376: Add the missing StorageCategoryViewModel test
`StorageCategoryViewModel` has no test. `toggleItemChecked`, `toggleAllChecked` and the `deleteItems` re-entry guard (`if (_uiState.value.isDeleting) return`) are the load-bearing logic of the storage cleanup screen.

**Ask:** add a test covering `toggleItemChecked`, `toggleAllChecked`, and the `deleteItems` re-entry guard.

**Files:** `(new) test/.../ui/settings/StorageCategoryViewModelTest.kt`

dogi:
## What & Why

`StorageCategoryViewModel` had no test coverage. `toggleItemChecked`, `toggleAllChecked`, and the `deleteItems` re-entry guard (`if (_uiState.value.isDeleting) return`) are the load-bearing logic of the storage cleanup screen, so add a test that exercises all three.

Closes #16376

## Changes

- `(new) app/src/test/java/org/ole/planet/myplanet/ui/settings/StorageCategoryViewModelTest.kt` — adds unit tests covering:
  - `toggleItemChecked` flips only the matching item's `isChecked` state (and flips back on a second toggle)
  - `toggleAllChecked` checks all items when not all checked, and unchecks all when all are checked
  - `deleteItems` re-entry guard — a second `deleteItems` call made while the first is in flight is dropped, so `deleteOfflineResources` runs exactly once
  - `deleteItems` emits the `deleteCompleteEvent` after deleting and clears `isDeleting`

The test follows existing project conventions: MockK, `MainDispatcherRule` with `StandardTestDispatcher`, `TestDispatcherProvider`, and `mockkObject(FileUtils)` for `getOlePath` — matching the patterns in `ResourcesViewModelTest`, `CourseDetailViewModelTest`, and `FreeSpaceWorkerTest`.

## Verification

No JDK 17 / Android SDK is available in this sandbox, so the Gradle `testDefaultDebugUnitTest` task could not be run locally. CI (`test.yml`) will run `./gradlew testDefaultDebugUnitTest`, which fails the build on any unit-test failure. The test logic was validated against `StorageCategoryViewModel`'s source:
- `toggleItemChecked` / `toggleAllChecked` are synchronous `_uiState.update` operations verified after `advanceUntilIdle()`.
- The re-entry-guard test relies on the first `deleteItems` setting `isDeleting = true` synchronously before the launched coroutine runs (with `StandardTestDispatcher`, the body is suspended until `advanceUntilIdle()`), so the second call observes `isDeleting == true` and returns early.

---

_This PR was created by an AI agent (OpenHands) on behalf of dogi._

@dogi can click here to [continue refining the PR](https://app.all-hands.dev/conversations/11c79f76-c4b9-4245-bd01-875ff9194d07)

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6791
-        versionName = "0.67.91"
+        versionCode = 6792
+        versionName = "0.67.92"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/test/java/org/ole/planet/myplanet/ui/settings/StorageCategoryViewModelTest.kt
+++ b/app/src/test/java/org/ole/planet/myplanet/ui/settings/StorageCategoryViewModelTest.kt
@@ -1,0 +1,136 @@
+package org.ole.planet.myplanet.ui.settings
+
+import android.content.Context
+import io.mockk.Runs
+import io.mockk.coEvery
+import io.mockk.coVerify
+import io.mockk.every
+import io.mockk.just
+import io.mockk.mockk
+import io.mockk.mockkObject
+import io.mockk.unmockkAll
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.flow.first
+import kotlinx.coroutines.test.StandardTestDispatcher
+import kotlinx.coroutines.test.advanceUntilIdle
+import kotlinx.coroutines.test.runTest
+import org.junit.After
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertFalse
+import org.junit.Assert.assertTrue
+import org.junit.Before
+import org.junit.Rule
+import org.junit.Test
+import org.ole.planet.myplanet.model.OfflineResourceItem
+import org.ole.planet.myplanet.repository.ResourcesRepository
+import org.ole.planet.myplanet.utils.FileUtils
+import org.ole.planet.myplanet.utils.MainDispatcherRule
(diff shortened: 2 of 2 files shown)


dogi:
**Review — rating: 90/100**

Test-only addition for a ViewModel that had no coverage, and the tests actually test things — worth saying explicitly, because several test PRs currently in this queue assert names rather than behaviour.

The standout is `deleteItems re-entry guard`. It uses `StandardTestDispatcher` deliberately: the first `deleteItems` call is observed *before* `advanceUntilIdle()` to catch `isDeleting == true` mid-flight, a second call is made against that state, and `coVerify(exactly = 1)` then proves the guard suppressed it. That test fails if the guard is removed, which is the whole point — and it's only possible because the dispatcher choice was right. An `UnconfinedTestDispatcher` would have run the coroutine to completion and made the re-entrant window unobservable.

`toggleAllChecked` is also seeded well: starting with `res2` already checked is what distinguishes "check all when not all checked" from "flip each item", so the mixed initial state is load-bearing rather than incidental. And `toggleItemChecked` asserts both that the target flipped *and* that the sibling didn't, then flips back.

Conventions match the repo: `MainDispatcherRule` + `TestDispatcherProvider` rather than hard-coded dispatchers, `mockkObject(FileUtils)` balanced by `unmockkAll()` in `@After`.

Risk is essentially nil — no production file is touched, so the worst case is a flaky test rather than a shipped defect.

**Nit:** `loadResources` is exercised only as a fixture by the private helper; nothing asserts the loaded state itself (item count, totals, or the arguments passed through to `getOfflineResourceItems`). Since the helper already stubs it, one assertion on the post-load `uiState` would cover the load path for free.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16387, opened 2026-08-27
community: smoother configurations repository tab views modelling (fixes #16326)

### LINKED ISSUE #16326: Route community/parent-code config reads and clearPreferences through ConfigurationsRepository
`CommunityTabViewModel` reads `sharedPrefManager.getParentCode()/getCommunityName()` directly while `getPlanetType()` on the next line already goes through `configurationsRepository`; `HomeCommunityDialogFragment` reads the same two keys directly; `SettingsViewModel` injects `SharedPrefManager` (:22) solely for one `clearPreferences()` call (:47); `LeadersViewModel` injects it solely for `getCommunityLeaders()`.

**Ask:** route all of these through `ConfigurationsRepository` instead of `SharedPrefManager` directly, for consistency with the existing `getPlanetType()` call. Note: `SyncActivity:274` already calls `prefData.clearPreferences()` separately — don't fold `clearPreferences` into a broader `clearAllData` in a way that double-clears there.

**Files:** `repository/ConfigurationsRepository.kt`, `repository/ConfigurationsRepositoryImpl.kt`, `ui/community/CommunityTabViewModel.kt`, `ui/community/LeadersViewModel.kt`, `ui/settings/SettingsViewModel.kt`

**Verification:** all four call sites confirmed.

dogi:
Routed community/parent-code config reads and clearPreferences through ConfigurationsRepository instead of accessing SharedPrefManager directly across UI components.

Fixes #16326

---
*PR created automatically by Jules for task [18126929720452606005](https://jules.google.com/task/18126929720452606005) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6881
-        versionName = "0.68.81"
+        versionCode = 6882
+        versionName = "0.68.82"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/repository/ConfigurationsRepository.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/ConfigurationsRepository.kt
@@ -12,6 +12,10 @@ interface ConfigurationsRepository {
     suspend fun clearAllData()
     suspend fun getMinApk(url: String, pin: String): ConfigurationResult
     fun getPlanetType(): String?
+    fun getParentCode(): String
+    fun getCommunityName(): String
+    fun getCommunityLeaders(): String
+    fun clearPreferences()
     suspend fun ensureServerUrlUpdated()
     suspend fun clearFirstRunStorageAndSetFlag(hasWritePermission: Boolean)
     suspend fun getQueuedDownloads(): List<String>
--- a/app/src/main/java/org/ole/planet/myplanet/repository/ConfigurationsRepositoryImpl.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/ConfigurationsRepositoryImpl.kt
@@ -393,6 +393,22 @@ class ConfigurationsRepositoryImpl @Inject constructor(
         return sharedPrefManager.getRawString("planetType")
     }
 
+    override fun getParentCode(): String {
+        return sharedPrefManager.getParentCode()
+    }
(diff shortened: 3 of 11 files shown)


dogi:
**Review — rating: 82/100**

Routes four community/config reads through `ConfigurationsRepository` so `CommunityTabViewModel`, `LeadersViewModel`, `SettingsViewModel` and `HomeCommunityDialogFragment` stop injecting `SharedPrefManager` directly. The methods are thin pass-throughs, but that's the documented layering in this repo (UI → Repository → data sources), and the payoff is concrete: three ViewModels become testable against a repository mock, which is what the three new test files demonstrate.

The tests are real rather than nominal. `LeadersViewModelTest` feeds actual leaders JSON and asserts the parsed `id` — it exercises `UserEntity.parseLeadersJson`, not just the delegation — and covers the empty-string branch separately. `SettingsViewModelTest` verifies both `clearAllData` **and** `clearPreferences` are called and that the event is emitted, which is the pairing that matters since dropping either would leave the app half-reset.

**⚠️ This conflicts with #16370, which I've just marked `ready`.** Both edit `HomeCommunityDialogFragment` in the same region, and they pull in different directions: this PR keeps `configurationsRepository` injected in the fragment and re-points the calls at it, while #16370 removes *both* injections and drives the fragment from `CommunityTabViewModel` instead. #16370 is the further-along end state — it makes this PR's fragment hunk moot.

The good news is the two compose cleanly in substance: #16370's fragment (VM-driven) plus this PR's `CommunityTabViewModel` (repo-driven) is a coherent end state, since #16370 doesn't touch the VM's internals. It's purely a textual conflict, so whichever lands second needs a rebase that keeps #16370's fragment body and this PR's ViewModel body. Worth not queueing them back to back in the drain.

**Two nits in the new tests:**

1. `LeadersViewModelTest` carries unused imports — `io.mockk.just`, `mockkObject`, `runs`, `unmockkObject`, `org.junit.Before`, `org.junit.After` — with no corresponding usages or setup/teardown methods, plus a stray double blank line. Worth a pass given the repo runs a `kotlin-importing` skill for exactly this.
2. Both `LeadersViewModelTest` and `SettingsViewModelTest` hand-roll an anonymous `DispatcherProvider` with all five dispatchers set to the same test dispatcher. `TestDispatcherProvider` already exists as shared test infra and is used elsewhere in this queue — reusing it drops ~7 lines from each file and keeps the two in sync if the interface gains a dispatcher.

**Verdict:** mergeable — resolve the merge order with #16370 rather than the code.

---

Okuro3499 (approved):

PR #16388, opened 2026-08-27
sync: smoother download repository file handling (fixes #16356)

### LINKED ISSUE #16356: Route DownloadRepositoryImpl's file-not-found logging through DiagnosticsRepository
`DownloadRepositoryImpl` imports `MainApplication.Companion.createLog` (:9) and calls it at :51/:53 — a data-layer repository reaching into the application singleton's static for a side effect (persisting an apk-log row) that the logging repository (`DiagnosticsRepository`) already owns.

**Ask:** inject `DiagnosticsRepository` and route these log calls through it instead of `MainApplication.createLog`. Note: this repository's test constructor is used at 8 sites, and there are two existing `MainApplication.createLog` verifications in tests — both need updating.

**Files:** `repository/DownloadRepositoryImpl.kt`, `test/.../DownloadRepositoryImplTest.kt`

**Verification:** confirmed the import and both call sites.

dogi:
fixes #16356

`DownloadRepositoryImpl` imported `MainApplication.Companion.createLog` and called it at two sites when a download returned a 404 — a data-layer repository reaching into the application singleton's static for a side effect (persisting an apk-log row) that `DiagnosticsRepository` already owns.

This injects `DiagnosticsRepository` (plus `TimeProvider`, which supplies the timestamp `createLog` previously generated internally via `coreDependenciesEntryPoint.timeProvider().now()`) and routes both 404 log calls through `diagnosticsRepository.saveLogToRoom(type, error, time)` instead of `MainApplication.createLog`. Both dependencies are already provided/bound in the Hilt graph (`RepositoryModule.bindDiagnosticsRepository`, `TimeModule.provideTimeProvider`), so the `@Inject constructor` wires up with no module changes.

**Files:**
- `repository/DownloadRepositoryImpl.kt` — drop the `MainApplication.Companion.createLog` import; add `diagnosticsRepository` and `timeProvider` constructor params; replace both `createLog(...)` call sites with `diagnosticsRepository.saveLogToRoom("File Not Found", …, "${timeProvider.now()}")`.
- `test/.../DownloadRepositoryImplTest.kt` — drop the `mockkObject(MainApplication)` setup/teardown and `MainApplication` import; supply a relaxed `DiagnosticsRepository` mock and a deterministic `TestTimeProvider`; update all constructor call sites; switch the two `MainApplication.createLog` verifications to `coVerify { diagnosticsRepository.saveLogToRoom("File Not Found", …, "123456789") }`.

**Verification:** `./gradlew testDefaultDebugUnitTest --tests "org.ole.planet.myplanet.repository.DownloadRepositoryImplTest"` — 10/10 tests pass (0 failures, 0 errors), including the two logging-verification tests.

---

@dogi can click here to [continue refining the PR](https://app.all-hands.dev/conversations/806569ad-dab6-4f4c-ad25-a188f7359ab5)


dogi:
**Review — rating: 88/100**

Replaces the global `MainApplication.createLog` with an injected `DiagnosticsRepository` plus a `TimeProvider` for the timestamp. Two dependencies added, one static reach-through removed, and the class becomes constructible in a test without touching the Application object.

**This is the pattern #16353 should have used.** That PR faced the same problem — a platform/global logging dependency making a class awkward to test — and solved it by deleting the diagnostic, which left a silent failure on an integrity check. Here the signal is preserved and *improved*: it lands in Room where it can actually be retrieved, rather than in logcat.

The test change is the strongest evidence it worked. `mockkObject(MainApplication)` and its `@After` teardown are gone entirely, and the assertions got stronger rather than weaker:

```kotlin
coVerify { diagnosticsRepository.saveLogToRoom("File Not Found", url, "123456789") }
```

The timestamp is now asserted, which was impossible when time came from a global. `TestTimeProvider(currentTime = 123456789L)` makes that deterministic.

I checked the failure-path safety, since this puts a suspend call inside a `catch` block: `DiagnosticsRepositoryImpl.saveLogToRoom` wraps its own body in try/catch and returns a Boolean rather than throwing, so the second `saveLogToRoom` — the one in the catch arm, where an exception would propagate uncaught — can't turn a download error into a crash. No dependency cycle either: `DiagnosticsRepository` pulls `ApkLogDao`/`SharedPrefManager`/`UserSessionManager` and nothing routes back to `DownloadRepository`.

**Nit:** the constructor is now four parameters and the test repeats the full instantiation at ten call sites. A small `private fun repository() = DownloadRepositoryImpl(...)` helper would make the next parameter addition a one-line change instead of a ten-line one.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16390, opened 2026-08-27
actions: smoother workflow release versioning (fixes #16349)

### LINKED ISSUE #16349: Remove the duplicated version-name parsing from release.yml
`release.yml` parses `versionName` out of `app/build.gradle` twice with the identical `sed` — once in `set release version` (:40) and again inside `build release APK and AAB` (:55) — in the same job, where `ANDROID_VERSION` is already exported. The `versionCode` read (:60) strips all non-digits, which breaks on a trailing comment.

**Ask:** route both version reads through the repo's existing `.github/scripts/version.sh read` (the canonical reader `automerge.sh` already uses) instead of duplicating the `sed` inline, and fix the versionCode read to not break on a trailing comment.

**Files:** `.github/workflows/release.yml` (version steps only)

**Verification:** confirmed both `sed` lines duplicate the same parse.

dogi:
Refactored `.github/workflows/release.yml` to call `.github/scripts/version.sh read app/build.gradle` in the `set release version` step, exporting both `ANDROID_VERSION` and `ANDROID_VERSION_CODE` to `$GITHUB_ENV`. Removed duplicate `versionName` `sed` parsing and fragile `versionCode` `grep | sed` line from `build release APK and AAB`.

Fixes #16349

---
*PR created automatically by Jules for task [12112477382309786126](https://jules.google.com/task/12112477382309786126) started by @dogi*

## Summary by CodeRabbit

- **Chores**
  - Improved the Android release process to reliably determine version names and codes.
  - Ensured release builds consistently use the configured application version, including existing “lite” variant handling.


dogi:
**Review — rating: 89/100**

I verified the contract this depends on, since an empty version here would silently produce a `v` tag and a broken release name. `version.sh read` prints exactly:

```
code=$cur_code
name=$cur_name
```

so `eval "$(…)"` populates `code` and `name`, which is what the workflow reads. It's executable (`-rwxr-xr-x`), and `release.yml` still checks out the repo, so it's on disk when the step runs.

**The replaced parser was genuinely fragile**, which makes this more than a tidy-up:

```bash
grep versionCode app/build.gradle | sed 's/[^0-9]//g'
```

That strips non-digits from **every** line containing `versionCode` and concatenates the survivors — a comment mentioning versionCode, or a flavor-level override, would silently yield a mangled number that still looks like a version code. `version.sh` anchors on `^[[:space:]]*versionCode[[:space:]]*=`, takes `head -1`, and `die`s under `set -euo pipefail` if either field is missing. Failing loudly beats publishing a wrong version code to the Play Store, where it can't be taken back.

**The consistency win is the real payoff.** `automerge.yml` (lines 107, 133) stages this same script and hands it to the drain as `VERSION_SH`, which is what *writes* the bumped version. Before this PR the release workflow parsed with its own inline sed/grep, so reader and writer could drift; now one script defines the format for both. That's precisely the drift that produces a tag not matching the built artifact.

Moving `ANDROID_VERSION_CODE` into the earlier step is also strictly safer — it's now set before anything could consume it, rather than mid-build.

**Verdict:** mergeable.

---

Okuro3499 (dismissed):
```
.github/workflows/release.yml:41 — eval "$(.github/scripts/version.sh read app/build.gradle)" does not check version.sh's exit status. If read_version calls die (e.g.           
    app/build.gradle is missing or its versionCode/versionName lines don't match), die writes to stderr and exits 1, so the command substitution yields an empty stdout, eval ""     
    succeeds (exit 0), and $code/$name end up unset. The job then continues with ANDROID_VERSION= and ANDROID_VERSION_CODE= empty, silently producing a broken release tag           
    (refs/tags/v), a garbage Play Store release name, and mangled artifact filenames — instead of failing loudly the way version.sh's die() was clearly designed to. Fix: capture and
    check the exit code explicitly, e.g.                                                                                                                                             
  version_output=$(.github/scripts/version.sh read app/build.gradle)                                                                                                                 
  eval "$version_output"                                                                                                                                                             
    (this fails under -e if the script itself exits non-zero, since the $(...) is now the direct RHS of a plain assignment rather than buried inside eval's argument). I verified    
    this failure mode locally: eval "$(failing_cmd)" under bash -eo pipefail does not propagate the inner command's non-zero exit and execution continues silently.                  
                                                                                                                                                                                     
    Note: this exact pattern (and the exact same latent bug) already exists in automerge.sh:484 (eval "$("$VERSION_SH" next ... | sed ...)", made worse there by piping through sed, 
    which also defeats pipefail). So this PR isn't introducing a novel risk — it's importing a pre-existing footgun into a second, arguably higher-stakes workflow. Worth fixing in  
    both places, but flagging it here since it's the file in this diff.         
```

dogi:
## Review: 87/100 — mergeable

More than DRY: the two expressions being replaced are both fragile in ways `version.sh` isn't, and one of them fails in a genuinely nasty way.

### The old `versionCode` extraction was a latent corruption of `$GITHUB_ENV`

```bash
echo ANDROID_VERSION_CODE=$(grep versionCode app/build.gradle | sed 's/[^0-9]//g') >> $GITHUB_ENV
```

`grep` is unanchored and unlimited. Today it matches exactly one line (`app/build.gradle:13`), so it works. But any *second* line containing the substring `versionCode` — a comment like `// bump versionCode before release`, or a second flavor block — makes `grep` emit two lines, `sed` strips non-digits from each, and the `echo` writes a **multi-line value** into `$GITHUB_ENV`. That file is parsed line-by-line as `KEY=value`, so the stray second line lands as a garbage env entry rather than failing loudly. A release publishing under a silently wrong version code is expensive to unwind given the Play Store save quota.

`version.sh read` is immune: anchored `^[[:space:]]*versionCode[[:space:]]*=[[:space:]]*([0-9]+)` plus `head -1`, and it `die`s if it finds nothing. Same for `versionName` (the old `sed` there was also unanchored, though nothing in the file currently false-matches it).

### The consistency argument is the real win

`automerge.sh` already uses `version.sh` to **write** the bump during a drain. After this, `release.yml` **reads** through the same parser — so the writer and reader can't drift on format. That's the class of bug that only shows up on a release, which is the worst time to find it.

### Verified

- `.github/scripts/version.sh` is mode `-rwxr-xr-x` in the tree, so calling it directly without `chmod +x` works. (Worth noting because #16366 needs the opposite treatment for `labels.sh`.)
- `version.sh read` emits exactly `code=<digits>` then `name=<version>`, so `eval` populating `${code}` / `${name}` is correct.
- `ANDROID_VERSION` is written to `$GITHUB_ENV` in *set release version* and consumed as a shell variable in the later *build* step — subsequent-step visibility, which is right.
- Moving `ANDROID_VERSION_CODE` earlier is harmless: its only consumer is *set playstore variables*, which runs after both.
- `VERSION="${ANDROID_VERSION}"` preserves the `-lite` suffix logic untouched.

### Nit: `eval` is avoidable

```bash
eval "$version_output"
```

`name` comes from `[^"]+` inside `app/build.gradle`, so a versionName containing shell metacharacters would be executed. The threat model is weak — writing that string requires commit access to `master`, which already means arbitrary execution via Gradle — but the parse is just as easy without `eval`:

```bash
while IFS='=' read -r k v; do
  case "$k" in
    code) echo "ANDROID_VERSION_CODE=$v" >> "$GITHUB_ENV" ;;
    name) echo "ANDROID_VERSION=$v" >> "$GITHUB_ENV" ;;
  esac
done < <(.github/scripts/version.sh read app/build.gradle)
```

### Can't be validated by this PR's CI

`release.yml` only runs on `master` push, so nothing here exercises the change — the green checks are build/test/labels only. Same caveat as #16311 and #16366 in this queue. Worth watching the first release run after merge; a failure surfaces at *set release version* with a clear `version.sh:` message rather than silently, which is itself an improvement over the old form.

**Verdict:** mergeable.

---

dogi:
## Review — rating 87/100 (unchanged)

The diff is the same one I reviewed at 87. Re-verified the script contract since the whole release depends on it:

```
case "${1:-}" in
    read)
        read_version "${2:?gradle file required}"
        echo "code=$cur_code"
        echo "name=$cur_name"
```

`version.sh read` emits `code=` and `name=`, which is what `eval "$version_output"` then `${name}`/`${code}` expects. The variable names line up (the function's internals are `cur_*`, but the dispatch renames them on output), so `ANDROID_VERSION` and `ANDROID_VERSION_CODE` are populated correctly. `ANDROID_VERSION_CODE` moves earlier in the job but is still set well before its only consumer at line 88, and `ANDROID_VERSION_NAME` keeps the `-lite` suffix logic.

Worth restating why this matters, since it reads as cosmetic. The line it replaces is:

```sh
echo ANDROID_VERSION_CODE=$(grep versionCode app/build.gradle | sed 's/[^0-9]//g')
```

`grep` matches *any* line containing `versionCode` and `sed` then strips non-digits across all of them concatenated. A second occurrence — a comment, a flavor block — silently yields a mashed-together number that gets shipped to the Play Store as a version code, with nothing failing. `version.sh` anchors on `^[[:space:]]*versionCode[[:space:]]*=`, takes `head -1`, and `die`s under `set -euo pipefail` when it finds nothing. Loud failure instead of a wrong upload. It also makes the release read versions through the same helper `automerge.sh` uses to write them, so the two can no longer drift.

Non-blocking: `eval` on command output is a smell. The values here are repo-controlled and anyone who can edit `build.gradle` already executes arbitrary code via Gradle in CI, so it is not an escalation — but having `version.sh` append to `"$GITHUB_ENV"` itself, or reading the two values without `eval`, would be tidier.

Heads-up on ordering: #16311 also edits `release.yml`, inserting a cache step immediately before `set release version` and adding a top-level `concurrency:` block. Different hunks, so it should auto-merge, but whichever lands second is worth a glance.

### Label state

This PR was carrying both `merge` and `review`, which are mutually exclusive. It has no star tag so it is not eligible for promotion through the normal merge gate, but a maintainer has already queued it at `merge`, which is the later stage — so I am keeping `merge` and dropping `review` rather than demoting it back to `ready` and pulling it out of the drain.

---

Okuro3499 (approved):

PR #16392, opened 2026-08-27
courses: less submission detail measure is more (fixes #16348)

### LINKED ISSUE #16348: Delete the hand-rolled RecyclerView measure pass in SubmissionDetailFragment
`setupRecyclerView` installs an anonymous `LinearLayoutManager` whose `onMeasure` loops over every item calling `getViewForPosition`/`addView`/`measureChild`/`getDecoratedMeasuredHeight`/`removeAndRecycleView` to sum a height — a full inflate-and-measure of the whole question list on every measure pass. The layout already declares `wrap_content` with `nestedScrollingEnabled=false` inside a `ScrollView`, which `RecyclerView` has supported natively since support-lib 23.2.

**Ask:** delete the hand-rolled measure override and rely on native `wrap_content` support. **Careful:** the same anonymous subclass also overrides `canScrollVertically()` to return `false` — keep that override (or otherwise verify the enclosing `ScrollView` still owns scrolling) or scrolling breaks.

**Files:** `ui/submissions/SubmissionDetailFragment.kt`

**Verification:** confirmed the whole anonymous subclass matches this description.

dogi:
Removed the custom onMeasure implementation in setupRecyclerView inside SubmissionDetailFragment.kt that was manually inflating and measuring all items, relying instead on RecyclerView's native wrap_content support while keeping canScrollVertically() = false.

Fixes #16348

---
*PR created automatically by Jules for task [5844794168783349441](https://jules.google.com/task/5844794168783349441) started by @dogi*


dogi:
**Review — rating: 90/100**

The layout confirms this override was reimplementing framework behaviour. `fragment_submission_detail.xml` gives `rv_questions_answers` `layout_height="wrap_content"` and `nestedScrollingEnabled="false"`, inside a `ScrollView` → vertical `LinearLayout`. That's the standard idiom for an inline non-scrolling list, and RecyclerView's auto-measure has handled `wrap_content` natively since 23.2.0 — with `canScrollVertically()` returning false, the height is treated as unbounded and every child is laid out, which is exactly what the manual measure was computing by hand.

Worth noting the deleted code wasn't merely redundant, it was working against the widget:

- It called `recycler.getViewForPosition(i)` for **every** item on every measure pass, then added, measured, and recycled each one. On a submission with many questions that's O(n) view creation per layout — recycling defeated by the code meant to help it.
- The blanket `catch (e: Exception)` fell back to `super.onMeasure` silently, so any failure in that loop turned into a quiet layout change with no signal.
- `addView` / `removeAndRecycleView` inside `onMeasure` mutates the LayoutManager's child list mid-measure, which is fragile enough that the try/catch was probably load-bearing.

Keeping `canScrollVertically() = false` is correct — that's what makes the parent `ScrollView` own the scrolling — and it doesn't interfere with auto-measure.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16394, opened 2026-08-27
test: smoother leaders view model loading (fixes #16380)

### LINKED ISSUE #16380: Add the missing LeadersViewModel test
`LeadersViewModel.loadLeaders` runs in `init` and either sets `emptyList()` or `UserEntity.parseLeadersJson(string)` — the whole data source of the screen, untested. `parseLeadersJson` uses `org.json`, so Robolectric is required.

**Ask:** add a test covering both branches of `loadLeaders` (empty-list case and the `parseLeadersJson` case), using Robolectric since `org.json` is involved.

**Files:** `(new) test/.../ui/community/LeadersViewModelTest.kt`

**Verification:** confirmed no such test file exists; `LeadersViewModel` matches the description exactly.

dogi:
## Summary

Adds the missing `LeadersViewModel` unit test. `LeadersViewModel.loadLeaders` runs in `init` and is the entire data source for the Leaders screen, yet it had no test coverage. This covers both of its branches.

## Changes

- **(new)** `app/src/test/java/org/ole/planet/myplanet/ui/community/LeadersViewModelTest.kt`

Two Robolectric tests (Robolectric is required because `UserEntity.parseLeadersJson` parses via `org.json`):

1. **empty-list branch** — when `SharedPrefManager.getCommunityLeaders()` returns `""`, `leaders` resolves to an empty list.
2. **parseLeadersJson branch** — when the leaders string is non-empty, `leaders` is parsed into `UserEntity` instances with the expected `id`/`name`/`firstName`/`lastName`/`email`.

Mirrors existing conventions:
- `MainDispatcherRule` + `TestDispatcherProvider` (`StandardTestDispatcher`) with `runTest(testDispatcher)` + `advanceUntilIdle()` to drive the coroutine launched in `init`, as in `NotificationsViewModelTest` / `UserProfileViewModelTest`.
- `@RunWith(RobolectricTestRunner::class)` + `@Config(application = Application::class)` (a plain `Application`, like `MeetupTest` / `EventsAdapterTest`) so Robolectric does not instantiate `MainApplication`, whose deferred init calls `SecurePrefs.warmUp()` and fails on Robolectric's absent `AndroidKeyStore`.
- The leaders JSON shape matches `UserEntityParseLeadersTest`.

## Verification

```
./gradlew testDefaultDebugUnitTest --tests "org.ole.planet.myplanet.ui.community.LeadersViewModelTest"
```

```
> Task :app:testDefaultDebugUnitTest
BUILD SUCCESSFUL
```

Result: `tests=2 failures=0 errors=0 skipped=0`.

Fixes #16380.

---

@dogi can click here to [continue refining the PR](https://app.all-hands.dev/conversations/8099f736-a7ef-4c86-b2fc-dfbcfd30787f)

### DIFF
--- a/app/src/test/java/org/ole/planet/myplanet/ui/community/LeadersViewModelTest.kt
+++ b/app/src/test/java/org/ole/planet/myplanet/ui/community/LeadersViewModelTest.kt
@@ -1,0 +1,74 @@
+package org.ole.planet.myplanet.ui.community
+
+import android.app.Application
+import io.mockk.every
+import io.mockk.mockk
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.test.StandardTestDispatcher
+import kotlinx.coroutines.test.advanceUntilIdle
+import kotlinx.coroutines.test.runTest
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertTrue
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.ole.planet.myplanet.services.SharedPrefManager
+import org.ole.planet.myplanet.utils.MainDispatcherRule
+import org.ole.planet.myplanet.utils.TestDispatcherProvider
+import org.robolectric.RobolectricTestRunner
+import org.robolectric.annotation.Config
+
+@OptIn(ExperimentalCoroutinesApi::class)
+@RunWith(RobolectricTestRunner::class)
+@Config(application = Application::class)
+class LeadersViewModelTest {
+
+    private val testDispatcher = StandardTestDispatcher()
+
+    @get:Rule
+    val mainDispatcherRule = MainDispatcherRule(testDispatcher)
+
+    private val sharedPrefManager: SharedPrefManager = mockk(relaxed = true)
+
+    @Test
+    fun `loadLeaders sets an empty list when community leaders string is empty`() = runTest(testDispatcher) {
+        every { sharedPrefManager.getCommunityLeaders() } returns ""
+
+        val viewModel = LeadersViewModel(sharedPrefManager, TestDispatcherProvider(testDispatcher))
+        advanceUntilIdle()
+
+        val leaders = viewModel.leaders.value
+        assertEquals(0, leaders?.size)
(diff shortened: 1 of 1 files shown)


dogi:
**Review — rating: 80/100 (duplicate)**

On its own terms this is a good test file — better written, in several specific ways, than the one it collides with. It uses the shared `TestDispatcherProvider` rather than hand-rolling an anonymous `DispatcherProvider`, carries no unused imports, pairs `StandardTestDispatcher` with `MainDispatcherRule(testDispatcher)` consistently, and asserts five parsed fields (`id`, `name`, `firstName`, `lastName`, `email`) rather than just the id — so it actually exercises `UserEntity.parseLeadersJson` instead of merely reaching it.

**But #16387 creates the same file at the same path**, `app/src/test/java/org/ole/planet/myplanet/ui/community/LeadersViewModelTest.kt`, as a new file — an add/add conflict that git can't resolve on its own. Worse than the textual clash, the two are semantically incompatible: #16387 changes `LeadersViewModel`'s constructor from `SharedPrefManager` to `ConfigurationsRepository`, and this test constructs `LeadersViewModel(sharedPrefManager, …)`. Whichever order they land in, this file stops compiling once #16387's refactor is in.

#16387 has to be the survivor: it carries the architectural change (routing community config reads through the repository, with four consumers updated) that this test would otherwise block, and it's already `ready`. Picking this PR instead would mean reverting that refactor to keep a test compiling, which is the tail wagging the dog.

**Recommend closing in favour of #16387.** Marking `duplicate`.

**Worth salvaging, though** — the three things this file does better are exactly the nits I raised on #16387's copy. Porting them over is a small edit and makes the surviving version strictly better:

1. Replace the hand-rolled anonymous `DispatcherProvider` with `TestDispatcherProvider(testDispatcher)`.
2. Drop the unused imports (`io.mockk.just`, `mockkObject`, `runs`, `unmockkObject`, `org.junit.Before`, `org.junit.After`).
3. Assert the full parsed `UserEntity` rather than size and id.

The only adjustment needed is stubbing `configurationsRepository.getCommunityLeaders()` in place of `sharedPrefManager.getCommunityLeaders()`.

---

PR #16395, opened 2026-08-27
enterprises: smoother repository reports csv view modelling (fixes #16374)

### LINKED ISSUE #16374: Let the enterprises CSV export read its own data instead of taking the UI's list
`exportReportsAsCsv(reports: List<MyTeam>, teamName)` takes already-loaded entities back from the UI; `EnterprisesReportsFragment:84` passes its own `reports` field assigned from the adapter callback, so the exported CSV silently depends on what the screen last rendered — export tapped before the flow emits writes a header-only file.

**Ask:** have the repository fetch its own data for the export instead of taking a pre-loaded list from the UI — fixes the header-only-file bug where export-before-first-emission produces an empty CSV.

**Files:** `repository/EnterprisesRepository.kt`, `repository/EnterprisesRepositoryImpl.kt`, `ui/enterprises/EnterprisesViewModel.kt`, `ui/enterprises/EnterprisesReportsFragment.kt`, `test/.../EnterprisesRepositoryImplTest.kt`

**Verification:** confirmed the whole call chain and the header-only-file bug.

dogi:
Refactored `EnterprisesRepository.exportReportsAsCsv` to take `teamId: String` instead of `reports: List<MyTeam>`. The repository now directly queries `teamDao.getByTeamIdAndDocType(teamId, "report")`, filtering out archived reports and sorting by `createdDate` descending before generating the CSV. Updated `EnterprisesViewModel`, `EnterprisesReportsFragment`, and `EnterprisesRepositoryImplTest` accordingly.

Fixes #16374

---
*PR created automatically by Jules for task [4028621344952507586](https://jules.google.com/task/4028621344952507586) started by @dogi*


dogi:
**Review — rating: 80/100**

The motivation holds up: the fragment previously handed the export whatever `reports` list it happened to be holding from the flow, so the CSV reflected a UI snapshot rather than the database. Reading fresh inside the repository means the export can't silently write stale rows, and it drops a parameter the caller only had by coincidence. The test update is a genuine improvement too — seeding an `archived` report and asserting it's excluded covers a filter that had no coverage before.

**The cost is a duplicated predicate on a data-export path.** `getReportsFlow` already does exactly:

```kotlin
.filter { it.status != "archived" }
.sortedByDescending { it.createdDate }
```

and this PR adds the same two lines to `exportReportsAsCsv`. Now "which reports count" is defined in two places, and they can drift — with the failure mode being an exported CSV that quietly disagrees with what the user sees on screen, which is worse than a visible bug because it looks authoritative.

**#16397, sitting right behind this one in the queue, is literally "Move non-archived report filter and ordering into TeamDao."** That's the fix for this duplication: once the predicate lives in a DAO query, both call sites should use it rather than each re-deriving it in Kotlin. Whichever of the two lands second should collapse this copy into that query — worth agreeing the order now rather than discovering it later.

**Nit:** the test seeds only one non-archived report, so the `sortedByDescending { createdDate }` half is unexercised. A second non-archived row with an earlier `createdDate` would pin the ordering, which matters because CSV row order is user-visible.

**Verdict:** mergeable — no defect, but plan the merge order with #16397 so the duplicated filter doesn't outlive it.

---

Okuro3499 (approved):

PR #16396, opened 2026-08-27
actions: smoother workflow automerge coauthoring (fixes #16375)

### LINKED ISSUE #16375: Replace the per-candidate grep dedup in coauthor crediting with an associative array
`coauthors.sh`'s dedup loop calls `printf | grep -qxF` twice per candidate (:83, :86), spawning two grep processes per candidate — O(n²) in collaborators and the slowest part of every automerge squash.

**Ask:** replace both `grep -qxF` lookups with Bash 4 associative-array membership checks (O(1) each).

**Files:** `.github/scripts/coauthors.sh`

**Verification:** confirmed both `grep -qxF` calls at `coauthors.sh:83` and `:86`.

dogi:
Replaces per-candidate `grep -qxF` lookups in `.github/scripts/coauthors.sh` with Bash 4 associative array membership checks (`declare -A collaborators` and `declare -A seen`), eliminating O(n²) process spawns during automerge squash.

Fixes #16375

---
*PR created automatically by Jules for task [1695303915522939231](https://jules.google.com/task/1695303915522939231) started by @dogi*


dogi:
**Review — rating: 70/100**

The `seen` half is clean and the details that usually go wrong here are right: `< <(…)` process substitution instead of a pipe keeps the `while` loop out of a subshell (a pipe would have silently discarded every array write), and both lookups use `${arr["$l"]:-}` — necessary under this script's `set -u`, where a bare unset-key read would abort. Dropping `sort -u` is also correct, since an associative array dedupes by construction and only membership is tested.

**But the `collaborators` rewrite breaks error propagation, and the line directly above it says that's the one thing not to do:**

```bash
# Needs push access; failing here beats crediting nobody.
```

The old form was a command substitution under `set -euo pipefail`:

```bash
collaborators=$(gh api "repos/$REPO/collaborators…" --paginate --jq … | tr … | sort -u)
```

A failing `gh api` — no push access, a 5xx, a rate limit — aborted the script, exactly as the comment intends. Process substitution has no such guarantee: `set -e` and `pipefail` don't apply to it, so a failed `gh api` now just yields no lines. The loop reads nothing, `collaborators` stays empty, every candidate fails the membership test, and the script exits 0 having credited nobody. That's the silent no-op the comment was written to prevent, and it's invisible — a squash commit with no `Co-authored-by` trailers looks the same as a PR that genuinely had no collaborators.

**Fix keeps both properties** — capture first, then fill the array from a here-string:

```bash
collab_raw=$(
    gh api "repos/$REPO/collaborators?per_page=100" --paginate \
        --jq '.[] | select(.type == "User") | .login' | tr '[:upper:]' '[:lower:]'
)
while IFS= read -r col; do
    [ -n "$col" ] && collaborators["$col"]=1
done <<<"$collab_raw"
```

The command substitution still aborts on failure under `set -e`/`pipefail`, and the loop still runs in the current shell so the array survives.

Worth noting the performance case here is stronger than the equivalent in #16384 — each old `grep` was a process spawn rather than an in-process scan — though with a handful of candidates per PR it's still not the reason to merge this. Clarity is.

**Verdict:** needs another round — keep the associative arrays, restore the failure propagation on the collaborator fetch.

---

dogi:
## Review: 86/100 — mergeable

Behavior-preserving and cleaner. The one thing that usually breaks this kind of rewrite is handled correctly.

### `set -u` is respected ✅

`coauthors.sh` runs under `set -euo pipefail`, and under `set -u` a bare `${collaborators["$l"]}` on a missing key **aborts the script**. Both lookups use the `:-` default:

```bash
if [[ -n "${seen["$l"]:-}" ]]; then
if [[ -z "${collaborators["$l"]:-}" ]]; then
```

That's the detail that would have turned a tidy-up into a broken automerge drain, and it's right.

### Equivalence checks

- **`sort -u` dropped:** fine. It existed only to dedupe for `grep -qxF` membership tests; assigning the same key twice into an associative array is idempotent, and ordering was never used.
- **`grep -qxF` → key lookup:** `-x -F` was a whole-line fixed-string match, which is exactly an exact key match. If anything the array is safer — no pattern semantics to get wrong.
- **Empty input:** `<<<"$collab_raw"` on an empty string yields one empty line, guarded by `if [ -n "$col" ]`. An empty collaborator set still means "credit nobody", same as before.
- **Failure behavior unchanged:** `collab_raw=$(gh api …)` is still a simple assignment, so a failing `gh api` still aborts under `set -e` — preserving the "failing here beats crediting nobody" intent in the comment above it.
- **`declare -A` needs bash 4+**; this runs on `ubuntu-24.04` (bash 5.2), and only via `automerge.yml`, so there's no portability exposure.

### Scope note on the framing

The title says "per-candidate grep dedup" and an "O(n²) index scan" is what the linked issue describes, but in practice the candidate list comes from co-author trailers in one PR body — usually a handful. The old code spawned two processes (`printf` + `grep`) per candidate per check; the saving is a dozen or so process spawns on a script that already makes two `gh` API round trips. So this is really a readability and robustness win, not a measurable speed-up. Worth saying plainly rather than leaning on the complexity class.

### Gap: nothing exercises this

`coauthors.sh` only runs inside the automerge drain (`automerge.yml:108`), so this PR's green checks say nothing about the change — it takes effect on the next drain, and a mistake there stops the queue.

`labels.sh` solved this with a `FILES_JSON` env hook that lets a test feed it fixture data instead of calling `gh api`. The equivalent here — a `COLLABORATORS_RAW` and `PR_JSON` override — would make this script testable and would have let this refactor prove itself. This PR is exactly the moment that hook earns its keep; worth a follow-up issue at minimum.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16397, opened 2026-08-27
enterprises: smoother reports dao filtering (fixes #16373)

### LINKED ISSUE #16373: Move the enterprises non-archived report filter and ordering into TeamDao
`EnterprisesRepositoryImpl.getReportsFlow` calls `teamDao.observeByTeamIdAndDocType(teamId,"report")` then `.filter { it.status != "archived" }.sortedByDescending { it.createdDate }` in memory, re-running a static predicate and ordering on every emission.

**Ask:** push the non-archived filter (`IFNULL(status,'') != 'archived'`) and the descending `createdDate` ordering into a new, narrower SQL query. `observeByTeamIdAndDocType` is shared with a `TeamsRepositoryImpl` transaction caller and must stay as-is — add a new DAO method rather than narrowing the existing one.

**Files:** `data/room/dao/TeamDao.kt`, `repository/EnterprisesRepositoryImpl.kt`, `test/.../EnterprisesRepositoryImplTest.kt`

**Verification:** confirmed at `EnterprisesRepositoryImpl.kt:87-99`.

dogi:
Pushed non-archived report status filtering (IFNULL(status, '') != 'archived') and descending createdDate ordering into a new observeNonArchivedReportsByTeamId query in TeamDao. Updated EnterprisesRepositoryImpl to consume this query flow directly and added unit tests in EnterprisesRepositoryImplTest.

Fixes #16373

---
*PR created automatically by Jules for task [7815129876896733136](https://jules.google.com/task/7815129876896733136) started by @dogi*


dogi:
**Review — rating: 82/100**

The SQL translation is exact, and it avoids the trap that makes this class of change dangerous. `status` is nullable, and Kotlin's `it.status != "archived"` keeps null-status rows — but a bare `status != 'archived'` in SQL evaluates to NULL for those, which is falsy, silently dropping them. `IFNULL(status, '') != 'archived'` reproduces the Kotlin across all three cases (null → kept, `'archived'` → dropped, anything else → kept). `createdDate` is a non-null `Long`, so `ORDER BY createdDate DESC` matches `sortedByDescending` too.

**The missing piece is a test of the SQL.** The new test stubs `observeNonArchivedReportsByTeamId` with a canned list and asserts the flow passes it through — that verifies delegation, not the predicate. The filter and ordering are exactly what moved into the database, so they're now the untested part. #16332 in this same queue handled the identical situation properly by exercising the real DAO against real rows (`ResourcesRepositoryLibrarySyncTest`), covering null-status, archived, and ordering. Same shape would work here and is the only way the `IFNULL` guard gets pinned.

**Two coordination notes, because this file is getting crowded.** Four open PRs now touch `EnterprisesRepositoryImpl` and its test:

- **#16363** rewrites the `distinctUntilChanged` block in this very function into `distinctByContent`, and rewrites the test's dispatcher setup to `TestDispatcherProvider(UnconfinedTestDispatcher())` — where this PR rewrites the same lines into a `mockk { every { default } … }`. That's a direct conflict in the test header.
- **#16367** also edits the test's field block (adding the injected `context`), and **#16395** adds a *duplicate* of the filter+sort you're deleting here into `exportReportsAsCsv`.

On #16395 specifically: this PR only adds the `observe*` (Flow) variant, so the export path can't reuse it as-is. A suspend sibling — `getNonArchivedReportsByTeamId(teamId): List<MyTeam>` with the same WHERE/ORDER BY — would let both call sites share one definition instead of #16395 reintroducing the Kotlin copy.

**Nit:** the neighbouring `getEligibleNextLeaderCandidates` expresses the same idea as `(status IS NULL OR status != 'archived')`. Both are correct; picking one idiom for the file would help the next reader.

**Verdict:** mergeable — add the Room-level test, and settle the merge order with #16363.

---

Okuro3499 (approved):

PR #16398, opened 2026-08-27
sync: smoother url utils base64 handling (fixes #16372)

### LINKED ISSUE #16372: Replace android.util.Base64 in UrlUtils with the JDK encoder
`basicAuthHeader` uses `android.util.Base64.encodeToString(..., NO_WRAP)`. Switching to `java.util.Base64` leaves the auth helper free of `android.util.*`.

**Ask:** replace `android.util.Base64` with `java.util.Base64.getEncoder()`. **Careful with padding:** `NO_WRAP` still includes `=` padding, unlike `withoutPadding()` — use the padded encoder (`Base64.getEncoder()`, not `.withoutPadding()`) or the header value changes and breaks auth.

**Files:** `utils/UrlUtils.kt`, `test/.../UrlUtilsTest.kt`

**Verification:** confirmed the Base64 call at `UrlUtils.kt:37`.

dogi:
Replaced `android.util.Base64` with `java.util.Base64.getEncoder()` in `UrlUtils.kt` so that `basicAuthHeader` is independent of `android.util.*`. Preserved padded Base64 encoding behavior and added unit test coverage in `UrlUtilsTest.kt`.

Fixes #16372

---
*PR created automatically by Jules for task [4037358810046416553](https://jules.google.com/task/4037358810046416553) started by @dogi*


dogi:
**Review — rating: 89/100**

The two encoders produce identical output here, which is the whole question. `android.util.Base64.encodeToString(bytes, NO_WRAP)` is the standard RFC 4648 alphabet, padded (no `NO_PADDING` flag), with line wrapping suppressed. `java.util.Base64.getEncoder()` is the Basic encoder: same alphabet, padded, never wrapped. So the emitted `Authorization` header is byte-identical — which matters, because a silently unpadded or wrapped credential would fail auth against CouchDB in a way that looks like a wrong password.

**`java.util.Base64` is API 26+, and this project's minSdk is exactly 26** — so it's available on every supported device, with no desugaring needed. Worth stating explicitly since that's the one thing that would make this unshippable.

The test is well chosen: it pins both a length that needs no padding (`user:pass` → `dXNlcjpwYXNz`) and one that needs two padding chars (`user:password` → `dXNlcjpwYXNzd29yZA==`). Padding is precisely where the two APIs could have diverged, so those two cases cover the real risk rather than just asserting the function returns something.

This is the same "get the platform type out of a util" theme as #16353 and #16385, done without their side effects — no behaviour change, nothing silently swallowed, no synchronization dropped, and a test that would catch a regression.

**Nit:** `import java.util.Base64` is inserted between `android.util.Log` and `androidx.annotation.VisibleForTesting`, so it sits above the `androidx.*` group while `java.net.URLEncoder` is further down with the other `java.*` imports. Moving it next to `URLEncoder` restores the grouping the rest of the file follows.

**Heads-up:** #16323 edits the `header` getter immediately above `basicAuthHeader`. Adjacent hunks, so expect a small conflict depending on merge order — #16323 is currently `change` pending the `ServerUrlMapper` fix, so this one will likely land first.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16346, opened 2026-08-27
all: less team user dao queries is more (fixes #16341)

dogi:
## Description
Removes two dead (uncalled) Room `@Query` methods that KSP still compiles and validates on every build:

- `TeamDao.observeByDocType` — `data/room/dao/TeamDao.kt`
- `UserDao.countByPlanetCode` — `data/room/dao/UserDao.kt`

Both methods had zero callers in `app/src/main` (verified via grep — matches found only at their declarations). Dead `@Query` methods are still compiled and validated by KSP on every build, so removing them shrinks the DAO surface and avoids maintaining unused SQL.

Fixes #16341

## Changes
- `app/src/main/java/org/ole/planet/myplanet/data/room/dao/TeamDao.kt`: deleted `observeByDocType`
- `app/src/main/java/org/ole/planet/myplanet/data/room/dao/UserDao.kt`: deleted `countByPlanetCode`

No behavioral change — no production or test code referenced either method.

## Verification
- `grep -rn "observeByDocType\|countByPlanetCode" app/src/main` → only declaration matches (now removed)
- `grep -rn "observeByDocType\|countByPlanetCode" app/src/test` → no matches
- `Flow` import retained in `TeamDao.kt` (still used by `observeAll` and `observeByTeamIdAndDocType`)

---
_This PR was created by an AI agent (OpenHands) on behalf of the repository contributor._

@dogi can click here to [continue refining the PR](https://app.all-hands.dev/conversations/0d3bdbaa-f07c-47cf-98e1-24c41f99f5ea)

PR #16399, opened 2026-08-27
resources: smoother repository dao querying (fixes #16371)

### LINKED ISSUE #16371: Replace two full-row loads in ResourcesRepository with column projections
`getMyLibIds` calls `myLibraryDao.getForUserPattern(...)`, which `SELECT *`s every `MyLibrary` row for the user, then emits only the ids; `getResourceTitlesMap` calls `getWithResourceId()` (`SELECT * FROM my_library WHERE resourceId IS NOT NULL`) but reads only `resourceId` and `title` — and it runs on every offline-resource render via `getOfflineResourceItems`.

**Ask:** add column-projection DAO queries for both (id-only for `getMyLibIds`; a typed `(resourceId, title)` projection for `getResourceTitlesMap`). `getForUserPattern` has other callers and must stay as-is — add a new narrower query rather than changing it.

**Files:** `data/room/dao/MyLibraryDao.kt`, `repository/ResourcesRepositoryImpl.kt`, `test/.../ResourcesRepositoryImplTest.kt`

**Verification:** both call sites confirmed.

dogi:
Replaces full-row MyLibrary SELECT * loads in ResourcesRepositoryImpl (getMyLibIds and getResourceTitlesMap) with targeted column projections (getIdsForUserPattern and getResourceTitles) to reduce memory overhead and improve performance.

Fixes #16371

---
*PR created automatically by Jules for task [17825775508381863252](https://jules.google.com/task/17825775508381863252) started by @dogi*


dogi:
**Review — rating: 87/100**

Both projections are exact narrowings — I compared the new queries against the ones they replace and the WHERE clauses are character-for-character identical:

- `getForUserPattern`: `WHERE userId LIKE :userPattern ESCAPE '\'` → `getIdsForUserPattern`, same predicate, `SELECT id`
- `getWithResourceId`: `WHERE resourceId IS NOT NULL` → `getResourceTitles`, same predicate, `SELECT resourceId, title`

Same row set, fewer columns, so there's no behavioural surface here at all. That matters because a projection PR is exactly where an accidentally-tightened predicate hides in plain sight.

The win is real rather than nominal: `my_library` rows are wide — title, description, attachment JSON, several type-converted list columns — and `getResourceTitlesMap` was materialising every row that has a `resourceId` just to keep two fields. `getMyLibIds` was doing the same to read one string per row.

`ResourceTitleProjection` as a top-level data class is the right shape for a Room POJO projection, and Room validates the column names against the query at compile time, so a mismatch fails KSP rather than surfacing at runtime — which is why I'm less worried here than I'd be about an untested *predicate* change.

**Nit — one orphan.** `getWithResourceId` had exactly one caller, the line this PR rewrites, so it has none once this lands. `getForUserPattern` is fine (still a dozen call sites). Deleting the dead query in this same PR would close the loop, same as #16347 and #16346 are doing elsewhere in the queue — Room generates an implementation for every `@Query`, so a dead one isn't free.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16422, opened 2026-08-27
teams: smoother members view binding (fixes #16416)

### LINKED ISSUE #16416: Compute the member display name once per row in MembersAdapter
**rating:** 63/100 &nbsp;·&nbsp; **proposed by:** deepseek(r2), glm(r2)

`onBindViewHolder` binds `if (member.toString() == " ") member.name else member.toString()` — calling UserEntity.toString() twice per row, allocating a fresh String each time, on every bind during scroll.

**Ask:** compute `member.toString()` once into a local variable and reuse it for both the comparison and the fallback branch.

**Files:** `ui/teams/members/MembersAdapter.kt`

**Verification:** Verified the double toString(). deepseek also claims getRoleAsString() is called repeatedly — it is called once; that half of its context is wrong.

dogi:
## Description

`MembersAdapter.onBindViewHolder` bound the member display name with `if (member.toString() == " ") member.name else member.toString()` — calling `UserEntity.toString()` twice per row on every bind during scroll.

As the review on this PR pointed out, `UserEntity.toString()` is just `return "$name"`, so `member.toString()` *is* `member.name`. That makes the conditional a no-op — both branches yield the same string — so the comparison it was hoisting could never discriminate. The strictly simpler fix is to bind `member.name` directly and delete the dead branch entirely (one call, no branch), rather than hoisting a comparison that cannot succeed.

A side note raised in review (pre-existing, not introduced here): when `name` is null, `"$name"` produces the literal `"null"`. Binding `member.name` directly instead of the string template renders empty text for a null name rather than the string `"null"`.

`getRoleAsString()` was already called once and is unchanged.

Fixes #16416

## Changes

- `app/src/main/java/org/ole/planet/myplanet/ui/teams/members/MembersAdapter.kt`: bind `tvTitle` to `member.name` directly; drop the redundant `member.toString()` conditional and its local.
- `app/src/test/java/org/ole/planet/myplanet/ui/teams/members/MembersAdapterTest.kt`: remove the whitespace-name fallback test, which encoded a distinction between two branches that produce the same value; the normal display-name binding test is retained.

## Verification

Tests follow the existing patterns in `MembersAdapterTest.kt` (Robolectric + `ApplicationProvider`, `submitList { ... }` + `onBindViewHolder`). Java/Android SDK is not available in this authoring environment, so CI (`test.yml`) will run `testDefaultDebugUnitTest` to verify.

---


dogi:
**Review — rating: 78/100**

The hoist itself is correct — one `toString()` per bind instead of two on the common path, on a scrolling list.

**But the conditional it's optimizing has no effect.** `UserEntity.toString()` is:

```kotlin
override fun toString(): String {
    return "$name"
}
```

So `member.toString()` *is* `member.name`, and the whole expression reduces to "if name is `" "`, use name; otherwise use name". Both branches yield the same string. The line can simply be:

```kotlin
binding.tvTitle.text = member.name
```

That's one call instead of two and removes the branch entirely — strictly better than hoisting a comparison that can't discriminate.

**This also explains why the new fallback test can't fail.** `testOnBindViewHolder_bindsWhitespaceNameFallbackToName` sets `name = " "`, so `toString()` returns `" "`, the guard matches, and the fallback yields `member.name` — which is also `" "`. The assertion passes on either branch. That isn't a fixable test setup: since `toString()` returns `name`, no input can make the two branches differ, so the test encodes a distinction that doesn't exist and will outlive the code it describes. The other test is fine — `"Alice Example"` genuinely exercises the else path.

Worth knowing while you're in here: when `name` is null, `"$name"` produces the literal string `"null"`, which fails the `== " "` check and renders **"null"** in `tvTitle`. Pre-existing, not introduced here, but collapsing to `member.name` (with an `.orEmpty()` or a real placeholder) would fix that at the same time.

**Verdict:** needs another round — delete the conditional rather than optimizing it, and drop the fallback test with it.

---

dogi:

---

dogi:
Followed up on @dogi's review. Pushed commit `bfeba0c9c026c3396505ffa56c3ddc9207d217fe` (and retitled the PR).

You were right that the hoisted conditional was optimizing something that can't discriminate: `UserEntity.toString()` is `return "$name"`, so `member.toString()` *is* `member.name`, and both branches of `if (memberDisplayName == " ") member.name else memberDisplayName` yield the same string. So instead of hoisting the comparison, I deleted it — `tvTitle` now binds `member.name` directly (one call, no branch), which is strictly better than the hoist.

Per your note, I also dropped `testOnBindViewHolder_bindsWhitespaceNameFallbackToName` — it asserted a distinction between two branches that produce the same value, so it couldn't fail and would outlive the code it described. The normal display-name binding test (`testOnBindViewHolder_bindsMemberDisplayName`) is kept.

On the null-name side note (pre-existing, not introduced here): with the template gone, a null `name` now renders empty text rather than the literal `"null"` — a side improvement, not a regression. I left it at `member.name` rather than adding `.orEmpty()` to keep this PR to the issue's scope; happy to add an explicit placeholder in a follow-up if you'd prefer one.

I didn't find any inline review threads from @Okuro3499 on this PR, so I acted on your review — point me at any further instructions and I'll follow up.

---
_This message was created by an AI agent (OpenHands) on behalf of the repository contributor._

dogi:
## Review: 85/100 — mergeable

The conditional really was dead, and removing it quietly fixes a small display bug.

### Why the old branch was a tautology

`UserEntity.toString()` is (`UserEntity.kt:183`):

```kotlin
override fun toString(): String {
    return "$name"
}
```

So `member.toString() == " "` is true only when `name` is the single-character string `" "` — and in that case the "fallback" `member.name` returns `" "` as well. **Both branches produced the same value.** Nothing was ever chosen by that check.

There's a fingerprint of what it was *meant* to do, though: `" "` is exactly what `"$firstName $lastName"` renders when both parts are blank. `getFullName()` (:150) still has that shape. So this check almost certainly dates from when `toString()` returned the full name, and it was left behind when `toString()` was narrowed to `"$name"` — a vestige rather than a mistake.

### It also fixes the null case

The one place the two expressions differ is when `name` is null:

- `master`: `member.toString()` → `"null"`, so the row literally displays the text **null**.
- this branch: `member.name` → `null`, and `TextView.setText(null)` renders empty.

That's a real (if cosmetic) improvement for members whose `name` never came down from the server, and it's the strongest reason to take this beyond tidiness. Worth putting in the description.

### The new test is a guard, not a regression test

`name = "Alice Example"` takes the `else` branch on `master` too (`"Alice Example" != " "`), so `testOnBindViewHolder_bindsMemberDisplayName` passes on both sides. Fine to keep, but the case that would actually **fail on `master`** is the one the change fixes:

```kotlin
val user = UserEntity(id = "user1", name = null)
// master renders "null"; this branch renders ""
assertEquals("", viewHolder.binding.tvTitle.text.toString())
```

Adding that turns the diff from "trust me, the branches were equal" into something CI enforces.

### Observation, explicitly out of scope

The row shows `member.name` — the account name — while `UserEntity` also offers `getFullName()` / `getFullNameWithMiddleName()`. Given the `" "` sentinel suggests this row once displayed a full name, it's worth a separate issue to decide which the team member list *should* show (full name with a fallback to `name` when both parts are blank). Not something to fold into this PR.

**Verdict:** mergeable.

---

Okuro3499 (dismissed):
```
File: app/src/main/java/org/ole/planet/myplanet/ui/teams/members/MembersAdapter.kt:106                                                                                           
Problem: member.toString() returns "$name", which for name == null evaluates to the literal string "null", not Kotlin null. So the old code's fallback branch (memberDisplayName 
== " " never true in practice) was actually always taking the else branch, binding "null" (4 visible characters) to tvTitle.text whenever a member had a null name. The new code 
binds member.name directly, i.e. Kotlin null, which clears/blanks the TextView instead. This is a real, if narrow, behavior change the commit message doesn't acknowledge ("both 
branches yield the same string" is false exactly in this case). It's very likely an improvement (blank is better than showing "null" for a user with no name), but it needs to be
called out as an intentional change, not asserted as a no-op refactor — and needs a regression test for that path, since it's the one case the diff actually alters.             
```

olevim:

olevim:
Followed up on @dogi's 85/100 review (no inline threads from @Okuro3499 exist on this PR).

## What the review said
@dogi noted the kept `testOnBindViewHolder_bindsMemberDisplayName` (name `"Alice Example"`) passes on both master and this branch, so it doesn't enforce the diff — the case that would actually **fail on master** is the one the change fixes: a null name renders the literal `"null"` via the old `"$name"` `toString()` template, but renders empty here via the direct `member.name` binding.

## Change (commit `a5b5811`)
- **`MembersAdapterTest.kt`**: added `testOnBindViewHolder_nullNameRendersEmpty` — a `UserEntity` with `name = null`, asserting `tvTitle` renders `""`. This fails on master (renders `"null"`) and passes on this branch, turning the behavior change from trust-me into something CI enforces.

No production code changed — the adapter already binds `member.name` directly. CI (`test.yml`) will run `testDefaultDebugUnitTest` to verify the new assertion.

dogi:
## Review — rating 87/100 (was 85)

The production change was already right; the added tests pin the part that matters.

`UserEntity.toString()` is `"$name"`, so the old condition compared a string template against `" "`:

```kotlin
binding.tvTitle.text = if (member.toString() == " ") member.name else member.toString()
```

That branch could only be taken when `name` was literally `" "` — and in that case both arms produce the same `" "`, so it never changed anything. Meanwhile the `else` arm was the actual bug: for a null name, `"$name"` renders the literal string **"null"** in the members list. Assigning `member.name` directly hands `setText` a null `CharSequence`, which renders empty.

`testOnBindViewHolder_nullNameRendersEmpty` is the test that earns its place — it fails on master, where the cell reads `"null"`, and the inline comment explains precisely why. `bindsMemberDisplayName` passes on master too, but it is a reasonable baseline to keep alongside it.

### Two notes

`#16529` is labelled `merge` and also touches `MembersAdapterTest.kt`. Whichever lands second will need a rebase — worth watching since both are queued.

Non-blocking, and a property of the whole file rather than this PR: assertions live inside the `submitList` commit callback.

```kotlin
adapter.submitList(list) {
    ...
    assertEquals("Alice Example", viewHolder.binding.tvTitle.text.toString())
}
```

`AsyncListDiffer` dispatches synchronously when the previous list is empty, so the callback does run here and the assertions execute. But if that precondition ever stopped holding, the callback would not fire and the test would pass with zero assertions rather than fail. Since every test in this file is written this way, it is not this PR's problem — but the file would be sturdier asserting after the callback rather than inside it.

Moving to `ready`.

---

Okuro3499 (approved):

PR #16423, opened 2026-08-27
all: smoother download url utils caching (fixes #16411)

### LINKED ISSUE #16411: Build the download URL list against one resolved base URL
**rating:** 65/100 &nbsp;·&nbsp; **proposed by:** codex(r3), glm(r3)

`DownloadUtils.downloadAllFiles` maps each library through `UrlUtils.getUrl(it)`, and each call re-derives `getUrl()` -> `dbUrl(spm())` -> spm() + baseUrl() string work for a base that is constant across the list; the result is then copied again into an ArrayList.

**Ask:** resolve the base URL once before the loop and reuse it for every item, instead of re-deriving `spm()`/`baseUrl()` per library. Avoid the redundant extra `ArrayList` copy if straightforward.

**Files:** `utils/DownloadUtils.kt`, `utils/UrlUtils.kt`

**Verification:** Verified the map+ArrayList shape. glm's version hoists the base URL (the real win); codex's only pre-sizes the ArrayList.

dogi:
## Summary

`DownloadUtils.downloadAllFiles` mapped each library through `UrlUtils.getUrl(it)`, and every call re-derived the constant base URL via `getUrl()` -> `dbUrl(spm())` -> `baseUrl(spm())` + `spm()` string work. The result was then copied into a second `ArrayList`.

This resolves the base URL **once** before the loop and reuses it for every item.

## Changes

- `UrlUtils.getUrl(id, file, base)` — new overload that builds `/resources/$id/$file` against a caller-supplied base URL, so the base is not re-derived per call. The existing `getUrl(id, file)` now delegates to it.
- `DownloadUtils.downloadAllFiles` — resolves `UrlUtils.getUrl()` a single time, then maps each library through the new overload via `mapTo(ArrayList())`, which builds the result `ArrayList` in one pass and drops the redundant intermediate `List` + `ArrayList` copy.

## Verification

- Added `UrlUtilsTest` coverage for the new `getUrl(id, file, base)` overload and that the base is resolved once across many libraries.
- Added `DownloadUtilsTest` coverage for `downloadAllFiles`: URLs are built against one resolved base, `spm.getCouchdbUrl()` is called exactly once regardless of list size, and the empty-input case still resolves the base once.

> Note: the runtime has no JDK/Android SDK available, so the tests could not be executed locally; they will run in CI.

Fixes #16411.

---

@dogi can click here to [continue refining the PR](https://app.all-hands.dev/conversations/028966a0-b90a-4ab6-bdc2-d7ac010dbc77)


dogi:
**Review — rating: 88/100**

Equivalence checks out exactly. The old `getUrl(library)` overload is `getUrl(library?.resourceId, library?.resourceLocalAddress)`, which expands to `"${getUrl()}/resources/$id/$file"` — so passing the same two fields plus a pre-resolved `base` produces an identical string, null handling included. The `getUrl(MyLibrary?)` overload still has eight other call sites, so nothing is orphaned.

The win is real: `getUrl()` walks `SharedPrefManager` (`isAlternativeUrl`, `getCouchdbUrl`, …) on every call, and that was happening once per library rather than once per download batch. `mapTo(ArrayList())` also drops the intermediate list that `ArrayList(list.map { })` was allocating.

**The test earns particular credit**: `verify(exactly = 1) { spm.getCouchdbUrl() }` pins the hoisting itself, so moving base resolution back into the loop fails the suite. That's a genuine regression test for a performance change — several test PRs currently in this queue assert only that the right values came out, which passes whether or not the optimization survives.

**Concrete follow-up:** `CourseStepFragment.kt:223` still has the identical pattern —

```kotlin
val urls = ArrayList(notDownloaded.map { UrlUtils.getUrl(it) })
```

— re-deriving the base once per item. The new three-arg overload applies verbatim there, and it's the same download path.

**Nits:** the two `downloadAllFiles` tests repeat the whole `unmockkObject` / `resetForTesting` / spm-stub / `init` block; a `@Before` or a small helper would DRY it. Also worth noting the empty-list case now resolves the base once where the old code resolved it zero times — harmless, and the test documents it.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16426, opened 2026-08-27
sync: smoother submissions repository photo dao uploading (fixes #16412)

### LINKED ISSUE #16412: Batch the photo uploaded-mark instead of one UPDATE per photo
**rating:** 64/100 &nbsp;·&nbsp; **proposed by:** deepseek(r2)

`PhotoUploader.uploadSubmitPhotos` calls `submissionsRepository.markPhotoUploaded(photoId, rev, id)` per successful photo, each issuing its own UPDATE — a 100-photo roll becomes 100 separate statements on the upload path.

**Ask:** add a batched variant that marks multiple photos uploaded in one call/transaction, and use it from `uploadSubmitPhotos` instead of one call per photo.

**Files:** `services/upload/PhotoUploader.kt`, `repository/SubmissionsRepositoryImpl.kt`

**Verification:** File and per-item mark verified; I did not measure the statement count.

dogi:
## Summary

`PhotoUploader.uploadSubmitPhotos` previously called `submissionsRepository.markPhotoUploaded(photoId, rev, id)` per successful photo, each issuing its own `UPDATE` — a 100-photo roll became 100 separate statements on the upload path.

This adds a batched variant that marks multiple photos uploaded in a single Room transaction and uses it from `uploadSubmitPhotos`, accumulating successful uploads per batch and marking them in one call instead of one call per photo.

## Changes

- `data/room/dao/SubmitPhotosDao.kt`
  - Added `markUploadedBatch(uploads)`, a `@Transaction`-annotated default method that runs the existing `markUploaded` for each photo inside one transaction.
  - Added an `UploadedPhoto` data class (local id, `_rev`, `_id`) to carry the per-photo result.
- `repository/SubmissionsRepository.kt` / `SubmissionsRepositoryImpl.kt`
  - Added `markPhotosUploadedBatch(uploads: List<UploadedPhoto>)` to the interface and implementation (no-op on empty input).
- `services/upload/PhotoUploader.kt`
  - Accumulate successful uploads per batch and call `markPhotosUploadedBatch` once per batch instead of `markPhotoUploaded` per photo.
  - Preserved the previous `photoId != null` guard so null local ids are skipped.

## Verification

- `services/upload/PhotoUploader.kt` — `markPhotosUploadedBatch` is the only mark call site in `uploadSubmitPhotos`; per-item `markPhotoUploaded` is no longer called there.
- `repository/SubmissionsRepositoryImpl.kt` — `markPhotosUploadedBatch` delegates to `submitPhotosDao.markUploadedBatch` in one call.

## Tests

- `UploadManagerTest` — updated `uploadSubmitPhotos uploads photos successfully` to assert the batched call is invoked with the right `UploadedPhoto` and that per-item `markPhotoUploaded` is not called; added `uploadSubmitPhotos batches photo mark calls once per batch instead of per photo` asserting a single `markPhotosUploadedBatch` call (not N) for a multi-photo batch.
- `SubmissionsRepositoryImplTest` — added `markPhotoUploaded delegates single photo to dao`, `markPhotoUploaded ignores null photo id`, `markPhotosUploadedBatch delegates batch to dao in one call`, and `markPhotosUploadedBatch does not call dao for empty batch`.

Fixes #16412.

---

@dogi can click here to [continue refining the PR](https://app.all-hands.dev/conversations/8846e4ae-e1aa-4f14-a4ef-abc705a6277b)


dogi:
**Review — rating: 84/100**

The core win is real and correctly implemented: `@Transaction` on a default interface method wraps the whole `forEach` in one SQLite transaction, so N photo marks cost one commit instead of N. That's a genuine I/O saving on a path that runs after every photo batch, not a micro-optimization.

Moving `successfulUploads.add(...)` out from under the `listener != null` guard is necessary rather than incidental — the list now drives the marking, so it has to be populated regardless of whether a listener is attached. Easy to get wrong; it's right here.

**The tests pin the actual claim.** `uploadSubmitPhotos batches photo mark calls once per batch` asserts `markPhotosUploadedBatch` is called `exactly = 1` for three photos *and* `markPhotoUploaded` `exactly = 0` — that fails if anyone reintroduces per-photo marking, which is exactly the regression this PR exists to prevent. The empty-batch short circuit is covered too.

**One trade-off worth being deliberate about: durability moved.** Previously each photo was marked the moment its own upload returned, so an exception escaping the loop or the process being killed mid-batch still left the already-uploaded photos correctly marked. Now nothing is persisted until the whole loop finishes. If the batch is interrupted after several successful POSTs, those photos are uploaded server-side but unmarked locally, and the next run re-POSTs them — creating duplicate remote documents rather than merely redoing work.

The window is bounded by the batch duration, and atomicity has its own value, so I don't think this blocks. But on an offline-first app where a long photo batch over a flaky connection is a normal case, it's worth a deliberate decision rather than a side effect. If duplicates would be a problem, chunking (mark every N photos) keeps most of the transaction win with a much smaller loss window.

**Nit — layering:** `UploadedPhoto` is nested inside `SubmitPhotosDao`, so `services/upload/PhotoUploader` and the `SubmissionsRepository` interface both import a type from `data/room/dao`. A DAO exposing a DTO that the service layer constructs points the dependency the wrong way; `model/` would be the natural home.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16427, opened 2026-08-27
teams: smoother events repository meetup dao querying (fixes #16421)

### LINKED ISSUE #16421: Return meetup member IDs directly from Room
**rating:** 61/100 &nbsp;·&nbsp; **proposed by:** codex(r1)

`EventsRepositoryImpl.getJoinedMembers` loads complete Meetup rows via `getMembersByMeetupId` only to `mapNotNull { it.userId }.distinct()` — a SELECT DISTINCT userId projection gives the same thing.

**Ask:** add a DAO query that selects distinct `userId` directly, and use it instead of loading full rows and reducing in Kotlin. Preserve the existing 400-id chunking and final `distinctBy` — the task explicitly requires keeping those.

**Files:** `repository/EventsRepositoryImpl.kt`, `data/room/dao/MeetupDao.kt`

**Verification:** Verified at EventsRepositoryImpl.kt:66-79; the existing 400-id chunking and final distinctBy must be preserved as the task says.

dogi:
## Summary

`EventsRepositoryImpl.getJoinedMembers` previously loaded complete `Meetup` rows via `getMembersByMeetupId` only to `mapNotNull { it.userId }.distinct()`. This adds a DAO query that selects `DISTINCT userId` directly from SQL and uses it instead of loading full rows and reducing them in Kotlin.

The existing 400-id chunking (`memberIds.chunked(400)`) and final `distinctBy { it.id }` are preserved, as required by the issue.

## Changes

- `data/room/dao/MeetupDao.kt`: add `getMemberUserIdsByMeetupId(meetupId)` — `SELECT DISTINCT userId FROM meetup WHERE meetupId = :meetupId AND userId IS NOT NULL AND userId != ''`.
- `repository/EventsRepositoryImpl.kt`: `getJoinedMembers` now calls the new DAO query instead of `getMembersByMeetupId`.
- `test/.../EventsRepositoryImplTest.kt`: update `getJoinedMembers` test to mock the new DAO method.

## Verification

`./gradlew testDefaultDebugUnitTest --tests "org.ole.planet.myplanet.repository.EventsRepositoryImplTest"` — all 10 tests pass, including `getJoinedMembers`.

Fixes #16421

@dogi can click here to [continue refining the PR](https://app.all-hands.dev/conversations/2d046aa7-95bb-4350-89d9-73ac7996ed51)


dogi:
**Review — rating: 87/100**

Exact narrowing: the new query keeps the same three predicates (`meetupId`, `userId IS NOT NULL`, `userId != ''`) and projects one column, with `DISTINCT` taking over from the Kotlin `.distinct()`. `Meetup` rows carry title, description, dates and type-converted list columns, all of which were being materialised to read a single string — and `DISTINCT` also cuts the rows transferred when a user appears several times.

**The detail I'd have expected to get lost is correctly kept:** `.takeUnless { it.isBlank() }` stays. SQL's `userId != ''` rejects only the empty string, so a whitespace-only id (`" "`) would slip through the query and still needs the Kotlin guard. Deleting it as "now redundant" would have been the easy mistake, and it wasn't made.

Ordering is a non-issue here, worth noting since it usually isn't: `getJoinedMembers` returns whatever `getUsersByAnyIds` yields, so the result order was already database-determined rather than derived from `memberIds`. Nothing changes.

**Nit — one orphan.** After this lands, `getMembersByMeetupId` has zero callers: the only two references are the production line and the test stub that this diff rewrites. Worth deleting in the same PR, as #16347 and #16399 are doing for their own superseded queries.

**Nit:** the updated test stubs `getMemberUserIdsByMeetupId` with `listOf("user1", "user2", "user1")` — a duplicate the `SELECT DISTINCT` can't actually produce. Harmless, and it does exercise the retained Kotlin dedup, but the fixture no longer reflects what the query can return. A whitespace-only id (`" "`) would be the more valuable case to stub, since that's the one the SQL genuinely can emit and the Kotlin filter exists to catch.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16428, opened 2026-08-27
courses: smoother submission view modelling (fixes #16417)

### LINKED ISSUE #16417: Do the submission newest-per-parent selection in one pass
**rating:** 62/100 &nbsp;·&nbsp; **proposed by:** codex(r2), jules(r1)

`SubmissionViewModel.filteredSubmissionsRaw` groups every filtered submission into lists, scans each group for its max, then re-groups a second time to build submissionCountMap — three passes over the same data in ViewModel state production.

**Ask:** compute the newest-per-parent selection and the count map in a single pass over the grouped data instead of three. Keep the existing tie-breaking rule for equal `lastUpdateTime` values exactly as-is.

**Files:** `ui/submissions/SubmissionViewModel.kt`

**Verification:** Verified at SubmissionViewModel.kt:68-88. codex keeps the work in the ViewModel and is precise about the equal-lastUpdateTime tie rule; jules instead proposes pushing the counts into a SubmissionDao GROUP BY, which is a bigger change for the same symptom.

dogi:
## Summary

`SubmissionViewModel.filteredSubmissionsRaw` previously computed the newest-per-parent selection and the count map across three passes over the grouped data:

1. `filtered.groupBy { it.parentId }` — grouped the submissions
2. `mapValues { it.value.maxByOrNull { it.lastUpdateTime } }` — scanned each group for its max to build `uniqueRawSubmissions`
3. `mapValues { it.value.size }.mapKeys { ... maxByOrNull ... }` — re-grouped a second time to build `submissionCountMap`, calling `maxByOrNull` once more per group

This refactors that into a **single pass** over the grouped data: one loop over the groups that simultaneously selects the newest submission (via the same `group.maxByOrNull { it.lastUpdateTime }` call) and records its count keyed by the newest submission's id.

## Behavior preserved

The existing tie-breaking rule for equal `lastUpdateTime` values is kept exactly as-is — the same `maxByOrNull { it.lastUpdateTime }` is applied per group (over the descending-sorted filtered list), so the first-encountered submission wins on ties, matching the prior behavior.

## Test

Added `testTieBreakingForEqualLastUpdateTime` to `SubmissionViewModelTest`, which places two submissions with identical `lastUpdateTime` under the same parent and asserts the first-encountered one is selected and that its count is correct. Existing tests (`testFilterModes`, `testQueryFilteringAndGrouping`, `testSubmissionCountsMapTracksLatestSubmission`, `testDistinctEmissions`) continue to pass.

All 5 tests in `SubmissionViewModelTest` pass.

## Verification

Confirmed at `SubmissionViewModel.kt:68-88` per the issue.

Fixes #16417

@dogi can click here to [continue refining the PR](https://app.all-hands.dev/conversations/df4f9f16-16fb-4fba-a833-439a6c82af53)


dogi:
**Review — rating: 88/100**

The old shape called `maxByOrNull` **twice per group** — once to build `uniqueRawSubmissions`, then again inside `mapKeys` to recover the same newest submission's id for the count map — plus `mapValues`/`mapKeys` allocating two more intermediate maps. One loop computing `newest` once and writing both outputs is a straight halving of that scan.

Equivalence holds where it matters:

- **Tie-breaking is preserved.** Both versions use `maxByOrNull`, which returns the *first* maximal element, so equal `lastUpdateTime` values resolve identically. That's the subtle property a rewrite like this most easily breaks (switching to a sort, or `maxWith`, would silently change which submission survives), and the new test pins it directly — two submissions sharing the newest timestamp under one parent, asserting `"1"` is kept with a count of 2.
- **The count map keeps its shape.** Key is still the newest submission's id, value still the group size, and the `HashMap<String?, Int>` key type matches what `mapKeys` produced before, so the `Triple` contract downstream is unchanged.
- `?: continue` replaces `filterNotNull()`. `maxByOrNull` only returns null for an empty group and `groupBy` never yields one, so both are unreachable — though the new form is marginally tighter, since the old `mapKeys` could in principle have emitted a `null`-keyed entry where this can't.

**Nit:** the test comment says the tie-breaker keeps the first entry "in the (descending-sorted) list", but at that point `filtered` isn't sorted — the `sortedByDescending` happens later, on `uniqueSubmissions`. What actually decides it is `groupBy` preserving encounter order plus `maxByOrNull` returning the first max. The assertion is right; the explanation would mislead whoever reads it next.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16429, opened 2026-08-27
courses: smoother ratings repository dao aggregating (fixes #16420)

### LINKED ISSUE #16420: Query rating summaries without loading every rating row
**rating:** 61/100 &nbsp;·&nbsp; **proposed by:** codex(r1)

`RatingsRepositoryImpl.getRatingSummary` loads all matching entities via `getByTypeAndItem` and computes count and average in Kotlin (`results.map { it.rate }.average()`), when only one user row plus two aggregates are needed.

**Ask:** add a Room `@Query` that computes `COUNT`/`AVG` directly (plus the one user-specific row still needed), and use it instead of loading and reducing every row in Kotlin.

**Files:** `repository/RatingsRepositoryImpl.kt`, `data/room/dao/RatingDao.kt`

**Verification:** Verified at RatingsRepositoryImpl.kt:44-60.

dogi:
## Summary

`RatingsRepositoryImpl.getRatingSummary` previously loaded every matching rating row via `RatingDao.getByTypeAndItem` and computed the count and average in Kotlin (`results.map { it.rate }.average()`), when only one aggregate row plus at most one user-specific row are needed.

This adds a Room `@Query` that computes `COUNT(*)` and `AVG(rate)` directly in SQLite, and reuses the existing `findByTypeUserItem` for the single user row.

## Changes

- **`data/room/dao/RatingDao.kt`**
  - Added a `RatingAggregate(totalCount: Int, averageRate: Double?)` projection data class.
  - Added `getAggregate(type, item)` — a `@Query` computing `COUNT(*) AS totalCount, AVG(rate) AS averageRate` in SQLite.
  - Removed the now-unused `getByTypeAndItem` (only `getRatingSummary` consumed it).
- **`repository/RatingsRepositoryImpl.kt`**
  - `getRatingSummary` now reads `ratingDao.getAggregate(...)` for the count/average and `ratingDao.findByTypeUserItem(...)` for the one user row, instead of loading and reducing every row in Kotlin.
  - A null `userId` skips the user-row lookup entirely.
- **`app/src/test/.../RatingsRepositoryImplTest.kt`**
  - Updated existing tests to mock `getAggregate` + `findByTypeUserItem` instead of `getByTypeAndItem`.
  - Added `getRatingSummary with null userId omits user lookup` covering the null-`userId` branch and verifying no user lookup is made.

## Verification

`./gradlew testDefaultDebugUnitTest --tests "org.ole.planet.myplanet.repository.RatingsRepositoryImplTest"` — all 12 tests pass; KSP compiled the new `@Query` (validating the SQL and projection column mapping).

## Performance impact

A summary request previously read N rating rows (full entity blobs, including the embedded `user` JSON) and reduced them in Kotlin; it now reads one aggregate row (two scalars) plus at most one user row — O(N) row loads → O(1).

---

Fixes #16420

@dogi can click here to [continue refining the PR](https://app.all-hands.dev/conversations/d50bd4f4-7f97-4401-a123-e3444d6cb1af)


dogi:
**Review — rating: 89/100**

Replacing "load every rating row for this item, then reduce in Kotlin" with `COUNT(*)`/`AVG(rate)` plus one indexed lookup for the user's own rating is the right trade — two narrow queries instead of materialising N full rows on a screen that renders a single summary.

I checked the way this could silently diverge, and it doesn't: SQLite's `AVG` skips NULLs while `COUNT(*)` counts them, so a nullable `rate` column would have made the average and the count disagree. `Rating.rate` is declared `var rate = 0` — a non-null `Int`, hence a NOT NULL column — so both aggregate over exactly the same rows. The empty case also lines up: `AVG` over zero rows is NULL and `COUNT(*)` is 0, which `?: 0f` maps to the old `if (totalRatings > 0) … else 0f`. That path is covered by the `RatingAggregate(0, null)` test.

Worth crediting: `getByTypeAndItem` is **deleted** rather than left behind. Several PRs in this queue supersede a DAO query and strand it; this one removes it, and the green build confirms nothing else called it.

The test updates show the new call pattern was actually reasoned about rather than mechanically patched. `getRatingSummary with null userId omits user lookup` asserts `coVerify(exactly = 0) { findByTypeUserItem(...) }`, pinning the skip; and `submitRating inserts new rating if not exists` switches to `returnsMany listOf(null, savedRating)` because `findByTypeUserItem` is now hit twice — once for the existence check, once building the summary. That's easy to get wrong by stubbing a single return and quietly asserting the wrong branch.

**Nit:** the two queries express the same logical filter with different operators — `getAggregate` uses `type IS :type AND item IS :item`, `findByTypeUserItem` uses `=`. Both callers pass non-null `type`/`itemId`, so behaviour is identical today, but the inconsistency invites someone to widen one signature to nullable later and get NULL-matching in one query and not the other.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16436, opened 2026-08-27
life: smoother personals repository view modelling (fixes #16432)

### LINKED ISSUE #16432: Replace the personals mutation lambda with an explicit repository intent
**rating:** 60/100 &nbsp;·&nbsp; **proposed by:** opus(r1)

`PersonalsRepository.updatePersonalResource(id, updater: (Personal) -> Unit)` hands a mutable Room entity out to the UI to be modified in place — PersonalsFragment:125 passes a lambda setting description and title, threaded through PersonalsViewModel:56. A data-layer write expressed as a UI callback: the repository cannot validate it, log it, or be reimplemented off-JVM.

**Ask:** replace the `updater: (Personal) -> Unit` lambda parameter with an explicit intent (e.g. named parameters or a small update-data-class for the fields that can change), removing the UI's ability to mutate the Room entity directly.

**Files:** `repository/PersonalsRepository.kt`, `repository/PersonalsRepositoryImpl.kt`, `ui/personals/PersonalsViewModel.kt`, `ui/personals/PersonalsFragment.kt`, `test/.../PersonalsRepositoryImplTest.kt`

**Verification:** Verified the whole lambda chain including the fragment call site.

dogi:
## Summary

Replaces the `updater: (Personal) -> Unit` lambda in `PersonalsRepository.updatePersonalResource` with an explicit `PersonalUpdate` data class, removing the UI's ability to mutate the Room entity directly.

Previously `PersonalsRepository.updatePersonalResource(id, updater)` handed a mutable Room entity out to the UI (`PersonalsFragment:125` passed a lambda that set `description` and `title`, threaded through `PersonalsViewModel:56`). A data-layer write expressed as a UI callback cannot be validated, logged, or reimplemented off-JVM. The repository now accepts a `PersonalUpdate(title, description)` intent where `null` means "leave the stored value unchanged", and applies the fields itself inside the repository layer.

## Changes

- `repository/PersonalsRepository.kt` — new `PersonalUpdate` data class; `updatePersonalResource(id, update: PersonalUpdate)` signature.
- `repository/PersonalsRepositoryImpl.kt` — applies `update.title` / `update.description` to the loaded entity, skipping `null` fields.
- `ui/personals/PersonalsViewModel.kt` — forwards `PersonalUpdate` instead of a `(Personal) -> Unit` lambda.
- `ui/personals/PersonalsFragment.kt` — constructs `PersonalUpdate(title = title, description = desc)` at the call site.
- `test/.../PersonalsRepositoryImplTest.kt` — updated the existing `updatePersonalResource` tests to the new signature and added coverage for the "leave unchanged fields alone when null" and "entity not found" paths.

## Verification

- Updated the whole lambda chain including the fragment call site.
- Existing `PersonalsRepositoryImplTest` tests rewritten for the explicit intent; added two new test cases covering partial updates and missing entities.
- Version bumped to `0.67.65` (`versionCode 6765`).

Closes #16432.

@dogi can click here to [continue refining the PR](https://app.all-hands.dev/conversations/587f8a39-b404-4c86-8326-b2271c445ed8)


dogi:
**Review — rating: 88/100**

This fixes a real layering problem rather than just reshaping an API. The old `updatePersonalResource(id, updater: (Personal) -> Unit)` handed a live Room `@Entity` to the UI and let `PersonalsFragment` mutate it in place — the fragment was writing directly to a persistence object, and the repository had no idea which fields it was allowed to change. `PersonalUpdate` makes the editable surface explicit and keeps `Personal` inside the repository, which is what the KDoc says and what the diff actually achieves.

Behaviour is preserved at the one call site: the fragment set both `title` and `description`, and `PersonalUpdate(title = title, description = desc)` applies exactly those. The not-found path (`personal?.let { … }`) is unchanged, and the `if (update.field != null)` guards implement "null means leave alone" faithfully.

The tests cover the right things — both lookup branches (`findByDocId` and the `findById` fallback), the new null-means-unchanged semantics, and the not-found case asserting `update` is never called. Dropping the old `updateCount` assertions is correct; they existed only to prove the lambda fired.

**Worth knowing about the sentinel:** `null` means "don't change", so this API can't set a field *to* null. That's fine as things stand — the fragment reads from `EditText`s, which yield `""` rather than null, so clearing a description still round-trips as an empty string and does update. It only becomes a limitation if something later needs to genuinely null a column, at which point the usual fix is a wrapper type rather than a bare nullable.

**Nit:** the diff carries the version bump (`6764` → `6765`) in `app/build.gradle`. That's the drain's mechanical edit rather than anything to review, but it does mean this branch has claimed `0.67.65` — if another PR merges and takes that version first, this one will conflict on those two lines.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16437, opened 2026-08-27
life: smoother key deduping (fixes #16435)

### LINKED ISSUE #16435: Make My Life de-duplication deterministic
**rating:** 59/100 &nbsp;·&nbsp; **proposed by:** codex(r1)

`LifeRepositoryImpl.dedupKey()` falls back to `System.identityHashCode(this)` when imageId, title and _id are all blank, so equivalent blank legacy rows dedup differently on each load and object identity leaks into a repository view model.

**Ask:** replace the `System.identityHashCode(this)` fallback with a deterministic key derived from the row's actual (even if blank) content, so equivalent blank rows dedup consistently across loads.

**Files:** `repository/LifeRepositoryImpl.kt`

**Verification:** Verified verbatim at LifeRepositoryImpl.kt:65-71.

dogi:
## Summary

`LifeRepositoryImpl.dedupKey()` fell back to `System.identityHashCode(this)` when `imageId`, `title`, and `_id` were all blank, so equivalent blank legacy rows deduped differently on each load — object identity leaked into a repository view model.

This replaces the non-deterministic fallback with a deterministic key derived from the row's actual (even if blank) content: `listOf(userId, isVisible, weight)`. Equivalent blank rows now dedup consistently across loads.

## Files

- `repository/LifeRepositoryImpl.kt` — replaced `System.identityHashCode(this)` fallback in `dedupKey()` with `listOf(userId, isVisible, weight)`.

## Test

- Added `getMyLifeByUserId_dedupsEquivalentBlankRowsAcrossLoads` in `LifeRepositoryImplTest.kt`, which feeds two equivalent blank `MyLife` rows (no `imageId`/`title`/`_id`) and asserts they collapse to a single row on every load.

Fixes #16435.

---
This PR was created by an AI agent (OpenHands) on behalf of the user.

@dogi can click here to [continue refining the PR](https://app.all-hands.dev/conversations/3cb0cdf9-b08d-42c0-bd86-dec9ef555149)


dogi:
**Review — rating: 86/100**

This is a genuine bug fix, and the title undersells it. `System.identityHashCode(this)` is per-*instance*, and Room materialises fresh objects on every query — so for any row lacking `imageId`, `title`, and `_id`, the fallback key was unique by construction and `distinctBy { it.dedupKey() }` could never merge anything. The dedup wasn't merely non-deterministic; for those rows it was a **no-op**, and the key also changed between loads of the same row.

`listOf(userId, isVisible, weight)` fixes it properly: `List` equality and `hashCode` are element-wise, so equivalent rows now produce equal keys. The heterogeneous return type (`String` from the first three branches, `List<Any?>` from the fallback) is fine here — a `String` can never equal a `List`, so the branches can't collide with each other.

The test is a real regression test: two separately constructed `blankRow()` instances with identical content, asserting a single row survives. Under `identityHashCode` that returns 2 and fails.

**Worth stating explicitly, since it's a behaviour change:** rows that previously *always* survived can now collapse. It only affects rows with no `imageId`, no `title` and no `_id` that also share `userId`, `isVisible` and `weight` — degenerate rows that are indistinguishable in the UI anyway, and `weight` normally differs per entry — so I think this is the intended semantics finally taking effect rather than data loss. But it's the kind of thing worth a line in the description.

**Two nits:**

1. The cross-load half of the test doesn't test what it says. `coEvery { … } returns listOf(blankRow(), blankRow())` evaluates the list once, so both loads receive the *same two instances*. To exercise "deterministic across loads with fresh objects", it needs `coAnswers { listOf(blankRow(), blankRow()) }` so each call builds new ones — which is precisely the scenario `identityHashCode` broke.
2. **Conflict heads-up:** #16368 (currently `ready`) rewrites `getMyLifeByUserId`'s signature immediately below this hunk — this diff still shows the old `ensureLatest` parameter, so it's based on pre-#16368. Adjacent hunks in the same file; whichever lands second will need a rebase.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16438, opened 2026-08-27
resources: smoother collections tags filtering (fixes #16410)

### LINKED ISSUE #16410: Hoist the search-string lowercase out of the collections filter lambda
**rating:** 65/100 &nbsp;·&nbsp; **proposed by:** codex(r2), glm(r2)

`filterTags` evaluates `charSequence.lowercase(Locale.ROOT)` **inside** the filter lambda, so the search string is re-lowercased once per tag on every keystroke.

**Ask:** hoist the `lowercase(Locale.ROOT)` call on the search string outside the filter lambda so it runs once per keystroke instead of once per tag.

**Files:** `ui/resources/CollectionsFragment.kt`

**Verification:** Verified at CollectionsFragment.kt:112-115.

dogi:
Hoisted `charSequence.lowercase(Locale.ROOT)` call on the search string outside of `list.filter` in `CollectionsFragment.kt` so that the search string is lowercased once per keystroke instead of once per tag.

Fixes #16410

---
*PR created automatically by Jules for task [11911479327397367514](https://jules.google.com/task/11911479327397367514) started by @dogi*


dogi:
**Review — rating: 87/100**

Correct and worth more than the diff size suggests. `lowercase(Locale.ROOT)` is pure and `Locale.ROOT` is constant, so hoisting is behaviour-preserving — and this is a search filter, so the lambda ran once per collection **on every keystroke**. That's N redundant allocations and conversions per character typed, now one.

The per-element `it.name?.lowercase(...)` can't be hoisted for obvious reasons, so this takes the half that was actually hoistable.

**Optional follow-up, not a request:** `it.name?.contains(charSequence, ignoreCase = true) == true` would drop the remaining N allocations too, since it compares in place rather than materialising a lowercased copy of every name. The caveat is that `ignoreCase` does char-by-char case folding rather than `Locale.ROOT` lowercasing, which can differ for scripts with special casing rules — and this app ships Arabic, Nepali and Somali locales, so that's a real (if narrow) semantic change rather than a free win. Worth measuring against a large collection list before deciding.

No test, which is right for a pure hoist with no observable behaviour change.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16439, opened 2026-08-27
courses: smoother activities repository visiting (fixes #16430)

### LINKED ISSUE #16430: Remove the repository-to-repository hop from course visit logging
**rating:** 60/100 &nbsp;·&nbsp; **proposed by:** codex(r1)

`ActivitiesRepositoryImpl.logCourseVisit` calls `userRepository.get().getUserByName(userId)` to fetch two provenance fields, widening the data-layer graph when the repository could read the same row through the UserDao abstraction other repositories already use.

**Ask:** replace the `userRepository.get().getUserByName(userId)` hop with a direct `UserDao` lookup, matching the pattern other repositories use.

**Files:** `repository/ActivitiesRepositoryImpl.kt`

**Verification:** Verified at ActivitiesRepositoryImpl.kt:86-105.

dogi:
## Summary

`ActivitiesRepositoryImpl.logCourseVisit` fetched two provenance fields (`parentCode`, `createdOn`) by calling `userRepository.get().getUserByName(userId)` — a repository-to-repository hop that widened the data-layer graph for a single-row lookup. This PR replaces that hop with a direct `UserDao.getByName(userId)` call, matching the pattern other repositories (e.g. `EventsRepositoryImpl`) already use.

## Changes

- `ActivitiesRepositoryImpl.kt`
  - Injected `UserDao` directly into the constructor (provided by `RoomModule`, no DI wiring changes needed).
  - `logCourseVisit` now looks the user up via `userDao.getByName(userId)` instead of `userRepository.get().getUserByName(userId)`.
  - The `Lazy<UserRepository>` dependency is retained — `recordSyncActivity` still uses `userRepository.get().getUserById(userId)`, which is outside the scope of this issue.

## Verification

- `ActivitiesRepositoryImplTest.logCourseVisit inserts course activity` updated to mock `userDao.getByName(...)` instead of `userRepository.getUserByName(...)`; asserts the inserted `CourseActivity` still carries the `parentCode`/`createdOn` provenance fields.
- Full `ActivitiesRepositoryImplTest` suite (24 tests) passes:
  ```
  ./gradlew testDefaultDebugUnitTest --tests "org.ole.planet.myplanet.repository.ActivitiesRepositoryImplTest"
  → BUILD SUCCESSFUL, 24 tests, 0 failures
  ```

## Checklist

- [x] Code builds successfully
- [x] Unit tests pass
- [x] No new dependencies introduced
- [x] Existing behavior preserved (same `UserDao.getByName` lookup `UserRepositoryImpl` already delegated to)

Closes #16430.

@dogi can click here to [continue refining the PR](https://app.all-hands.dev/conversations/ab3a07ed-68a7-4474-8c13-b62ff96c7348)


dogi:
**Review — rating: 89/100**

Zero behaviour change, and I verified rather than assumed it: `UserRepositoryImpl.getUserByName` (line 117) is a pure one-line delegation to `userDao.getByName(name)`, which is exactly what this now calls. Same query, same `name = :name LIMIT 1` semantics — notably **not** the neighbouring `getByNameIgnoreCase`, which would have silently changed matching for users whose stored name differs in case.

Cutting out the middleman is the right call here. It drops a `Lazy<UserRepository>` hop — indirection that usually exists only to break a Hilt cycle — for a lookup that was never more than a DAO call, and it matches the documented pattern of repositories injecting the DAOs they need directly. `userRepository` still has other uses in this class, so the injection stays load-bearing rather than becoming dead weight.

The test swaps its stub to `userDao.getByName` and the constructor gains the new parameter in the same diff, so the hand-built test instance and the Hilt graph stay in step.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16440, opened 2026-08-27
login: smoother achievements editing (fixes #16407)

### LINKED ISSUE #16407: Hoist the achievement resource dialog's dup-check into a Set
**rating:** 66/100 &nbsp;·&nbsp; **proposed by:** glm(r2), kimi(r2)

`createResourceList` loops `for (i in list.indices)` calling `prevList.contains(list[i].title)` where prevList is a List — a linear scan inside a linear loop, O(n*m) over resources x previously selected titles.

**Ask:** convert `prevList`'s title lookup to a `Set` (or build a `Set` of titles once before the loop) so the containment check is O(1).

**Files:** `ui/user/EditAchievementFragment.kt`

**Verification:** Verified at EditAchievementFragment.kt:477-482. glm's version cites the exact method and predicate; kimi's cites a slightly wrong enclosing function name.

dogi:
Hoisted the achievement resource dialog's duplicate check lookup into a Set in `EditAchievementFragment.kt`. Refactored `setUpOldAchievement`, `showResourceListDialog`, and `createResourceList` signatures and implementation to use `Set<String?>`, reducing containment checks during resource list creation to O(1).

Fixes #16407

---
*PR created automatically by Jules for task [18088940552520620276](https://jules.google.com/task/18088940552520620276) started by @dogi*


dogi:
**Review — rating: 89/100**

Checked the only consumer, and the swap is safe for the two reasons that matter. `createResourceList` uses `prevList` exclusively as `prevList.contains(list[i].title)` inside the loop — it's never iterated and its size is never read — so `HashSet` losing insertion order and collapsing duplicate titles is unobservable. That's what makes plain `HashSet` fine here rather than needing `LinkedHashSet` defensively.

The complexity win is real rather than theoretical: the loop runs over the *entire* resource library (`resourcesRepository.getAllLibraries()`), so it was O(library × previously-selected) linear scans on a dialog that can open against a large library. Now it's one hash lookup per row.

Threading `Set<String?>` through all three signatures — `setUpOldAchievement`, `showResourceListDialog`, `createResourceList` — rather than converting at the call site is the right call; it makes the membership-only contract visible in the types instead of leaving a `List` that invites someone to start iterating it later.

No test, which is appropriate for a data-structure swap with no observable behaviour change.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16441, opened 2026-08-27
all: smoother file utils url resolving (fixes #16409)

### LINKED ISSUE #16409: Parse the resource URL once in FileUtils.getSDPathFromUrl
**rating:** 66/100 &nbsp;·&nbsp; **proposed by:** glm(r2)

`getSDPathFromUrl` calls both `getIdFromUrl(url)` and `getResourceRelativePathFromUrl(url)`, and each independently does `url.toUri().pathSegments` + `indexOf('resources')` — the same URI parsed and scanned twice per call, on a path reached from `checkFileExist` during InlineResourceAdapter bind.

**Ask:** parse the URL's path segments once and share the result between the id and relative-path extraction instead of parsing twice.

**Files:** `utils/FileUtils.kt`

**Verification:** The three functions exist and the double parse is consistent with their bodies. Shares a file with the storage-stats task — sequence them.

dogi:
This change optimizes `FileUtils.getSDPathFromUrl` so that `url?.toUri()?.pathSegments` is parsed once and shared between ID extraction (`getIdFromSegments`) and relative-path extraction (`getResourceRelativePathFromSegments`), eliminating duplicate URI parsing and path scans during file existence checks.

Fixes #16409

---
*PR created automatically by Jules for task [7412718303610633158](https://jules.google.com/task/7412718303610633158) started by @dogi*


dogi:
**Review — rating: 82/100**

The core change is right: `getSDPathFromUrl` was calling `getIdFromUrl(url)` and `getResourceRelativePathFromUrl(url)`, each of which independently did `url.toUri().pathSegments` — so every resource path parsed the URI twice. Extracting segment-based helpers and parsing once is the correct fix, and pushing the shared `getFileNameFromSegments` down removes a third parse on the fallback path.

I checked the substitution that could have changed behaviour: `Uri.lastPathSegment` (old `getFileNameFromUrl`) versus `pathSegments.lastOrNull()` (new). Android implements the former as exactly the latter, returning null for an empty path, so they're equivalent. Null and empty-path cases also land the same: a null URL yields `""` from both old and new, and a URL like `http://host` with no `resources` segment falls through to the filename branch and produces `""` either way.

**Nit — the PR orphans a function and then keeps it.** `getResourceRelativePathFromUrl` had exactly one caller: the line in `getSDPathFromUrl` that this diff rewrites to call `getResourceRelativePathFromSegments` directly. It's retained as a private delegating wrapper with no callers left, which Kotlin will flag as "never used". Deleting it finishes the refactor.

**Conflict heads-up:** #16340 also rewrites `getSDPathFromUrl` (renaming `createFilePath` → `resolveFilePath` and removing its `mkdirs`). These two touch the same function and will conflict. #16340 is currently `change` pending its download-path fix, so this will likely land first — worth knowing when that one comes back.

**Minor:** the new helpers add three more `e.printStackTrace()` calls. Consistent with the rest of this file, but the opposite direction from #16388, which routed this kind of reporting through `DiagnosticsRepository`. Not worth blocking on; worth knowing the file is drifting against the trend.

**Verdict:** mergeable.

---

Okuro3499 (dismissed):
```
File: app/src/main/java/org/ole/planet/myplanet/utils/FileUtils.kt, getResourceRelativePathFromUrl (lines 125-133)                                                               
Problem: Dead code — this private function is no longer called anywhere (confirmed via grep across the repo); getSDPathFromUrl now goes through                                  
getResourceRelativePathFromSegments directly.
```

dogi:
## Review — rating 82/100

Clean refactor. `getSDPathFromUrl` parsed the same URL up to three times — once in `getIdFromUrl`, once in `getResourceRelativePathFromUrl`, and a third time in the `getFileNameFromUrl` fallback. Now it parses once and threads the segment list down. `Uri.parse` plus `getPathSegments` does real string scanning and list allocation, and each old call built a fresh `Uri` so nothing was shared, so this is a genuine if modest win on a hot path.

I walked the equivalence case by case, since this decides where downloaded files land on disk:

| Input | Master | This PR |
|---|---|---|
| null url | `getIdFromUrl` → `""`; `getResourceRelativePathFromUrl` early-returns `getFileNameFromUrl(null)` → `""` | `segments == null` → `""` / `""` |
| `toUri()` throws | both helpers catch independently; the relative-path fallback re-parses and throws again into its own catch → `""` | caught once in `getSDPathFromUrl`, `segments` null → `""` |
| `.../resources/<id>/a/b.pdf` | id `<id>`, path `a/b.pdf` decoded | identical |
| no `resources` segment | falls back to `getFileNameFromUrl` | falls back to `getFileNameFromSegments` |

The last row is the one worth confirming: `Uri.getLastPathSegment()` is implemented as `pathSegments.isEmpty() ? null : pathSegments[size - 1]` over the same segment list, so swapping it for `segments.lastOrNull()` is exact, including the trailing-slash case where Android drops the empty final segment. Decoding stays asymmetric in the same way as before — the id segment is returned raw, the relative path is URL-decoded — so no callers shift.

Keeping `getFileNameFromUrl` and `getIdFromUrl` public with unchanged behaviour is the right call; they have callers outside this file.

### Worth adding a test

There is already a `FileUtilsTest.kt`, and `getSDPathFromUrl` feeds `DownloadService:365` and `DownloadWorker:114` — it decides the on-disk destination for every downloaded resource. A silent behaviour change here misplaces files rather than throwing, which is the kind of thing that surfaces weeks later as "the resource downloaded but won't open". Two cases would cover the refactor:

- `.../resources/<id>/sub%20dir/file%20name.pdf` → asserts the id, and that the relative path is decoded to `sub dir/file name.pdf`
- a URL with no `resources` segment → asserts the fallback resolves to the decoded last segment

Not blocking, since I am satisfied the behaviour is preserved, but this is exactly the function where a pinned expectation pays for itself.

Moving to `ready`.

---

Okuro3499 (approved):

PR #16442, opened 2026-08-27
all: smoother notifications view modelling (fixes #16402)

### LINKED ISSUE #16402: Parallelize the notifications enrichment lookups
**rating:** 68/100 &nbsp;·&nbsp; **proposed by:** deepseek(r3)

`NotificationsViewModel.loadNotifications` runs four independent suspend DB lookups strictly serially on the Main scope; task-team-names, join-request-details and the unread badge count have no dependency on each other and can be async/awaited together.

**Ask:** parallelize the four independent lookups in `loadNotifications` using `coroutineScope { async { } }`.

**Verification:** Verified the four call sites in loadNotifications. Note this file is also the target of two other shipped tasks — sequence them.

dogi:
Parallelized four independent database lookups in NotificationsViewModel.loadNotifications (task team names by IDs, task team names by titles, join request details, and unread count) using coroutineScope { async { ... } }. Added unit test coverage in NotificationsViewModelTest to verify concurrent execution and map merging.

Fixes #16402

---
*PR created automatically by Jules for task [12801607045272148142](https://jules.google.com/task/12801607045272148142) started by @dogi*


dogi:
**Review — rating: 80/100**

The concurrency is correctly structured: four `async` blocks inside `coroutineScope`, so all four repository lookups overlap, the scope waits for all of them, and a failure in any one cancels the siblings and propagates rather than leaving orphaned work. Four sequential DB round trips collapsing into one concurrent batch is a real latency win on a screen that loads on every open.

**The change worth being deliberate about is that `taskTitles` is no longer filtered.** The old code built the title list *after* awaiting the ID lookup:

```kotlin
val taskTitles = taskNotifications
    .filter { it.relatedId.isNullOrEmpty() || !taskTeamNames.containsKey(it.relatedId) }
    .mapNotNull { parseTaskDate(it.message)?.first }
```

so tasks already resolved by ID never contributed a title, and when everything resolved by ID the title query was skipped entirely (`taskTitles.isNotEmpty()` guard). That filter is impossible to keep under parallelism — you can't filter on a result you haven't awaited — so `getTaskTeamNamesByTaskTitles` now runs on **every** task title, every time. The query count is the same, but it's wider, and it always fires. Whether that's a net win depends on how often tasks resolve by ID; concurrency probably still pays for it, but it's more total DB work rather than strictly less.

Worth noting the merge precedence flipped with it — the old order let titles override IDs, the new one lets IDs override titles. The two maps are keyed in different spaces (task id vs. task title), so nothing actually collides, but the new direction is the safer one to have landed on: an ID match is more precise than a title match, so if the key spaces ever did overlap, IDs winning is what you'd want.

The test is good on the part that matters most — it stubs `getTaskTeamNamesByTaskTitles(listOf("Task By Id", "Task By Title"))`, so it explicitly documents the now-unfiltered title list, and it verifies the combined map produces the right output.

**Nit:** despite the name, `testLoadNotificationsParallelLookups` doesn't test parallelism — it asserts the four calls happen and the results merge, which passes equally against a sequential implementation. `runTest`'s virtual clock makes the real assertion cheap: stub each lookup with a `delay`, then assert total elapsed `currentTime` matches the slowest rather than the sum.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16443, opened 2026-08-27
teams: less members leader handle is more (fixes #16406)

### LINKED ISSUE #16406: Drop the redundant second member fetch in MembersFragment.handleMakeLeader
**rating:** 67/100 &nbsp;·&nbsp; **proposed by:** deepseek(r3)

`handleMakeLeader` calls `getJoinedMembersWithVisitInfo(teamId)` and `updateData(members, false)` and then immediately calls `loadMembers()`, which re-queries the same thing one line later. The eager fetch is dead work and hard-codes isLoggedInUserTeamLeader=false, unlike the real path.

**Ask:** remove the redundant eager fetch + `updateData(members, false)` call and let the subsequent `loadMembers()` call do the (correct) refresh.

**Files:** `ui/teams/members/MembersFragment.kt`

**Verification:** Verified verbatim at :157-161 — the discarded fetch and the wrong leader flag are both real.

dogi:
Removed redundant getJoinedMembersWithVisitInfo(teamId) and updateData(members, false) calls in MembersFragment.handleMakeLeader, allowing the subsequent loadMembers() call to fetch members and accurately compute leader status.

Fixes #16406

---
*PR created automatically by Jules for task [3123538855405110573](https://jules.google.com/task/3123538855405110573) started by @dogi*


dogi:
**Review — rating: 88/100**

**Not a duplicate of #16456** — I'd flagged these two as a possible collision from their titles ("remove redundant member fetch" vs "reuse resolved user"), and having read both diffs that suspicion was wrong. This one edits `handleMakeLeader` (~line 156); #16456 edits the imports, field block, setup, and `loadMembers`. Disjoint hunks, disjoint concerns, and they merge independently. Both are rated on their own merits.

The removed lines aren't merely redundant — they were **producing a wrong intermediate state**:

```kotlin
val members = teamsRepository.getJoinedMembersWithVisitInfo(teamId)
membersAdapter?.updateData(members, false)   // isLeader hard-coded false
loadMembers()                                 // immediately re-fetches and passes the real isLeader
```

`loadMembers()` does the same fetch, computes `isLeader` properly, and calls `updateData(members, isLeader)`. So the deleted code paid for a second full `getJoinedMembersWithVisitInfo` round trip *and* pushed `isLeader = false` into the adapter first — a brief window where the leader-only controls would be hidden from someone who had just been made leader, before the correct value landed a moment later.

Deleting it removes the duplicate query and that flicker in one go. `loadMembers()` fully covers the behaviour, so nothing is lost.

No test, which is fine here — this is a deletion whose behaviour is entirely subsumed by the surviving call on the next line.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16444, opened 2026-08-27
teams: smoother courses filtering (fixes #16408)

### LINKED ISSUE #16408: Convert the team-course availability filter from List to Set membership
**rating:** 66/100 &nbsp;·&nbsp; **proposed by:** glm(r2)

`showAddCourseDialog` computes `allCourses.filter { it.courseId !in existingIds }` where existingIds is the `List<String>` returned by getTeamCourseIds — `!in` on a List is a linear scan, so the Add Course dialog runs O(all courses x already-added courses) on open.

**Ask:** convert `existingIds` to a `Set<String>` before the filter so membership checks are O(1).

**Files:** `ui/teams/courses/TeamCoursesFragment.kt`

**Verification:** Verified at TeamCoursesFragment.kt:107-109.

dogi:
Converts `existingIds` to a `Set<String>` in `TeamCoursesFragment.kt` before filtering available courses. This changes the `!in` membership check from linear $O(N)$ scan to constant time $O(1)$ lookup per item.

Fixes #16408

---
*PR created automatically by Jules for task [624141280455479068](https://jules.google.com/task/624141280455479068) started by @dogi*


dogi:
**Review — rating: 86/100**

Correct and the right shape. `!in` resolves to `contains`, which is a linear scan on a `List` and a hash lookup on a `Set`, so the filter over `getAllCourses()` goes from O(all courses × team courses) to O(all courses + team courses). Membership is the only thing `existingIds` is used for, so losing order and collapsing any duplicates is unobservable.

Being straight about the magnitude: a team's course list is usually small, so the linear scan wasn't necessarily hurting — and `.toSet()` does allocate. The win scales with the course catalogue rather than the team, which is the direction that actually grows on a mature planet, and the conversion is cheap enough that it's worth taking regardless.

No test, which is right for a data-structure swap with no observable behaviour change.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16445, opened 2026-08-27
all: smoother base permission usage handling (fixes #16404)

### LINKED ISSUE #16404: Drop the per-call Java reflection from the usage-stats permission check
**rating:** 68/100 &nbsp;·&nbsp; **proposed by:** kimi(r3)

`BasePermissionActivity.getUsagesPermission` resolves AppOpsManager methods via `Class.getMethod` + `Method.invoke` on every call, with a Q-vs-legacy branch, plus a catch-all exception surface — for an API that is public and directly callable at this project's minSdk 26.

**Ask:** replace the reflection with direct calls.

**Files:** `base/BasePermissionActivity.kt`

**Verification:** Verified verbatim at BasePermissionActivity.kt:50-67. One correction the task does not make: on Q+ the reflective target is `unsafeCheckOpNoThrow`, so a direct call needs either @RequiresApi(29) or the deprecated `checkOpNoThrow`; the task's step 3 (delete the Q branch) picks the latter without saying so.

dogi:
Replaced Java reflection in `BasePermissionActivity.getUsagesPermission` with direct calls to `AppOpsManager.unsafeCheckOpNoThrow` for Android Q+ and `AppOpsManager.checkOpNoThrow` for pre-Q. Added comprehensive unit tests in `BasePermissionActivityTest`.

Fixes #16404

---
*PR created automatically by Jules for task [1861409647708060952](https://jules.google.com/task/1861409647708060952) started by @dogi*


dogi:
**Review — rating: 89/100**

The reflection was working around a problem that doesn't exist here. `checkOpNoThrow` has been available since API 19 and `unsafeCheckOpNoThrow` since API 29 — with minSdk 26 and compileSdk 37, both resolve at compile time, and the existing `SDK_INT >= Q` branch already guards the runtime call. So `getMethod(...)`/`invoke(...)` bought nothing while costing type safety.

Dropping the `catch (Exception)` is also safe rather than merely tidy: these are the *NoThrow* variants, which signal failure by returning `MODE_ERRORED` instead of throwing. The old handler left `mode = -1`, which matched neither `MODE_DEFAULT` nor `MODE_ALLOWED` and so returned false — an unreachable path that existed only because reflection could throw. Remove the reflection and the handler has nothing left to catch. `getSystemService` sits outside the old `try` in both versions, so nothing changes there.

The three tests cover all three branches of the return logic — `MODE_ALLOWED`, `MODE_ERRORED`, and `MODE_DEFAULT` falling through to `checkCallingOrSelfPermission`. Stubbing both `unsafeCheckOpNoThrow` and `checkOpNoThrow` with the same value is the right call: it keeps the tests independent of whichever SDK level Robolectric reports, rather than silently exercising only one arm.

The `android.util.Log` import stays needed — three other `Log` calls remain in the file — so no orphaned import, which has been a recurring leftover elsewhere in this queue.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16456, opened 2026-08-27
teams: smoother base members user caching (fixes #16405)

### LINKED ISSUE #16405: Reuse the resolved user in MembersFragment instead of re-querying per refresh
**rating:** 67/100 &nbsp;·&nbsp; **proposed by:** opus(r2)

The signed-in user is resolved three ways for one screen: BaseTeamFragment.onCreate loads it into the inherited `user` field, MembersFragment.onViewCreated calls `userSessionManager.getUserModel()` again (:77), and `loadMembers()` calls it a third time (:103). getUserModel() is a Room userDao.getById, and loadMembers() has five call sites.

**Ask:** reuse the already-resolved `user` field from `BaseTeamFragment` instead of re-querying via `userSessionManager.getUserModel()` at the redundant call sites.

**Files:** `ui/teams/members/MembersFragment.kt`

**Verification:** Verified the onViewCreated coroutine and loadMembers().

dogi:
Reuses the already-resolved user property inherited from BaseTeamFragment in MembersFragment instead of re-querying userSessionManager.getUserModel() during onViewCreated and loadMembers(). Also cleans up the unused userSessionManager field injection.

Fixes #16405

---
*PR created automatically by Jules for task [7412718303610633896](https://jules.google.com/task/7412718303610633896) started by @dogi*


dogi:
**Review — rating: 62/100**

**Not a duplicate of #16443** — I'd flagged the two as a possible collision from their titles, and the diffs show that was wrong: #16443 touches only `handleMakeLeader`, this one touches the imports, field block, setup, and `loadMembers`. No overlap, so both are rated independently.

The intent is right: `BaseTeamFragment` already resolves the current user, so `MembersFragment` separately injecting `UserSessionManager` to re-fetch it is duplication worth removing.

**But `user` isn't available when this reads it.** `BaseTeamFragment` populates it asynchronously (lines 41–43):

```kotlin
override fun onCreate(savedInstanceState: Bundle?) {
    super.onCreate(savedInstanceState)
    teamId = requireArguments().getString("id", "")
    lifecycleScope.launch {
        user = userRepository.getUserModel()   // suspends on the DB read
        loadTeamDetails()
    }
}
```

`getUserModel()` suspends, so that coroutine parks and `onCreate` returns immediately. Android then proceeds to `onCreateView`/`onViewCreated` — and this PR converts the setup block from a coroutine into a **synchronous** read:

```kotlin
val resolvedUser = user ?: UserEntity()
requestsAdapter?.setUser(resolvedUser)
membersAdapter?.setUserId(resolvedUser.id)
```

At that point `user` is normally still null, so `resolvedUser` becomes a blank `UserEntity()` and `membersAdapter` gets `setUserId("")`. The `?: UserEntity()` fallback is what makes this nasty — it swallows the failure instead of throwing, so the symptom isn't a crash.

`loadMembers()` has the same problem as a race rather than a near-certainty: it runs in a coroutine whose first statement suspends, so `user` *may* be set by the time `val currentUserId = user?.id` executes — or may not. The old `userSessionManager.getUserModel()` was an authoritative suspend read that always returned the real user.

**The user-visible consequence** is in the next line:

```kotlin
val isLeader = members.any { it.user.id == currentUserId && it.isLeader }
```

With `currentUserId` null or `""`, `isLeader` is false, so an actual team leader intermittently loses their leader-only controls depending on whether the base's coroutine happened to win the race.

**Fix:** await the base's resolution rather than reading the field synchronously. Either keep the setup work in a coroutine that observes the user (exposing it from `BaseTeamFragment` as a `StateFlow<UserEntity?>` or a `Deferred<UserEntity?>` and collecting/awaiting it), or have the base assign `user` before the suspend point. Dropping the redundant `UserSessionManager` injection is still the right goal — it just needs a value that's actually ready.

**Verdict:** needs another round.

---

dogi:
## Review: 84/100 — mergeable

The substitution is genuinely equivalent, and dropping the duplicate injection is the right cleanup.

**`UserSessionManager.getUserModel()` was a pure passthrough** (`UserSessionManager.kt:38-40`):

```kotlin
suspend fun getUserModel(): UserEntity? {
    return userRepository.getUserModel()
}
```

So routing through `BaseTeamFragment`'s existing `userRepository` reaches the same DAO read by a shorter path. That's what makes removing the `@Inject lateinit var userSessionManager` field safe rather than a behavior change — worth stating, because "swap one service for another" normally deserves suspicion.

**The memoization is the actual win.** `user` is already a `var user: UserEntity? = null` on `BaseTeamFragment:19`, populated in `onCreate`. On `master`, `MembersFragment` ignored it and re-read the user from the database twice more — once in `onViewCreated`, once per `loadMembers()` call. Since `onCreate`'s coroutine populates `user` first in the normal case, both of those now hit the cached value.

**Two things worth knowing, neither blocking:**

- **The dedupe is best-effort, not a lock.** `ensureUserResolved()` is a check-then-act around a suspending call, and `onCreate`'s coroutine and `onViewCreated`'s coroutine both run on Main. If the second starts before the first resumes, both observe `user == null` and both fetch. Harmless — same value, idempotent assignment — it just doesn't always save the read. Not worth a mutex here; worth not being surprised by it in a profile.
- **The cached `UserEntity` can go stale after a sync.** `loadMembers()` only reads `.id`, which can't change, so that path is safe. But `requestsAdapter?.setUser(resolvedUser)` receives the whole entity, so a roles change pulled by a background sync wouldn't be reflected until the fragment is recreated. Importantly this isn't new exposure: `BaseTeamFragment.user` is already cached from `onCreate` and used that way by every other subclass — this PR makes `MembersFragment` consistent with the hierarchy rather than introducing a fresh staleness window. If that window ever matters, it should be fixed once in the base class for all of them.

**Housekeeping checks:** `UserEntity` is still imported and used (`?: UserEntity()`); `javax.inject.Inject` and the `UserSessionManager` import are correctly dropped with the field.

**No test.** `ensureUserResolved()` is a natural unit to pin — "second call doesn't re-hit the repository" is one `coVerify(exactly = 1)` — but it lives on an abstract Fragment, which makes it awkward without Robolectric. Fair to skip; a follow-up that moves this resolution into a ViewModel would make it testable and would also fix the staleness note above.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16457, opened 2026-08-27
all: smoother server reachability time providing (fixes #16401)

### LINKED ISSUE #16401: Add a ServerReachabilityProvider test and move its cache TTL onto TimeProvider
**rating:** 69/100 &nbsp;·&nbsp; **proposed by:** copilot(r3)

`ServerReachabilityProvider` injects OkHttpClient, ServerUrlMapper and DispatcherProvider but uses `System.currentTimeMillis()` for both the cache read and write, against the project's TimeProvider convention, and has no unit test at all.

**Ask:** inject `TimeProvider` and use it for the cache TTL read/write instead of `System.currentTimeMillis()`; add a unit test covering the cache hit/miss/expiry behavior. Do NOT merge this with `MainApplication`'s separate `HttpURLConnection` reachability cache — they are distinct.

**Files:** `utils/ServerReachabilityProvider.kt`, `(new) test/.../utils/ServerReachabilityProviderTest.kt`

**Verification:** Verified the file exists and has no test. The task correctly warns not to merge in MainApplication's separate HttpURLConnection reachability cache.

dogi:
Injected TimeProvider into ServerReachabilityProvider to manage cache TTL checks and writes via timeProvider.now(). Added unit tests in ServerReachabilityProviderTest to cover cache hit, miss, and expiry behavior.

Fixes #16401

---
*PR created automatically by Jules for task [14313847186258640838](https://jules.google.com/task/14313847186258640838) started by @dogi*


dogi:
## Review: inject `TimeProvider` into `ServerReachabilityProvider`

**Rating: 88/100** — correct, behavior-preserving, and it buys real test coverage for a class that had none.

### Correctness

The production change is two mechanical substitutions of `System.currentTimeMillis()` → `timeProvider.now()` inside `isServerReachable`. `SystemTimeProvider.now()` *is* `System.currentTimeMillis()` and `TimeModule.provideTimeProvider()` already binds it, so runtime behaviour is bit-identical and the new constructor parameter resolves through Hilt without any module edit. `@Singleton` on the provider is unaffected.

The tests hold up against the real implementation:

- `UrlMapping(url, null)` matches the declared `UrlMapping(primaryUrl, alternativeUrl = null, extractedBaseUrl = null)` — compiles.
- The fallback test's `match { it.url.toString() == url || it.url.toString() == "$url/" }` is the right defensive shape: `Request.Builder().url("http://primary.com")` normalises through `HttpUrl` to a trailing slash, so the bare-form arm alone would have missed.
- `tryConnect` runs `withContext(dispatcherProvider.io)`, and `TestDispatcherProvider` wires `io` to the same `StandardTestDispatcher` the `runTest` is driving, so the suspension resolves inside the test body rather than deadlocking.
- `Response` is mocked `relaxed = true`, which covers the `close()` that `.use { }` invokes on the way out.
- `timeProvider` is a test-class field, and JUnit4 constructs a fresh instance per test method, so `currentTime` genuinely resets to `100_000` between the TTL-hit and TTL-expiry cases rather than leaking.

The TTL boundary tests are the valuable ones: 10 s → one `newCall`, 30_001 ms → two. That is exactly the assertion that was structurally impossible before this change, since `System.currentTimeMillis()` can't be moved 30 seconds without sleeping.

### One direction question

`TimeProvider`'s own KDoc says:

> Use `now` for timestamps and time-based decisions persisted across restarts. For measuring elapsed durations within a single process, use `android.os.SystemClock.elapsedRealtime()` directly instead — it is monotonic and immune to wall-clock jumps.

A 30-second in-memory reachability TTL is precisely "measuring elapsed duration within a single process", so by the letter of that doc this call site is the one `TimeProvider` says *not* to use it for. In practice the pre-existing code already used wall clock, so this PR doesn't make the clock-jump exposure worse — a backward NTP correction makes `now() - checkedAt` negative and pins a cache entry as fresh until the clock catches up, and that was true before too. But it's worth deciding as a project which way this class should go: either `elapsedRealtime()` (correct, and then the TTL needs a different injection seam to stay testable), or `TimeProvider` and a note that the KDoc's carve-out is advisory. I'd take this PR as-is either way — the tests are the thing that was missing, and they'd survive a later swap of the seam.

### Nits

- `REACHABILITY_CACHE_TTL_MS` stays a private instance `val` in SCREAMING_SNAKE, so the tests hard-code `30_000` in comments and magic numbers. Not this PR's mess, but hoisting it to a `companion object` `const` would let the tests reference it and stop the two from drifting.
- `isPrimaryServerReachable does not use cache` doesn't stub `serverUrlMapper` at all — correct, since that path never touches it, but an explicit `verify(exactly = 0) { serverUrlMapper.processUrl(any()) }` would state that intent the way the blank-URL test does.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16458, opened 2026-08-27
lifel: smoother personals repository device name providing (fixes #16433)

### LINKED ISSUE #16433: Take the Application Context out of the personals upload path
**rating:** 59/100 &nbsp;·&nbsp; **proposed by:** copilot(r1)

`PersonalsRepositoryImpl` takes an @ApplicationContext Context solely so uploadPersonalDocument can call `NetworkUtils.getCustomDeviceName(context)` when serializing — pinning the personals domain repository to Android for one string.

**Ask:** remove the `@ApplicationContext Context` dependency from `PersonalsRepositoryImpl` — either inject the device name directly (e.g. via a small provider abstraction) or resolve it at a layer that's already Android-aware, so the repository itself no longer needs `Context`.

**Files:** `repository/PersonalsRepository.kt`, `repository/PersonalsRepositoryImpl.kt`, `ui/personals/PersonalsViewModel.kt`, `ui/personals/PersonalsFragment.kt`, `test/.../PersonalsRepositoryImplTest.kt`

**Verification:** Not independently re-verified beyond file existence; the described shape is consistent with the rest of the repository layer.

dogi:
## Summary

`PersonalsRepositoryImpl` took an `@ApplicationContext Context` solely so `uploadPersonalDocument` could call `NetworkUtils.getCustomDeviceName(context)` when serializing the document — pinning the personals domain repository to Android for a single string. Notably, `NetworkUtils.getCustomDeviceName(context)` **ignores the context argument entirely** and just delegates to `SharedPrefManager.getCustomDeviceName()`.

This removes the `@ApplicationContext Context` dependency from `PersonalsRepositoryImpl`, replacing it with a small `DeviceNameProvider` abstraction injected via Hilt. The Android-aware lookup now lives in the provider's implementation (`SharedPrefDeviceNameProvider`), so the repository itself no longer needs `Context`.

## Changes

- **New** `utils/DeviceNameProvider.kt`: an injectable `DeviceNameProvider` interface + `SharedPrefDeviceNameProvider` implementation (reads from `SharedPrefManager`) + a Hilt `DeviceNameModule` binding. Mirrors the existing `TimeProvider`/`DispatcherProvider` pattern.
- `repository/PersonalsRepositoryImpl.kt`: drop `@ApplicationContext Context`, inject `DeviceNameProvider` instead; `uploadPersonalDocument` now calls `deviceNameProvider.getCustomDeviceName()`. Removes the `Context`/`NetworkUtils` imports.
- `test/.../PersonalsRepositoryImplTest.kt`: mock `DeviceNameProvider` instead of `Context`, and add a test asserting the custom device name is sourced from the provider and lands in the serialized request body.

The `PersonalsRepository` interface, `PersonalsViewModel`, and `PersonalsFragment` are unchanged — they already depend on the repository interface and don't construct the impl directly.

## Verification

Ran `./gradlew testDefaultDebugUnitTest --tests "org.ole.planet.myplanet.repository.PersonalsRepositoryImplTest"` locally — all 19 tests pass (including the new one). The Hilt graph compiles cleanly with the new binding.

Fixes #16433.

_Note: This PR was created by an AI agent (OpenHands) on behalf of the repo maintainers._

@dogi can click here to [continue refining the PR](https://app.all-hands.dev/conversations/e9ed6aa6-0f6a-49f6-909d-e4286bfe3937)


dogi:
## Review: `DeviceNameProvider` seam for `PersonalsRepositoryImpl`

**Rating: 88/100** — behaviour-identical, and it deletes a dependency that was never doing anything.

### Correctness

The substitution is exact, and in a stronger way than the PR description claims. `NetworkUtils.getCustomDeviceName(context)` is:

```kotlin
    val spm = coreEntryPoint.sharedPrefManager()
    return spm.getCustomDeviceName()
}
```

The `context` parameter is **ignored** — the lookup goes through `NetworkUtils`' own lazily-built `CoreDependenciesEntryPoint` (rooted at `MainApplication.context`) to `SharedPrefManager.getCustomDeviceName()`, which is a plain `pref.getString(CUSTOM_DEVICE_NAME, "") ?: ""`. `SharedPrefDeviceNameProvider` injects the same `SharedPrefManager` and calls the same method, so the returned string is identical for every input, and the repository loses both an `@ApplicationContext` it wasn't using and a hop through an `EntryPointAccessors` static singleton in favour of ordinary constructor injection. That's a net simplification, not just a relocation.

`SharedPrefManager` is `@Inject constructor(@ApplicationContext context, gson)`, so the new `@Binds` resolves without touching any other module, and `@Singleton` on the binding matches the singleton it wraps.

### Test

The new test is the right assertion — it captures the posted `JsonObject` and checks `customDeviceName` came from the provider, so it would catch a regression that silently re-read the old static.

One detail the setUp comment gets slightly wrong: it hedges that `NetworkUtils` statics are kept mocked "in case shared helpers (e.g. `Personal.serialize`) reach into it indirectly." They don't reach in *in case* — `Personal.serialize` unconditionally calls `NetworkUtils.getUniqueIdentifier()` and `NetworkUtils.getDeviceName()` (`Personal.kt:47-48`), so those two stubs are load-bearing and should be stated as such. The `getCustomDeviceName(any<Context>())` stub, on the other hand, is now genuinely dead — nothing under test reaches it — and dropping that one line also lets the `android.content.Context` import go from the test file.

### Nits

- The Hilt `@Module` (`DeviceNameModule`) is declared in `utils/` alongside the interface, whereas every other module in the project lives in `di/`. `TimeProvider` — which the KDoc explicitly says this mirrors — keeps its interface in `utils/` and its binding in `di/TimeModule.kt`. Following that split would keep the module inventory in `di/` complete.
- This converts 1 of the 18 `NetworkUtils.getCustomDeviceName(...)` call sites. `TeamsRepositoryImpl`, `SubmissionsRepositoryImpl`, `ActivitiesRepositoryImpl`, `UserRepositoryImpl`, `UploadManager`, and the `model/` serializers all still hold a `Context` purely to pass to a function that discards it. Worth a follow-up issue so the new seam doesn't become a one-off — and worth eventually deleting the `Context` parameter from `NetworkUtils.getCustomDeviceName` itself, since it is pure decoration today.

### Coordination

`PersonalsRepositoryImpl` and `PersonalsRepositoryImplTest` are contended right now — #16436 and #16470 both touch them. This PR rewrites the constructor signature and the test's `setUp`, so whichever of the three lands second will need its base merged in and CI re-run rather than a clean auto-merge.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16459, opened 2026-08-27
teams: smoother surveys repository eligibility filtering (fixes #16424)

### LINKED ISSUE #16424: Push survey eligibility filtering into Room and make the exclusion lookup a Set
**rating:** 61/100 &nbsp;·&nbsp; **proposed by:** codex(r1), codex(r2)

`SurveysRepositoryImpl.getTeamOwnedSurveys`/`getAdoptableTeamSurveys` load every survey via the broad `ExamDao.getByType("surveys")` and filter in Kotlin; `getAdoptableTeamSurveys` additionally builds `excludedIds` by list concatenation and calls `excludedIds.contains(it.id)` per survey.

**Ask:** push the eligibility filtering into a Room `@Query` instead of loading every survey and filtering in Kotlin, and convert `excludedIds` to a `Set` so the exclusion check is O(1) instead of a linear scan.

**Files:** `repository/SurveysRepositoryImpl.kt`, `data/room/dao/ExamDao.kt`

**Verification:** File and method names verified. codex proposed the SQL push-down in r1 and the Set fix in r2 — the Set half is the cheaper, lower-risk one; counted once.

dogi:
Moved team-owned and adoptable team survey filtering logic from Kotlin into Room DAO SQL queries in ExamDao. Updated getAdoptableTeamSurveys to construct excludedIds as a Set<String> for O(1) exclusion lookups. Added unit test coverage for new Room queries and updated repository tests.

Fixes #16424

---
*PR created automatically by Jules for task [2991893772100732926](https://jules.google.com/task/2991893772100732926) started by @dogi*


dogi:
**Review — rating: 89/100**

Both predicates translate faithfully, including the cases that usually bite when moving filters into SQL:

- `teamId = :teamId` matches Kotlin's `it.teamId == teamId` even for NULL `teamId` rows — SQL yields NULL (falsy) and Kotlin yields false, so both exclude.
- `id IN (:submissionIds)` with an **empty** collection is safe here specifically because SQLite permits an empty list and evaluates `x IN ()` as false — which reduces the `OR` to the teamId clause, exactly matching Kotlin's `emptySet.contains(...)`.
- The `NOT IN` NULL trap is avoided: `excludedIds` is built from `getTeamSubmissionExamIds` plus `mapNotNull { it.sourceSurveyId }`, so no NULL can enter the list. One NULL in there would make `id NOT IN (…)` evaluate to NULL for *every* row and silently return no adoptable surveys at all.

**The DAO tests are exactly what this kind of change needs.** `ExamDaoTest` drives the real queries against real rows and covers the type filter (the `type = "other"` row correctly excluded), the `teamId OR submissionId` branch, and the shareable-plus-exclusion path. I asked for precisely this on #16397, where the moved predicate was only covered by a mock; here the SQL itself is under test.

**Two nits:**

1. **The no-argument `getAdoptableTeamSurveys()` overload is redundant.** Since `x IN ()` is false in SQLite, `id NOT IN ()` is true for every row, so the `excludedIds` version already returns the full shareable set when the collection is empty — making the `if (excludedIds.isEmpty())` branch in the repository unnecessary. Keeping it is defensible if you'd rather not lean on SQLite's non-standard empty-list handling, but then it's worth a comment saying so, since otherwise it reads as an oversight.
2. **The `IN` list isn't chunked.** Each element binds a SQLite variable, and `UserRepositoryImpl.getUsersByIds` already chunks at 400 for exactly this reason. `excludedIds` is bounded by the number of surveys a team has touched, so it's unlikely to approach the limit — but the codebase has a precedent worth matching if this ever grows.

`.toSet()` on `excludedIds` is a good addition beyond dedup: it also keeps duplicate submissions from multiplying the bind parameters.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16460, opened 2026-08-27
all: smoother notifications group view modelling (fixes #16403)

### LINKED ISSUE #16403: Regroup notifications only when notifications change, not on every selection tap
**rating:** 68/100 &nbsp;·&nbsp; **proposed by:** opus(r2)

`groupedItems` combines four flows — _notifications, _selectedIds, _collapsedGroups, _expandedGroups — and rebuilds everything through `buildGroupedList` on each emission: a full groupBy with a `type.lowercase(Locale.ROOT)` per notification, a re-derived ordered type list and a per-group unread recount. Selecting one item in multi-select re-groups and re-counts the entire inbox, though grouping depends solely on _notifications.

**Ask:** restructure so grouping/lowercasing only recomputes when `_notifications` changes; selection/collapse/expand state changes should not trigger a full regroup.

**Verification:** Verified: combine at :50, groupBy+lowercase at :232-242.

dogi:
Decouples notification grouping, type lowercasing, ordering, and unread counts from UI state updates by maintaining a separate `groupedNotifications` flow mapped from `_notifications`. Selection, expansion, and collapse state changes now combine with pre-grouped notifications to produce list items without performing full regroup operations.

Fixes #16403

---
*PR created automatically by Jules for task [872571780335414568](https://jules.google.com/task/872571780335414568) started by @dogi*


dogi:
**Review — rating: 89/100**

Well-targeted. `groupedItems` combined four flows, so *any* selection tap, collapse, or expand re-ran the full `groupBy` — including a `lowercase(Locale.ROOT)` per notification and the type ordering. Selection is the most frequent interaction on that screen, and it was paying for a complete regroup every time. Splitting the grouping into a flow derived only from `_notifications` means selection changes now rebuild just the flat list, which they genuinely have to since `isSelected` is per-item.

I walked the equivalence and it holds: the `groupBy` predicate, the `TYPE_ORDER.filter { … } + grouped.keys.filter { … }).distinct()` ordering, the `unreadCount` computation, the `isExpanded` precedence (expanded → collapsed → `unreadCount > 0`), and `inSelectionMode` are all unchanged. The empty-input short circuit is preserved on both sides of the split. `isExpanded` correctly stayed in `buildGroupedList` rather than moving into the group, since it depends on the collapsed/expanded sets.

**The new test is the right one to have written.** The specific risk in this refactor is caching *too much* — pulling grouping into a separate flow could easily have left selection state stale. `testToggleSelectionUpdatesSelectionStateOnListItems` asserts both `isSelected` and `isSelectionMode` flip on and back off, which is exactly the failure mode a mis-scoped cache would produce.

**Nit:** `mapNotNull { grouped[type] ?: return@mapNotNull null }` can't actually yield null — `orderedTypes` is derived from `grouped.keys`, so the lookup always hits. A plain `map` says that more clearly.

**Conflict heads-up:** #16442 (currently `ready`) also edits this file and appends its new test at the same insertion point in `NotificationsViewModelTest.kt` (right after line 229). The production hunks are far apart, but the test files will collide — whichever lands second needs a rebase there.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16465, opened 2026-08-27
resources: less repository enriched libraries is more (fixes #16453)

### LINKED ISSUE #16453: Privatize getEnrichedLibraries on ResourcesRepository
**rating:** 56/100 &nbsp;·&nbsp; **proposed by:** kimi(r1)

`ResourcesRepository.getEnrichedLibraries` is on the public interface but is called only from inside ResourcesRepositoryImpl (by getResourceListModels), and it returns a `LibraryWithMetadata` holding raw JsonObject ratings — an internal join pipeline that downstream layers should not be able to depend on.

**Ask:** remove `getEnrichedLibraries` from the public `ResourcesRepository` interface and make it private to the impl. **Important:** the task's stated justification is wrong — `ResourcesRepositoryImplTest.kt:225` DOES call `repository.getEnrichedLibraries(false, ...)` through the interface, so that test must move or change with it.

**Files:** `repository/ResourcesRepository.kt`, `repository/ResourcesRepositoryImpl.kt`

**Verification:** Premise holds for production code but the task's justification is wrong: it claims 'the public caller set was empty from a source-tree grep', and ResourcesRepositoryImplTest.kt:225 calls repository.getEnrichedLibraries(false, ...) through the interface. That test has to move or change with it.

dogi:
## Summary

`ResourcesRepository.getEnrichedLibraries` was on the public repository interface but is only called internally — by `ResourcesRepositoryImpl.getResourceListModels`. It returns a `LibraryWithMetadata` holding raw `JsonObject` ratings, an internal join pipeline that downstream layers should not depend on.

This change removes `getEnrichedLibraries` from the `ResourcesRepository` interface and makes it `private` on `ResourcesRepositoryImpl`, encapsulating the internal join pipeline behind the existing public `getResourceListModels` entry point.

## Changes

- `repository/ResourcesRepository.kt` — removed `getEnrichedLibraries` from the interface.
- `repository/ResourcesRepositoryImpl.kt` — changed `getEnrichedLibraries` from `override` to `private`.
- `repository/ResourcesRepositoryImplTest.kt` — the existing test called `repository.getEnrichedLibraries(false, ...)` through the interface. Since the method is now private, the test was retargeted onto the public `getResourceListModels` method (which drives the same code path) and renamed accordingly. The test still asserts the public-not-user DAO branch is exercised.

## Verification

Ran the unit tests for the affected class locally:

```
./gradlew testDefaultDebugUnitTest --tests "org.ole.planet.myplanet.repository.ResourcesRepositoryImplTest"
```

Result: BUILD SUCCESSFUL — 36 tests, 0 failures, 0 errors. The retargeted test `getResourceListModels fetches public-not-user items when not my course lib` passes.

_Note on the issue's stated justification:_ the issue body's premise that the public caller set was empty did not hold for the test source tree — `ResourcesRepositoryImplTest.kt:225` called `getEnrichedLibraries` through the interface. That test was moved onto `getResourceListModels` as part of this change, as the issue's own verification note required.

Closes #16453.

---

@dogi can click here to [continue refining the PR](https://app.all-hands.dev/conversations/e11f37fb-b994-40d7-baa2-a2bb3cc3a196)


dogi:
**Review — rating: 88/100**

Narrowing the interface is the right move: `getEnrichedLibraries` returned `LibraryWithMetadata`, an enrichment type that only exists to feed `getResourceListModels`, so exposing it on the repository contract invited callers to depend on an intermediate representation. Removing an interface member is also self-verifying — any external caller fails the build, and both flavors plus both test shards are green.

**The test change is a genuine improvement rather than a mechanical fixup.** It previously reached past the public API to assert on `getEnrichedLibraries` directly; it now drives `getResourceListModels` and asserts the same outcome. Same coverage of the enrichment path, but through the surface callers actually use — so the test survives future refactors of the private helper instead of pinning it.

**Nit that would finish the job:** `LibraryWithMetadata` is *declared* in `ResourcesRepository.kt` (line 11), and after this change nothing in that file references it — the only remaining user is the now-private method in the impl. So the type that just became an implementation detail still lives in the public interface's file. Moving it beside the impl (or into `model/`) completes the encapsulation this PR is arguing for.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16471, opened 2026-08-27
sync: smoother repository error handling (fixes #16336)

### LINKED ISSUE #16336: Replace printStackTrace with tagged Log.e on the sync path
Sync failures never reach logcat: `SyncManager` has 5 naked `printStackTrace()` (218, 383, 406, 417, 553) despite already importing `android.util.Log`; `LoginSyncManager` has 10; `SyncRepositoryImpl` has 2 (plus indexed `for (i in 0 until array.size())` `JsonArray` walks at 139-144 and 185-192 that Kotlin can iterate directly).

**Ask:** replace all `printStackTrace()` calls in these three files with tagged `Log.e(...)`, and convert the indexed `JsonArray` walks in `SyncRepositoryImpl` to idiomatic Kotlin iteration.

**Files:** `services/sync/SyncManager.kt`, `services/sync/LoginSyncManager.kt`, `repository/SyncRepositoryImpl.kt`

**Verification:** `printStackTrace` counts confirmed exactly: SyncManager 5, LoginSyncManager 10, SyncRepositoryImpl 2.

dogi:
Replaced naked printStackTrace() calls in SyncManager (5 calls), LoginSyncManager (10 calls), and SyncRepositoryImpl (2 calls) with tagged Log.e(...) logging so sync failures reach logcat. Also refactored indexed JsonArray walks in SyncRepositoryImpl to idiomatic Kotlin iteration.

Fixes #16336

---
*PR created automatically by Jules for task [9995105938434153279](https://jules.google.com/task/9995105938434153279) started by @dogi*


dogi:
**Review — rating: 85/100**

A real improvement, and the valuable part isn't the API swap — it's that each call site gets a **specific** message. `printStackTrace()` writes to `System.err` untagged and at no severity, so on a sync path with a dozen catch blocks you get a stack trace with no indication of which phase died. "Resource batch failed" / "Library sync failed" / "Authentication encoding failed" turn each one into something filterable and diagnosable.

The incidental loop rewrites are correct — `JsonArray` implements `Iterable<JsonElement>`, so `for (element in array)` is equivalent to the indexed form and drops the repeated `get(i)`. They are scope creep for a PR titled after `printStackTrace`, though; if this needs a rebase for any reason, they'd be worth splitting.

Mocking `Log.e` in `LoginSyncManagerTest` is the necessary consequence of the change and follows the existing `mockkStatic(Base64::class)` pattern in that file.

**One direction question worth settling once, rather than per-PR.** This adds `android.util.Log` to three more classes, while #16388 — merged into `ready` earlier in this queue — solved the same problem in `DownloadRepositoryImpl` by routing through the injected `DiagnosticsRepository`, which lands the record in Room where it can actually be retrieved from a device in the field rather than only in logcat. And #16353 was sent back precisely for deleting a diagnostic instead of rehoming it. On the sync path, where failures happen on someone else's phone in a low-connectivity deployment, diagnostics is arguably the more useful destination.

`Log.e` is a strict improvement over `printStackTrace` either way, so this isn't a blocker — but with #16595 queued to do the same conversion across config, secure-store, file, camera and download-worker paths, it's worth deciding now whether the target is `Log.e` or `DiagnosticsRepository`, so the two PRs don't standardise on different answers.

**Nit:** the tag is a repeated string literal at each call site; a `private const val TAG` per class is the usual shape, and would keep `SyncManager`'s error tag from drifting away from its existing `"SyncPerf"` debug tag by accident.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16475, opened 2026-08-27
dashboard: smoother voice date count querying (fixes #16455)

### LINKED ISSUE #16455: Replace the dashboard voice-date list with a count query
**rating:** 55/100 &nbsp;·&nbsp; **proposed by:** jules(r1)

`DashboardViewModel` calls `voicesRepository.getCommunityVoiceDates(...)` twice (:292, :293) and consumes the results only as `uniqueDates.size` for voiceCount (:313, :321) — materializing unique-date lists to produce two Ints on the dashboard load path.

**Ask:** add a count query that returns the unique-date count directly and use it for both call sites, instead of materializing the date lists just to read `.size`. Note the original task only addresses one of the two calls — handle both.

**Files:** `ui/dashboard/DashboardViewModel.kt`, `repository/VoicesRepository.kt`, `repository/VoicesRepositoryImpl.kt`, `data/room/dao/NewsDao.kt`

**Verification:** Verified the call sites and the `.size`-only consumption. The task addresses only one of the two calls.

dogi:
## What

`DashboardViewModel.evaluateChallengeDialog` called `voicesRepository.getCommunityVoiceDates(...)` twice and consumed the results only as `.size` for `voiceCount` and `allVoiceCount` — materializing unique-date lists on the dashboard load path just to read two Ints.

This replaces both call sites with a single SQL `COUNT` over `DISTINCT` calendar days, so the unique-date lists are no longer built.

> Note: the original task addressed only one of the two calls. This PR handles **both** call sites, as the issue body asks.

## Changes

- **`data/room/dao/NewsDao.kt`** — add `countDistinctCommunityVoiceDates` and `countDistinctCommunityVoiceDatesForUser`. Each runs `COUNT(*)` over a subquery selecting `DISTINCT strftime('%Y-%m-%d', time / 1000, 'unixepoch')` for rows in the time window whose `viewIn` JSON contains a `community` section entry (matching the in-memory `isCommunitySection` filter the repo used before).
- **`repository/VoicesRepository.kt`** — replace `getCommunityVoiceDates(...): List<String>` with `getCommunityVoiceDateCount(...): Int`.
- **`repository/VoicesRepositoryImpl.kt`** — implement the count by delegating to the two DAO queries (user-scoped when `userId != null`, all-community otherwise); drop the now-unused `isCommunitySection`, `getDateFromTimestamp`, `dateFormatter` helpers and their `java.time` imports.
- **`ui/dashboard/DashboardViewModel.kt`** — use `getCommunityVoiceDateCount` for both `voiceCount` and `allVoiceCount`.

## Tests

- `NewsDaoTest`: Room integration tests asserting the count query returns the unique-day count (collapsing same-day posts, excluding non-community posts) and is correctly scoped by `userId`.
- `VoicesRepositoryImplTest`: the count delegates to the all-community query when `userId` is null and to the user-scoped query otherwise.
- `DashboardViewModelTest`: `evaluateChallengeDialog` calls the count method exactly once per call site (user-scoped and all-community).

All 25 affected tests pass (`./gradlew testDefaultDebugUnitTest` for the three classes).

Fixes #16455.

_This PR was created by an AI agent (OpenHands) on behalf of the maintainer team._

@dogi can click here to [continue refining the PR](https://app.all-hands.dev/conversations/fd50f159-931c-4855-8d30-5b192987bf50)


dogi:
## Review: push community-voice date counting into SQL

**Rating: 88/100** — a real win (a windowed table scan + per-row JSON parse collapses to one aggregate), and the DAO test exercises the actual SQL rather than a mock. One semantic change is worth a decision before merge.

### The `LIKE` predicate holds

Substring-matching JSON in SQL is the kind of shortcut that usually breaks on a formatting difference, so I traced every writer of `News.viewIn`:

- `VoicesRepositoryImpl:432` (sync path) — `plainGson.toJson(JsonUtils.getJsonArray("viewIn", doc))`
- `VoicesRepositoryImpl:228, 266` (label add/remove) — `gson.toJson(array)`
- `News.getViewInJson` (local post creation) — `JsonUtils.gson.toJson(viewInArray)`

`NetworkModule.providePlainGson()` is a bare `Gson()`, `provideGson()` is a `GsonBuilder` with only `excludeFieldsWithModifiers` + `serializeNulls`, and `JsonUtils.gson` is a bare `Gson()`. None enables `setPrettyPrinting`, so every path writes `"section":"community"` with no whitespace. Key order doesn't matter either — `getViewInJson` emits `_id, section, name` while a synced doc can emit any order, and the pattern only anchors on the one pair. There's precedent for exactly this in the same file: `isAlreadyShared` (`:79-81`) already does `viewIn?.contains("\"_id\":\"$viewInId\"")`.

So the predicate is sound as written, and the DAO comment correctly identifies *why* it's sound. Worth keeping that comment glued to the query — it is the only thing standing between this and a silent zero if someone ever swaps in a pretty-printing `Gson`.

The rest of the SQL checks out: `time / 1000` is integer division on positive epoch millis, `strftime('%Y-%m-%d', …, 'unixepoch')` buckets by day, and `COUNT(*) FROM (SELECT DISTINCT …)` is the right shape (`COUNT(DISTINCT strftime(…))` would have been equivalent and slightly cheaper to read, but this is fine). The test timestamps are accurate — `1735257600000` really is 2024-12-27T00:00:00Z, `+86400000` is the 28th, and `1735430400000` is the 29th — and the `day2SameDay` and `teamPost` rows pin down the two behaviours that matter.

### The one behaviour change: local day → UTC day

This is the part I'd want called out explicitly. The old code bucketed with:

```kotlin
Instant.ofEpochMilli(timestamp).atZone(ZoneId.systemDefault()).format("yyyy-MM-dd")
```

— **device-local** calendar days. The new SQL uses `'unixepoch'`, which is **UTC**. For any user not on UTC, the day boundaries move by their offset, so posts made in the evening (west of UTC) or early morning (east of UTC) can merge into or split out of a neighbouring day. That directly feeds `voiceCount >= 5`, the gate on the challenge dialog — a user with exactly five qualifying local days could land on four UTC days and stop seeing the prompt, or vice versa.

The window itself (`startTime`/`endTime` are hard-coded epoch constants in `DashboardViewModel:284-285`) was always absolute, so only the *bucketing* changes, and it's a fixed historical window for a one-off challenge, so the blast radius is small. But it is a change, and the PR description presents this as a pure optimisation. Two clean resolutions:

1. **Keep parity** — add SQLite's `'localtime'` modifier: `strftime('%Y-%m-%d', time / 1000, 'unixepoch', 'localtime')`. That reproduces `ZoneId.systemDefault()` exactly, is a one-token change, and stays a single aggregate query. (The DAO test's UTC-based expectations would then need a fixed TZ, which Robolectric can pin.)
2. **Keep UTC** — fine if the challenge is meant to be evaluated consistently regardless of where the device is, but then say so in the PR body and the DAO comment, so the next reader doesn't "fix" it back.

I'd take (1), since it makes the change genuinely behaviour-preserving and the whole point of this PR is that nothing observable moves.

Minor related note: SQLite's `LIKE` is ASCII case-insensitive by default, whereas the old `optString("section") == "community"` was case-sensitive. A hypothetical `"Section":"Community"` would now count. Server data is lowercase, so this is theoretical — mentioning it only so it isn't a surprise.

### Nits

- `NewsDao.getInTimeRange` and `getInTimeRangeForUser` have no callers left anywhere in `main` or `test` after this change — the replaced function was their only consumer. Worth deleting in this PR while the context is fresh, rather than leaving two Room-generated queries nobody calls.
- The `DashboardViewModelTest` addition verifies the two delegations fire once each, but doesn't assert the `voiceCount >= 5` gate that the counts exist to feed. Since the ViewModel now receives an `Int` directly instead of a list whose `.size` it took, a test asserting `ChallengeDialogData.voiceCount == 5` would be cheap and would cover the actual wiring.
- Repository-level tests mock the DAO and so can't catch a `strftime`/`LIKE` regression — but `NewsDaoTest` runs against real Room, so that gap is already covered. Good split.

**Verdict:** mergeable once the UTC-vs-local question is settled one way or the other — either resolution is fine, but it shouldn't merge as an unremarked change.

---

dogi:
Correction to my own numbers above: I rated this **88** while simultaneously saying it shouldn't merge until the UTC-vs-local bucketing is settled. Those two don't agree, and the verdict is the one I stand behind.

The local-day → UTC-day shift is a correctness change to a user-facing gate (`voiceCount >= 5`), not a stylistic note, and this PR presents itself as a pure optimisation. Revising to **76/100** and moving to `change`. Everything else in the review stands — the `LIKE` predicate is sound, the SQL is right, and the DAO test is the good kind. Adding `'localtime'` to both `strftime` calls (and pinning the test's timezone) should be enough to bring it straight back to `review` at the original score.

---

dogi:

---

dogi:
@dogi addressed in [a28d709](https://github.com/open-learning-exchange/myplanet/commit/a28d70939).

**UTC -> local day bucketing.** Both `strftime` calls now carry `'localtime'`, so the `COUNT(DISTINCT ...)` buckets by the device's local calendar day -- matching the prior in-memory `getDateFromTimestamp`, which used `ZoneId.systemDefault()`. The `voiceCount >= 5` gate keeps its meaning for devices off UTC instead of silently shifting to a UTC-day window. The DAO comment now states this is *why* `'localtime'` is there (the one thing standing between this and a silent regression).

**Test determinism.** `NewsDaoTest` now pins `TimeZone` to UTC in `@Before` (restored in `@After`), following the existing `TimeUtilsTest` pattern, so the day-bucketing assertions don't depend on the runner's locale. The test timestamps keep >= 24 h day gaps, so the expected counts hold under any offset -- the pin just removes runner-to-runner variance.

Everything else from the review stands untouched: the `LIKE` predicate (no pretty-printing Gson writes `viewIn`), the `time / 1000` integer division, and the `COUNT(*) FROM (SELECT DISTINCT ...)` shape.

_This comment was created by an AI agent (OpenHands) on behalf of the maintainer team._

dogi:
## Review: 85/100 — mergeable

Substantially reworked since my last pass, and the pushdown is now a faithful translation rather than a rewrite. I verified the two things that could have silently broken the challenge gate.

### The `LIKE` filter matches every writer of `viewIn` ✅

Swapping a JSON parse for a substring match only works if every producer emits compact JSON. There are four writers, and all four do:

| writer | serializer | compact? |
|---|---|---|
| `VoicesRepositoryImpl:432` (sync-in) | `@PlainGson` → `Gson()` | ✅ |
| `VoicesRepositoryImpl:228` (share to community) | `provideGson()` — `GsonBuilder().excludeFieldsWithModifiers(…).serializeNulls()`, **no** `setPrettyPrinting()` | ✅ |
| `VoicesRepositoryImpl:266` (unshare) | same `gson` | ✅ |
| `News.getViewInJson` (:236) | `JsonUtils.gson` → `Gson()` | ✅ |

So `"section":"community"` with no space is what's actually on disk. Key **ordering** doesn't matter either — `getViewInJson` writes `_id` before `section` while `:222` writes `section` first, and a substring match on the pair is indifferent to both. Multi-entry arrays also behave identically: the old `isCommunitySection` returned true on *any* matching element, and so does `LIKE`. `viewIn IS NULL` yields NULL → excluded, matching the old `?.let { } ?: false`.

### The date bucketing preserves its meaning ✅

`strftime('%Y-%m-%d', time / 1000, 'unixepoch', 'localtime')` buckets by the device's local calendar day, matching the old `Instant.ofEpochMilli(t).atZone(ZoneId.systemDefault())`. Integer division truncates toward zero, which is floor for positive epoch millis. The DAO comment spelling this out is exactly the right thing to leave behind, since this is a user-facing gate (`voiceCount >= 5`) where an off-by-one day changes who sees the dialog.

### The win is real

`evaluateChallengeDialog` runs on the dashboard path and previously loaded **every** news row in the window into memory, `org.json`-parsed each one's `viewIn`, mapped to formatted date strings, and `distinct()`ed — twice (user-scoped and global). Now it's two `COUNT(*)`s. On a device with a large news table that's a meaningful startup saving.

### Two things worth addressing

**1. The format dependency is untested.** The correctness of the `LIKE` rests entirely on Gson staying compact, and nothing enforces it: `NewsDaoTest` hand-writes `"[{\"section\":\"community\",…}]"` as a literal. If someone ever swaps in a pretty-printing Gson for `viewIn`, the count silently becomes 0, the challenge dialog silently stops appearing, and **every test still passes**. Building the fixture through the production path instead —

```kotlin
viewIn = plainGson.toJson(JsonArray().apply { add(JsonObject().apply {
    addProperty("section", "community"); addProperty("_id", "planet@parent")
}) })
```

— makes the test fail the moment the serializer changes shape, which is the failure you want to be loud.

**2. `TimeZone.setDefault(UTC)` probably doesn't pin what the comment claims.** SQLite's `'localtime'` modifier resolves through the C library's timezone (the `TZ` environment), not the JVM's `TimeZone.getDefault()`. So the comment "Pin the timezone so the day-bucketing COUNT queries are deterministic across runners" is likely not doing that. The tests happen to pass regardless — I worked the fixtures through UTC−5 and the distinct-day count is still 2, because the timestamps are spaced far enough apart — but the guarantee is accidental. Either drop the claim, or choose timestamps that would straddle a day boundary in some zone and assert that they don't.

### Cleanup left behind

After this lands, `NewsDao.getInTimeRange` and `getInTimeRangeForUser` have **no remaining callers** — `VoicesRepositoryImpl:305/307` was their only use and this PR deletes that block. Worth removing them in the same change (or noting deliberately why they stay), since a dead `SELECT *` next to the new `COUNT` is exactly what a future reader would reach for by mistake.

**Tests otherwise good:** the DAO tests cover same-day collapsing, non-community exclusion at an identical timestamp, and user scoping — all genuinely new behavior. The repository tests pin the null/non-null branch selection with `coVerify(exactly = 0)` on the other query, which is the right shape.

**Verdict:** mergeable.

---

Okuro3499 (dismissed):
```
 NewsDaoTest.kt (app/src/test/.../NewsDaoTest.kt:22-24): fix or drop the TimeZone.setDefault() pinning — it doesn't influence the SQL query's 'localtime' resolution (confirmed by
    probe), so it's misleading test infrastructure. Either (a) actually verify 'localtime' bucketing across real timezone offsets on-device/CI (e.g. running the shard with          
    TZ=America/New_York ./gradlew ... to see if results change), or (b) sidestep SQLite's tz handling entirely by computing local-day boundary timestamps in Kotlin                  
    (ZoneId.systemDefault()) and passing them as query bounds instead of relying on 'localtime' inside SQL — that preserves the exact old semantics without depending on native      
    SQLite timezone resolution. 
```

olevim:

olevim:
@dogi @Okuro3499 Followed up on the review in [`3cd76b1`](https://github.com/open-learning-exchange/myplanet/commit/3cd76b1). Three things, all proportional:

**1. Dropped the misleading `TimeZone.setDefault(UTC)` pin** (Okuro3499's `CHANGES_REQUESTED` + dogi point #2). SQLite's `'localtime'` resolves through the C library's `TZ`, not the JVM's `TimeZone.getDefault()`, so the pin didn't control what its comment claimed — the determinism was accidental. The production `'localtime'` modifier is kept (it's the faithful translation of the old `ZoneId.systemDefault()` device-local bucketing); Okuro3499's option (b) — passing local-day boundary timestamps as query bounds — would not reproduce the `DISTINCT`-calendar-day count, so it isn't equivalent. Instead the fixtures rely on offset-robustness: `day1`/`day2` are exactly 24 h apart, so under *any* fixed offset they land on distinct local calendar days → count is always 2; `day2SameDay` is +1h and stays with `day2`; `teamPost` is excluded by section. So the assertions hold without a TZ pin, and the new comment records why none is needed.

**2. Built the `viewIn` fixtures through the production serializer** (dogi point #1). The count tests now build `viewIn` via `Gson().toJson(JsonArray{JsonObject{section=community,_id=planet@parent}})` — the same compact `Gson()` (no `setPrettyPrinting()`) every writer of `News.viewIn` uses — instead of hand-written JSON literals. The output is byte-identical to the literals (`[{"section":"community","_id":"planet@parent"}]`), so the `LIKE '%"section":"community"%'` still matches and the assertions are unchanged; the difference is that a future pretty-printing Gson would now fail the test loudly instead of silently zeroing the count and leaving every test green.

**3. Deleted the dead `getInTimeRange`/`getInTimeRangeForUser`** (dogi cleanup). They lost their only caller in this PR — the replaced `VoicesRepositoryImpl.getCommunityVoiceDates` block — and a dead `SELECT *` next to the new `COUNT` is exactly the trap a future reader would reach for by mistake.

Not done: dogi's nit about a `voiceCount == 5` gate assertion in `DashboardViewModelTest`. The `shouldPrompt` gate depends on `LocalDate.now()` falling in a hardcoded **past** window (2024-11-30 → 2025-01-16); today is 2026-08-28, so `dialogData` is always `null` regardless of counts. Asserting the gate value would mean mocking `LocalDate.now()`, which is invasive clock-mocking for a one-off expired challenge — not proportional to the benefit. The existing test already pins that each count method fires once per call site.

CI is running on the new commit; the `NewsDaoTest` carrying the fixture change is in the test shards.

_This comment was created by an AI agent (OpenHands) on behalf of the maintainer team._

dogi:
## Review — rating 88/100 (was 85)

Pushing this into SQL is clearly right, and the two assumptions it rests on both hold. I verified them rather than taking the comments on trust, since a silent mismatch here would quietly change who gets the challenge dialog.

### The `LIKE` on `viewIn` is safe because every writer is compact

This was the assumption I most wanted to check — a substring match against serialized JSON is normally fragile. It holds because **every** path that writes `viewIn` re-serializes through Gson rather than storing server text:

```kotlin
// VoicesRepositoryImpl.buildNewsFromJson — the sync path
news.viewIn = plainGson.toJson(JsonUtils.getJsonArray("viewIn", doc))
```

The server's formatting never survives; the array is parsed and re-emitted locally. The local-creation path (`News.getViewInJson` → `JsonUtils.gson.toJson`) and the label add/remove paths (`gson.toJson(array)`) are the same. And `grep setPrettyPrinting` over `app/src/main` returns nothing, so all three Gson instances emit `{"_id":"…","section":"community",…}` with no spaces. There is also precedent: `VoicesRepositoryImpl:81` already does `viewIn?.contains("\"_id\":\"$viewInId\"")` in production.

Worth noting the pattern is also not prone to false positives — `"section":"community-extra"` does not contain `"section":"community"` because of the closing quote.

Building the fixtures through `Gson().toJson(...)` rather than a hand-written string literal is the detail that makes this durable: if the serializer ever changes shape, `NewsDaoTest` fails instead of the dashboard silently under-counting.

### The timezone swap is a real difference, but not reachable here

`strftime(..., 'localtime')` resolves through SQLite's C-library timezone, whereas the old `getDateFromTimestamp` used the JVM's `ZoneId.systemDefault()`. Those can diverge if an app calls `TimeZone.setDefault()` — `grep TimeZone.setDefault app/src/main` returns nothing, so they cannot drift here.

The test fixtures are genuinely offset-robust, not just asserted to be: 24 h spacing means 12-27/12-28 00:00 UTC collapse to two distinct local days under UTC-5 (12-26, 12-27) and under UTC+13 (12-27, 12-28) alike, and the range predicate is on raw epoch `time`, so it is timezone-independent regardless.

### The win

Master loaded every `News` row in the window into memory, ran `org.json.JSONArray` over `viewIn` per row, formatted a date per row, then `.distinct()` — and `DashboardViewModel` does this **twice** (user-scoped and community-wide) on the challenge-dialog path. That is now two aggregate queries.

The `LIKE '%…%'` leading wildcard means no index can be used, so it is still a full scan of `news` — but master scanned exactly the same rows *and* deserialized JSON for each one, so this is strictly better. Not worth restructuring for.

Test coverage is thorough at all three layers: real in-memory Room for the SQL itself (distinct-day collapse, non-community exclusion, user scoping), `coVerify(exactly = 0)` on the sibling method to pin the delegation branch, and a ViewModel test asserting one call each for the user and community counts.

Moving to `ready`.

---

Okuro3499 (approved):

PR #16476, opened 2026-08-27
life: smoother achievements references info handling (fixes #16446)

### LINKED ISSUE #16446: Hoist the layout inflater and attach achievement rows to their parent
**rating:** 58/100 &nbsp;·&nbsp; **proposed by:** opus(r2)

`showAchievementAndInfo` and `showReference` each rebuild their container from scratch and call `LayoutInflater.from(activity)` once per row inside the loop, then inflate with no parent — discarding each row layout's own layout_* params before addView. showAchievementAndInfo is re-entrant from its own delete handler, so removing one attachment re-inflates every remaining row.

**Ask:** hoist the `LayoutInflater.from(activity)` call out of the per-row loop, and inflate each row with its parent (`attachToRoot = false`) so the row layout's own `layout_*` params are preserved.

**Files:** `ui/user/EditAchievementFragment.kt`

**Verification:** File verified; the per-row inflater and parentless inflate are consistent with the surrounding code. Not line-by-line re-verified.

dogi:
Hoisted `LayoutInflater.from(activity)` outside of per-row inflation loops in `EditAchievementFragment.kt` and updated `inflate` calls to pass the parent container (`binding.llAttachment` and `binding.llOtherInfo`) with `attachToRoot = false`. This preserves layout parameters and improves performance.

Fixes #16446

---
*PR created automatically by Jules for task [17037628316127523100](https://jules.google.com/task/17037628316127523100) started by @dogi*


dogi:
**Review — rating: 89/100**

Two changes, and the second is a real layout fix rather than the style tidy-up the title suggests.

Hoisting `LayoutInflater.from(activity)` out of the loop is the obvious win — it's a `getSystemService` lookup that was running once per row.

**Passing the parent is the substantive part.** Inflating with no root discards the row layout's own `layout_width`/`layout_height`, because there's no parent to turn them into LayoutParams — so `addView` afterwards applies the `LinearLayout`'s defaults (wrap_content × wrap_content). A row declared `match_parent` was silently rendering at wrap_content. `inflate(inflater, binding.llAttachment, false)` lets the parent generate the correct params from the XML.

`attachToRoot = false` is required here and is right: both loops call `addView` explicitly (`binding.llAttachment.addView(...)`, `binding.llOtherInfo.addView(...)`). With `true` the view would already have a parent and the subsequent `addView` would throw.

Also correctly scoped — `AlertReferenceBinding.inflate(LayoutInflater.from(activity))` a few lines below is left alone, which is right: that view goes into a dialog via `setView`, where a null root is the appropriate call.

No tests, which is fair — this is layout-params behaviour with no instrumented test source set in the repo, so there's nothing meaningful to assert in a unit test.

Worth noting #16440 also touches this file but in `setUpOldAchievement`/`createResourceList` (~lines 362–490), well clear of these hunks, so the two should merge independently.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16477, opened 2026-08-27
life: smoother dictionary repository inserting (fixes #16434)

### LINKED ISSUE #16434: Give the dictionary population a mutex so concurrent calls can't both parse
**rating:** 59/100 &nbsp;·&nbsp; **proposed by:** codex(r1)

`DictionaryRepositoryImpl.insertDictionaryData` checks `dictionaryDao.count()` before parsing and inserting, but two concurrent calls can both observe zero and both do the work. LifeRepositoryImpl already has a seedMutex for exactly this shape.

**Ask:** add a mutex around the check-then-insert sequence in `insertDictionaryData`, following the same `seedMutex` pattern already used in `LifeRepositoryImpl`.

**Files:** `repository/DictionaryRepositoryImpl.kt`

**Verification:** Verified the check-then-insert at DictionaryRepositoryImpl.kt:30-36 and the absence of any lock.

dogi:
## Summary

`DictionaryRepositoryImpl.insertDictionaryData` performed a check-then-insert (`dictionaryDao.count()` → parse → `insertAll`) with no synchronization, so two concurrent calls could both observe a count of zero and both parse and insert the dictionary file.

This wraps the check-then-insert sequence in a `seedMutex`, following the same pattern already established in `LifeRepositoryImpl.seedMyLifeIfEmpty`. The `count()` re-check now happens inside `withLock`, so only one call proceeds to parse/insert; any concurrent caller re-checks after the first releases the lock and short-circuits to `AlreadyPopulated`.

## Files

- `repository/DictionaryRepositoryImpl.kt` — added `seedMutex` (`kotlinx.coroutines.sync.Mutex`) and wrapped the check-then-insert block in `seedMutex.withLock { }`.
- `repository/DictionaryRepositoryImplTest.kt` — added a test covering overlapping `insertDictionaryData` calls, asserting the dictionary is parsed/inserted exactly once and the second caller returns `AlreadyPopulated`.

## Verification

- `./gradlew testDefaultDebugUnitTest --tests "org.ole.planet.myplanet.repository.DictionaryRepositoryImplTest"` — all 5 tests pass (4 pre-existing + the new concurrent test).
- The `FileMissing` early-return stays outside the lock (it needs no DB access and shouldn't be serialized behind inserters).

Closes #16434.

---

@dogi can click here to [continue refining the PR](https://app.all-hands.dev/conversations/ed635d2c-c609-4e46-8996-9c5d6b5bad2b)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved dictionary initialization to prevent duplicate population when multiple requests run concurrently.
  * Ensured dictionary data is inserted only once, with subsequent attempts recognized as already populated.

* **Tests**
  * Added coverage verifying safe behavior during simultaneous dictionary initialization requests.


dogi:
## Review: guard dictionary seeding with a `Mutex`

**Rating: 78/100** — the production change is right, but the test that exists to prove it doesn't exercise it and would pass on `master` unchanged.

### The fix itself is correct

The load-bearing fact is scoping, so I checked it: `RepositoryModule:84-86` binds `DictionaryRepositoryImpl` with `@Binds @Singleton` into `SingletonComponent`. That's what makes an *instance* field work as a guard — one repository instance app-wide means one `seedMutex`, so two `DictionaryActivity` instances (a relaunch, or a recreate while a seed is in flight) genuinely serialise against each other. Had the binding been unscoped, each injection point would have gotten its own mutex and the change would have been decorative. Worth a comment on the field noting the dependency, since a future "this doesn't need to be a singleton" cleanup would silently un-fix this.

The race is real and worth closing: `DictionaryEntity.id` is a fresh `UUID.randomUUID()` per row, so a double-seed doesn't collide on conflict — it duplicates the entire dictionary.

Control flow through the refactor holds up. `return@withLock DictionaryLoad.AlreadyPopulated` returns from the lock lambda, `withLock` propagates that value, and the `seedMutex.withLock { … }` expression is the tail of the `withContext` lambda, so the result reaches the caller unchanged. The `try`/`catch` moved *inside* the lock, which is the right side — `withLock`'s `finally` releases on the exception path.

### The concurrency test doesn't test concurrency

```kotlin
private val testDispatcher = UnconfinedTestDispatcher()
…
val results = listOf(
    async { dictionaryRepository.insertDictionaryData() },
    async { dictionaryRepository.insertDictionaryData() }
).map { it.await() }
```

Under `UnconfinedTestDispatcher`, `async` starts the coroutine immediately in the calling frame, and there is no real suspension anywhere along the path to yield at:

- `withContext(dispatcherProvider.io)` — `TestDispatcherProvider` wires `io` to the *same* dispatcher instance, so there's no interceptor change and no dispatch.
- `seedMutex.withLock` — uncontended, so `lock()` takes the fast path and never suspends.
- `dictionaryDao.count()` and `insertAll()` are `coEvery { … } answers { … }`, which return without suspending.

So the first `async` runs `count() → 0 → insertAll → inserted = true → Inserted` to completion *before the second `async` is even constructed*, and the second then sees `count() == 1`. The two coroutines never overlap, the mutex is never contended, and deleting `seedMutex` entirely produces the exact same `Inserted` / `AlreadyPopulated` split and the same `coVerify(exactly = 1)`. The test is green either way, which means it cannot protect the fix it was written for.

Making it discriminating is a small change — force an interleave by giving the check a real suspension point and using a dispatcher that doesn't run eagerly:

```kotlin
private val testDispatcher = StandardTestDispatcher()   // for this test
…
coEvery { dictionaryDao.count() } coAnswers { delay(10); if (inserted.get()) 1L else 0L }
```

With that, both coroutines get past the `count() == 0` check before either inserts. **Without** the mutex, `insertAll` fires twice and both results are `Inserted` — the assertions fail. **With** it, the second waits, re-reads `count() == 1`, and returns `AlreadyPopulated`. That version earns the name in the test.

(`runTest`'s virtual clock makes the `delay(10)` free, so this costs nothing in wall time.)

### Nits

- `java.util.concurrent.atomic.AtomicBoolean` is referenced fully qualified inline; every other type in the file is imported. Same for the pre-existing `io.mockk.verify` style elsewhere in the suite — worth keeping the file consistent.
- Alternative worth a sentence in the PR body: the check-then-insert could instead be made atomic at the DB layer via `AppDatabase.withTransaction`, which would hold even if the binding's scope ever changed. The mutex is simpler and adequate given `@Singleton`; just noting the choice was a choice.
- `e.printStackTrace()` survives the move. Not this PR's job — #16471 and #16388 are converging on tagged `Log.e` for repository catch blocks — but this is one of the sites those will want.

**Verdict:** needs a change. The production diff is fine as-is; the ask is a test that fails without the mutex. Once the new test actually interleaves, this goes straight back to `review` in the high 80s.

---

dogi:

---

dogi:
Thanks @dogi — addressed in 6f2a366.

**Discriminating test.** The concurrent test now uses a `StandardTestDispatcher` sharing `runTest`'s scheduler instead of `UnconfinedTestDispatcher`, so the two `async` coroutines are queued rather than run eagerly. The yield point is the `insertAll` mock (`coAnswers { delay(10); inserted.set(true) }`): both coroutines reach and pass the `count() == 0` check before either completes the insert, so deleting the mutex makes both insert. I verified this directly — with the mutex removed the test fails with `expected:<1> but was:<2>` (two inserts); with it restored, all 5 tests pass. The `delay` runs on the shared virtual clock, so it costs no wall time.

(One note on the suggested placement: putting the `delay` in `count()` rather than `insertAll` doesn't actually discriminate — the `if (inserted.get())` is evaluated *after* the delay resumes, so the second coroutine sees the first's insert and short-circuits to `AlreadyPopulated` even without the mutex. Moving the suspension point to `insertAll` is what opens the window where both observe `count == 0`.)

**Nits.** `AtomicBoolean` is now imported instead of fully-qualified, and a comment on `seedMutex` records the `@Singleton` dependency (an unscoped binding would hand each injection point its own mutex and silently re-open the race).

The `AppDatabase.withTransaction` alternative is noted; the per-instance mutex stays simpler and is adequate given the `@Singleton` binding. `e.printStackTrace()` is left for #16471 / #16388.

_This comment was created by an AI agent (OpenHands) on behalf of the repository maintainer._

dogi:
## Review: 88/100 — mergeable

Well-executed fix, and the concurrency test is the best one I've seen in this queue.

### The load-bearing assumption is documented *and* true ✅

The comment stakes correctness on the binding being `@Singleton`, so I checked: `RepositoryModule:84-86` is

```kotlin
@Binds
@Singleton
abstract fun bindDictionaryRepository(impl: DictionaryRepositoryImpl): DictionaryRepository
```

Confirmed. Writing that dependency down rather than leaving it implicit is exactly right — it's the kind of thing that gets silently broken by an unrelated DI cleanup.

**One gap in it, though:** the scope is on the `@Binds` *method*, not the class. Anyone injecting the concrete `DictionaryRepositoryImpl` instead of the `DictionaryRepository` interface gets a fresh unscoped instance with its own mutex, and the race quietly returns. Only `DictionaryActivity:85` injects it today, and it uses the interface. Adding `@Singleton` to the class itself would make the guarantee independent of how it's injected — cheap belt-and-braces for something the comment already identifies as fragile.

### The bug is real and the damage is permanent

Only one caller, so the window is narrow — but it's reachable: back out of `DictionaryActivity` while it's seeding (a file read, a full Gson parse, and a bulk insert of the whole dictionary — seconds, not milliseconds), re-enter, and two seeds overlap.

The consequence is worse than a wasted insert. Entities are built with `id = UUID.randomUUID().toString()`, so a second seed can't collide with the first — `insertAll` happily writes a **complete duplicate dictionary**, and every subsequent lookup returns every word twice. It's persistent, and nothing in the app would clean it up short of the destructive-migration resync. That's a solid justification for a mutex on a path that looks, at a glance, single-threaded.

### The test genuinely fails on `master`

The construction is careful in the ways that matter:

- A `StandardTestDispatcher` (rather than the file's default `UnconfinedTestDispatcher`) so the two `async` coroutines are *queued* and can interleave instead of running to completion eagerly — the comment calls this out, which is the difference between a real concurrency test and one that accidentally serializes.
- `delay(10)` inside the stubbed `insertAll` as the explicit yield point, with `inserted` flipping only *after* it, so both coroutines observe `count() == 0` on `master` and both insert.
- Sharing `testScheduler` between `runTest` and the dispatcher means the delay costs no wall-clock time.
- Both `async` calls go through the **same** `concurrentRepository` instance, which is what makes it a test of the shared mutex rather than of two independent ones.

On `master` that yields two `Inserted` results and `insertAll` called twice, so `assertEquals(1, results.count { it is Inserted })` and `coVerify(exactly = 1)` both fail. Proper regression test.

- **The lock spans the whole load**, not just the check-and-insert: file read, Gson parse, and bulk insert all happen under it. That's necessary for correctness here, but it means a second concurrent caller blocks for the full seed duration rather than returning quickly. Fine for this call site (the activity is showing a loading state anyway) — worth knowing if `insertDictionaryData` ever gets called from a sync worker.
- `e.printStackTrace()` survives in the catch. Pre-existing and untouched by this diff, but it's the pattern #16595 is systematically replacing; a tagged `Log.e` here would fit that direction.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16480, opened 2026-08-27
all: smoother feedback messages caching (fixes #16473)

### LINKED ISSUE #16473: Cache the parsed messages JsonArray in the Feedback model
**rating:** 50/100 &nbsp;·&nbsp; **proposed by:** devin(r2)

`messageList` and `message` re-parse the messages JSON string on every access while feedback list/detail adapters read them repeatedly during bind; `messages` is only mutated through setMessages, so an @Ignore transient cache is safe for Room.

**Ask:** add an `@Ignore` transient cache for the parsed messages `JsonArray`, invalidated in `setMessages`, so `messageList`/`message` stop re-parsing the JSON string on every access.

**Files:** `model/Feedback.kt`

**Verification:** Verified the getters and setMessages exist and both guard on `ar.size() > 0`.

dogi:
## Summary

`messageList` and `message` re-parsed the `messages` JSON string on every access. Feedback list/detail adapters read them repeatedly during bind, so each bind re-ran `JsonParser.parseString`.

This adds an `@Ignore @Transient` cache (`cachedMessages`) for the parsed `JsonArray` in `model/Feedback.kt`. It is memoized lazily in `parsedMessages()` and invalidated whenever `messages` is reassigned — including through `setMessages`, which assigns the field and triggers the custom setter.

## Changes

- `model/Feedback.kt`
  - Added a custom setter on `messages` that clears `cachedMessages` on every assignment.
  - Added `@get:Ignore @Ignore @Transient private var cachedMessages: JsonArray?` and a `parsedMessages()` helper that returns the cached array, parsing only on first access.
  - `messageList` and `message` now read via `parsedMessages()` instead of calling `JsonParser.parseString` each time.
  - Kept the existing `ar.size() > 0` guards intact.

## Verification

- `./gradlew testDefaultDebugUnitTest --tests "org.ole.planet.myplanet.model.FeedbackTest"` — all 11 tests pass (0 failures, 0 errors).
- Added tests:
  - `testMessagesCachedAcrossAccesses` — verifies the parsed array is memoized and reused across repeated `messageList`/`message` reads (same instance, never re-parsed).
  - `testCacheInvalidatedOnMessagesChange` — verifies reassigning `messages` invalidates the cache so the next read reflects the new content.
  - `testCacheInvalidatedOnSetMessages` — verifies `setMessages(JsonArray)` invalidates the cache.

Fixes #16473.

---

@dogi can click here to [continue refining the PR](https://app.all-hands.dev/conversations/1b1974ce-b8fa-434e-b862-6c40fc004a9d)


dogi:
## Review: memoize the parsed `messages` `JsonArray` on `Feedback`

**Rating: 74/100** — the implementation is careful and correct, but the memoization is largely ineffective against the path it targets, and it puts mutable derived state on a Room `@Entity` to buy it.

### What's right

The invalidation is airtight, which is the part these caches usually get wrong. `messages` gets a custom setter that nulls `cachedMessages`, and the only other writer — `setMessages(JsonArray?)` — assigns through that same property (`this.messages = JsonUtils.gson.toJson(messages)`), so it invalidates too. I checked the repository writers: `FeedbackRepositoryImpl:68` and `:108` both go through `setMessages(JsonArray)`, and `:154`/`:157` construct fresh instances. There is no path that mutates the backing string behind the setter's back.

The cached array is also never handed out — `parsedMessages()` is private and both consumers (`message`, `messageList`) only read from it — so no caller can mutate the shared instance. And `@Transient` keeps it out of Gson (both `provideGson()`'s `excludeFieldsWithModifiers(TRANSIENT, …)` and the bare `JsonUtils.gson`, which excludes transient by default), while `@Ignore` keeps it out of Room — the green KSP build is real evidence there, since an un-ignored `JsonArray` field would have failed the Room processor outright.

### Why it doesn't buy much

`Feedback.message` and `Feedback.messageList` have exactly one consumer in the app — `FeedbackDetailActivity:57-60`:

```kotlin
collectLatestWhenStarted(viewModel.feedback) { fb ->
    fb?.let {
        …
        activityFeedbackDetailBinding.tvMessage.text =
            if (TextUtils.isEmpty(it.message)) "N/A" else it.message   // 2 parses
        replyAdapter?.submitList(it.messageList)                        // 1 parse
```

`FeedbackAdapter` doesn't touch either property; `FeedbackReplyAdapter:25` reads `FeedbackReply.message`, a different class. So the entire hot path is three parses of a small JSON array per emission of a single-row flow.

And crucially, the memo is **per instance**: Room hands back a freshly constructed `Feedback` on every re-query, so each new emission starts with a cold cache regardless. The cache only helps *within* one emission (3 parses → 1) and across a `repeatOnLifecycle` re-subscribe that replays the same instance. That's a real but very small win, and the PR body reads as though it eliminates repeated parsing across the screen's lifetime, which it doesn't.

The equivalent of the dominant saving is one line at the call site:

```kotlin
val msg = it.message
activityFeedbackDetailBinding.tvMessage.text = if (TextUtils.isEmpty(msg)) "N/A" else msg
```

That removes the duplicate parse with no entity-level state, no custom property setter, and no reflection in tests.

### The cost side

- The triple annotation `@get:Ignore @Ignore @Transient` on a *private* property is doing more than it needs to (the private property has no meaningful public getter to target) and is the kind of incantation that gets copied without understanding. If the field stays, one `@Ignore @Transient` with a comment explaining why both are needed would age better.
- `cachedMessages` is unsynchronized. Today `Feedback` instances look confined to the collecting coroutine, so this is fine — but that's an invariant nothing states, and entities do get passed around.
- The codebase now has two different memoization idioms on entities: `News` guards with `rawViewIn != viewIn` re-check (`VoicesAdapter:610`), this one uses a setter hook. Worth picking one if this pattern is going to spread.
- Both new cache tests reach into a private field by reflection. That works, but it welds the tests to the field name — a rename breaks them for no behavioural reason. `assertSame` on the objects returned by two consecutive `messageList` reads would test the same property through the public surface.

### What would change my mind

If there's a profile or a trace showing this parse actually costs something — a feedback thread with hundreds of replies, or a consumer I've missed — say so and I'll happily move this back; the implementation is good enough to merge on that basis. As it stands the diff adds persistent-entity state and reflective tests to save two `JsonParser` calls per screen open.

**Verdict:** needs a change. Either land the one-line call-site hoist instead, or justify the cache with a real measurement.

---

dogi:

---

dogi:
Followed up on @dogi's review (74/100, "needs a change"). I couldn't find any inline reviews from @Okuro3499 on this PR, so I worked through @dogi's points. Pushed in `a7277bc`:

- **Call-site hoist (the "dominant saving").** `FeedbackDetailActivity:56-58` was reading `it.message` twice — once for `TextUtils.isEmpty(...)` and once for the value — so the single hot-path consumer parsed the backing string twice per emission. Hoisted it into a local `val message`, so it parses once. No entity-level state needed for that win.
- **Dropped the redundant `@get:Ignore`.** `cachedMessages` is `private`, so it has no public getter for `@get:Ignore` to target — it was an incantation that would get copied without understanding. Reduced it to `@Ignore @Transient`, matching the existing convention on `Submission`/`MyCourse`. Added a short comment explaining why *both* annotations are needed (`@Ignore` → Room, `@Transient` → Gson) and that the cache is confined to the owning coroutine, so it needs no synchronization.
- **Tests no longer reach into the private field by reflection.** `messageList` allocates a fresh list on every read and `message` returns a `String`, so there's no public instance to `assertSame` on (the literal suggestion wouldn't hold). Instead `testMessagesCachedAcrossAccesses` now verifies through the public surface that the backing string is parsed **exactly once** across many `message`/`messageList` reads (`mockkStatic(JsonParser)` + `verify(exactly = 1)`). A rename of the private field no longer breaks the test.

Kept the `@Ignore @Transient` cache itself, since the issue (#16473) explicitly asked for it and it does give a real — if small — win (3 parses → 1 per emission, and replay-safe across a `repeatOnLifecycle` resubscribe). The invalidation tests (`testCacheInvalidatedOnMessagesChange`, `testCacheInvalidatedOnSetMessages`) are unchanged; they already exercise the public surface.

The "two memoization idioms across entities" note I'm deliberately leaving out of this PR's scope — it's a cross-cutting convention decision better made separately than bolted onto a single cache.

Couldn't run the unit tests locally (no JDK/Android SDK in this environment); relying on the PR's `test.yml` CI to confirm `FeedbackTest` stays green.

---
_Generated by an AI agent (OpenHands) on behalf of the maintainer._

dogi:
## Review: 86/100 — mergeable

The invalidation gap from my earlier pass is closed, and closed in the right place.

### Invalidation is complete ✅

Routing it through the property setter is what makes this safe:

```kotlin
var messages: String? = null
    set(value) { field = value; cachedMessages = null }
```

Both write paths land there — direct assignment, and `setMessages(JsonArray?)`, which assigns `this.messages` and therefore trips the same setter. Room writes through the generated `setMessages(String)` (a distinct JVM overload from `setMessages(JsonArray)`), so hydration from the database also clears the cache. There's no way to change `messages` that bypasses invalidation, which is exactly the property that was missing before.

### Both annotations are needed, and both are present ✅

`@Ignore` keeps `cachedMessages` out of Room's schema; `@Transient` keeps it out of Gson — and that second one is load-bearing, since `provideGson()` uses `excludeFieldsWithModifiers(Modifier.FINAL, Modifier.TRANSIENT, Modifier.STATIC)` and `JsonUtils.gson` is a plain `Gson()` which also skips transient fields. Without `@Transient` the cache would have started serializing into upload payloads.

### The tests earn their keep

`testMessagesCachedAcrossAccesses` stubs `JsonParser.parseString` and asserts `verify(exactly = 1)` across twelve reads spanning **both** derived views. On `master` that's twelve parses, so it fails there — a real regression test, not a guard. The two invalidation tests likewise fail without the custom setter. Nice detail that the fixture array has two elements while `messageList` asserts size 1, which correctly reflects the `for (i in 1 until ar.size())` skip of the first entry.

**1. The thread-confinement claim is an assumption, not an invariant.** The comment says the instance "is confined to the collecting coroutine that owns this entity." That holds today because Room hands out fresh objects per query, so instances aren't shared. But nothing enforces it — if a `Feedback` were ever handed to the upload path while the UI reads it, `cachedMessages` is a non-volatile field written without synchronization. The damage would be mild (a redundant parse; reference writes are atomic on ART), so I wouldn't add a lock — just be aware the comment is describing current usage rather than something the type guarantees.

**2. `parsedMessages()` returns the cached instance, not a copy.** Both call sites are read-only today, and it's `private`, so it's contained. Worth flagging only because this codebase just reasoned about the same hazard in the opposite direction: `Achievement.parseStringListToJsonArray` deliberately `deepCopy()`s each cached `JsonElement` before handing it out, precisely so a caller's mutation can't corrupt the shared cache (#16385 even added a test pinning that). If a future caller starts mutating the returned array here, the cache and the `messages` string silently diverge. A one-line comment on `parsedMessages()` saying "callers must not mutate" would be enough.

**Small note:** the `val message = it.message` hoist in `FeedbackDetailActivity` is now redundant — with the cache in place the second read is free — but it reads more clearly, so no objection.

**Contention:** #16554 also modifies `Feedback.kt`. Worth a re-check on whichever lands second.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16481, opened 2026-08-27
sync: smoother time logger starting (fixes #16462)

### LINKED ISSUE #16462: Guard the SyncTimeLogger hot log calls with Log.isLoggable
**rating:** 54/100 &nbsp;·&nbsp; **proposed by:** devin(r2)

`Log.d("SyncPerf", ...)` is called on every sync step with interpolated strings that invoke formatElapsed/formatTime and String.format even when the tag is disabled, so release builds pay the allocation for output never displayed.

**Ask:** guard the hot `Log.d("SyncPerf", ...)` calls with `Log.isLoggable` so the string interpolation and formatting work is skipped when the tag is disabled.

**Files:** `utils/SyncTimeLogger.kt`

**Verification:** The SyncPerf Log.d calls are confirmed present (also visible in TransactionSyncManager). Shares a file with the flatten task — sequence them.

dogi:
Guarded hot `Log.d("SyncPerf", ...)` logging calls in `SyncTimeLogger.kt` with `Log.isLoggable("SyncPerf", Log.DEBUG)` checks and moved expensive string interpolations and formatting calculations inside the blocks so release builds skip work when the log tag is disabled.

Fixes #16462

---
*PR created automatically by Jules for task [14563631510091977101](https://jules.google.com/task/14563631510091977101) started by @dogi*


dogi:
## Review: guard `SyncTimeLogger` logcat writes with `Log.isLoggable`

**Rating: 84/100** — this targets a genuinely hot path and the data-collection side is untouched. One consequence needs to be stated out loud before it merges, because it isn't obvious from the diff.

### The path really is hot

`isLogging` is set unconditionally by `SyncManager:137` on every sync — there is no build-type or preference gate on it. So today, on **release** devices, every API call and every DB operation formats an elapsed timestamp, a duration, an emoji, and an endpoint into a string and writes it to logcat. `logApiCall` and `logDbOperation` fire once per request and once per batch write across a full sync; that's thousands of `formatElapsed`/`formatTime` calls and string concatenations per sync, all discarded by anyone not watching logcat. Guarding them is worth doing.

### The consequence to state: these logs are now off by default

`Log.isLoggable(tag, DEBUG)` returns true only when the tag's level is at or below DEBUG, and Android's **default level is INFO**. `Log.d` itself is not gated by anything — it always writes. So this diff doesn't just skip string building when nobody's listening; it turns SyncPerf from *always emitted* into *never emitted*, on debug builds included, until someone runs:

```
adb shell setprop log.tag.SyncPerf DEBUG
```

That's the standard Android idiom and it may well be exactly what you want — but the PR title reads "guard hot log calls", which sounds like a pure optimisation, and the next person to debug a sync will find the trace gone with no clue why. Two ways to close that:

1. Put the `setprop` line in the PR body and in a comment above the first guard, so it's discoverable from the code.
2. If debug builds should keep the trace unconditionally, use `BuildConfig.DEBUG || Log.isLoggable("SyncPerf", Log.DEBUG)` — you keep the release-build saving, developers keep the output, and `setprop` still works for release debugging.

I'd take (2) plus the comment from (1).

Worth noting the diagnostics that actually persist are unaffected: `saveSummaryToRoom`, `detailedLogs`, `apiCallTimes`, `dbOperationTimes`, `processTimes` and both `AtomicInteger` counters all stay outside the guards. `apiCallCounter.incrementAndGet()` and `dbOpCounter.incrementAndGet()` in particular are correctly left un-guarded — moving those inside would have silently corrupted `#N` numbering and the summary. That's the mistake this class of change usually makes, and this PR doesn't make it.

### Smaller points

- `"SyncPerf"` appears as a literal at 12+ sites and now in the guards too. `private const val TAG = "SyncPerf"` would make it impossible for a guard and its `Log.d` to drift apart onto different tags — which would silently re-enable or permanently disable a block.
- `logDetail` moves `timeProvider.now()` inside the guard. Correct (the value only fed the log line), and `now()` is pure so there's no side effect to lose — just flagging it as a real behavioural change hidden in a "move the local" hunk.
- Each guard re-queries the tag level rather than caching it. That's deliberate and right — caching would break live `setprop` toggling — but it's worth a word in the PR body so nobody "optimises" it into a `val` later.
- This introduces `Log.isLoggable` to the codebase; nothing else uses it today. If the pattern is going to spread (and there's a lot of unconditional `Log.d` around), a tiny helper in `utils/` would keep the idiom consistent rather than hand-rolled per file.

### Coordination

#16487 also rewrites `SyncTimeLogger` (mistake/summary flattening). Different regions, but whichever lands second wants a base merge and a CI re-run rather than a clean auto-merge.

**Verdict:** mergeable. The off-by-default behaviour is a legitimate choice — just make it a stated one rather than a discovered one.

---

Okuro3499 (approved):

PR #16486, opened 2026-08-27
sync: smoother upload coordinator retrying (fixes #16464)

### LINKED ISSUE #16464: Build the retry-queue payload map only when a retryable failure exists
**rating:** 53/100 &nbsp;·&nbsp; **proposed by:** deepseek(r3)

`queueRetryableFailures` (:266) and `queueRetryableFailuresRoom` (:471) both build `preparedUploads.associateBy { it.localId }` over every prepared upload before checking whether any failure is retryable — in the common error-free case the map is allocated and never used.

**Ask:** move the `associateBy { it.localId }` map construction below the retryable-failure guard in both methods, so it is only built when there is actually a retryable failure to queue.

**Files:** `services/upload/UploadCoordinator.kt`

**Verification:** Verified both associateBy lines sit above their `errors.filter { it.retryable }.forEach` guards.

dogi:
## Summary

`UploadCoordinator.queueRetryableFailures` and `queueRetryableFailuresRoom` both built `preparedUploads.associateBy { it.localId }` over every prepared upload *before* checking whether any failure was retryable. In the common error-free case (or when every failure is non-retryable), the map was allocated and never used.

This moves the `associateBy` map construction below the retryable-failure guard in both methods, so it is only built when there is actually a retryable failure to queue. The filtered retryable errors are computed once and reused for both the early return and the loop.

## Files changed

- `services/upload/UploadCoordinator.kt` — moved `payloadMap` allocation below the `errors.filter { it.retryable }` guard in `queueRetryableFailures` and `queueRetryableFailuresRoom`; added an early return when there are no retryable failures.

## Tests

Added `UploadCoordinatorTest` covering both the legacy (`upload`) and Room (`uploadRoom`) retry paths:

- all uploads succeed → retry queue **not** invoked
- failures are non-retryable (HTTP 404) → retry queue **not** invoked
- failures are retryable (HTTP 500) → retry queue invoked once per item with the prepared payload

All new tests pass (`./gradlew testDefaultDebugUnitTest --tests "...UploadCoordinatorTest"`), and the existing `UploadManagerTest` / `UploadConfigsTest` suites continue to pass.

Fixes #16464.

---

@dogi can click here to [continue refining the PR](https://app.all-hands.dev/conversations/c925a78c-a395-4acd-a91d-fd832863df4f)


dogi:
## Review: build the retry-queue payload map only when a retryable failure exists

**Rating: 87/100** — the production change is trivially safe, and the far larger contribution is that `UploadCoordinator` now has a test file at all.

### Correctness

`errors.filter { it.retryable }` is hoisted out of the statement and an early return added when it's empty. When the filtered list is empty the old `forEach` was a no-op, so the only observable difference is that `preparedUploads.associateBy { it.localId }` no longer runs — and its result was reachable only from inside that `forEach`. Nothing else in either function consumes `payloadMap`. Behaviour-identical.

The saving is one HashMap build over a full batch on the happy path of every upload, in both the Room and legacy paths. Small per call, but it's the common case and the change costs nothing.

Worth noting for context: `RetryQueue.queueFailedOperation` already guards internally (`RetryQueue:36`, `if (!error.retryable) … return`), so the filter here was never load-bearing for correctness — it's purely about not doing the work. That makes this an unusually low-risk diff.

### The tests are the real value

`UploadCoordinator` is ~478 lines of orchestration and had no test file before this PR. Five tests now pin down the retry-queueing contract across both entry points, and the HTTP codes they use are grounded in the actual classifier rather than guessed — `UploadCoordinator:215` sets `retryable = response.code() >= 500`, so 404 → not queued and 500 → queued is exactly right, and the `exactly = 2` counts match the two-item batches. The named-argument `coVerify` block is a nice touch: it asserts `endpoint`, `httpMethod`, `dbId` and `modelClassName` are threaded through correctly, not just that *something* was queued.

### Nits

- At `:263`, `val uploadType = config.modelClass.simpleName ?: "Unknown"` is still computed above the new early return, so the guard skips the map build but not this. Trivial cost, but moving it below the guard would make the two functions symmetric — the sibling at `:471` has no such line stranded above its guard.
- The two functions are now near-identical: filter, guard, `associateBy`, `forEach { queueFailedOperation(...) }`. This PR makes that duplication more visible than it was. A single private `queueRetryableErrors(uploadType, endpoint, errors, preparedUploads)` would collapse both and guarantee they can't drift — worth doing while the code is fresh, though I wouldn't hold the PR for it.
- `retrofit2.Response`, `okhttp3.ResponseBody` and `UploadedItemResult`'s companions are referenced fully-qualified inline rather than imported, which reads against the rest of the suite.
- `okhttp3.ResponseBody.create(null, "error")` is the deprecated static overload; `"error".toResponseBody(null)` is the OkHttp 5 form. It compiles, but `build.yml` runs with `--warning-mode all`, so this adds noise to every build log.

### Coordination

#16495 (parallelize `UploadCoordinator` batch HTTP uploads) rewrites large parts of this same file and is in the same review batch. These two will conflict textually even though they're independent in intent — whichever merges second needs its base merged in and CI re-run. Given this one is a three-line diff and #16495 is structural, landing this first is the cheaper order.

**Verdict:** mergeable.

---

Okuro3499 (approved):

dogi:
Superseded by #16495, which now carries this change.

Both PRs added `app/src/test/java/.../upload/UploadCoordinatorTest.kt`. Two branches creating the same new path is an add/add merge: git has no common ancestor for the file, so it merges against an empty base and every difference conflicts. Restructuring the file contents does not avoid it — an identical scaffold with the tests in disjoint regions still conflicts, and extra shared context between them only splits it into more hunks. Whichever of the two merged second would have been retired as `conflict`.

The source changes never overlapped (`queueRetryableFailures` here vs `uploadBatch` there), so the test file was the only real collision.

Carried over verbatim into #16495 as commit 00cd97d:

- the early return in `queueRetryableFailures` / `queueRetryableFailuresRoom`, so `preparedUploads.associateBy { it.localId }` is built only when there is a retryable failure to queue
- all five retry-queue tests (no retry when all succeed, no retry on 404, one retry per item on 500, across both the legacy and Room paths), adapted to that file's per-test coordinator construction and reusing its fixtures instead of duplicating them

`Fixes #16464` has moved to #16495, so the issue still closes on merge. Nothing is lost by closing this.

---

PR #16487, opened 2026-08-27
sync: smoother time logger summary generating (fixes #16461)

### LINKED ISSUE #16461: Hoist the mistake/summary flattening in SyncTimeLogger
**rating:** 54/100 &nbsp;·&nbsp; **proposed by:** codex(r3), deepseek(r2), deepseek(r3)

`buildSummary` flattens apiCallTimes six times (:245, :246, :266, :267, :286, :289), allocating a throwaway List over every logged entry each time, to compute scalars at the end of every sync.

**Ask:** flatten `apiCallTimes` once into a local and reuse it across all six sites in `buildSummary`, instead of re-flattening per scalar.

**Files:** `utils/SyncTimeLogger.kt`

**Verification:** Verified all six flatten() calls. deepseek(r3) covers all six; deepseek(r2) covers two; codex covers the generateSummary pair. deepseek proposed it in two rounds — counted once.

dogi:
Flatten apiCallTimes and dbOperationTimes once into local variables in generateSummary() and reuse them across all scalar calculations to avoid repeated throwaway list allocations. Also add unit tests for generateSummary().

Fixes #16461

---
*PR created automatically by Jules for task [13182967188167027853](https://jules.google.com/task/13182967188167027853) started by @dogi*


dogi:
## Review: hoist the mistake/summary flattening in `SyncTimeLogger`

**Rating: 86/100** — correct, and it quietly makes the summary *more* internally consistent, not just cheaper.

### Correctness

`apiCallTimes.values.flatten()` went from three evaluations (`totalApiTime`, `successfulCalls`, `apiPercentage`) to one; `dbOperationTimes.values.flatten()` likewise (`totalDbTime`, `totalDbItems`, `dbPercentage`). `flatten()` on a `ConcurrentHashMap`'s values produces a fresh `List` each time, so this replaces six list allocations — each proportional to the total number of API calls and DB ops in the sync — with two.

There's a subtler benefit worth naming, because it's the argument for the change beyond allocation count: all three API-derived statistics now read from **one snapshot**. Before, `successfulCalls` and `totalApiTime` each took their own snapshot, so a late `logApiCall` landing between them could produce a summary where the success count and the time total described different sets of calls. `isLogging = false` is set in `stopLogging` before `generateSummary` runs, so that race is closed anyway today — but the hoisted form is structurally immune rather than incidentally safe.

Moving the flatten above the `if (apiCallTimes.isNotEmpty())` guard means an empty map is now flattened on a sync with no API calls. That's an empty-list allocation on a path that's about to build a multi-kilobyte string — not worth a thought.

### Follow the hoist one step further

`totalApiCalls` and `totalDbOps` were left behind:

```kotlin
val totalApiCalls = apiCallTimes.values.sumOf { it.size }   // still a separate traversal
val totalApiTime  = allApiCallLogs.sumOf { it.duration }
```

`allApiCallLogs.size` and `allDbOpLogs.size` are exactly equal to those sums by construction, and using them would (a) drop the last redundant traversal and (b) make the count and the time provably describe the same list — which is the property the rest of this diff is establishing. Same two lines in the DB block.

### Test

Good test, and the arithmetic is checkable by hand: `startTime = 1000`, `endTime = 2000` → 1000 ms total; API `300 + 200 = 500` → 50.0%; DB `400` → 40.0%; other 10.0%. The success/failure split (`2 (Success: 1, Failed: 1)`) exercises the `successfulCalls` path that this diff touches, and `Total items processed: 5` covers `totalDbItems`. That's real coverage of the hoisted expressions rather than a smoke test.

Two things about it:

- **No `@After unmockkAll()`.** `mockkStatic(Log::class)` is installed and never torn down, so the static mock leaks to whatever runs next in the same fork — including `testExtractProcessName` in this very class. Every other MockK-using test in this suite pairs `mockkStatic`/`mockkObject` with an `@After`. Worth adding before this becomes someone else's flaky-test hunt, especially with CI running two shards.
- Only `Log.d(any(), any())` is stubbed. That's sufficient today (the exercised paths only reach `Log.d`), but an unstubbed static call under `mockkStatic` throws rather than no-ops, so any future log line at a different level inside `stopLogging` will fail this test with a confusing error. Stubbing `Log.e`/`Log.w` alongside would make it robust.

### On widening `generateSummary` to `internal`

This is a visibility relaxation purely to create a test seam. I'd accept it — `internal` is still module-scoped, the function is pure with respect to the logger's collected state, and asserting on the produced string is genuinely the right way to test it. Just noting it's a deliberate trade rather than an incidental edit, since `private` → `internal` on a class member tends to get copied as precedent.

### Coordination

#16481 (`Log.isLoggable` guards) rewrites `startLogging`, `stopLogging`, `logApiCall`, `logDbOperation` and `logDetail` in this same file. Different regions from this diff — `generateSummary` is untouched there — so a merge should resolve cleanly, but whichever lands second still wants its base merged and CI re-run.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16489, opened 2026-08-27
all: smoother crash log store caching (fixes #16463)

### LINKED ISSUE #16463: Reuse the shared filename parse in CrashLogStore
**rating:** 53/100 &nbsp;·&nbsp; **proposed by:** deepseek(r2)

`isValidLogFile` and `loadPendingLogs` each parse the same `<epoch>_<type>.log` shape independently (removeSuffix / indexOf('_') / substring), so every load re-derives fields validation just computed on the same iteration.

**Ask:** extract the `<epoch>_<type>.log` filename parse into one shared helper that returns the parsed fields, and have both `isValidLogFile` and `loadPendingLogs` use it instead of each parsing independently.

**Files:** `utils/CrashLogStore.kt`

**Verification:** File exists; the duplicated parse was not line-verified.

dogi:
Extracted filename parsing logic into parseLogFile helper in CrashLogStore and updated isValidLogFile and loadPendingLogs to avoid duplicate parsing operations.

Fixes #16463

---
*PR created automatically by Jules for task [2697094455050800770](https://jules.google.com/task/2697094455050800770) started by @dogi*


dogi:
## Review: reuse the shared filename parse in `CrashLogStore`

**Rating: 89/100** — removes a genuine double-parse and tightens the exception scope as a side effect.

### Correctness

`loadPendingLogs` previously parsed each filename twice: once in `filter { isValidLogFile(it) }` and again inside the `mapNotNull` body to recover `time` and `type`. Now `parseLogFile` runs once per file and returns both. Same set of files accepted, same values extracted.

The field mapping is the trap in this refactor and the PR steps around it: `ParsedLogInfo` declares `(time, type)` while `PendingLog` takes `(file, type, time, error)` — inverted. The call site uses `PendingLog(file, parsed.type, parsed.time, file.readText())`, which is correct. Worth reordering `ParsedLogInfo` to `(type, time)` anyway so the two records read in the same order and a future positional construction can't silently swap them.

The parsing guards are preserved exactly: `separator <= 0` still rejects a name with no `_` and one that starts with `_` (so `substring(0, separator)` can't be empty), and `toLongOrNull()` still gates the timestamp. A name like `123_.log` yields `type = ""` today and still does — unchanged, if slightly odd.

### An improvement the PR doesn't claim

Pulling the parse out of the `try` narrows the catch to `file.readText()` — the only expression in that block that can actually throw. Before, a hypothetical `substring` failure would have been swallowed into `printStackTrace()` and dropped; now filename parsing is total (returns `null`) and the `catch` covers only I/O. That's a real robustness gain, and it's the kind of thing worth putting in the PR body since it's easy to read the diff as pure deduplication.

### One nit on the `save()` path

`isValidLogFile` survives — correctly, it's still used at `:33` for the `MAX_PENDING_FILES` cap — but it's now `parseLogFile(file) != null`, so counting pending files allocates a `ParsedLogInfo` (and does an extra `substring` for `type`) per file, only to discard it.

That path matters a little more than it looks. This file's own doc comment says reports are written "before any coroutine or Room machinery runs, because app shutdown can happen before launched coroutines finish persisting rows" — it's the crash/ANR handler, deliberately dependency-free. In absolute terms it's ≤20 tiny short-lived allocations (`MAX_PENDING_FILES = 20`), so I wouldn't hold the PR for it. But if you want the guard to stay allocation-free, keeping `isValidLogFile` as its own cheap predicate and having `parseLogFile` be the richer variant costs one small duplicated check and keeps the crash path exactly as lean as it was.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16495, opened 2026-08-27
sync: smoother upload coordinating (fixes #16467)(fixes #16464)

### LINKED ISSUE #16464: Build the retry-queue payload map only when a retryable failure exists

### LINKED ISSUE #16467: Parallelize the UploadCoordinator batch HTTP uploads
**rating:** 52/100 &nbsp;·&nbsp; **proposed by:** copilot(r2)

`uploadBatch` and `uploadBatchRoom` walk `batch.forEach` and await each postUpload/putUpload sequentially, where network latency dominates; ActivitiesRepositoryImpl already uses Semaphore(6)+async for the same pattern.

**Ask:** parallelize the per-item uploads in `uploadBatch`/`uploadBatchRoom` using the `Semaphore(6)` + `async` pattern already used in `ActivitiesRepositoryImpl`. This is the highest-risk task on the list — the 409 recovery path, cancellation, and result ordering must all survive the change, so preserve them explicitly and test them.

**Files:** `services/upload/UploadCoordinator.kt`

**Verification:** Verified the sequential `batch.forEach` at :146 and :395. Higher risk than the rest of this list: the 409 recovery path, cancellation and result ordering all have to survive.

dogi:
Parallelized per-item HTTP uploads in UploadCoordinator batch processing using Semaphore(6) and coroutineScope + async/awaitAll.

This PR now also carries the retry-queue guard that was previously #16486, so both `UploadCoordinator` optimisations ship as one unit.

## Why they were merged into one PR

Both changes added `app/src/test/java/.../upload/UploadCoordinatorTest.kt`. Two branches creating the same new path is an **add/add** merge: git has no common ancestor for the file, so it merges against an empty base and every difference between the two versions conflicts. That is not fixable by rearranging the contents — an identical scaffold with the new tests in disjoint regions still conflicts, and adding shared context between them only splits it into more hunks. Whichever PR merged second would have been retired as `conflict`.

The source changes themselves never overlapped (`uploadBatch` vs `queueRetryableFailures`), so only the test file forced the decision.

## Changes

**`services/upload/UploadCoordinator.kt`**
- per-item HTTP uploads in `uploadBatch` / `uploadBatchRoom` run concurrently under a `Semaphore(6)`, preserving 409 conflict recovery, IOException retryability, cancellation and result ordering
- `queueRetryableFailures` / `queueRetryableFailuresRoom` filter retryable errors once and return early when there are none, so `preparedUploads.associateBy { it.localId }` is only built when there is actually a retryable failure to queue. Previously that map was allocated on every batch, including the common error-free case.

**`UploadCoordinatorTest.kt`** — 12 tests:
- bounded concurrency and result ordering, on both the legacy and Room paths
- 409 conflict recovery, and IOException during recovery treated as retryable
- cancellation propagation
- retry queue not invoked when all uploads succeed, and when failures are non-retryable (404)
- retry queue invoked once per item for retryable failures (500), on both paths

Shared fixtures are reused across both groups rather than duplicated.

## Verification

Full unit suite run locally against master merged in: **213 classes, 1529 tests, 0 failures** attributable to this change, with all 12 `UploadCoordinatorTest` cases passing.

One unrelated pre-existing flake was seen once in that run and did not reproduce: `DispatcherProviderIntegrationTest` failed with `UncaughtExceptionsBeforeTest`, which is coroutine-exception pollution leaking from an earlier test in the same fork. It passes in isolation, passes alongside `UploadCoordinatorTest`, and passes on a re-run of the identical tree. Worth chasing separately — it is a latent hazard in the suite, not in this diff.

Fixes #16467
Fixes #16464

Supersedes #16486.


dogi:
## Review: parallelize `UploadCoordinator` batch HTTP uploads

**Rating: 72/100** — the parallelization itself is built correctly, but there's an unremarked change to retry classification that loses a retry, and the headline concurrency test can't fail.

### The parallelization is sound

`coroutineScope { batch.map { async { … } }.awaitAll() }` is the right shape: `awaitAll` preserves list order, so `succeeded` still comes back in batch order (the tests assert this, correctly); a `BatchItemResult` sealed class replaces the two shared `MutableList`s, so there's no concurrent mutation left; `CancellationException` is rethrown in every branch, so `coroutineScope` still cancels siblings and propagates; and both entry points already run inside `withContext(dispatcherProvider.io)`, so the children inherit the IO pool rather than landing somewhere unexpected.

### The defect: 409-recovery failures stop being retryable in `uploadRoomBatch`

In the **old** `uploadRoomBatch`, the 409 branch had no inner `try`:

```kotlin
} else if (response.code() == 409) {
    val getResponse = uploadRepository.fetchExistingDoc(…)   // exceptions escape to the outer catch
```

so an `IOException` from that recovery fetch fell through to the function's outer `catch (e: IOException)` and produced `UploadError(…, retryable = true)`.

The **new** version wraps that branch in its own `try`:

```kotlin
} catch (e: Exception) {
    BatchItemResult.Error(UploadError(…, "Document exists (409) but fetch failed: …", retryable = false, httpCode = 409))
}
```

An `IOException` is an `Exception`, so a transient network blip during conflict recovery now yields **`retryable = false`** and `RetryQueue.queueFailedOperation` drops it (`RetryQueue:36`). The item isn't permanently lost — it stays unmarked locally and gets picked up by the next full sync — but it loses its place in the retry queue, and that's a real behavioural regression for the exact failure mode (flaky network) the retry queue exists to absorb.

I can see the argument that this *harmonises* the Room path with the legacy `uploadBatch`, which always had that inner catch. If that's the intent, fine — but then it belongs in the PR description, and I'd argue the legacy path is the one that's wrong, not the Room one. The safer resolution either way:

```kotlin
    throw e
} catch (e: IOException) {
    BatchItemResult.Error(UploadError(preparedItem.localId, e, retryable = true, httpCode = 409))
} catch (e: Exception) {
    BatchItemResult.Error(UploadError(…, retryable = false, httpCode = 409))
}
```

in **both** functions.

### The concurrency test can't fail

```kotlin
assertTrue("Max concurrent requests was $maxConcurrentRequests", maxConcurrentRequests in 1..6)
```

`1..6` is satisfied by `1` — i.e. by fully sequential execution. Delete the `async`, delete the `Semaphore`, revert to `batch.forEach`, and this assertion still passes. It's the one test that's supposed to prove the PR's thesis, and it asserts only the upper bound.

With 10 items, `delay(10)` inside the mock, and `Semaphore(6)`, the observed maximum is deterministically 6 under the test scheduler:

```kotlin
assertEquals("expected the semaphore to be saturated", 6, maxConcurrentRequests)
```

That fails on `master` and on any regression to sequential execution, which is what you want. (The unsynchronized `maxConcurrentRequests` read/write is fine here — everything runs on the single test dispatcher thread.)

### Design points

- **`Semaphore(6)` is per-call, and hardcoded twice.** `UploadCoordinator` is `@Singleton`, so if two upload types are ever driven concurrently, each batch gets its own 6 permits and the real ceiling is 6 × N. `NetworkModule` sets `MAX_REQUESTS_PER_HOST = 20`, so a few concurrent types would saturate OkHttp's per-host queue. A single injected/instance-level `Semaphore` with a named constant would bound it globally and stop the two literals drifting apart.
- **`beforeUpload`/`afterUpload` now run concurrently, and outside the permit.** No config in the codebase supplies either today (both default to `null` in `UploadConfig` and `RoomUploadConfig`, and nothing overrides them), so this is latent rather than live — but it's worth a comment on those fields, because the first config that adds an `afterUpload` writing to Room will get concurrent invocation with no indication from the signature.
- **`coroutineContext.ensureActive()` is now vestigial.** As the first statement inside each `async`, it checks once at coroutine start rather than between items. Cancellation still works via the suspending calls, so nothing is broken — but the line no longer does what it was put there to do, and reads as though it does.
- The two batch functions were already near-duplicates and this diff roughly doubles each. Now would be a good moment to extract the shared per-item body; as it stands the 409 fix above has to be applied twice, which is exactly how the two paths diverged in the first place.

### Conflict with #16486 — add/add on the same new file

#16486 also creates `app/src/test/java/org/ole/planet/myplanet/services/upload/UploadCoordinatorTest.kt`, from scratch, with different content. Two PRs adding the same new path is an add/add conflict git cannot resolve on its own, so whichever lands second will need the file merged by hand — the two test suites are complementary (retry-queueing contract vs. parallelism/ordering/409), so the merge should keep both sets of cases rather than picking one.

#16486 is a three-line production diff; landing it first and rebasing this one onto it is much cheaper than the reverse.

### Nits

- `okhttp3.ResponseBody.create(null, "")` is the deprecated overload (`"".toResponseBody(null)` is the OkHttp 5 form); `build.yml` runs `--warning-mode all`.
- `io.mockk.mockkObject` / `io.mockk.unmockkObject` are fully qualified inline while the rest of MockK is imported.
- `tearDown` calls both `unmockkAll()` and `unmockkObject(UrlUtils)`; the first already covers the second.

**Verdict:** needs a change. Restore retryability for `IOException` on the 409-recovery path (or state the harmonisation intent), and tighten the concurrency assertion to `assertEquals(6, …)`. The core rewrite is good work and I'd be happy to see it land right after.

---

dogi:
## Review: 84/100 — mergeable

Much stronger than the earlier version. The two things that make batch parallelisation dangerous — result ordering and error containment — are both handled, and there's now 370 lines of test where there was none.

### Ordering and containment ✅

- `batch.map { async { … } }.awaitAll()` returns in **list** order, not completion order. That matters because `succeeded`/`failed` feed `markUploaded` and the retry queue downstream. The test asserts `local-1..local-10` map to `remote-1..remote-10` explicitly, which is the right thing to pin.
- Every path inside the `async` returns a `BatchItemResult` rather than throwing, with `CancellationException` deliberately rethrown. So `coroutineScope` never cancels siblings on a single item's failure — per-item error collection survives the rewrite. That's the subtle one: `coroutineScope` + `async` propagates the *first* exception and cancels the rest, which would have silently converted "one bad item" into "lost batch".
- Cancellation is preserved and tested.

### It also fixes a real retryability bug

In `uploadBatch`, `master`'s 409-recovery block caught everything with one `catch (e: Exception)` and marked it `retryable = false`. So a transient network blip *during conflict recovery* permanently classified the item as non-retryable and it never re-entered the retry queue. The new inner `catch (e: IOException) → retryable = true` fixes that, and `upload 409 conflict recovery handles network IOException as retryable` pins it.

(The Room path reached the same `retryable = true` via its outer catch on `master`, so there it's not a behavior fix — but it now logs and carries `httpCode = 409`, which is better error metadata.)

### The concurrency contract change has no current caller — but should be documented

`config.beforeUpload` and `config.afterUpload` are invoked **outside** the semaphore, so with a batch of N all N callbacks can run concurrently, where `master` ran them strictly one at a time between requests. I grepped `UploadConfigs` and every other config site: **nothing sets either hook today** — both are `null` everywhere. So this is latent, not live.

It's still worth a line of KDoc on those two fields saying they may be invoked concurrently for items in the same batch. The next person to add an `afterUpload` that writes to Room or mutates a shared accumulator would otherwise inherit a race that no test would catch.

### Notes

- **`Semaphore(6)` is a duplicated magic number.** It appears identically in both methods with no name or rationale. Worth a `private const val MAX_CONCURRENT_UPLOADS = 6` — and worth reconciling with `NetworkModule.MAX_REQUESTS_PER_HOST = 20`, since OkHttp is already imposing a per-host ceiling. Right now there are two independent limits and neither references the other.
- **The duplication between `uploadBatch` and `uploadRoomBatch` got worse.** They were near-identical before; each is now ~100 lines of the same structure, and the 409 handling had to be fixed in both places (which is exactly how the two paths drifted in the first place). Extracting the shared per-item body would make the next fix a one-place change.
- The test's `maxConcurrentRequests` is a plain `var` written from several coroutines while `activeRequests` is an `AtomicInteger`. Safe here because a single-threaded `StandardTestDispatcher` confines them, but the mismatch invites a copy-paste into a multi-threaded context later.

**`assertEquals(6, maxConcurrentRequests)`** is a good assertion — it fails on `master` (max would be 1) *and* would fail if the semaphore were removed entirely, so it pins both the parallelism and its bound.

**Contention:** #16486 also touches `UploadCoordinator` and adds a file at the same path (`UploadCoordinatorTest.kt`) — that's an **add/add conflict**, not a clean merge. Whichever lands second will need its test file reconciled by hand.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16496, opened 2026-08-27
teams: smoother tasks notification worker testing (fixes #16451)

### LINKED ISSUE #16451: Collect valid notified task IDs during notification delivery
**rating:** 56/100 &nbsp;·&nbsp; **proposed by:** codex(r3)

`TaskNotificationWorker.doWork` iterates pending tasks to show notifications, then traverses them again with `mapNotNull { it.id }.filter { it.isNotBlank() }` to build the ids to mark — two extra collections and a second pass over the same list.

**Ask:** collect the valid task ids during the existing notification-delivery loop instead of re-traversing the list afterwards.

**Files:** `services/TaskNotificationWorker.kt`

**Verification:** Verified at TaskNotificationWorker.kt:46-58.

dogi:
Optimize TaskNotificationWorker to collect valid non-blank task IDs in a single pass during the notification delivery loop, avoiding a second pass over the tasks list and intermediate collection allocations. Added TaskNotificationWorkerTest to verify task ID collection behavior.

Fixes #16451

---
*PR created automatically by Jules for task [11627911660480366639](https://jules.google.com/task/11627911660480366639) started by @dogi*


dogi:
## Review: collect valid task IDs during notification delivery

**Rating: 88/100** — fuses a redundant second pass into the existing loop, and deletes a `mapNotNull` that was quietly lying about the model.

### Correctness

`TeamTask.id` is declared `var id: String = ""` — **non-null**. So the old line:

```kotlin
val taskIds = tasks.mapNotNull { it.id }.filter { it.isNotBlank() }
```

had a `mapNotNull` that could never filter anything, and it allocated two intermediate lists across two extra traversals. The replacement collects into a single `mutableListOf<String>()` inside the loop that was already iterating `tasks`, applying the same `isNotBlank()` guard.

Order is preserved (same iteration order, appended in sequence), no de-duplication existed before and none is introduced, and `markTasksNotified` is still skipped entirely when the list ends up empty. Behaviour-identical.

The removal of the phantom `mapNotNull` is worth as much as the traversal saving: it was signalling "`id` can be null" to every future reader of this worker, and it can't be.

### Tests

First test file for this worker, and the assertions are discriminating rather than decorative — `coVerify(exactly = 1) { teamsRepository.markTasksNotified(listOf("task_1", "task_3")) }` pins the exact list, so it catches both a dropped blank-filter and an order change, and the second test covers the all-blank case that must not call the repository at all. Both would fail if the filter were lost in a future edit.

### One gap worth closing

Fusing "deliver the notification" and "collect the id" into one loop creates a specific new failure mode: someone later reading `if (id.isNotBlank())` as a per-task guard and moving the notification call inside it, silently dropping notifications for blank-id tasks. Nothing in the suite would catch that — `task2` (blank id) should still get a notification, and no test asserts it does.

The fix is one line, but it needs the notification manager hoisted first, since it's currently an anonymous mock returned inline:

```kotlin
private val notificationManager: NotificationUtils = mockk(relaxed = true)
…
every { NotificationUtils.getInstance(any()) } returns notificationManager
…
verify(exactly = 3) { notificationManager.showNotification(any()) }   // all three, blank id included
```

As written, `every { … } returns mockk(relaxed = true)` hands back a fresh mock per call, so there's nothing to verify against.

### Nits

- The loop body now does two jobs. It's the right trade here (one traversal instead of three), but a one-line comment saying the blank check gates *only* the id collection, not delivery, would make the intent unmissable at the point where it matters.
- `TeamTask.title` is nullable while `id` isn't (`task.title.orEmpty()` two lines up). Not this PR's problem, just noting the model's nullability is uneven and that's what invited the bogus `mapNotNull` in the first place.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16497, opened 2026-08-27
courses: smoother exams questions initializing (fixes #16478)

### LINKED ISSUE #16478: Hoist the locale and fix the loop re-initialization in ExamQuestion
**rating:** 50/100 &nbsp;·&nbsp; **proposed by:** devin(r2)

`setCorrectChoiceArray` calls `Locale.getDefault()` on every iteration, and `insertCorrectChoice` resets correctChoiceList and re-parses the correctChoice array **inside** the choices loop, making array-style parsing O(n) instead of O(1). Both run on every exam sync.

**Ask:** hoist the `Locale.getDefault()` call out of the `setCorrectChoiceArray` iteration, and move the `correctChoiceList` reset + `correctChoice` array parse out of the choices loop in `insertCorrectChoice` so array-style parsing is O(1) instead of O(n).

**Files:** `model/ExamQuestion.kt`

**Verification:** File exists; the reset-inside-loop was not line-verified, but it is the more interesting half of the claim if it holds.

dogi:
Hoisted `Locale.getDefault()` call out of the loop in `setCorrectChoiceArray` and moved the `correctChoice` array check and parsing out of the `choices` loop in `insertCorrectChoice`.

Fixes #16478

---
*PR created automatically by Jules for task [8756492853024670307](https://jules.google.com/task/8756492853024670307) started by @dogi*


dogi:
## Review: hoist locale and fix loop re-initialization in `ExamQuestion`

**Rating: 82/100** — three separate improvements in one small diff, including a latent NPE guard the title doesn't mention. The gap is that nothing tests the control-flow change.

### 1. The loop restructure is the real content

Old `insertCorrectChoice`, for a question whose `correctChoice` is a JSON array, re-ran this **once per choice**:

```kotlin
myQuestion?.correctChoiceList = mutableListOf()
myQuestion?.setCorrectChoiceArray(JsonUtils.getJsonArray("correctChoice", question), myQuestion)
```

Each iteration reset the list to empty and refilled it from the same source, so the final value was correct but built N times — including N calls to `JsonUtils.getJsonArray` and N full rebuilds of the lowercased list. Hoisting the `isJsonArray` test out of the loop makes it once. Same end state, and the `else` branch keeps the original per-choice scan (still no `break`, so last match wins — unchanged).

### 2. `question.has("correctChoice")` closes a real NPE

This is worth calling out because the PR body doesn't. `JsonObject.get(key)` returns `null` for an absent key, so the old condition:

```kotlin
if (question["correctChoice"].isJsonArray)
```

throws NPE whenever `correctChoice` is missing. The call site makes that shape reachable:

```kotlin
val isMultipleChoice = type?.startsWith("select") == true && question.has("choices")
if (isMultipleChoice) insertCorrectChoice(question["choices"].asJsonArray, question, this)
```

The guard is `type` starts with `select` **and** `choices` exists — nothing requires `correctChoice`. A survey question is exactly that shape: `select`-typed, with choices, and no correct answer. Whether it fires today depends on whether the Planet server always emits a `correctChoice` field for survey questions (it may, which would be why nobody has hit this), but the parser accepts documents it would crash on, and after this PR it doesn't. That's the most valuable line in the diff.

### 3. Locale hoist

`Locale.getDefault()` out of the append loop in `setCorrectChoiceArray`. Cheap either way, but free and correct — and it also makes the loop use one consistent locale rather than re-reading a mutable global per iteration.

### One behaviour change to state

When `choices` is an **empty** array and `correctChoice` **is** a JSON array, the two versions differ:

- **Before:** the `for` never ran, so `correctChoiceList` was left untouched (null).
- **After:** the array branch runs unconditionally, so `correctChoiceList` gets populated from `correctChoice`.

I think the new behaviour is the right one — the set of correct answers shouldn't depend on the choices list being non-empty — but it is a change, it's reachable (a `select` question with `"choices": []`), and it isn't mentioned. Worth a line in the PR body.

### The gap: no tests

`insertExamQuestions` / `insertCorrectChoice` are pure static functions over `JsonObject`, with no Android or Room dependency — about as cheap to test as anything in this repo. This PR changes their control flow in three ways and adds none. Four cases would lock all of it down:

- `correctChoice` as an array with 3 choices → `correctChoiceList` matches, and (the point of the fix) it's built once.
- `correctChoice` as a scalar matching one choice `id` → single-element list with that choice's `res`.
- `correctChoice` **absent** with non-empty `choices` → no exception, `correctChoiceList` untouched. *This one fails on `master`.*
- `correctChoice` as an array with **empty** `choices` → documents whichever behaviour you decide is intended above.

Given this feeds exam and survey grading, that coverage is worth more than the loop saving.

**Verdict:** mergeable. I'd rather it merged with the four tests, but nothing here regresses and the NPE guard is a net improvement over what's on `master` today.

---

Okuro3499 (approved):

PR #16501, opened 2026-08-27
life: smoother repository dao visibility filtering (fixes #16447)

### LINKED ISSUE #16447: Push the my-life dashboard visibility filter into MyLifeDao
**rating:** 58/100 &nbsp;·&nbsp; **proposed by:** copilot(r1), glm(r2)

`getMyLifeForDashboard` loads all rows via `myLifeDao.getByUserId` then `filter { it.isVisible }`, and again on the cache/seed paths; MyLifeDao has no visible-only query. The cached branch additionally does a two-pass `filter{}.map{}` where one pass would do.

**Ask:** add a visible-only query to `MyLifeDao` and use it for the dashboard path instead of loading all rows and filtering in Kotlin; also collapse the cached branch's two-pass `filter{}.map{}` into one pass.

**Files:** `data/room/dao/MyLifeDao.kt`, `repository/LifeRepositoryImpl.kt`, `test/.../LifeRepositoryImplTest.kt`

**Verification:** Verified the cached filter+map at LifeRepositoryImpl.kt:93-99. copilot pushes the predicate into SQL (the bigger win); glm only collapses the two-pass chain.

dogi:
This PR pushes the `isVisible` filter for dashboard items into `MyLifeDao` SQL query (`getVisibleByUserId`) and collapses the two-pass `filter {}.map {}` chain on cached items into a single `mapNotNull` pass in `LifeRepositoryImpl`. Corresponding unit tests in `LifeRepositoryImplTest` have been updated and expanded.

Fixes #16447

---
*PR created automatically by Jules for task [5419815247527632798](https://jules.google.com/task/5419815247527632798) started by @dogi*


dogi:
## Review: push the my-life dashboard visibility filter into `MyLifeDao`

**Rating: 70/100** — the DAO query is right, but swapping the filter into SQL silently changes the early-return condition and the dedup ordering. Neither is stated or tested.

### The query itself is correct

`MyLife.isVisible` is `var isVisible: Boolean = false` — non-null, stored by Room as INTEGER 0/1 — so `AND isVisible = 1` is exactly equivalent to `filter { it.isVisible }`, with no NULL trap. The rest of the predicate is copied verbatim from `getByUserId`, so the user-scoping semantics (including the `:userId IS NULL OR userId IS NULL` arm) are unchanged. Pushing the filter into SQLite instead of loading every row and discarding some in Kotlin is the right direction.

### Change 1: the early-return condition moved

```kotlin
- val allForUser = getMyLifeByUserId(effectiveUserId, ensureLatest = false)
- if (allForUser.isNotEmpty()) return allForUser.filter { it.isVisible }
+ val visibleForUser = getVisibleMyLifeByUserId(effectiveUserId)
+ if (visibleForUser.isNotEmpty()) return visibleForUser
```

The guard changed from **"this user has rows"** to **"this user has *visible* rows"**. For a user who has hidden every item, the old code returned an empty list immediately; the new code falls through into the SharedPreferences cache, then into `seedMyLifeIfEmpty`, then into a second full DAO read.

I traced where that lands, and the final answer is still empty — `seedMyLifeIfEmpty` guards on `countByUserId(userId) == 0`, which is false when hidden rows exist, so nothing is re-seeded, and the trailing `seeded.filter { it.isVisible }` re-applies the filter. So this is not a wrong-data bug today. But two things follow:

- **It's a performance regression on the path the PR is optimising.** In the all-hidden state, every dashboard load now does: the visible query, a SharedPreferences read, a JSON parse, a `countByUserId` query, a second full `getByUserId`, and a SharedPreferences **write** (`cacheMyLifeItems`) — where it used to do one query and return. That's strictly more work than `master`.
- **It depends on the cache never being stale.** If the SharedPreferences cache holds `isVisible = true` for an item the user has since hidden, the new code returns it and hidden items reappear on the dashboard; the old code could never reach the cache in that state. `cacheMyLifeItems` is called on the visibility-update path, so this should hold — but it's now a load-bearing invariant that wasn't one before, and it rests on *every* mutation path calling `cacheMyLifeItems`. Worth confirming rather than assuming.

The cheap fix keeps both the SQL filter and the original semantics:

```kotlin
val visibleForUser = getVisibleMyLifeByUserId(effectiveUserId)
if (visibleForUser.isNotEmpty()) return visibleForUser
if (myLifeDao.countByUserId(effectiveUserId) > 0) return emptyList()   // has rows, all hidden
// …cache / seed fall-through unchanged
```

One `COUNT(*)` in the rare empty-result case, and the fall-through means exactly what it did before: "this user has no rows at all."

### Change 2: dedup now runs after the filter, not before

- **Before:** `getByUserId` → `distinctBy { dedupKey() }` → `filter { isVisible }`
- **After:** `getVisibleByUserId` (filtered in SQL) → `distinctBy { dedupKey() }`

`dedupKey()` is `imageId ?: title ?: _id ?: identityHashCode`. So when two rows collide on that key — same `imageId`, or same `title` with a blank `imageId` — and one is visible while the other is hidden:

- **Before:** `distinctBy` keeps whichever comes first in `ORDER BY weight`, possibly the hidden one, and `filter` then drops it → the item **disappears** from the dashboard.
- **After:** SQL has already removed the hidden twin, so the visible one survives → the item **appears**.

The new behaviour is arguably the better one, and I'd keep it — but it's a real change, it's reachable (the presence of `distinctBy` at all, plus #16437 "make My Life de-duplication deterministic" being open right now, says collisions are a live concern), and nothing in the diff or the tests acknowledges it.

### Tests

The new test covers only the happy path — visible rows exist, they're returned. Note that `getMyLifeForDashboard_validJson` had to be re-stubbed from `getByUserId` to `getVisibleByUserId` returning empty, which is exactly the fall-through this PR widens, and no test pins down what *should* happen there. Two cases would cover both changes above:

```kotlin
@Test fun getMyLifeForDashboard_rowsExistButAllHidden_doesNotConsultCache()
@Test fun getMyLifeForDashboard_visibleAndHiddenShareDedupKey_keepsVisible()
```

The first fails on this branch with the current code and passes with the `countByUserId` guard.

### Smaller point

`cached.filter { it.isVisible }.map { … }` → `cached.mapNotNull { if (item.isVisible) MyLife(…) else null }` saves one intermediate list, which is fine, but `filter { }.map { }` read more clearly and this is a cold path (cache miss on the DB only). If you keep it, `mapNotNull { item -> item.takeIf { it.isVisible }?.let { … } }` says the same thing without the bare `else null`.

### Coordination

`LifeRepositoryImpl` is contended: #16368 (move seed-if-empty into the repository) and #16437 (deterministic de-duplication) are both open against it, and #16437 touches `dedupKey`/`distinctBy` directly — the same interaction as Change 2 above. Whichever lands second needs a base merge, and the dedup semantics should be settled in one place rather than negotiated across three PRs.

**Verdict:** needs a change. Add the `countByUserId` guard to preserve the early return, and cover the all-hidden and dedup-collision cases. The SQL filter itself is good and should stay.

---

dogi:
## Review: 82/100 — mergeable, but coordinate with #16096 first

The pushdown is careful in the one place it would have been easy to get wrong, and it quietly improves a dedup edge case. My main concern is a collision with another PR in the queue.

### The `countByUserId` guard is the load-bearing detail ✅

The naive version of this change is a regression, and it's avoided. On `master`:

```kotlin
val allForUser = getMyLifeByUserId(effectiveUserId, ensureLatest = false)
if (allForUser.isNotEmpty()) {
    return allForUser.filter { it.isVisible }   // may be empty — but still returns
}
// cache only reached when the user has NO rows at all
```

So "rows exist, all hidden" returns empty **without** consulting the SharedPreferences cache. Filtering in SQL destroys that distinction — an all-hidden user looks identical to a user with no rows — and would have resurrected stale cached items on the dashboard for someone who had deliberately hidden everything. The `countByUserId(effectiveUserId) > 0` check restores it exactly, and `getMyLifeForDashboard_rowsExistButAllHidden_doesNotConsultCache` pins it with `coVerify(exactly = 0)` on the preferences read. That's the right test for the right reason.

Cost is one extra `COUNT` only on the empty path; the common path drops from "load all rows including hidden" to one filtered query.

### A genuine (unadvertised) improvement in dedup

`distinctBy { it.dedupKey() }` now runs over **visible rows only**, where `master` deduped across all rows and *then* filtered. If a hidden row and a visible row share a dedup key and the hidden one sorts first by `weight`, `master` kept the hidden one, then filtered it out — so **neither** appeared and the visible item silently vanished from the dashboard. Now the visible one survives. Worth calling out in the description; it's a real bug fix hiding inside a perf change.

One nit: `getMyLifeForDashboard_visibleAndHiddenShareDedupKey_keepsVisible` doesn't actually exercise that — the DAO now returns only visible rows, so the mock has nothing hidden to collide with. It's really asserting "a visible item survives dedup". Either rename it or drop it; as written the name promises coverage that isn't there.

### ⚠️ Conflict with #16096 on the DAO predicate

The new query copies the existing predicate verbatim:

```sql
WHERE (:userId IS NULL OR userId IS NULL OR userId = :userId) AND isVisible = 1
```

**#16096 is changing exactly that predicate**, because `userId IS NULL` makes every guest-owned row match for *every* user — leaking guest entries into other users' My Life lists. If #16096 lands first, this PR reintroduces the leak in a second query that its fix doesn't touch, and the bug comes back on the dashboard path only — which is harder to notice than the original.

These two need to be reconciled deliberately: either this PR adopts #16096's corrected predicate, or it lands first and #16096 updates both queries. Given #16096 already carries the star tag and is at `merge`, adopting its predicate here is the safer order.

### Test hygiene

The last test now stubs both `getVisibleByUserId` → `emptyList()` **and** `getByUserId` → `listOf(item1, item2)`, but the new code path never calls `getByUserId` — it calls `countByUserId`, which falls to the relaxed-mock default. That stub is dead and misleading about which query drives the assertion. Worth stubbing `countByUserId` explicitly instead so the test states its own preconditions.

**Minor:** the `filter{}.map{}` → `mapNotNull { if (…) … else null }` rewrite in the cache branch is behaviourally identical and saves one intermediate list, but reads worse than the original. I'd keep `filter {}.map {}`.

**Verdict:** mergeable on its own terms — but settle the predicate question against #16096 before merging, or the guest-row leak returns through the back door.

---

Okuro3499 (approved):

PR #16502, opened 2026-08-27
resources: smoother repository files downloading (fixes #16493)

### LINKED ISSUE #16493: Replace the dead null-guard in ResourcesRepositoryImpl.downloadFiles
**rating:** 45/100 &nbsp;·&nbsp; **proposed by:** glm(r3)

`var files = libraryList`, then `if (files == null) files = getAllLibrariesToSync()`, then `val safeFiles = files ?: emptyList()` — after the null branch the value is never null, so the elvis is dead and safeFiles is just files.

**Ask:** remove the dead elvis/`safeFiles` indirection — after the null branch assigns a value, `files` is non-null, so use it directly.

**Files:** `repository/ResourcesRepositoryImpl.kt`

**Verification:** File and function exist; the exact three-line shape was not re-verified.

dogi:
## What

Removes the dead null-guard in `ResourcesRepositoryImpl.downloadFiles`.

The method held a redundant null guard:

```kotlin
var files = libraryList
if (files == null) {
    files = getAllLibrariesToSync()
}
val safeFiles = files ?: emptyList()   // dead: files is never null here
...
return safeFiles
```

After the `if (files == null)` branch assigns a value, `files` is never null, so the `files ?: emptyList()` elvis is dead code and `safeFiles` is just an alias for `files`. This replaces the mutable-null-then-elvis pattern with a single `val` and uses it directly:

```kotlin
val files = libraryList ?: getAllLibrariesToSync()
val urls = DownloadUtils.downloadAllFiles(files)
...
return files
```

## Why

`var` + null-check + elvis is a three-line way to write a one-line coalescing assignment. The elvis can never fire, so it's misleading dead code that suggests `getAllLibrariesToSync()` could still return null (it returns `List<MyLibrary>`, non-null). Collapsing it to `val files = libraryList ?: getAllLibrariesToSync()` keeps the same behavior — including the null fallback to `getAllLibrariesToSync()` — with no dead branch.

Closes #16493.

## How

- `repository/ResourcesRepositoryImpl.kt` — `downloadFiles` now uses `libraryList ?: getAllLibrariesToSync()` and returns `files` directly; the `safeFiles` indirection is gone.
- `app/src/test/.../ResourcesRepositoryImplTest.kt` — added two tests covering both paths:
  - `downloadFiles returns provided list directly when not null` — verifies the provided list is returned as-is and `getAllLibrariesToSync()` is not called.
  - `downloadFiles falls back to getAllLibrariesToSync when list is null` — verifies the null path calls `myLibraryDao.getSyncable()` and returns the synced list.

## Verification

```
./gradlew testDefaultDebugUnitTest --tests "org.ole.planet.myplanet.repository.ResourcesRepositoryImplTest"
```

`ResourcesRepositoryImplTest`: 38 tests, 0 failures, 0 errors (both new tests pass).

---

This PR was created by an AI agent (OpenHands) on behalf of the repository maintainer.

@dogi can click here to [continue refining the PR](https://app.all-hands.dev/conversations/1f93f77c-7365-42b4-9a26-49332fdc1b9e)


dogi:
## Review: simplify `downloadFiles` null handling

**Rating: 90/100** — removes genuinely dead code and fixes a formatting defect, with tests that pin both branches.

### Correctness

`ResourcesRepository.getAllLibrariesToSync()` is declared `suspend fun getAllLibrariesToSync(): List<MyLibrary>` — non-null. So in the old code:

```kotlin
var files = libraryList
if (files == null) files = getAllLibrariesToSync()
val safeFiles = files ?: emptyList()   // can never be null here
```

the `?: emptyList()` was unreachable: after the `if`, `files` is non-null on both paths. The compiler couldn't prove it because `files` was a `var` of nullable type, which is exactly why the elvis got written. Switching to `val files = libraryList ?: getAllLibrariesToSync()` gives the compiler the non-null type directly and the fallback disappears on its own. Behaviour is identical for both a null and a non-null argument.

The diff also repairs the function signature's indentation — `override suspend fun downloadFiles` was flush against the left margin, outside the class body's indent level. Small, but it was the kind of thing that makes a file look corrupted in review.

### Tests

Both branches are covered, and the assertions are the right ones: `assertSame(library, result[0])` proves the caller's list is returned as-is rather than copied, and `coVerify(exactly = 0) { myLibraryDao.getSyncable() }` proves the DAO isn't consulted when a list is supplied — which is the actual behavioural contract, not just the return value.

Worth noting the test setup change is an improvement in its own right: the two trailing `mockk(relaxed = true)` positional arguments in `setUp` are now named `configurationsRepository` and `dispatcherProvider`, so they can be stubbed and verified. That's what makes these tests possible at all.

### Nits

- `import org.junit.After` is added but no `@After` method appears — the teardown is done with `try`/`finally` inside each test instead. Either drop the import or move the `unmockkObject` calls into an `@After`; as written it's an unused import and `build.yml` runs with `--warning-mode all`.
- `CoroutineScope(SupervisorJob())` stands in for `MainApplication.applicationScope`, so the `launch` block runs on `Dispatchers.Default` outside the test's control and the test may complete before it does. It's harmless here (`checkServerAvailability()` is stubbed `false`, so the body no-ops), but a `TestScope(testScheduler)` would keep it deterministic and let a future test assert on what that branch does.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16506, opened 2026-08-27
all: smoother download utils channels configuring (fixes #16483)

### LINKED ISSUE #16483: One-shot guard for DownloadUtils.createChannels
**rating:** 49/100 &nbsp;·&nbsp; **proposed by:** copilot(r2)

`createChannels` is invoked from every buildInitialNotification, buildProgressNotification and completion builder plus DownloadService/DownloadWorker; it null-checks channels but still makes three getNotificationChannel system calls per progress tick.

**Ask:** add a one-shot guard so `createChannels` does its work once per process instead of making three `getNotificationChannel` system calls on every progress tick.

**Files:** `utils/DownloadUtils.kt`, `test/.../DownloadUtilsTest.kt`

**Verification:** File exists; the per-tick call count was not verified.

dogi:
Adds a process-level flag `@Volatile channelsCreated` in `DownloadUtils.kt` so system calls to `getNotificationChannel` and `createNotificationChannel` are executed only once per process runtime instead of on every progress update/tick. Also adds unit testing and a `@VisibleForTesting` reset function.

Fixes #16483

---
*PR created automatically by Jules for task [8578941020028719482](https://jules.google.com/task/8578941020028719482) started by @dogi*


dogi:
## Review: process-level one-shot guard for `DownloadUtils.createChannels`

**Rating: 88/100** — this is a better optimisation than the diff makes it look, because of where `createChannels` is called from.

### Why it's worth doing

`createChannels` isn't only called at service startup. It's called from three places inside `DownloadUtils` itself:

- `:60` — `buildInitialNotification`
- `:83` — the progress-notification builder
- `:111` — the completion-notification builder

plus `DownloadService:111`, `DownloadService:195`, and `DownloadWorker:54`. So during an active download, every notification rebuild was making **three `getNotificationChannel` binder round-trips** to `NotificationManagerService` before doing anything useful. Guarding on a process-level flag removes all of them after the first call. That's a real saving on a genuinely repeated path, not a theoretical one.

### Correctness

The cached fact is safe to cache because nothing invalidates it: there is no `deleteNotificationChannel` call anywhere in the codebase, so a channel created once persists for the process's lifetime (and beyond — channels survive app restarts, which just makes the guard's first call cheap too). A user disabling a channel doesn't remove it, and `getNotificationChannel` would still have returned non-null in that case, so the old code wouldn't have recreated it either. No behaviour change.

`@Volatile` is the right annotation. Two threads can still both pass the guard and both run the body — but `createNotificationChannel` on an existing channel is idempotent, and each block is already gated by its own `getNotificationChannel(...) == null` check, so the worst case is duplicated work, never inconsistent state. Worth a one-line comment saying that's a deliberate accept rather than an oversight, since the next reader will wonder why it isn't `synchronized`.

### Test

`verify(exactly = 3)` after two calls, then `exactly = 6` after a reset and a third — that's discriminating: without the guard the first assertion sees 6 and fails. The `@Before`/`@After` resets are necessary given the flag is process-global and the suite runs two shards in one JVM per fork; leaving it set would leak into any other test that touches `createChannels`.

### Nit

`resetChannelsCreatedForTesting()` adds a production function that exists only for tests. It's correctly marked `@VisibleForTesting internal`, so this is acceptable — but marking the field itself `@VisibleForTesting internal var channelsCreated` and assigning it directly from the test would achieve the same with one less member on the object's surface.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16507, opened 2026-08-27
resources: smoother collections testing (fixes #16492)

### LINKED ISSUE #16492: Remove the dead filteredList field and the list-copy churn in CollectionsFragment
**rating:** 46/100 &nbsp;·&nbsp; **proposed by:** devin(r3)

An unused `filteredList` property, plus `currentTagDataList` rebuilt with `.toMutableList()` at five assignment sites although it is only ever reassigned, never mutated in place.

**Ask:** delete the unused `filteredList` property and drop the `.toMutableList()` copies at the `currentTagDataList` assignment sites, since the list is only ever reassigned and never mutated in place.

**Files:** `ui/resources/CollectionsFragment.kt`

**Verification:** The `.toMutableList()` at the filterTags assignment is verified; the unused field and the other four sites were not. Shares a file with the lowercase-hoist task.

dogi:
## What & Why

Fixes #16492.

`CollectionsFragment` carried an unused `filteredList` property, and `currentTagDataList` was rebuilt with `.toMutableList()` at five assignment sites even though the field is only ever **reassigned**, never mutated in place. The copies were pure churn.

This PR:
- deletes the unused `filteredList: ArrayList<TagEntity>` property
- drops the `.toMutableList()` copies at all five `currentTagDataList` assignment sites (Success state, `filterTags`, `onParentTagClicked`, `onCheckboxTagSelected`, `onCheckedChanged`)
- narrows the field type from `MutableList<TagData>` to `List<TagData>`, since only whole-list reassignment happens

## Changes

**`app/src/main/java/org/ole/planet/myplanet/ui/resources/CollectionsFragment.kt`**
- removed `private var filterededList: ArrayList<TagEntity> = ArrayList()`
- `currentTagDataList` is now `private var currentTagDataList: List<TagData> = emptyList()`
- five `buildTagDataList(...).toMutableList()` calls became `buildTagDataList(...)`

## Verification

Added `CollectionsFragmentTest` exercising the `buildTagDataList` logic that the assignment sites depend on:
- builds a plain list of parents
- expansion state carries forward across reassignment (parent → parent + child)
- selection state from `selectedItemsList` is reflected

Ran the new test plus the existing `CollectionsViewModelTest`:

```
./gradlew :app:testDefaultDebugUnitTest \
  --tests "org.ole.planet.myplanet.ui.resources.CollectionsFragmentTest" \
  --tests "org.ole.planet.myplanet.ui.resources.CollectionsViewModelTest"
```

All tests pass (3 new + 4 existing, 0 failures).

---

This PR was created by an AI agent (OpenHands) on behalf of the maintainer.

@dogi can click here to [continue refining the PR](https://app.all-hands.dev/conversations/5e2e9bfc-1650-47fc-9898-1fc2e27e8847)


dogi:
## Review: dead field cleanup in `CollectionsFragment`

**Rating: 89/100** — both changes check out against the file, including the one that could plausibly have broken.

### `filteredList` is genuinely dead

Declared at `:38` and referenced nowhere else in the file — not read, not written, not passed. (The similarly-named `filteredParentList` at `:115` is a different, local variable.) Straight deletion.

### `MutableList` → `List` is safe, and I verified the part that matters

`currentTagDataList` is only ever *reassigned*, never mutated in place — there's no `.add`/`.remove`/`.clear` on it anywhere. Its only read is inside `buildTagDataList`, which walks it to rebuild `parentMap` so `isExpanded` survives a rebuild.

The thing worth checking was whether dropping `.toMutableList()` breaks either the expansion carry-forward or `ListAdapter.submitList`:

- **Carry-forward:** `.toMutableList()` was a *shallow* copy, so the `TagData.Parent` instances were already shared between the copy and `buildTagDataList`'s return value. Removing the copy changes no object identity, so `parentMap[it.tag.id]` still finds the same `Parent` carrying the user's `isExpanded` flag. Unchanged.
- **`submitList`:** `buildTagDataList` allocates a fresh `mutableListOf<TagData>()` on every call and returns it, so each assignment is still a distinct instance. `ListAdapter` short-circuits when handed the same reference twice — that risk would be real if `buildTagDataList` cached, and it doesn't. Diffing still runs on every update.

So the five `.toMutableList()` calls were pure copies of a list that was about to be discarded, on paths that fire on every keystroke in the filter (`:115`), every expand/collapse (`:152`), every selection toggle (`:162`), and every switch flip (`:172`). Removing them is free.

Tightening the declared type from `MutableList` to `List` is the more valuable half: it makes "nothing mutates this" a compiler-enforced fact rather than a convention, which matters precisely because handing a mutable list to `submitList` is a classic `ListAdapter` footgun.

### Tests

Good that they exist — this fragment had none — and they test the real risk: `buildTagDataList carries expanded children forward across reassignments` is exactly the invariant the `.toMutableList()` removal could have broken, and it drives it through the same reassign-then-rebuild sequence `onParentTagClicked` uses.

The cost is heavy reflection: two private fields, one private method, plus a hand-rolled `findField` walking the superclass chain. That works, but it welds the tests to field and method *names* — a rename breaks them with a `NoSuchFieldException` rather than a compile error, which is a much worse failure mode. Since `buildTagDataList` is a pure function of `list` / `childMap` / `selectedItemsList` / `MainApplication.isCollectionSwitchOn` and touches no view state, the cleaner shape would be to lift it out of the fragment into an `internal` top-level function taking those as parameters. The fragment gets simpler, the tests lose all the reflection, and the coverage is identical.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16508, opened 2026-08-27
life: smoother item moving (fixes #16505)

### LINKED ISSUE #16505: Remove the redundant intermediate list in LifeAdapter.onItemMoveFinished
**rating:** 44/100 &nbsp;·&nbsp; **proposed by:** minimax(r2)

`val finalList = list.toList()` is created and passed to both reorderCallback and submitList, although submitList makes its own copy.

**Ask:** drop the redundant `list.toList()` intermediate in `onItemMoveFinished`, since `submitList` already makes its own copy.

**Files:** `ui/life/LifeAdapter.kt`

**Verification:** File exists; not line-verified. Shares a file with the ViewBinding task.

dogi:
## Summary

`LifeAdapter.onItemMoveFinished` originally created a redundant intermediate copy with `val finalList = list.toList()` and passed it to both `reorderCallback` and `submitList`. The first version of this PR dropped that copy entirely and passed the live `MutableList` directly to both consumers.

Review feedback (\@dogi, 72/100) flagged that removing the copy lets the **same** `MutableList` instance be shared between `AsyncListDiffer`'s background diff and the IO coroutine captured by `updateMyLifeListOrder` — risking `IndexOutOfBoundsException: Inconsistency detected` from the layout manager under a fast drag. A defensive copy on a shared-mutable-state boundary that costs one `ArrayList` of a dozen references, once per completed drag, is the copy you keep.

This revision keeps a **single** immutable `toList()` snapshot (the same one copy `master` had) and nulls `dragList` **before** the callback so nothing can race the list that escapes the gesture — no live `MutableList` is reachable from either `reorderCallback` or `submitList`.

## Changes

- `app/src/main/java/org/ole/planet/myplanet/ui/life/LifeAdapter.kt` — `onItemMoveFinished` now takes an immutable `dragList?.toList() ?: return` snapshot, nulls `dragList` first, then passes that immutable list to `reorderCallback` and `submitList`.
- `app/src/test/java/org/ole/planet/myplanet/ui/life/LifeAdapterTest.kt` — tests covering move + `onItemMoveFinished` ordering, the no-op case when no drag occurred, and that `reorderCallback` receives the reordered list; removed a stale comment that claimed the drag list is passed directly.

## Verification

No Android SDK / JDK toolchain is available in this environment to run the unit tests locally (`./gradlew testDefaultDebugUnitTest`). The change is minimal and behavior-preserving: `submitList` copies its input, and the restored `toList()` is the defensive snapshot that decouples the gesture buffer from both the differ and the reorder coroutine. CI (`test.yml`) runs `testDefaultDebugUnitTest` on push.

Closes #16505.

---


dogi:
## Review: drop the `toList()` copy in `onItemMoveFinished`

**Rating: 72/100** — correct as the code stands today, but it trades a real safety property for a saving that doesn't exist, and the aliasing it introduces is wider than the diff suggests.

### It is correct today

I traced the three things that could break:

1. **The adapter won't mutate the escaped list.** `dragList = null` runs immediately after, and the next gesture starts with `dragList = currentList.toMutableList()` (`:111`) — a fresh copy every time. So the instance handed to `submitList` is never touched by `onItemMove` again.
2. **The callback can't mutate it through its declared type.** `reorderCallback: (List<MyLife>) -> Unit` is read-only; mutating requires a deliberate downcast.
3. **Nothing downstream mutates it.** `LifeFragment:61` → `viewModel.updateMyLifeListOrder(list)` → `lifeRepository.updateMyLifeListOrder(list)`, all read-only.

So no bug ships. But:

### What the copy was buying

After this change the *same* `MutableList` instance is simultaneously:

- `AsyncListDiffer`'s committed list (`submitList(list)`), diffed on a background thread, and
- captured by a coroutine that outlives the call — `updateMyLifeListOrder` does `viewModelScope.launch { withContext(dispatcherProvider.io) { … } }`, so the reference is read on the IO dispatcher well after `onItemMoveFinished` returns.

`.toList()` made "nothing can mutate what the adapter is displaying" a local, one-line, checkable fact. Without it, that guarantee is distributed across three files and depends on `onItemMove` always re-copying, `dragList = null` always running first, and every future consumer of `reorderCallback` never casting. Mutating a `ListAdapter`'s backing list out from under it is one of the more unpleasant RecyclerView failures — `IndexOutOfBoundsException: Inconsistency detected` from the layout manager, usually reproducing only under a fast drag.

### And what it costs

One `ArrayList` copy of a dozen-odd `MyLife` *references*, once per completed drag gesture — a user-driven action that happens at human speed and is already doing a DiffUtil pass plus a Room write on the same tick. There is no measurable saving here, and the PR doesn't claim a measurement.

That's the whole objection: a defensive copy on a shared-mutable-state boundary is exactly the copy you keep when it costs nothing.

If the goal is to remove the redundancy rather than the defence, the clean version keeps the guarantee and still drops a copy — change `dragList` to be built as an immutable result:

```kotlin
override fun onItemMoveFinished() {
    val finalList = dragList?.toList() ?: return
    dragList = null                 // null it first, so nothing can race the callback
    reorderCallback(finalList)
    submitList(finalList)
}
```

Same one copy as `master`, but no live `MutableList` is reachable from either consumer.

### Tests

Good that `LifeAdapter` now has any, and running them under `AndroidJUnit4` against the real adapter (rather than mocking `ListAdapter`) is the right call. The no-drag case (`dragList == null` → callback not invoked) is a genuine branch worth pinning.

One mismatch though — this comment:

```kotlin
// The drag list is passed directly (no redundant toList() copy) to both callbacks
assertEquals(3, receivedList!!.size)
```

The assertion checks the *size*, which is 3 either way. Nothing in the suite distinguishes "passed directly" from "passed as a copy", so all three tests pass unchanged on `master`. If direct-passing is the property being asserted, it needs `assertSame`; if it isn't, the comment should go, because right now it documents a guarantee the test doesn't hold.

**Verdict:** needs a change. Either restore the copy in the shape above, or make the case with a measurement — right now the diff gives up a real invariant for nothing.

---

dogi:

---

dogi:
@dogi thanks for the review — agreed. Dropping the defensive copy traded a checkable local fact for a guarantee spread across three files (the differ'''s background diff vs. the IO coroutine captured by `updateMyLifeListOrder`). One `ArrayList` of a dozen references, once per completed drag, is the cheap copy you keep.

Implemented your suggested clean version in 4e33a69:

```kotlin
override fun onItemMoveFinished() {
    val finalList = dragList?.toList() ?: return
    dragList = null                 // null first, so nothing can race the escaped list
    reorderCallback(finalList)
    submitList(finalList)
}
```

- Restores a single immutable `toList()` snapshot (same one copy as `master`) — no live `MutableList` is reachable from either `reorderCallback` or `submitList`.
- Nulls `dragList` before invoking the callback, closing the window where the gesture'''s working buffer could be observed alongside the escaped reference.
- Removed the now-stale test comment claiming the drag list is passed directly; the ordering/size assertions still hold and remain valid.

No `@Okuro3499` review comments or inline threads were present on the PR or the linked issue, so there were no additional improvement instructions to apply beyond your review.

No JDK/Android toolchain in this environment, so `testDefaultDebugUnitTest` will run in CI on push.

_This comment was created by an AI agent (OpenHands) on behalf of the repository maintainers._

dogi:
## Review: 80/100 — mergeable

The reordering is correct and strictly safer, and the new test file gives `LifeAdapter` its first coverage. The one gap is that nothing tests the actual change.

### What the reorder buys

Clearing `dragList` **before** invoking the callbacks closes a re-entrancy hole. On `master`:

```kotlin
dragList?.let { list ->
    val finalList = list.toList()
    reorderCallback(finalList)   // ← if this re-enters onItemMoveFinished…
    submitList(finalList)
    dragList = null              // ← …dragList is still non-null here
}
```

A synchronous `reorderCallback` that reached back into the adapter would find `dragList` still populated and fire `reorderCallback` a second time with the same list — or recurse. The new order returns immediately on re-entry. Standard "clear state before you call out" discipline, and the `?: return` early exit reads better than the nested `let`.

Being straight about the value: this is **hardening, not a fix for an observed bug**. `reorderCallback` currently persists the new order through a coroutine, so it doesn't re-enter synchronously today. The change is still worth taking — it removes a footgun for whoever makes that callback synchronous — but the description shouldn't imply a user-visible bug was fixed.

One behavioral consequence worth noting: if `reorderCallback` throws, `master` left `dragList` populated (so a subsequent call could retry) while this leaves it null (drag state discarded). Discarding is the better default — a half-applied reorder shouldn't be replayable — just be aware it changed.

**`submitList(finalList)` is safe from the `AsyncListDiffer` reference-equality short-circuit**, since `finalList` is a fresh `toList()` copy each time rather than the adapter's current list instance.

### The tests don't cover the change

All three pass on `master`, because none of them makes `reorderCallback` re-enter the adapter — which is the only observable difference. They're useful characterization tests for `onItemMove`/`onItemMoveFinished` (and the class had none), but they'd be green with the diff reverted.

The test that would fail on `master` is small:

```kotlin
var invocations = 0
lateinit var adapter: LifeAdapter
adapter = LifeAdapter(
    context = context,
    mDragStartListener = noopDrag,
    visibilityCallback = { _, _ -> },
    reorderCallback = { invocations++; adapter.onItemMoveFinished() }  // re-entrant
)
adapter.submitList(listOf(lifeItem("a", 0), lifeItem("b", 1)))
adapter.onItemMove(0, 1)
adapter.onItemMoveFinished()
assertEquals(1, invocations)   // master: 2 (or recurses)
```

That turns "trust me, clearing first is safer" into something CI enforces.

**Minor:** the tests rely on `submitList` dispatching synchronously, which holds only because the previous list is empty. Fine as written; worth knowing if a test ever submits twice.

**Contention:** #16096 also touches `LifeAdapter.kt` around this method and is already labelled `merge`. Whichever lands second wants a re-check — and if #16096 goes first, rebase before merging this.

**Verdict:** mergeable. Adding the re-entrancy test would make it clearly so.

---

Okuro3499 (approved):

dogi:
Merged `master` in to clear the conflict, and added the re-entrancy test from your 80/100 review.

## Conflict

The only conflict was `LifeAdapterTest.kt`, add/add — `master` and this branch each created that file independently. Resolved as a union rather than by picking a side, since the two sets cover different things:

- kept `master`'s four: the ViewBinding holder test, title binding, visibility alpha, and its two-item reorder test, along with its `RobolectricTestRunner` / `@Config` / `setUp` scaffolding
- kept this branch's two distinct cases: the three-item move (`0→2`, asserting `onItemMove` returns true) and the no-drag no-op
- dropped this branch's `reorderCallback receives the reordered drag list directly` — an exact duplicate of `master`'s `onItemMove reorders list and onItemMoveFinished propagates order`, which is strictly stronger since it also asserts the callback fires exactly once

`LifeAdapter.kt` auto-merged cleanly: `master`'s ViewBinding view holder plus this branch's `onItemMoveFinished` rewrite. Diffed against both parents to confirm neither side was lost.

## The re-entrancy test

You were right that nothing tested the actual change — all the existing tests are green with the diff reverted. Added in c7c2401:

```kotlin
@Test
fun `onItemMoveFinished clears drag state before the callback so re-entry is a no-op`() {
    var invocations = 0
    lateinit var reentrantAdapter: LifeAdapter
    reentrantAdapter = adapterWith {
        invocations++
        reentrantAdapter.onItemMoveFinished()
    }

    reentrantAdapter.submitList(listOf(lifeItem("a", 0), lifeItem("b", 1)))
    reentrantAdapter.onItemMove(0, 1)
    reentrantAdapter.onItemMoveFinished()

    assertEquals(1, invocations)
}
```

Verified in both directions rather than just asserting it passes. Against `master`'s implementation it fails with `StackOverflowError` — your "(or recurses)" is what actually happens, since `dragList` is never nulled before `reorderCallback` runs, so the nested `let` re-enters unboundedly rather than firing twice. Against the current implementation, 7/7 green.

Also ran the full `LifeAdapterTest` class locally under `testDefaultDebugUnitTest` (7 tests, 0 failures) — the earlier "no toolchain available" note on this PR no longer applies.

## One note

#16096 is still open and still touches `LifeAdapter.kt` around this method, so the contention you flagged stands for whichever lands second.

The `conflict` label is now stale — the merge resolved it — so that can come off and `automerge` go back on whenever you're ready.

---

PR #16510, opened 2026-08-27
teams: smoother voices label managing (fixes #16472)

### LINKED ISSUE #16472: Reduce per-label allocations in VoicesLabelManager.formatLabelValue
**rating:** 51/100 &nbsp;·&nbsp; **proposed by:** devin(r2)

`formatLabelValue` chains `replace("_"," ").replace("-"," ")` and calls `Locale.getDefault()` inside the joinToString lambda for every word of every chip label, during list scrolling.

**Ask:** hoist the `Locale.getDefault()` call out of the `joinToString` lambda and collapse the chained `replace` calls into a single pass, so each label costs fewer allocations during scroll.

**Files:** `services/VoicesLabelManager.kt`

**Verification:** File exists; not line-verified. Shares a file with the showChips task.

dogi:
Hoisted Locale.getDefault() out of the joinToString lambda in formatLabelValue and collapsed chained replace("_", " ").replace("-", " ") calls into a single pass using a reusable Regex.

Fixes #16472

---
*PR created automatically by Jules for task [18292236725910664968](https://jules.google.com/task/18292236725910664968) started by @dogi*


dogi:
## Review: reduce per-label allocations in `formatLabelValue`

**Rating: 81/100** — correct, and the locale hoist is a real (if small) win. The allocation claim in the title is roughly a wash, and the path is colder than it looks.

### Correctness

`raw.replace("_", " ").replace("-", " ")` → `raw.replace(Regex("[_-]"), " ")` is exactly equivalent. The `-` sits last in the character class, where it's a literal hyphen rather than a range operator — the one way this could have gone wrong, and it didn't. Every `_` and `-` is still replaced independently, in one pass instead of two.

Hoisting `Locale.getDefault()` into a local is the best part of the diff: it was previously called **twice per word** inside the `joinToString` lambda (once for `lowercase`, once for `titlecase`), so a three-word label made six calls into `Locale.getDefault()`. Now it's one per invocation, and as a bonus all words in a label are guaranteed to use the same locale even if the default were swapped mid-call.

Moving `whitespaceRegex` above the function it's used in is cosmetic but correct — `by lazy` initialisation order in a companion doesn't depend on declaration position, so this changes nothing at runtime and just reads better.

### On the allocation claim

Worth being precise, since it's the stated goal:

- **Before:** `replace("_", " ")` and `replace("-", " ")` each allocate a `StringBuilder` and a result `String` → ~4 allocations.
- **After:** `Regex.replace` allocates a `Matcher`, a `StringBuilder`, and a result `String` → ~3 allocations.

So one fewer object, but a `Matcher` is a heavier object than the short intermediate `String` it displaces, and the regex engine is slower per character than `String.replace`'s `indexOf` scan for single-character literals. For label-length strings this is a wash at best. The honest framing is "one pass instead of two, and clearer intent" — which is a fine reason on its own.

### The path is a fallback

`formatLabelValue` is only reached from `getLabel`:

```kotlin
private fun getLabel(s: String): String = reverseLabels[s] ?: formatLabelValue(s)
```

Every label that exists in `Constants.LABELS` hits the `reverseLabels` map and never reaches this function. So `formatLabelValue` runs only for labels *not* in the known set — the custom/unknown fallback. That doesn't make the change wrong, but it means neither the allocation nor the locale saving is going to show up in a profile. Worth setting expectations in the PR body rather than presenting it as a hot-path fix.

### Coordination

`VoicesLabelManager` is contended — #16482 rewrites `showChips` in the same file (and currently carries `change` for a cache-key bug). Different regions, so a merge should be clean, but whichever lands second wants a base merge and a CI re-run.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16513, opened 2026-08-27
sync: smoother preview flowing (fixes #16468)

### LINKED ISSUE #16468: Coalesce the parallel-phase SyncManager status emissions
**rating:** 52/100 &nbsp;·&nbsp; **proposed by:** copilot(r2)

Phase 1 launches ~14 tables in parallel and each completion writes `_syncStatus.value = SyncStatus.Syncing(...)`, so concurrent writers thrash the StateFlow collectors (login/sync UI) with near-simultaneous updates and string allocations.

**Ask:** coalesce the parallel-phase status emissions so ~14 concurrent table completions don't each push a separate `SyncStatus.Syncing(...)` write to the StateFlow — batch or throttle them into meaningfully fewer updates.

**Files:** `services/sync/SyncManager.kt`, `test/.../SyncManagerTest.kt`

**Verification:** File and parallel-phase shape verified; the emission count was not measured. Shares a file with the sync-logging task.

dogi:
## Summary

Phase 1 launches ~14 tables in parallel and Phase 3 launches library shelves in parallel; each completion wrote its own `SyncStatus.Syncing(...)` to the `_syncStatus` StateFlow.

The first iteration put a `SyncStatusCoalescer` on the producer side. @dogi's review showed that was net-negative — see the rewrite rationale below. This revision drops the coalescer and moves throttling to the collector.

## What changed (this revision)

- **Dropped `SyncStatusCoalescer`** entirely (and its test). `_syncStatus` is a `MutableStateFlow`, which already conflates: a slow/suspended collector only ever resumes on the latest value, so no UI consumer can observe the ~14 intermediate writes — the cost being removed was ~14 atomic stores spread across a multi-second network phase.
- **Reverted `SyncManager` to a final `class`**: removed the `open` modifier, the `protected open fun emitSyncStatus` test seam, and the `SYNC_STATUS_COALESCE_MS` companion. Restored the direct `_syncStatus.value =` writes in Phase 1 (tables) and Phase 3 (shelves). `SyncManager` is back to byte-for-byte its pre-PR state.
- **Moved throttling to the one collector that does real per-status redraws** — `SyncActivity`'s sync progress dialog — via `.sample(SYNC_STATUS_SAMPLE_MS)` (150 ms) on the collected `syncManager.syncStatus` flow. `DashboardActivity` only reacts to `Success`, so it's untouched.

## Why the producer-side coalescer was dropped

The review was right on four counts:

1. **StateFlow already conflates** — `value =` is an atomic store that skips equal values, and a slow collector never sees intermediate emissions. The quantity the old test optimized (write count) was invisible through the flow itself, which is why the test had to subclass `SyncManager` and override a `protected` seam to count it.
2. **The coalescer cost more than it saved** — every `report()` cancelled a `Job` and launched a new coroutine + `delay` timer, replacing an atomic store with scheduler churn.
3. **Data race** — `pending` was a plain `var` mutated from concurrent `async` workers under a `Semaphore(6)` on `dispatcherProvider.io`, so two reports could interleave and orphan an uncancellable deferred.
4. **In production it coalesced nothing** — the 14 table completions are seconds apart, not within the 150 ms quiet window, so each report armed a timer, waited it out unchallenged, and emitted — one per table as before, *plus* 150 ms of added latency on every progress update. The burst only appeared under the test's instant-return mocks.

Consumer-side `.sample()` has no concurrency surface, opens no class, adds no test seam, and on a hot `StateFlow` never permanently drops a value (it emits the latest at each tick), so terminal `Success`/`Error` states still land.

## Verification

- `./gradlew :app:testDefaultDebugUnitTest --tests "org.ole.planet.myplanet.services.sync.*"` — all sync-package tests pass (SyncManagerTest's 3 original tests; coalescer test removed with the class).
- `:app:compileDefaultDebugKotlin` — main sources (incl. `SyncManager` + `SyncActivity`) compile cleanly.

## Files

- `services/sync/SyncManager.kt` — reverted to pre-PR state (final class, direct writes).
- `ui/sync/SyncActivity.kt` — `.sample(150L)` on the collected `syncStatus` flow; `SYNC_STATUS_SAMPLE_MS` constant; `@OptIn(ExperimentalCoroutinesApi)`.
- `services/sync/SyncStatusCoalescer.kt` — deleted.
- `test/.../SyncStatusCoalescerTest.kt` — deleted.
- `test/.../SyncManagerTest.kt` — reverted to pre-PR state (removed the seam-based write-count test).

Fixes #16468.

---


dogi:
## Review: coalesced parallel-phase status emissions

**Rating: 52/100** — the problem being solved is already solved by `StateFlow`, the fix costs more than what it replaces, the coalescer has a data race on the field it uses for coordination, and in production it will coalesce nothing.

### `StateFlow` already conflates

The premise is that "~14 near-simultaneous completions would otherwise each thrash the StateFlow." But `_syncStatus` is a `MutableStateFlow`, and:

- `value =` is an atomic store that **skips entirely** when the new value `equals` the current one;
- `StateFlow` is conflated, so a collector that is slow or suspended **never observes** intermediate values — it resumes on the latest.

So no UI consumer can see the 14 intermediate writes today. The cost being removed is 14 atomic CAS operations spread across a multi-second network phase. There is nothing there to reclaim.

This is also why the test has to subclass `SyncManager` and override `emitSyncStatus` to count writes: the quantity being optimised is **invisible through the StateFlow itself**, because the flow conflates it. A metric that no consumer can observe is a strong signal that the optimisation has no observable effect.

### The coalescer costs more than it saves

```kotlin
fun report(status: …) {
    latest.set(status)
    val s = scope ?: return
    pending?.cancel()
    pending = s.launch { delay(intervalMillis); flush() }
}
```

Every `report` **cancels a Job and launches a new coroutine**. For 14 completions that's 14 coroutine launches, 14 `Job` allocations, 13 cancellations, and 14 `delay` timer registrations — replacing 14 atomic stores. This is strictly more work, more allocation, and more scheduler churn than the code it replaces.

### The data race

`pending` is a plain `var`, mutated from `report`, which is called concurrently from the `async` workers — and in Phase 3 explicitly from `async(dispatcherProvider.io)` under a `Semaphore(6)`, so genuinely from multiple threads.

Two concurrent `report` calls can interleave as:

1. T1 reads `pending` = J0, cancels J0
2. T2 reads `pending` = J0, cancels J0 (no-op)
3. T1 assigns `pending` = J1
4. T2 assigns `pending` = J2 — **J1 is now orphaned and uncancellable**

`stop()` cancels only whatever `pending` currently holds, so J1 survives and fires `flush()` 150 ms later. It happens to be contained because `coroutineScope` awaits its children and `flush()` uses `getAndSet(null)`, so the orphan usually no-ops — but that's luck, not design. `latest` is an `AtomicReference` while `pending`, which needs the same protection, is unguarded. If this class stays, `pending` needs to be an `AtomicReference<Job?>` with `getAndSet`, or the whole `report` body needs a lock.

### In production it will coalesce nothing

The 150 ms quiet window only merges reports that land **within 150 ms of each other**. Phase 1 syncs 14 CouchDB tables over the network; those completions are seconds apart, not milliseconds. So each report will arm a timer, wait out the full 150 ms unchallenged, and emit — one emission per table, exactly as before, plus **150 ms of added latency on every progress update**.

The burst behaviour only appears under the test's `coEvery { transactionSyncManager.syncDb(any()) } returns 0`, which makes all 14 tables complete instantaneously. The test isn't wrong about what it measures; it's measuring a scenario the mocks created.

### Opening the production class for a test seam

`class SyncManager` → `open class SyncManager` plus `protected open fun emitSyncStatus` widens the inheritance surface of a `@Singleton` orchestrator so a test can subclass it. That's a real concession, and it's inconsistent even within the diff: Phase 1's initial write and the parallel writes route through `emitSyncStatus`, but Phase 2 (`:174`) and Phase 4 still assign `_syncStatus.value` directly. A subclass therefore sees a partial picture of status writes, which will mislead the next person who uses this hook.

`val emissionCount = AtomicInteger(0)` on `SyncStatusCoalescer` is also public production state that only tests read.

### What I'd suggest instead

If the goal is to reduce UI churn during parallel phases — a reasonable goal — do it on the **consumer** side, where it's a one-liner and has no concurrency surface at all:

```kotlin
syncManager.syncStatus
    .sample(150)        // or .conflate() / .debounce(150)
    .collect { … }
```

That throttles what the UI redraws without touching `SyncManager`, without opening the class, without a new concurrent utility, and without adding latency to the terminal value (which `sample` still delivers on completion). If instead the goal is to reduce *writes*, the honest answer is that `StateFlow` already makes them free.

**Verdict:** needs a change. I'd drop `SyncStatusCoalescer` and the `open`/`protected` widening, and move throttling to the collector — or, if there's profiling showing the writes actually cost something, share it, because that would change my read of this entirely.

---

dogi:

---

dogi:
Thanks @dogi — agreed on all four points, and @Okuro3499's instructions hadn't landed as a review thread yet (no inline/review comments were present), so I followed your review. Pushed a rewrite in 2abc260.

**What changed:**

- Dropped `SyncStatusCoalescer` (and its test) entirely.
- Reverted `SyncManager` to a final `class` — removed the `open` modifier, the `protected open fun emitSyncStatus` test seam, and the `SYNC_STATUS_COALESCE_MS` companion. Phase 1 + Phase 3 are back to direct `_syncStatus.value =` writes. `SyncManager` is now byte-for-byte its pre-PR state.
- Moved throttling to the one collector that does real per-`Syncing` redraws — `SyncActivity`'s progress dialog — via `.sample(SYNC_STATUS_SAMPLE_MS)` (150 ms) on the collected `syncManager.syncStatus` flow. `DashboardActivity` only reacts to `Success`, so it's left alone.

**On your specific points:**

- *StateFlow already conflates* — you're right; that's exactly why the old test had to subclass `SyncManager` and override a `protected` hook to count writes. The seam is gone now.
- *Costs more than it saves* — removed; no more per-report coroutine launches / cancellations.
- *Data race on `pending`* — removed with the class.
- *Coalesces nothing in production* — removed; no more 150 ms latency on every seconds-apart progress update.
- *Opening the production class for a test seam* — reverted; `SyncManager` is final again.

On the consumer side: `.sample()` on a hot `StateFlow` emits the latest value at each tick and never permanently drops a value, so terminal `Success`/`Error` states still land (at most ~150 ms later). No new concurrent utility, no inheritance surface. The diff is now just `SyncActivity.kt` (+6/-1).

Verified with `./gradlew :app:testDefaultDebugUnitTest --tests "org.ole.planet.myplanet.services.sync.*"` (all pass) and `:app:compileDefaultDebugKotlin`.

No profiling to share — so per your verdict this is the consumer-side throttle path, not the "writes actually cost something" path.

_This reply was written by an AI agent (OpenHands) on behalf of the repository contributor._

dogi:
## Review: 85/100 — mergeable

This is a much better answer to #16468 than the earlier producer-side coalescer, and it addresses every objection I raised against that version.

**The layer is right now.** Throttling belongs on the collector, not in `SyncManager`. The previous approach added shared mutable state (`pending`) inside the producer with no synchronization, and fought `StateFlow`'s own conflation. A single `.sample()` in front of `collectWhenStarted` introduces **no shared state, no race, and no change to `SyncManager` at all** — the burst is absorbed where the cost actually is, at the UI redraw.

**Terminal statuses aren't lost.** The failure mode to check with `sample` is a dropped final value: `sample` emits the latest value per window, so a value arriving just before the flow ends can vanish. That can't bite here — `syncStatus` is a `StateFlow`, which never completes, so a trailing `Success`/`Error` is emitted at the next tick rather than discarded. The cost is up to 150 ms of latency on dialog dismissal and the completion toast, which is imperceptible. Worth stating explicitly in the KDoc, though, since "sample can swallow the last emission" is the first thing a reviewer should worry about.

**150 ms is a sensible number**, it's a named constant, and it carries a comment explaining what it's for. That's the right shape.

**Re-subscription is fine:** `collectWhenStarted` rebuilds the chain on every foreground transition, so each gets a fresh sample window. The current status is then delivered up to 150 ms after resume — again imperceptible for a progress dialog.

### Notes

- **`lastSyncStatus` is now dead weight.** `MutableStateFlow` doesn't emit when assigned an equal value, and `sample` only emits when something new arrived since the last tick — so `if (status == lastSyncStatus) return@collectWhenStarted` can never fire. It predates this PR, but this is the natural moment to drop it; leaving it suggests duplicates are expected when they aren't.
- **`@OptIn(ExperimentalCoroutinesApi::class)` is applied to the whole abstract class** for one operator in one line. Scoping it to `onCreate` (or the flow expression) keeps the opt-in from silently covering future additions to a class this large.
- **No test.** Fair — the change lives in an abstract `Activity.onCreate`, and testing it would mostly be testing `kotlinx.coroutines`. If you wanted one cheaply, extracting `syncManager.syncStatus.sample(SYNC_STATUS_SAMPLE_MS)` into an internal val would let a `runTest` assert that a burst of N statuses inside one window yields a single emission, on virtual time.

**Verdict:** mergeable.

---

Okuro3499 (changes requested):
```
File: app/src/main/java/org/ole/planet/myplanet/ui/sync/SyncActivity.kt:35 and the import kotlinx.coroutines.ExperimentalCoroutinesApi at line 34.                               
  - Problem: sample() requires opting into kotlinx.coroutines.FlowPre
```

olevim:

olevim:
Thanks @Okuro3499 — addressed in 8f114e0.

You're right: `sample(periodMillis: Long)` is annotated `@FlowPreview` in kotlinx-coroutines 1.11.0, not `@ExperimentalCoroutinesApi`. The previous `@OptIn(ExperimentalCoroutinesApi::class)` opted into the wrong marker.

Swapped the `@OptIn` (and its import) to `FlowPreview::class` in `SyncActivity`. This also matches the three sibling files in the same package that already opt into `FlowPreview` for `sample()`/`debounce()` (`RealtimeSyncMixin`, `LoginActivity`, `GuestLoginExtensions`).

@dogi's review of the revised approach (the collector-side `.sample(150L)`) was 85/100 mergeable, so no further structural change was warranted — only this opt-in correction.

dogi:
## Review — rating 86/100 (was 85)

Same approach as before, now with the magic number promoted to a documented constant and `@OptIn(FlowPreview::class)` hoisted to the class. Sampling on the collector side is the right call — it leaves `SyncManager` free to publish status as often as it likes and puts the redraw budget where the redraws actually happen.

The one thing that makes `sample` risky is that it drops a pending value when the upstream completes, so a terminal status can vanish. I traced it and that cannot happen here:

- `syncStatus` is a `StateFlow`, which never completes. The only termination is collector cancellation at lifecycle STOP, and `collectWhenStarted` re-subscribes on START, at which point the StateFlow immediately replays its current value.
- `SyncStatus.Success` (set in `destroy()`) and `SyncStatus.Error` are *resting* states, not transients — nothing overwrites them on a timer.
- The only thing that clears them is `resetSyncStatus()`, and its sole callers are `SyncActivity.kt:183` and `:190`, both **inside the status collector's own `when` branches**. The reset is caused by observing the terminal state, so it cannot race ahead of it.

`sample` also only re-emits when a new value arrives, so an idle `StateFlow` does not generate a redraw every 150 ms, and the existing `lastSyncStatus` guard still filters duplicates on top of that.

Moving to `ready`.

---

Okuro3499 (approved):

PR #16514, opened 2026-08-27
courses: smoother rating utils factoring (fixes #16491)

### LINKED ISSUE #16491: Deduplicate the shared display logic in the two CourseRatingUtils.showRating overloads
**rating:** 46/100 &nbsp;·&nbsp; **proposed by:** glm(r3)

Both overloads end with the identical three display lines (average text, rating-count string, ratingBar value) differing only in how the values were sourced, so a rating-display tweak has to be made twice.

**Ask:** extract the shared three-line display tail into one private helper both `showRating` overloads call, so the display logic exists in a single place.

**Files:** `utils/CourseRatingUtils.kt`

**Verification:** File exists; the duplicated triple was not line-verified.

dogi:
Extracted the common 3-line display tail in both CourseRatingUtils.showRating overloads into a private helper method renderRating and added unit tests in CourseRatingUtilsTest.

Fixes #16491

---
*PR created automatically by Jules for task [3368059556704297610](https://jules.google.com/task/3368059556704297610) started by @dogi*


dogi:
## Review: deduplicate shared display logic in `CourseRatingUtils.showRating`

**Rating: 89/100** — a clean extraction with no behavioural surface, plus coverage for the overload that had none.

### Correctness

The two `showRating` overloads differ only in how they *derive* `averageRating` / `totalRatings` / `userRating` — one from a `RatingSummary`, the other from a `when` over its inputs. The final three statements were byte-identical in both, and `renderRating` is those three statements moved verbatim: same order, same `?: 0f` / `?: 0` defaults, same `userRating ?: averageRating ?: 0f` precedence chain. Nothing else in either overload reads or writes the views after the call, so extraction is behaviour-preserving.

This is the good kind of deduplication: the duplicated block was the part most likely to drift (three view writes with a specific fallback chain), and now a change to the fallback logic can't apply to one overload and not the other.

### Tests

The `RatingSummary` overload previously had no direct coverage — the existing tests exercised the other one. Both new cases assert through the shared path, and `userRating = 5` with `averageRating = 4.2f` is a well-chosen pair because it pins the precedence: the bar takes the *user's* rating, not the average. A test where they're equal would have proved nothing.

### Nits

- `org.ole.planet.myplanet.repository.RatingSummary` is written out fully-qualified twice, including the awkward `null as org.ole.planet.myplanet.repository.RatingSummary?` disambiguation cast. Importing `RatingSummary` turns that into `null as RatingSummary?` and makes the constructor call readable. The cast itself is necessary — with two overloads a bare `null` is ambiguous — so only the qualification is worth changing.
- `String.format(Locale.getDefault(), "%.2f", …)` means `average.text = "4.20"` only holds in locales using `.` as the decimal separator; under `de_DE` it would be `"4,20"` and both new assertions would fail. Robolectric defaults to `en_US` so CI is stable, and this is pre-existing (the older `"0.00"` assertion has the same exposure) — but if the suite ever runs under a different default locale these will be the first to break. Passing an explicit `Locale.US` in the test setup, or asserting with a locale-derived expected string, would close it.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16516, opened 2026-08-27
sync: smoother download service queue caching (fixes #16485)

### LINKED ISSUE #16485: Read the download-queue sets and file name once in onDownloadComplete
**rating:** 47/100 &nbsp;·&nbsp; **proposed by:** deepseek(r3), devin(r2), glm(r2)

`onDownloadComplete` reads PRIORITY_DOWNLOADS_KEY (:502) and then calls `getRemainingCount()` (:503), which reads both PRIORITY and PENDING again and allocates a union set — so the priority set is read twice per completion. `getFileNameFromUrl(url)` (a URI parse plus URLDecoder.decode) is also called twice with the same url.

**Ask:** read the priority set once and reuse it (rather than having `getRemainingCount()` re-read it), and call `getFileNameFromUrl(url)` once per completion instead of twice.

**Files:** `services/DownloadService.kt`

**Verification:** Verified verbatim at DownloadService.kt:176-180 and 500-516. glm covers both the double set read and the double filename parse; deepseek only the sets; devin only the union allocation inside getRemainingCount.

dogi:
## Summary

`onDownloadComplete` was reading the priority download-queue set twice per completion and parsing `getFileNameFromUrl(url)` twice:

- `remainingPriority` read `PRIORITY_DOWNLOADS_KEY` (DownloadService.kt:502), then `getRemainingCount()` read both `PRIORITY_DOWNLOADS_KEY` and `PENDING_DOWNLOADS_KEY` again (DownloadService.kt:177-178) — so the priority set was read twice.
- `getFileNameFromUrl(url)` (a URI parse + `URLDecoder.decode`) was called at DownloadService.kt:506 and again at DownloadService.kt:515 with the same url.

## Changes

- `getRemainingCount()` now takes an optional `priorityUrls` parameter; when supplied it reuses the already-read priority set instead of re-reading it. Existing call sites (processDownloadQueue, the batch-download notification, sendNotification) keep working with the default `null`.
- `onDownloadComplete` reads the priority set once into `priorityUrls`, derives both `remainingPriority` and `remaining` (via `getRemainingCount(priorityUrls)`) from it, and computes `fileName = getFileNameFromUrl(url)` once, reusing it for both `Download.fileName` and the notification content text.

Closes #16485.

## Verification

- Added a regression test `DownloadServiceOnDownloadCompleteTest` that invokes `onDownloadComplete` against a spied `DownloadService` with a mocked `SharedPreferences` and mocked `FileUtils`, then asserts `getStringSet(PRIORITY_DOWNLOADS_KEY, …)` is read exactly once and `getFileNameFromUrl(url)` is called exactly once, and that the broadcast `Download` carries the decoded filename.
- Verified the test fails against the original code (priority set read 2×) and passes with this change.
- Ran the existing `DownloadServiceTest` and `DownloadServiceUrlSelectionTest` — all still pass (17 tests total, 0 failures).

```
DownloadServiceOnDownloadCompleteTest ... 1 test, 0 failures
DownloadServiceTest ..................... 6 tests, 0 failures
DownloadServiceUrlSelectionTest ........ 10 tests, 0 failures
```

_Note: this PR was created by an AI agent (OpenHands) on behalf of the maintainer._

@dogi can click here to [continue refining the PR](https://app.all-hands.dev/conversations/90678d50-b8da-4529-a88c-7310c42a382f)


dogi:
## Review: read queue sets and filename once in `onDownloadComplete`

**Rating: 84/100** — the production change is a genuine consistency fix as well as a saving, and unusually for this queue, the tests actually fail on `master`. The test's mechanism is the weak part.

### Correctness

Three changes, all sound:

1. **`getRemainingCount(priorityUrls: Set<String>? = null)`** — the default keeps every existing caller working unchanged (it re-reads as before), and `onDownloadComplete` now passes the set it already read.
2. **One `PRIORITY_DOWNLOADS_KEY` read instead of two.** This is worth more than the saved lookup: previously `remainingPriority` and `getRemainingCount()` each read the set independently, so a concurrent write between them could produce a `remainingPriority` and a `remaining` describing *different* queue states — and those two values jointly decide `completeAll`. Reading once makes the pair consistent by construction. That's the real win, and the PR body undersells it as a perf tweak.
3. **`getFileNameFromUrl(url)` hoisted.** `this.fileName = fileName` inside the `apply` block is the correct disambiguation once a local of the same name exists — easy to get subtly wrong (a bare `fileName = fileName` would have been a self-assignment on the local), and it's right here.

### Tests

Both assertions are discriminating: `verify(exactly = 1)` on the priority-set read and on `getFileNameFromUrl` would each see **2** on `master`. That's the property I want and it's frequently missing — good.

`@Config(sdk = [34])` is fine: 34 is present in `robolectricSdkJars` (`34: '14-robolectric-10818077-i7'`), so the `-ProbolectricOffline=true` staging CI uses resolves it. Note the sibling `DownloadServiceTest` writes the same level as `Build.VERSION_CODES.UPSIDE_DOWN_CAKE` — worth matching so the two files for the same class don't look like they target different SDKs.

### The brittleness

The test reaches into Kotlin's compiled internals:

```kotlin
val prefsField = DownloadService::class.java.getDeclaredField("preferences\$delegate")
prefsField.set(service, kotlin.lazyOf(mockPreferences))
```

`preferences$delegate` is a synthetic field whose name is a compiler implementation detail of `by lazy`. If `preferences` ever stops being a lazy delegate — or is renamed, or the property is refactored to constructor injection — this fails with `NoSuchFieldException` rather than a compile error, and the message won't point anywhere useful. That's on top of seven more private fields and one private method reached reflectively.

It works, and I can see why: `DownloadService` is a `Service` with `lateinit` injection and no seam. But the amount of machinery here is a signal about the class, not the test. `onDownloadComplete`'s logic — given a priority set, a pending set, a processed set, and a session count, compute `remaining`, `remainingPriority`, and `completeAll` — is pure. Lifting that into an internal function taking those four inputs would let this test drop every reflection call and assert on returned values instead of on mock interactions, and the service would just call it. Worth considering as a follow-up rather than a blocker.

### Nit

`android.util.Log` is referenced fully-qualified five times in `setUp` while everything else is imported.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16519, opened 2026-08-27
chat: smoother clipboard caching (fixes #16518)

### LINKED ISSUE #16518: Cache the ClipboardManager lookup in the chat adapter

`ChatAdapter.copyToClipboard` does `context.getSystemService(CLIPBOARD_SERVICE) as ClipboardManager` on every long-press of a chat message.

**Ask:** cache the `ClipboardManager` lookup instead of resolving the system service on every long-press.

**Files:** `ui/chat/ChatAdapter.kt`

dogi:
## Summary

`ChatAdapter.copyToClipboard` resolved the `ClipboardManager` system service on every long-press of a chat message. This caches the lookup behind a `lazy` property so the service is resolved only once per adapter instance.

Fixes #16518

## Changes

- `ui/chat/ChatAdapter.kt`: added a `clipboardManager` `by lazy` property and updated `copyToClipboard` to use it instead of calling `context.getSystemService(CLIPBOARD_SERVICE)` each time.
- `ui/chat/ChatAdapterTest.kt`: added a Robolectric test that binds a query holder, triggers the long-press copy twice, and asserts the expected text lands on the clipboard.

## Verification

The runtime lacks a JDK, so the unit tests could not be run locally in this session. CI (`test.yml`) will run `testDefaultDebugUnitTest` on push.

## Notes

@dogi can click here to [continue refining the PR](https://app.all-hands.dev/conversations/c9e25d01-5126-4edb-abc9-55cc37d15224)


dogi:
## Review: cache the `ClipboardManager` lookup in `ChatAdapter`

**Rating: 84/100** — correct, introduces no new leak (which is the thing worth checking here), and the tests actually fail on `master`.

### The leak question

Caching a system service on an adapter is the kind of change that deserves a second look, because `ClipboardManager` is fetched through a `CachedServiceFetcher` and holds a reference back to the `Context` it was created from. So `ChatAdapter` → `ClipboardManager` → `Context` is a new retention edge.

It's harmless here only because `ChatAdapter` already holds `context` directly as a constructor property. The adapter's lifetime already pins that `Context`; the lazy field adds a second path to something already reachable, not a new root. If the adapter were ever changed to take an application context or a `WeakReference`, this field would quietly reintroduce the strong reference — worth a comment on the property saying it deliberately shares the constructor `context`'s lifetime.

`by lazy` defaults to `SYNCHRONIZED`, and `copyToClipboard` runs on the main thread from a long-press, so there's no initialisation race. The `as ClipboardManager` cast is still deferred to first use, exactly as before.

### Value

Honest framing: `getSystemService` is a map lookup into `SystemServiceRegistry` plus a cached-fetcher array index, and `copyToClipboard` fires once per user long-press. This is not a hot path and the saving isn't measurable. What the change actually buys is that the lookup-and-cast noise is out of the method body, leaving `copyToClipboard` to express only what it does. That's a fine reason on its own — it just isn't a performance fix, and the PR body reads as though it is.

### Tests

`copyToClipboard resolves the clipboard service only once across copies` is the good kind: `CountingClipboardContext` counts `getSystemService(CLIPBOARD_SERVICE)` calls and asserts exactly **1** after three copies, where `master` would give **3**. It fails without the change, which is the property most of these caching PRs miss. Returning a stable cached instance from the wrapper so the adapter and the assertion see the same `ClipboardManager` is the right detail to have got right.

`@Config(sdk = [33])` resolves — 33 is present in `robolectricSdkJars` (`33: '13-robolectric-9030017-i7'`), so the offline SDK staging CI uses covers it.

### Nits

- Both tests reach `copyToClipboard` by reflection. It's one stable private method so the brittleness is low, but `internal` would remove the reflection entirely and cost nothing — the class is already `internal`-visible to the test source set.
- The second test rebuilds the `Method` handle rather than reusing the existing `invokeCopyToClipboard` helper defined a few lines above, so the same reflection is written twice in one file.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16520, opened 2026-08-27
sync: smoother auto worker time providing (fixes #16490)

### LINKED ISSUE #16490: Use the injected clock for AutoSyncWorker's sync-timestamp writes
**rating:** 46/100 &nbsp;·&nbsp; **proposed by:** glm(r3)

AutoSyncWorker injects timeProvider and reads it at :55, but writes timestamps back with `Date().time` at :139 and :157 — defeating the injectable clock the test layer relies on and keeping java.util.Date in the sync path.

**Ask:** replace the two `Date().time` writes with the already-injected `timeProvider`, so the worker's timestamp writes go through the injectable clock and `java.util.Date` leaves the sync path.

**Files:** `services/AutoSyncWorker.kt`

**Verification:** Verified all three lines and that they are the file's only Date references.

dogi:
Replaced direct Date().time calls with the injected timeProvider in AutoSyncWorker's timestamp write methods (setLastSync and setLastUsageUploaded), and removed the unused java.util.Date import.

Fixes #16490

---
*PR created automatically by Jules for task [4292828048108967534](https://jules.google.com/task/4292828048108967534) started by @dogi*


dogi:
## Review: use the injected `timeProvider` for `AutoSyncWorker` sync timestamps

**Rating: 88/100** — finishes a conversion that was left half-done, and the substitution is exact.

### Correctness

`SystemTimeProvider.now()` is `System.currentTimeMillis()`, and `Date().time` is the same call behind an allocation. So runtime behaviour is identical and the change also drops two short-lived `Date` objects.

The important detail is that `timeProvider` was **already** a constructor-injected property of this worker (`:45`) and already in use at `:55`:

```kotlin
val currentTime = timeProvider.now()
```

So before this PR the worker read time from two different sources: an injectable one for its scheduling decision, and `Date()` for the two timestamps it persists. That's worse than either choice on its own — under a test that pins the clock, `currentTime` would be the fake value while `setLastSync` recorded the real one, which is exactly the sort of skew that makes a "did we sync recently?" assertion pass for the wrong reason. Routing all three through the same provider makes the worker's notion of "now" single-valued.

Removing the now-unused `java.util.Date` import is correct — there are no other `Date` references in the file.

### Follow-up worth naming

`SyncManager:117` writes the **same preference** with the same expression:

```kotlin
sharedPrefManager.setLastSync(Date().time)
```

and `SyncManager` already injects a `TimeProvider` too. So `lastSync` is currently written from two places, one of which is now clock-injectable and one of which isn't — which means a test that pins the clock still can't assert on that preference end-to-end. Same one-line change; worth doing in this PR or immediately after, since leaving it makes the codebase harder to reason about than either state alone.

### Gap

No test. The whole value of this change is that the timestamp becomes assertable, and nothing asserts it. A `TestTimeProvider(currentTime = 12345L)` plus `verify { sharedPrefManager.setLastSync(12345L) }` would be a handful of lines and would lock in the property the PR exists to create. `TestTimeProvider` already exists in the test source set and is used elsewhere, so there's no infrastructure to build.

Not blocking — the change is a strict improvement over `master` either way — but a conversion whose only purpose is testability should probably arrive with the test.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16521, opened 2026-08-27
courses: smoother repository sorting (fixes #16512)

### LINKED ISSUE #16512: Sort filtered courses without allocating lowercase titles

`CoursesRepositoryImpl.getCoursesBySearchAndFilter` sorts with `it.courseTitle?.lowercase() ?: ""`, allocating a lowercase string during comparisons for every retained course; a case-insensitive comparator preserves the ordering without the copies.

**Ask:** replace the `lowercase()`-per-comparison sort key with a case-insensitive comparator, preserving the existing ordering without allocating a lowercase copy per course.

**Files:** `repository/CoursesRepositoryImpl.kt`

dogi:
## Summary

`CoursesRepositoryImpl.filterCourses` (the `getCoursesBySearchAndFilter` path) sorted the retained courses with `sortedBy { it.courseTitle?.lowercase() ?: "" }`, which allocates a fresh lowercase `String` for every retained course on each comparison. This replaces that sort key with a case-insensitive comparator, preserving the existing ordering without allocating a lowercase copy per course.

## Changes

- `repository/CoursesRepositoryImpl.kt`: `.sortedBy { it.courseTitle?.lowercase() ?: "" }` → `.sortedWith(compareBy(String.CASE_INSENSITIVE_ORDER) { it.courseTitle ?: "" })`. This mirrors the pattern already used in `ui/surveys/SurveysViewModel.kt` for `SortOption.TITLE_ASC`.
- `repository/CoursesRepositoryImplTest.kt`: added `filterCourses sorts titles case-insensitively`, which feeds mixed-case titles (`banana`, `Apple`, `cherry`) and asserts the output is ordered `Apple`, `banana`, `cherry`.

## Verification

- `./gradlew testDefaultDebugUnitTest --tests "org.ole.planet.myplanet.repository.CoursesRepositoryImplTest"` — BUILD SUCCESSFUL, all 16 tests pass (0 failures), including the new case-insensitive sort test.

Fixes #16512.

---

@dogi can click here to [continue refining the PR](https://app.all-hands.dev/conversations/76746f1f-06bd-4aa1-9c74-44486f336297)


dogi:
## Review: sort filtered courses without a per-item `lowercase()`

**Rating: 87/100** — a bigger saving than the title suggests, with one semantic nuance worth recording.

### The saving is per *comparison*, not per item

`sortedBy { selector }` is `sortedWith(compareBy(selector))`, and `compareBy` invokes the selector **on every comparison**, not once per element. So `it.courseTitle?.lowercase()` was allocating a new `String` roughly `2 · n · log₂ n` times per sort — for a 200-course library that's ~3,000 throwaway strings, on a path that reruns on every debounced search keystroke and every grade/subject/tag filter change.

`compareBy(String.CASE_INSENSITIVE_ORDER) { it.courseTitle ?: "" }` keeps the same per-comparison selector call, but the selector is now just a null check and the case-insensitivity happens inside the comparator without allocating. That's the whole win, and it's a real one — worth stating in the PR body, because "per-item lowercase" undersells it by a log factor.

`.toList()` at the end is still required (the chain is a `Sequence`), and null/empty titles still sort first via `?: ""` in both versions.

### One nuance: the two orderings aren't identical for all input

`String.lowercase()` (no argument) applies full Unicode case mapping under `Locale.ROOT`, which can change string *length* — e.g. `İ` (U+0130) lowercases to `i` + combining dot above, two code units. `String.CASE_INSENSITIVE_ORDER` compares char-by-char via `Character.toUpperCase`/`toLowerCase`, so it never changes length and handles those characters differently.

For ASCII titles the two orderings agree exactly. They can diverge for titles containing characters with non-1:1 case mappings — relevant here because the app ships Arabic, Spanish, French, Nepali and Somali locales and course titles come from the server. In practice this affects sort *position* of a rare title, not correctness of the list contents, and `CASE_INSENSITIVE_ORDER` is the more conventional choice for a display sort. Just worth a line in the PR body so it's a recorded decision rather than an accident.

(If locale-correct ordering ever matters here, `Collator.getInstance()` is the real answer — but it's slower than either of these, so I'd only reach for it if users report mis-sorted non-Latin titles.)

### Test

Covers the intended behaviour (`Apple` / `banana` / `cherry` interleaving cases). Note it passes on `master` too — `sortedBy { lowercase() }` produces the same order for ASCII — so it's a characterisation test locking in behaviour across the refactor rather than a regression guard for the change itself. That's the right kind of test to add here; just don't read a green run as evidence the comparator swap was exercised. A case where the two genuinely differ would be hard to write and not worth it.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16525, opened 2026-08-27
login: smoother activities repository bulk inserting (fixes #16511)

### LINKED ISSUE #16511: Collect the sync activity lookup keys in a single pass

`ActivitiesRepositoryImpl.insertLoginActivitiesFromSync` walks documentList three separate times with map/filter/distinct pipelines to derive remote IDs, login times and user names before two DAO lookups.

**Ask:** derive the remote IDs, login times and user names in a single pass over `documentList` instead of three separate map/filter/distinct walks.

**Files:** `repository/ActivitiesRepositoryImpl.kt`

dogi:
## Summary

`ActivitiesRepositoryImpl.insertLoginActivitiesFromSync` previously walked `documentList` three separate times with `map`/`filter`/`distinct` pipelines to derive the remote IDs, login times, and user names before the two DAO lookups. This derives all three lookup-key sets in a single pass over `documentList`, preserving the distinctness and non-empty filtering of each set.

## Changes

- `repository/ActivitiesRepositoryImpl.kt`: replaced the three `map`/`filter`/`distinct` chains with one `for` loop that populates `LinkedHashSet`s for ids, login times, and user names (LinkedHashSet keeps insertion order while deduping, matching the prior `distinct()` behavior).
- `ActivitiesRepositoryImplTest.kt`: added tests covering the single-pass collection, deduplication, design-doc/empty-key filtering, early return, and the fallback-lookup skip path.

## Verification

- `./gradlew testDefaultDebugUnitTest --tests "org.ole.planet.myplanet.repository.ActivitiesRepositoryImplTest"` — all 34 tests pass.

Closes #16511

_Note: this PR was created by an AI agent (OpenHands) on behalf of the repository maintainer._

@dogi can click here to [continue refining the PR](https://app.all-hands.dev/conversations/23389daa-57de-4007-b252-82051b1dca62)


dogi:
## Review: collect sync activity lookup keys in a single pass

**Rating: 89/100** — behaviour-preserving, and the ordering equivalence (the part that could have broken) holds exactly.

### Correctness

Three chains of `documentList.map { … }.filter { … }.distinct()` collapse into one loop filling three `LinkedHashSet`s.

The ordering is the thing worth checking, because both DAO calls assert on list *contents in order* downstream. `distinct()` is implemented over a `LinkedHashSet` and keeps first-occurrence order; `LinkedHashSet.add` in iteration order gives the same sequence. So `ids`, `loginTimes` and `userNames` come out identical, element for element, to what `master` produced.

The allocation difference is larger than one traversal: each old chain built three intermediate lists (map → filter → distinct), so nine lists across three full passes of `documentList`. Now it's one pass, three sets, and three `.toList()` copies at the call sites. On a sync batch of a few hundred login documents that's a real reduction.

One reordering to note and dismiss: `loginTimes`/`userNames` are now computed *before* the `getByRemoteIds` DAO call rather than after it. `JsonUtils.getLong` and `getString` are pure reads over the already-parsed `JsonObject`s, so hoisting them across the suspend call changes nothing observable.

### Tests

Five tests on a method that had none, and they're well chosen: dedup across repeated documents, design-doc/empty-id skipping, the early return, and the "fallback lookup skipped when `loginTime`/`user` are absent" branch — that last one is the easiest to break in a future edit and the least obvious from reading the code.

`coVerify(exactly = 1) { getByRemoteIds(listOf("a1", "a2")) }` asserts the exact list *and its order*, which is what makes these more than smoke tests — a `HashSet` instead of `LinkedHashSet` would fail them. Worth noting they also pass on `master` (the old chains produce the same lists), so they're locking behaviour in across the refactor rather than proving the refactor fixed something.

### Pre-existing, not this PR

`getByRemoteIds(ids)` and `getByLoginTimesAndUserNames(loginTimes, userNames)` bind one SQL variable per element. A large sync batch could push those past SQLite's parameter limit, and `UserDao.getUsersByIds` already chunks at 400 for exactly this reason. Unchanged here — the sets are the same size the lists were — but this PR is the natural place from which to notice it, and a chunked variant would be a good follow-up.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16526, opened 2026-08-27
all: smoother version utils testing (fixes #16504)

### LINKED ISSUE #16504: Split VersionUtilsTest so the pure JVM tests leave Robolectric
**rating:** 44/100 &nbsp;·&nbsp; **proposed by:** copilot(r3)

VersionUtilsTest costs 5.7s for the class and mixes pure string compare/parse tests with Context/PackageManager tests under @RunWith(RobolectricTestRunner::class) — the pure tests pay the sandbox cost for nothing, and production already separates the two groups.

**Ask:** split the pure string compare/parse tests out into a new plain-JUnit `VersionUtilsPureTest` (no Robolectric runner), leaving only the Context/PackageManager tests in the Robolectric-backed `VersionUtilsTest`.

**Files:** `test/.../VersionUtilsTest.kt`, `(new) test/.../VersionUtilsPureTest.kt`

**Verification:** The file exists and mixes both kinds; the 5.7s figure is from the agent's CI log reading, which I could not re-run.

dogi:
Extracted pure string comparison and version parsing tests from VersionUtilsTest into a plain JUnit test class VersionUtilsPureTest (removing RobolectricTestRunner). Retained Context and PackageManager tests in VersionUtilsTest.

Fixes #16504

---
*PR created automatically by Jules for task [2642459091247354070](https://jules.google.com/task/2642459091247354070) started by @dogi*


dogi:
## Review: split `VersionUtilsTest` into `VersionUtilsPureTest`

**Rating: 90/100** — a pure move that buys real CI wall-clock, and the split line is drawn in the right place.

### The split is correct

Everything moved — `compareVersions`, `isVersionAllowed`, `parseApkVersionString` — operates on `String`s and touches no Android API. `VersionUtils` is an `object` whose initialiser is trivial, and its `android.os.Build` / `PackageManager` references live inside `getVersionCode`/`getVersionName`, which stay behind in the Robolectric class. So `VersionUtilsPureTest` genuinely runs as plain JUnit4 with no sandbox, and the two tests that need a framework (`getVersionCode_should_return_0_on_NameNotFoundException`, and the `@Config(sdk = …)` cases pinning `Build.VERSION_CODES`) keep their runner.

The bodies are moved verbatim, including the explanatory comments — the one about `compareVersions` parsing *every* part so `"2.0"` vs `"1.x"` still throws, and the one documenting the `0.0.12` → `0012` → `12` collision in `parseApkVersionString`, are both worth preserving and were. Import cleanup on the remaining file matches what's still used.

### Why it's worth it

Robolectric builds a sandbox and loads an `android-all-instrumented` jar per test class. Thirteen assertions over string comparison were paying that cost for nothing. With CI running `test.yml` in two shards specifically to cut wall clock — and `-ProbolectricOffline=true` staging SDK jars precisely because Robolectric startup was a measured pain point — moving pure tests out of the sandbox is directly aligned with how this repo has decided to spend its CI budget.

It also sets a pattern worth repeating: several other classes in this suite mix pure and framework-dependent tests under one `@RunWith(RobolectricTestRunner::class)`.

### Conflict to sequence — this one is concrete

#16503 (`version: smoother segment compare without boxed ints`) adds two tests to `VersionUtilsTest.kt`:

- `compareVersions_should_compare_multi_digit_segments_numerically_not_lexicographically`
- `compareVersions_should_order_by_segment_count_when_prefix_matches`

inserted immediately after `compareVersions_should_not_throw_on_insufficient_version_parts` — a method **this PR deletes from that file**. So the two will conflict textually, and resolving it by keeping both sides verbatim would be wrong: those two new tests are pure `compareVersions` assertions and belong in `VersionUtilsPureTest`, not in the Robolectric class.

Cleanest order is to land this split first, then rebase #16503 and have it add its two tests to `VersionUtilsPureTest`. Doing it the other way round means resolving the conflict by hand and remembering to relocate them.

### Note on the label

This is tagged `enormous`, but it's a move — the size labeller sums additions and deletions, so relocating N lines counts as 2N. Not something to change (the rule is deliberate and documented), just worth knowing the label overstates the review burden here: the substantive diff is zero.

**Verdict:** mergeable.

---

Okuro3499 (dismissed):
```
android.content.pm.PackageManager$NameNotFoundException
	at org.ole.planet.myplanet.utils.VersionUtilsTest.getVersionName_should_return_empty_string_on_NameNotFoundException(VersionUtilsTest.kt:74)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.robolectric.RobolectricTestRunner$HelperTestRunner$1.evaluate(RobolectricTestRunner.java:524)
	at org.robolectric.internal.SandboxTestRunner.executeInSandbox(SandboxTestRunner.java:494)
	at org.robolectric.internal.SandboxTestRunner.access$900(SandboxTestRunner.java:67)
	at org.robolectric.internal.SandboxTestRunner$7.evaluate(SandboxTestRunner.java:442)
	at org.junit.runners.BlockJUnit4ClassRunner$1.evaluate(BlockJUnit4ClassRunner.java:100)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:366)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:103)
	at org.robolectric.internal.SandboxTestRunner.access$600(SandboxTestRunner.java:67)
	at org.robolectric.internal.SandboxTestRunner$6.evaluate(SandboxTestRunner.java:333)
	at org.robolectric.internal.SandboxTestRunner$3.evaluate(SandboxTestRunner.java:233)
	at org.robolectric.internal.SandboxTestRunner$5.lambda$evaluate$0(SandboxTestRunner.java:317)
	at org.robolectric.internal.bytecode.Sandbox.lambda$runOnMainThread$0(Sandbox.java:101)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:317)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
	at java.base/java.lang.Thread.run(Thread.java:1583)
android.content.pm.PackageManager$NameNotFoundException
	at org.ole.planet.myplanet.utils.VersionUtilsTest.getVersionCode_should_return_0_on_NameNotFoundException(VersionUtilsTest.kt:25)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.robolectric.RobolectricTestRunner$HelperTestRunner$1.evaluate(RobolectricTestRunner.java:524)
	at org.robolectric.internal.SandboxTestRunner.executeInSandbox(SandboxTestRunner.java:494)
	at org.robolectric.internal.SandboxTestRunner.access$900(SandboxTestRunner.java:67)
	at org.robolectric.internal.SandboxTestRunner$7.evaluate(SandboxTestRunner.java:442)
	at org.junit.runners.BlockJUnit4ClassRunner$1.evaluate(BlockJUnit4ClassRunner.java:100)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:366)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:103)
	at org.robolectric.internal.SandboxTestRunner.access$600(SandboxTestRunner.java:67)
	at org.robolectric.internal.SandboxTestRunner$6.evaluate(SandboxTestRunner.java:333)
	at org.robolectric.internal.SandboxTestRunner$3.evaluate(SandboxTestRunner.java:233)
	at org.robolectric.internal.SandboxTestRunner$5.lambda$evaluate$0(SandboxTestRunner.java:317)
	at org.robolectric.internal.bytecode.Sandbox.lambda$runOnMainThread$0(Sandbox.java:101)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:317)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
	at java.base/java.lang.Thread.run(Thread.java:1583)
```

dogi:
## Review — rating 82/100

Good hygiene PR, correctly executed. The split is the interesting part and the detail that matters was handled right.

### The Robolectric split is done correctly

Thirteen tests that exercise pure string/int functions were paying for a Robolectric sandbox they never used. Moving them to plain JUnit is a real if modest CI win given this project already shards tests for wall clock.

The subtle part is which tests had to *stay*. Replacing `e.printStackTrace()` with `Log.w(TAG, …, e)` means `getVersionCode`/`getVersionName` now touch `android.util.Log`, and this project does not set `returnDefaultValues`, so those tests would die with *"Method w in android.util.Log not mocked"* under plain JUnit. They correctly remain in the Robolectric file while only the Log-free tests moved. Loading the `VersionUtils` object from plain JUnit is fine — `TAG` is a compile-time constant and no pure function reaches a `Log` call.

I diffed the moved block against the removed block; it is a verbatim move with nothing dropped, and the now-unused `assertTrue`/`assertFalse` imports were cleaned up.

### The documented bug is worse than the comment says

`parseApkVersionString_should_document_latent_multi_leading_zero_behavior` is a good instinct, but it picks the least alarming example. The implementation is:

```kotlin
var vsn = raw.replace("v", "")
vsn = vsn.replace(".", "")
val cleaned = if (vsn.startsWith("0")) vsn.replaceFirst("0", "") else vsn
return cleaned.toIntOrNull()
```

Stripping dots without preserving field widths means **any single-digit minor or patch shifts the result by a factor of ten**, and this project's own scheme is about to produce one. Per `.github/scripts/version.sh`, `versionCode = major*10000 + minor*100 + patch`:

| version | true versionCode | `parseApkVersionString` |
|---|---|---|
| 0.67.67 (today) | 6767 | 6767 ✅ |
| 0.68.0 (next minor rollover) | 6800 | **680** ❌ |
| 0.67.7 | 6707 | **677** ❌ |

It works right now only because both components happen to be two digits. The next patch rollover to `0.68.0` breaks the `minapk` gate at `ConfigurationsRepositoryImpl:141` by an order of magnitude — the app would compare 680 against a server value of 6800 and conclude it is catastrophically out of date. The collision the test documents (`0.1.2` and `0.0.12` both → 12) is the same root cause, just a less consequential instance.

I would not hold this PR for it — the bug is pre-existing and this PR is the first thing to write it down. But please open an issue, because the test currently asserts the broken behaviour as expected, which is exactly the kind of thing that makes a future fix look like a regression. The fix is to parse the three components and apply `version.sh`'s formula rather than string-mangling.

Nit: `VersionUtilsPureTest` is an odd name next to the repo's `<Class>Test` convention — something like `VersionUtilsComparisonTest` would read better.

Moving to `ready`.

---

Okuro3499 (approved):

PR #16529, opened 2026-08-27
teams: smoother members date formatting (fixes #16524)

### LINKED ISSUE #16524: Hoist the members-list DateTimeFormatter
**rating:** 41/100 &nbsp;·&nbsp; **proposed by:** kimi(r1), kimi(r2)

`MembersAdapter` builds a `DateTimeFormatter.ofPattern(...).withZone(...)` per adapter instance (:34) and formats lastVisitDate at bind (:113); TimeUtils.getFormattedDate(Long?) already exists as the shared formatter for this timestamp.

**Ask:** drop the per-adapter `DateTimeFormatter` and format `lastVisitDate` through the existing shared `TimeUtils.getFormattedDate(Long?)` helper instead.

**Files:** `ui/teams/members/MembersAdapter.kt`

**Verification:** Verified the field at :34, the bind-time format, and that TimeUtils.getFormattedDate(Long?) exists at TimeUtils.kt:79. Same agent proposed it in two rounds with two different fixes — counted once.

dogi:
## Summary

`MembersAdapter` built a `DateTimeFormatter.ofPattern(...).withZone(...)` per adapter instance and formatted `lastVisitDate` at bind time. This drops the per-adapter formatter and routes the timestamp through the existing shared `TimeUtils.getFormattedDate(Long?)` helper, removing duplicated formatting logic.

Fixes #16524.

## Changes

- Removed the per-adapter `dateFormatter` field (`DateTimeFormatter.ofPattern(TimeUtils.DATE_FORMAT).withZone(ZoneId.systemDefault())`) from `MembersAdapter`.
- Removed now-unused imports (`java.time.Instant`, `java.time.ZoneId`, `java.time.format.DateTimeFormatter`).
- Format `lastVisitDate` at bind time via `TimeUtils.getFormattedDate(memberData.lastVisitDate)`.

## Verification

- Added two unit tests to `MembersAdapterTest`:
  - `testOnBindViewHolder_lastVisitDate_usesSharedTimeUtilsFormatter` — asserts the bound `tvLastVisit` text contains the value produced by `TimeUtils.getFormattedDate(timestamp)`.
  - `testOnBindViewHolder_nullLastVisitDate_showsNoVisit` — asserts the `no_visit` string is shown when `lastVisitDate` is null.
- Ran `./gradlew :app:testDefaultDebugUnitTest --tests "org.ole.planet.myplanet.ui.teams.members.MembersAdapterTest"` — all 6 tests pass (4 existing + 2 new), 0 failures.

## Notes

The shared helper uses `TimeUtils`'s `defaultDateFormatter` (`EEEE, MMM dd, yyyy`, UTC), so the displayed format for last-visit dates changes from `dd MMM yyyy` (system zone) to the shared format. This consolidates formatting into the single shared helper as requested by the issue.

@dogi can click here to [continue refining the PR](https://app.all-hands.dev/conversations/c2a03c66-bd1e-48ae-b73e-2df7044e3191)


dogi:
## Review: last-visit date formatting via shared `TimeUtils`

**Rating: 48/100** — this is not the refactor it's presented as. `TimeUtils.getFormattedDate` uses a **different pattern and a different timezone** from the formatter it replaces, so the members list will render visibly different dates — and the new test can't detect it.

### The two formatters are not equivalent

Removed from `MembersAdapter`:

```kotlin
DateTimeFormatter.ofPattern(TimeUtils.DATE_FORMAT).withZone(ZoneId.systemDefault())
// TimeUtils.DATE_FORMAT == "dd MMM yyyy"          →  "11 Mar 2024"
// zone: system default
```

What `TimeUtils.getFormattedDate(date: Long?)` actually uses (`TimeUtils:46-48`, `:79-86`):

```kotlin
private val defaultDateFormatter by lazy {
    DateTimeFormatter.ofPattern("EEEE, MMM dd, yyyy", defaultLocale).withZone(utcZone)
}
// utcZone = ZoneId.of("UTC")                       →  "Monday, Mar 11, 2024"
```

Two independent changes, neither mentioned:

1. **Display format.** `"dd MMM yyyy"` → `"EEEE, MMM dd, yyyy"`. Every row in the joined-members list goes from `11 Mar 2024` to `Monday, Mar 11, 2024` — roughly triple the width, in a `RowJoinedUserBinding` row that also carries name, visit count and role. That's a layout question, not just a cosmetic one.
2. **Timezone.** System-default → **UTC**. For a user at UTC−5, a visit recorded at 20:00 local on 10 March now displays as 11 March. Every member west of UTC will see last-visit dates that disagree with what the user remembers, and it'll be off by one for a chunk of each day.

`TimeUtils.DATE_FORMAT` — the constant the adapter was using — is now unreferenced by this call path, which is the tell that the "shared" function isn't the same thing.

### The test can't catch either change

```kotlin
val expected = TimeUtils.getFormattedDate(timestamp)
assertTrue(viewHolder.binding.tvLastVisit.text.toString().contains(expected))
```

The expectation is computed by calling **the same function under test**. This passes for any format `getFormattedDate` might return, including one nobody intended — it asserts only "the adapter calls this function", never "the adapter renders this text". A literal expectation is what's needed:

```kotlin
assertTrue(viewHolder.binding.tvLastVisit.text.toString().contains("11 Mar 2024"))
```

That version fails on this branch, which is exactly why it's the one worth having. (It would need a fixed timezone/locale in the test to stay stable — which is itself a useful forcing function for deciding what zone this should render in.)

The `no_visit` test is fine and does assert a real string.

### What the fix looks like

If the goal is genuinely to stop hand-rolling a formatter in the adapter — a good goal — `TimeUtils` already has the machinery. `formatterFor(pattern, zone, locale)` is private but caches per `(pattern, zone, locale)` in a `ConcurrentHashMap`, so a small public wrapper preserves today's output exactly:

```kotlin
// TimeUtils
fun getFormattedShortDate(date: Long): String =
    formatterFor(DATE_FORMAT, ZoneId.systemDefault()).format(Instant.ofEpochMilli(date))
```

Then `MembersAdapter` calls that, the pattern constant stays load-bearing, the formatter is shared and cached process-wide rather than per-adapter-instance, and nothing on screen moves.

If instead the long UTC format *is* wanted here, that's a product decision — but it needs to be stated in the PR, checked against the row layout, and justified for the timezone shift.

**Verdict:** needs a change. The refactor is worth doing; this particular substitution changes user-visible output in two ways and ships with a test that is structurally incapable of noticing.

---

dogi:

---

dogi:
@dogi Thanks for the review — all three points addressed in 0c28dc6928f092a4e269fb1cdacb62eabf52775a.

**1. Display format + timezone now preserved.** The substitution to `getFormattedDate(Long?)` was wrong: it uses the long `EEEE, MMM dd, yyyy` pattern in UTC, not the `dd MMM yyyy` / system-default zone the per-adapter formatter used. Added `TimeUtils.getFormattedShortDate(Long)` — a cached wrapper around `formatterFor(DATE_FORMAT, ZoneId.systemDefault())` — and the adapter now calls that. This hoists the per-adapter formatter into the shared, process-wide `ConcurrentHashMap` cache (the actual goal of the issue) while keeping the on-screen output byte-identical to the pre-PR code. `DATE_FORMAT` stays load-bearing.

**2. The tautological test is fixed.** It previously asserted the bound text `contains(TimeUtils.getFormattedDate(timestamp))` — a re-call of the function under test, so it passed for any format. Replaced with a literal expectation (`11 Mar 2024`) pinned to a fixed `TimeZone`/`Locale` (UTC / US), restored in a `finally`. This fails if the pattern or zone regresses — exactly the forcing function you described. The `no_visit` test was already correct and is unchanged.

**3. Verification.** `./gradlew :app:testDefaultDebugUnitTest --tests MembersAdapterTest --tests TimeUtilsTest` — MembersAdapterTest 6/6 pass (incl. both last-visit tests), TimeUtilsTest 25/25 pass, 0 failures.

_This reply was created by an AI agent (OpenHands) on behalf of the repository maintainers._

dogi:
## Review: 86/100 — mergeable

Both problems from my earlier pass are fixed, and fixed the right way rather than by adjusting the test to match.

### The formatting is now genuinely identical ✅

The earlier version reused `defaultDateFormatter`, which is `"EEEE, MMM dd, yyyy"` in **UTC** — a different pattern *and* a different zone from what `MembersAdapter` rendered, so last-visit dates would have changed shape and could land on the wrong day. This version adds a dedicated `getFormattedShortDate` instead:

```kotlin
formatterFor(DATE_FORMAT, ZoneId.systemDefault())      // "dd MMM yyyy", system zone
```

versus what it replaces:

```kotlin
DateTimeFormatter.ofPattern(TimeUtils.DATE_FORMAT).withZone(ZoneId.systemDefault())
```

Same pattern constant, same zone. Behaviour-preserving.

**The formatter cache is correctly keyed**, which matters more than it looks: `FormatterKey(pattern, zone, locale)` includes the zone and the locale, and both are evaluated per call, so a locale or timezone change produces a fresh entry rather than serving a stale formatter. That's also what makes the test's `TimeZone.setDefault` pinning actually work — `ZoneId.systemDefault()` resolves through `TimeZone.getDefault()`, so the pin reaches the formatter. (Worth contrasting with #16475 in this queue, where the same pinning idiom does *not* reach SQLite's `'localtime'`.)

### The test is a real assertion now ✅

The literal `"11 Mar 2024"` replaces the earlier self-referential expectation that called the function under test — which could only ever agree with itself. `DATE_FORMAT` is `"dd MMM yyyy"`, so the literal is right, and the comment saying why it's a literal is worth keeping. Locale and timezone are pinned and restored in a `finally`. The null-`lastVisitDate` companion test covers the other branch.

### Small side benefit

The per-instance `dateFormatter` field is gone, so all `MembersAdapter` instances now share one cached formatter instead of constructing one each.

### Nits

- `Locale.getDefault()` (via `formatterFor`'s default) vs `DateTimeFormatter.ofPattern(String)`'s `Locale.getDefault(Locale.Category.FORMAT)` — these can diverge if an app sets a separate FORMAT-category locale. Android effectively never does, and the explicit form is the more predictable of the two, so this is a footnote rather than a concern.
- `e.printStackTrace()` in the new function matches its neighbours in `TimeUtils`, but runs against the direction #16595 is establishing.
- The second test's `assertTrue(...contains(expected))` would be tighter as `assertEquals`, and it fully-qualifies `android.content.Context` / `org.ole.planet.myplanet.R` inline. The new `java.util.*` imports are also placed after the project imports, against the file's existing ordering.

**Contention:** #16422 modifies the same `MembersAdapter.kt` (the `tvTitle` binding) **and** adds cases to the same `MembersAdapterTest.kt`. Different hunks, but both test files grow at the end — expect to reconcile whichever lands second. #16422 is already at `ready`.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16530, opened 2026-08-27
login: smoother achievements payload handling (fixes #16523)

### LINKED ISSUE #16523: Lift the handled-payload set to companion scope in AchievementsAdapter

`onBindViewHolder(holder, position, payloads)` builds `setOf(PAYLOAD_PHONE, PAYLOAD_EMAIL)` per partial-bind call — an allocation on every diff callback.

**Ask:** lift the `setOf(PAYLOAD_PHONE, PAYLOAD_EMAIL)` into companion-object scope so it is allocated once instead of on every partial-bind call.

**Files:** `ui/user/AchievementsAdapter.kt`

dogi:
## Summary

`AchievementsAdapter.onBindViewHolder(holder, position, payloads)` built `setOf(PAYLOAD_PHONE, PAYLOAD_EMAIL)` on every partial-bind call — an allocation on every diff callback. This lifts that set into the companion object as `HANDLED_PAYLOADS` so it is allocated once.

## Changes

- `ui/user/AchievementsAdapter.kt`
  - Hoisted `setOf(PAYLOAD_PHONE, PAYLOAD_EMAIL)` into the companion object as the private `HANDLED_PAYLOADS` property (allocated once at class-load).
  - `onBindViewHolder(..., payloads)` now subtracts `HANDLED_PAYLOADS` instead of allocating a fresh set per call.

## Tests

Added coverage in `app/src/test/java/org/ole/planet/myplanet/ui/user/AchievementsAdapterTest.kt` for the partial-bind payload path:

- `testOnBindViewHolder_withPhonePayload_updatesOnlyPhone` — phone payload updates phone only.
- `testOnBindViewHolder_withEmailPayload_updatesOnlyEmail` — email payload updates email only.
- `testOnBindViewHolder_withBothPayloads_updatesBothFields` — both payloads update both fields.
- `testOnBindViewHolder_withUnknownPayload_fallsBackToFullBind` — unknown payload falls back to a full bind.
- `testOnBindViewHolder_withNestedPayloadList_flattensAndHandlesKnownPayloads` — nested list of payloads is flattened and handled.

All 8 tests in the suite pass (`./gradlew testDefaultDebugUnitTest --tests "org.ole.planet.myplanet.ui.user.AchievementsAdapterTest"`).

## Notes

- No version bump; this is a perf-only refactor of an existing allocation with no schema/behavior change.
- Behavior is unchanged — the set contents and the subtract/fallback logic are identical; only the allocation site moved.

Closes #16523.

_This pull request was created by an AI agent (OpenHands) on behalf of the repository maintainer._

@dogi can click here to [continue refining the PR](https://app.all-hands.dev/conversations/53c6e8c4-0269-4a1d-8a45-4f1f79490f5f)


dogi:
## Review: achievements payload handling

**Rating: 89/100** — a one-line hoist that's correctly placed, and five tests that are worth considerably more than the hoist.

### Correctness

`setOf(PAYLOAD_PHONE, PAYLOAD_EMAIL)` moves from a local in `onBindViewHolder` to a companion `private val HANDLED_PAYLOADS`. `PAYLOAD_PHONE` and `PAYLOAD_EMAIL` are `const val` in the same companion, so they're compile-time constants inlined before the property initialiser runs — no initialisation-order hazard, which is the one way a companion hoist like this goes wrong.

The set is only read (`payloadSet.subtract(HANDLED_PAYLOADS)`), so sharing one instance across all adapter instances is safe. Behaviour is identical.

The saving is real if small: this is the payload overload of `onBindViewHolder`, which runs per row on every partial rebind, so a two-element `LinkedHashSet` was being allocated and discarded on each one.

### The tests are the substance here

Five cases covering the partial-bind contract, and they assert the property that actually matters — that a targeted payload updates *only* its field:

```kotlin
adapter.onBindViewHolder(holder, 0, mutableListOf<Any>(PAYLOAD_PHONE))
assertEquals("123-456-7890", binding.tvRefPhone.text.toString())
assertNotEquals("john@example.com", binding.tvRefEmail.text.toString())
assertNotEquals("John Doe", binding.tvRefName.text.toString())
```

Pre-seeding the views with `"stale"` and then asserting the untouched fields *stay* stale is the right technique — without it, a full rebind would satisfy the positive assertion and the test would prove nothing.

The unknown-payload → full-bind fallback and the nested-list flattening case are the two branches most likely to be broken by a future edit (`flatPayloads.any { it !is String }` in particular is easy to get wrong), and both are now pinned.

Worth noting these pass on `master` too — the hoist doesn't change behaviour — so they're locking in the existing contract rather than proving a fix. That's the correct role for them, and it's what makes the hoist safe to land.

### Nit

The PR title frames this as payload-handling work, but the production diff is a single allocation hoist; the tests are the contribution. Worth saying so in the body so a reviewer knows where to look — and so the tests get credit rather than reading as incidental.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16531, opened 2026-08-27
teams: smoother tasks assignees view modelling (fixes #16509)

### LINKED ISSUE #16509: Deduplicate unknown task assignees during collection

`TeamsTasksFragment.refreshTaskList` builds a list of assignee IDs, filters it, then calls `distinct()` — intermediate collections on every task refresh where an insertion-ordered set collects them in one pass.

**Ask:** collect the assignee IDs into an insertion-ordered set in one pass instead of the build-list → filter → `distinct()` pipeline, dropping the intermediate collections.

**Files:** `ui/teams/tasks/TeamsTasksFragment.kt`

dogi:
Deduplicate unknown task assignees into an insertion-ordered LinkedHashSet in a single pass in TeamsTasksFragment to eliminate intermediate collection allocations.

Fixes #16509

---
*PR created automatically by Jules for task [2396491362478856543](https://jules.google.com/task/2396491362478856543) started by @dogi*


dogi:
## Review: deduplicate unknown task assignees during collection

**Rating: 88/100** — equivalent, one pass instead of three, and the lambda rename that makes it readable is the detail that had to be right.

### Correctness

Old:

```kotlin
list.mapNotNull { it.assignee }
    .filter { it.isNotBlank() && !knownAssigneeIds.contains(it) }
    .distinct()
```

New:

```kotlin
list.mapNotNullTo(LinkedHashSet()) { task ->
    task.assignee?.takeIf { it.isNotBlank() && it !in knownAssigneeIds }
}
```

Same set, same order. `distinct()` is backed by a `LinkedHashSet` and keeps first-occurrence order, and collecting directly into a `LinkedHashSet` gives the identical sequence — which matters because the result feeds `getUsersByIds` and the caller then maps results back by id.

`it !in knownAssigneeIds` and `!knownAssigneeIds.contains(it)` compile to the same call. The null-assignee case still drops out (`?.` yields null → `mapNotNullTo` skips), and blank/known assignees now drop out via `takeIf` returning null instead of via a separate `filter` pass.

Renaming the outer lambda parameter to `task` is what makes this work — with the default `it`, the inner `takeIf { it.isNotBlank() }` would have shadowed and read ambiguously even though it would still compile to the right thing. Good call.

Allocation-wise: three intermediate lists become one `LinkedHashSet`, over one traversal instead of three.

### Signature widening

`fetchAssigneeNames(assigneesToFetch: List<String>)` → `Collection<String>` with a `.toList()` at the `getUsersByIds` boundary. Correct — it lets the caller pass the set without materialising a list first, and the conversion happens once at the point where a `List` is actually required.

Worth noting `UserRepository.getUsersByIds` chunks its input (at 400) before hitting the DAO, so the `.toList()` isn't feeding an unbounded `IN (…)` clause.

### Gap

No test. The change is behaviour-preserving so nothing regresses, but `assigneesToFetch` is the input to a network/DB fetch whose results are keyed back by id — a case with duplicate assignees across tasks, some already in `knownAssigneeIds`, asserting the exact collection passed to `fetchAssigneeNames`, would take a few lines and would pin both the dedup and the order-sensitivity. Not blocking.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16535, opened 2026-08-27
all: smoother network utils handling (fixes #16528)

### LINKED ISSUE #16528: Fix the mislabeled network-id local in NetworkUtils.getCurrentNetworkId
**rating:** 40/100 &nbsp;·&nbsp; **proposed by:** glm(r3)

`var ssid = -1` is assigned `connectionInfo.networkId` and returned — the name lies about both type and meaning in the connectivity path.

**Ask:** rename the mislabeled `ssid` local to something that reflects what it actually holds (a network id `Int`). Pure rename, zero behavior change.

**Files:** `utils/NetworkUtils.kt`

**Verification:** File exists; not line-verified. Pure rename, zero behavior change.

dogi:
## Summary

Renamed the mislabeled `ssid` local in `NetworkUtils.getCurrentNetworkId()` to `networkId`.

`var ssid = -1` was assigned `connectionInfo.networkId` (an `Int` network id) and returned by the function, but the variable name `ssid` falsely suggested it held the Wi-Fi SSID string. The name lied about both the type (`Int`, not `String`) and the meaning (a network id, not an SSID) in the connectivity path.

Pure rename, zero behavior change.

## Files
- `app/src/main/java/org/ole/planet/myplanet/utils/NetworkUtils.kt`

## Tests
Added unit tests for `getCurrentNetworkId` in `NetworkUtilsMockTest.kt` covering the connectivity path that the renamed local participates in:
- returns `-1` when no active network
- returns `-1` when transport is not wifi
- returns `-1` when wifi transport but connectionInfo is null
- returns `-1` when ssid is null or empty
- returns the network id when wifi connected with a valid ssid

All `NetworkUtilsMockTest` tests pass (14 tests, 0 failures, 0 errors).

Closes #16528.

---

@dogi can click here to [continue refining the PR](https://app.all-hands.dev/conversations/5471066b-030d-458d-af97-be4d4aa0c5a8)


dogi:
## Review: rename the mislabeled `ssid` local to `networkId`

**Rating: 88/100** — the rename fixes a genuinely misleading name, and the five tests are the real contribution.

### The rename is warranted

The local was named `ssid` but only ever held `connectionInfo.networkId` — an `Int` handle to a saved network configuration, which is a completely different thing from the SSID string. The function's return type (`Int`) already contradicted the name. Renaming to `networkId` makes the declaration, the assignment, and the return agree.

Pure local rename, no signature or behaviour change.

### Tests

Five cases covering every exit: no active network, non-WiFi transport, WiFi with null `connectionInfo`, WiFi with a null SSID, and the success path returning `42`. That's complete branch coverage of a function that had none, and it's the kind of function where the fallbacks matter — three separate conditions each silently yield `-1`, and before this PR nothing distinguished them.

They also pin down something worth having on record: the SSID is used purely as a **guard**, not as data. `networkId` is returned only when `connectionInfo.ssid` is non-empty, so a WiFi connection whose SSID is hidden or unavailable yields `-1` even though a valid `networkId` exists. That's pre-existing behaviour and reasonable as a "are we really on a known WiFi network" proxy — but it's the sort of thing that reads as a bug to the next person, and now the test says it's deliberate.

### Note (pre-existing, not this PR)

`WifiInfo.getNetworkId()` and `WifiManager.getConnectionInfo()` are both deprecated — hence the existing `@Suppress("DEPRECATION")`. On API 29+ `getNetworkId()` returns `-1` for callers without location permission, and the app doesn't declare `ACCESS_FINE_LOCATION`. So on modern devices this function may return `-1` on the success path too, regardless of the guard. Worth an issue to check whether callers depend on a real value here; nothing for this PR to do.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16539, opened 2026-08-27
courses: smoother repository member filtering (fixes #16517)

### LINKED ISSUE #16517: Dedupe course membership in one collection pass
**rating:** 42/100 &nbsp;·&nbsp; **proposed by:** codex(r2), deepseek(r2)

`MyCourse.setUserId` filters into a mutable list, does a linear `contains`, conditionally appends, then runs `distinct()` — an extra full pass and list on every course save.

**Ask:** collect the user ids into an insertion-ordered `Set` in one pass. **Do NOT** simply drop the trailing `.distinct()` on the grounds that `contains` guarantees uniqueness — that reasoning is wrong, because the pre-existing list can already contain duplicates.

**Files:** `model/MyCourse.kt`

**Verification:** Verified at MyCourse.kt:53-60. Take codex's insertion-ordered-Set version: deepseek's step 1 ("drop the trailing .distinct() — contains already guarantees uniqueness") is wrong, because the pre-existing list can already contain duplicates.

dogi:
Refactor MyCourse.setUserId to collect non-blank user IDs into an insertion-ordered LinkedHashSet in a single pass, filtering out blank entries and deduplicating pre-existing entries while preserving insertion order.

Fixes #16517

---
*PR created automatically by Jules for task [5496159564222505517](https://jules.google.com/task/5496159564222505517) started by @dogi*


dogi:
## Review: dedupe course membership in one collection pass

**Rating: 89/100** — equivalent, and the equivalence rests on a `LinkedHashSet` property that the tests correctly pin down.

### Correctness

Old:

```kotlin
val current = this.userId.orEmpty().filter { !it.isNullOrBlank() }.toMutableList()
if (!current.contains(userId)) current.add(userId)
this.userId = current.distinct()
```

New:

```kotlin
val set = this.userId.orEmpty().filterTo(LinkedHashSet()) { !it.isNullOrBlank() }
set.add(userId)
this.userId = set.toList()
```

The subtle part is what happens when `userId` is **already present**. The old code guarded with `contains` and skipped the append; the new code calls `add` unconditionally and relies on `LinkedHashSet.add` returning `false` **without re-positioning** the existing entry. If `LinkedHashSet` had access-order semantics (as `LinkedHashMap(accessOrder = true)` does) this would silently move the id to the end and change the persisted order. It doesn't — insertion order is fixed at first insert — so the two are identical.

`testSetUserId_existingUser_preservesInsertionOrderWithoutDuplicates` is exactly the test that property needs, and `testSetUserId_deduplicatesPreExistingAndPreservesInsertionOrder` covers the other half: pre-existing duplicates and blanks in the stored list. Worked through by hand, `["user1", "", "user2", "user1", "   "] + "user3"` gives `["user1", "user2", "user3"]` under both versions.

Allocation-wise: three intermediate lists (filter → toMutableList → distinct) plus a linear `contains` scan become one `LinkedHashSet` and one `toList`, with O(1) membership. `setUserId` runs per course per user during shelf sync, so it's a real if modest saving on a path that repeats.

### Tests

Five on a model method that had none, covering the null/blank/whitespace early return, the null-list case, both dedup orderings, and `removeUserId`. They pass on `master` too, so they're characterisation tests locking behaviour across the refactor — which is the right role, and it's what makes the `LinkedHashSet` swap safe to land.

### Nit

`it.isNullOrBlank()` inside the filter implies the element type is nullable, but every caller and test supplies `List<String>`. If `MyCourse.userId` is declared `List<String>?` rather than `List<String?>?`, `isNullOrBlank()` is doing a null check that can't fire and `isBlank()` would say what's meant. Worth a look while this code is open — if the element type genuinely is nullable, the `filterTo` target is a `LinkedHashSet<String?>` and it'd be clearer to state that explicitly.

**Verdict:** mergeable.

---

Okuro3499 (dismissed):
```
(MyCourse.setUserId) that has no production caller. Real course-membership add/remove flows go through       
  CoursesRepositoryImpl.mergeUserIds() (repository/CoursesRepositoryImpl.kt:219,341) and a direct .filter { it != userId } (line 369), not through                              
  MyCourse.setUserId/removeUserId. So this branch adds test coverage and a micro-optimization for dead code, while the actual dedup logic on the hot path (mergeUserIds) is     
  untouched and unreviewed here.                                           
```

dogi:
## Review — rating 84/100

Clean refactor of two copies of the same merge-user-id logic, and the equivalence holds.

Master built a list, did an O(n) `contains` check, appended, then called `distinct()` — which internally allocates a `LinkedHashSet` anyway. So the old code was doing the set work twice while looking like it was avoiding it. Collapsing to a single `filterTo(LinkedHashSet())` gives the same result because `LinkedHashSet` preserves insertion order exactly as `distinct()` does (first occurrence wins), and `add` is idempotent so the `contains` guard was redundant.

Worth noting the type cleanup that came along with it: master's `filter { !it.isNullOrBlank() }` was a null check on a `List<String>`, where the element can never be null. `isNotBlank()` is the honest predicate.

The user-id lists here are small — the users on one device who added a course — so this is about clarity rather than performance. Fine; the old version read as if it were guarding against something.

### The tests are the right kind

Both new test sets pass on master as well, which is exactly what you want from a pure refactor: they are characterization tests pinning behaviour that had **no coverage at all** before (`MyCourseTest` is a new file, and `setUserId`/`removeUserId` were previously untested). The fixture `listOf("user1", "", "user2", "user1")` is well chosen — it exercises blank filtering, pre-existing duplicates, and order preservation in one case, so a future "simplification" that reaches for `toSet()` instead of `LinkedHashSet` and loses ordering will fail loudly.

Capturing the upserted entity with `slot<MyCourse>()` and asserting the merged list, rather than just verifying the call happened, is the detail that makes the repository tests worth having.

Moving to `ready`.

---

Okuro3499 (approved):

PR #16543, opened 2026-08-27
dashboard: smoother surveys text coloring (fixes #16522)

### LINKED ISSUE #16522: Hoist the dashboard survey list text color

`DashboardSurveysAdapter.onBindViewHolder` calls `ContextCompat.getColor(holder.itemView.context, R.color.daynight_textColor)` on every row bind of the survey-picker dialog.

**Ask:** hoist the `ContextCompat.getColor(...)` lookup out of `onBindViewHolder` so the color is resolved once rather than on every row bind.

**Files:** `ui/dashboard/DashboardSurveysAdapter.kt`

dogi:
Hoisted `ContextCompat.getColor(..., R.color.daynight_textColor)` from `onBindViewHolder` into `SurveyViewHolder` in `DashboardSurveysAdapter` to avoid redundant color lookups on every row bind. Added Robolectric unit tests in `DashboardSurveysAdapterTest`.

Fixes #16522

---
*PR created automatically by Jules for task [8459317755225466695](https://jules.google.com/task/8459317755225466695) started by @dogi*


dogi:
## Review: hoist the dashboard survey list text color

**Rating: 80/100** — safe in this adapter's specific circumstances, and the tests are solid, but they pull in Mockito, which this project neither declares nor uses by convention.

### Correctness

Moving `ContextCompat.getColor(itemView.context, R.color.daynight_textColor)` from `onBindViewHolder` to a `val` on the ViewHolder resolves it once per holder instead of once per bind.

The thing to check is that `daynight_textColor` is a **theme-dependent** resource — its resolved value differs between light and dark mode. Caching a resolved color on a ViewHolder is only safe if holders can't outlive a UI-mode change. Here they can't:

- A night-mode switch (via `AppCompatDelegate.setDefaultNightMode` or a system change) recreates the Activity by default, which rebuilds the RecyclerView and all its holders.
- This adapter is scoped to an `AlertDialog` (it takes one in its constructor and dismisses it on click), so it's short-lived and goes away with the Activity anyway.

So no staleness in practice. Worth a one-line comment on the field saying the value is theme-dependent and relies on holder recreation — because the general pattern *does* break if a `RecycledViewPool` is ever shared across configuration changes, and this is the kind of line that gets copied to an adapter where that's true.

Value-wise: `ContextCompat.getColor` is a resource lookup with `TypedValue` resolution — cheap but not free, and it was running on every bind. For a small dialog list the saving is negligible; the readability improvement in `onBindViewHolder` (five lines to one) is the more honest justification.

### The Mockito problem

```kotlin
import org.mockito.Mockito.mock
import org.mockito.Mockito.verify
```

Two issues, and the second is the real one:

1. **Convention.** `CLAUDE.md` states the testing convention as "MockK (not Mockito)". The suite is overwhelmingly MockK; this adds a fourth file (after `SubmissionViewModelTest`, `CoursesAdapterTest`, `ResourcesAdapterTest`) going the other way.
2. **Mockito is not a declared dependency.** It appears nowhere in `app/build.gradle` or `gradle/libs.versions.toml` — it's on the test classpath only transitively (`androidx.arch.core:core-testing` pulls `mockito-core`). That works today, which is why CI is green, but it means a Dependabot bump to `core-testing` that changes or drops that transitive edge breaks these tests with a `NoClassDefFoundError` and no obvious cause. Given the repo runs daily Dependabot on Gradle deps, that's a live risk.

`mockDialog` is used for exactly two things — being passed to the constructor and having `dismiss()` verified. Both are one-liners in MockK:

```kotlin
private val mockDialog: AlertDialog = mockk(relaxed = true)
…
verify { mockDialog.dismiss() }
```

That removes the undeclared dependency and matches the rest of the suite. If Mockito is genuinely wanted, it should be declared explicitly in the version catalog rather than inherited.

### Tests otherwise

Good: the first asserts both the hoisted field **and** that it's actually applied (`viewHolder.textView.currentTextColor`), so a hoist that forgot to use the field would fail. The second drives a real `RecyclerView` through measure/layout to get a bound holder rather than faking one, which is the right way to test the click path.

**Verdict:** mergeable. I'd swap the two Mockito calls for MockK before merging — it's a two-line change that removes a real fragility.

---

Okuro3499 (approved):

PR #16544, opened 2026-08-27
login: smoother configurations repository server availability checking (fixes #16538)

### LINKED ISSUE #16538: Parse the server-availability response without building a token list
**rating:** 38/100 &nbsp;·&nbsp; **proposed by:** codex(r1), deepseek(r2)

`checkServerAvailability` reads the whole body, splits on ",", drops trailing empties and allocates a list purely to test whether it has eight entries.

**Ask:** determine whether the response has eight entries without building the intermediate token list (count the separators / iterate instead of `split` + `dropLastWhile` + list allocation).

**Files:** `repository/ConfigurationsRepositoryImpl.kt`

**Verification:** File and function exist; the split/dropLastWhile chain was not line-verified. Shares a file with the config-logging task.

dogi:
## Summary

`checkServerAvailability` read the whole response body, split it on `","`, dropped trailing empties with `dropLastWhile { it.isEmpty() }`, and allocated a `List<String>` purely to test whether it had eight entries.

This PR replaces that chain with `countCommaEntries`, which determines whether the response has eight entries **without building the intermediate token list**: it strips trailing commas, then counts separators in a single pass and returns `commaCount + 1` (or `0` for null/empty/commas-only bodies).

## Behavior preservation

The new helper is equivalent to the original `split(",").dropLastWhile { it.isEmpty() }.size`:
- null / empty / all-commas bodies → `0` (server unavailable)
- trailing commas are dropped (e.g. `"1,2,3,4,5,6,7,8,,"` → 8 entries)
- internal empty tokens are kept (e.g. `"a,,,b"` → 4 entries)

## Changes

- `repository/ConfigurationsRepositoryImpl.kt` — replaced the `split` + `dropLastWhile` + list allocation with a call to the new `countCommaEntries` helper.
- `app/src/test/.../ConfigurationsRepositoryImplTest.kt` — added edge-case tests covering trailing commas, all-commas, empty body, internal empty tokens, and the exact-seven boundary.

## Verification

`./gradlew testDefaultDebugUnitTest --tests "org.ole.planet.myplanet.repository.ConfigurationsRepositoryImplTest"` — **29 tests, 0 failures, 0 errors** (includes the 6 new edge-case tests).

Fixes #16538

---

@dogi can click here to [continue refining the PR](https://app.all-hands.dev/conversations/a96fee77-762c-439a-bcf5-03221a12d951)


dogi:
## Review: parse the server-availability response without building a token list

**Rating: 87/100** — I worked the equivalence through case by case and it holds exactly, including the awkward ones.

### Equivalence

Old: `ss?.split(",")?.dropLastWhile { it.isEmpty() }?.size ?: 0`
New: `countCommaEntries(ss)` — strip trailing `,`, count remaining `,`, add 1 (or 0 if nothing remains).

| body | old | new |
|---|---|---|
| `null` | `0` | `0` |
| `""` | `split` → `[""]` → drop → `[]` → `0` | `end == 0` → `0` |
| `"a"` | `1` | `0 commas + 1` → `1` |
| `"a,"` | `["a",""]` → drop → `1` | strip → `"a"` → `1` |
| `",,,"` | `["","","",""]` → drop all → `0` | strip → `end == 0` → `0` |
| `"1,…,8,,"` | 10 → drop 2 → `8` | strip 2 → 7 commas + 1 → `8` |
| `"a,,,b,,,,,,,h"` | 11 (no trailing empty) | 10 commas + 1 → `11` |

The one that would break a naive rewrite is the last: **internal** empty entries must still count, only **trailing** ones are dropped. `dropLastWhile` says that, and the strip-then-count formulation preserves it because stripping only touches the suffix. Good.

The `>= 8` boundary is unchanged, and `"1,2,3,4,5,6,7"` → 7 → false is now pinned.

### Tests

Five cases, and they're the right five — the boundary (7 vs 8), trailing commas, all-commas, empty body, and internal empties. They pass on `master` too, so they're characterisation tests locking the contract across the rewrite, which is exactly what a hand-rolled replacement for a stdlib chain needs.

### Trade-off worth naming

This swaps a two-line idiomatic expression for an eleven-line hand-rolled scanner. The saving is the intermediate `List<String>` and its substrings — real, but `checkServerAvailability` runs on login/sync entry, not in a loop, and the body is a short server response. So the performance case is thin.

What does justify it is legibility of intent: `split(",").dropLastWhile { it.isEmpty() }.size` requires the reader to reconstruct "count entries, ignoring trailing empties", whereas `countCommaEntries` names it. I'd keep the change on that basis and drop the perf framing from the PR body — and add a one-line KDoc on the helper stating the trailing-vs-internal rule, since that's the invariant a future edit is most likely to break.

### Pre-existing

`dbCount >= 8` is an unexplained magic number — presumably the count of CouchDB databases a healthy Planet server exposes. Worth a named constant with a comment while this code is open; the tests now depend on it and nothing says where 8 comes from.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16549, opened 2026-08-27
all: smoother server config utils networking (fixes #16537)

### LINKED ISSUE #16537: Parse the local-network host once in ServerConfigUtils
**rating:** 38/100 &nbsp;·&nbsp; **proposed by:** codex(r3)

`isLocalNetwork` derives its host with two chained `split` calls and their temporary lists before the address checks, on the server-selection path.

**Ask:** derive the host once without the two chained `split` calls and their temporary lists, before the address checks.

**Files:** `utils/ServerConfigUtils.kt`

**Verification:** Verified verbatim at ServerConfigUtils.kt:72-80.

dogi:
## Summary

`ServerConfigUtils.isLocalNetwork` derived its host with two chained `split(":")` / `split("/")` calls on the server-selection path, each allocating a temporary `List` before the address checks.

Per issue #16537, this derives the host once without the two chained `split` calls and their temporary lists, using `substringBefore` chained calls instead. `substringBefore` returns the whole string when its delimiter is absent — matching the previous `firstOrNull()` fallthrough behavior (`?: url`) — so the result is identical while avoiding the intermediate list allocations.

## Files changed

- `utils/ServerConfigUtils.kt` — `isLocalNetwork` host derivation.
- `utils/ServerConfigUtilsTest.kt` — added `getDefaultProtocol_returnsHttpForLocalNetworkHosts` and `getDefaultProtocol_returnsHttpsForNonLocalHosts` covering the local-network detection (via the public `getDefaultProtocol`, since `isLocalNetwork` is private), including hosts with ports and path segments.

## Verification

`./gradlew testDefaultDebugUnitTest --tests "org.ole.planet.myplanet.utils.ServerConfigUtilsTest"` — 11 tests, 0 failures.

Fixes #16537.

@dogi can click here to [continue refining the PR](https://app.all-hands.dev/conversations/af519294-3334-4b8c-8253-c8c15c712f28)


dogi:
## Review: derive the local-network host without chained splits

**Rating: 90/100** — exactly equivalent, drops two list allocations and one piece of dead code, and the new tests pin the interesting boundaries.

### Equivalence

```kotlin
- val host = url.split(":").firstOrNull()?.split("/")?.firstOrNull() ?: url
+ val host = url.substringBefore(':').substringBefore('/')
```

`split` always returns at least one element, so both `firstOrNull()` calls could never return null — which makes the `?: url` fallback unreachable. `split(":")[0]` and `substringBefore(':')` agree in every case, including when the delimiter is absent (both return the whole string), and the same holds for the `/` step.

Worked through the orderings that could diverge:

- `"192.168.1.73:5984"` → `"192.168.1.73"` both ways
- `"192.168.1.73/db/path"` → `"192.168.1.73"` both ways
- `"192.168.1.73/db:8080"` (delimiters in the other order) → old: `split(":")` → `"192.168.1.73/db"` → `split("/")` → `"192.168.1.73"`; new: same two steps, same result

Two intermediate `List<String>` allocations become zero — `substringBefore` returns a substring view of the original.

### Tests

Good boundary selection. `"172.31.255.255"` (last address of the private `172.16.0.0/12` block → local) versus `"172.32.0.1"` (first address outside it → remote) is exactly the pair that catches an off-by-one in `localNetworkRegex`, and `"192.168.1.73/db/path"` covers the path-stripping branch this diff touches. `"raspberrypi.local"` and `"localhost"` cover the non-numeric arms.

They pass on `master` too, so they're characterisation tests locking behaviour across the rewrite rather than proving a fix — which is the right role for a pure-equivalence refactor.

### Pre-existing, worth knowing

Neither version handles a URL that still carries its scheme: `"http://192.168.1.73"` → `substringBefore(':')` → `"http"` → no local prefix matches → `https://`. Unchanged by this PR (the old chain did the same), and every caller appears to pass a bare host, so nothing is broken today. But it's a latent trap given the function is named for URLs and takes a parameter called `url`. A one-line `removePrefix("http://").removePrefix("https://")` at the top, or renaming the parameter to `host`, would make the contract honest.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16551, opened 2026-08-27
all: smoother notifications repository batch filtering (fixes #16548)

### LINKED ISSUE #16548: Eliminate the intermediate filter lists in the NotificationsRepositoryImpl batch methods
**rating:** 36/100 &nbsp;·&nbsp; **proposed by:** devin(r2)

`getJoinRequestDetailsBatch`, `getTaskTeamNamesByTaskTitles` and `getTaskTeamNamesByTaskIds` build `map{}.filter{}.distinct()` chains and an un-presized Triple list, on every notification badge or list refresh.

**Ask:** collapse the `map{}.filter{}.distinct()` chains in these three methods so the intermediate lists are not allocated, and pre-size the Triple list.

**Files:** `repository/NotificationsRepositoryImpl.kt`

**Verification:** The three methods exist; the chains were not line-verified.

dogi:
## Summary

Collapses the `map{}.filter{}.distinct()` chains in the three NotificationsRepositoryImpl batch methods so the intermediate lists are no longer allocated on every notification badge/list refresh, and pre-sizes the `Triple` list.

## Changes

`repository/NotificationsRepositoryImpl.kt`:

- `getTaskTeamNamesByTaskIds` — replaced `tasks.mapNotNull { it.teamId }.filter { it.isNotEmpty() }.distinct()` with a single accumulation pass into a `LinkedHashSet`, preserving insertion-ordered deduplication without allocating two intermediate lists.
- `getTaskTeamNamesByTaskTitles` — same `LinkedHashSet` accumulation applied to the `teamId` collection.
- `getJoinRequestDetailsBatch` — collapsed both the `joinRequests.map { it.teamId }.filter { ... }.distinct()` and the `intermediateList.map { it.second }.filter { ... }.distinct()` chains into `LinkedHashSet` accumulation passes, and pre-sized the `Triple<String, String, String>` list to `joinRequests.size` via `ArrayList(...)` instead of the un-sized `mutableListOf`.

Behavior is unchanged: deduplication is still insertion-ordered (matching `distinct()`'s order), empty/blank ids are still skipped, and the empty-list short-circuits are untouched.

## Tests

Added unit tests in `NotificationsRepositoryImplTest.kt` covering the three refactored methods:

- mapping results and skipping empty/blank `teamId`s
- deduplication of team ids and user ids passed downstream (verified via `coVerify` on the deduplicated `List`)
- the `Unknown User` / `Unknown Team` fallbacks in `getJoinRequestDetailsBatch`
- empty-input short-circuits

The environment lacks a JDK/Android SDK so the suite could not be run locally; the new tests mirror the existing MockK + `runTest` patterns in the file.

fixes #16548

@dogi can click here to [continue refining the PR](https://app.all-hands.dev/conversations/4595a4f8-102b-481e-8a3f-7dded10f0c45)


dogi:
## Review: batch id dedup in the notifications repository

**Rating: 87/100** — three equivalent conversions plus nine tests on methods that had none. The idiom chosen is more verbose than the one a sibling PR uses for the same job.

### Correctness

All three sites replace `mapNotNull { … }.filter { it.isNotEmpty() }.distinct()` with a `LinkedHashSet` filled in one pass. `distinct()` is `LinkedHashSet`-backed and keeps first-occurrence order, so the resulting sequences are identical — which matters because each feeds a `getTeamNamesByIds` / `getUsersByIds` lookup whose results are then keyed back by id.

Per site, three intermediate lists collapse to one set. `mutableListOf<Triple<…>>()` → `ArrayList(joinRequests.size)` is a capacity hint only, no semantic change, and the size is exact since every join request contributes at most one entry.

### Idiom: prefer the `mapNotNullTo` form

```kotlin
tasks.forEach { task -> task.teamId?.takeIf { it.isNotEmpty() }?.let { teamIds.add(it) } }
```

This is a two-statement declaration plus a triple-chained `?.takeIf?.let`, repeated three times. #16531, doing exactly the same job in `TeamsTasksFragment`, uses:

```kotlin
val teamIds = tasks.mapNotNullTo(LinkedHashSet()) { task ->
    task.teamId?.takeIf { it.isNotEmpty() }
}
```

Same single pass, same set, but it's one expression, the `val` is initialised where it's declared, and the `?.let { add(it) }` tail disappears. Since both PRs are landing into the same codebase in the same window, matching the idiom is worth doing — otherwise the next person sees two shapes for one pattern and doesn't know which is intended.

### Tests

Nine, on three methods with no prior coverage, and the assertions are the discriminating kind:

```kotlin
coVerify { teamsRepository.get().getTeamNamesByIds(listOf("teamA")) }
```

asserts the exact deduplicated list, so a `HashSet` (order-losing) or a missing dedup would fail. The empty-input tests assert the DAO is *not* called at all (`coVerify(exactly = 0)`), which pins the early returns. And `getJoinRequestDetailsBatch falls back to Unknown for missing user and team` covers the `"Unknown User"` / `"Unknown Team"` defaults that are easy to break silently.

They pass on `master` too — expected for an equivalence refactor — so they're locking the contract rather than proving a fix.

### Nits

- `import java.util.LinkedHashSet` is unnecessary: `kotlin.collections.LinkedHashSet` is a typealias available without an import, and it's what the other two PRs in this batch rely on.
- The tests write `org.ole.planet.myplanet.model.TeamTask()` and `org.ole.planet.myplanet.model.UserEntity(...)` fully qualified inline, eight times between them, while `JoinRequestInfo` is imported. Importing both would cut a lot of noise from the new block.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16555, opened 2026-08-27
all: smoother android decrypter utils payload checking (fixes #16536)

### LINKED ISSUE #16536: Compare decrypted payload prefixes without slicing
**rating:** 39/100 &nbsp;·&nbsp; **proposed by:** codex(r3)

`AndroidDecrypter.decrypt` creates a prefix slice solely to compare the payload against the IV, then a second slice to strip it; an offset comparison removes the comparison copy while preserving legacy-ciphertext support.

**Ask:** replace the prefix-slice-then-compare with an offset comparison so the comparison copy is not allocated. Preserve legacy-ciphertext support — that path must keep working.

**Files:** `utils/AndroidDecrypter.kt`

dogi:
Optimized `AndroidDecrypter.decrypt` to replace prefix slicing with offset-based comparison (`startsWith`) and offset decryption (`cipher.doFinal(encryptedBytes, offset, length)`), avoiding intermediate byte array allocations while preserving legacy ciphertext decoding.

Fixes #16536

---
*PR created automatically by Jules for task [16189289072933537380](https://jules.google.com/task/16189289072933537380) started by @dogi*

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved encrypted data handling when payloads include the initialization vector.
  - Preserved compatibility with legacy ciphertext-only data.
  - Prevented decryption issues caused by redundant initialization vector bytes.


dogi:
## Review: compare decrypted payload prefixes without slicing

**Rating: 89/100** — removes two array copies from the decrypt path, and the existing test suite already covers both branches it touches.

### Correctness

Two allocations disappear:

1. `encryptedBytes.sliceArray(0 until ivBytes.size)` — a copy of the prefix, made solely to feed `contentEquals`. Replaced by `startsWith`, which compares in place. The new helper's `array.size < prefix.size` guard reproduces the old `encryptedBytes.size >= ivBytes.size` short-circuit exactly, so a payload shorter than the IV still falls to the legacy branch rather than throwing.
2. `encryptedBytes.sliceArray(ivBytes.size until encryptedBytes.size)` — a copy of the entire ciphertext. Replaced by `cipher.doFinal(encryptedBytes, ivBytes.size, encryptedBytes.size - ivBytes.size)`, which is the standard `javax.crypto.Cipher` offset overload and produces the same output as decrypting the copied range.

The `else` branch is byte-for-byte what it was. So the only observable change is fewer allocations — and the second one scaled with payload size, which is the more valuable of the two.

### Already covered

`AndroidDecrypterTest` has round-trip tests for both formats — the IV-prefixed path and the legacy no-prefix path (`:27` and `:41`) — plus zero-key and truncated-key cases. Those exercise exactly the two branches this diff rewrites, and they're green, which is real evidence that the offset-based `doFinal` is equivalent rather than merely plausible. Good position to be refactoring from.

### On the comparison not being constant-time

`startsWith` returns early on the first mismatching byte, so it leaks timing information about how many leading bytes matched. That's fine here and worth stating so nobody "fixes" it later: the value being compared is the **IV**, which is not secret — it's transmitted in the clear as the payload prefix by design. The old `contentEquals` was equally variable-time, so nothing changed.

The rule to keep in mind is that if this pattern is ever reused to compare an authentication tag or MAC, it must become `MessageDigest.isEqual`. A one-line comment on the helper saying it's for non-secret prefixes only would be cheap insurance, given `startsWith` is a tempting name to reuse.

### Nit

`ex.printStackTrace()` in the surrounding `catch` survives — same conversion #16471/#16388/#16595 are working through elsewhere, and a decrypt failure is exactly the kind of thing you'd want tagged in logcat. Not this PR's scope.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16554, opened 2026-08-27
all: smoother emptiness checking (fixes #16553)

dogi:
## Summary

Completes the `isEmpty()`/`isNotEmpty()` sweep of every `size()`/`size`-vs-`0` emptiness check in `app/src/main`, as requested in #16553 and extended per @dogi's review (the prior commits converted 6 of 23 sites — leaving two idioms with no rule separating them). This change converts the remaining 17, so the codebase uses one form throughout.

All 23 sites are now converted. Each conversion is behavior-preserving.

### The original 6 sites (#16553)

| File | Site | Before → After |
|------|------|----------------|
| `utils/JsonUtils.kt` — `addJson` | `value.keySet().size > 0` | `value.keySet().isNotEmpty()` |
| `repository/TagsRepositoryImpl.kt` — `createTag` | `attachedTo.size > 0` | `attachedTo.isNotEmpty()` |
| `repository/ConfigurationsRepositoryImpl.kt` | `rows.size() > 0` | `rows != null && !rows.isEmpty()` |
| `services/sync/TransactionSyncManager.kt` | `arr.size() == 0` | `arr.isEmpty()` |
| `model/Feedback.kt` — `messageList` | `ar.size() > 0` (guard) | guard removed; loop runs directly |
| `model/Feedback.kt` — `message` | `ar.size() > 0` | `!ar.isEmpty()` |

### The 17 sites added to complete the sweep (commit da18c6f)

All are `JsonArray` emptiness comparisons (`size() > 0` → `!isEmpty()`, `size() == 0` → `isEmpty()`):

| File | Before → After |
|------|----------------|
| `repository/CoursesRepositoryImpl.kt:813` | `correctChoiceArray.size() > 0` → `!correctChoiceArray.isEmpty()` |
| `repository/VoicesRepositoryImpl.kt:203` | `array.size() > 0` → `!array.isEmpty()` |
| `repository/VoicesRepositoryImpl.kt:262` | `filtered.size() == 0` → `filtered.isEmpty()` |
| `repository/VoicesRepositoryImpl.kt:490` | `ar.size() > 0` → `!ar.isEmpty()` |
| `repository/UserRepositoryImpl.kt:1222` | `element.asJsonArray.size() > 0` → `!element.asJsonArray.isEmpty()` |
| `repository/SyncRepositoryImpl.kt:115` | `array.size() > 0` → `!array.isEmpty()` |
[...]

PR #16558, opened 2026-08-27
teams: smoother voices view modelling (fixes #16557)

### LINKED ISSUE #16557: Cache the label map and use the null-safe empty idiom in VoicesViewModel
**rating:** 33/100 &nbsp;·&nbsp; **proposed by:** minimax(r2)

`filterNews` rebuilds a `labelDisplayToValue` map from the `Constants.LABELS` constant on every call, and `downloadReferencedResources` uses `(news?.imagesArray?.size() ?: 0) > 0`.

**Ask:** hoist the `labelDisplayToValue` map so it is built once rather than on every `filterNews` call, and replace the `(… ?.size() ?: 0) > 0` check with the null-safe empty idiom. **Note:** the original task's own steps hedge between two different lookup strategies, which suggests the site was read loosely — read the actual code and pick one coherent approach rather than following those steps literally.

**Files:** `ui/voices/VoicesViewModel.kt`

**Verification:** File exists; not line-verified. The task's own steps 3-5 hedge between two different lookup strategies, which is a sign the site was read loosely.

dogi:
## Summary

`VoicesViewModel.filterNews` rebuilt a `labelDisplayToValue` map from the `Constants.LABELS` constant on every call, and `downloadReferencedResources` used the `(… ?.size() ?: 0) > 0` pattern to check for images.

This PR:
- **Hoists the label map.** The static name→value mapping from `Constants.LABELS` is now built once as a hoisted `labelDisplayToValue` val instead of being re-copied into a local map on every `filterNews` call. The per-news *dynamic* augmentation (display names for label values not present in the static set) still has to be computed per call, since it depends on the current list — so `filterNews` now builds only that dynamic part and resolves `selectedLabel` against the hoisted static map first, falling back to the dynamic map. This is one coherent lookup strategy rather than re-seeding the whole map each time.
- **Uses the null-safe empty idiom.** `(news?.imagesArray?.size() ?: 0) > 0` is replaced with `!news?.imagesArray?.isEmpty() == true`, the project's preferred null-safe-empty check. `News.imagesArray` is already a non-null `JsonArray` (it returns an empty `JsonArray()` when `images` is null), so this reads cleanly and keeps the null-safety on `news` itself.

## Files
- `ui/voices/VoicesViewModel.kt`

## Tests
Adds unit tests in `VoicesViewModelTest.kt`:
- static label filter resolves `Offer` / `Help wanted` through the hoisted map
- dynamic label filter resolves an unknown label value via `VoicesLabelManager.formatLabelValue`
- `downloadReferencedResources` skips news with empty/null images and only downloads referenced resource ids
- `downloadReferencedResources` is a no-op for a list of only `null` news

Closes #16557.

_This PR was created by an AI agent (OpenHands) on behalf of openhands._

@dogi can click here to [continue refining the PR](https://app.all-hands.dev/conversations/0e8c6d83-65bb-4d11-9e06-275f2625ac16)


dogi:
## Review: label map caching in `VoicesViewModel`

**Rating: 85/100** — the resolution semantics survive the split into static + dynamic maps, and the second hunk uses the right `JsonArray` idiom.

### The static/dynamic split preserves precedence

The subtlety is which map wins when a dynamic label's *display name* collides with a static one.

- **Before:** one merged map, seeded from `Constants.LABELS`, then `putIfAbsent(labelName, label)` for dynamic labels — so an existing static entry was never overwritten.
- **After:** two maps, and `labelDisplayToValue[selectedLabel] ?: dynamicLabelDisplayToValue[selectedLabel]` — static consulted first.

Same precedence, reached differently. And `labelDisplayToValue = Constants.LABELS` is a correct alias rather than a copy: the old loop was `Constants.LABELS.forEach { (labelName, labelValue) -> map[labelName] = labelValue }`, i.e. an identity copy of a `displayName → value` map. So the hoist is sound.

The win is that `filteredNews` is a `combine(...)`-derived `StateFlow` — it recomputes on every search-query change, every label change, and every news-list emission. Rebuilding a copy of `Constants.LABELS` on each of those (i.e. per keystroke) now doesn't happen at all.

### `imagesArray` hunk

`(news?.imagesArray?.size() ?: 0) > 0` → `news?.imagesArray?.isEmpty() == false` is equivalent across all three cases: null → `null == false` → false; empty → `true == false` → false; non-empty → `false == false` → true.

Worth noting this uses `JsonArray.isEmpty()` **directly** — the right form. #16554 is landing the same kind of cleanup elsewhere using `asList().isNotEmpty()`, which allocates a view; this file gets it right and is the better precedent of the two.

### Two things I'd change

- **`containsValue` is a linear scan.** `labelDisplayToValue.containsValue(label)` walks the map's values, and it runs once per label per news item. `Constants.LABELS` is small so this is fine in practice, but the intent — "is this a known label value?" — is exactly what `Constants.LABEL_VALUE_TO_NAME` answers in O(1), and that map is already used on the next line. `if (!Constants.LABEL_VALUE_TO_NAME.containsKey(label))` says the same thing, is constant-time, and removes the value-scan entirely.
- **`imagesArray` is read twice** on the same news item (`isEmpty()` then `get(0)`). If it's a computed property that parses the `images` JSON string on each access — the naming suggests it is — that's two parses per item on the download path. Pre-existing (the old code did the same), but this hunk is already touching the line, so hoisting it into a local would be free:
  ```kotlin
  val images = news?.imagesArray
  if (images?.isEmpty() == false) { val ob = images[0].asJsonObject; … }
  ```

### Tests

Four, and the dynamic-label one is the valuable one: it feeds a label value that isn't in `Constants.LABELS`, derives the expected display name through `VoicesLabelManager.formatLabelValue`, and asserts the filter resolves it — which is precisely the path the static/dynamic split could have broken. The `downloadReferencedResources` pair covers the empty-images skip and the null-news case with a `coVerify(exactly = 0)`, so both branches of the changed condition are exercised.

Nit: `org.ole.planet.myplanet.services.VoicesLabelManager` is written fully qualified inline in one test while everything else in the file is imported.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16559, opened 2026-08-27
teams: smoother requests view modelling (fixes #16541)

### LINKED ISSUE #16541: Drop the nested coroutine for post-approval team activity
**rating:** 37/100 &nbsp;·&nbsp; **proposed by:** deepseek(r2)

`RequestsViewModel.respondToRequest` launches a child `launch { teamsRepository.recordTeamActivity() }` inside an already-running coroutine after success — an unsupervised fire-and-forget where a direct suspend call would do.

**Ask:** replace the nested `launch { teamsRepository.recordTeamActivity() }` with a direct suspend call, since it already runs inside a coroutine — dropping the unsupervised fire-and-forget.

**Files:** `ui/teams/members/RequestsViewModel.kt`

**Verification:** Verified verbatim at RequestsViewModel.kt:57-59.

dogi:
Removed the nested coroutine fire-and-forget launch block around teamsRepository.recordTeamActivity() in RequestsViewModel.respondToRequest, replacing it with a direct suspend call inside the existing coroutine scope.

Fixes #16541

---
*PR created automatically by Jules for task [13876799199991336987](https://jules.google.com/task/13876799199991336987) started by @dogi*


dogi:
## Review: drop the nested coroutine for post-approval team activity

**Rating: 88/100** — the inner `launch` bought nothing, and removing it is behaviour-neutral. Worth recording *why* it's neutral, because the call it wraps does network I/O.

### What the nested `launch` did and didn't do

```kotlin
viewModelScope.launch {
    …
    if (result.isSuccess) {
        _successAction.emit(Unit)
        launch { teamsRepository.recordTeamActivity() }   // ← removed
    }
}
```

The inner `launch` provided none of the three things a nested launch usually buys:

- **No parallelism** — it was the last statement in the block, so nothing ran concurrently with it.
- **No separate lifetime** — as a child of the outer coroutine it was cancelled by exactly the same events, at exactly the same moments.
- **No error isolation** — `viewModelScope` is `SupervisorJob`-backed at the top, but that supervision applies to *its* direct children, not to a grandchild launched inside one. An exception escaping `recordTeamActivity` was unhandled either way.

So it was ceremony. Removing it means the outer coroutine now stays alive for the duration of the call instead of spawning a sibling that stays alive for the same duration. Net: identical.

### Worth knowing what's behind that call

`recordTeamActivity()` delegates to `syncTeamActivities()` (`TeamsRepositoryImpl:900-904`), which does real network work — `MainApplication.isServerReachable(...)` probes for the primary URL and possibly the alternative, plus preference writes.

Two things follow, both fine:

- `_successAction.emit(Unit)` fires **before** it, so the UI already has its success signal; the awaited network round-trip doesn't delay anything the user sees.
- The probe runs off the main thread despite `viewModelScope.launch` defaulting to `Dispatchers.Main.immediate`, because `ServerReachabilityProvider.tryConnect` wraps its work in `withContext(dispatcherProvider.io)`. That was true before this change too.

### If detachment was the actual intent

If the goal ever *was* "record this even if the user navigates away", the nested `launch` never achieved it — a child of `viewModelScope` dies with the ViewModel just as the parent does. That would need `@ApplicationScope` (which this project already provides via `DispatcherModule`):

```kotlin
applicationScope.launch { teamsRepository.recordTeamActivity() }
```

Worth deciding explicitly: today, approving a member and immediately backing out of the screen can cancel the activity record mid-probe, both before and after this PR. If that's acceptable, this diff is right as-is; if not, the fix is the application scope, not the nested launch.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16564, opened 2026-08-27
all: smoother utilities hex mapping (fixes #16547)

### LINKED ISSUE #16547: Replace String.format with BigInteger.toString(16) in Utilities.toHex
**rating:** 36/100 &nbsp;·&nbsp; **proposed by:** devin(r2)

`toHex` uses `String.format("%x", BigInteger(1, it))`, paying Formatter overhead whenever sync builds per-user database names.

**Ask:** replace `String.format("%x", BigInteger(1, it))` with `BigInteger(1, it).toString(16)`, which is equivalent for a positive BigInteger and skips the Formatter overhead.

**Files:** `utils/Utilities.kt`

**Verification:** Verified verbatim at Utilities.kt:87-89. Equivalent output for a positive BigInteger.

dogi:
Replaced String.format("%x", BigInteger(1, it)) with BigInteger(1, it).toString(16) in Utilities.toHex to avoid Formatter overhead. Added unit tests for Utilities.toHex.

Fixes #16547

---
*PR created automatically by Jules for task [16132280190545842137](https://jules.google.com/task/16132280190545842137) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6884
-        versionName = "0.68.84"
+        versionCode = 6885
+        versionName = "0.68.85"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/utils/Utilities.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/utils/Utilities.kt
@@ -85,7 +85,7 @@ object Utilities {
     }
 
     fun toHex(arg: String?): String {
-        return arg?.toByteArray()?.let { String.format("%x", BigInteger(1, it)) } ?: ""
+        return arg?.toByteArray()?.let { BigInteger(1, it).toString(16) } ?: ""
     }
 
     fun normalizeText(str: String): String {
--- a/app/src/test/java/org/ole/planet/myplanet/utils/UtilitiesTest.kt
+++ b/app/src/test/java/org/ole/planet/myplanet/utils/UtilitiesTest.kt
@@ -69,4 +69,11 @@ class UtilitiesTest {
         verify(exactly = 1) { mockToast.show() }
         verify(exactly = 1) { anyConstructed<Handler>().post(any()) }
     }
+
+    @Test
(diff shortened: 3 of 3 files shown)


dogi:
## Review: use `BigInteger.toString(16)` in `Utilities.toHex`

**Rating: 90/100** — equivalent output, and it removes an implicit-locale dependency from a function whose result becomes a CouchDB database name.

### Why this is more than cosmetic

`toHex` isn't formatting something for display. It builds server-side identifiers:

```kotlin
// UserRepositoryImpl:621, :644
val table = "userdb-${Utilities.toHex(model.planetCode)}-${Utilities.toHex(model.name)}"
// TransactionSyncManager:124
"userdb-${userModel.planetCode?.let { Utilities.toHex(it) }}-${userModel.name?.let { Utilities.toHex(it) }}"
```

`String.format("%x", …)` with no `Locale` argument uses `Locale.getDefault()`. `BigInteger.toString(16)` is locale-independent by contract. This app ships `ar`, `es`, `fr`, `ne` and `so` locales — Nepali and Arabic being the ones with non-ASCII digit systems — so pinning the output to be locale-free on a code path that constructs a remote database name is worth doing on its own merits. A mangled name here wouldn't fail loudly; it would target a database that doesn't exist and break sync only for users in those locales, which is about the worst failure shape available.

It also silences the standard `String.format` without-locale lint, and skips the `Formatter` machinery for a two-call path.

### Equivalence

Both produce lowercase hex with no leading zeros and no sign prefix (`BigInteger(1, bytes)` forces a positive signum, so `toString(16)` can't emit `-`). For an empty byte array both yield `"0"` — which the test pins, and which is the one case where "no leading zeros" produces something surprising.

The `?: ""` for null input is untouched.

### Test

Covers the three meaningful inputs: a normal string (`"hello"` → `68656c6c6f`, verifiable by hand), the empty string (`"0"`, the degenerate `BigInteger.ZERO` case), and null (`""`). That's the full behaviour of the function.

Nit: `org.junit.Assert.assertEquals` is written fully qualified three times; the file already uses JUnit assertions elsewhere, so importing it would match.

### Worth a follow-up

Since these hex values become database names, a round-trip test against a known `planetCode`/`name` pair — asserting the exact `userdb-…-…` string the server expects — would be more valuable than testing `toHex` in isolation. That's the assertion that would catch a future change to this function before it reaches a user's sync.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16565, opened 2026-08-27
teams: smoother voices image caching (fixes #16552)

### LINKED ISSUE #16552: Cache the imagesArray size and use the isEmpty idiom in VoicesAdapter
**rating:** 35/100 &nbsp;·&nbsp; **proposed by:** jules(r1), minimax(r2)

`imagesArray.size()` is called three times in the image-layout branch — the `if`, the `else-if` and the `for` bound.

**Ask:** hoist `imagesArray.size()` into a local reused by all three sites, and swap the emptiness check to `!imagesArray.isEmpty`.

**Files:** `ui/voices/VoicesAdapter.kt`

**Verification:** Verified all three calls at VoicesAdapter.kt:875-883. minimax's version also swaps in `!imagesArray.isEmpty`; jules's only hoists the local.

dogi:
## What

`VoicesAdapter.loadImage()` called `imagesArray.size()` three times in the image-layout branch (the `if` guard, the `else-if` single-image check, and the `for` loop bound). This hoists the size into a single local `size` reused by all three sites and swaps the emptiness check to the `!imagesArray.isEmpty` idiom.

## Why

Repeated `JsonArray.size()` calls on the same array recompute the count each time. Caching it in a local avoids the redundant calls, and `!isEmpty` expresses the intent more directly than `size() > 0`.

## Changes

- `ui/voices/VoicesAdapter.kt` — in `loadImage`, hoist `imagesArray.size()` into `val size` and reuse it for the single-image check (`size == 1`) and the `for` loop bound (`0 until size`); replace `imagesArray.size() > 0` with `!imagesArray.isEmpty`.

## Tests

Adds `VoicesAdapterImagesTest`, a Robolectric test that drives `VoicesAdapter` through the `PAYLOAD_IMAGES_CHANGED` bind path (which exercises `loadImage` in isolation) and asserts the three branches of the `imagesArray` layout logic:

- `empty imagesArray renders neither image view nor image container` — empty array leaves both `imgNews` and `llNewsImages` `GONE` and requests no library resources.
- `single image array entry requests exactly that resourceId` — a one-entry array requests exactly that `resourceId`.
- `multiple image array entries show the image container and request each resourceId in order` — a multi-entry array sets `llNewsImages` `VISIBLE` and requests each `resourceId` in order.

All three pass:
```
./gradlew :app:testDefaultDebugUnitTest --tests "org.ole.planet.myplanet.ui.voices.VoicesAdapterImagesTest" -ProbolectricOffline=true
```

Fixes #16552.

---

@dogi can click here to [continue refining the PR](https://app.all-hands.dev/conversations/95966949-41da-43f7-925a-c66d0eba988b)


dogi:
## Review: image array size caching in `VoicesAdapter`

**Rating: 82/100** — correct, though the caching saves nothing measurable; the tests are the real contribution, and one of them can hang rather than fail.

### Correctness

`size() > 0` → `!imagesArray.isEmpty` (Kotlin property syntax over Gson's `JsonArray.isEmpty()`), `size() == 1` → `size == 1`, and `0 until imagesArray.size()` → `0 until size`. `imagesArray` isn't mutated inside the loop, so the hoisted value can't go stale. Equivalent.

On the framing though: `JsonArray.size()` is `elements.size()` — an O(1) field read on the backing `ArrayList` — and `0 until imagesArray.size()` already evaluated it once to construct the range, not per iteration. So the old code made three constant-time calls and the new one makes one plus an `isEmpty`. There's no allocation and no traversal being saved. The honest justification is that reading `size` once reads better; "caching" oversells it.

Minor: `val size = imagesArray.size()` sits above the `!isEmpty` guard, so it's computed on the empty path too. Irrelevant in cost, slightly odd in ordering.

Using `isEmpty` directly is the right `JsonArray` idiom — worth noting because #16554 is landing the same style of cleanup elsewhere via `asList().isNotEmpty()`, which allocates a view. This file picks the better form.

### Tests

First coverage of `VoicesAdapter`'s image-binding path, and the three cases are well chosen: empty array (neither view shown, no resource requested), single entry (goes down the `loadLibraryImage` branch), and multiple entries (container made visible, **each id requested in order**). Capturing resource ids through the injected `getLibraryResourceFn` rather than mocking the loader is a clean way to observe the branch without touching Glide.

### The one thing I'd change before merge

```kotlin
adapter.submitList(listOf(news)) { committed = true }
while (!committed) {
    org.robolectric.shadows.ShadowLooper.idleMainLooper()
}
```

If the commit callback never fires — a `ListAdapter` short-circuit, a background executor that never runs under the current Robolectric looper mode — this spins forever. A hung job is much worse than a failed one: it burns the whole CI timeout with no diagnostic, and `test.yml` runs two shards where one stuck fork stalls the matrix.

`AsyncListDiffer` posts to the main looper, so a bounded form gives the same result and fails loudly:

```kotlin
shadowOf(Looper.getMainLooper()).idle()
assertTrue("submitList never committed", committed)
```

or, if a loop is genuinely needed, cap the iterations and assert afterwards.

### Nits

- Heavy inline qualification: `io.mockk.mockk`, `com.google.gson.JsonArray`, `org.robolectric.shadows.ShadowLooper`, `android.content.Context`, `org.ole.planet.myplanet.databinding.RowNewsBinding` are all written out rather than imported — about a dozen occurrences in a 109-line file.
- `bindForImages` asserts `imgNews.visibility == GONE` for every caller, and the first test then asserts it again. Either the helper shouldn't assert, or the test shouldn't repeat it.

**Verdict:** mergeable — but I'd bound that `while` loop first; an unbounded spin in a CI test is the kind of thing that only bites once it's forgotten about.

---

Okuro3499 (approved):

PR #16566, opened 2026-08-27
resources: smoother storage dialog handling (fixes #16546)

### LINKED ISSUE #16546: Use BottomSheetDialog.behavior instead of the internal design_bottom_sheet lookup
**rating:** 37/100 &nbsp;·&nbsp; **proposed by:** devin(r3)

StorageBreakdownFragment, StorageCategoryDetailFragment and AddResourceFragment each look up Material's internal `com.google.android.material.R.id.design_bottom_sheet` FrameLayout and call `BottomSheetBehavior.from(it)`, when BottomSheetDialog exposes `behavior` directly.

**Ask:** replace the internal `design_bottom_sheet` lookup + `BottomSheetBehavior.from(it)` in all three fragments with `BottomSheetDialog`'s public `behavior` property.

**Files:** `ui/settings/StorageBreakdownFragment.kt`, `ui/settings/StorageCategoryDetailFragment.kt`, `ui/resources/AddResourceFragment.kt`

**Verification:** All three files exist; the onCreateDialog bodies were not line-verified.

dogi:
Replaced internal com.google.android.material.R.id.design_bottom_sheet frame layout lookup and BottomSheetBehavior.from(it) calls in StorageBreakdownFragment, StorageCategoryDetailFragment, and AddResourceFragment with BottomSheetDialog's public behavior property.

Fixes #16546

---
*PR created automatically by Jules for task [15806425692458510633](https://jules.google.com/task/15806425692458510633) started by @dogi*


dogi:
## Review: use `BottomSheetDialog.behavior` in `BottomSheetDialogFragment` subclasses

**Rating: 91/100** — the modern Material idiom, correct across all three call sites, and it removes a real hazard along the way.

### Correctness

`BottomSheetDialog.getBehavior()` has existed since Material Components 1.1.0 and this project is on 1.14.0. It's lazy — `getBehavior()` calls `ensureContainerAndBehavior()` if the behavior isn't built yet — so it's safe to touch in `onCreateDialog`, before the dialog is shown. That's exactly the usage Material's own documentation shows, so no `onShow` deferral is needed.

Setting `state = STATE_EXPANDED` pre-layout is handled by `BottomSheetBehavior` (it records the pending state and applies it when the view is laid out). If anything this is a small **UX improvement**: with `setOnShowListener` the sheet was laid out at peek height and then expanded, which can show a visible jump on slower devices; setting it before the sheet is attached means it comes up expanded.

`setHideable(true)` → `isHideable = true` in `AddResourceFragment` is the same call in Kotlin property form, and it needs to stay because `skipCollapsed` is only meaningful on a hideable sheet.

### Two things it fixes beyond terseness

- **`setOnShowListener` is a single-slot setter.** Occupying it purely to configure the behavior meant any future need for a show listener in these fragments would have silently clobbered the sheet configuration (or vice versa). Freeing it removes that trap.
- **`AddResourceFragment` called `BottomSheetBehavior.from(it)` three times** to set three properties. Now one `apply` block.

Dropping `findViewById(com.google.android.material.R.id.design_bottom_sheet)` also removes a dependency on a Material *internal* resource id — the kind of thing that breaks silently on a library upgrade, and exactly why `getBehavior()` was added.

### Import handling is right

`FrameLayout` is removed from all three files, and `DialogInterface` is removed from the two where it becomes unused (`StorageBreakdownFragment`, `StorageCategoryDetailFragment`) but **kept** in `AddResourceFragment`, where it's still needed at `:175` for `dialog.setButton(…) { _: DialogInterface?, _: Int -> }`. Easy to get wrong in a sweep like this; it's correct here.

### Note

No tests, which is reasonable — this is dialog plumbing whose contract is the Material API, and the behaviour object isn't observable without an instrumented test (and there's no `androidTest` source set). The three call sites are identical in shape, which is its own kind of verification.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16567, opened 2026-08-27
teams: smoother tasks members assigning (fixes #16561)

### LINKED ISSUE #16561: Use the ViewModel's member-count query before opening the task-assignment dialog
**rating:** 31/100 &nbsp;·&nbsp; **proposed by:** jules(r3)

`TeamsTasksFragment` fetches the full joined-member list just to check whether the team has members before opening the assignment dialog; a count would short-circuit the user-model loads.

**Ask:** use the existing `getJoinedMemberCount` (on TeamsMembersRepository) via the ViewModel to test for members before opening the assignment dialog, instead of loading the full joined-member list.

**Files:** `ui/teams/tasks/TeamsTasksFragment.kt`, `ui/teams/tasks/TeamsTasksViewModel.kt`

**Verification:** Files exist and `getJoinedMemberCount` exists on TeamsMembersRepository; the fragment call site was not line-verified. jules filed the fragment change and the ViewModel passthrough as two tasks.

dogi:
## Description

`TeamsTasksFragment` opened the task-assignment dialog even when there was no one to assign to, leaving the user staring at an empty picker. The first cut tried to short-circuit on a `getJoinedMemberCount` query, but the review pointed out two problems: it added a `COUNT` query to the common path (a team *with* members now ran COUNT **and** the full member load), and it still let the empty dialog through when joined members existed but none had names (count > 0, filtered list empty).

This revision drops the count query entirely and guards on the list that the dialog actually renders.

## Changes

- `ui/teams/tasks/TeamsTasksFragment.kt` — both assignment-dialog entry points (the "assign member" button in `showTaskAlert` and the "more" assign action in `onClickMore`) now share a single `private suspend fun loadAssignableMembers(): List<UserEntity>?`. It runs the existing `getJoinedMembers` query, applies the existing name filter, and returns `null` (after showing `no_members_task`) when the filtered list is empty. One query instead of two, and the empty picker is avoided both when the team has no members and when its members all have blank names.
- `ui/teams/tasks/TeamsTasksViewModel.kt` — removed the `getJoinedMemberCount` passthrough; the repository method itself is unchanged and still used by `TeamDetailFragment` and `RequestsViewModel`.
- `ui/teams/tasks/TeamsTasksViewModelTest.kt` — removed the test for the dropped passthrough (14 tests remain).

## Verification

The environment here has no JDK/Android SDK, so the test suite could not be re-run locally. The change is mechanical and reuses the existing `getJoinedMembers` / filter / `no_members_task` toast already present in both call sites; CI (`test.yml`, `build.yml`) will run `testDefaultDebugUnitTest` and the assemble matrix on push.

Addresses the review feedback on this PR (dogi): guard on `filteredUserList.isEmpty()` after the existing query rather than a count query in front of it, and lift the duplicated block into one helper so the two entry points cannot diverge.

Fixes #16561.

---


dogi:
## Review: short-circuit the task assignment dialog on joined-member count

**Rating: 70/100** — the UX goal is right (a toast beats an empty dialog), but the implementation adds a query to the common path and still lets the empty dialog through in a case it was meant to catch.

### It isn't a short-circuit; it's an extra query

```kotlin
if (teamsTasksViewModel.getJoinedMemberCount(teamId) == 0) { … return@launch }
val userList = teamsTasksViewModel.getJoinedMembers(teamId)
```

When the team **has** members — the overwhelmingly common case — this now runs a `COUNT` query *and* the full member load, where before it ran one query. When the team has none, it avoids materialising an empty list, which costs nothing. So the change makes the common path measurably slower and the rare path no faster. The title frames it as an optimisation; it's the opposite.

### The guard doesn't cover the case the dialog actually depends on

The dialog is populated from `filteredUserList`, not `userList`:

```kotlin
val userList = teamsTasksViewModel.getJoinedMembers(teamId)
val filteredUserList = userList.filter { user -> user.getFullName().isNotBlank() || !user.name.isNullOrBlank() }
```

So a team with, say, three joined members whose names are all blank gives `getJoinedMemberCount(teamId) == 3` → the guard passes → `filteredUserList` is empty → **the empty dialog appears anyway, with no toast**. That's the same defect the PR exists to fix, just reached by a different route.

### Both problems have the same one-line fix

Drop the count query and guard on the list you're about to render:

```kotlin
val userList = teamsTasksViewModel.getJoinedMembers(teamId)
val filteredUserList = userList.filter { user -> user.getFullName().isNotBlank() || !user.name.isNullOrBlank() }
if (filteredUserList.isEmpty()) {
    Toast.makeText(context, R.string.no_members_task, Toast.LENGTH_SHORT).show()
    return@launch
}
```

One query instead of two, and it covers both "no members" and "no *nameable* members". `getJoinedMemberCount` then needs no new ViewModel wrapper at all — the repository method already exists (`TeamsMembersRepository:17`, used by `RequestsViewModel` and `TeamDetailFragment`), so nothing is lost by not adding another caller.

### Duplication

The guard is copy-pasted at `:128` and `:364`, on top of the `getJoinedMembers` → filter → build-dialog block that was already duplicated between those two sites. This PR makes the duplicated region longer. Since both call sites want identical behaviour, extracting a single `private suspend fun loadAssignableMembers(): List<UserEntity>?` that returns null after showing the toast would collapse both and guarantee they can't diverge — which matters here, because the fix above has to be applied twice otherwise.

### Good parts

`R.string.no_members_task` already exists and is translated in all five non-default locales (`ar`, `es`, `fr`, `ne`, `so`) — no localisation debt introduced, which is easy to get wrong. The ViewModel test is a clean delegation check.

**Verdict:** needs a change. Guard on `filteredUserList.isEmpty()` after the existing query instead of adding a count query in front of it, and lift the duplicated block while you're there.

---

dogi:

---

dogi:
@dogi Good catch on both points — implemented as suggested in f4072619b40703570e3388f2a7922424ecfde264.

**Dropped the count query, guard on the filtered list.** Both assignment-dialog entry points (`showTaskAlert`'s assign button and `onClickMore`) now share one `private suspend fun loadAssignableMembers(): List<UserEntity>?` that runs the existing `getJoinedMembers` query, applies the existing name filter, and returns `null` (after the `no_members_task` toast) when the filtered list is empty. So:

- the common path (team has members) runs **one** query, not a COUNT + full load;
- the empty dialog is avoided both when the team has no members **and** when it has members whose names are all blank (count > 0 but `filteredUserList` empty) — the gap the count guard left open.

**Removed the `getJoinedMemberCount` ViewModel wrapper** and its test, since it's no longer needed here. The repository method itself is untouched and still used by `TeamDetailFragment` and `RequestsViewModel`.

I didn't see an inline thread from @Okuro3499 to follow up on (none on the PR yet); happy to adjust further if more comes through.

_This reply was generated by an AI agent (OpenHands) on behalf of the user._

dogi:
## Review: 85/100 — mergeable

Much better than the earlier count-based short-circuit. This is a straight extraction of two byte-identical blocks, and the control flow survives intact.

**The `null`-means-handled contract is the right shape here.** `loadAssignableMembers()` shows the toast and returns `null`; both call sites do `?: return@launch`, which reproduces the original `return@launch` exactly. The filter predicate, the toast, and the string resource are all copied verbatim, so there's nothing to diff behaviourally.

### Worth a follow-up: the filter it now centralises is very nearly a no-op

```kotlin
userList.filter { user -> user.getFullName().isNotBlank() || !user.name.isNullOrBlank() }
```

`getFullName()` is `"$firstName $lastName"` (`UserEntity:150`), and Kotlin string templates render `null` as the literal text `"null"`. So for a user with both name parts null, `getFullName()` returns `"null null"` — which is emphatically **not blank**, and the user passes the filter. The only users this excludes are ones whose `firstName` and `lastName` are both *empty strings* (giving `" "`) **and** whose `name` is null or blank.

So the "no members to assign" toast almost never fires, and users with no usable display name still reach the assignee dialog — where they'd render as `"null null"`. That's pre-existing and out of scope here, but this PR is what makes it fixable in one place instead of two, which is a fair argument for the extraction on its own.

Related: #16422 in this queue removes a dead `member.toString() == " "` check that's a vestige of the same `"$firstName $lastName"` behaviour. The `" "`-when-empty quirk is showing up in more than one place; a `UserEntity.displayNameOrNull()` helper would retire both.

### Notes

- `Toast.makeText(context, …)` inside a `suspend` function will NPE if the fragment is detached when the coroutine resumes. Both original sites had the same exposure and `viewLifecycleOwner.lifecycleScope` cancels on view destruction, so this isn't new — just inherited.
- No test. `loadAssignableMembers` is a `private suspend` member of a Fragment, so it isn't reachable from a unit test as written. If the filter is fixed later, pulling the predicate out as a pure function (`fun List<UserEntity>.assignable(): List<UserEntity>`) would make that fix testable — worth doing at the same time.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16574, opened 2026-08-27
courses: smoother selecting (fixes #16556)

### LINKED ISSUE #16556: Count instead of filtering in the CoursesAdapter selection helpers
**rating:** 33/100 &nbsp;·&nbsp; **proposed by:** jules(r3)

`areAllSelected` builds `currentList.filter { ... }` only to read `.size`, and `selectAllItems` rebuilds the same filtered list — an allocation on every selection-checkbox tap.

**Ask:** replace the `filter { ... }.size` in `areAllSelected` with a `count { ... }`, and avoid rebuilding the same filtered list in `selectAllItems`.

**Files:** `ui/courses/CoursesAdapter.kt`

**Verification:** Verified at CoursesAdapter.kt:195-206.

dogi:
Optimized selection helpers in CoursesAdapter.kt by replacing `currentList.filter { ... }.size` with `currentList.count { ... }` in `areAllSelected()` and using `currentList.filterTo(selectedItems) { ... }` in `selectAllItems()` to avoid intermediate list allocations on selection checkbox taps.

Fixes #16556

---
*PR created automatically by Jules for task [7580200885663053405](https://jules.google.com/task/7580200885663053405) started by @dogi*


dogi:
## Review: optimize `CoursesAdapter` selection helpers to prevent list allocations

**Rating: 88/100** — both rewrites are exactly equivalent and each removes a full intermediate list from a path that runs on every selection change.

### `areAllSelected`

```kotlin
- val selectableCourses = currentList.filter { isMyCourseLib || !it.isMyCourse }
- return selectedItems.size == selectableCourses.size && selectableCourses.isNotEmpty()
+ val count = currentList.count { isMyCourseLib || !it.isMyCourse }
+ return count > 0 && selectedItems.size == count
```

`filter { … }.size` → `count { … }` drops the list entirely; only the tally was ever used. Reordering the conjunction to put `count > 0` first is a free short-circuit and reads better as "there is something selectable, and all of it is selected" — the empty-list guard is now the precondition rather than an afterthought, which is the right emphasis since an empty selectable set returning `true` would be the bug.

### `selectAllItems`

```kotlin
- val selectableCourses = currentList.filter { isMyCourseLib || !it.isMyCourse }
- selectedItems.addAll(selectableCourses)
+ currentList.filterTo(selectedItems) { isMyCourseLib || !it.isMyCourse }
```

`filterTo` appends directly to the destination instead of building a list to copy from. `selectedItems.clear()` runs immediately above, so the destination starts empty and ends with the same contents in the same order (or the same set, if `selectedItems` is set-backed — either way `addAll` and `filterTo` agree).

Both helpers are called on selection changes and menu invalidation, and `currentList` can be the whole course library, so this is a real if modest saving on a UI-thread path.

### Gap

No test, and `CoursesAdapterTest` already exists. The behaviour most worth pinning is the one the reordering touches:

```kotlin
@Test fun `areAllSelected returns false when nothing is selectable`()
```

i.e. `isMyCourseLib = false` with a list where every course has `isMyCourse = true` → `count == 0` → must be `false`, not `true`. That's the case where a naive `selectedItems.size == count` without the guard would wrongly report "all selected" for an empty set and enable a select-all menu item that does nothing. The guard is present and correct here; a test would keep it that way.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16488, opened 2026-08-27
courses: smoother progress submissions repositories dao querying (fixes #16484)

dogi:
## Summary

Removes the identity `map { it }` allocations that copy DAO lists with no transformation, so the DAO lists are used directly instead of being reallocated. This was reported in #16484 across the course-progress path (full-list allocations with no transformation, immediately before grouping/indexing).

## Changes

Eight no-op `map { it }` / `map { it }`-equivalent copies were deleted:

**`repository/ProgressRepositoryImpl.kt`** (5 sites)
- `getCourseProgress` — `courseStepDao.getByCourseIds(courseIds).map { it }` → direct use
- `fetchCourseData` — `examDao.getByCourseIds(courseIds).map { it }` → direct use
- `fetchCourseData` — `submissionDao.getExamSubmissionsByUser(userId).map { it }` before `groupBy` → dropped
- `submissionMap` — `answerDao.getBySubmissionIds(submissionIds).map { it }` → direct use
- `submissionMap` — `questionDao.getByIds(questionIds).map { it }` → direct use

**`repository/SubmissionsRepositoryImpl.kt`** (3 sites)
- `getExamsByIds` — `examDao.getByIds(examIds).map { it }` → direct use
- `saveSubmission` — `submission.answers?.map { it }.orEmpty()` → `submission.answers.orEmpty()`
- `getSurveysByCourseId` — `examDao.getByCourseIdAndType(courseId, "survey").map { it }` → direct use

## Verification

- Each removed call was a pure identity transform producing a new list with identical elements, so behavior is preserved.
- Existing unit tests already exercise these code paths (`ProgressRepositoryImplTest.testGetCourseProgress`, `fetchCourseData_executes_successfully`/`fetchCourseData_*` flows, and `SubmissionsRepositoryImplTest.saveSubmission upserts submission through Room`).
- Builds and unit tests run via CI (`test.yml`) on push.

This PR was created by an AI agent (OpenHands) on behalf of the repository maintainers.

fixes #16484

@dogi can click here to [continue refining the PR](https://app.all-hands.dev/conversations/d018875a-d305-498c-9ec3-76a56d3ce6eb)

PR #16578, opened 2026-08-27
all: smoother repositories mapping (fixes #16560)

### LINKED ISSUE #16560: Drop the redundant intermediate collections in five repository mapping chains
**rating:** 32/100 &nbsp;·&nbsp; **proposed by:** jules(r2)

Small allocation trims: TagsRepositoryImpl.getLinkIdsForTagNames (`getByNames(...).map { it.id }` before a mapNotNull), LifeRepositoryImpl.updateMyLifeListOrder (`mapIndexed { ... to index }.toMap()`), SubmissionsRepositoryImpl.getUniqueSurveys (`uniqueSurveys.values.toList()`), FeedbackRepositoryImpl (a map/associateBy/map chain), CoursesRepositoryImpl (`mapValues { entry -> entry.value.map { it } }`).

**Ask:** drop the redundant intermediate collections at all five sites. **Note:** only the CoursesRepositoryImpl one (:865) was line-verified — the other four are plausible from file-and-shape but unverified, so read each site before changing it and skip any whose premise does not actually hold.

**Files:** `repository/TagsRepositoryImpl.kt`, `repository/LifeRepositoryImpl.kt`, `repository/SubmissionsRepositoryImpl.kt`, `repository/FeedbackRepositoryImpl.kt`, `repository/CoursesRepositoryImpl.kt`

**Verification:** The CoursesRepositoryImpl mapValues copy is verified (:865). The rest are file-and-shape plausible but were not line-verified; each is a 2-6 line change with essentially no measurable payoff. Shipped as one entry because none justifies its own PR.

dogi:
Eliminated redundant collection allocations in repository mapping chains:
- CoursesRepositoryImpl: Removed redundant .mapValues { entry -> entry.value.map { it } } copy on groupBy result.
- FeedbackRepositoryImpl: Replaced Pair intermediate list allocation in insertFeedbackList with zip transformation.
- LifeRepositoryImpl: Replaced mapIndexed.toMap() with buildMap in updateMyLifeListOrder.
- TagsRepositoryImpl: Deferred .map { it.id } transformation until after checking if matching tags list is non-empty.
- SubmissionsRepositoryImpl: Verified and skipped as uniqueSurveys.values.toList() creates a single list directly.

Fixes #16560

---
*PR created automatically by Jules for task [11062840607882888787](https://jules.google.com/task/11062840607882888787) started by @dogi*


dogi:
## Review: drop redundant intermediate collections in repository mapping chains

**Rating: 90/100** — four independent cleanups, each a provable equivalence, each removing real allocation.

### The four

**`CoursesRepositoryImpl:865`** — `.mapValues { entry -> entry.value.map { it } }` deleted. An identity copy of *every group* produced by `groupBy`, so proportional to the total step count, not the group count. This is the same Realm-era `.map { it }` residue #16488 is clearing elsewhere, and it was on the list of remaining sites I flagged there — good follow-through.

**`FeedbackRepositoryImpl.insertFeedbackList`** — `map { it to id }` → a plain `ids` list plus `zip(ids) { json, id -> … }`. The old form allocated *n* `Pair` objects and then a second list just to extract `.second`. `zip` with a transform lambda never materialises the pairs at all. Sizes match by construction (`ids` is derived from `jsonObjects`), so `zip`'s min-length truncation can't drop anything, and order is preserved.

**`LifeRepositoryImpl.updateMyLifeListOrder`** — `mapIndexed { … item._id to index }.toMap()` → `buildMap(list.size) { forEachIndexed { put(...) } }`. Removes *n* `Pair`s and the intermediate list, and pre-sizes the map. Duplicate-key semantics are identical: `toMap()` on a pair list lets later entries overwrite earlier ones, and so does `put` in iteration order. Both produce a `LinkedHashMap`, so `idToIndex.keys` on the next line still iterates in list order — which matters, because that ordering feeds the weight assignment.

**`TagsRepositoryImpl.getLinkIdsForTagNames`** — the emptiness check moves from the projection to the source. `matchingTags.isEmpty()` is true exactly when `matchingTags.map { it.id }.isEmpty()` is, so the guard is unchanged; the difference is that the no-match path no longer builds an id list before discarding it. It also reads better — checking the thing rather than a projection of it.

### No tests, and that's defensible here

Each change is a local equivalence over a pure expression, and the surrounding methods already have coverage (`CoursesRepositoryImplTest`, `LifeRepositoryImplTest`, the feedback repository tests). The one I'd have most wanted a test for is the `LifeRepositoryImpl` map, since duplicate `_id` values and iteration order both feed the persisted weights — but the semantics are identical under both forms, so a test would be locking in behaviour rather than guarding this diff.

### Coordination

Two of these files are contended, though in different regions:

- `LifeRepositoryImpl` — #16501 (`change`) edits `getMyLifeForDashboard` (~`:77`), #16437 and #16368 also touch this file. This PR is at `:38`.
- `TagsRepositoryImpl` — #16554 (`change`) edits `:150`. This PR is at `:55-62`.

Should merge cleanly, but a base merge and CI re-run for whichever lands later.

### Follow-up

`.map { it }` still survives at `CoursesRepositoryImpl:436, :443, :455, :528, :529, :879` and `TeamsRepositoryImpl:209, :303, :370, :1291`. Between this PR and #16488 the pattern is nearly gone; one more sweep would finish it.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16581, opened 2026-08-27
all: smoother storage categories sharing (fixes #16570)

### LINKED ISSUE #16570: Consolidate the storage-category extension metadata inside the storage UI package
**rating:** 29/100 &nbsp;·&nbsp; **proposed by:** kimi(r1)

StorageBreakdownFragment hard-codes the category definitions including each extension set, while StorageCategoryDetailFragment re-carries those sets as Bundle args (ARG_EXTENSIONS/ARG_ALL_KNOWN) and pushes them into the ViewModel — extension semantics duplicated across two fragments and threaded as two raw Set<String>.

**Ask:** consolidate the category/extension metadata into one place inside the storage UI package, so the extension sets are not duplicated across the two fragments and threaded as raw `Set<String>` Bundle args. **Note:** the Bundle-arg threading was not verified and this is ~80 changed lines for a pure-internal tidy — read the actual call sites first and keep the change contained.

**Files:** `ui/settings/StorageBreakdownFragment.kt`, `ui/settings/StorageCategoryDetailFragment.kt`, `ui/settings/StorageCategoryViewModel.kt`

**Verification:** The three files exist and the categories list is in StorageBreakdownFragment as described; the Bundle-arg threading was not verified. ~80 changed lines for a pure-internal tidy.

dogi:
Consolidates the storage category/extension metadata into a single `StorageCategories` object inside the storage UI package, so the extension sets are no longer duplicated across the two fragments or threaded as raw `Set<String>` Bundle args (fixes #16570).

## What changed

**New: `StorageCategories.kt`** — single source of truth inside `ui/settings/` holding the category list (`nameRes` + `extensions`), the derived `allKnownExtensions` union, the `OTHER_INDEX` catch-all constant, and an `indexOf(extension)` lookup used by the storage scan.

**`StorageBreakdownFragment.kt`** — the hard-coded category list (with its inline extension sets and the lazy `extensionToIndex` map) is replaced by a mapping from `StorageCategories.all`; the scan resolves extensions through `StorageCategories.indexOf`. Only scan-specific mutable state (`sizeBytes`, `fileCount`) stays local.

**`StorageCategoryDetailFragment.kt`** — the `ARG_EXTENSIONS`/`ARG_ALL_KNOWN` Bundle args are gone. The fragment now receives just a `categoryIndex` (an Int) and resolves the extension sets from `StorageCategories` when loading the ViewModel, eliminating the duplicated extension semantics.

`StorageCategoryViewModel` is unchanged — it already took the sets as plain parameters, so no signature change was needed.

## Why this approach

- The issue asked for the metadata to be consolidated **inside the storage UI package**, so the new object lives beside the two fragments rather than in a repository or model layer.
- Keeping `categoryIndex` (Int) as the Bundle arg instead of the extension sets keeps the fragment's process-death restoration trivial and avoids re-serializing set semantics across fragments.
- `CategoryData` (scan state) is kept in `StorageBreakdownFragment` since it is per-scan mutable UI state, not static metadata.

## Testing

- **`StorageCategoriesTest`** (new, 6 tests): canonical category order, the catch-all category is last and empty, `allKnownExtensions` union contents, `indexOf` resolution for known extensions, fallback to `OTHER_INDEX` for unknown extensions, and no extension mapped to more than one category.
- All tests pass (`./gradlew testDefaultDebugUnitTest --tests "org.ole.planet.myplanet.ui.settings.StorageCategoriesTest"` → BUILD SUCCESSFUL, 6/6 green).
- Behavior is unchanged: the category sets, the "other" fallback semantics, and the delete-all flow (which relies on `extensions.isEmpty()` meaning "not any known extension") are identical to before.

---
*This PR was created by an AI agent (OpenHands) on behalf of the user.*

@dogi can click here to [continue refining the PR](https://app.all-hands.dev/conversations/0ea40bd7-cf17-46fc-8f56-f3f9f041ed17)


dogi:
## Review: share storage-category metadata across the breakdown and detail fragments

**Rating: 89/100** — the right shape for this: one declaration of the categories, and an index through the `Bundle` instead of serialised extension lists.

### What it fixes

Before, `StorageBreakdownFragment` owned the category/extension table and then **shipped it through fragment arguments** — `putStringArrayList(ARG_EXTENSIONS)` plus `putStringArrayList(ARG_ALL_KNOWN)` — so the detail sheet's notion of "what counts as a PDF" was whatever the breakdown sheet happened to serialise at click time. Two failure modes disappear with this PR: the arguments can no longer be stale or partial after process death and re-creation, and the extension sets can't drift between the two files. Passing a single `Int` is also a much smaller `Bundle`.

Hoisting `extensionToIndex` into the shared object means it's built once per process rather than per fragment instance, and `allKnownExtensions` likewise — the old code recomputed `categories.dropLast(1).flatMap { it.extensions }.toSet()` inside `populateCategoryRows`, i.e. on every render of the sheet.

### Tests

Six, and they cover the invariants rather than just the values:

- canonical order of `all`
- only the last category is the empty catch-all
- `allKnownExtensions` is exactly the union
- `indexOf` for each known category, and the unknown fallback
- **no extension appears in two categories** — a genuine property test, and the one most likely to catch a careless future addition

`assertEquals(R.string.storage_other, all[OTHER_INDEX].nameRes)` is doing important work: it's what stops the hardcoded `OTHER_INDEX = 4` from silently pointing at the wrong entry.

### Three things worth tightening

- **`const val OTHER_INDEX: Int = 4`.** A literal index into a list declared immediately above it. Insert a category anywhere but the end and every unknown extension lands in the wrong bucket. The test catches it, but `val OTHER_INDEX = all.lastIndex` makes it structurally impossible and costs nothing (it doesn't need to be `const`).
- **`indexOfFirst { it.nameRes == category.nameRes }`** in the click listener is a lookup for something already known. `categories` is built by mapping `StorageCategories.all` in order, so position *is* the index — the `filter { it.fileCount > 0 }` is what discards it. Iterating with the index keeps it:
  ```kotlin
  categories.forEachIndexed { index, category ->
      if (category.fileCount == 0) return@forEachIndexed
      …
      itemBinding.root.setOnClickListener {
          StorageCategoryDetailFragment.newInstance(label = name, categoryIndex = index)…
      }
  }
  ```
  That also removes the (currently theoretical) ambiguity if two categories ever shared a `nameRes`.
- **`indexOf` is case-sensitive**, which the test documents (`indexOf("MKV")` → other). The only caller lowercases first (`file.extension.lowercase()`), so nothing is broken — but the function is now public API on a shared object, and the next caller won't necessarily know. `extensionToIndex[extension.lowercase()]` inside `indexOf` would make it safe by construction.

### Coordination

`StorageBreakdownFragment` and `StorageCategoryDetailFragment` are contended right now:

- **#16566** rewrites `onCreateDialog` in both files (`BottomSheetDialog.behavior`) — different region, should merge cleanly.
- **#16607** ("keep storage-category selection state in its ViewModel") targets the same pair and is likely to overlap the detail fragment's argument handling directly. Worth sequencing this one first, since #16607 will want to build on the `categoryIndex` argument rather than the old extension lists.

**Verdict:** mergeable.

---

Okuro3499 (approved):

Issue #16337, opened 2026-08-27
Drop the dead retry-queue API surface


dogi:
`RetryQueue.queueFailedOperations` (:56), `getPendingCount` (:83) and `resetAllPending` (:104) have zero production callers; `RetryRepository.resetAllPending` exists only to serve that dead facade method, and `RetryDao.resetPendingRetryTime` (:38) exists only to serve that. Three layers of interface kept alive by nothing.

**Ask:** delete all four dead methods across the chain (`RetryQueue`, `RetryRepository`/`RetryRepositoryImpl`, `RetryDao`) and their now-unused test coverage.

**Files:** `services/retry/RetryQueue.kt`, `repository/RetryRepository.kt`, `repository/RetryRepositoryImpl.kt`, `data/room/dao/RetryDao.kt`, `test/.../RetryRepositoryImplTest.kt`

**Verification:** confirmed the only references outside the chain itself are in one test.

PR #16583, opened 2026-08-27
feedback: smoother pending query documenting no count variant (fixes #16580)

### LINKED ISSUE #16580: Add count queries to FeedbackDao and RetryDao
**rating:** 22/100 &nbsp;·&nbsp; **proposed by:** minimax(r1)

Both DAOs expose `getPending(): List<...>` with no count variant, so callers that only need a size load the rows.

**Ask:** add count queries only where a caller genuinely needs one. **Two caveats, both load-bearing:** (1) the payoff is unestablished — the original task hedges on whether any caller actually does `.size` on the result, so FIRST check the call sites and, if none of them only need a size, close this without adding an unused query rather than adding dead API. (2) The RetryDao half collides with the in-flight "drop the dead retry-queue API surface" work (issue #16337), which is removing retry API — do not re-add surface there; if in doubt, do the FeedbackDao half only and note why.

**Files:** `data/room/dao/FeedbackDao.kt`, `data/room/dao/RetryDao.kt`, `+ FeedbackRepository/RetryRepository and impls`

**Verification:** The DAOs exist. Both tasks hedge on whether any caller actually does `.size` ('If fetchPendingItems calls .size on the result, update it'), so the payoff is unestablished. Note the RetryDao half collides with the dead-retry-surface task above.

dogi:
fixes #16580

## Summary

Investigated whether `FeedbackDao` and `RetryDao` need a count variant of their `getPending()` queries, per the issue's two load-bearing caveats. Conclusion: **no count queries are added** — adding them would be dead API, exactly the outcome the issue anticipates and instructs ("close this without adding an unused query rather than adding dead API").

### FeedbackDao.getPending() — no count-only caller (caveat 1)

The sole production call chain:

```
FeedbackDao.getPending()
  -> FeedbackRepositoryImpl.getPendingFeedback()
    -> UploadConfigs.kt:201  fetchPendingItems = { feedbackRepository.getPendingFeedback() }
      -> UploadCoordinator.queryItemsToUpload() / queryItemsToUploadRoom()
```

`UploadCoordinator` invokes `fetchPendingItems` to get the **full list** of pending feedback rows, then `mapNotNull { ... }` each row into a `PreparedUpload` to serialize and HTTP-upload it. The only size-like use is `itemsToUpload.isEmpty()` (an early-return guard), performed **after** loading rows the coordinator needs anyway. No caller only needs a `.size`, so a `COUNT(*)` would never replace the row fetch — it would be unused. Per caveat 1, no count variant is added.

A short KDoc is added on `FeedbackDao.getPending()` recording this deliberate omission so the same non-issue isn't reopened by a future contributor/agent.

### RetryDao — left untouched (caveat 2)

- `RetryDao.getPending(now)` is consumed by `RetryQueueWorker.doWork()`, which `chunked(BATCH_SIZE).forEach { processOperation(...) }` — it iterates and processes every returned op. It needs the full list; a count would not replace it.
- `RetryDao` **already** has `getActiveCount()` (`COUNT(*) WHERE status='pending' OR status='in_progress'`), which backs the one genuine count need: the Settings retry-queue snapshot (`getRetryQueueSnapshot` → `SettingsActivity`, where the count drives the "...and N more" line and button enable state). So the real count need is already satisfied.
- Per caveat 2, no retry surface is re-added: #16337 ("Drop the dead retry-queue API surface") is actively **removing** retry API. Adding a new `RetryDao` count query now would collide with that in-flight removal.

## Changes

- `data/room/dao/FeedbackDao.kt`: added a 3-line KDoc on `getPending()` documenting why no count variant exists.

No behavior change; no test needed (documentation-only, per the contribution guidelines).

## Why not add the queries

| DAO | Genuine count-only caller? | Action |
|-----|----------------------------|--------|
| FeedbackDao | No — only caller needs full rows to upload | None (documented) |
| RetryDao | Already served by existing `getActiveCount()`; `getPending` callers need rows | None (collides with #16337) |

@dogi can click here to [continue refining the PR](https://app.all-hands.dev/conversations/05b697da-84a3-4862-a536-d929775f3499)

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/data/room/dao/FeedbackDao.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/data/room/dao/FeedbackDao.kt
@@ -17,9 +17,9 @@ interface FeedbackDao {
     @Query("SELECT * FROM feedback WHERE owner IS :owner ORDER BY openTime DESC")
     fun getByOwnerFlow(owner: String?): Flow<List<Feedback>>
 
-    // No count variant by design: the only caller (UploadCoordinator via
-    // FeedbackRepository.getPendingFeedback) needs the full rows to serialize and upload each
-    // one, so a COUNT(*) here would be dead API.
+    // No count variant by design: consumers (UploadCoordinator via
+    // FeedbackRepository.getPendingFeedback) serialize and upload each returned row, so they
+    // need the full list — a COUNT(*) here would be dead API.
     @Query("SELECT * FROM feedback WHERE isUploaded = 0")
     suspend fun getPending(): List<Feedback>
 
--- a/app/src/main/java/org/ole/planet/myplanet/data/room/dao/FeedbackDao.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/data/room/dao/FeedbackDao.kt
@@ -17,6 +17,9 @@ interface FeedbackDao {
     @Query("SELECT * FROM feedback WHERE owner IS :owner ORDER BY openTime DESC")
     fun getByOwnerFlow(owner: String?): Flow<List<Feedback>>
 
+    // No count variant by design: the only caller (UploadCoordinator via
+    // FeedbackRepository.getPendingFeedback) needs the full rows to serialize and upload each
+    // one, so a COUNT(*) here would be dead API.
     @Query("SELECT * FROM feedback WHERE isUploaded = 0")
     suspend fun getPending(): List<Feedback>
 


dogi:
## Review: document that `FeedbackDao.getPending` has no count variant by design

**Rating: 86/100** — the claim in the comment is the whole deliverable, and I verified it end to end. It's accurate.

### The call chain is exactly as stated

- `feedbackDao.getPending()` — one production caller: `FeedbackRepositoryImpl:86`.
- `getPendingFeedback()` — one production caller: `UploadConfigs:201`, `fetchPendingItems = { feedbackRepository.getPendingFeedback() }`.
- That config is consumed by `UploadCoordinator`, which serialises and POSTs each row individually.

So the consumer genuinely needs the rows, not a tally, and a `COUNT(*)` variant would have no caller. The comment's reasoning holds.

### Why this is worth a comment rather than nothing

The codebase has a visible habit of pairing list queries with count queries — `MyLifeDao.countByUserId`, `TeamsMembersRepository.getJoinedMemberCount`, `VoicesRepository.countTeamChats`, and #16475 is adding two more right now. Against that backdrop, the *absence* of `countPending()` reads like an oversight, and the natural "fix" is to add one. Recording the reason costs three lines and saves the next person the call-chain trace I just did.

### One suggestion on wording

"the only caller" is the kind of claim that rots silently — a second caller can appear without anyone touching this comment. Two small changes would make it age better:

- Anchor it: "as of #16583, the only caller is …" or drop the count and say "callers need the full rows to serialize and upload each one", which stays true even if a second uploader appears.
- Name the constraint rather than the current state: the reason a count is useless here isn't that there's one caller, it's that the consumer is an uploader that must read every row.

Neither is blocking — an accurate comment that may drift is better than the nothing that was there.

**Verdict:** mergeable.

---

Okuro3499 (commented):
this just a comment

dogi:
@openhands follow up on my review and/or @Okuro3499 s improvement

---

olevim:

olevim:
Followed up on @dogi's wording suggestion (and checked for @Okuro3499's instructions — see below).

**@dogi's review — addressed (non-blocking wording fix).** The KDoc on `FeedbackDao.getPending()` was reworded to name the *constraint* rather than the rotting current-state claim:

```diff
- // No count variant by design: the only caller (UploadCoordinator via
- // FeedbackRepository.getPendingFeedback) needs the full rows to serialize and upload each
- // one, so a COUNT(*) here would be dead API.
+ // No count variant by design: consumers (UploadCoordinator via
+ // FeedbackRepository.getPendingFeedback) serialize and upload each returned row, so they
+ // need the full list — a COUNT(*) here would be dead API.
```

The reasoning a count is useless here is now stated as the property of the consumer ("serialize and upload each returned row → need the full list"), which stays true even if a second uploader is added later, rather than as "the only caller," which drifts silently. Commit: `dcc97fdd`.

**@Okuro3499's review — no actionable instruction found.** The review is a `COMMENTED` review whose body is "this just a comment", with zero inline comments, so there were no improvement instructions to follow. If a concrete suggestion was intended, happy to act on it once it lands.

No behavior change — documentation-only, as before.

_This comment was created by an AI agent (OpenHands) on behalf of the repository maintainers._

dogi:
## Review — rating 80/100

A three-line comment documenting a deliberate absence. I traced the claim rather than taking it at face value, and it holds exactly:

```
FeedbackDao.getPending()
  └─ FeedbackRepositoryImpl.getPendingFeedback()   (bare passthrough, :85-86)
       └─ UploadConfigs.kt:201  fetchPendingItems = { feedbackRepository.getPendingFeedback() }
```

That is the **only** consumer, and the surrounding `RoomUploadConfig` runs `serializer` over each item and `markUploaded` over each result — so it needs the rows themselves. A `COUNT(*)` variant would have no caller. The comment is accurate and sits where someone would look before adding one.

This is the right kind of comment: it records why something *isn't* there, which is the category that otherwise gets rediscovered and "fixed" every few months.

### Two small refinements

The comment says "consumers" (plural) where there is exactly one. And it names `UploadCoordinator`, but the actual reference lives in `UploadConfigs.kt:201` — pointing there directly would save the next reader a hop:

```kotlin
// No count variant by design: the sole consumer (UploadConfigs.Feedback via
// FeedbackRepository.getPendingFeedback) serializes and uploads each returned row, so it
// needs the full list — a COUNT(*) here would be dead API.
```

Worth noting the same reasoning applies to roughly a dozen sibling `fetchPendingItems` configs — `getPendingNewsLogUploads`, `getPendingCourseProgressUploads`, `getPendingApkLogs` and so on all follow this fetch-all-then-upload-each shape. If this question keeps coming up, a single note on the shared upload path would cover them all rather than one comment per DAO. Not a reason to hold this.

Moving to `ready`.

---

PR #16584, opened 2026-08-27
Reuse already-loaded step resources in refreshInlineResources (fixes #16569)

### LINKED ISSUE #16569: Reuse the already-loaded step resources in refreshInlineResources
**rating:** 30/100 &nbsp;·&nbsp; **proposed by:** deepseek(r3)

`CourseStepFragment.refreshInlineResources` issues a second `resourcesRepository.getAllStepResources(stepId)` although the fragment already holds the same list in its `resources` field from onViewCreated.

**Ask:** reuse the already-loaded `resources` field in `refreshInlineResources` instead of re-querying `getAllStepResources(stepId)`. **Careful with ordering:** the resources must actually be loaded before this path runs — verify that holds before removing the query. The second query was not line-verified, so read the site first.

**Files:** `ui/courses/CourseStepFragment.kt`

**Verification:** File exists; the second query was not line-verified. Shares a file with the counts task. The task itself flags the ordering hazard (resources must be loaded first).

dogi:
Updated `CourseStepFragment.kt` to reuse the already-loaded `resources` property in `refreshInlineResources()` rather than querying `resourcesRepository.getAllStepResources(stepId)` again. Also added a safety check using `::resources.isInitialized` before submitting to the adapter.

Fixes #16569

---
*PR created automatically by Jules for task [6553897503783031811](https://jules.google.com/task/6553897503783031811) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/ui/courses/CourseStepFragment.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/courses/CourseStepFragment.kt
@@ -229,12 +229,10 @@ class CourseStepFragment : BaseContainerFragment(), ImageCaptureCallback {
     }
 
     private fun refreshInlineResources() {
-        viewLifecycleOwner.lifecycleScope.launch {
-            val updatedResources = resourcesRepository.getAllStepResources(stepId)
-            resources = updatedResources
-            inlineResourceAdapter?.submitList(updatedResources)
-            fragmentCourseStepBinding.resourceDownloadProgress.visibility = View.GONE
+        if (::resources.isInitialized) {
+            inlineResourceAdapter?.submitList(resources)
         }
+        fragmentCourseStepBinding.resourceDownloadProgress.visibility = View.GONE
     }
 
     private fun hideTestIfNoQuestion(isTestPresent: Boolean, isSurveyPresent: Boolean) {
--- a/app/src/main/java/org/ole/planet/myplanet/ui/courses/CourseStepFragment.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/courses/CourseStepFragment.kt
@@ -230,7 +230,6 @@ class CourseStepFragment : BaseContainerFragment(), ImageCaptureCallback {
 
     private fun refreshInlineResources() {
         viewLifecycleOwner.lifecycleScope.launch {
-            if (!::resources.isInitialized) return@launch
             val updatedResources = resourcesRepository.getAllStepResources(stepId)
             resources = updatedResources
             inlineResourceAdapter?.submitList(updatedResources)
--- a/app/src/main/java/org/ole/planet/myplanet/ui/courses/CourseStepFragment.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/courses/CourseStepFragment.kt
@@ -229,10 +229,13 @@ class CourseStepFragment : BaseContainerFragment(), ImageCaptureCallback {
     }
 
     private fun refreshInlineResources() {
-        if (::resources.isInitialized) {
-            inlineResourceAdapter?.submitList(resources)
+        viewLifecycleOwner.lifecycleScope.launch {
+            if (!::resources.isInitialized) return@launch
+            val updatedResources = resourcesRepository.getAllStepResources(stepId)
+            resources = updatedResources
+            inlineResourceAdapter?.submitList(updatedResources)
+            fragmentCourseStepBinding.resourceDownloadProgress.visibility = View.GONE
         }
-        fragmentCourseStepBinding.resourceDownloadProgress.visibility = View.GONE
     }
 
     private fun hideTestIfNoQuestion(isTestPresent: Boolean, isSurveyPresent: Boolean) {


dogi:
## Review: reuse already-loaded step resources in `refreshInlineResources`

**Rating: 35/100** — this removes the only mechanism that refreshes the inline resource rows after a download, and the replacement is a no-op. After this change `onDownloadComplete()` does nothing but hide the spinner.

### The one caller is `onDownloadComplete`

```kotlin
override fun onDownloadComplete() {
    super.onDownloadComplete()
    refreshInlineResources()
}
```

That's the sole call site (`:338`). It fires **after a resource download finishes**, and its entire job is to make the rows reflect the new on-disk state — `resourceLocalAddress`, `resourceOffline`, `downloadedRev` — which live in Room and were just written by the download pipeline.

`resources` is assigned exactly once, at `:113`, from `loadStepData()` during `onViewCreated`. So at the moment `onDownloadComplete` runs, the in-memory list holds the **pre-download** row objects. There is nothing already-loaded to reuse: the data is stale by construction at precisely the moment this callback fires. The `getAllStepResources(stepId)` call that was deleted is what picked up those changes.

### And the replacement can't rebind even if the data were fresh

```kotlin
if (::resources.isInitialized) {
    inlineResourceAdapter?.submitList(resources)
}
```

`AsyncListDiffer.submitList` begins with a **reference-equality** short-circuit — if the new list is the same instance as the current one, it returns immediately without diffing or dispatching any updates. `resources` is never reassigned after `:113`, so the adapter is being handed the identical instance it already holds. The call returns without touching a single view holder.

Net effect of the new `refreshInlineResources()`: `resourceDownloadProgress.visibility = View.GONE`.

The user-visible consequence: download a resource inline from a course step, the spinner disappears, and the row still shows its pre-download state (offering a download rather than an open) until the fragment is destroyed and recreated.

### What I'd suggest instead

If the goal is to avoid a redundant DB read, the query isn't redundant — it's the refresh. Options that keep the behaviour:

1. **Keep the re-query** and just add the guard that this PR usefully introduces:
   ```kotlin
   private fun refreshInlineResources() {
           val updated = resourcesRepository.getAllStepResources(stepId)
           resources = updated
           inlineResourceAdapter?.submitList(updated)
           fragmentCourseStepBinding.resourceDownloadProgress.visibility = View.GONE
       }
   }
   ```
   The `::resources.isInitialized` idea is worth keeping in spirit — it guards the case where a download completes before `loadStepData` has assigned `resources` — but as a guard on the *read*, not a replacement for it.
2. **Make it reactive.** `ResourcesRepository` already exposes Flow-based queries (`getMyLibraryFlow`, `getRecentResources`, `observeOpenedResourceIds`). A `Flow`-backed step-resource query collected with `collectWhenStarted` would update the list on any DB change, and `onDownloadComplete` could drop to just hiding the spinner — which is what this PR wants it to be, but correct.

**Verdict:** needs a change. The `isInitialized` guard is a good idea; the removal of the re-query is the bug.

---

dogi:
## Review: 55/100 — needs a change

Good news first: the earlier revision's `submitList(resources)` no-op is gone — the DB re-query is back and the adapter now receives a fresh list, so the "downloaded resources never update their row" bug I flagged is resolved.

What's left is a single added line, and it does more harm than good.

### The guard protects against something that can't happen

```kotlin
private fun refreshInlineResources() {
        if (!::resources.isInitialized) return@launch
        val updatedResources = resourcesRepository.getAllStepResources(stepId)
        resources = updatedResources
        inlineResourceAdapter?.submitList(updatedResources)
        fragmentCourseStepBinding.resourceDownloadProgress.visibility = View.GONE
    }
}
```

The body only ever **assigns** `resources` (:234) — it never reads it. Assigning a `lateinit var` is always legal, so `UninitializedPropertyAccessException` was never reachable from this function. The binding access is safe too: `viewLifecycleOwner.lifecycleScope` cancels the coroutine when the view goes away, so `fragmentCourseStepBinding` can't be touched after `onDestroyView`.

### And when it does fire, it leaves the progress spinner stuck

`refreshInlineResources()` is called only from `onDownloadComplete()` (:338), and its last statement hides the download indicator:

```kotlin
fragmentCourseStepBinding.resourceDownloadProgress.visibility = View.GONE
```

If a download completes before `resources` is assigned at :113 — e.g. a download kicked off on a previous visit finishing just after this fragment attaches, before the initial load emits — the new early return skips **all** of it, including that line. `master` re-queries, updates the list, and hides the spinner. This branch leaves `resourceDownloadProgress` visible with nothing to clear it.

Narrow window, but the symptom is a permanently spinning progress bar on a course step, which is precisely the sort of thing that gets reported as "the app is stuck".

### The title describes the previous revision

"Reuse already-loaded step resources in refreshInlineResources" is what the *earlier* version did — reuse `resources` instead of re-querying — and that's the approach that produced the no-op. The current diff keeps the re-query. Since the queue squash-merges on the PR title, this needs updating regardless of what happens to the guard.

### Recommendation

Drop the line. `master`'s `refreshInlineResources` is already correct, and with the guard removed this PR is empty — which is the right outcome if the goal was the reuse optimization and that turned out not to work.

If there's a real crash driving this, it isn't in `refreshInlineResources` — the reads of `resources` are at :121, :176, :192 and :196, and those are the ones worth guarding (or, better, initializing `resources` to `emptyList()` and dropping `lateinit` entirely, which removes the whole class of question).

**CI** — build (default + lite), both shards, labels green. Nothing here exercises the guard.

**Verdict:** remove the guard and retitle; or close if the reuse idea is being abandoned.

---

dogi:
## Review — rating 20/100 (was 55) — recommend closing

This PR is now empty. Three commits, `mergeable_state: clean`, and zero changed files against `master` — the change has been fully reverted.

To be clear about what that low rating means: **reverting was the right call.** The score reflects that there is nothing here to merge, not that the work was bad.

### Why the premise does not hold

Confirming for the record, since #16569 is still open. `master`'s version is already correct:

```kotlin
private fun refreshInlineResources() {
        val updatedResources = resourcesRepository.getAllStepResources(stepId)
        resources = updatedResources
        inlineResourceAdapter?.submitList(updatedResources)
        fragmentCourseStepBinding.resourceDownloadProgress.visibility = View.GONE
    }
}
```

The re-query is the whole point of the function. It runs *after a download completes*, precisely to pick up resources that were not present when `resources` was first populated. Reusing the already-loaded property would make the refresh a no-op — the newly downloaded files would never appear until the fragment was recreated. The "redundant query" the issue identifies is load-bearing.

The `::resources.isInitialized` guard was the more concrete problem: it wrapped the `submitList` *and* the `visibility = View.GONE`, so on the uninitialised path the download spinner would never be hidden and would sit on screen indefinitely.

### Recommendation

Close this PR, and close #16569 as invalid. The issue describes the re-query as redundant, which it is not, so any future attempt at it will rediscover the same dead end. If there is a real concern underneath it, it is probably "`refreshInlineResources` re-queries the full step-resource list on every download completion, which is O(downloads) queries" — that would be worth an issue in its own words, and the fix would be batching or debouncing the refresh, not removing the query.

I am not closing anything myself — flagging it for a maintainer.

Moving to `change` to get it out of the review queue; the intended outcome is closure rather than another revision.

---

PR #16589, opened 2026-08-27
courses: smoother steps data handling (fixes #16568)

### LINKED ISSUE #16568: Use the step-exam and resource counts instead of the lists in the course step UI
**rating:** 30/100 &nbsp;·&nbsp; **proposed by:** jules(r1), jules(r3)

`CourseStepFragment` renders `getString(R.string.retake_test, stepExams.size)` (:245/:247) and `getString(R.string.resources_size, resources.size)` (:121) off full lists loaded through CourseStepData.

**Ask:** use counts rather than the full lists for these three string-formatting call sites. **Note:** the payoff here is small — the lists are loaded for rendering anyway — so keep the change minimal and do not restructure the loading. Ignore jules's r3 variant (skip formatting when empty); it is a different and weaker claim about layout passes.

**Files:** `ui/courses/CourseStepFragment.kt`

**Verification:** All three call sites verified. The payoff is small — the lists are loaded for rendering anyway — and jules's r3 variant (skip formatting when empty) is a different, weaker claim about layout passes.

dogi:
Updated CourseStepFragment to use step-exam and resource counts directly from CourseStepData when formatting resources_size, retake_test, and take_test string resources.

Fixes #16568

---
*PR created automatically by Jules for task [11492460560897570467](https://jules.google.com/task/11492460560897570467) started by @dogi*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved course step resource and assessment handling.
  * Resource button counts now accurately reflect the available resources.
  * Empty resource, exam, and survey lists are handled more reliably.
  * Improved visibility behavior for tests without questions.


dogi:
## Review: use step-exam and resource counts in course step UI string formatting

**Rating: 66/100** — this is a no-op that adds an unused parameter and lets one function read the same quantity from two places. Nothing is broken; nothing is gained either.

### Both substitutions are identity

At `:113-114`, immediately above the changed lines:

```kotlin
resources = data.resources
stepExams = data.stepExams
```

So at `:118` `resources.size` **is** `data.resources.size`, and at `:122` `data.stepExams.size` **is** `stepExams.size`. Neither value can differ from what it replaces. The rendered strings are byte-identical.

### The new parameter has no caller and creates a split

```kotlin
private fun hideTestIfNoQuestion(
    isTestPresent: Boolean,
    isSurveyPresent: Boolean,
    stepExamCount: Int = stepExams.size
) {
    …
    if (stepExams.isNotEmpty()) {              // guard: field
        getString(R.string.retake_test, stepExamCount)   // count: parameter
```

`hideTestIfNoQuestion` has exactly one call site (`:122`), which now passes the value explicitly — so the default is dead, and the parameter exists to carry a value identical to the field it sits beside.

More importantly, the guard and the count now come from different sources. Today they agree because the field was assigned from the same `data` eight lines earlier. But the signature now *permits* a caller to pass a count that disagrees with `stepExams`, and the failure would be quiet: the button appears (guard says non-empty) showing "Take test (0)". A private helper reading its own field consistently can't do that.

If the intent is to make `hideTestIfNoQuestion` independent of fragment state — a reasonable goal — the fix is to pass the *list* and drop the field read entirely:

```kotlin
private fun hideTestIfNoQuestion(isTestPresent: Boolean, isSurveyPresent: Boolean, stepExams: List<StepExam>)
```

Then the guard and the count are the same value by construction. Half-parameterising is the worst of the three options.

### The `data.resources.size` change is fine

Reading from the local `data` rather than the just-assigned field is marginally clearer at that site — the reader doesn't have to scroll up to confirm the field holds what they think. I'd keep that half.

### Coordination

#16584 also edits `CourseStepFragment` (`refreshInlineResources`, `:231-238`) and currently carries `change`. Different regions from this diff (`:118-122`, `:240-250`), so a merge should be clean, but both are in flight against the same file.

**Verdict:** needs a change — drop the `stepExamCount` parameter, or pass the list so the guard and the count can't diverge. Keeping the `data.resources.size` line is fine either way.

---

dogi:
## Review: 72/100 — needs a change

The substitutions are safe, but they're a no-op, the title implies otherwise, and the change stops just short of the problem that's actually worth fixing in this file.

### Both reads were already correct

The assignments happen immediately above the reads:

```kotlin
step = data.step
resources = data.resources          // :113
stepExams = data.stepExams          // :114
stepSurvey = data.stepSurvey
…
fragmentCourseStepBinding.btnResources.text =
    getString(R.string.resources_size, resources.size)     // :121 — same value as data.resources.size
hideTestIfNoQuestion(data.hasExam, data.hasSurvey)          // :122 — reads stepExams, assigned at :114
```

So `resources.size` and `data.resources.size` are the same list, and `hideTestIfNoQuestion`'s reads of the `stepExams` field (:243, :245, :247) see exactly `data.stepExams`. There's no stale value and no ordering hazard — nothing observable changes.

That makes the title misleading: **"Use step-exam and resource counts in course step UI string formatting"** reads as a fix to what the counts show, and the counts are identical before and after. The squash-merge would put that claim in the history. What the diff actually does is parameterize `hideTestIfNoQuestion` instead of having it reach for fields — worth doing, but it should be described as that.

### The `lateinit` reads that *can* fail are untouched

`stepExams` is `private lateinit var` (:58), and passing it into `hideTestIfNoQuestion` removes three reads that were never at risk (they run after :114 in the same block). The reads that *are* exposed remain:

- **:96** — `if (stepExams.isEmpty()) true else null`
- **:275** — `if (stepExams.isNotEmpty())`

Both sit outside the `onViewCreated` coroutine, so either can run before :114 assigns and throw `UninitializedPropertyAccessException`. Same for `resources` at :176 and :196. If the goal is to make this fragment's field access safe, that's where the work is — and the cheapest fix is to drop `lateinit` entirely:

```kotlin
private var stepExams: List<StepExam> = emptyList()
private var resources: List<MyLibrary> = emptyList()
```

That removes the whole class of crash and makes the parameterization in this PR unnecessary rather than merely tidy.

### Also

- **No test.** Nothing here is observable, so there's nothing to assert — which is itself a signal about the change's value.
- **Contention:** #16584 also modifies `CourseStepFragment.kt` (adding an `::resources.isInitialized` guard in `refreshInlineResources`). Different hunk, but the two PRs are circling the same `lateinit` problem from opposite ends — #16584 adds a guard, this one removes a read. Doing the `= emptyList()` change once would make both unnecessary.

**Verdict:** retitle to describe the refactor, and consider folding this into the `lateinit` → default-value fix that both this and #16584 are working around.

---

dogi:
## Review — rating 76/100 (was 72)

The `lateinit` → `emptyList()` conversion is a genuine improvement over the previous revision, which was a pure rearrangement. But I traced every read site and it still does not fix a reachable crash — and it walks past a real bug in the same six lines.

### No reachable read happens before assignment

I checked all nine read sites of `resources` / `stepExams` / `stepSurvey` against their callers:

| Read | Reachable before assignment? |
|---|---|
| `resources.size` (121), `setupInlineResources` (176, 192), `autoDownloadResources` (196), `hideTestIfNoQuestion` (243-251), `setListeners` (275, 287) | No — `setListeners()`/`setupInlineResources()`/`autoDownloadResources()` are called at lines 143-145, *inside* the same guarded block, after the assignments at 112-115 |
| `stepExams.isEmpty()` in `launchSaveCourseProgress` (96) | No — reached from line 168 (inside the block) or from `setMenuVisibility`, which returns early on `!::step.isInitialized`, additionally requires `userHasCourse` (false until the block runs), and is wrapped in try/catch |
| `refreshInlineResources` (231-238), the one out-of-band caller via `onDownloadComplete` | No — it *assigns* `resources` before using it |

So this is defensively correct rather than a fix. That is still worth something: `onDownloadComplete` already demonstrates that this fragment gets called into out of band, so the next such entry point would crash on master and degrade gracefully here. Passing `data.resources` / `data.stepExams` directly also reads the source of truth instead of a field assigned three lines earlier, which is marginally clearer. I have no objection to any of it.

The objection is the same one as last time: the title says "Use step-exam and resource counts in course step UI string formatting", which describes a no-op, and this becomes the commit message.

### The real bug, two lines up

`onViewCreated` launches on `viewLifecycleOwner.lifecycleScope` and then gates *everything* on a one-shot check:

```kotlin
loadDataJob = viewLifecycleOwner.lifecycleScope.launch {
    user = userRepository.getUserModel()
    val data = loadStepData()
    if (viewLifecycleOwner.lifecycle.currentState.isAtLeast(Lifecycle.State.STARTED)) {
        step = data.step
        resources = data.resources
        ...
        setListeners()
        setupInlineResources()
        autoDownloadResources()
```

`lifecycleScope.launch` does not suspend at STOPPED — only `repeatOnLifecycle`/`withStarted` do. So if the user backgrounds the app while `loadStepData()` is in flight, the coroutine resumes anyway, the check fails, and the **entire block is skipped**: no field assignments, no listeners, no inline resources, no description text. `loadDataJob` has now completed, so nothing retries. The user returns to the foreground and sits on a permanently blank step until they navigate away and back.

That is the bug in this code, and it is the one that makes the `lateinit` fields dangerous in the first place. The fix is to suspend rather than skip:

```kotlin
viewLifecycleOwner.lifecycle.withStarted {
    step = data.step
    ...
}
```

`withStarted` waits for STARTED instead of dropping the work, so a backgrounded load completes correctly on return.

### Suggestion

Fold that into this PR — it is the same block you are already editing, it makes the `emptyList()` defaults coherent rather than incidental, and it gives the PR a title that describes something real. Failing that, retitle to say this is a defensive-defaults change and open a separate issue for the `isAtLeast` skip.

Moving to `change`.

---

dogi:
## Review — rating 86/100 (was 76)

This now fixes the bug instead of walking past it, and the restructuring around `withStarted` shows the constraints were actually worked through rather than the call being swapped in.

### The blank-step bug is fixed

`lifecycleScope.launch` does not pause at STOPPED, so master's one-shot `isAtLeast(STARTED)` check meant that backgrounding the app during `loadStepData()` skipped the *entire* setup block — no field assignments, no `setListeners()`, no inline resources, no description — and `loadDataJob` then completed, so nothing ever retried. Returning to the foreground left a permanently blank step. `withStarted` suspends until STARTED instead of dropping the work.

Two consequences of that swap were handled correctly, and neither is obvious:

- **`getCourseTitleById` had to move out.** `withStarted` takes a non-suspending block, so the suspend call could not stay inside. Hoisting it to `val title = …` before the block is required — and is incidentally better, since the DB read no longer waits on the lifecycle.
- **`withResumed` had to move out too**, for the same reason. The ordering still holds: `withStarted` suspends until its block has *run*, so `userHasCourse` is assigned by the time `if (userHasCourse)` is evaluated below it. Worth noting this also fixes a second silent skip — progress saving previously never happened at all if the fragment was not STARTED at that instant, and now runs on return.

With `withStarted` in place the `emptyList()` defaults become genuinely defensive rather than load-bearing, which is the right relationship: no reachable read precedes assignment, and a future out-of-band entry point (the file already has one in `onDownloadComplete`) degrades instead of throwing.

CI is green on both shards and both flavors.

### Two notes

`import androidx.lifecycle.Lifecycle` is now unused — `Lifecycle.State.STARTED` was its only reference. A warning rather than an error, so CI stays green, but worth removing.

The title now *undersells* the PR. "Use step-exam and resource counts in course step UI string formatting" describes the no-op part; the substance is "fix course step rendering blank when the app is backgrounded during load". That is the opposite of the usual problem, but it still costs findability — nobody debugging a blank course step will find this commit. Worth retitling before merge; I am not holding the PR for it.

Not blocking, but if you want the fix pinned: a Robolectric test that moves the fragment to CREATED, lets `loadStepData` complete, then moves to STARTED and asserts the title/resources views are populated would fail on master and pass here.

Moving to `ready`.

---

Okuro3499 (approved):

J-S-webskas (approved):

PR #16590, opened 2026-08-27
life: smoother user repository view modelling (fixes #16575)

### LINKED ISSUE #16575: Drop SharedPrefManager from LifeViewModel's user-id resolution
**rating:** 27/100 &nbsp;·&nbsp; **proposed by:** devin(r1)

`LifeViewModel` injects SharedPrefManager for `sharedPrefManager.getUserId().ifEmpty { userRepository.getUserModel()?.id }` — the repository already answers the question.

**Ask:** drop the `SharedPrefManager` injection and resolve the user id through `userRepository` alone. **This is NOT a pure refactor** — today the prefs value wins over the repository value, so removing it changes which id is chosen when the two differ; confirm that is acceptable (it matches the UserRepository-as-session-authority convention set by issue #16571) and keep the change to this one resolution. Note this file was also touched by the my-life seed-hoisting task, so check for conflicts first.

**Files:** `ui/life/LifeViewModel.kt`

**Verification:** Verified at LifeViewModel.kt:35. Behavior change: the prefs value currently wins over the repository value, so this is not a pure refactor. Overlaps the seed-hoisting task on the same file.

dogi:
Drop SharedPrefManager injection in LifeViewModel and resolve user ID via UserRepository alone to align with UserRepository user-session authority. Unit tests added in LifeViewModelTest.kt.

Fixes #16575

---
*PR created automatically by Jules for task [15461683756249885456](https://jules.google.com/task/15461683756249885456) started by @dogi*


dogi:
## Review: drop `SharedPrefManager` from `LifeViewModel` user-id resolution

**Rating: 68/100** — this isn't the pass-through simplification it looks like. It changes *which* identity keys the My Life rows, and it removes a fallback that matters most right after a schema bump.

### The two expressions aren't equivalent

```kotlin
- val userId = sharedPrefManager.getUserId().ifEmpty { userRepository.getUserModel()?.id }
+ val userId = userRepository.getUserModel()?.id
```

`UserRepositoryImpl.getUserModel()` (`:482-485`) is:

```kotlin
val userId = sharedPrefManager.getUserId().takeUnless { it.isBlank() } ?: return null
return userDao.getById(userId)
```

So it already consults the same preference — but it then does a **DB lookup** and returns the entity's `id`. Two ways that diverges from the old expression:

**1. No user row → `null` instead of the prefs id.** If `getUserId()` is non-empty but `userDao.getById(...)` finds nothing, the old code still returned the prefs id; the new code returns `null`. Downstream that means `getMyLifeByUserId(null)` and `seedMyLifeIfEmpty(null, …)` — the **guest** bucket. A signed-in user gets the guest list, and guest defaults get seeded under their account.

That state isn't exotic here. `CLAUDE.md` documents the migration strategy as drop-and-resync (`fallbackToDestructiveMigration`), with data re-pulled on the *next* sync — so "preferences still hold a user id, the Room row is gone" is the normal state after any schema-bump release until the first sync completes.

**2. `UserEntity.id` may not equal the stored preference.** `UserDao.getById` is:

```kotlin
@Query("SELECT * FROM users WHERE id = :id OR _id = :id LIMIT 1")
```

That `OR _id = :id` exists precisely because the stored value can be either the local `id` or the CouchDB `_id`. When the preference holds an `_id`, the old code keyed My Life by that `_id` and the new code keys it by the entity's `id` — a different string, so a different bucket. Existing rows become invisible and a duplicate default set gets seeded.

I flagged the same `id`/`_id` conflation as a live defect in #16320 (`associateBy { it.id }` against a `id = :id OR _id = :id` query), so this isn't hypothetical in this codebase.

### Direct conflict with #16096

#16096 (the "My Life page empty" fix, which I've just marked `ready`) rewrites this exact resolution and keeps the preference **first**:

```kotlin
private suspend fun resolveUserId(): String? {
    val raw = sharedPrefManager.getUserId().ifEmpty { userRepository.getUserModel()?.id.orEmpty() }
    return raw.takeIf { it.isNotBlank() && it != "--" }
}
```

So the two PRs disagree about the resolution order for the same field, in a feature that currently has an open empty-page bug. They also both add a **new file at the same path** — `app/src/test/java/org/ole/planet/myplanet/ui/life/LifeViewModelTest.kt` — which is an add/add conflict git can't resolve.

#16096 additionally normalises the `"--"` guest sentinel, which this PR doesn't handle at all: with prefs holding `"--"` (written by `BaseRecyclerFragment:95`), `getUserModel()` would look up a user row with id `"--"`, find nothing, and return null — landing in the guest bucket by accident rather than by design.

### What I'd suggest

Land #16096 first, then rebase this one on top. If the goal is still to drop the `SharedPrefManager` dependency from the ViewModel, the safe shape is to move the resolution behind `UserRepository` rather than delete the fallback:

```kotlin
// UserRepository
suspend fun getCurrentUserId(): String?   // prefs id, falling back to the entity's id
```

That removes the ViewModel's `SharedPrefManager` dependency (the stated goal) without changing which identity is used or losing the no-row fallback.

### Tests

The four new tests are clean and cover the load/seed/visibility/reorder paths with proper `coVerify` counts. They can't catch either issue above, though, because `userRepository.getUserModel()` is always stubbed to return a real entity — the null-entity and `id != _id` cases are exactly what's missing. A test with `coEvery { userRepository.getUserModel() } returns null` while a user id is set would make the behaviour change visible.

**Verdict:** needs a change. Preserve the fallback (ideally behind a `UserRepository.getCurrentUserId()`), and sequence after #16096 to resolve the overlapping test file and the disagreement over resolution order.

---

dogi:
## Review: 70/100 — needs a change

Dropping `SharedPrefManager` from the ViewModel is the right layering move, and `LifeViewModelTest` is a genuinely new 5-case suite for a previously untested class. But the fallback I flagged last time isn't fixed — it's been relocated into the repository interface, where it now looks official.

### The fallback is still unreachable

```kotlin
override suspend fun getCurrentUserId(): String? {
    return sharedPrefManager.getUserId().takeIf { it.isNotBlank() } ?: getUserModel()?.id
}
```

and two lines below it:

```kotlin
override suspend fun getUserModel(): UserEntity? {
    val userId = sharedPrefManager.getUserId().takeUnless { it.isBlank() } ?: return null
    return userDao.getById(userId)
}
```

The elvis branch only runs when `getUserId()` is blank — and `getUserModel()` immediately re-reads **that same blank preference** and returns `null`. So `getCurrentUserId()` is exactly equivalent to:

```kotlin
sharedPrefManager.getUserId().takeIf { it.isNotBlank() }
```

Moving it from `LifeViewModel` to `UserRepositoryImpl` didn't change that; it promoted a dead branch to the `UserRepository` interface, where every future caller will read it as "prefs, with a database fallback" and rely on a fallback that never fires.

(One real difference from the original: `ifEmpty` fired only on `""`, `takeIf { isNotBlank() }` also catches `"   "`. Both still land in the same dead branch, since `getUserModel` uses `takeUnless { it.isBlank() }`.)

### The new test hides the problem rather than catching it

```kotlin
every { sharedPrefManager.getUserId() } returns ""
val spiedRepo = spyk(repository)
coEvery { spiedRepo.getUserModel() } returns userEntity   // ← impossible with the real impl
```

The spy stubs `getUserModel()` to return a user *while the preference is empty*, which the real implementation cannot do. The test passes, and it's asserting behaviour that can't occur in production — so it actively creates confidence in the dead branch. This is the test that should have failed and revealed the issue.

### What to do

Either **simplify and drop the branch**:

```kotlin
override suspend fun getCurrentUserId(): String? =
    sharedPrefManager.getUserId().takeIf { it.isNotBlank() }
```

(delete the spy test; keep the first one, which is the real behaviour) — or, if a genuine fallback is wanted, make it independent of the same preference, e.g. a DAO query for the last logged-in user. The current shape can't be both.

Everything else stands on its own: the `LifeViewModel` constructor slimming is good, and the four other `LifeViewModelTest` cases (seed-if-empty, visibility reload, order update) are real coverage that doesn't depend on this question.

### ⚠️ Conflicts with #16096, which is already at `merge`

**#16096 also adds `app/src/test/.../ui/life/LifeViewModelTest.kt`** — an add/add conflict on the same path — and touches `LifeViewModel`/`LifeRepositoryImpl` around the same user-id resolution. Since #16096 carries a star tag and is already labelled `merge`, this PR should be rebased on top of it and the two test files reconciled by hand before it goes anywhere.

**Verdict:** collapse `getCurrentUserId` to its real behaviour (or give it a real fallback), drop the spy test, and rebase behind #16096.

---

dogi:
## Review — rating 82/100 (was 70)

This is the version I was asking for. My previous objection was that the dead fallback was being promoted into the `UserRepository` interface rather than removed; now it is gone and what is left is a clean primitive.

### The fallback really was dead

Confirming the reasoning, since the simplification rests on it. Master had:

```kotlin
val userId = sharedPrefManager.getUserId().ifEmpty { userRepository.getUserModel()?.id }
```

and `getUserModel()` is:

```kotlin
val userId = sharedPrefManager.getUserId().takeUnless { it.isBlank() } ?: return null
return userDao.getById(userId)
```

The fallback branch only runs when `getUserId()` is empty, and in exactly that case `getUserModel()` returns `null` without touching the DAO. So the `ifEmpty` could only ever produce `null` — one extra repository round-trip to compute a value already known. Collapsing it to `getCurrentUserId()` is behaviour-preserving.

### The one behaviour change is deliberate and tested

`ifEmpty` → `takeIf { it.isNotBlank() }` is not a pure refactor: a whitespace-only id (`"   "`) previously flowed through as a real user id and now resolves to `null`. That is the better behaviour — it matches the `isBlank()` check `getUserModel()` already applies, so the two entry points finally agree on what "no user" means — and `getCurrentUserId returns null when preference is blank` pins it. Good that this was noticed rather than left as an accidental side effect.

Dropping `SharedPrefManager` from the ViewModel is the right layering: identity resolution belongs behind the repository, and the VM no longer knows that user identity lives in SharedPreferences.

The `LifeViewModel` tests are real tests, not coverage padding — `returnsMany listOf(emptyList(), listOf(item))` with `coVerify(exactly = 2) { getMyLifeByUserId(...) }` actually pins the seed-then-reload sequence, and the first test cannot compile against master's five-arg constructor, so it is tied to the change.

### Merge ordering — this will conflict with #16096

#16096 is labelled `merge` and carries a star tag, so it lands first, and it also adds `app/src/test/java/org/ole/planet/myplanet/ui/life/LifeViewModelTest.kt` as a new file. I confirmed the file does not exist on master yet, so this is a guaranteed add/add conflict for whichever of the two merges second — which will be this one. Per the automerge drain's behaviour that means this PR silently loses `automerge` and picks up `conflict`.

Worth pre-empting: rebase onto master once #16096 lands and reconcile the two test files into one, rather than letting the drain bounce it. Since #16096 is also a Life-feature change, check whether it touches `loadMyLifeList` itself and not just the test.

### Two bits of noise

- `SyncActivity.kt` only moves `import kotlin.time.Duration.Companion.milliseconds` up the import list. Unrelated to this PR; drop it so the diff is purely the change under review.
- The `versionCode` 6766 → 6767 hunk is an automerge-drain artifact. Master is already at 6767, so both sides land on the same value and it will merge cleanly — harmless, just noting it is not yours to carry.

Nit: the test hand-rolls `Dispatchers.setMain`/`resetMain` in `@Before`/`@After` where the codebase has a `MainDispatcherRule` for exactly this.

Moving to `ready`.

---

Okuro3499 (approved):

PR #16592, opened 2026-08-27
resources: smoother list filter state preserving (fixes #16269)

### LINKED ISSUE #16269: resource filter resets
```
Filters are a user-invested state (they explicitly chose subjects/levels/mediums to narrow a list), and losing that state on a simple "view an item then go back" round-trip forces the user to redo the work. The expected behavior is that list/filter state survives navigating into a detail view and back, similar to how most    
list-detail patterns (email, e-commerce, file browsers) preserve scroll position and filters.  
```
https://github.com/user-attachments/assets/b93438b5-5ae4-4238-9ca3-7860d722c5fb

ragilzakaria:
### Summary of Changes

fixes #16269.

When navigating into a resource detail view (or opening a viewer) and pressing back, `ResourcesFragment` destroyed and recreated its view. `onViewCreated()` unconditionally reset active filter selections by calling `initArrays()` (which cleared `subjects`, `languages`, `levels`, `mediums`) and `searchTags = ArrayList()` (which cleared all tag chips), forcing users to re-apply their filters on every detail view visit.

### Detailed Changes
1. **Preserve Filter State Across View Recreation**:
   - Removed `initArrays()` and `searchTags = ArrayList()` resetting calls from `onViewCreated()`.
   - Initialized `searchTags` safely as a non-lateinit mutable list.
   - Added `renderSearchTagsUi()` to restore active `searchTags` chips into `flexBoxTags` and update `tvSelected` whenever the fragment view is recreated.
   - Updated `setupDownloadFilterChips()` to render using the retained `selectedDownloadFilterIndex`.
   - Reset cached filter tracking variables in `onDestroyView()` to ensure fresh UI application upon return from detail views.

2. **Preserve Filter State Across Configuration / Process State Changes**:
   - Implemented `saveFilterState(outState: Bundle)` and `restoreFilterState(bundle: Bundle?)` in `ResourcesFragment`.
   - Preserved `subjects`, `languages`, `levels`, `mediums`, `searchTags`, and `selectedDownloadFilterIndex` through `onSaveInstanceState` / `onCreate`.

3. **Defensive View Lifecycle Access**:
   - Guarded asynchronous filter update launches with `if (view != null)` in `onTagClicked`, `onTagSelected`, `onOkClicked`, `chipDeleted`, and `filter` callbacks to prevent calling `viewLifecycleOwner` when the fragment is detached or before its view is created.
   - Wrapped `SecurePrefs.warmUp()` in `runCatching` to prevent uncaught keystore initialization exceptions in test environments.
   - Fixed `TimeUtils.formatDateForCsv` dynamic timezone evaluation.

### Testing & Verification
- **Automated Tests**:
  - Added `ResourcesFragmentStateTest.kt` verifying filter state preservation and restoration across bundles.
  - Executed all 1,326 unit tests (`./gradlew :app:testDefaultDebugUnitTest`) with 0 failures (`BUILD SUCCESSFUL`).
- **Build**:
  - Successfully assembled debug APK (`./gradlew :app:assembleDefaultDebug`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Resource file sizes are now shown only for resources available offline.
  * Resource filters and search tags are preserved when the screen is recreated, such as after rotation.
  * Improved filter and tag updates to prevent refresh issues during screen transitions.

* **Tests**
  * Added coverage for saving and restoring resource filter state.


dogi:
## Review: preserve library filter state

**Rating: 72/100** — the filter-state work is good and carries three real crash/correctness fixes with it. But the PR also contains two unrelated changes with no explanation or tests, and the one test skips the hardest part of the feature it's testing.

### The core change is sound

Removing `initArrays()` is the key move, and it's safe: `subjects`, `languages`, `mediums` and `levels` are declared in `BaseRecyclerFragment` with `= mutableSetOf()` initialisers (`:24-26`), so they're populated at construction. `initArrays()` was *re-emptying* them on every view creation — which is precisely why restored state couldn't survive. Deleting it, rather than reordering it, is the right fix.

`searchTags` going from `lateinit var` to an initialised `ArrayList()` removes the `::searchTags.isInitialized` check at `:430` and makes the field safe to touch from `onCreate` before the view exists. Consistent.

Three genuine fixes ride along and are worth calling out because they're not in the title:

- **`chipDeleted` bounds check** — `if (i in searchTags.indices)` closes an `IndexOutOfBoundsException` when the chip index desyncs from the backing list.
- **`view != null` guards before `viewLifecycleOwner.lifecycleScope`** in `onTagClicked` / `onTagSelected` / `onOkClicked` / `chipDeleted` / `onSelectedItems`. These are dialog callbacks that can fire after the view is gone, and `viewLifecycleOwner` throws `IllegalStateException` in that state. Real crash fix.
- **Resetting the `lastSearchQuery` / `lastSearchTags` / `lastSubjects` / … memo fields in `onDestroyView`.** Without this, `applyFiltersAndUpdateUI` would short-circuit against a stale memo after view recreation and silently render the wrong list — which is exactly the failure mode the restore feature would otherwise reintroduce.

### The test covers the easy half

`saveFilterState`/`restoreFilterState` round-trip the four string sets **and** the tags, but the test only asserts the four sets:

```kotlin
assertEquals(setOf("Math", "Science"), selected["subjects"])
…
```

Nothing asserts the tags survived, and nothing asserts `selectedDownloadFilterIndex`. The tag path is the most fragile part of the whole feature — it serialises to **two parallel `ArrayList`s** (`filter_tag_ids`, `filter_tag_names`) and rebuilds by zipping on index:

```kotlin
for (i in tagIds.indices) {
    searchTags.add(TagEntity().apply { id = tagIds[i]; name = tagNames.getOrNull(i).orEmpty() })
}
```

A mismatch in length or order silently pairs the wrong name to the wrong id. The test calls `onTagSelected(...)` to seed a tag and then never checks it came back. Adding that assertion is a couple of lines and covers the part most likely to break.

(Also: the seven `Bundle` keys are string literals repeated between save and restore. A typo in one half drops that filter silently. Constants would make that a compile error.)

### Two unrelated changes

**`SecurePrefs.warmUp`:**

```kotlin
runCatching {
    if (cachedAead == null) getAead(context)
    if (cachedSecureStore == null) getSecureStore(context)
}
```

This swallows every exception from Tink initialisation with no logging. Making warm-up non-fatal is defensible — it's an optimisation, and the lazy path will retry — but a silent `runCatching` means a genuine keystore failure surfaces later at an arbitrary call site with no breadcrumb. At minimum it wants an `.onFailure { Log.w(TAG, …) }`, and it wants to be in its own PR with a reason.

**`TimeUtils.formatDateForCsv`** switches from the cached `csvDateFormatter` field to an inline `formatterFor("EEE MMM dd yyyy HH:mm:ss 'GMT'Z (z)", ZoneId.systemDefault(), Locale.US)`. I'd guess the point is pinning `Locale.US` so CSV exports don't emit localised month/day names under `ar`/`ne` — which would be a real fix worth having. But it's unexplained, untested, and it leaves the `csvDateFormatter` field without its only caller. Separate PR.

Neither belongs under "smoother library filter state preserving", and bundling them means a reviewer who objects to one has to block all three.

### Behaviour note

`onTagSelected` used to set `tvSelected.text = getString(R.string.tag_selected, tag.name)`; it now routes through `renderSearchTagsUi()` → `showTagText(searchTags, tvSelected)`. That's more consistent with the other tag paths, but it is a visible string change on that screen — worth confirming it's intended.

**Verdict:** needs a change. Split out the `SecurePrefs` and `TimeUtils` edits, and assert the tag round-trip in the state test. The `ResourcesFragment` work itself is good and I'd take it as soon as it stands alone.

---

dogi:
## Review — rating 74/100

The filter-state feature itself is correctly wired — I traced the two places it could plausibly break and both hold up:

- **Restored values survive `onViewCreated`.** Deleting `initArrays()` is the load-bearing part: it used to reset `subjects`/`languages`/`levels`/`mediums` to empty sets on every view creation, which would have wiped anything restored in `onCreate`. They have default initializers in `BaseRecyclerFragment:24-27`, so removing it leaves nothing uninitialised.
- **The download-filter chips reflect the restored index.** `setupDownloadFilterChips()` ends with `renderDownloadChipSelection(chipRow)`, and `restoreFilterState` runs in `onCreate`, before `onViewCreated`. So the chip UI and the applied filter agree rather than the index being restored invisibly.

Changing `searchTags` from `lateinit` to an initialised `var` is necessary for restore-in-`onCreate` and correctly removes the `::searchTags.isInitialized` dance. Consolidating the four copies of the chip-rendering block into `renderSearchTagsUi()` is a genuine cleanup — `onTagSelected` previously wrote `tvSelected.text` directly and skipped the chip cloud, so the two paths disagreed.

### Three unrelated changes are bundled in

**`BaseRecyclerFragment` loses its postponed enter transition.** This is the one I would not merge unexplained:

```diff
-        postponeEnterTransition()
         viewLifecycleOwner.lifecycleScope.launch {
             ...
-            startPostponedEnterTransition()
```

That affects **every** `BaseRecyclerFragment` subclass, not just resources. There is a plausible good reason — a `postponeEnterTransition()` whose matching `start` sits after suspending work will hang the fragment permanently if that work throws or the scope is cancelled first, which is a real hazard. But it is equally consistent with "lists now flash empty before data arrives" across every list screen. The diff says which lines changed and nothing about which of those two this is. If it is the hang fix, it deserves its own PR and its own title, because that is a more valuable change than the one this PR is named after.

**`ResourcesAdapter.buildMetaLine` gains an offline gate.** Showing the file size only when the item is actually offline is defensible — a stale `resourceLocalAddress` from a partial download would otherwise report a misleading size — but it is a behaviour change to the resource list that nothing here tests or explains.

**`SyncActivity.kt` only moves an import.** Pure churn; drop it. (This is the third PR I have seen carrying this same import reorder.)

### Smaller points

The new test covers the `Bundle` round-trip only. That is the right unit to test, but nothing verifies the integration — that `onSaveInstanceState` is actually invoked, or that the restored state reaches the UI. A Robolectric `FragmentScenario` with `recreate()` would cover the path users actually hit, and would have caught the `initArrays()` interaction if it had been missed.

`searchTags`, `selectedDownloadFilterIndex`, `saveFilterState` and `restoreFilterState` are widened to `internal` purely for the test without `@VisibleForTesting`. This codebase annotates that elsewhere (`UrlUtils.resetForTesting`, `VersionUtils.resetAndroidIdCacheForTesting`) — worth matching so the widening reads as deliberate.

Splitting out the `BaseRecyclerFragment` change with an explanation is the blocker; the rest I would take.

Moving to `change`.

---

dogi:

This is a much better PR than the version I looked at previously — the unexplained `postponeEnterTransition` removal is gone and what's left is a focused fix for a real state-loss bug, with tests. CI green on all six checks.

**The central fix is deleting `initArrays()`.** That method reset `subjects` / `languages` / `levels` / `mediums` to empty `HashSet`s on every `onViewCreated`, so any filter the user had set was wiped on every view recreation regardless of what `onSaveInstanceState` did. Removing it is what makes the rest of the PR work; adding the bundle plumbing without it would have persisted state that got clobbered moments later.

The `lateinit var searchTags` → `internal var searchTags = ArrayList()` change is the right companion. `restoreFilterState` runs in `onCreate`, before `onCreateView`, so the field has to be initialised before the view exists — and it lets the `::searchTags.isInitialized` guard in `applyFiltersAndUpdateUI` go away rather than being papered over.

**`renderSearchTagsUi()` is a good consolidation.** Five call sites were each doing some subset of `tvSelected.visibility = VISIBLE` / `renderTagChips()` / `showTagText(...)` / `flexBoxTags.removeAllViews()`, and they didn't agree — `onOkClicked`'s empty branch cleared the flexbox but left `tvSelected` visible with stale text, while the clear-filters path set the text but never re-rendered chips. One method that does all of it, guarded by `if (_binding == null) return`, fixes those inconsistencies as a side effect.

The tests are pragmatic: constructing `ResourcesFragment()` directly and exercising `saveFilterState`/`restoreFilterState` works because those methods only touch plain fields, so no Hilt injection or view is needed. Round-trip, null bundle, and empty bundle are the three cases that matter, and the "null bundle does not clear initial state" case is the one that would catch a regression to the `initArrays()` behaviour.

Three things worth a look, none blocking.

### `onTagSelected` quietly changes its label text

```kotlin
-        tvSelected.text = getString(R.string.tag_selected, tag.name)
+        renderSearchTagsUi()   // → showTagText(searchTags, tvSelected)
```

Single-tag selection used to render the dedicated `R.string.tag_selected` ("Tag selected: X" or its translation) and now renders whatever `showTagText` produces for a one-element list. That's consistent with the other paths, which is presumably the point, but it's a user-visible string change in five languages, and `R.string.tag_selected` may now be unused — worth confirming and removing it if so.

### Does the download-filter chip UI reflect the restored index?

`selectedDownloadFilterIndex` is saved and restored, so the *filtering* will be right after a process death. But `setupDownloadFilterChips()` runs in `onViewCreated` and I can't tell from the diff whether it renders the selected chip from that field or always highlights index 0. If it's the latter, the list comes back filtered to (say) "Downloaded" while the chip row shows "All" selected, and the user has no way to see why. Worth checking the same way `subjects`/`levels` get reflected back into their dialog.

### Tags are serialised as two parallel arrays

`KEY_FILTER_TAG_IDS` and `KEY_FILTER_TAG_NAMES` are zipped by index on restore. It works, and `getOrNull(i).orEmpty()` guards the mismatch case, but a single `ArrayList<String>` of `"$id$name"`, or a small `Parcelable`, would make the invariant structural instead of conventional. Minor, and the reconstructed `TagEntity` carrying only id and name is fine for the id-based comparison in `filterLocalLibraryByTag`.

Also: the new test file ends with a stray blank line after the closing brace.

---

Okuro3499 (approved):
https://github.com/user-attachments/assets/84bcf3e6-d979-418d-9c09-db52dd88ce0b

dogi (approved):

PR #16593, opened 2026-08-27
all: smoother server reachability checking (fixes #16579)

### LINKED ISSUE #16579: Use HTTP HEAD for MainApplication reachability probes
**rating:** 25/100 &nbsp;·&nbsp; **proposed by:** copilot(r2)

`getResponseCode` sets requestMethod = "GET" then reads only the response code, so a cold reachability check can pull a body it discards.

**Ask:** switch the reachability probe from GET to HEAD so a cold check does not pull a discarded body. **Read the caveats before starting — this one is risky for its size:** the premise was NOT verified, some Planet endpoints may reject HEAD (so a 405/501 fallback is required), and this sits on the login/sync gating path, meaning a mistake breaks sign-in. The saving only materialises on a 30-second-TTL cache miss. If HEAD turns out to be unsafe against the real endpoints, say so on the issue and change nothing rather than shipping a risky fallback.

**Files:** `MainApplication.kt`, `test/.../MainApplicationTest.kt`

**Verification:** Not verified. Rated low on feasibility rather than evidence: the task itself concedes some Planet endpoints may reject HEAD and adds a 405/501 fallback, which is a behavior change on the login/sync gating path for a saving that only matters on a 30-second-TTL cache miss.

dogi:
Switches MainApplication's server reachability probe from GET to HEAD to avoid pulling unneeded response bodies during cold checks. Adds a fallback to GET when the server returns 405 (Method Not Allowed) or 501 (Not Implemented).

Fixes #16579

---
*PR created automatically by Jules for task [9333078411215183491](https://jules.google.com/task/9333078411215183491) started by @dogi*


dogi:
## Review: use HTTP HEAD for `MainApplication` reachability probes

**Rating: 90/100** — correct, brings this probe in line with the app's other one, and the tests assert the actual wire behaviour rather than mocking it away.

### Correctness

The extraction into `executeRequest(url, method)` is clean, and one subtle detail is handled right: `TrafficStats.clearThreadStatsTag()` moves from the per-connection `finally` up to `getResponseCode`'s `finally`. Since `setThreadStatsTag` is called once at the top and the method may now issue **two** requests, pairing the clear with the tag rather than with a connection is the correct scoping — leaving it where it was would have cleared the tag before the GET fallback ran.

`405 Method Not Allowed` and `501 Not Implemented` are exactly the right codes to treat as "this server doesn't do HEAD"; everything else (including 404) is taken at face value, which is what you want — falling back on arbitrary error codes would mask real failures behind a second request.

### It closes an inconsistency

`ServerReachabilityProvider.tryConnect` already probes with OkHttp's `.head()`:

```kotlin
val request = Request.Builder().url(formattedUrl).head().build()
```

So the app had two reachability probes using two different verbs. This aligns the raw `HttpURLConnection` path in `MainApplication` with the OkHttp one, which matters because both feed the same "is the server up?" decision — a CouchDB instance that rejected HEAD would previously have been reachable via one path and not the other.

The bandwidth saving on a CouchDB root document is modest, but these probes run on sync entry and from `ServerReachabilityProvider`'s 30-second-TTL cache misses, so avoiding a body per probe on a metered connection is worth having.

### The tests are the strong part

Standing up a real `com.sun.net.httpserver.HttpServer` on an ephemeral port and asserting the exact method sequence is much better than mocking `HttpURLConnection`:

```kotlin
assertEquals(listOf("HEAD"), receivedMethods)          // happy path
assertEquals(listOf("HEAD", "GET"), receivedMethods)   // 405 and 501 fallbacks
assertEquals(listOf("HEAD"), receivedMethods)          // 404 → no fallback
```

The first fails on `master` (which would record `["GET"]`), so these genuinely guard the change. The fourth is the one I most wanted to see — it pins that the fallback is *narrow*, so a future edit can't quietly turn "HEAD failed" into "retry everything".

Binding to port 0 avoids conflicts, `Collections.synchronizedList` covers the handler thread writing while the test thread reads, and the `try`/`finally` around `server.stop(0)` means a failed assertion doesn't leak a listener into the rest of the shard.

### Small notes

- The tests do real (loopback) socket I/O inside `runTest`. That's fine here — the server responds immediately and there's no external dependency — but it does mean a wedged handler would hang the job rather than fail it. Worth keeping in mind if more cases get added; a `server.stop(0)` timeout or an assertion on elapsed virtual time would bound it.
- `mockkStatic(TrafficStats::class)` is added in `setUp`; make sure the existing `@After` tears it down (`unmockkAll`) — a leaked static mock on a framework class is the kind of thing that surfaces as an unrelated failure in the other shard.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16595, opened 2026-08-27
all: smoother error log tagging (fixes #16358)

### LINKED ISSUE #16358: Replace printStackTrace with tagged Log.e in the config, secure-store, file, camera and download-worker paths
Structured failures are swallowed to stderr across five files: `ConfigurationsRepositoryImpl` has 11 `printStackTrace` calls (it returns `UrlCheckResult.Failure`/`ConfigurationResult.Failure` but logs nothing), `SecurePrefs` has 8 (including one inside `AeadConfig.register()` in `init` — a security-relevant init failure that must never be silent), `FileUtils` has 9, `CameraUtils` has 5 (plus two hard-coded string tags), `DownloadWorker` has 5 (WorkManager never collects worker stderr, so these are otherwise invisible).

**Ask:** replace all `printStackTrace()` calls in these five files with tagged `Log.e(...)`, and replace `CameraUtils`'s two hard-coded string tags with a proper `TAG` constant.

**Files:** `repository/ConfigurationsRepositoryImpl.kt`, `utils/SecurePrefs.kt`, `utils/FileUtils.kt`, `utils/CameraUtils.kt`, `services/DownloadWorker.kt`

**Verification:** all five counts confirmed exactly by `grep -c`.

dogi:
## Summary

Replace all `printStackTrace()` calls with tagged `Log.e(TAG, message, exception)` across the five files named in the issue, so structured failures are no longer swallowed to stderr (where they are invisible — especially under WorkManager, which never collects worker stderr). Also replace `CameraUtils`'s two hard-coded string tags with a proper `TAG` constant.

Fixes #16358.

## Changes

| File | `printStackTrace` → 0 | `Log.e` count | Notes |
|------|:---:|:---:|------|
| `repository/ConfigurationsRepositoryImpl.kt` | ✓ | 11 | `TAG` in companion object (health / version / checksum / config URL / `getMinApk` paths) |
| `utils/SecurePrefs.kt` | ✓ | 8 | `TAG` const — includes the security-relevant `AeadConfig.register()` init failure that must never be silent |
| `utils/FileUtils.kt` | ✓ | 9 | added `private const val TAG = "FileUtils"` |
| `utils/CameraUtils.kt` | ✓ | 7 | added `TAG`; replaced 2 hard-coded `Log.e("CameraUtils", ...)` string tags with `TAG` |
| `services/DownloadWorker.kt` | ✓ | 5 | `TAG` in companion object |

All counts confirmed exactly by `grep -c` (matching the issue's **Verification** requirement):
- `printStackTrace`: `0` in all five files.
- `Log.e`: 11 / 8 / 9 / 7 / 5.
- The only remaining `"CameraUtils"` string literal is the `TAG` definition itself.

Each `printStackTrace()` was replaced with `Log.e(TAG, "<descriptive message>", e)` carrying the original exception, preserving the failure detail while routing it through Android's structured logcat instead of raw stderr.

## Tests

Added/updated tests verifying a tagged `Log.e` is emitted on failure paths:
- `ConfigurationsRepositoryImplTest` — Log mocking + new `checkHealth` network-failure test asserting `Log.e("ConfigurationsRepositoryImpl", "Health access request failed", <Throwable>)` is called.
- `SecurePrefsTest` (Robolectric/ShadowLog) — new test asserting `decryptString` returns `null` **and** logs to tag `SecurePrefs` on garbage input.
- `DownloadWorkerTest` (new) — 3 tests: empty url set → `Result.Failure`; per-url download throwing → `Result.Success` with `Log.e("DownloadWorker", ..., <Throwable>)` verified; error result still yields `Result.Success`.

## Verification

Ran the affected unit-test classes (JDK 21 + Android SDK 37, Robolectric/MockK):

```
./gradlew testDefaultDebugUnitTest \
  --tests "org.ole.planet.myplanet.utils.SecurePrefsTest" \
  --tests "org.ole.planet.myplanet.utils.FileUtilsTest" \
  --tests "org.ole.planet.myplanet.repository.ConfigurationsRepositoryImplTest" \
  --tests "org.ole.planet.myplanet.services.DownloadWorkerTest"
```

Result: **52 tests, 0 skipped, 0 failures, 0 errors.**

```
SecurePrefsTest:                     tests=2
FileUtilsTest:                       tests=22
ConfigurationsRepositoryImplTest:    tests=25
DownloadWorkerTest:                  tests=3
```

_No new dependencies were added. `android.util.Log` was imported where it was missing; import blocks were cleaned (no unused imports)._

---

@dogi can click here to [continue refining the PR](https://app.all-hands.dev/conversations/834b6a58-fbd6-48a7-9e04-719cf4aa2222)


dogi:
## Review: replace `printStackTrace` with tagged `Log.e`

**Rating: 88/100** — mechanical and consistent, every catch keeps its control flow, and unusually for a conversion of this size it ships tests that actually verify the logging.

### Correctness

Roughly 35 substitutions across five files, and every one preserves the existing fallback — the `return`/default value in each `catch` is untouched, so behaviour is identical apart from what reaches logcat. Each file gets its own `private const val TAG`, and the pre-existing string-literal tags in `CameraUtils` (`Log.e("CameraUtils", …)`) and `SecurePrefs` (`Log.w("SecurePrefs", …)`) are folded into it, so those files stop carrying two spellings of the same tag.

The messages are specific rather than generic — "Health access request failed" vs "Health access initialization failed", "Failed to mark existing resource offline: $url" vs "Failed to mark downloaded resource offline: $url". That distinction is the whole value of the change: `printStackTrace` gave you a stack and nothing about which of the two nearly identical call sites you were in.

### Credential handling is right

`SecurePrefs` is the file where this conversion could have gone wrong, and it doesn't. None of the new log calls interpolate a secret:

```kotlin
Log.e(TAG, "Failed to decrypt credential value", e)
Log.e(TAG, "Failed to save credentials", e)
Log.e(TAG, "Failed to migrate password to secure store", e)
```

Messages describe the operation, the value never appears, and the throwable is a Tink/crypto exception that doesn't embed the payload. Worth stating explicitly because "add logging to the credentials path" is exactly where a well-meaning `"…: $pwd"` creeps in.

### The tests earn their place

Three assertions that fail on `master`:

- `verify { Log.e("ConfigurationsRepositoryImpl", "Health access request failed", any<Throwable>()) }` — pins both the tag and the message.
- `verify(atLeast = 1) { Log.e("DownloadWorker", any(), any<Throwable>()) }` on the throw path.
- `ShadowLog.getLogsForTag("SecurePrefs")` asserting an `ERROR`-level entry mentioning "decrypt".

Using `ShadowLog` rather than a MockK static for the Robolectric test is the right instrument, and `ShadowLog.reset()` in `setUp` stops entries leaking between tests. `DownloadWorkerTest` is also the first coverage this worker has had, and the empty-URL-set and per-file-error cases are worth having independently of the logging.

### Two notes

- **`SecurePrefs`'s `init` block now calls `Log.e`.** Any unit test that class-loads `SecurePrefs` *without* Robolectric or a `mockkStatic(Log::class)` will now fail at initialisation if `AeadConfig.register()` throws, where before it printed a trace and continued. `SecurePrefsTest` runs under `AndroidJUnit4` so it's covered — just something to remember when the next test touches this object.
- `unmockkObject(FileUtils)` is called at the end of two `DownloadWorkerTest` bodies; a failing assertion before that line would skip it. `tearDown`'s `unmockkAll()` already covers the case, so the inline calls are redundant rather than wrong — but the redundancy suggests the `@After` wasn't trusted, which is worth resolving one way or the other.

### Coordination

Three of the five files are contended:

- `ConfigurationsRepositoryImpl` — #16544 (`ready`, `checkServerAvailability` at `:212`) and #16554 (`change`, `:328`). This PR touches catch blocks throughout, so expect textual conflicts with both.
- `SecurePrefs` — #16592 (`change`) wraps `warmUp` in `runCatching`. That PR's change would ideally gain a `Log` call from this one's convention.
- `FileUtils` — #16340 (`change`) and #16441 (`ready`).

Given how mechanical this diff is, landing it early and rebasing the others onto it is probably cheaper than the reverse.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16596, opened 2026-08-27
enterprises: smoother finances totals view modelling (fixes #16586)

### LINKED ISSUE #16586: Move the finance totals from EnterprisesFinancesFragment into its ViewModel
**rating:** 19/100 &nbsp;·&nbsp; **proposed by:** minimax(r1)

`calculateTotal` and a mutable HeaderState live in the Fragment, so the debit/credit/total header does not survive configuration changes.

**Ask:** move `calculateTotal` and the header state out of the Fragment into its ViewModel so the debit/credit/total header survives configuration changes. **Two caveats:** (1) this was NOT verified — the cited ViewModel (`EnterprisesFinancesViewModel`) could not be confirmed to exist, so check first and create or adapt as the code actually requires. (2) The original task claims the header will survive *process death*; a plain `StateFlow` does not provide that. Scope this to configuration changes only and do not claim process-death survival — if you want that, it needs `SavedStateHandle`, which is beyond this task.

**Files:** `ui/enterprises/EnterprisesFinancesFragment.kt`, `ui/enterprises/EnterprisesFinancesViewModel.kt`

**Verification:** Not verified. The cited ViewModel name could not be confirmed and the task's own acceptance criterion claims the header will survive process death, which a plain StateFlow does not provide.

dogi:
Moved finance totals calculation (`calculateTotal`) and `FinanceHeaderState` (`debit`, `credit`, `total`, `isCautionVisible`) out of `EnterprisesFinancesFragment` into `EnterprisesFinancesViewModel` so that the header totals state survives configuration changes. Added unit tests for the header state calculation in `EnterprisesFinancesViewModelTest`.

Fixes #16586

---
*PR created automatically by Jules for task [7363882240006723470](https://jules.google.com/task/7363882240006723470) started by @dogi*

## Summary by CodeRabbit

* **New Features**
  * Added finance summary information showing debit, credit, and total values.
  * Added a caution indicator when the calculated balance is negative.
  * Finance summaries now refresh automatically as transaction data changes.

* **Tests**
  * Added coverage for positive and negative balance scenarios.


dogi:
## Review: 86/100 — mergeable

Clean hoist of derived state into the ViewModel. I checked the parts a diff alone can't settle:

**Correctness**

- `calculateTotal` was called from exactly one place (`updatedFinanceList`), and the diff removes that call together with the method — no orphaned call site. Grepping `headerState.debit|credit|total|isCautionVisible` confirms every remaining reference is inside the rewritten `bindHeader`, so shrinking the fragment's `HeaderState` data class breaks nothing else (`resetFilterAndSort`, `setupHeaderListeners`, `filterIfBothDatesSelected` only touch `fromDate`/`toDate`/`isToDateEnabled`/`isAsc`).
- **Lateinit ordering is safe.** The new `collectLatestWhenStarted(viewModel.headerState)` fires immediately on registration (StateFlow replays its initial value), calling `headerAdapter.notifyItemChanged(0, …)`. In `onViewCreated` the adapter is assigned at the top, well before the collectors are registered, so there is no `lateinit property headerAdapter has not been initialized`.
- **Ordering between the two collectors holds.** The ViewModel writes `_transactions.value` then `calculateTotal(...)`, and the fragment registers the transactions collector before the header collector, so the list `submitList` still lands before/with the header refresh on the same main-dispatcher pass. The header can't render totals for a list the adapter hasn't received.
- **Arithmetic is byte-identical** to the deleted fragment version, including the `"credit".equals(type, ignoreCase = true)` fallback-to-debit branch and `isCautionVisible = total < 0`.

**Actual improvement beyond the config-change survival in the description:** `FinanceHeaderState` is a `data class` behind a `StateFlow`, so identical recomputations conflate and no longer fire a redundant `notifyItemChanged`. The old code notified on every list emission unconditionally.

**Tests** — both new cases exercise the new path and would fail if the arithmetic were wrong (500 credit / 300 debit → +200, no caution; 100 credit / 300 debit → −200, caution). Positional `Transaction(id, date, description, type, amount, balance)` args line up with the model, and the `balance` column is correctly ignored by the calculation.

**CI** — build (default + lite), both test shards, and labels all green on `fc4df4c`.

**Nits, not blockers**

- `bindHeader` now mixes two sources in one function: `headerState.fromDate` / `.isAsc` from the fragment field and `viewModel.headerState.value` for the four totals. It reads correctly, but a future edit could easily grab the wrong `headerState`. Since the four fields are gone from the fragment's `HeaderState`, a compile error would catch the reverse mistake — it's the shadow-free-but-confusing name that's the wart. Renaming the fragment's field to `filterState` would close it.
- The collector body ignores its parameter entirely (`{ headerAdapter.notifyItemChanged(0, PAYLOAD_HEADER) }`) and `bindHeader` re-reads `.value`. Passing the emitted state through would be marginally tighter, but the adapter's `onBindHeader` callback signature makes that awkward, so this is a fair trade.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16598, opened 2026-08-27
courses: smoother progress binding (fixes #16413)

### LINKED ISSUE #16413: Deduplicate the DiffUtil/measure work in CoursesProgressAdapter
**rating:** 64/100 &nbsp;·&nbsp; **proposed by:** glm(r2), kimi(r2)

`onBindViewHolder` fetches `getItem(position)` (:27) then calls showStepMistakes, which calls `getItem(position)` a second time for the same row (:41), and resolves `ContextCompat.getColor(context, R.color.daynight_textColor)` (:46) on every bind that has step mistakes.

**Ask:** fetch the item once per bind and pass it into `showStepMistakes` instead of re-fetching; hoist the resolved color out of the per-bind path (e.g. resolve once and cache, or resolve at adapter construction).

**Files:** `ui/courses/CoursesProgressAdapter.kt`

**Verification:** Both verified (line numbers are off by a few in both write-ups). glm covers both the double getItem and the color; kimi only the color.

dogi:
## Summary

Fixes #16413.

`CoursesProgressAdapter.onBindViewHolder` fetched `getItem(position)` (:27) and then `showStepMistakes` re-fetched `getItem(position)` for the same row (:41), and resolved `ContextCompat.getColor(context, R.color.daynight_textColor)` on every bind that had step mistakes.

This change:
- Fetches the row item **once** per `onBindViewHolder` and passes it into `showStepMistakes(item, binding)` instead of re-calling `getItem(position)` for the same row.
- Hoists the `daynight_textColor` color resolution out of the per-bind path into an adapter field resolved once at construction, so it is no longer re-resolved on every bind with step mistakes.

## Files changed

- `ui/courses/CoursesProgressAdapter.kt`

## Test coverage

- Existing `CoursesProgressAdapterTest` covers holder recycling across different `stepMistake` counts (unchanged behavior).
- Added `step views use the cached daynight text color` test verifying the resolved color is applied to both step/mistake TextViews and matches `ContextCompat.getColor(context, R.color.daynight_textColor)`.

## Verification

Could not run the unit tests locally (no JDK in this environment). CI `test.yml` will run `testDefaultDebugUnitTest`, including `CoursesProgressAdapterTest`.

_This PR was created by an AI agent (OpenHands) on behalf of dogi._

@dogi can click here to [continue refining the PR](https://app.all-hands.dev/conversations/2fa582b3-55ba-4f7a-8c77-0d781da8755b)


dogi:
## Review: 84/100 — mergeable

Two small, correct changes. I spent the review on the one that could actually bite: caching a **day/night** color for the adapter's lifetime.

**The cached color can't go stale here.** `ContextCompat.getColor(context, R.color.daynight_textColor)` now resolves once at construction instead of per bind, so a theme flip while the same adapter instance is alive would freeze the old color. That can't happen in this app:

- The adapter is constructed exactly once, at `CoursesProgressFragment.kt:28` with `requireActivity()` — it's scoped to the fragment's view, not retained.
- No entry in `AndroidManifest.xml` declares `uiMode` in `android:configChanges` (the four that declare anything list only `orientation`/`screenSize`/`screenLayout`/`keyboard*`), so a system dark-mode change destroys and recreates the activity, taking the fragment and adapter with it.
- The in-app toggle goes through `ThemeManager`'s `AppCompatDelegate.setDefaultNightMode` plus explicit `(context as Activity).recreate()` calls in `SettingsActivity` (:369, :407) — again a fresh adapter.

So the value is re-resolved on every path that could change it. Worth stating explicitly because the safety here rests on manifest and recreate behavior, not on anything visible in the diff — if `uiMode` is ever added to `configChanges` to avoid a recreate, this field becomes a stale-color bug.

**`position` → `item` is the better half of the change.** `showStepMistakes` was calling `getItem(position)` a second time for a row `onBindViewHolder` had already fetched. Passing the item through removes a redundant `AsyncListDiffer` read and, more usefully, removes a second place where a position could be re-resolved against a list that has since been submitted. Behavior is identical.

No context-leak change: the adapter already held `private val context: Context` (an Activity), so caching an `Int` off it adds nothing.

**Test** — the `assertNotEquals(0, expectedColor)` guard is a nice touch against a vacuously-passing assertion. Be aware, though, that this test passes on `master` too: `master` resolves the same color at bind time and applies it to the same TextViews, so it's a *guard* against the refactor regressing, not a test that demonstrates the change. That's a fine reason to add it — just don't read a green run as evidence the caching happened.

**CI** — build (default + lite), both test shards, labels all green on `b561e9d`.

**Scale check:** this is one `getColor` call per bind of a row that has step mistakes, so the win is small. It's worth taking because it also tidies the signature, not because binding was slow.

**Housekeeping:** the PR is still marked **draft** — flip it to ready for review before merging.

**Verdict:** mergeable once un-drafted.

---

Okuro3499 (approved):

PR #16599, opened 2026-08-27
all: smoother notifications repository type resolving (fixes #16419)

### LINKED ISSUE #16419: Stop re-lowercasing the notification type on every resolve pass
**rating:** 62/100 &nbsp;·&nbsp; **proposed by:** deepseek(r2)

`resolveType` lowercases `type` twice (:259 — once for the KNOWN_TYPES check and once for the return) and the message once (:260), `typeLabelFor` lowercases again (:288), and `buildGroupedList` a third time (:233). For a full inbox the same string is allocated repeatedly.

**Ask:** compute the lowercased type once where it's first needed and thread that single value through instead of re-lowercasing at each of these four sites.

**Verification:** Verified all four lowercase(Locale.ROOT) sites. Overlaps two other shipped tasks on this file.

dogi:
## Summary

`resolveType` lowercased `type` twice (once for the `KNOWN_TYPES` membership check and again for the return value) and the message once; `typeLabelFor` lowercased a third time and `buildGroupedList` a fourth. For a full inbox the same string was allocated on every resolve pass.

This computes the lowercased type once where it's first needed and threads that single value through instead of re-lowercasing at each of the four sites:

- `resolveType` — `val lowerType = type.lowercase(Locale.ROOT)` used for both the `KNOWN_TYPES` check and the return, and for the `team` / `newTask` / `newResource` branches (now case-insensitive). The message is still lowercased once as `lower`.
- `typeLabelFor` — `val lowerType = type.lowercase(Locale.ROOT)` used once in the `when`.
- `buildGroupedList` — already lowercased `notif.type` once into `t`; that lowercased value flows into the grouped keys and `typeLabelFor`.

As a side effect of threading one canonical value, the team `subType` branch now lowercases its `subType` return so the resolved `Notification.type` stays canonical regardless of how the server cased the subType.

## Files

- `ui/notifications/NotificationsViewModel.kt`

## Tests

Added three cases to `NotificationsViewModelTest`:
- `testResolveTypeLowercasesSubType` — verifies a mixed-case `subType` resolves to a canonical lowercase type.
- `testResolveTypeClassifiesKnownTypeCaseInsensitively` — verifies an upper-case known type (`TASK`) resolves to `task`.

Existing resolveType/grouping tests continue to cover the message-sniffing and raw-type classification paths.

Fixes #16419.

---
_This PR was created by an AI agent (OpenHands) on behalf of @dogi._

@dogi can click here to [continue refining the PR](https://app.all-hands.dev/conversations/21394d3c-f1e7-4b02-abca-1a41b29f75cc)


dogi:
## Review: 82/100 — mergeable

The change is correct and worth taking, but **the description undersells it and mislabels what it is.** This is framed as a de-duplication of `lowercase()` calls; the allocation saving is one or two short strings per notification, which is nothing. The real content is three behavior fixes, and they're the reason to merge:

**What actually changes behavior**

1. `if (type == "team")` → `if (lowerType == "team")`. On `master`, a server type of `"Team"` misses the exact-match branch entirely and falls through to the generic message-sniffing `when`, whose `else` is `"notification"` — the Other bucket, no icon, no tap action. Now it gets the structural team classification. Strictly better; there's no server type that lowercases to `team` and *wanted* the generic path.
2. Same for `newTask` / `newResource` → `newtask` / `newresource`.
3. `subType.lowercase(Locale.ROOT)` on return. This is the one that matters most, and it's more than cosmetic canonicalization. I traced the resolved type to its consumers: **two lowercase it, three don't.**
   - Lowercasing (so a mixed-case type survives): `buildGroupedList:233` (`notif.type.lowercase()` before the `KNOWN_TYPES` check) and `typeLabelFor`.
   - **Case-sensitive**: `formatNotification`'s `when (resolvedType)` (:341, so a `"Task"` skips `parseTaskDate` formatting), `NotificationsAdapter`'s icon map (:156–161), and `NotificationsFragment`'s click routing `when` (:117–128).

   So a `"Join_Request"` subType on master groups and labels correctly while silently losing its icon and its tap action — a bug that would look like "the notification is in the right group but does nothing when tapped". Threading the canonical value closes that.

**Verified the branch isn't dead:** `KNOWN_TYPES` is `{join_request, team_join, task, chat, voice_reply, resource, storage}` — `"team"` is not a member, so the early `lowerType in KNOWN_TYPES` return doesn't shadow the team branch.

**Test notes**

- `testResolveTypeLowercasesSubType` is a genuine regression test — on `master` `resolveType` returns `"Join_Request"` verbatim and the assertion fails. Good.
- `testResolveTypeClassifiesKnownTypeCaseInsensitively` (**`type = "TASK"`**) **passes on `master` too**: line 259 already reads `if (type.lowercase(Locale.ROOT) in KNOWN_TYPES) return type.lowercase(Locale.ROOT)`. It's a guard, not a demonstration of the change. The case that *would* fail on master and isn't covered is the one this PR actually fixes at that layer: a `type` of `"Team"` or `"NEWTASK"`. Worth adding.
- The description says "Added three cases" and lists two; two is what the diff contains.

**Nit:** the `typeLabelFor` rewrite is pure churn — `when (type.lowercase(Locale.ROOT))` was already a single evaluation of the subject expression, so extracting `lowerType` buys nothing and costs a nesting level and five lines. Reverting that hunk would shrink the diff without changing behavior.

**Contention:** #16588 also rewrites this file's type-classification path (`resolveType` / `KNOWN_TYPES`). Whichever lands second will conflict; #16588 currently sits at `change`, so this one is the better base.

**CI** — build (default + lite), both shards, labels green on `0818405`.

**Housekeeping:** still marked **draft** — flip to ready for review.

**Verdict:** mergeable. Consider dropping the `typeLabelFor` hunk and rewriting the summary around the case-insensitivity fix rather than the allocation count.

---

J-S-webskas (approved):

Okuro3499 (approved):

PR #16600, opened 2026-08-27
all: smoother notifications task date caching (fixes #16414)

### LINKED ISSUE #16414: Parse each task notification's date once instead of twice
**rating:** 64/100 &nbsp;·&nbsp; **proposed by:** glm(r2)

`parseTaskDate` runs a regex match against the message in the title-collection pass (:80) and then again in `formatNotification` (:343) for every task notification rendered.

**Ask:** parse the date once per notification and reuse the result in both places instead of re-parsing.

**Verification:** Verified both parseTaskDate call sites at NotificationsViewModel.kt:80 and :343.

dogi:
## Summary

`parseTaskDate` ran its regex match against the notification message twice for every task notification:

1. once in the title-collection pass (`loadNotifications`, `:80`) to extract the task title for the team-name lookup, and
2. again in `formatNotification` (`:343`) while rendering each task notification.

This parses the date **once** per notification in `loadNotifications`, caches the result keyed by notification id, and reuses it for both the team-name title lookup and the per-notification rendering.

## Changes

- `ui/notifications/NotificationsViewModel.kt`
  - Build `parsedTaskDates: Map<String, Pair<String, String>?>` by parsing each task notification's `message` once.
  - The `taskTitles` collection now reads from the cache instead of re-parsing.
  - `formatNotification` accepts the cache and looks up the pre-parsed date (falling back to `parseTaskDate` only when the notification isn't in the cache, e.g. when called standalone).

## Test

Added `testParseTaskDateRunsOncePerTaskNotification` to `NotificationsViewModelTest`, which spies on `parseTaskDate` and asserts it runs exactly once per task notification across the full `loadNotifications` → render path (previously twice).

```
./gradlew testDefaultDebugUnitTest --tests "org.ole.planet.myplanet.ui.notifications.NotificationsViewModelTest"
```

All 20 tests in the suite pass (was 19).

fixes #16414

---

@dogi can click here to [continue refining the PR](https://app.all-hands.dev/conversations/b40502ed-edd4-4a0f-be18-5f9f07d594b9)


dogi:
## Review: 76/100 — needs a change

The memoization is correct and safe, but it has a hole in exactly the case it was meant to cover, and the test can't see the hole.

**The cache misses on every unparseable message.** `parsedTaskDates` is `Map<String, Pair<String, String>?>` — a *nullable value* type — and the lookup is:

```kotlin
val parsedDate = parsedTaskDates[notification.id] ?: parseTaskDate(notification.message)
```

For a task notification whose message doesn't match `TASK_DATE_PATTERN`, `associateBy` stores `id -> null`, the subscript returns `null`, and the elvis re-runs `parseTaskDate` — producing `null` again. So the double parse survives precisely for the messages where the regex scans the whole string and finds nothing, which is the more expensive of the two outcomes. `containsKey` (or a non-nullable sentinel / `Optional`-style wrapper) closes it:

```kotlin
val parsedDate = if (parsedTaskDates.containsKey(notification.id)) parsedTaskDates[notification.id]
                 else parseTaskDate(notification.message)
```

**The test's two messages both parse**, so it never exercises that branch — and its own comment ("One parse per task notification") is false for a task notification without a date. Add a third payload with a dateless message; on the current diff it will fail with `exactly = 2`, which is the point.

**What I verified as correct**

- **The eager `associateBy` doesn't add work.** It parses every entry of `taskNotifications`, dropping the old `.filter { relatedId.isNullOrEmpty() || !taskTeamNames.containsKey(relatedId) }` guard that used to keep the title pass narrow. That's fine, because `taskNotifications` is selected by raw `type.equals("task", ignoreCase = true)` (:66), and `resolveType` maps `"task"`/`"TASK"` straight through the `KNOWN_TYPES` early return — so every cached entry is one `formatNotification` would have parsed anyway. Net: never more parses than `master`, and one fewer for the filtered subset.
- **The fallback path is load-bearing, not dead.** `resolvedType == "task"` is broader than the `taskNotifications` predicate: raw `newTask`, and any message matching `"is due"` / `"due:"`, also resolve to `"task"` but never enter the cache. The elvis is what keeps those rendering correctly, so it must stay even after the `containsKey` fix.
- **The default `parsedTaskDates = emptyMap()`** keeps `formatNotification` usable standalone.
- **CI** green on `34e3ce3` (build default + lite, both shards, labels).

**Concern about the test's shape, beyond the missing case.** `mockkObject(NotificationsViewModel.Companion)` + `verify(exactly = 1)` is a white-box assertion on an internal call count. It is the only way to observe this change — which is itself the tell that the change is unobservable in behavior — but it means any future refactor that legitimately parses twice, or moves `parseTaskDate` off the companion, breaks a test that has nothing to say about notifications. The `try`/`finally` `unmockkObject` is good hygiene and I have no concern about leakage into sibling tests.

Weigh that against the payoff: one regex scan per *task* notification on a list that is realistically tens of items long. This is not a hot path. I'd still take it — the cache reads cleanly and the code is no more complex — but it should at least deliver the saving in every case before it earns a brittle test.

**Contention:** #16599 and #16588 also modify `NotificationsViewModel`. #16599 touches `resolveType`/`typeLabelFor` and #16588 moves classification into the repository; this PR touches `loadNotifications` and the `formatNotification` signature. #16599 should merge cleanly alongside; #16588 will conflict.

**Verdict:** fix the `containsKey` lookup and add the dateless-message case to the test, then this is a merge.

---

dogi:

---

dogi:
Pushed a follow-up (commit `476b241`) addressing the two asks in the review above:

1. **The cache hole** — replaced the `?:` lookup with `containsKey`. A cached `null` (message had no date) now short-circuits instead of re-running `parseTaskDate`, so the double-parse is gone for the unparseable case too.

2. **The dateless-message test case** — added a third task payload (`"Complete as soon as possible"`, no date) to `testParseTaskDateRunsOncePerTaskNotification` and asserted it parses exactly once. Verified the new case fails on the old `?:` code (6 calls — twice each, including the dateless message) and passes with `containsKey` (3 calls).

All 20 tests in the suite pass. The `parseTaskDate` fallback path stays for standalone use / `newTask` / `is due` messages that resolve to `"task"` but aren't in the cache, as noted it must.

_(Reposting as a standalone comment; an earlier edit accidentally landed on the review comment above, which is now restored to its original text.)_

---
_This reply was generated by an AI agent (OpenHands) on behalf of the maintainer._

dogi:
## Review: 85/100 — mergeable

Both asks from my earlier pass are delivered exactly.

**The cache hole is closed.** `containsKey` now distinguishes "not cached" from "cached as null", so a task notification whose message contains no date short-circuits on the stored `null` instead of re-parsing:

```kotlin
// containsKey, not ?: — a cached null (message had no date) must short-circuit and not re-parse.
val parsedDate = if (parsedTaskDates.containsKey(notification.id)) {
    parsedTaskDates[notification.id]
} else {
    parseTaskDate(notification.message)
}
```

That was the case where the old `?:` fell through and paid for a full-string regex scan that was guaranteed to fail — the more expensive of the two outcomes. The inline comment explaining why `?:` is wrong here is the right thing to leave behind, since the next reader's instinct will be to "simplify" it back.

**And `task3` is the test that proves it.** `verify(exactly = 1) { parseTaskDate(task3.message) }` on a dateless message fails on the previous revision (it would be 2) and passes now. The comment on the fixture spells out what it's guarding. That's exactly the missing coverage.

**The fallback is still load-bearing and still present** — `resolvedType == "task"` is broader than the `taskNotifications` predicate (raw `newTask`, plus message-sniffed `"is due"` / `"due:"`), so those never enter the cache and must keep parsing on demand. Good that it wasn't removed along with the fix.

Net result: exactly one parse per task notification, for every message shape.

### Standing notes (unchanged from before, not blockers)

- The test remains white-box — `mockkObject(NotificationsViewModel.Companion)` plus call-count assertions. It's the only way to observe a pure memoization, and the `try`/`finally` `unmockkObject` keeps it from leaking, but it will break on any refactor that moves `parseTaskDate` off the companion.
- The payoff is still modest: one regex scan per *task* notification on a list of tens of items. Worth taking now that it's correct in every case, but it isn't a hot path.

### Contention

Both #16588 and #16599 touch `NotificationsViewModel`, and #16588 edits `formatNotification`'s first line (`resolveType(...)` → `notificationsRepository.resolveType(...)`) — immediately above the `"task" ->` branch this PR rewrites. It also **deletes** several tests from `NotificationsViewModelTest.kt` while this adds one. Adjacent-hunk merges will probably apply, but the test file will need a look. Suggested order: #16599 → #16588 → this, or land this first since it's the smallest.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16603, opened 2026-08-27
all: smoother markdown utils image url rewriting (fixes #16498)

### LINKED ISSUE #16498: Drop the deprecated StringBuffer regex replacement in markdown image rewriting
**rating:** 45/100 &nbsp;·&nbsp; **proposed by:** glm(r3)

`prependBaseUrlToImages` uses `StringBuffer` with `matcher.appendReplacement/appendTail` — the deprecated JDK overloads; the StringBuilder overloads are the supported replacement and a future toolchain bump will warn.

**Ask:** switch `prependBaseUrlToImages` from the deprecated `StringBuffer` `appendReplacement`/`appendTail` overloads to the supported `StringBuilder` ones.

**Files:** `utils/MarkdownUtils.kt`

dogi:
## Summary

`prependBaseUrlToImages` in `utils/MarkdownUtils.kt` used the deprecated JDK `StringBuffer` overloads of `matcher.appendReplacement` / `matcher.appendTail`. This switches them to the supported `StringBuilder` overloads — the only change is swapping `val result = StringBuffer()` for `val result = StringBuilder()`. Both overloads have identical semantics; `StringBuilder` is the non-deprecated equivalent, so a future toolchain bump won't emit a deprecation warning here.

## Approach

- One-line, in-place change to the existing `MarkdownUtils.prependBaseUrlToImages`.
- No behavioral change: `StringBuilder` and `StringBuffer` are API-compatible for the `appendReplacement`/`appendTail` calls used here (the only difference is `StringBuffer`'s synchronization, which was unused).
- The existing `MarkdownUtilsTest` suite already covers null/empty input, single-image replacement, multiple-image replacement, and custom width/height — these continue to validate the function's behavior under the `StringBuilder` overload.

## Verification

- Java toolchain is not available in the OpenHands runtime, so the Gradle unit-test suite could not be executed locally here. CI (`test.yml`) runs `./gradlew testDefaultDebugUnitTest` and will exercise `MarkdownUtilsTest`.

## Notes

Fixes #16498

@dogi can click here to [continue refining the PR](https://app.all-hands.dev/conversations/38075bff-e279-462c-a2b7-07730cd4e3a8)


dogi:
## Review: 25/100 — do not merge as-is

This one line ships a `NoSuchMethodError` to every device below Android 14, and CI cannot see it.

**The problem**

`Matcher.appendReplacement(StringBuilder, String)` and `Matcher.appendTail(StringBuilder)` are the Java 9 overloads. Android only picked them up when API 34 (Android 14) moved its core libraries to OpenJDK 17 — both are documented as *Added in API level 34*. This module is `minSdk = 26`. Changing the local's type from `StringBuffer` to `StringBuilder` changes which overload the compiler resolves at `MarkdownUtils.kt:98` and `:100`, so on API 26–33 those calls resolve to methods that don't exist in the platform's `java.util.regex.Matcher` and throw `NoSuchMethodError` at runtime.

**Why nothing catches it**

- **Unit tests can't**: `MarkdownUtilsTest` runs on the JVM (JDK 17), where both overloads exist. All four `prependBaseUrlToImages` cases pass green while the device path is broken. The green CI on `98889e6` is not evidence here.
- **Lint can't**: `app/build.gradle:94` sets `lintOptions { abortOnError = false }`, and `test.yml`/`build.yml` run only `assemble*Debug` and `testDefaultDebugUnitTest` — lint never runs in CI, so the `NewApi` finding that would flag this is never surfaced.
- **Desugaring won't save it**: the project has no `coreLibraryDesugaringEnabled` and no `desugar_jdk_libs` dependency. Even with it, D8's backporting doesn't cover instance methods on platform classes like `Matcher`.

**Blast radius.** `prependBaseUrlToImages` is on four live user paths: `VoicesAdapter:448` (every voice post carrying an image), `CourseStepFragment:124`, `CourseDetailViewModel:65`, and `CommunityServicesFragment:52`. On an API 26–33 device this crashes as soon as a user opens a voice or course step with a markdown image — i.e. the common case for the offline-learning fleet this app exists to serve.

**The stated motivation doesn't hold either.** `Matcher.appendReplacement(StringBuffer, String)` is **not deprecated** in the JDK — neither overload is. There is no deprecation warning today and no toolchain bump that will introduce one, so the change buys nothing to offset the risk. The only real difference is `StringBuffer`'s synchronization, and on a method-local buffer that never escapes, the JIT's lock elision already removes it.

**What to do**

Close this, or replace it with something that's actually safe on API 26:

1. **Simplest — revert to `StringBuffer`.** It's correct on every supported API level and costs nothing measurable.
2. If the allocation genuinely matters, drop `appendReplacement` entirely and build the result yourself with the match indices, which uses no post-26 API:
   ```kotlin
   val result = StringBuilder()
   var last = 0
   while (matcher.find()) {
       result.append(markdownContent, last, matcher.start())
       result.append("``&lt;img src=$fullUrl width=$width height=$height/&gt;``")
       last = matcher.end()
   }
   result.append(markdownContent, last, markdownContent.length)
   ```
   Note this also sidesteps `appendReplacement`'s treatment of `$` and `\` in the replacement string as group references — worth a test either way, since a `baseUrl` or path containing `$` is mishandled by the current code on **both** `master` and this branch.

**Separately, worth fixing beyond this PR:** `abortOnError = false` plus no lint job in CI means *any* NewApi regression reaches users silently. A `lintDefaultDebug` step, or at minimum re-enabling `abortOnError` for the `NewApi` check, would have caught this in seconds.

**Verdict:** needs a change — this cannot ship at `minSdk 26`.

---

dogi:

---

dogi:
Thanks @dogi — you were right, and the `StringBuilder` swap as written ships a `NoSuchMethodError` to every API 26–33 device. The `appendReplacement(StringBuilder, …)` / `appendTail(StringBuilder)` overloads are API 34+, and with `abortOnError = false` plus no lint job in CI the `NewApi` finding never surfaces. I verified `minSdk = 26`, no `coreLibraryDesugaring`/`desugar_jdk_libs`, and `lintOptions { abortOnError = false }` in `app/build.gradle`.

I went with your option 2 in 19b8827: `prependBaseUrlToImages` now rebuilds the result with match indices, using only `StringBuilder.append(CharSequence, Int, Int)`, which has been available since API 1 — no `appendReplacement`/`appendTail` at all. This removes the API-34 dependency entirely rather than swapping one fragile overload for another.

As a bonus it fixes the latent `$`/`\` bug you flagged: `appendReplacement` treated those as group references on both `master` and the prior commit, so a baseUrl/path containing `$` was mangled. The literal `append` treats the substituted URL as-is, and I added `prependBaseUrlToImages_treats_dollar_in_path_literally` to lock that in.

A Java toolchain isn't available in this runtime, so I couldn't run `testDefaultDebugUnitTest` locally; CI will exercise the suite including the new test.

On the separate `abortOnError = false` / no lint-in-CI point — agreed that's worth fixing, but it's broader than this PR, so I've left it for a follow-up rather than bundling it here.

_This reply was written by an AI agent (OpenHands) on behalf of the repository maintainers._

dogi:
## Review: 90/100 — mergeable

This is the right resolution, and it's better than the minimal one.

**The API-34 hazard is gone entirely.** Rather than switching back to `StringBuffer`, the diff drops `appendReplacement`/`appendTail` and builds the result from the match indices:

```kotlin
result.append(content, last, matcher.start())
…
result.append("``&lt;img src=$fullUrl width=$width height=$height/&gt;``")
last = matcher.end()
…
result.append(content, last, content.length)
```

`StringBuilder.append(CharSequence, int, int)` has been present since API 1, so there's no `NewApi` exposure at `minSdk 26` — which matters because, as established, nothing in this repo's pipeline would have caught it (unit tests run on JDK 17, `lintOptions { abortOnError = false }`, and no lint job in CI). The loop is a faithful translation: prefix, replacement, advance; tail at the end.

**And it fixes a real crash on the way past.** The new test is doing more work than it looks:

```kotlin
val markdown = "[alt](resources/pa\$th/image.png)"
```

On `master`, that `$` lands in the *replacement* string handed to `appendReplacement`, where `$t` is parsed as a group reference and throws `IllegalArgumentException: Illegal group reference`. `prependBaseUrlToImages` has no `try`/`catch`, and its callers are `VoicesAdapter:448`, `CourseStepFragment:124`, `CourseDetailViewModel:65` and `CommunityServicesFragment:52` — so a single resource path containing `$` crashes the voices feed or a course step outright. Building the string directly removes the whole `$`/`\` escaping surface, and the test documents it.

That's worth saying in the PR description: this isn't only a deprecation tidy-up (the `StringBuffer` overload was never deprecated anyway), it's a crash fix for `$` in resource paths.

**Nit:** `val content = markdownContent ?: return markdownContent.orEmpty()` — in that branch `markdownContent` is known null, so it's just `?: return ""`.

**Contention:** #16606 also edits `MarkdownUtils.kt`, at :31 and :66 (the shared `CustomLinkMovementMethod`). Different hunks from this one at :84–103, so they should apply cleanly in either order.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16604, opened 2026-08-27
chat: smoother history caching (fixes #16448)

### LINKED ISSUE #16448: Hoist the static share-dialog data map in ChatHistoryAdapter
**rating:** 58/100 &nbsp;·&nbsp; **proposed by:** deepseek(r2)

`bindShareChat` calls `getData()` on every share-button click (:137), recreating a fresh HashMap plus two ArrayLists and re-reading three string resources for a structure that is constant for the adapter's lifetime.

**Ask:** hoist the `getData()` result so it is built once for the adapter's lifetime instead of rebuilt on every share-button click.

**Files:** `ui/chat/ChatHistoryAdapter.kt`

**Verification:** Verified the getData() call at :137 and its definition at :300. Shares a file with the first-query-hoist task.

dogi:
Hoisted the share dialog data map in ChatHistoryAdapter using lazy initialization so the map is built once for the adapter lifetime rather than being re-created on every share-button click.

Fixes #16448

---
*PR created automatically by Jules for task [8267956439723199592](https://jules.google.com/task/8267956439723199592) started by @dogi*


dogi:
## Review: 84/100 — mergeable

Correct hoist, and it quietly fixes something more interesting than the caching it advertises.

**The map's iteration order is user-visible, and this makes it deterministic.** `generateFlatList` iterates `dataMap` with `for ((groupTitle, children) in dataMap)` and appends group headers in iteration order, so the map's ordering *is* the share dialog's group ordering. `getData()` built a `HashMap`, whose order for two keys falls out of the hash codes of the **localized** strings — meaning "Share with community" vs "Share with team/enterprise" could appear in different order in English than in Arabic or Nepali, with nothing in the code hinting at it. `mapOf(...)` returns a `LinkedHashMap`, so the order is now insertion order (community first) in every locale. That's a real, if small, fix and worth mentioning in the description; it also means the dialog's group order may visibly flip in some locales after this lands, which is the intended outcome rather than a regression.

**The unchecked cast goes away too.** `getData() as? Map<String, List<String>> ?: emptyMap()` was casting a value already declared `Map<String, List<String>>` — an unchecked cast that could only ever succeed, and whose `?: emptyMap()` fallback was unreachable. Good riddance.

**Caching localized strings is safe here.** The `by lazy` memoizes six `getString` results for the adapter's lifetime, which would be stale if the app language changed under a live adapter. It can't: the adapter is constructed fresh inside `ChatHistoryFragment.observeScreenData()` (:163) with `requireContext()`, and a locale change recreates the activity and with it the fragment. Same reasoning that makes the existing `daynightGreyColor` lazy safe.

**`LazyThreadSafetyMode.NONE` is correct** — the only reader is `bindShareChat`, on the main thread. Minor inconsistency that the sibling `daynightGreyColor` lazy on the line above uses the default `SYNCHRONIZED`; matching them either way would be tidier.

**Nits**

- `val dataMap = shareDataMap` is a pointless alias now that there's no cast; the three uses below could read `shareDataMap` directly.
- No test. `ChatHistoryAdapter` has no coverage on `master` — but see below.

**Contention:** #16601 also modifies `ChatHistoryAdapter.kt` (the `onBindViewHolder` title/fallback block around :101–111) and adds a new `ChatHistoryAdapterTest`. The hunks don't overlap — this PR touches :56, :137 and :306 — so both should apply cleanly in either order, but whichever lands second should be re-checked. If #16601 lands first, a small case asserting the group order out of `generateFlatList` would slot straight into its new test file and pin the determinism this PR introduces.

**CI** — build (default + lite), both shards, labels green on `511375b`.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16605, opened 2026-08-27
teams: smoother voices notifications repositories dao querying (fixes #16418)

### LINKED ISSUE #16418: Use the existing countTeamChats instead of loading every news row for a notification count
**rating:** 62/100 &nbsp;·&nbsp; **proposed by:** jules(r1), minimax(r2)

`TeamsVoicesViewModel.getFilteredNews` calls `voicesRepository.getFilteredNews(teamId)` and reads `newsList.size` purely to pass to `updateTeamNotification` — loading every news row plus all columns to produce one Int, when `VoicesRepository.countTeamChats(teamId): Long` is already backed by SQL COUNT.

**Ask:** use `countTeamChats(teamId)` for the notification-count call instead of `getFilteredNews(teamId).size`. The full list is still needed elsewhere by the caller — only replace the count derivation, don't remove the list fetch if something else in the same flow still needs it.

**Files:** `ui/teams/voices/TeamsVoicesViewModel.kt`

**Verification:** Verified the call site and that countTeamChats exists (VoicesRepository.kt:43). minimax names the existing count method; jules only gestures at 'an optimized count metric'. Note the list is still returned to the caller, so this saves the column read only if the two paths are actually split.

dogi:
## Summary

Fixes the team-chat **unread badge** so it compares like for like.

The badge is `hasChat = lastCount < chatCount`, computed in `NotificationsRepositoryImpl.getTeamNotifications`. The watermark (`lastCount`) is written by `TeamsVoicesViewModel.getFilteredNews` as `newsList.size` — the size of the top-level feed the user sees (`getTopLevelByTeam`: top-level only, both the `viewableBy`/`viewableId` path and the `viewIn LIKE` path that locally-created team messages use). But `chatCount` came from `getTeamChatViewableIds`, which counted a **different set** — `viewableBy`/`viewableId` only (so `viewIn`-only local posts were invisible to it), **replies included** (no `replyTo` filter), and exact case. The two operands measured different populations, so the comparison was meaningless in both directions:

- A team with any replies had `chatCount > lastCount` permanently → the badge never cleared (the reported symptom).
- A team whose posts were all created on-device (`viewIn` only) had `chatCount == 0` while `lastCount > 0` → the badge never appeared.

### Fix
Make `chatCount` use the **same predicate** as the watermark: replace `getTeamChatViewableIds` with `NewsDao.countTopLevelByTeam`, a SQL `COUNT(*)` mirroring `getTopLevelByTeam` (top-level only, both visibility paths, `COLLATE NOCASE`), exposed via `VoicesRepository.countTopLevelByTeam` and called per team in `getTeamNotifications`. Now both sides of `lastCount < chatCount` are the same query, so the badge reflects new top-level posts since the feed was last opened and clears once it is opened.

### Why the ViewModel is unchanged
`getFilteredNews` already fetches the list (the caller needs it), so `newsList.size` is free. The earlier iteration of this PR replaced it with a SQL `COUNT` for the same value — a no-op second round-trip for a number already in hand (the original `countTeamChats`/`countTopLevelByTeam` predicates are provably equal to `newsList.size` here). So the ViewModel is reverted to master; the real fix is on the badge-read side.

## Files
- `app/src/main/java/org/ole/planet/myplanet/data/room/dao/NewsDao.kt` — replace `getTeamChatViewableIds` with `countTopLevelByTeam`
- `app/src/main/java/org/ole/planet/myplanet/repository/VoicesRepository.kt` / `VoicesRepositoryImpl.kt` — expose `countTopLevelByTeam(teamId)`
- `app/src/main/java/org/ole/planet/myplanet/repository/NotificationsRepositoryImpl.kt` — `chatCount` now uses `countTopLevelByTeam` per team
- `app/src/main/java/org/ole/planet/myplanet/ui/teams/voices/TeamsVoicesViewModel.kt` — unchanged (left at master's `newsList.size`)

## Tests
- `TeamChatBadgeIntegrationTest` (new, Room-backed): posts top-level messages and replies and asserts
  - the badge **clears** after the feed is opened, even with replies present (the reported symptom — would fail on the old code),
  - the badge **appears** when a new top-level post arrives after the feed was opened,
  - the badge **ignores replies** (they aren't part of the top-level feed),
  - `chatCount` equals the displayed top-level feed size.
- `NewsDaoTest.countTopLevelByTeam_matches_getTopLevelByTeam_size` — the COUNT predicate matches the list across viewIn-only (local), viewableBy/viewableId (synced), reply-exclusion, and other-team cases.
- `TeamsVoicesViewModelTest` — verifies `getFilteredNews` writes `newsList.size` as the watermark and issues no extra count query.

`./gradlew testDefaultDebugUnitTest --tests "org.ole.planet.myplanet.repository.TeamChatBadgeIntegrationTest" --tests "org.ole.planet.myplanet.ui.teams.voices.TeamsVoicesViewModelTest" --tests "org.ole.planet.myplanet.data.room.dao.NewsDaoTest" --tests "org.ole.planet.myplanet.repository.NotificationsRepositoryImplTest" --tests "org.ole.planet.myplanet.repository.VoicesRepositoryImplTest"` — all green.

Closes #16418.

_Note: this PR was created by an AI agent (OpenHands) on behalf of the myPlanet maintainers._

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved team chat notification badges by comparing unread counts with the number of top-level posts displayed.
  * Replies are no longer incorrectly counted as new top-level chat messages.
  * Badge behavior now remains accurate when opening feeds or receiving new posts.

* **Tests**
  * Added coverage for team-specific filtering, post counting, replies, and end-to-end notification badge behavior.


dogi:
## Review: 90/100 — mergeable, and more valuable than the description claims

**The framing is wrong, and in the PR's favour.** As a performance change this is a net *loss*: `getFilteredNews(teamId)` still runs and materializes every row, so `newsList` is already in memory and `newsList.size` is free — the diff **adds** a `SELECT COUNT(*)` round trip that `master` didn't make. Nothing is avoided.

What it actually fixes is a **broken watermark comparison**, and that's worth merging on its own. Tracing both sides of the badge:

- **The writer** (`updateTeamNotification` → `TeamNotification.lastCount`) was fed `getFilteredNews(teamId).size` = `NewsDao.getTopLevelByTeam`:
  ```sql
  WHERE (replyTo IS NULL OR replyTo = '')
    AND ((viewableBy = 'teams' COLLATE NOCASE AND viewableId = :teamId COLLATE NOCASE)
         OR viewIn LIKE :teamPattern ESCAPE '\')
  ```
- **The reader** (`getTeamNotifications`, :311–330) computes `chatCount` from `getTeamChatViewableIds`:
  ```sql
  SELECT viewableId FROM news WHERE viewableBy = 'teams' AND viewableId IN (:teamIds)
  ```
  and lights the badge with `hasChat = notification.lastCount < chatCount`.

Those are **different populations**, so the comparison was meaningless in both directions:

1. `getTopLevelByTeam` excludes replies; `getTeamChatViewableIds` counts them. Any team whose chat has replies has `lastCount` permanently below `chatCount` — **the badge never clears, no matter how many times you open the team**.
2. Conversely, `getTopLevelByTeam` also picks up posts matched by `viewIn LIKE :teamPattern` (and uses `COLLATE NOCASE`), which the reader's query does not. A team with enough `viewIn`-shared posts gets `lastCount` pushed *above* `chatCount` — **the badge never lights**.

`countTeamChats` is `SELECT COUNT(*) FROM news WHERE viewableBy = 'teams' AND viewableId = :teamId` — the exact same predicate as the reader's query, same collation, same reply handling. So after this change the watermark and the counter finally measure the same set and the badge behaves. That's the merge case; I'd rewrite the summary around it.

**Other checks**

- `.toInt()` on the `Long` count is safe — this is a per-team row count, nowhere near overflow.
- The teamId threaded through is the same `getEffectiveTeamId()` value the fragment passes (`TeamsVoicesFragment:121`), so writer and reader key on the same id.
- The added COUNT query is one indexed aggregate per screen open. Fine.

**Test is a genuine regression test.** `coVerify(exactly = 0) { updateTeamNotification(teamId, newsList.size) }` with a 2-element list against a stubbed count of `7L` fails on `master` (which calls it with `2`). Both the positive and negative assertions are load-bearing.

**Worth a follow-up, not a blocker:** the two queries stay correct only by convention — nothing stops a future edit to `getTeamChatViewableIds` or `countTeamChats` from drifting apart again and silently re-breaking the badge. A comment on `NewsDao.countTeamChats` noting it must match `getTeamChatViewableIds`, or better, deriving both from one predicate, would make that durable.

**CI** — build (default + lite), both shards, labels green on `4370b6d`.

**Verdict:** mergeable.

---

J-S-webskas (dismissed):

Okuro3499 (dismissed):
```
File: app/src/main/java/org/ole/planet/myplanet/data/room/dao/NewsDao.kt:81-82 (countTeamChats), used from                                                                       
    app/src/main/java/org/ole/planet/myplanet/ui/teams/voices/TeamsVoicesViewModel.kt:60                                                                                             
    Problem: countTeamChats's predicate (viewableBy='teams' AND viewableId=:teamId) is a strict subset of getTopLevelByTeam's predicate (also includes viewIn LIKE :teamPattern, the 
    path all locally-created team messages actually use per News.createNews). The notification count derived from it no longer reflects the list actually shown, so the unread badge 
    becomes wrong for normal usage.
```

dogi:
Good catch — you're right that `countTeamChats`'s predicate (`viewableBy='teams' AND viewableId=:teamId`) is a strict subset of `getTopLevelByTeam`'s and omits the `viewIn LIKE` path that locally-created team messages actually use (`News.createNews` + `getViewInJson` set `viewIn`, not `viewableBy`/`viewableId`), so the notification count would have drifted from the displayed feed.

Fixed in e4d0f5c: added `NewsDao.countTopLevelByTeam(teamId, teamPattern)`, a SQL `COUNT(*)` that mirrors `getTopLevelByTeam`'s exact predicate (top-level only, both the `viewableBy`/`viewableId` and `viewIn LIKE` visibility paths), exposed via `VoicesRepository.countTopLevelByTeam` and used in `TeamsVoicesViewModel.getFilteredNews`. `countTeamChats` is left untouched — it's still used by `getTeamNotifications` for its chat-count comparison, which intentionally counts `viewableId` rows.

Tests: a Room test (`countTopLevelByTeam_matches_getTopLevelByTeam_size`) proves the COUNT equals the list size across the viewIn-only (locally-created), viewableBy/viewableId (synced), reply-exclusion, and other-team cases; the ViewModel test now asserts `countTopLevelByTeam` is used and `countTeamChats` is not. Both suites green locally.

_This reply was created by an AI agent (OpenHands) on behalf of the maintainers._

dogi:
## Review — rating 45/100

This change computes the same number it already had, and the mismatch it is meant to fix is on the other side of the comparison.

### `countTopLevelByTeam` is provably equal to `newsList.size`

`getFilteredNews` is a bare passthrough to the query the new COUNT mirrors:

```kotlin
override suspend fun getFilteredNews(teamId: String): List<News> {
    return newsDao.getTopLevelByTeam(teamId, teamIdPattern(teamId))
}
```

And the two DAO predicates are identical, character for character, differing only in `SELECT COUNT(*)` vs `SELECT *` and the dropped `ORDER BY`:

```sql
WHERE (replyTo IS NULL OR replyTo = '')
  AND ((viewableBy = 'teams' COLLATE NOCASE AND viewableId = :teamId COLLATE NOCASE)
       OR viewIn LIKE :teamPattern ESCAPE '\')
```

So in `getFilteredNews`, `voicesRepository.countTopLevelByTeam(teamId)` returns exactly `newsList.size` — the value already in hand. The PR replaces a free in-memory `.size` with a second round-trip to the database for the same answer.

The new DAO test makes this explicit rather than hiding it: `assertEquals(listSize.toLong(), count)` is a proof of the no-op. And the ViewModel test only distinguishes the two by mocking `countTopLevelByTeam` to return 7 while the list has 2 elements — a divergence that cannot occur in production.

### The actual bug is the watermark/badge population mismatch

The badge is decided in `NotificationsRepositoryImpl.getTeamNotifications`:

```kotlin
val chatViewableIds = voicesRepository.getTeamChatViewableIds(teamIds)
// … counted by grouping …
```

where `getTeamChatViewableIds` is:

```sql
SELECT viewableId FROM news WHERE viewableBy = 'teams' AND viewableId IN (:teamIds)
```

Compare the two populations being compared against each other:

| | `chatCount` (the badge side) | `lastCount` (the watermark this PR writes) |
|---|---|---|
| replies | **included** | excluded (`replyTo IS NULL OR ''`) |
| locally-created `viewIn`-only posts | **excluded** | included (the `viewIn LIKE` branch) |
| case sensitivity | exact `'teams'` / `viewableId` | `COLLATE NOCASE` on both |

They count different things, so the comparison is meaningless in both directions:

- A team with any replies has `chatCount > lastCount` permanently → **the badge never clears**, which is the reported symptom.
- A team whose posts were all created on-device (`viewIn` only, no `viewableBy`) has `chatCount == 0` while `lastCount > 0` → the badge never appears.

Writing a *differently derived* Population-B number into `lastCount` cannot fix a comparison whose other operand is Population A. The fix has to make both sides measure the same set — either give the badge a batched count with the `getTopLevelByTeam` predicate (a `GROUP BY` variant of the new query, keyed by team, so it stays one query for N teams), or write a watermark shaped like `getTeamChatViewableIds`. The former is better, since the badge should track the feed the user actually sees.

The `countTopLevelByTeam` query added here is genuinely useful as the building block for that batched version — it is the right predicate, just wired to the wrong side.

Once both sides agree, the test worth having is an integration one: post a top-level message and a reply, assert the badge clears after the feed is opened. That is the behaviour that is broken today, and neither new test would catch it.

Moving to `change`.

---

dogi:
You were right on both counts, and the earlier change was wrong.

**1. The ViewModel change was a no-op.** `countTopLevelByTeam` mirrors `getTopLevelByTeam` exactly, so in `getFilteredNews` it returns `newsList.size` — a value already in hand. It replaced a free in-memory `.size` with a second DB round-trip. I reverted the ViewModel to master's `newsList.size`; the list is fetched for the caller regardless, so `.size` is the right and free answer there. (The test that asserted the divergence mocked `countTopLevelByTeam` to 7 against a 2-element list — exactly the impossible-in-production split you called out; it's gone.)

**2. The real bug is the badge comparison.** `getTeamNotifications` compared `lastCount` (the watermark, top-level + both visibility paths + NOCASE) against `chatCount` from `getTeamChatViewableIds` (viewableBy/viewableId only, replies included, exact case) — different populations, so the badge never cleared for teams with replies and never appeared for viewIn-only local posts. Fixed in 3e3854c: `chatCount` now uses `countTopLevelByTeam` per team — the same predicate as the watermark — so both sides of `lastCount < chatCount` are the same query. `getTeamChatViewableIds` (the buggy query) is removed.

I didn't batch the viewIn path into one GROUP BY: the `viewIn LIKE :teamPattern` is per-team (each team has its own escaped pattern), so a single LIKE can't cover N teams. The badge read now does one indexed COUNT per team — cheap, and honest about the per-team pattern. If you'd rather keep it to one query, the alternative is to load the top-level rows once and group in memory, but that's the load-every-row shape the issue set out to avoid, so I went with per-team COUNTs.

The integration test you asked for is in: `TeamChatBadgeIntegrationTest` (Room-backed) posts a top-level message and a reply, opens the feed, and asserts the badge clears; a second case asserts a new top-level post makes it appear; a third asserts replies alone don't. `badge clears after the feed is opened, even with replies present` is the case that would have failed on the old code.

_This reply was created by an AI agent (OpenHands) on behalf of the maintainers._

dogi:
## Review — rating 84/100 (was 45)

This is the right fix. The previous revision changed the watermark; this one changes the **badge side**, which is where the mismatch actually lived.

### Both operands now measure the same set

```kotlin
// NotificationsRepositoryImpl.getTeamNotifications
```

`chatCount` and `lastCount` now both resolve to `getTopLevelByTeam`'s predicate with the same `teamIdPattern(teamId)` — top-level only, `COLLATE NOCASE`, and covering both the `viewableBy`/`viewableId` and `viewIn LIKE` paths. `hasChat = lastCount < chatCount` finally compares like for like, so:

- replies no longer inflate `chatCount` → the badge clears
- locally-created `viewIn`-only posts are no longer invisible → the badge appears

Reverting `TeamsVoicesViewModel` to `newsList.size` is also correct: since `getFilteredNews` *is* `getTopLevelByTeam`, the list size is already the right number and a second query would buy nothing. The test asserting `coVerify(exactly = 0) { voicesRepository.countTopLevelByTeam(teamId) }` pins that, which is a nice way to encode "do not reintroduce the redundant query".

Dropping `getTeamChatViewableIds` from the DAO, interface and impl rather than leaving it beside its replacement is the right call.

### The integration test is what this needed

`TeamChatBadgeIntegrationTest` wires a real in-memory `AppDatabase` through the real `VoicesRepositoryImpl` and `NotificationsRepositoryImpl` and asserts the actual `hasChat` outcome. The four cases map exactly onto the failure modes:

- badge clears after opening the feed *with a reply present* — the reported bug, and it fails on master
- badge appears for a new top-level post arriving via the `viewIn` path — the case the old `chatCount` could not see at all
- replies do not raise the badge
- `countTopLevelByTeam` equals `getFilteredNews().size`

That last one is the invariant the whole fix rests on, checked against real SQL rather than asserted in a comment.

### One regression to acknowledge: this is now N+1

Master fetched every team's count in a single query:

```kotlin
val chatViewableIds = voicesRepository.getTeamChatViewableIds(teamIds)   // ... WHERE viewableId IN (:teamIds)
```

The replacement loops:

```kotlin
for (teamId in teamIds) {
}
```

That is one query per team, each a full scan of `news` (the `viewIn LIKE '%…%'` leading wildcard means no index can help). `getTeamNotifications` runs from `BaseDashboardFragment:252` on dashboard render, so a user in eight teams now does eight table scans where they previously did one.

N is a user's team count, so this is unlikely to be visible in practice — and correct-but-slower beats fast-but-wrong. But the comment above the loop still says nothing about the change in shape, and the block it replaced was explicitly labelled "in a single query". Worth either batching it (a `GROUP BY viewableId` pass for the `viewableBy` branch plus a second pass for `viewIn`, merged in memory) or noting the tradeoff in the comment so the next reader knows it was deliberate.

### Cleanup

`countTeamChats` now has **no callers** anywhere in `app/src/main` — it was only reachable through the path this PR replaced. Remove it from `NewsDao`, `VoicesRepository` and `VoicesRepositoryImpl`, or it becomes the next piece of dead API someone has to reason about.

Nits in the new test: several fully-qualified types inline (`org.ole.planet.myplanet.data.room.dao.NewsDao`, `java.util.UUID`, `kotlinx.coroutines.test.UnconfinedTestDispatcher`) instead of imports, an unused `import org.ole.planet.myplanet.di.PlainGson`, and `time = System.nanoTime()` on a field that holds epoch millis — harmless for ordering here, but confusing to read.

Moving to `ready`.

---

Okuro3499 (approved):

PR #16606, opened 2026-08-27
all: smoother markdown utils link movement caching (fixes #16499)

### LINKED ISSUE #16499: Reuse one CustomLinkMovementMethod instance in MarkdownUtils
**rating:** 45/100 &nbsp;·&nbsp; **proposed by:** copilot(r2)

`setMarkdownText` allocates a new CustomLinkMovementMethod on every call although Markwon itself is a cached singleton; list/detail binds that set markdown repeatedly create short-lived movement methods.

**Ask:** hoist the `CustomLinkMovementMethod` into a single reused instance (matching how Markwon itself is already cached) instead of allocating a new one on every `setMarkdownText` call.

**Files:** `utils/MarkdownUtils.kt`, `test/.../MarkdownUtilsTest.kt`

**Verification:** File exists; not line-verified. Shares a file with the StringBuffer task.

dogi:
## Description

`MarkdownUtils.setMarkdownText` was allocating a brand-new `CustomLinkMovementMethod` on every call, even though the Markwon instance itself is already cached as a singleton. List and detail bindings that set markdown repeatedly created many short-lived movement-method objects.

This hoists `CustomLinkMovementMethod` into a single reused instance (a `private val` on the `MarkdownUtils` object, matching how `markwonInstance` is already cached) instead of allocating a new one per `setMarkdownText` call.

## Changes

- `utils/MarkdownUtils.kt`: added `private val linkMovementMethod = CustomLinkMovementMethod()` and assigned that shared instance in `setMarkdownText` instead of `CustomLinkMovementMethod()`.
- `test/.../MarkdownUtilsTest.kt`: added `setMarkdownText_reuses_same_movement_method_instance`, which binds two separate `TextView`s and asserts both receive the same `MovementMethod` instance.

## Verification

Ran locally:

```
./gradlew testDefaultDebugUnitTest --tests "org.ole.planet.myplanet.utils.MarkdownUtilsTest"
```

All 7 tests in `MarkdownUtilsTest` pass, including the new reuse test.

```
tests="7" skipped="0" failures="0" errors="0"
```

## Related Issue

Fixes #16499

---

@dogi can click here to [continue refining the PR](https://app.all-hands.dev/conversations/71e63b16-e916-4b60-82b1-ba4cb99dc845)


dogi:
## Review: 88/100 — mergeable

Correct, and it matches what the framework itself does — `LinkMovementMethod.getInstance()` is a process-wide static singleton shared by every `TextView` in the system. The only question worth asking is whether *this subclass* is safe to share, and it is.

**`CustomLinkMovementMethod` is genuinely stateless.** I read the whole class (`MarkdownUtils.kt:103–127`): no fields, no constructor body, no captured `Context`. Its single override takes `widget`, `buffer` and `event` as parameters and derives everything — `totalPaddingLeft/Top`, `scrollX/Y`, `layout`, the `CustomImageSpan` lookup — from the passed-in widget and buffer, then delegates to `super`. There is no per-TextView state to be corrupted by sharing, and because it holds no `Context`, promoting it to a field on the `MarkdownUtils` **object** (i.e. process-lifetime) leaks nothing. Those two properties are the whole ballgame here; had the class cached the last widget or a context, this change would have been a leak or a cross-row bug.

**Class-init timing is fine.** The field is eagerly initialized in `MarkdownUtils`'s `<clinit>`, so merely touching any member of the object — including the pure-string `prependBaseUrlToImages` — now constructs an `android.text.method.LinkMovementMethod`. That would blow up in a plain JVM unit test against android.jar stubs, but `MarkdownUtilsTest` is the only test in the tree that references `MarkdownUtils` and it's `@RunWith(RobolectricTestRunner::class)`, so real framework classes are present. Worth knowing if a non-Robolectric test of the markdown string helpers is ever added — it would need Robolectric solely because of this field. A `by lazy` would sidestep that, though it's not required today.

**Test is a real regression test** — `assertSame` on the movement methods of two separately-bound `TextView`s fails on `master`, where each `setMarkdownText` allocates a fresh instance.

**Nits**

- `ApplicationProvider.getApplicationContext<android.content.Context>()` fully-qualifies the type inline; `import android.content.Context` would read better.
- The saving is one small object per `setMarkdownText` call. Real in list bindings, but modest — the honest framing is "match the framework's singleton idiom", which this does.

**Contention:** #16603 also edits `MarkdownUtils.kt`, at `:89` (the `StringBuffer`/`StringBuilder` line) — a different hunk, so no textual conflict. Note that #16603 is currently at `change`; nothing here depends on it.

**CI** — build (default + lite), both shards, labels green on `cee6a74`.

**Verdict:** mergeable.

---

J-S-webskas (approved):

Okuro3499 (approved):

PR #16607, opened 2026-08-27
login: smoother storage category detail view modelling (fixes #16449)

### LINKED ISSUE #16449: Keep storage-category selection state in its ViewModel
**rating:** 57/100 &nbsp;·&nbsp; **proposed by:** deepseek(r2), opus(r1)

`StorageCategoryDetailFragment` reads `viewModel.uiState.value.items` in two click listeners (:104, :111) to re-derive state the ViewModel already owns and hands the lists back to `viewModel.deleteItems(...)`; `updateSelectionState` recomputes `items.count { it.isChecked }` (:159) in the view layer on every emission. Reading `.value` off a StateFlow inside a click listener is the classic stale-snapshot bug.

**Ask:** move the selection logic into the ViewModel (e.g. `deleteSelected()`/`deleteAll()` methods and a `checkedCount` on the state) so the fragment stops reading `.value` in click listeners and stops recomputing the count in the view layer.

**Files:** `ui/settings/StorageCategoryViewModel.kt`, `ui/settings/StorageCategoryDetailFragment.kt`

**Verification:** Verified all four line references. opus moves the selection into the ViewModel (deleteSelected/deleteAll) and fixes the stale read; deepseek only adds a checkedCount field to the state.

dogi:
## Description

Fixes #16449 — moves the storage-category selection logic into `StorageCategoryViewModel` so the fragment stops reading `uiState.value` in click listeners and stops recomputing the checked count in the view layer.

The previous `StorageCategoryDetailFragment` read `viewModel.uiState.value.items` in two click listeners (`:104`, `:111`) to re-derive state the ViewModel already owned, then handed the lists back to `viewModel.deleteItems(...)`. `updateSelectionState` also recomputed `items.count { it.isChecked }` (`:159`) in the view layer on every emission. Reading `.value` off a StateFlow inside a click listener is the classic stale-snapshot bug.

## Changes

**`ui/settings/StorageCategoryViewModel.kt`**
- Added `deleteSelected()` and `deleteAll()` methods that derive the deletion target from the ViewModel's own state. `deleteItems(...)` is now `private`.
- Added `checkedCount` / `allChecked` computed properties to `StorageCategoryUiState` so the count lives on the state the ViewModel owns.
- The delete buttons no longer need to re-derive the list of items to delete; they just call `viewModel.deleteSelected()` / `viewModel.deleteAll()`.

**`ui/settings/StorageCategoryDetailFragment.kt`**
- The delete-button click listeners now delegate to `viewModel.deleteSelected()` / `viewModel.deleteAll()` instead of filtering `uiState.value.items` and passing the result back into `deleteItems`. They still read `uiState.value` only for the confirmation-dialog count (a pure display value), not for the deletion target list.
- `updateSelectionState` now takes the `StorageCategoryUiState` and reads its `checkedCount` / `allChecked` instead of recomputing `items.count { it.isChecked }` in the view layer.
- Removed the now-unused private `deleteItems(...)` helper from the fragment.

## Verification

- New unit test `StorageCategoryViewModelTest` (8 cases) covers `loadResources`, `toggleItemChecked` + `checkedCount`, `toggleAllChecked` (check all → uncheck all), `deleteSelected` (deletes only checked, and is a no-op with no selection), `deleteAll` (deletes every loaded item, and is a no-op when empty), and the `deleteCompleteEvent` emission.
- `./gradlew testDefaultDebugUnitTest --tests "org.ole.planet.myplanet.ui.settings.StorageCategoryViewModelTest"` → **8 tests, 0 failures, 0 errors**.

## Checklist

- [x] Code builds and unit tests pass
- [x] No new secrets / credentials introduced
- [x] Follows existing patterns (MockK + `runTest` + `TestDispatcherProvider`, `mockkObject(FileUtils)` for the `getOlePath` call)

---

@dogi can click here to [continue refining the PR](https://app.all-hands.dev/conversations/ebf6870c-e35a-4a03-9e51-6d94a678fb73)


dogi:
## Review: 87/100 — mergeable

Good encapsulation change, and the new test file is the strongest part of it — `StorageCategoryViewModel` had **zero** coverage before this.

**Correctness checks I actually ran down**

- **`checkedCount` / `allChecked` are body properties, not constructor params**, so they stay out of the generated `equals`/`hashCode`/`copy`. That matters: `StorageCategoryUiState` sits behind a `MutableStateFlow` whose conflation and `update` semantics depend on data-class equality. Putting derived values in the constructor would have been a subtle mess. This is right.
- **`deleteCompleteEvent` is a `Channel(Channel.BUFFERED)` + `receiveAsFlow()`**, not a zero-replay `SharedFlow`. That's why `assertEquals(Unit, viewModel.deleteCompleteEvent.first())` in the last test doesn't hang on an emission that already happened — the buffered element is still there when `.first()` subscribes. Worth stating, because the same test against a `SharedFlow(replay = 0)` would deadlock and time out.
- **The empty guards are defensive, not load-bearing.** When `state.isEmpty`, the fragment sets `binding.actionButtons.visibility = GONE` (:130), so `deleteAll()` can't be reached from the UI with an empty list. The guards are still correct to have, and the two "does not delete" tests pin them.
- **`hasLoaded` doesn't break the empty-list test** — it's per-instance and `@Before` builds a fresh ViewModel, so re-stubbing `getOfflineResourceItems` to `emptyList()` before the first `loadResources` call takes effect.

**Two behavior changes worth surfacing in the description**

1. **The deletion target is now read at confirm time, not at click time.** `deleteSelected()` re-derives from `_uiState.value` when it runs, whereas the old code captured `selected` before showing the dialog. This is the improvement the summary is reaching for, but note it's now *slightly* inconsistent with the dialog text: `confirmDelete` is still passed a count snapshotted at click time. The bottom sheet is modal so the two can't diverge in practice — just don't let a later change assume they're the same read.
2. **The old fragment-side `deleteItems` had `if (_binding == null) return`**, so confirming after the view was destroyed (e.g. a rotation with the dialog up) silently dropped the deletion. The VM now performs it regardless. That's more correct — the user confirmed, `viewModelScope` outlives the view, and the recreated view still collects `deleteCompleteEvent` and dismisses — but it *is* a change, and the description doesn't mention it.

**Nits**

- `allChecked` calls `checkedCount`, which is itself an O(n) `count {}`, so `updateSelectionState` now makes two passes over `items` where the old code made one. Irrelevant at storage-list sizes, but `val checkedCount = state.checkedCount` then `checkedCount == state.items.size && ...` would avoid it. More importantly, neither property is memoized — anything that reads them in a loop pays each time.
- The test uses `Dispatchers.setMain(testDispatcher)` / `resetMain()` by hand. The repo has `MainDispatcherRule` for exactly this and it's the documented convention — the checklist claims "follows existing patterns", and it does for MockK/`TestDispatcherProvider`, but not here.
- `@Before fun setup() = runTest { ... }` wraps a body with no suspending calls; the `runTest` is inert. Plain `@Before fun setup() { ... }` is clearer.

**Contention:** #16581 also touches the storage settings fragments. Whichever lands second will need a re-check against this file.

**CI** — build (default + lite), both shards, labels green on `364500e`.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16609, opened 2026-08-27
life: smoother health examination formatting (fixes #16452)

### LINKED ISSUE #16452: Reuse buildApkLog-style consolidation for the exam-condition JSON parse in HealthExaminationAdapter
**rating:** 56/100 &nbsp;·&nbsp; **proposed by:** kimi(r1)

`HealthExaminationAdapter.showAlert` parses the exam's conditions JSON inline in the UI layer, re-running Gson parsing per examination row; MyHealth is the data-shape owner and a companion helper puts JSON semantics somewhere unit tests can reach.

**Ask:** move the inline conditions-JSON parsing out of `HealthExaminationAdapter.showAlert` into a companion helper on `MyHealth`, so the JSON semantics live with the data-shape owner and are unit-testable.

**Files:** `model/MyHealth.kt`, `ui/health/HealthExaminationAdapter.kt`

**Verification:** File paths verified; the inline parse was not line-verified.

dogi:
## Summary

`HealthExaminationAdapter.showAlert` previously parsed the examination's `conditions` JSON inline in the UI layer, re-running Gson parsing per examination row and keeping the JSON semantics out of reach of unit tests. The inline `.asBoolean` predicate also silently discarded *all* conditions whenever a single value was `null` or non-boolean — a real bug on CouchDB-sourced data.

This moves the parsing into a companion helper on the data-shape owner and hardens value handling:

- Added `HealthExamination.formatConditions(conditions: String?): String` on `HealthExamination`'s existing `companion object` (which already reads/writes the `conditions` blob in `fromJson`/`serialize`) — parses the JSON object, returns the names of conditions flagged `true` joined by `", "`, and returns `""` for null/blank/malformed/non-object input.
- Switched the predicate from `conditionsMap[it].asBoolean` to `JsonUtils.getBoolean(it, conditionsMap)`, matching `HealthRepositoryImpl.getExaminationConditions`. This tolerates a missing/null/non-boolean value (treated as `false`) instead of throwing and blanking every condition — so a true flag now survives a malformed sibling.
- `HealthExaminationAdapter.showAlert` calls `HealthExamination.formatConditions(realmExamination.conditions)`; the `MyHealth` import it had gained is dropped.
- The catch `printStackTrace()`s instead of swallowing silently, matching the sibling conditions-blob handler on master.

## Why `HealthExamination`

The issue asks for the JSON semantics to live with the data-shape owner and be unit-testable. `HealthExamination` owns the `conditions` field and its `companion object` already serializes/deserializes that exact blob — so the read now sits next to the write, with no new companion object. The helper is pure (no Android dependencies), so it runs in plain JUnit.

## Files

- `app/src/main/java/org/ole/planet/myplanet/model/HealthExamination.kt` — added `formatConditions` to the existing `companion object`.
- `app/src/main/java/org/ole/planet/myplanet/model/MyHealth.kt` — restored to holder-only form (the helper no longer lives here).
- `app/src/main/java/org/ole/planet/myplanet/ui/health/HealthExaminationAdapter.kt` — replaced inline parse with helper call; dropped the `MyHealth` import.
- `app/src/test/java/org/ole/planet/myplanet/model/HealthExaminationTest.kt` — new unit tests covering null, blank, all-true, all-false, mixed, a `null` value alongside a `true` flag, a non-boolean value alongside a `true` flag, malformed, non-object, and empty-object inputs.

## Verification

No JDK is available in this sandbox, so the test suite could not be executed locally. The new test class (`HealthExaminationTest`) targets `./gradlew testDefaultDebugUnitTest --tests "org.ole.planet.myplanet.model.HealthExaminationTest"` and follows the existing pure-unit-test pattern used by `ExamQuestionTest` / `RatingTest`. CI (`test.yml`) will run it.

Behavior is preserved for well-formed input, and improved for malformed values: a single bad value no longer discards the good ones. A malformed or non-object document still yields `""`.

Fixes #16452.

@dogi can click here to [continue refining the PR](https://app.all-hands.dev/conversations/67531591-4639-4f2d-a174-50703d38fd7a)


dogi:
## Review: 78/100 — needs a change

The extraction itself is faithful and the tests are the best part. Two things should change before this lands, and one of them is a real bug the extraction is perfectly placed to fix.

### 1. `MyHealth` is not the data-shape owner — `HealthExamination` is

The summary argues the helper belongs on "the data-shape owner for health". But `MyHealth` (`model/MyHealth.kt`) is a 14-line holder of `profile` / `userKey` / `lastExamination` plus a nested `MyHealthProfile` — **it has no `conditions` field and never touches this JSON**. The field is `HealthExamination.conditions` (`model/HealthExamination.kt:28`), the call site passes `realmExamination.conditions`, and `HealthExamination` **already has a `companion object`** whose `deserialize` (:73) and `serialize` (:97) read and write that exact blob:

```kotlin
myHealth.conditions = JsonUtils.gson.toJson(JsonUtils.getJsonObject("conditions", act))
…
JsonUtils.addJson(`object`, "conditions", JsonUtils.gson.fromJson(health.conditions, JsonObject::class.java))
```

Moving `formatConditions` there puts the read next to the write, needs no new `companion object`, and drops the `import org.ole.planet.myplanet.model.MyHealth` the adapter just gained. `MyHealthTest.kt` becomes `HealthExaminationTest.kt` with no other change.

### 2. There is already a parser for this blob, and it's the more correct one

`HealthRepositoryImpl.getExaminationConditions` (:94–108) parses the same field:

```kotlin
val conditions = gson.fromJson(examination.conditions, JsonObject::class.java)
for (key in conditions.keySet()) {
    result[key] = JsonUtils.getBoolean(key, conditions)
}
```

So this PR creates a **third** copy of the parse rather than converging on the existing one — which matters because the two disagree on robustness. `JsonUtils.getBoolean` tolerates a missing or non-boolean value; `conditionsMap[it].asBoolean` **throws** on `JsonNull` (`UnsupportedOperationException`) or on an object/array value, and the blanket `catch` then discards **every** condition:

| input | repository path | `formatConditions` |
|---|---|---|
| `{"diabetes": true, "asthma": null}` | `{diabetes=true, asthma=false}` | `""` — the true flag vanishes |
| `{"diabetes": true, "asthma": {}}` | `{diabetes=true, asthma=false}` | `""` |

A `null` value in a CouchDB-sourced blob is entirely plausible, and the failure is silent: the user sees a blank conditions line on an examination that has conditions. This is **pre-existing** — `master`'s inline code behaves identically, so it's not a regression — but this PR is the moment to fix it, and the fix is one expression:

```kotlin
.filter { JsonUtils.getBoolean(it, conditionsMap) }
```

That also makes the KDoc accurate; as written, "Any parse failure … yields an empty string" understates it — a single malformed *value* discards all the good ones.

Please add the two cases that pin this: `{"diabetes": true, "asthma": null}` and a non-boolean value, both asserting `diabetes` still comes through.

### What checks out ✅

- **Behavior is genuinely preserved for valid input.** `master` seeded `conditionsText = ""` and only assigned after `joinToString` completed, so a mid-parse throw left it empty — identical to the extracted `catch` returning `""`.
- **`formatConditions("   ")` → `""` works, but not via the guard.** `isNullOrEmpty()` doesn't catch a blank string, so it reaches Gson; `fromJson` on a whitespace-only document hits EOF and returns `null`, so the `!= null` branch produces `""`. The test passes, just for a subtler reason than it looks — `isNullOrBlank()` would make the intent explicit and skip the parse.
- **`formatConditions("[\"diabetes\"]")` → `""`** via the `ClassCastException` Gson throws casting a `JsonArray` to `JsonObject`, caught by the blanket handler. Correct, if incidental.
- **No orphaned import.** Dropping the bare `JsonUtils` import is right; `com.google.gson.JsonObject` is still used at adapter lines 121, 135 and 172.
- **Tests are good**: 9 cases, and comparing `split(", ").toSet()` rather than an exact string keeps them independent of key order.
- **CI** — build (default + lite), both test shards, labels all green on `25d1d60`. (`mergeable_state` reads `unstable`, but no check is red.)

### Nit

`catch (e: Exception) { "" }` swallows silently with an unused binding, which now cuts against the direction #16595 just took (tagged `Log.e` instead of silent/`printStackTrace` handling). A `Log.e("MyHealth", "…", e)` here would make a malformed blob diagnosable instead of invisible.

**Verdict:** move the helper to `HealthExamination`'s existing companion and switch the predicate to `JsonUtils.getBoolean` (with the two extra tests); then this is a clean merge.

---

dogi:

---

dogi:
Thanks for the review @dogi — addressed both substantive points and the nit's spirit in `22af0cf`. No JDK in this sandbox so CI (`test.yml`) will run `HealthExaminationTest`.

**1. Moved the helper to `HealthExamination` (accepted).** You're right that `MyHealth` never touches this blob. `formatConditions` now lives on `HealthExamination`'s existing `companion object`, right above `fromJson`/`serialize`, which already read/write `conditions`. `MyHealth.kt` is back to its holder-only form, and the `MyHealth` import the adapter gained is gone. `MyHealthTest.kt` → `HealthExaminationTest.kt`.

**2. Switched to `JsonUtils.getBoolean` + added the two tests (accepted).** This was the real bug and the extraction was exactly the place to fix it. The old `conditionsMap[it].asBoolean` threw on a `JsonNull`/object/array value and the blanket catch then discarded *every* condition — so `{"diabetes": true, "asthma": null}` returned `""`, blanking the line on an examination that had conditions. Now `.filter { JsonUtils.getBoolean(it, conditionsMap) }` evaluates each value safely (matching `HealthRepositoryImpl.getExaminationConditions`), and a malformed sibling no longer takes the true flags down with it. Added both pinning cases: `{"diabetes": true, "asthma": null}` and `{"diabetes": true, "asthma": {}}`, each asserting `diabetes` still comes through.

**Minor:** `isNullOrEmpty()` → `isNullOrBlank()` so the blank guard is explicit and skips the parse. KDoc updated to state that a missing/non-boolean value is treated as `false` (tolerated), not that any failure blanks everything.

**Nit (`Log.e` vs silent swallow) — kept the helper pure, but made it diagnosable.** The catch now `printStackTrace()`s instead of swallowing silently (was `catch (e: Exception) { "" }` with an unused binding). I did *not* switch to `android.util.Log.e`: the helper is deliberately Android-free so `HealthExaminationTest` stays plain JUnit (no `mockkStatic(Log::class)`/Robolectric), which is the property the "What checks out ✅" section praised. `#16595` (the `Log.e` direction) is also still open and its scope is config/secure-store/file/camera/download-worker — it doesn't touch this conditions blob, whose on-master sibling (`getExaminationConditions`) still uses `printStackTrace()`. So `printStackTrace()` matches the current merged convention for this exact blob while keeping the unit test pure. Happy to revisit if `#16595` lands and sweeps health too.

Still leaving this as draft, as before.

_This reply was created by an AI agent (OpenHands) on behalf of the maintainer._

dogi:
## Review: 89/100 — mergeable

Everything from my earlier pass is addressed, and the two tests I asked for are exactly the two that fail on the previous revision.

**Moved to the right owner.** `formatConditions` now sits in `HealthExamination`'s existing `companion object`, beside the `fromJson`/`serialize` pair that reads and writes the same `conditions` field. `MyHealth` — which has no such field — is untouched, and the adapter's new import points at the class that actually owns the data.

**The tolerant predicate is in.** `JsonUtils.getBoolean(it, conditionsMap)` replaces `conditionsMap[it].asBoolean`, so a single malformed value no longer discards every other flag. The KDoc says so and names `HealthRepositoryImpl.getExaminationConditions` as the behaviour being matched, which is exactly the right cross-reference to leave for the next reader — the two parsers of this blob now agree by design rather than by luck.

**The regression tests are precise:**

- `formatConditions_nullValueDoesNotDiscardTrueFlags` — `{"diabetes": true, "asthma": null}` → `{diabetes}`
- `formatConditions_nonBooleanValueDoesNotDiscardTrueFlags` — `{"diabetes": true, "asthma": {}}` → `{diabetes}`

Both returned `""` before. That was the user-visible bug — a blank conditions line on an examination that has conditions — and it's now pinned.

`isNullOrBlank()` also replaces `isNullOrEmpty()`, so the whitespace case is handled by the guard rather than incidentally by Gson returning null at EOF.

### Heads-up: #16494 will break one of these tests

`HealthExaminationTest` is plain JUnit with no Robolectric, and `formatConditions_nonBooleanValueDoesNotDiscardTrueFlags` drives `JsonUtils.getBoolean` into `safeGet`'s catch block (`JsonObject.getAsBoolean()` throws `UnsupportedOperationException`; the `null` case doesn't, since `getBoolean` short-circuits on `el !is JsonNull`).

Today `safeGet`'s catch calls `e.printStackTrace()` — harmless off-device. **#16494 changes it to `Log.d(...)`**, and with no `returnDefaultValues` in `app/build.gradle`'s `unitTests` block, an unmocked `android.util.Log` call throws *"Method d in android.util.Log not mocked"*. So whichever of these two lands second, this test class will need `mockkStatic(Log::class)` (as `JsonUtilsTest` does in that PR) or a Robolectric runner. Worth a note on both PRs so it isn't discovered as a mystery red build.

**Nit:** the catch is back to `e.printStackTrace()`. That does preserve the cause, which is what I was after — but it's the exact pattern #16595 is systematically replacing. A tagged `Log.e("HealthExamination", "…", e)` would fit the direction (and would make the Robolectric question above moot by forcing the decision now).

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16618, opened 2026-08-28
courses: smoother submissions landscaping (fixes #16617)

### LINKED ISSUE #16617: fix mySubmissions cant be scrolled on landscape mode
https://github.com/user-attachments/assets/56bde0b2-ef21-43d4-be9e-517ea61ba7d8

ragilzakaria:
https://github.com/user-attachments/assets/122d8b97-b40f-45a5-b5fa-3e7391c62bc6

fixes #16617 

## Summary by CodeRabbit

- **New Features**
  - Added a dedicated landscape layout for the My Submission screen.
  - Added landscape support for searching and reviewing exam or survey submissions.

- **Bug Fixes**
  - Improved app stability when saving logs or initializing services encounters an error.
  - Prevented secure preferences startup issues from interrupting app use.

- **Style**
  - Refined submission headers, search spacing, column alignment, text sizing, and row layouts.
  - Improved readability of submission titles, statuses, submitters, dates, and submission counts.


dogi:
## Review: 68/100 — needs a change

The landscape fix is correct. The problem is that it's one of **four unrelated changes** in a PR titled for only one of them.

### The actual fix is right ✅

`rv_mysurvey` going from `layout_height="match_parent"` to `0dp` + `layout_weight="1"` is the correct fix: as the last child of a vertical `LinearLayout`, `match_parent` made the RecyclerView demand the full parent height *after* the header/search/title rows had already consumed space, pushing its bottom off-screen so the last rows were unreachable. Weighted zero-height gives it exactly the remainder. The new `layout-land/` variant with a horizontal search row and tighter paddings is a sensible landscape treatment.

### But three other things are riding along

**1. `TimeUtils` — six `by lazy` formatters deleted in favour of `formatterFor(...)` calls.**

I checked all six substitutions and they're faithful, including the subtle one: `tzFormatter` had *no* explicit locale, and the replacement correctly passes `locale = null` so `formatterFor` uses `ofPattern(pattern)` rather than `ofPattern(pattern, locale)`. `csvDateFormatter`'s `Locale.US` is preserved too.

**This is a real bug fix, and it isn't mentioned anywhere.** `defaultLocale` is `get() = Locale.getDefault()` — a computed property. Captured once inside a `by lazy`, the formatter froze the locale at first use, so switching the app language mid-session left dates rendering in the *previous* language until the process restarted. `formatterFor` reads the locale per call and makes it part of the cache key, so a language change now produces a correctly-localised formatter. For an app shipping six languages that's worth its own PR and its own line in the changelog — not a silent rider on a layout fix.

**2. `MainApplication` — `catch (t: Throwable) { false }` around both `saveLog*ToRoom` calls.**

Catching `Throwable` is defensible *here* specifically, because these run from the uncaught-exception handler where a secondary failure would mask the original crash. But it's completely silent: if the diagnostics pipeline breaks (Hilt entry point unavailable, Room closed), every crash log is dropped and nothing anywhere says so. At least `Log.w(TAG, "…", t)` before returning false.

Also note `SecurePrefs.warmUp` now has an internal `try`/`catch` **and** an external `runCatching` at the call site — one of the two is redundant.

**3. `row_mysurvey.xml`** — restructuring `submission_count` into a nested `LinearLayout` with the title, and 18sp → 16sp throughout. A visual redesign, reasonable on its own, unrelated to scrolling.

### Why this matters beyond tidiness

The `enormous` label is doing real work here: a reviewer looking at "mysubmissions landscape scroll" has no reason to scrutinise a `Throwable` swallow in the crash-logging path or a rewrite of every date formatter in the app. Those changes each deserve their own review, and the two non-layout ones are the riskiest things in the diff. Splitting into three PRs — layout fix, `TimeUtils` locale fix, `MainApplication` hardening — would get each of them a fair look and let the layout fix merge immediately.

### Also

- **No tests.** The `TimeUtils` change in particular is testable and worth pinning: format a fixed instant, change `Locale.setDefault`, format again, assert the output changed. That test fails on `master` and is the proof the locale bug is fixed.
- **Contention:** #16529 also modifies `TimeUtils` (adding `getFormattedShortDate` built on `formatterFor`) and is at `ready`. Same direction, but this PR deletes the `by lazy` block it sits next to — expect a conflict.
- `tools:context` corrected from `.ui.submission.` to `.ui.submissions.` — good catch, worth keeping wherever these hunks land.

**Verdict:** split it. The layout fix is ready to merge on its own today; the `TimeUtils` locale fix deserves its own PR and a test; the `Throwable` swallow needs logging before it goes anywhere.

---

dogi:
## Review — rating 85/100 (was 68)

The unrelated hardening bundle is gone and this is now purely the landscape fix it was always titled. That is the right call — #16630 has since merged and carries the correct version of that tree on master, so dropping it here avoids a guaranteed conflict.

### The scroll fix is correct

```diff
         <androidx.recyclerview.widget.RecyclerView
             android:id="@+id/rv_mysurvey"
             android:layout_width="match_parent"
-            android:layout_height="match_parent" />
+            android:layout_height="0dp"
+            android:layout_weight="1" />
```

That is the actual bug. As the last child of a vertical `LinearLayout`, `match_parent` asks for the full parent height *without* accounting for the header, search row and column-title row above it, so the RecyclerView extends past the bottom edge and its final rows are unreachable. In portrait the preceding rows are short enough that it mostly goes unnoticed; in landscape they eat most of the screen and the list becomes visibly unscrollable at the end. `0dp` + `layout_weight="1"` makes it take exactly the remaining space.

Worth noting this fixes portrait too, not just landscape — the same overflow was there, just less obvious.

### View-binding parity checks out

The risk with adding a `layout-land` variant is a missing id: view binding generates a nullable field for any id absent from one configuration, and the fragment then NPEs in that orientation only. I compared all three sets:

- portrait declares 12 ids
- the new landscape file declares the same 12
- `SubmissionsFragment` binds 10 of them (`etSearch`, `llSearch`, `rbExam`, `rbSurvey`, `rgSubmission`, `rvMysurvey`, `title`, `tlSearch`, `tvFragmentInfo`, `tvMessage`)

Every bound id exists in both layouts, so there is no orientation-specific null. Good.

The landscape variant's own design choices are sensible — putting the radio group and search field side by side rather than stacked reclaims vertical space, which is the scarce dimension there, and dropping the title row from 18sp to 14sp matches.

Fixing `tools:context=".ui.submission.SubmissionsFragment"` → `.ui.submissions.` is tools-only with no runtime effect, but it repairs IDE preview and navigation.

### One maintenance note

`layout-land/fragment_my_submission.xml` duplicates ~185 lines of the portrait layout, so the two now have to be kept in step by hand — a future id addition needs to land in both or it reintroduces exactly the nullability hazard above. That is inherent to configuration-qualified layouts and I would not restructure for it here; just worth a comment at the top of each file pointing at the other.

Moving to `ready`.

---

Okuro3499 (approved):

dogi (approved):

PR #16619, opened 2026-08-28
all: smoother flexbox library migrating (fixes #16914)

ragilzakaria:
## Summary

Replaces the local binary `app/libs/flexbox-1.0.0.aar` with the official Maven dependency (`com.google.android.flexbox:flexbox:3.0.0`) managed via Gradle version catalog.

## Changes Made

- **Removed Local AAR**: Deleted `app/libs/flexbox-1.0.0.aar`.
- **Dependency Management**: Added `flexbox = "3.0.0"` to `gradle/libs.versions.toml` and updated `app/build.gradle`.
- **Test Hardening**: Added error handling safeguards in `MainApplication.kt` and `DiagnosticsRepositoryImpl.kt` to prevent non-DI Robolectric tests from being interrupted by background SQLite/Room logging operations without SQLite bindings.
- **Coroutines Teardown**: Updated `NotificationActionReceiverTest.kt` to safely restore `MainApplication.applicationScope` during `tearDown()`.

## Verification

- Ran `./gradlew compileDefaultDebugKotlin compileDefaultDebugJavaWithJavac` (Passed)
- Ran `./gradlew testDefaultDebugUnitTest` (All 1,324 unit tests passed)

## Summary by CodeRabbit

* **Chores**
  * Added support for the Flexbox layout library to enable improved interface layout capabilities.


dogi:
## Review — rating 84/100

Removing a 75 KB binary blob from version control in favour of a resolvable Maven coordinate is straightforwardly good, and the jump from a 2017 build to the current release is overdue.

### I went looking for a runtime break and there isn't one

The risk with swapping a vendored AAR for a major-version-newer artifact is a *precompiled* consumer: `app/libs/ChipCloud-3.0.5.aar` is still vendored, `fileTree(include: ['*.jar', '*.aar'], dir: 'libs')` still picks it up, and an AAR in a fileTree carries no POM, so nothing declares what it links against. If ChipCloud had been compiled against flexbox 1.0.0's API, a 3.0.0 swap would surface as `NoSuchMethodError` at chip-render time — invisible to compilation, and invisible to CI since this project runs no instrumented tests.

I unpacked the AAR and checked:

```
$ for f in $(find . -name '*.class'); do strings "$f" | grep -o "com/google/android/flexbox/[A-Za-z$]*"; done | sort -u
(no output)
```

ChipCloud contains **zero** references to flexbox. Its constructor takes a plain `ViewGroup` — the app happens to hand it a `FlexboxLayout` at all four call sites, but the library never links against the type. So there is no precompiled consumer and no hidden ABI coupling. Concern withdrawn.

The app's own flexbox usage is compiled, not vendored — `BaseDashboardFragment` constructs `FlexboxLayout.LayoutParams` directly — so any 1.0.0→3.0.0 API break there is a compile error. Both flavors build green.

Keeping the `fileTree` line is correct, since `ChipCloud-3.0.5.aar` still needs it.

### Worth knowing while you are here

That same unpack turned up something relevant to the companion PR:

```
android/support/v4/content/
android/support/v4/graphics/drawable/
```

ChipCloud still links against the **legacy support library**. That is pre-existing and not this PR's problem, but it is a much sharper argument for #16620 than "tidy up the libs folder" — with Jetifier gone in AGP 9, a vendored AAR referencing `android.support.v4` is living on borrowed time. Worth putting in #16620's description.

### Two notes

Eleven-plus layouts use `FlexboxLayout`, and flexbox 3.0.0 is a rewrite relative to 1.0.0 (AndroidX migration in 2.x, then further cleanup). Nothing in the diff can break at compile time, but wrap/alignment defaults are the kind of thing that shifts quietly across a jump this size. A quick visual pass over the dashboard cards, resource search tags and the health examination screen before merge would be cheap insurance — CI cannot see any of it.

Trivial: the `flexbox = "3.0.0"` entry is appended after the `test*` versions in `[versions]` rather than grouped with the other library versions.

Moving to `ready`.

---

Okuro3499 (dismissed):
dashboard ui breaks items are no longer visible
https://github.com/user-attachments/assets/c07b15d9-e5f5-4ca1-8874-a390eccea346

dogi:
## Review — rating 82/100 → `ready`

Getting a binary AAR out of `app/libs/` and onto a versioned Maven coordinate is worth doing on its own, and the accompanying code changes fix a genuine bug rather than just keeping the compiler happy. CI green on all five checks.

**The `params` → `createChipLayoutParams()` change is the real fix here**, and it's easy to miss among the layout churn. On master a single `FlexboxLayout.LayoutParams` instance was created once by `lazy` and handed to *every* `addView` call across all four dashboard cards — courses, library, teams and my-life chips all shared one object. `FlexboxLayout` writes per-child flex state back into its children's LayoutParams during measure, so one instance shared by N children means every chip's flex values are whatever the last measured child left behind. Returning a fresh instance per chip is correct, and `flexShrink = 0f` is what actually makes the `nowrap` row scroll instead of squeezing every chip down to fit.

**I checked the migration surface before worrying about the major version jump.** Flexbox usage across `app/src/main/java` is only `FlexboxLayout`, `FlexboxLayout.LayoutParams`, `FlexDirection.ROW`, the `flexDirection` property and plain `addView` — in `BaseDashboardFragment`, `AchievementFragment`, `EditAchievementFragment` and the search-tags layout. No `FlexboxLayoutManager`, no `FlexboxItemDecoration`, none of the RecyclerView integration where 1.x→3.x actually changed shape. So 1.0.0 → 3.0.0 touches nothing this app calls, and `minSdk 26` clears 3.0.0's floor comfortably.

Also worth noting for anyone worried about the other AAR: `implementation fileTree(include: ['*.jar', '*.aar'], dir: 'libs')` stays in `app/build.gradle`, so `ChipCloud-3.0.5.aar` keeps loading exactly as before. And ChipCloud carries no `com/google/android/flexbox/` references of its own, so deleting the local flexbox AAR doesn't pull the rug from under it.

### The tablet layouts change behaviour, not just markup

This is the part I'd want confirmed with eyes on a device. The four `layout-sw600dp/home_card_*.xml` files go from:

```xml
<ScrollView>  <!-- vertical -->
    <FlexboxLayout android:layout_width="match_parent" app:flexWrap="wrap" />
```

to a `HorizontalScrollView` with `flexWrap="nowrap"`. On tablets that converts each dashboard card from a **multi-row wrapped grid that scrolled vertically** into a **single row that scrolls sideways** — the phone behaviour. On a 10" tablet the card has real vertical space, and a user with fifteen courses goes from seeing them all at a glance to scrolling horizontally through one strip.

That may well be the intent (consistency across form factors, and it's what the phone layouts already did). But it isn't implied by "migrate flexbox off a local AAR", and it's the largest user-visible change in the diff, so it deserves a line in the description and a look on a tablet.

### Smaller

- `app:alignContent="stretch"` is set on all eight FlexboxLayouts alongside `flexWrap="nowrap"`. `alignContent` only governs how *multiple lines* are distributed on the cross axis, so with `nowrap` there is exactly one line and the attribute does nothing. Harmless, but it reads as though it's doing something.
- The `findViewById<FlexboxLayout>(...)?.flexDirection` safe-call additions turn a crash into a silent no-op if an id ever goes missing from a layout variant. Consistent with `renderMyLibrary` above it, so I'd keep it — just noting that a missing flexbox now fails invisibly rather than loudly.
- The `updateTeamNotifications` rewrite (nested `let` → `?: continue`) is a readability win but unrelated to the migration; it widens the diff for reviewers looking for flexbox changes.

---

Okuro3499 (changes requested):
```
 Problem: Tablet dashboard cards changed from a wrapping multi-row grid (ScrollView + flexWrap="wrap") to a single non-wrapping horizontally-scrolling row (HorizontalScrollView +
    flexWrap="nowrap", plus flexShrink = 0f in BaseDashboardFragment.createChipLayoutParams() forcing every chip to its full fixed width regardless of how many exist). On a tablet  
    (dashboard_chip_width = 150dp) with a realistic number of courses/library items/teams, a user who previously saw most items at a glance in a wrapped grid must now               
    horizontal-scroll through a single long row. This is a real, unreviewed UX regression, not a refactor-neutral change.   
```
https://github.com/user-attachments/assets/acc751ee-e7a7-4f84-aad4-f106c3e1c17e

Okuro3499:
Screenshot_20260904_141653

Okuro3499 (approved):

dogi:
**Review — rating: 88/100**

Three distinct wins bundled here, and the dependency swap is the least interesting of them.

**Getting off the vendored AAR is overdue.** `app/libs/flexbox-1.0.0.aar` was pulled in by the catch-all `implementation fileTree(include: ['*.jar', '*.aar'], dir: 'libs')` at `app/build.gradle:206` — a binary in the repo with no version metadata, no transitive dependencies and no way for Dependabot to see it. Replacing it with `com.google.android.flexbox:flexbox:3.0.0` through the version catalog makes it a tracked dependency like everything else, and jumps 1.0.0 → 3.0.0 (the AndroidX-migrated line) in the process. Both flavors build and the tests pass, so the `FlexboxLayout` / `FlexDirection` API surface this app uses came through the major bump unchanged.

**The shared `LayoutParams` was a real bug.** `params` was a single `by lazy` instance handed to every `addView` call across courses, library, teams and My Life. `ViewGroup.LayoutParams` are per-child state that the parent writes into during measure — `FlexboxLayout` in particular stores computed main/cross sizes on them — so every chip on the dashboard was sharing one mutable object. `createChipLayoutParams()` gives each child its own, which is the only correct arrangement. `flexShrink = 0f` then pins the chip width instead of letting the layout squeeze items.

**The XML change is the one that fixes what users see.** `flexWrap="wrap"` inside a `HorizontalScrollView` is a contradiction: the scroll view gives its child an unbounded width, so there's no boundary to wrap against and the wrap setting mostly produces surprises. `nowrap` plus an explicit `flexDirection="row"` states what these rows actually are, and `fillViewport="true"` gives the flexbox the viewport's height so the existing `alignItems="stretch"` has something to stretch to.

**Nits:** with the AAR gone, `app/libs/` is empty, so the `fileTree(... dir: 'libs')` line at `build.gradle:206` is now dead and can go with it. And now that the layouts declare `app:flexDirection="row"`, the three `findViewById<FlexboxLayout>(...)?.flexDirection = FlexDirection.ROW` assignments in `onViewCreated` are redundant — the null-safe calls make them harmless, but they're setting what the XML already says.

**Verdict:** mergeable.

---

dogi:
**Promoted to `merge`** — Okuro3499 approved `dad70e78`, which is still the head, and my review stands at 88/100 with no changes since.

I re-checked the tablet regression from the earlier changes-requested, since that was the substantive objection. It's resolved in the cleanest possible way: `layout-sw600dp/home_card_*.xml` is **byte-identical to master** — still `ScrollView` + `flexWrap="wrap"`, so tablets keep the wrapping multi-row grid. Only the phone `layout/` variant moves to `HorizontalScrollView` + `flexDirection="row"` + `flexWrap="nowrap"`, which is where `wrap` was contradictory to begin with. The fix is scoped to the layout that was actually wrong.

---

dogi (approved):

PR #16622, opened 2026-08-28
dashboard: smoother back press navigating (fixes #16611)

### LINKED ISSUE #16611: Improvement: Single child fragment back press triggers logout prompt instead of popping backstack
In DashboardActivity.kt, the custom back press callback checks supportFragmentManager.backStackEntryCount > 1 before calling FragmentNavigator.popBackStack(). When a user navigates from the home dashboard into a single child fragment (where backStackEntryCount equals 1), pressing the Android system back button skips popping the backstack and instead shows the "Press back again to logout" prompt. Back navigation should check backStackEntryCount > 0 (or verify the active fragment) to return to the dashboard before offering the logout prompt.

ragilzakaria:
## Summary

Fixes the back-press handling in `DashboardActivity` so that pressing the system back button while in a single child fragment properly pops the backstack and navigates back to the root dashboard instead of skipping to the "Press back again to logout" prompt.

## Cause of Issue

In `DashboardActivity.kt`, `addBackPressCallback` checked `supportFragmentManager.backStackEntryCount > 1` before invoking `FragmentNavigator.popBackStack()`. When a user navigated into a single child fragment (where `backStackEntryCount` is 1), the condition `1 > 1` failed, causing the app to prompt for logout rather than popping the backstack.

## Solution

- Added `isAtRootDashboard()` helper in `DashboardActivity` to determine if the currently displayed fragment is the root dashboard (`BellDashboardFragment` or `InactiveDashboardFragment`).
- Updated the back-press callback to check `supportFragmentManager.backStackEntryCount > 0 && !isAtRootDashboard() -> FragmentNavigator.popBackStack(supportFragmentManager)`.
- Added unit test suite in `FragmentNavigatorTest.kt` to ensure backstack replacement, additions, and popping operations behave as expected.

## Verification

- Ran `./gradlew compileDefaultDebugKotlin compileDefaultDebugJavaWithJavac` (Passed)
- Ran `./gradlew testDefaultDebugUnitTest` (All 1,328 unit tests passed)

fixes #16611 

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved dashboard back navigation so returning from nested screens behaves more reliably.
  - Prevented startup, secure-preference initialization, diagnostics logging, and database connection failures from unexpectedly interrupting the app.
  - Improved resilience when system time services are unavailable by using safe fallback values.
  - Standardized CSV date formatting across locales and time zones for more consistent exports.
- **Tests**
  - Added coverage for fragment navigation and back-stack behavior.


dogi:
## Review — rating 62/100

The back-press fix is correct and worth having. It is buried in an unrelated hardening bundle that I would not merge as-is, and which duplicates changes already sitting in #16618 and #16620.

### The back-press fix is right

Master used the back-stack depth as a proxy for "am I at the root":

```kotlin
supportFragmentManager.backStackEntryCount > 1 -> FragmentNavigator.popBackStack(...)
else -> promptLogout()
```

With exactly one entry — dashboard, then one feature fragment pushed — back press falls through to `promptLogout()` instead of navigating back. That is #16611, and it is a genuinely bad user-facing bug: back on a normal screen offers to log you out.

Replacing the depth heuristic with an explicit root check is the correct fix rather than an off-by-one tweak, because the `> 1` only worked when the root fragment was itself on the back stack. `isAtRootDashboard()` reads `R.id.fragment_container`, which is the same container used at lines 321, 638 and 725, so it is looking at the right thing. I walked the transitions (root → A → B → back → back → back) and it behaves correctly at each step.

One gap it does not close: a fragment shown with `addToBackStack = false` leaves `backStackEntryCount == 0`, so back press there still prompts logout. Same as master, so not a regression — worth a follow-up issue.

### The bundled changes are the problem

Six of the eight changed files have nothing to do with back-press navigation, and this is the third PR carrying this same hardening tree (#16618 and #16620 both have versions of it). Whichever lands first will leave the others conflicting on the same lines.

**`catch (Throwable)` in suspend functions swallows cancellation.** `saveLogToRoom` and `saveLogsToRoom` are `suspend` and now wrap their body in `catch (e: Throwable) { false }`. `CancellationException` is a `Throwable`, so a cancelled coroutine is converted into a normal `false` return and the coroutine completes instead of cancelling — structured concurrency is broken for every caller. The same applies to the bare `catch (_: Throwable) {}` blocks in `createLog` and `persistCriticalLog`. If the intent is "never let logging crash the app", the catch needs to rethrow cancellation:

```kotlin
    throw e
} catch (e: Throwable) {
    false
}
```

**`initializeDatabaseConnection` swallowing everything is the same objection I raised on #16620.** If `appDatabase.openHelper.writableDatabase` throws, the app now continues to run against a database it could not open, and every downstream failure gets reported as some unrelated symptom. A failure to open the DB should be visible.

**`TimeUtils.formatDateForCsv` is an unflagged behaviour change.** I checked and the output is identical — `formatterFor` builds `ofPattern(pattern, locale).withZone(zone)`, exactly matching the old `csvDateFormatter`. But master's `by lazy` froze `ZoneId.systemDefault()` at first use, whereas this re-reads it per call, so CSV timestamps now follow a runtime timezone change. That is arguably the better behaviour, which is why it deserves to be its own PR with that stated, not a silent line inside a navigation fix. Also, `csvDateFormatter` is left behind and is now dead — remove it.

### The new test does not test the change

`FragmentNavigatorTest` exercises `FragmentNavigator.replaceFragment`/`popBackStack`, which already worked and which this PR does not touch. Nothing covers `isAtRootDashboard()` or the `handleOnBackPressed` branch selection — the actual change. Since `isAtRootDashboard()` is a pure function of the current fragment, a Robolectric test that puts a `BellDashboardFragment` and then a plain fragment in the container and asserts which branch runs would be straightforward, and would be worth more than all four navigator tests.

### What I would do

Split this. A PR containing only the `DashboardActivity` change plus a test for it is an easy `ready` from me — it fixes a real bug in about ten lines. The `catch (Throwable)` hardening should be one PR (with cancellation rethrown, and `initializeDatabaseConnection` left alone or handled visibly), reconciled against #16618 and #16620 so the three stop overlapping. The `TimeUtils` change should be its own, with the dead field removed.

Moving to `change`.

---

dogi:
## Review — rating 85/100 (was 62)

Both asks from the last pass are done: the unrelated hardening tree and the `TimeUtils` change are gone (correctly — #16630 has since merged and carries that work properly on master), and the back-press logic now has a real test.

Extracting `isRootDashboardFragment` to a companion function is the move that made testing possible:

```kotlin
fun isRootDashboardFragment(fragment: Fragment?): Boolean {
    return fragment == null || fragment is BellDashboardFragment || fragment is InactiveDashboardFragment
}
```

`DashboardActivity` needs Hilt and a full inflation to construct, so a predicate hanging off the instance would have stayed untestable. As a static it takes a `Fragment?` and `isRootDashboardFragment_identifiesRootFragmentsCorrectly` exercises all four cases directly — null, both root types, and a feature fragment. That test fails if someone adds a third dashboard root and forgets the predicate, which is exactly the regression worth guarding.

The fix itself is unchanged and still right: `backStackEntryCount > 1` used depth as a proxy for "am I at the root", so a single pushed feature fragment fell through to `promptLogout()` — back on an ordinary screen offering to log you out.

### One limitation worth naming

The two navigation tests re-derive the branch condition rather than invoking it:

```kotlin
val shouldPop = activity.supportFragmentManager.backStackEntryCount > 0 &&
    !DashboardActivity.isRootDashboardFragment(currentFrag)
assertTrue(shouldPop)
```

That is a copy of the production expression, so editing `handleOnBackPressed` would not fail these tests. They still earn their place — they prove the predicate behaves correctly against a *real* `FragmentManager` stack across push and pop, which the unit test alone does not — but the `when` branch itself remains uncovered. Given what it takes to stand up `DashboardActivity`, I think that is the right trade; just worth a comment in the test file so nobody reads it as end-to-end coverage.

`FragmentNavigatorTest` is fine now that it is supplementary rather than standing in for a test of the change.

### Still open, and unchanged from last time

A fragment shown with `addToBackStack = false` leaves `backStackEntryCount == 0`, so back there still prompts logout. Same as master, so not a regression — worth its own issue.

And the interaction with **#16626** stands: that PR rewrites `openCallFragment` so tab switches restack rather than pop, which would make this condition walk back through duplicate entries. #16626 is currently `change` for that reason; whichever lands second should re-check this branch.

Moving to `ready`.

---

Okuro3499 (approved):

dogi (approved):

PR #16623, opened 2026-08-28
teams: smoother interactive team tasks status and board handling (fixes #16616)

### LINKED ISSUE #16616: Improvement: Interactive Kanban/checklist board for Team Tasks
Team tasks are currently displayed in a basic list. Upgrading the Team Tasks view into an interactive checklist/board with clear status states (To Do, In Progress, Completed), due dates, and member assignment badges will improve team project coordination.

ragilzakaria:
## Summary
Enhances Team Tasks with interactive status badges (**To Do**, **In Progress**, **Completed**), board category filters, and direct status toggling.

## Changes
- Added dedicated task status filtering toggles (`btn_all`, `btn_todo`, `btn_in_progress`, `btn_completed`, `btn_my`) in [fragment_teams_tasks.xml](file:///Users/ragilzakaria/myplanet/app/src/main/res/layout/fragment_teams_tasks.xml).
- Added interactive colored status badge chip (`status_badge`) in [row_task.xml](file:///Users/ragilzakaria/myplanet/app/src/main/res/layout/row_task.xml).
- Implemented `setTaskStatus` in [TeamsRepository](file:///Users/ragilzakaria/myplanet/app/src/main/java/org/ole/planet/myplanet/repository/TeamsRepository.kt), [TeamsRepositoryImpl](file:///Users/ragilzakaria/myplanet/app/src/main/java/org/ole/planet/myplanet/repository/TeamsRepositoryImpl.kt), and [TeamsTasksViewModel](file:///Users/ragilzakaria/myplanet/app/src/main/java/org/ole/planet/myplanet/ui/teams/tasks/TeamsTasksViewModel.kt).
- Updated [TeamsTasksAdapter](file:///Users/ragilzakaria/myplanet/app/src/main/java/org/ole/planet/myplanet/ui/teams/tasks/TeamsTasksAdapter.kt) and [TeamsTasksFragment](file:///Users/ragilzakaria/myplanet/app/src/main/java/org/ole/planet/myplanet/ui/teams/tasks/TeamsTasksFragment.kt) to support clicking status badges to cycle task states (To Do ➔ In Progress ➔ Completed).
- Added unit tests in [TeamsTasksViewModelTest](file:///Users/ragilzakaria/myplanet/app/src/test/java/org/ole/planet/myplanet/ui/teams/tasks/TeamsTasksViewModelTest.kt).

## Fixes
fixes #16616 

## Testing

## Summary by CodeRabbit

* **New Features**
  * Added task statuses: To Do, In Progress, and Completed.
  * Added color-coded status badges to task rows.
  * Users can advance task statuses directly from the task list.
  * Added filters for each status, including improved “All” and “My Tasks” views.
* **Improvements**
  * Task completion and status updates now remain synchronized.
  * Added localized “To Do” labels in supported languages.
* **Tests**
  * Added coverage for status transitions, filtering, serialization, and updates.


dogi:
## Review — rating 60/100

The feature work shows real care in the places people usually miss — `bindingAdapterPosition` resolved at click time rather than captured at bind, `status` added to both the `DiffUtil` comparator and `TaskSnapshot` so the change actually redraws and is not swallowed by the snapshot short-circuit. But there is a data-shape bug that will hide existing tasks, and the PR again carries the unrelated hardening bundle.

### Existing tasks will vanish from the new tabs

`TeamTask.status` is populated from the server as:

```kotlin
task.status = JsonUtils.getString("status", obj)
```

`JsonUtils.getString` returns **`""`** for a missing field, not `null`. So every task synced from a server that does not send `status` — which is every task that exists today — carries `status = ""`.

Now trace the filters:

```kotlin
todoTasks:       !completed && (status == null || status == "to_do")   // "" → no
inProgressTasks: !completed && status == "in_progress"                  // "" → no
completedTasks:  completed || status == "completed"                     // "" → no
```

An incomplete task with `status = ""` matches **none** of the three new tabs. It is visible only under "All" (and "My Tasks" if assigned). Since "To Do" is the tab users will treat as the default working view, the practical effect is that pre-existing tasks look deleted.

The fix is to treat blank as to-do, and it wants to be one predicate rather than three literals:

```kotlin
private fun TeamTask.effectiveStatus(): String = when {
    completed || status == "completed" -> "completed"
    status == "in_progress" -> "in_progress"
    else -> "to_do"                     // covers null, "", and "to_do"
}
```

That also fixes the adapter, which has the same three-way branch inline, and the badge cycle.

### The status strings should not be literals

`"to_do"`, `"in_progress"`, `"completed"` appear as bare strings across `TeamsRepositoryImpl`, `TeamsTasksAdapter`, and `TeamsTasksFragment`. These values are persisted and round-trip through CouchDB, so a typo in any one of them silently misfiles tasks with no compile error. An enum with a serialized-name mapping, or at minimum shared constants, is worth it here.

### Missing translations

`status_todo` is new and is added only to `values/strings.xml`. Per the project convention it needs an entry in each of `values-ar`, `values-es`, `values-fr`, `values-ne`, `values-so`. (`status_in_progress` and `status_completed` already existed, so only the one string is affected.)

### The hardening bundle again

This is the fourth PR carrying the same unrelated tree — #16618, #16620, #16622 and now this one all modify `MainApplication`'s `catch (Throwable)` blocks, `SecurePrefs.warmUp`, `DiagnosticsRepositoryImpl`'s `Exception` → `Throwable`, `TimeUtils.formatDateForCsv`, and the `NotificationActionReceiverTest` scope save/restore. They will conflict with each other on identical lines.

The same two objections apply as on #16622: `saveLogToRoom`/`saveLogsToRoom` are `suspend` and `catch (e: Throwable) { false }` swallows `CancellationException`, converting a cancelled coroutine into a normal `false` return and breaking structured concurrency; and `initializeDatabaseConnection` swallowing everything means the app runs on a database it could not open. Please land that tree once, in its own PR, with cancellation rethrown.

### Testing

`setTaskStatus calls repository` covers the trivial passthrough. Nothing covers the badge's cycle order, the three filter predicates, or the `setTaskCompletion` change that now also writes `status` — which is where the `""` bug lives. The filters are pure functions over a `List<TeamTask>`; they are cheap to test and would have caught this.

Minor: the badge remains visually identical for `nonTeamMember`, who cannot click it — worth hiding or dimming.

Moving to `change`.

---

dogi:
## Review — rating 88/100 (was 60)

Every finding from the last pass is addressed, and the data bug is fixed at the root rather than patched at each call site.

### The `status == ""` gap is closed

```kotlin
fun TeamTask.effectiveStatus(): String = when {
    completed || status == TaskStatus.COMPLETED.value -> TaskStatus.COMPLETED.value
    status == TaskStatus.IN_PROGRESS.value -> TaskStatus.IN_PROGRESS.value
    else -> TaskStatus.TODO.value          // null, "", and "to_do" all land here
}
```

That is the fix. Since `TeamTask.createTask` populates `status` via `JsonUtils.getString`, which returns `""` for an absent field, every already-synced task carried `""` and matched none of the three tab predicates — invisible everywhere except "All". Routing all five filters through `effectiveStatus()` means those tasks now show under To Do.

And it is properly pinned. `tab filtering correctly categorizes all tasks including legacy empty status` puts `task1LegacyTodo` (`status = ""`) in the fixture and asserts it appears in `filterTodoTasks`. That test fails against the previous revision, which is exactly what a regression guard for this should do. The `effectiveStatus` unit test covers all seven combinations including the two where the `completed` flag disagrees with `status`.

Hoisting the filters into a companion object is what made that testable — they were private instance methods depending on `user?.id` before.

The `TaskStatus` enum replaces the bare `"to_do"`/`"in_progress"`/`"completed"` literals, so a typo in one of the five places these values are compared is now a compile error rather than a silently misfiled task. `nextStatus()` centralises the cycle, and `nextStatus cycles in order` covers it including the empty-status entry point.

All five translations are present (ar, es, fr, ne, so) alongside the default.

Two smaller things I noted last time also handled: `binding.statusBadge.alpha = if (nonTeamMember) 0.6f else 1.0f` gives the non-clickable badge a visual affordance, and the `DiffUtil` comparator now compares `effectiveStatus()` rather than raw `status` — so a row whose status is `null` in one snapshot and `""` in the next no longer rebinds pointlessly.

The unrelated hardening tree is gone, which is right: #16630 merged and carries that work correctly on master.

### Two minor notes

`TaskStatus.fromValue()` appears to have no callers — `effectiveStatus()` compares `.value` strings directly. Either wire it in (it would make `effectiveStatus` read more cleanly as a `when` over the enum) or drop it before it becomes stale.

The repository tests construct `TeamsRepositoryImpl` with eighteen mocked constructor arguments, so adding a dependency to that class breaks two tests that do not care about it. Unavoidable without Hilt test infrastructure, and the tests earn their keep by capturing the upserted `TeamTask` and asserting `completed`/`status`/`isUpdated` stay consistent — just be aware of the maintenance cost.

Naming nit: an enum constant called `TODO` sits close to Kotlin's `TODO()`. It resolves correctly here, but `TO_DO` would read less ambiguously.

Moving to `ready`.

---

Okuro3499 (approved):
https://github.com/user-attachments/assets/3bd2e6eb-cc3f-491a-8737-f8242eae6938

Okuro3499 (changes requested):
```
 File: app/src/main/java/org/ole/planet/myplanet/model/TeamTask.kt, function serialize(task, user)                                                                                
    Problem: The outgoing JSON object never includes task.status, so a task moved to "In Progress" (or any status) is never persisted server-side. Combined with                     
    bulkInsertTasksFromSync's unconditional overwrite of local rows from server docs, this state can silently revert on the next sync, breaking the feature across devices/reinstalls
    — the same failure mode RoomModule's drop-and-resync strategy already warns about for unsynced writes.
```

PR #16624, opened 2026-08-28
search: smoother offline full text search matching (fixes #16615)

### LINKED ISSUE #16615: Improvement: Full-text offline search across downloaded resources and course step transcripts
Finding specific topics offline is currently limited to exact title and tag matches. Implementing Room SQLite Full-Text Search (FTS5) for downloaded resource titles, descriptions, and course step transcripts will allow learners to perform rapid offline keyword searches even with spelling variations or partial terms.

ragilzakaria:
## Summary
Improves offline search capabilities by implementing full-text search across titles, descriptions, transcripts, authors, publishers, tags, and course step contents.

## Changes
- Enhanced [ResourcesSearchUtils](file:///Users/ragilzakaria/myplanet/app/src/main/java/org/ole/planet/myplanet/utils/ResourcesSearchUtils.kt) with multi-field full-text search matching and prioritized match ranking (Title > Author/Publisher > Tags > Description/Transcripts).
- Added `searchCourseSteps` utility for searching within course steps.
- Updated [CoursesRepositoryImpl](file:///Users/ragilzakaria/myplanet/app/src/main/java/org/ole/planet/myplanet/repository/CoursesRepositoryImpl.kt) to perform offline full-text search across course titles, descriptions, and step transcripts.
- Added unit tests in [ResourcesSearchUtilsTest](file:///Users/ragilzakaria/myplanet/app/src/test/java/org/ole/planet/myplanet/utils/ResourcesSearchUtilsTest.kt).

## Fixes
fixes #16615 

## Testing
- Executed `./gradlew testDefaultDebugUnitTest` - all 1,325 unit tests pass.

## Summary by CodeRabbit

* **New Features**
  * Improved course search ranking, prioritizing title-prefix matches, followed by title and content matches.
  * Course searches now include descriptions and step content, with support for multi-word and diacritic-normalized queries.
  * Resource searches now include metadata such as descriptions, authors, publishers, tags, and subjects.
  * Search results are sorted alphabetically within each relevance group, while blank searches preserve the original result order.


dogi:
## Review — rating 64/100

The search work is sound in design — the three-tier ranking (title prefix → title contains → content contains) preserves the prefix-first behaviour from #16608 rather than flattening everything into one relevance bucket, which is the mistake this kind of change usually makes.

Two things I checked that came out fine: `MyCourse.courseSteps` is `@Ignore`, so I expected the step search to be dead, but `mapCourses` populates it from `courseStepDao.getByCourseIds(courseIds)` before `filterCourses` runs, so the step text really is available there. Worth confirming the same for `searchCourses` — its `allCourses` argument only carries steps if the caller passed a `mapCourses` result, and if any caller passes raw DAO output the step matching silently never fires.

### `searchCourseSteps` is dead code

```
$ grep -rn "searchCourseSteps" app/src/main/java/
NO CALLERS IN MAIN
```

It is added, tested, and never called. Either wire it into the course-step UI this PR is meant to serve, or drop it — an unused public helper with passing tests reads as covered functionality and will mislead the next person.

### The hardening bundle, for the fifth time

#16618, #16620, #16622, #16623 and now this one all carry the same unrelated tree: `MainApplication`'s `catch (Throwable)` blocks, `SecurePrefs.warmUp`, `DiagnosticsRepositoryImpl`'s `Exception` → `Throwable`, `TimeUtils.formatDateForCsv`, and the `NotificationActionReceiverTest` scope save/restore. Five PRs editing identical lines will conflict with each other whichever order they land in.

Same two objections as before: `saveLogToRoom`/`saveLogsToRoom` are `suspend`, so `catch (e: Throwable) { false }` swallows `CancellationException` and breaks structured concurrency for callers; and `initializeDatabaseConnection` swallowing everything lets the app run against a database it failed to open. Please land that tree once, on its own, with cancellation rethrown.

### Cost per keystroke

`filterCourses` now runs `contains` across the description and every step of every course, and `searchCourses` builds a fresh `"$title $description $stepsText"` per course. Both run on each keystroke against the full course list, and the joined string is allocated and discarded every time. `searchLocalModels` does the same with `tags.joinToString(" ")`.

The existing `courseTitleNormal` / `titleNormal` fields exist precisely to avoid re-normalising per query — the new secondary fields have no equivalent. Precomputing a normalized content blob when courses are mapped, rather than per search, would keep this from degrading as libraries grow. Not blocking, but this is the direction the code was already moving and this PR moves against it.

### Smaller items

The two `searchList` overloads are ambiguous-looking: the four-parameter version has a default for `secondarySelectors`, so it also matches a three-argument call. Kotlin resolves it to the three-parameter overload (candidates not using defaults win), so it compiles — but the three-parameter version is now redundant and the pair invites confusion. Delete it and let the default do the work.

Test nits: `mockk(relaxed = true) { every { … } returns …; every { … } returns … }` with semicolon-chained statements on one line is hard to read, and `testSearchCourseSteps` uses fully-qualified `org.ole.planet.myplanet.model.CourseStep` instead of an import.

Moving to `change`.

---

dogi:
## Review — rating 86/100 (was 64)

All three findings addressed, and the performance one was fixed better than I suggested.

**The dead `searchCourseSteps` is gone**, along with the redundant three-argument `searchList` overload — only the four-parameter version with a defaulted `secondarySelectors` remains, so the resolution ambiguity goes with it. The unrelated hardening tree is gone too, which is right now that #16630 has merged.

**The per-item allocation is gone.** I had suggested precomputing a normalized content blob; short-circuiting is the better answer:

```kotlin
val matchesSecondary = normalizedQueryParts.all { part ->
    primary.contains(part) || secondarySelectors.any { selector ->
        val value = selector(item)
        value != null && Utilities.normalizeText(value).contains(part)
    }
}
```

No `"$title $description $steps"` string is built, and `any {}` exits at the first selector that hits. The `secondarySelectors.isNotEmpty()` guard also skips tier three entirely for callers that pass none. Same for the course path, which now uses `||` over description and steps rather than joining them.

The three-tier ranking (title prefix → title contains → content contains) is preserved rather than flattened, which is the thing that would have quietly undone #16608's prefix-first ordering.

### The tests are discriminating

`testSearchListRanking` is well constructed — "Physics Advanced" (prefix), "Advanced Physics" (title contains), "Quantum Computing" with `body` containing "physics" (content only) — and asserts the exact order `1, 2, 3`. It fails if any tier is merged or reordered.

`search ranks prefix matches first, then title contains, then content contains` does the same through the real repository, and `filterCourses matches title, description, and step content` covers all three match sources in one assertion set. Both fail on master, where only title matched.

**The two search paths normalize differently.** `search()` runs everything through `Utilities.normalizeText` (accent folding), while `filterCourses` uses `contains(searchText, ignoreCase = true)` with no normalization. So searching "matemática" behaves differently depending on which entry point the screen uses. That asymmetry predates this PR for titles, but the PR extends it to description and step content. Worth aligning at some point — probably by routing `filterCourses` through the same normalized comparison.

**The tag selector still allocates.** `{ it.tags.joinToString(" ") { tag -> tag.name.orEmpty() } }` builds a string per item, though only for items that already failed both title tiers. That is a much smaller set than before, so it is fine in practice; a `tags.any { it.name?.contains(part) }` selector would remove it entirely if this ever shows up in a profile.

Moving to `ready`.

---

Okuro3499 (changes requested):
```
CoursesRepositoryImpl.search() (line 250) is dead code — the actual PR goal doesn't reach users. CoursesRepository.search() has zero callers in app/src/main/java (verified by  
search); the only path wired to the courses UI is filterCourses() (CoursesFragment.kt:241,626 → CoursesViewModel.filterCourses → CoursesRepositoryImpl.filterCourses, line 293).
All the "smoother"/normalized (diacritic-insensitive, accent-insensitive, prefix-ranked) matching logic and its test coverage went into search(), which nothing calls.          
Meanwhile filterCourses() — the real path — only gained the description/step contains(..., ignoreCase = true) additions, with no normalization, so accented/diacritic queries   
still won't match in the live course search, despite that being the headline claim of commit 6a20f6acd. Fix: either port the normalized/ranked logic from search() into         
filterCourses() (and delete the now-pointless duplicate search()/matchesAllParts), or wire filterCourses to call search()'s logic directly. As-is, this branch ships a          
well-tested method nobody runs. 
```

dogi:
## Review — rating 74/100 → `change`

Two good instincts here. **Deleting `CoursesRepository.search()` is a clean win** — I grepped `app/src/main` and it had zero production callers, so the only thing keeping it alive was its own test. Folding the ranking into `filterCourses` means one code path instead of two that had drifted apart (the old `filterCourses` did a plain `contains` with no relevance ordering while `search()` did prefix-then-contains). And extending matching to descriptions, step titles, authors and tags is the right call for an offline library where the title alone is often not what the learner remembers.

The tests are good too — the diacritics case (`"algébra"` finding `"Algebra I"`) and the three-tier ranking assertion are exactly what this needs. CI green on all six checks.

Three things to fix.

### Search results lost their alphabetical ordering

Before, every result came back through:

```kotlin
.sortedWith(compareBy(String.CASE_INSENSITIVE_ORDER) { it.courseTitle ?: "" })
```

Now that sort only runs on the **empty-query** path. With a query, the return is `startsWithQuery + titleContainsQuery + contentContainsQuery`, and within each bucket the order is whatever `courseDao.getAll()` happened to return — unordered SQLite row order. So the three tiers are ranked but their contents are arbitrary, and two searches for the same term can come back differently after a sync rewrites rows. Sorting each bucket before concatenating keeps the new relevance tiers and restores predictability.

### `normalizeText` runs inside the per-query-part loop

```kotlin
val matchesContent = normalizedQueryParts.all { part ->
    title.contains(part) ||
        (description != null && description.contains(part)) ||
        (steps != null && steps.any { step ->
            step.stepTitle?.let { Utilities.normalizeText(it) }?.contains(part) == true ||
                step.description?.let { Utilities.normalizeText(it) }?.contains(part) == true
        })
}
```

`description` is hoisted correctly, but every step's title and description is re-normalized **once per query part**. A two-word query against a course with 20 steps is 80 `normalizeText` calls for that one course, and this runs across every course on each debounced keystroke.

`ResourcesSearchUtils.searchList` has the same shape, and worse:

```kotlin
val matchesSecondary = normalizedQueryParts.all { part ->
    primary.contains(part) || secondarySelectors.any { selector ->
        val value = selector(item)
        value != null && Utilities.normalizeText(value).contains(part)
    }
}
```

One of those selectors is `{ it.tags.joinToString(" ") { tag -> tag.name.orEmpty() } }` — so a string join *and* a normalize per query part per item. Normalizing the secondary values once per item, before the `all {}`, fixes both.

### `searchList`'s contract changed silently

The old signature carried a comment that is now deleted:

> Requires normalizedTitleSelector to return a pre-normalized (and lowercased) string

The new `primarySelector` normalizes internally. That's the better design, but the one caller still passes `it.library.titleNormal ?: ...normalizeText(...)` — so `titleNormal`, already normalized at write time, gets normalized a second time on every search. Either drop the normalization from the selector or document that the parameter is now raw text. (It's only reachable through `searchLocalModels`, so the signature change itself is safe — that's the sole production caller.)

### Two small things in the tests

- In `filterCourses matches title, description, and step content with diacritic normalization`, the comment on `c2` says `// Title match`, but `"math overview"` doesn't contain `"algebra"` — it reaches the list through its *description*, i.e. the content bucket, same as `c3`. The assertion passes because both land in `contentContainsQuery` in DAO order. Since these comments are the documentation of intended ranking, it's worth getting them right.
- `course3.apply { courseSteps = mutableListOf(...) }` is dead setup: `mapCourses` unconditionally overwrites `courseSteps` from `courseStepDao.getByCourseIds`, which the test also stubs with equivalent data. It reads as though the in-memory field is what's under test when it's actually the DAO path.

Sort the buckets and hoist the normalization and I'd move this to `ready`.

---

dogi:
## Re-review — rating 68/100 → `change`

All three points from my last review are fixed, and fixed well:

- **Bucket-internal ordering restored.** `titleComparator` is applied to all three buckets in `filterCourses` and `comparator` to all three in `searchList`, with two new tests (`filterCourses sorts results alphabetically within each relevance bucket`, `testSearchListSortsAlphabeticallyWithinRelevanceBuckets`) pinning it.
- **Normalization hoisted.** `stepTexts` is now built once per course via `flatMap` and `secondaryValues` once per item via `mapNotNull`, both outside the `all { }`. That removes the per-query-part re-normalization, including the tag `joinToString`.
- **Contract documented and honoured.** The KDoc states the selectors take raw or pre-normalized text, and `searchLocalModels`' primary selector now passes the raw title instead of `titleNormal`, so nothing is normalized twice.

The test-comment correction is careful work too — the diacritics test now labels `c2` and `c3` as content matches and reorders the expectation to `c1, c3, c2` because "Advanced Calculus" sorts before "Math Overview" within the content bucket. That's the assertion following the code rather than the other way round.

Unfortunately the sort fix went one step too far and broke a different feature.

### The empty-query branch now re-sorts, overriding the user's chosen order

```kotlin
val comparator = compareBy<T, String>(String.CASE_INSENSITIVE_ORDER) { primarySelector(it).orEmpty() }
if (trimmedQuery.isEmpty()) return list.sortedWith(comparator)
```

Master returned `list` untouched here. The caller is:

```kotlin
// ResourcesFragment.kt:863
private fun filterLocalLibraryByTag(models: List<ResourceListModel>, s: String, tags: List<TagEntity>): List<ResourceListModel> {
    var filteredList = ResourcesSearchUtils.searchLocalModels(models, s)
```

`filterLocalLibraryByTag` runs on **every** filter pass, with `s` being whatever is in the search box — empty in the normal case. And `ResourcesViewModel` owns a user-facing sort: `toggleSortOrder` (DATE) and `toggleTitleSortOrder` (TITLE), each with an ascending flag, driven by the *order by date* / *order by title* buttons in the bottom sheet. Those handlers sort `allResourceModels` and then call `applyFiltersAndUpdateUI`, which routes straight back through `filterLocalLibraryByTag`.

So with an empty search box — the state the screen is in almost all the time — tapping **order by date** now sorts the list and then has it immediately re-sorted alphabetically by title before it reaches the adapter. The date sort, and the descending direction of the title sort, both become no-ops.

Your own new assertion locks this in:

```kotlin
val resultEmpty = ResourcesSearchUtils.searchLocalModels(models, "")
assertEquals("Apple Juice", resultEmpty[0].item.title)
```

The fix is narrow: restore `if (trimmedQuery.isEmpty()) return list` in `searchList`. The relevance-bucket sorting you added for the *non-empty* path is the valuable part and should stay — it's only the empty-query path that has a caller with its own opinion about order.

Note the asymmetry with `filterCourses`: there the empty-query path already sorted alphabetically before this PR, so keeping `filteredCandidates.sortedWith(titleComparator)` is correct and unchanged behaviour. It's only `searchList` that acquired a new sort its caller didn't ask for.

Everything else here is ready to go.

---

dogi:
**Review — rating: 78/100**

The feature is worth having and the three-tier ranking is a sensible model: prefix match on the title, then all query parts in the title, then all query parts across title plus secondary fields, each tier sorted alphabetically. Extending it to description, author, publisher, tags and subject for resources, and to description plus step titles/descriptions for courses, is a real improvement over title-only matching. I checked `mapCourses` populates `courseSteps` from `courseStepDao.getByCourseIds` before `filterCourses` runs, so the step-content branch isn't dead code — good, because an `@Ignore` field would have been an easy way to ship half a feature. Deleting the now-unused `CoursesRepository.search` is a clean removal.

**The problem is that `searchList` now throws away the precomputed normalization.** The old signature was explicit about its contract:

```kotlin
// Requires normalizedTitleSelector to return a pre-normalized (and lowercased) string
fun <T> searchList(list: List<T>, query: String, normalizedTitleSelector: (T) -> String?)
```

and `searchLocalModels` honoured it by passing `it.library.titleNormal` — a stored, already-normalized column. The new version takes a raw selector and normalizes inside the loop:

```kotlin
val primary = primarySelector(item)?.let { Utilities.normalizeText(it) }.orEmpty()
```

with `searchLocalModels` now passing `it.item.title` raw. So `titleNormal` is no longer consulted at all, and `normalizeText` runs once per item per keystroke on incremental search, plus once per secondary field for every item that missed on title — description, author, publisher, a `joinToString` over tags, and a `joinToString` over subjects, all built fresh each time. On a large library that's a lot of transient string work in the search path, which is the one place a "smoother search" PR shouldn't add any.

The root cause is that `primarySelector` is doing two jobs: it supplies the text to *match* (which wants the normalized form) and the text to *sort* by (which wants the display form, for `String.CASE_INSENSITIVE_ORDER`). Those can't both be one selector without normalizing on the fly. Splitting them fixes it cleanly:

```kotlin
fun <T> searchList(
    list: List<T>,
    query: String,
    normalizedPrimarySelector: (T) -> String?,   // pre-normalized, e.g. library.titleNormal
    displaySelector: (T) -> String? = normalizedPrimarySelector,
    secondarySelectors: List<(T) -> String?> = emptyList()
)
```

`filterCourses` already gets this right — `item.courseTitleNormal ?: item.courseTitle?.let { normalizeText(it) }` prefers the stored column and only falls back — so the two halves of the same PR disagree about whether the precomputed column is worth using. The secondary fields have no normalized counterpart and will always need computing, but at minimum they shouldn't be recomputed for items that already matched, and hoisting the tag/subject joins would help.

Behaviour changes that look right: `filterCourses` dropping the raw `contains(searchText, ignoreCase = true)` filter in favour of normalized matching makes accented titles behave consistently with the resource search, and re-sorting within each tier makes the result order deterministic instead of inheriting upstream order. Both are documented in the new KDoc.

CI green 5/5, and the test coverage for the new tiers is solid.

**Verdict:** needs another round — keep using `titleNormal` rather than re-normalizing every title on every keystroke; splitting the match selector from the sort selector is the way to get both.

---

PR #16626, opened 2026-08-28
dashboard: smoother navigating (fixes #16613)

### LINKED ISSUE #16613: fragile fragment backstack logic and magic state counter in DashboardElementActivity
DashboardElementActivity.openCallFragment uses a custom counter variable c and branching logic (if (c > 0 && c > 2)) to pop or replace fragments by tag. When users switch across multiple tabs (Courses, Library, Surveys, Teams, Community), fragments can fail to swap properly or remain attached under other fragments. The navigation logic should be streamlined using standard FragmentManager replacement or Android Jetpack Navigation.

ragilzakaria:
## Summary
Streamlines and standardizes fragment navigation across `openCallFragment` in [DashboardElementActivity](file:///Users/ragilzakaria/myplanet/app/src/main/java/org/ole/planet/myplanet/ui/dashboard/DashboardElementActivity.kt).

## Changes
- Refactored `openCallFragment` to utilize [FragmentNavigator.replaceFragment](file:///Users/ragilzakaria/myplanet/app/src/main/java/org/ole/planet/myplanet/ui/dashboard/FragmentNavigator.kt) consistently.
- Added fragment container visibility checks to ensure seamless UI transitions.

## Fixes
fixes #16613 

## Testing


dogi:
## Review — rating 58/100

Deleting the `var c` counter is absolutely the right instinct. That code was indefensible — `if (c > 0 && c > 2)` is just `c > 2`, `if (tag != "")` sat inside a branch that had already established it, and an activity-scoped mutable counter was gating fragment transactions. Nobody could reason about it.

But the replacement is not behaviour-preserving, and the change that got lost is the one that mattered.

### Tab switching now grows the back stack and loses state

Master, when a fragment with that tag existed but was not visible:

```kotlin
FragmentNavigator.popBackStack(fragmentManager, tag, 0)
```

It **popped back to the existing instance**. The PR falls through to `replaceFragment(..., addToBackStack = true)`, creating a fresh fragment and pushing a new entry.

So Dashboard → Library → Courses → Library now pushes a third back-stack entry with a brand-new `ResourcesFragment`, where master returned to the existing one. Two consequences:

- The back stack grows without bound as the user taps between bottom-nav tabs. Backing out of the app becomes N presses where N is however many times they switched tabs.
- Per-tab state — scroll position, active filters, search text — is discarded on every return to a tab.

The old implementation was incoherent, but its *intent* (pop to an existing tab rather than restack it) is standard bottom-navigation behaviour and is worth keeping. The counter can go without taking that with it:

```kotlin
val existing = tag?.takeIf { it.isNotEmpty() }?.let { fragmentManager.findFragmentByTag(it) }
when {
    existing?.isVisible == true -> return
    existing != null -> FragmentNavigator.popBackStack(fragmentManager, tag, 0)
    else -> FragmentNavigator.replaceFragment(
        fragmentManager, R.id.fragment_container, newFragment, addToBackStack = true, tag = tag
    )
}
```

That is the same simplification minus the regression.

### Interaction with #16622

#16622 rewrites `handleOnBackPressed` to `backStackEntryCount > 0 && !isAtRootDashboard()`. Its behaviour depends directly on how deep the back stack gets, so these two PRs need to be reasoned about together — if this one lands as written, back from a deep tab-switching stack walks through a pile of duplicate fragments instead of prompting logout.

### The tag fix is good, and overlaps #16627

Changing index 3 from `"survey"` to `"teams"` is correct — `TeamFragment` registered under a `"survey"` tag meant any `findFragmentByTag("teams")` missed it, which is very likely *why* the reuse path appeared broken and the counter got bolted on. Worth calling out in the description as the actual root cause.

Note #16627 is titled "dashboard: smoother teams tab navigation tagging" and looks like it makes the same fix. Please reconcile — one of the two should carry it.

### The hardening bundle, seventh copy

`MainApplication`, `SecurePrefs.warmUp`, `DiagnosticsRepositoryImpl`, `TimeUtils.formatDateForCsv`, `NotificationActionReceiverTest` — identical to #16618, #16620, #16622, #16623, #16624, #16625. Same objections: `catch (Throwable)` in `suspend` functions swallows `CancellationException`, and `initializeDatabaseConnection` swallowing everything lets the app run on a database it could not open. One PR, once, with cancellation rethrown.

No tests accompany a change to the app's central navigation entry point. `openCallFragment` is Robolectric-testable — asserting that returning to a previously opened tag does not add a back-stack entry would pin exactly the behaviour at issue here.

Moving to `change`.

---

dogi:
## Review — rating 88/100 (was 58)

The regression is fixed and the fix is now guarded by a test that would have caught it.

### Pop-to-existing is back

```kotlin
val existing = tag?.takeIf { it.isNotEmpty() }?.let { fragmentManager.findFragmentByTag(it) }
when {
    existing?.isVisible == true -> return
    existing != null -> FragmentNavigator.popBackStack(fragmentManager, tag, 0)
    else -> FragmentNavigator.replaceFragment(fragmentManager, R.id.fragment_container, newFragment, addToBackStack = true, tag = tag)
}
```

That keeps the behaviour the old `var c` counter was incoherently trying to express — returning to a tab pops back to its existing instance rather than restacking a fresh one — while the counter itself is gone. Tab switching no longer grows the back stack without bound, and per-tab state survives.

### The test is the right test

`testOpenCallFragment_switchingTabsAndReturningPopsBackStackInsteadOfRestacking` walks Library → Courses → Library and asserts:

```kotlin
assertEquals(1, fm.backStackEntryCount)
assertSame(libraryFragment, currentFragment)
```

`assertSame` on the original instance is what makes this worth having — it proves the *same* fragment was restored, not merely that something tagged `"library"` is on screen. That is the property that carries scroll position and filter state, and it fails against the previous revision of this PR.

Standing up `DashboardElementActivity` for this is not trivial — it extends `SyncActivity` → `ProcessUserDataActivity` — so the `@AndroidEntryPoint` test subclass with a stub content view and an injected `SharedPrefManager` for `UrlUtils.init` is a reasonable amount of scaffolding for real coverage of the app's central navigation entry point. Both test shards are green, so the Hilt test-activity processing works.

`testOnClickTabItems_usesTeamsTagForTeamFragment` also pins the `"survey"` → `"teams"` correction, which is very likely the original cause of the reuse path never finding the teams tab — the thing the counter was bolted on to compensate for.

Making `navigationView.menu.findItem(...)?.isChecked` null-safe is a small unrelated but welcome hardening; `findItem` is nullable and master would have thrown if a menu item were ever absent.

### Notes

The duplicate tag fix in **#16627** is now redundant — that PR is closed, so nothing to do, just confirming this PR carries it.

The interaction with **#16622** is now benign rather than a hazard: that PR's `backStackEntryCount > 0 && !isAtRootDashboard()` was concerning only while this one grew the stack per tab switch. With pop-to-existing restored, the stack stays shallow and the two compose correctly. Both are `ready`; either merge order works.

Worth watching that this test's setup cost (full Hilt + Robolectric + a three-level activity hierarchy, four times) does not become a drag on the shard it lands in — the repo already shards for wall clock.

Moving to `ready`.

---

Okuro3499 (approved):

dogi (approved):

PR #16637, opened 2026-08-31
resources: smoother repository html offline marking (fixes #16636)

### LINKED ISSUE #16636: html resources require multiple downloads
Not able to view html resource as even after download completion attempting to view triggers a fresh download

Okuro3499:
fixes #16636
Detect already-downloaded HTML resources during sync and mark them offline. Reconcile HTML resource state before opening in WebView. Add cover image preview support for HTML resources in both ResourcesAdapter and InlineResourceAdapter. Make DownloadService.onDownloadComplete suspend and mark resources offline by resource ID.

## Summary by CodeRabbit

* **New Features**
  * Improved offline handling for HTML resources, including automatic status reconciliation.
  * Added cover image previews for offline HTML resources in course and resource listings.
  * HTML resources now recognize existing local files and display them as available offline.

* **Bug Fixes**
  * Improved resource download completion tracking to ensure offline status updates reliably.
  * Added fallback handling when HTML cover images are unavailable.


dogi:
## Review — rating 72/100

The core problem is real and the download-path fix is the best part of this PR. Three things need attention before merge.

### The suspend refactor is a genuine fix, not just plumbing

```diff
-private fun onDownloadComplete(url: String) {
+private suspend fun onDownloadComplete(url: String) {
     if ((outputFile?.length() ?: 0) > 0) {
-        appScope.launch {
-            try { resourcesRepository.markResourceOfflineByUrl(url) } catch …
-        }
+        try { resourcesRepository.markResourceOfflineByUrl(url) } catch …
```

Master fired the offline-marking into `appScope` and returned immediately, so `onDownloadComplete` finished before the row was updated — a race where a UI refresh triggered by download completion could read the resource as still not-offline. Making the chain suspend closes it. Worth calling out in the description; it is more valuable than the preview work.

`markResourceOfflineByResourceId`'s guard is also right:

```kotlin
if (relativePath != entryFile) return
```

Multi-file HTML resources download many assets, and only the entry file should flip the flag — otherwise the first stylesheet to land marks the whole resource offline.

I confirmed `FileUtils.resolveHtmlEntryFile` (FileUtils.kt:108) and `MyLibraryDao.getByResourceId` both already exist, so nothing here is dangling. Both flavors and both test shards are green.

### 1. `reconcileHtmlResourceOffline` does not reconcile — it asserts

```kotlin
override suspend fun reconcileHtmlResourceOffline(resourceId: String) {
    val library = myLibraryDao.getByResourceId(resourceId) ?: return
    if (library.isResourceOffline()) return
    …
    library.resourceOffline = true
```

There is no filesystem check. It sets the flag unconditionally. It happens to be *correct today* because the only caller is inside `if (indexExists)` in `BaseContainerFragment:196`, where line 195 already verified `resolveHtmlEntryFile(directory, items.openWhichFile)?.exists() == true`. But the name promises a check the body does not perform, and the next caller will reasonably trust it.

Either move the existence check inside (it is two lines, and the function already computes `entryFile`), or rename to `markHtmlResourceOffline` so the precondition is the caller's job by contract.

### 2. `findHtmlCoverImage` re-walks the directory on every bind

```kotlin
val images = resourceDir.walkTopDown()
    .maxDepth(4)
    .filter { it.isFile && it.extension.lowercase() in previewImageExtensions }
    .toList()
```

For an offline HTML site this is a recursive walk collecting *every* image, then a `File.length()` stat on each for `maxByOrNull`. `ResourcesAdapter` calls it from `adapterScope.launch` on every `bind` with no cache — and unlike `InlineResourceAdapter`'s other preview paths, there is no `getFileCacheKeyIfExist` equivalent here. Scrolling a library containing several HTML resources will walk those trees repeatedly.

It is off the main thread, so it will not ANR, but it will burn I/O and battery and make previews flicker in on every scroll-back. Cache the resolved cover per resource id, or resolve it once at download time and store the path on `MyLibrary` alongside `resourceLocalAddress`.

### 3. Filesystem I/O moved into a model factory

The HTML self-heal block now lives in `MyLibrary.createFromResource`, which means a `@Entity` model's static factory reads `MainApplication.context` and stats the filesystem — during sync deserialization, once per HTML document. That is repository work in the model layer, it adds a blocking `exists()` to every sync batch, and it makes the factory untestable without a real context and a real directory.

`ResourcesRepositoryImpl` already has `reconcileHtmlResourceOffline`; running that over HTML rows after an insert batch would put the logic where it belongs and let it be tested. Also, the `catch (e: Exception) { e.printStackTrace() }` there runs against the direction the codebase just moved in (#16630's `runBestEffort`, #16494's tagged logging) — at minimum use `Log.w` with a tag.

### Testing

Nothing covers the new behaviour. The only test change adapts existing reflection to the new suspend signature. Three additions would be cheap and would pin the parts most likely to drift:

- `findHtmlCoverImage` is a pure function over a directory — a temp dir with `cover.png`, a larger `photo.jpg`, and a nested image verifies both the name-hint preference and the largest-file fallback.
- `markResourceOfflineByResourceId` with `relativePath` equal to and different from `openWhichFile` pins the entry-file guard.
- `reconcileHtmlResourceOffline` on an already-offline row asserts the early return.

Moving to `change` — the fix is sound, but the uncached directory walk on a scrolling list and the unguarded marker are worth resolving first.

---

dogi:

Reconciling HTML resources that are on disk but were never marked offline is a real user-facing fix, and the approach is right: reconcile at the three points where the gap can appear (sync upsert, course-library upsert, and immediately before opening the viewer). CI is green on all five checks, and the new `FileUtilsTest` / `ResourcesRepositoryImplTest` cases actually pin the behaviour rather than restating it.

The narrowing of `markResourceOfflineByUrl` is the part I like most — gating on `relativePath == entryFile` means a stylesheet or image finishing its download no longer flips the whole resource to offline. The two new tests cover exactly that pair.

### `MainApplication.context` where an injected `Context` already exists

`ResourcesRepositoryImpl` already takes `@param:ApplicationContext private val context: Context` as its first constructor parameter, and the file already builds this exact directory from it:

```kotlin
// ResourcesRepositoryImpl.kt:522 (master)
val baseDir = File(context.getExternalFilesDir(null), "ole/$resourceId")
```

`reconcileHtmlResourceOffline` builds the same path from `MainApplication.context` instead. That is the only reason the new tests need `mockkObject(MainApplication)` and a `finally { unmockkObject(...) }` — switch to the injected `context` and the mocking block disappears along with the global-state dependency. The only existing `MainApplication` reference in this file is `applicationScope` at line 420, so this introduces a new kind of coupling rather than following one.

### The reconcile loop is copy-pasted into `CoursesRepositoryImpl`

`reconcileHtmlLibraries` in `ResourcesRepositoryImpl` and the new `libraries.forEach { … }` block in `CoursesRepositoryImpl.kt:856-864` are the same eight lines — same `mediaType == "HTML"` guard, same null-`resourceId` skip, same `try`/`catch`/`Log.w`. Promoting `reconcileHtmlLibraries(libraries: List<MyLibrary>)` onto the `ResourcesRepository` interface would collapse the courses side to a single call and keep the "which libraries need reconciling" rule in one place.

### The reconcile pass re-reads rows it already holds

`reconcileHtmlLibraries` is handed the `librariesToUpsert` list, then calls `reconcileHtmlResourceOffline(resourceId)`, which immediately does `myLibraryDao.getByResourceId(resourceId)` — one DAO round trip per HTML library, on the sync path, for rows that were just passed in. Same for the courses side. An overload taking the `MyLibrary` directly would keep the single-id entry point for `BaseContainerFragment` while making the bulk path allocation-free.

### `catch (e: Exception)` now swallows cancellation

`onDownloadComplete` used to run inside `appScope.launch`; this PR makes it a `suspend fun` called from the download coroutine. `kotlinx.coroutines.CancellationException` is an `Exception`, so this block now absorbs cancellation of the download itself:

```kotlin
try {
    resourcesRepository.markResourceOfflineByUrl(url)
} catch (e: Exception) {
    e.printStackTrace()
}
```

This is the same taxonomy #16630 landed on master. A `catch (e: CancellationException) { throw e }` ahead of the `Exception` arm restores the old semantics.

### `findHtmlCoverImage` doesn't honour its own hint priority

`previewImageNameHints` reads as a preference order (`cover`, `thumbnail`, `thumb`, …), but the selection is `images.firstOrNull { hints.any { … } }` — first file in **walk order** matching *any* hint. A bundle containing both `assets/screenshot1.png` and `cover.png` picks whichever the directory walk reaches first. Iterating hints in the outer loop would make the list mean what it looks like. (`thumbnail` is also redundant given `contains` and `thumb`, and `contains("cover")` matches `discover.png`.)

Both adapters launch `showHtmlPreview` on the adapter-wide `adapterScope` rather than a holder-scoped job, so a slow `walkTopDown` can paint a cover into a recycled row — but that matches the existing `showPdfPreview` call sites exactly, so it's pre-existing precedent rather than something this PR introduces.

None of the above is a merge blocker; the injected-`Context` swap and the cancellation re-throw are the two I'd take before this lands.

---

dogi (approved):

PR #16639, opened 2026-08-31
sync: smoother upload pipelines unifying (fixes #16638)

### LINKED ISSUE #16638: duplicated pipeline
```
UploadCoordinator.kt has a fully duplicated pipeline: upload()/uploadRoom() (34-133 vs 319-382), uploadBatch/uploadBatchRoom (135-227 vs 384-445), queryItemsToUpload/...Room   
(257-279 vs 463-483) — ~230 lines of copy-pasted batch/HTTP/409/retry logic, acknowledged in a comment at line 312 as "until Realm is retired."                                 
RetryQueueWorker.processOperationInternal (176-249) reimplements the same PUT/POST/409/5xx logic a third time.  
```

Okuro3499:
fixes #16638
Extract UploadPipelineConfig interface from UploadConfig and RoomUploadConfig, eliminating the duplicated batch/HTTP/retry pipeline in UploadCoordinator. Both config types now implement the shared interface, and UploadCoordinator routes them through a single runPipeline method.

## Summary by CodeRabbit

* **Improvements**
  * Unified upload processing across supported storage systems.
  * Improved upload batching, retries, filtering, serialization, and persistence.
  * Upload status and error handling now use clearer model labels.
  * Added support for preserving uploaded-item results through the shared upload workflow.


J-S-webskas (approved):

dogi:

This collapses ~170 lines of deliberately duplicated pipeline into one path, and the seam it picks is the right one: `UploadPipelineConfig` abstracts exactly the two things that actually differed (the model label, and how results get persisted), leaving the batch/HTTP/409/retry loop shared. `UploadConfig.persistUploaded` keeping the `StepExam`/`Submission` dispatch and `RoomUploadConfig.persistUploaded` delegating to its DAO lambda means neither side lost behaviour. CI green on all five checks.

I walked the merged `queryItemsToUpload` against both originals. The Realm path's old two-pass structure (serialize `Simple` in pass 1, `Async` in pass 2, `else -> preSerialized ?: return@mapNotNull null`) is equivalent to the new single pass, because `UploadSerializer.serialize` is non-null-returning — so the `?:` arm was unreachable and the split bought nothing. `shouldFilter` likewise reproduces `filterGuests && guestUserIdExtractor != null && userId.startsWith("guest")` exactly. Good simplification, not just relocation.

One deliberate note this PR overrides is worth naming explicitly, since it was load-bearing:

> the batch/HTTP loop is intentionally duplicated rather than shared, so the still-live Realm path carries zero risk; the two converge once every uploadable model is on Room and the Realm path is deleted.

The convergence is happening before that stated precondition. I think that's defensible here — the Realm path is now a `persistUploaded` override rather than a parallel pipeline, so the blast radius is small — but it does mean the Realm upload path is being exercised through freshly-shared code, which brings me to the one real gap.

### No test covers the merged pipeline

`app/src/test/java/org/ole/planet/myplanet/services/upload/` contains only `UploadConfigsTest.kt`; there is no `UploadCoordinatorTest`. So the batch loop, the 409-refetch branch, `updateDatabaseBatch`'s failure reconciliation, and `queueRetryableFailures` all move from two untested copies to one untested copy, with nothing asserting equivalence. This is the path that persists user submissions and exam answers — a test that runs a `RoomUploadConfig` and a `UploadConfig` through `runPipeline` with a fake `UploadRepository` and asserts each lands in the right `persistUploaded` would make this refactor verifiable rather than merely plausible.

### Smaller points

- The guest-filter log lost its subject: `"Filtering out guest user item: $userId"` → `"Filtering out item from upload"`. Since `shouldFilter` returns `Boolean`, the id isn't available at the call site any more. Worth keeping the diagnostic, either by logging inside `shouldFilter` or by having it return the reason.
- `persistUploaded(uploadRepository, results)` passes `UploadCoordinator`'s own injected `uploadRepository` into the config so that `UploadConfig` can call back into it, while `RoomUploadConfig` ignores the parameter entirely. It works, but the dependency is travelling in a circle. Having `RoomUploadConfig` carry its lambda and `UploadConfig` carry a `UploadUpdateType` — with the coordinator doing the dispatch — would keep `uploadRepository` on the one object that already holds it.
- `modelClassName` in the retry queue changes from `config.modelClass.java.simpleName` to `KClass.simpleName` on the Realm side. Identical for `StepExam` and `Submission`, so no behaviour change today; just noting it in case anything downstream keys off the exact string for a nested class later.

Approving on the strength of the equivalence review — but the coordinator test is the thing that would let the next refactor here be safe by construction rather than by careful reading.

---

dogi (approved):

PR #16641, opened 2026-08-31
all: smoother callback listeners unifying (fixes #16640)

### LINKED ISSUE #16640: identical callback interfaces
```
Five callback interfaces are structurally identical no-arg "changed" events (OnMemberChangeListener, OnTeamUpdateListener, OnSecurityDataListener, OnFeedbackSubmittedListener, 
OnNotificationsListener) — collapsible to one fun interface. 
```

Okuro3499:
fixes #16640
Replace OnFeedbackSubmittedListener, OnMemberChangeListener, OnSecurityDataListener, and OnTeamUpdateListener with a single OnChangedListener fun interface, updating all usage sites.

## Summary by CodeRabbit

* **Refactor**
  * Consolidated several specialized update callbacks into a single reusable change-notification mechanism.
  * Updated feedback, team, membership, and security-data flows to use the unified callback.
  * Preserved existing behavior, including refreshing lists and notifying screens after successful updates.
  * Simplified callback usage by supporting direct lambda-based notifications.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6887
-        versionName = "0.68.87"
+        versionCode = 6888
+        versionName = "0.68.88"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/callback/OnChangedListener.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/callback/OnChangedListener.kt
@@ -1,0 +1,5 @@
+package org.ole.planet.myplanet.callback
+
+fun interface OnChangedListener {
+    fun onChanged()
+}
--- a/app/src/main/java/org/ole/planet/myplanet/callback/OnFeedbackSubmittedListener.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/callback/OnFeedbackSubmittedListener.kt
@@ -1,5 +1,0 @@
-package org.ole.planet.myplanet.callback
-
-interface OnFeedbackSubmittedListener {
-    fun onFeedbackSubmitted()
-}
--- a/app/src/main/java/org/ole/planet/myplanet/callback/OnMemberChangeListener.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/callback/OnMemberChangeListener.kt
@@ -1,5 +1,0 @@
-package org.ole.planet.myplanet.callback
-
-interface OnMemberChangeListener {
-    fun onMemberChanged()
-}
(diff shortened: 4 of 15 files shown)


dogi:
## Review — rating 80/100

Sound consolidation, and it is complete rather than partial — which is the thing I usually have to ask for.

I enumerated every interface in `callback/` with exactly one no-argument method:

```
OnFeedbackSubmittedListener.kt: fun onFeedbackSubmitted()
OnMemberChangeListener.kt:      fun onMemberChanged()
OnSecurityDataListener.kt:      fun onSecurityDataUpdated()
OnTeamUpdateListener.kt:        fun onTeamDetailsUpdated()
```

Those four are exactly the set this PR merges. Nothing structurally identical is left behind, so the codebase ends up consistent instead of half-migrated. Both flavors and both test shards are green, so none of the 34 references on master dangle.

Making it a `fun interface` is what earns the change its keep — `BecomeMemberActivity` drops a five-line anonymous object for a lambda, and the previously awkward "one class implements two identical interfaces" arrangement in `TeamDetailFragment` becomes explicit wiring.

### The cost, and a one-line mitigation

Collapsing four types into one means `TeamPagerAdapter` now takes **two parameters of the same type**:

```kotlin
private val onMemberChangeListener: OnChangedListener,
private val teamUpdateListener: OnChangedListener
```

and the only call site passes them positionally:

```kotlin
TeamPagerAdapter(
    this, pageConfigs, team?._id,
    OnChangedListener { onMemberChanged() },
    OnChangedListener { onTeamDetailsUpdated() }
)
```

Before this PR the distinct types made those two arguments unswappable. Now transposing them compiles cleanly and silently misroutes both callbacks — member changes refresh team details and vice versa, which would present as a confusing "the list sometimes doesn't update" bug rather than anything obviously wrong.

Named arguments restore the safety for free:

```kotlin
TeamPagerAdapter(
    parentFragment = this,
    pages = pageConfigs,
    teamId = team?._id,
    onMemberChangeListener = OnChangedListener { onMemberChanged() },
    teamUpdateListener = OnChangedListener { onTeamDetailsUpdated() },
)
```

Worth doing here, since this is precisely the site where the type distinction was doing real work.

### Two smaller notes

`TeamDetailFragment` moving from `override fun onMemberChanged()` to a private method plus explicit lambda wiring is an improvement in its own right — the old arrangement relied on the fragment implicitly satisfying two interfaces, which was easy to miss when reading `TeamPagerAdapter`'s constructor.

`FeedbackListFragment` still implements the interface directly, so its override is now `onChanged()`. That reads as considerably less informative than `onFeedbackSubmitted()` at the implementation site. Not worth blocking, but a one-line KDoc on `OnChangedListener` explaining it means "something this screen displays changed, re-query" would help, since the name alone invites confusion with `androidx.lifecycle.Observer.onChanged`.

Moving to `ready`.

---

J-S-webskas (approved):

dogi (approved):

PR #15951, opened 2026-08-20
teams: smoother repository update requesting (fixes #15568)

J-S-webskas:
https://github.com/user-attachments/assets/68ac7003-a41d-4c87-a000-c0e55a47276a

## Summary by CodeRabbit

* **New Features**
  * Added support for selecting, previewing, uploading, and displaying team profile images when creating or updating teams.
  * Redesigned team list cards with profile images, improved spacing, metadata, descriptions, and action controls.
  * Improved team creation layout with a profile image area and updated spacing.

* **Usability**
  * Plan content can now scroll within its screen.
  * Updated default date formatting to `MM/dd/yyyy`.
  * Simplified team list presentation by removing the visible table header.
  * Clarified visit counts by adding a “visits” label.

Fixes #15568

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #16646, opened 2026-08-31
courses: filter stays active (fixes #16271)

### LINKED ISSUE #16271: courses filter resets
```
Filters are a user-invested state (they explicitly chose to narrow a list), and losing that state on a simple "view an item then go back" round-trip forces the user to redo the work. The expected behavior is that list/filter state survives navigating into a detail view and back, similar to how most list-detail patterns (email, e-commerce, file browsers) preserve scroll position and filters.
```
https://github.com/user-attachments/assets/933c43d8-d32b-4d7c-876d-98e19e3ca622

J-S-webskas:
https://github.com/user-attachments/assets/75b81a64-f367-438f-8c68-8f2ba3affe53

## Summary by CodeRabbit

* **New Features**
  * Course filters now persist and are automatically restored when returning to the Courses screen.
  * Previously selected search text, grade, subject, tags, and progress filters remain applied after course data reloads.
* **Bug Fixes**
  * Prevented restored filter settings from being cleared or triggering duplicate filtering when the Courses screen initializes.
  * Added a database schema revision to support reliable data resynchronization.

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/data/room/AppDatabase.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/data/room/AppDatabase.kt
@@ -125,7 +125,7 @@ import org.ole.planet.myplanet.model.UserEntity
         Answer::class,
         MyTeam::class,
     ],
-    version = 9,
+    version = 10,
     exportSchema = false,
 )
 @TypeConverters(Converters::class)
--- a/app/src/main/java/org/ole/planet/myplanet/ui/courses/CourseFilterController.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/courses/CourseFilterController.kt
@@ -91,6 +91,48 @@ class CourseFilterController(
         onScrollToTop()
     }
 
+    fun restoreFilterState(state: FilterState, tags: List<TagEntity> = emptyList()) {
+        if (::etSearch.isInitialized && etSearch.text.toString() != state.searchText) {
+            etSearch.setText(state.searchText)
+        }
+        if (::spnGrade.isInitialized) {
+            val gradeAdapter = spnGrade.adapter
+            if (gradeAdapter != null) {
+                var found = false
+                for (i in 0 until gradeAdapter.count) {
+                    if (gradeAdapter.getItem(i).toString() == state.grade || (state.grade.isEmpty() && i == 0)) {
+                        spnGrade.setSelection(i)
+                        found = true
+                        break
+                    }
+                }
+            }
+        }
+        if (::spnSubject.isInitialized) {
+            val subjectAdapter = spnSubject.adapter
+            if (subjectAdapter != null) {
+                for (i in 0 until subjectAdapter.count) {
+                    if (subjectAdapter.getItem(i).toString() == state.subject || (state.subject.isEmpty() && i == 0)) {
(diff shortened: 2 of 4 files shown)


dogi:
## Review — rating 35/100 → `duplicate`

**Recommend closing in favour of #16647.** These two PRs implement the same feature — persisting the course filter across navigation — across the same three files, with the same `restoreFilterState` / `currentFilterState` / `filterCoursesInternal` design. #16647 is the same idea executed cleanly, and it does not carry the defect below. I'm not closing this myself; that's the author's call.

### The blocking issue: a schema-less `AppDatabase` version bump

```diff
-    version = 9,
+    version = 10,
```

Nothing else in this diff touches an entity, a DAO, or `Converters` — the only other files are `CourseFilterController.kt`, `CoursesFragment.kt`, and `CoursesViewModel.kt`, all pure UI. `RoomModule` builds with `fallbackToDestructiveMigration(true)`, and destructive fallback fires on a version *change*, so this bump does exactly one thing: **drops and rebuilds the local database on every existing install**, discarding any unsynced local writes — pending uploads, drafts, submissions not yet pushed — for a UI filter-persistence change that needs no schema at all.

That's the mirror image of the omitted-bump problem in #15951 and #15699, and it's just as invisible in testing: a fresh install can't show it, and the rebuild looks like a successful resync. Dropping the two-line hunk is the whole fix.

### Beyond that, this revision is the rougher of the two

- **Dead `found` variable.** In the `spnGrade` block, `var found = false` is assigned and never read; the structurally identical `spnSubject` block below it has no such variable. Copy-paste residue.
- **`cachedAllCourses` is never used.** `filterCoursesInternal` takes `cachedAllCourses: List<MyCourse>?` and the reload path passes `allCourses` into it, but the body ignores the parameter and unconditionally calls `coursesRepository.filterCourses(...)` + `getMyCourses(...)`. So the reload path does a second full repository round trip immediately after fetching `allCourses` / `validCourses` / `myCourses` — and the unused parameter is evidence that avoiding exactly that was the intent. #16647 drops the parameter rather than leaving the trap in place.
- **Eight positional parameters.** `filterCoursesInternal(isMyCourseLib, userId, searchText, selectedGrade, selectedSubject, tagNames, progressFilter, allCourses, progressMap, tagsMap)` — two adjacent `String` triples with no named arguments at either call site. #16647 passes a `FilterState` instead, which removes the swap hazard entirely.
- **Stray indentation** on `         return if (isMyCourseLib) {` (nine spaces).

### One bug both PRs share, worth carrying to the winner

`restoreFilterState(state, tags = emptyList())` is called with only the state, so `searchTags` is cleared and never repopulated, and `state.tagNames` is ignored outright. `currentState()` then builds `tagNames = searchTags.mapNotNull { it.name }` — empty. In *this* PR the resulting first emission is swallowed by the widened guard (`if (savedFilter.isActive || !state.isActive)`), so the ViewModel keeps its tags and the list stays tag-filtered while `refreshTagText()` renders an empty chip line: **the list is filtered by tags the UI says aren't selected**, and the user has no control to clear them. Whichever PR proceeds needs `restoreFilterState` to resolve `state.tagNames` back into `TagEntity`s.

CI is green on all five checks, and `CoursesViewModelTest.kt` exists but gains no case for the new persistence behaviour.

---

PR #16647, opened 2026-08-31
courses: smoother filter controller view modelling (fixes #16271)

J-S-webskas:
Refactored restoreFilterState in CourseFilterController to break down state restoration into modular helper methods, lowering cyclomatic complexity. Reduced parameter count of filterCoursesInternal in CoursesViewModel by using FilterState object.

---
*PR created automatically by Jules for task [13028885277691387066](https://jules.google.com/task/13028885277691387066) started by @J-S-webskas*

## Summary by CodeRabbit

* **New Features**
  * Course filters now persist while navigating and are restored when reopening the Courses screen.
  * Restored filters include search text, grade, subject, tags, and progress criteria.
  * Course results automatically reload using the active filter selections.
  * Filtered course lists continue to reflect progress-based and course-library criteria.

* **Bug Fixes**
  * Tag selections are preserved and displayed correctly when filters are restored.
  * Tags continue to match correctly when available tags are refreshed, including tags without IDs.


dogi:
## Review — rating 66/100 → `change`

This is the better of the two implementations of course-filter persistence — see #16646, which I've marked `duplicate` in favour of this one. Compared to that revision this PR extracts `restoreSearchText` / `restoreSpinnerSelection` / `restoreTags` instead of inlining three loops, passes a `FilterState` into `filterCoursesInternal` rather than eight positional parameters, carries no dead `found` variable and no unused `cachedAllCourses`, and — importantly — **does not bump `AppDatabase.version`**, which in the sibling PR would have dropped every existing user's local database for a UI-only change. The extraction of the shared filter/progress logic out of `filterCourses` into one reusable `filterCoursesInternal` is a genuine improvement over master, independent of the feature. CI green on all five checks.

Holding it at `change` for one functional gap in the feature itself.

### Tags are not restored — and are then destroyed

`restoreFilterState(state, tags = emptyList())` is called from the fragment with a single argument, so `tags` takes its default. `restoreTags(emptyList())` clears `searchTags` and adds nothing, and `state.tagNames` — which is part of the `FilterState` being restored — is never consulted. `currentState()` then builds `tagNames = searchTags.mapNotNull { it.name }`, so the state published by `_filterState.value = currentState()` has an empty tag list.

Unlike #16646, this PR leaves the fragment's first-emission guard as `if (!state.isActive)`. The restored state *is* active (search/grade/subject survived), so the guard doesn't fire and the emission falls straight through to:

```kotlin
viewModel.filterCourses(
    isMyCourseLib, model?.id, state.searchText, state.grade,
    state.subject, state.tagNames, state.progressFilter
)
```

`filterCourses` opens with `currentFilterState = filterState`, so the saved tag selection is overwritten with the empty list on the first emission after restore. Net effect: filter by a tag, open a course, come back — search text, grade and subject return, tags are silently gone and cannot be recovered.

The fix belongs in `restoreFilterState`: resolve `state.tagNames` back into `TagEntity` instances (the controller would need the available tag list, which is what the unused `tags` parameter is presumably reaching for) rather than accepting a parameter no caller supplies. As it stands, `tags` is dead across the whole codebase.

### `restoreSpinnerSelection` fires the spinner listener

`spnGrade` and `spnSubject` share `spinnerListener`, whose `onItemSelected` calls `_filterState.value = currentState()` (CourseFilterController.kt:77-79). `setSelection(i)` to a position other than the current one dispatches that callback, so restoring a non-default grade or subject publishes an extra state — after the first-emission bookkeeping has already run — with whatever `searchTags` holds at that moment. It lands on the same empty-tags value here, so it's not currently a second distinct bug, but it makes the restore sequence order-dependent in a way that will bite once tags are wired up. Detaching the listeners around the restore (the controller already nulls them in its teardown at lines 198-199) would make `restoreFilterState` atomic.

### `currentFilterState` doesn't survive process death

`CoursesFragment` uses `by viewModels()`, so the ViewModel is fragment-scoped. That's fine for the navigate-away-and-back case this PR targets, and it's the right scope for it. Worth knowing that it does not cover configuration-change-plus-process-death; `SavedStateHandle` would, if that turns out to matter.

### Tests

`CoursesViewModelTest.kt` exists and gains nothing. The two assertions that would lock this feature down are cheap: that `filterCourses` records `currentFilterState`, and that a subsequent course reload with an active `currentFilterState` routes through `filterCoursesInternal` rather than `processCourses`. A third asserting round-trip tag preservation would fail today, which is exactly the point.

Fix the tag restoration and I'd move this to `ready`.

---

dogi:

Both findings from my last review are fixed, and the tests I asked for are here.

**Tags are actually restored now.** `restoreTags(state.tagNames, availableTags)` reads the saved names instead of ignoring them, so `currentState()` rebuilds a `FilterState` carrying the real tag list. That closes the loop I described: previously the first emission after restore fed an empty `tagNames` into `filterCourses`, which overwrote `currentFilterState` and destroyed the saved tag selection permanently. Filter by tag, open a course, come back — the tags survive.

Worth noting the fragment's first-emission guard is unchanged (`if (!state.isActive)`), and that's now fine: the restored state is active, so it falls through to `viewModel.filterCourses(...)` — but with the correct tag names, so the write back to `currentFilterState` is idempotent instead of destructive. The bug was never the guard, it was what the guard was letting through.

**The spinner restore is atomic.** Detaching `onItemSelectedListener` from both spinners around the restore and reattaching after is exactly the fix — `setSelection` to a non-current position dispatches `onItemSelected`, which calls `_filterState.value = currentState()` mid-restore. Doing it before `restoreTags` runs means that stray emission would have carried an empty tag list. Removing the window removes the whole class of ordering bug rather than the one instance.

**And the ViewModel behaviour is pinned.** The three new tests cover what I asked for — `currentFilterState` is recorded by `filterCourses`, a subsequent `loadCourses` routes through `filterCoursesInternal` with the saved filter (`coVerify(atLeast = 2)` on the same `filterCourses` call is a neat way to assert the reload path), and tag names round-trip. CI green on all five checks.

One leftover, not blocking.

### `availableTags` is still a parameter nobody supplies

```kotlin
fun restoreFilterState(state: FilterState, availableTags: List<TagEntity> = emptyList())
```

`CoursesFragment` calls `filterController.restoreFilterState(savedFilter)`, so `availableTags` is always empty and the `availableTags.find { it.name == name }` branch never executes — every restored tag is the synthetic `TagEntity().apply { this.name = name }` from the fallback.

For the courses path that's harmless: `currentState()` reads `searchTags.mapNotNull { it.name }`, `refreshTagText()` renders names, and `CoursesRepository.filterCourses` matches on names, so a name-only entity round-trips correctly. The two things worth checking are whether the tag-picker dialog marks previously-selected tags as checked by `id` (a synthetic tag has none, so they'd appear unchecked), and whether anything downstream expects a fully-populated `TagEntity`.

Either pass the real tag list from the fragment so the matching branch does its job, or drop the parameter and make the synthetic construction the documented behaviour. Right now it's a half-wired seam that reads as if it resolves real tags when it never does.

The process-death caveat from last time still stands and is still fine — `by viewModels()` covers navigate-away-and-back, which is what this PR targets; `SavedStateHandle` would be needed for more, and that's a separate ask.

---

Okuro3499 (changes requested):
```
Problem: restoreFilterState's availableTags parameter defaults to emptyList() and the only call site (CoursesFragment.kt:227) never supplies a real tag list. So every tag restored
  after a view recreation becomes a synthetic TagEntity().apply { name = name } with id = "". This list becomes filterController.searchTags, which is later passed by reference into 
  CollectionsFragment.getInstance(filterController.searchTags, "courses") at CoursesFragment.kt:301. CollectionsFragment.buildTagDataList computes selectedIds =                     
  selectedItemsList.mapNotNull { it.id }.toHashSet() (CollectionsFragment.kt:121) and compares against the real, DB-loaded tag entities' non-empty ids                               
  (CollectionsFragment.kt:129,136). Since the restored tags all have id = "", none of them ever matches, so:                                                                         
  - Reopening the tag picker after a config change/tab switch shows the previously-selected tags as unchecked, even though the course list is still correctly filtered by name — a   
    visible, confusing inconsistency between tvSelected (still shows the tag names) and the picker's checkboxes.                                                                     
  - Because TagEntity has no equals() override, onCheckboxTagSelected's selectedItemsList.contains(tag) (CollectionsFragment.kt:156) uses reference equality, so tapping an          
    already-selected tag row won't be recognized as "already selected" and can add a duplicate entry into selectedItemsList (self-heals only because setTags dedupes by name on      
    commit — but the interaction is still broken during the picker session).
```

Okuro3499 (approved):

dogi:
## Re-review — rating 68/100 → `change`

**The tag restoration is properly solved now.** Carrying `tags: List<TagEntity>` on `FilterState` alongside `tagNames`, populating it from `searchTags.toList()` in `currentState()`, and preferring it in `restoreTags` means the real entities survive the round trip instead of being reconstructed from names. That's a better answer than the one I suggested — no lookup needed at all on the happy path. The `CollectionsFragment` reconciliation (remapping `selectedItemsList` onto the freshly-loaded tag objects) closes the matching gap on the dialog side.

But the change that makes that reconciliation work introduces a defect in a model class used across the app.

### `TagEntity.equals`/`hashCode` violate the contract

```kotlin
override fun equals(other: Any?): Boolean {
    ...
    return if (id.isNotEmpty() && other.id.isNotEmpty()) id == other.id
           else !name.isNullOrEmpty() && name == other.name
}
override fun hashCode(): Int =
    if (id.isNotEmpty()) id.hashCode() else name?.hashCode() ?: 0
```

Take `A(id="1", name="X")` and `B(id="", name="X")` — the synthetic name-only tag your own fallback creates:

- `A.equals(B)` takes the *else* branch (B's id is empty) and returns **true**.
- `A.hashCode()` is `"1".hashCode()`, `B.hashCode()` is `"X".hashCode()` — **different**.

Equal objects with unequal hash codes. Any `HashSet<TagEntity>`, `Map` keyed by tag, `distinct()`, `toSet()` or `groupBy` will behave incorrectly — `set.contains(B)` can miss an already-present `A`, and adding both yields a set containing two "equal" elements.

Equality is also non-transitive: with `C(id="2", name="X")`, `A == B` and `B == C` are both true while `A == C` is false, because that pair takes the id branch.

Today's call sites happen to dodge it — `selectedItemsList` is a list scanned with `indexOfFirst`, and `selectedIds`/`selectedNames` are `HashSet<String>` rather than of tags — so this probably doesn't misbehave yet. That's exactly what makes it worth fixing now rather than after someone writes `tags.toSet()`.

The straightforward fix is to make both consistent: key on `id` when present and on `name` otherwise, in *both* methods, and treat an id-bearing tag and a name-only tag as different objects — then do the name-based reconciliation explicitly at the one place that needs it (`CollectionsFragment`) rather than by overloading `equals`.

### Nothing tests the new equality

The three ViewModel tests are unchanged from the last revision and don't touch `TagEntity`. Given the subtlety above, a handful of assertions — id vs id, name vs name, id-vs-name, and `hashCode` agreement for any pair that compares equal — would be cheap and would have caught this.

### Smaller

- `availableTags` is *still* a parameter no caller supplies. It's now a second-tier fallback behind `state.tags`, so it's harmless, but the `availableTags.find { it.name == name }` branch remains unreachable in production. Either wire the fragment to pass the loaded tags or drop it.
- The re-indentation of the `collectLatestWhenStarted(viewModel.coursesState)` block in `CoursesFragment` is a genuine fix (the old nesting was wrong), but it lands as ~20 lines of whitespace churn in a diff about filter persistence — worth calling out in the description so reviewers don't hunt for behaviour changes in it.

Fix the equality contract and I'd move this back to `ready`; everything else here is in good shape.

---

dogi:
## Re-review — rating 76/100

**The `equals`/`hashCode` contract violation from my last review is properly fixed.** Adding the `else -> false` arm for the mixed case (one side has an id, the other doesn't) is the right call, and it makes the relation sound:

- both ids non-empty → compare ids → both hash `id.hashCode()`
- both ids empty → compare non-empty names → both hash `name.hashCode()`
- mixed → never equal, and unequal objects are free to hash differently

That's reflexive (via the `this === other` fast path), symmetric, transitive, and consistent with `hashCode`. `FilterState` is a `data class` that now carries `List<TagEntity>`, so its generated `equals` depends on this being right — worth having got correct.

Four things below, one of which is user-visible.

### 1. The restored progress filter doesn't reach the chip UI

`restoreFilterState` sets `progressFilter = state.progressFilter`, but nothing re-renders the progress chip row afterwards. The ordering makes this unavoidable as written:

- `CoursesFragment` line 169 — `setupCourseFilterChips()`, which ends by calling `renderCourseChipSelection(chipRow)`
- line ~225 — `filterController.restoreFilterState(savedFilter)`

Chips are painted before the state is restored, and `renderCourseChipSelection` reads `filterController.currentState().progressFilter`, so the row keeps showing "All" highlighted while the list is filtered to "In Progress". The user sees a filtered list with no filter apparently selected, and tapping "In Progress" then looks like a no-op.

Calling `renderCourseChipSelection` after the restore fixes it.

### 2. `loadCourses` queries the database twice under an active filter

```kotlin
val validCourses = ...            // queried
val myCourses = ...               // queried
...
if (currentFilterState.isActive) {
    filterCoursesInternal(...)    // queries filterCourses() and getMyCourses() again
} else {
    processCourses(isMyCourseLib, userId, validCourses, myCourses, ...)
}
```

In the active-filter branch, `validCourses` and `myCourses` are computed and then discarded — `filterCoursesInternal` re-runs `coursesRepository.filterCourses(...)` and `getMyCourses(...)` itself. `testLoadCourses_withActiveCurrentFilterState_usesFilterCoursesInternal` asserts `coVerify(atLeast = 2)`, which pins the duplicate rather than flagging it. Since this PR family is about making the courses screen smoother, doubling the query on every load with a filter active is worth avoiding — hoist the branch above the two queries.

### 3. Two different notions of `TagEntity` equality now coexist

`TagEntity.equals` says mixed id/no-id is **not** equal. `CollectionsFragment.onCheckboxTagSelected` says it **is**:

```kotlin
if (tag.id.isNotEmpty() && selected.id.isNotEmpty()) selected.id == tag.id
else !tag.name.isNullOrEmpty() && selected.name == tag.name   // mixed → name match
```

The looser rule is presumably deliberate, so a restored `TagEntity().apply { name = ... }` (no id) matches a real tag loaded from the DB. But now `contains(tag)` and `indexOfFirst { ... }` disagree on the same pair, and the next person to reach for `contains` will get the other answer. Pick one: either make `equals` implement the lenient rule, or drop `equals` and route every comparison through a named helper like `TagEntity.matches(other)`.

Relatedly, name-based matching collides. `buildTagDataList` now marks a tag selected when `selectedNames.contains(parentTag.name)`, and `restoreTags` dedupes by name — so two collections sharing a name select and deselect together. The reconciliation step in the `Success` branch mitigates it by swapping no-id tags for real ones, but it falls back to `?: selected` when nothing matches, so no-id tags can still survive into the comparison.

### 4. Reattaching the spinner listener synchronously

```kotlin
restoreSpinnerSelection(spnGrade, state.grade)   // calls spinner.setSelection(i)
...
spnGrade.onItemSelectedListener = listener       // reattached immediately
```

Detaching the listeners around the restore is the right instinct. But `Spinner.setSelection(int)` doesn't fire `onItemSelected` synchronously — it requests a layout and the callback arrives on a later pass, by which point the listener is back in place. So the restore can still emit a spurious filter event. `lastState` dedup will usually absorb it, which is why this hasn't shown up, but the guard isn't doing what it looks like it's doing. `spinner.post { spinner.onItemSelectedListener = listener }` is the usual fix.

### Tests

`restoreFilterState` is the headline of the PR and nothing exercises it — not the listener detach/reattach, not the two-source tag restore, not the name dedup. The three tests added to `CoursesViewModelTest` cover `currentFilterState` bookkeeping, and two of them overlap heavily.

Worth knowing: #16701 adds a `CourseFilterControllerTest` that inflates the real layout and drives the controller directly. That's the right harness for `restoreFilterState`, whichever PR lands first.

### Heads-up on #16701

That PR edits `CoursesFragment.kt` at the same two places this one does — the `filterController.filterState` collector around line 225 and the `btn_clear_tags` listener around line 335 — so these will conflict. It also adds the `renderCourseChipSelection(chipRow)` call into the collector that finding 1 above asks for, so whichever merges second gets a smaller job. Worth coordinating with @ragilzakaria on ordering.

Marking `change` on finding 1; the rest are worth addressing but wouldn't block on their own.

---

dogi:
**Review — rating: 55/100**

I can't assess the feature here, because the branch's history has diverged from master in a way that makes this PR unmergeable as it stands — and dangerously so, since git will merge it **without a single conflict**.

**What's actually wrong.** The branch has 6848 commits ahead of the merge base and **zero merge commits**: it was rebased onto master rather than having master merged into it, which rewrote master's history under new SHAs. Git can therefore no longer find a recent common ancestor — the merge base dates to 2026-09-03 and master is 93 commits past it. That's why GitHub shows 282 files changed for what is, by your own commits, six commits of work:

```
1a662e004  added renderCourseChipSelection
4f5606447  fixed equal contract and reconciliation
bd4892ed8  fixed tag
436ec1b81  Reconciling the restored synthetic tags against DB-loaded tags
9a904c8b7  Restore searchTags from FilterState.tagNames into TagEntity objects
4c338603f  refactor: reduce complexity of restoreFilterState and parameter count
```

**What merging would actually do.** Diffing the branch against *current* master: **355 files, 3035 insertions, 12307 deletions.** Among them `.github/workflows/build.yml`, `release.yml` and `test.yml`, `.github/dependabot.yml`, `.claude/settings.json`, `.gitmodules`, `CLAUDE.md`, `AGENTS.md`, `README.md`, `app/build.gradle`, `gradle/libs.versions.toml`, `docs/CODE_STYLE_GUIDE.md` and `docs/TESTING.md` — plus it re-adds `app/libs/flexbox-1.0.0.aar` and `app/libs/ChipCloud-3.0.5.aar`, vendored AARs master has since dropped in favour of Maven coordinates. Roughly a week of master would be silently rolled back, and `git merge-tree origin/master` reports **0 conflicts**, so nothing would flag it during the merge. CI on the PR is green because the branch builds a self-consistent older tree; that tells you nothing about the post-merge state.

**What to do.** Bring master in — `git merge origin/master` on the branch, not another rebase — resolve whatever comes up, and push. That restores a real merge base and collapses the diff to your six commits. Alternatively, rebase the six commits onto *current* `origin/master` and force-push the branch, which gets the same result if you'd rather keep linear history. Once the diff shows only `CourseFilterController`/`CoursesFragment`-area files plus their tests, this is reviewable on its merits.

One sequencing note for afterwards: #16701 touches the same course filter chip area and is small, green and self-contained. It should land first, and this branch should be rebuilt on top of it.

**Verdict:** needs another round — merge current master into the branch so the diff reflects your changes rather than a rollback of everyone else's.

---

dogi:
**Re-review — rating: 76/100** (was 55)

**The branch is fixed.** It's rebased onto current master — zero commits behind, merge base 2026-09-08 — and the diff is now your six commits over six files (231 insertions), not a 355-file rollback of everyone else's work. That was the whole of my last review and it's fully resolved, so this is the first time the change itself is reviewable.

**The filter-persistence fix is real and worth having.** `CoursesViewModel.currentFilterState` plus the `filterCoursesInternal` extraction means a background data refresh now re-applies the active filter instead of calling `processCourses` with the unfiltered list and silently wiping what the user had selected. That was a genuine bug, and reusing one function for both the refresh path and the explicit-filter path is the right way to make it impossible to reintroduce. The `restoreFilterState` work — detaching spinner listeners around the restore so it doesn't re-fire the filter, then reattaching — is careful in the way this kind of code has to be.

**The problem is `TagEntity` identity: the PR introduces four different rules for it, and the one it puts in `equals` is the one the feature can't use.**

```kotlin
// 1. TagEntity.equals — strict
if (id.isNotEmpty() && other.id.isNotEmpty()) id == other.id
else if (id.isEmpty() && other.id.isEmpty()) !name.isNullOrEmpty() && name == other.name
else false
```
```kotlin
// 2. CollectionsFragment.onCheckboxTagSelected — lenient
if (tag.id.isNotEmpty() && selected.id.isNotEmpty()) selected.id == tag.id
else !tag.name.isNullOrEmpty() && selected.name == tag.name
```
```kotlin
// 3. CollectionsFragment.buildTagDataList — union
selectedIds.contains(parentTag.id) || selectedNames.contains(parentTag.name)
```
```kotlin
// 4. CourseFilterController.restoreTags — name only
if (name != null && seenNames.add(name)) { searchTags.add(tag) }
```

Rules 2, 3 and 4 all have to match a restored name-only tag (`TagEntity().apply { this.name = name }`, id `""`) against a DB-loaded tag that *does* have an id. That's the central case this feature exists to handle. Rule 1 returns **false** for exactly that pairing — the `else` branch. So `equals`/`hashCode` are added to express tag identity, then can't be used for it, and the work is done by three hand-rolled comparisons that don't agree with each other or with `equals`.

That's not a live bug today — I checked, and nothing existing depends on `TagEntity` equality: `ResourcesTagsAdapter`'s `DiffUtil` callback compares `tag.id` and `tag.name` field-by-field rather than with `==`, and the one `MutableList.remove(tag)` at `ResourcesFragment.kt:717` removes an instance that is in the list by reference anyway. But it's a trap set for later: the next person to write `searchTags.contains(tag)` or drop these into a `HashSet` will get rule 1's answer while every nearby call site implements rule 2 or 3.

The fix is to pick one rule and have it be `equals`. If it's the lenient one the feature needs, `hashCode` has to hash on `name` unconditionally — hashing by `id` when present, as it does now, would mean an id-bearing tag and its name-only counterpart land in different buckets and never meet in a hash-based collection, so the lenient equals would be unreachable there. With that in place, call sites 2 and 3 collapse to `contains` / `indexOf`, and `restoreTags`' `seenNames` set becomes a plain `distinct()`.

If you'd rather not put a lenient rule on a Room `@Entity` — a fair position, since these are mutable `var` fields and mutating one after it lands in a `HashSet` loses it — then drop `equals`/`hashCode` entirely and give `TagEntity` a single `fun matches(other: TagEntity): Boolean` that all three sites call. Either way the goal is one rule with one name.

Two smaller things:

- `CoursesFragment` shadows `chipRow`: `val chipRow = view?.findViewById<LinearLayout>(R.id.chip_filter_row)` is declared, then a second `val chipRow = requireView().findViewById<LinearLayout>(...)` inside the `if (savedFilter.isActive)` block. The outer one is already in scope and already null-checked at its use site.
- The refresh path still ends in `e.printStackTrace()`. Pre-existing rather than added here, but #16782 landed specifically to remove these, so it's worth taking while you're in the file.

The `CollectionsFragment` reconciliation — swapping synthetic name-only tags for the real DB objects once the list loads — is the right shape, and it's what makes the whole restore work end to end.

CI green (6 checks including the Dependabot config check).

**Verdict:** needs another round — consolidate the four tag-identity rules into one. Everything else here is ready.

---

dogi:
**Re-review — rating: 87/100** (was 76)

The four competing tag-identity rules are now one, and you picked the option that avoids the trap.

`equals`/`hashCode` are gone from the entity; in their place a single `matches`:

```kotlin
    return if (id.isNotEmpty() && other.id.isNotEmpty()) id == other.id
    else !name.isNullOrEmpty() && name == other.name
}
```

and every site now calls it — `restoreTags`, `addTag` and `setTags` through the new `addTagInternal`, plus the reconciliation, both `buildTagDataList` branches, and `onCheckboxTagSelected`. The hand-rolled `selectedIds`/`selectedNames` sets and the `seenNames` dedupe are all gone.

**Keeping this off `equals` is what makes it correct, and it's worth being explicit about why:** the lenient rule can't satisfy the equals contract. A(id=`t1`,name=`X`) matches B(id=``,name=`X`), B matches C(id=`t2`,name=`X`), but A doesn't match C — not transitive. As `equals` on a Room entity that could land in a `HashSet`, that's a latent bug; as a named predicate used only in `any {}` / `indexOfFirst {}` / `find {}` linear scans, it's just the domain rule, correctly stated. It's also symmetric, which matters because `onCheckboxTagSelected` flipped the receiver from `tag` to the list element when it moved over — same answer either way.

The rest of the PR holds up from the last pass: the branch is still properly rebased, and `CoursesViewModel.currentFilterState` plus the `filterCoursesInternal` reuse still fixes the real bug, where a background refresh called `processCourses` with the unfiltered list and silently wiped the user's filter.

Two nits carried over, neither worth another round:

- The shadowed `chipRow` is still there, and slightly more visible now: `val chipRow = requireView().findViewById<LinearLayout>(R.id.chip_filter_row)` at line 225, then the identical `requireView().findViewById(...)` re-declared four lines down inside `if (savedFilter.isActive)`. Since `requireView()` is non-null, both the inner `if (chipRow != null)` and the later `chipRow?.let { … }` are also redundant.
- `CoursesViewModel.kt:148` still ends its catch in `e.printStackTrace()`. Pre-existing, and #16782 landed specifically to remove these.

`buildTagDataList` now does `selectedItemsList.any { it.matches(tag) }` per tag instead of a precomputed set lookup — O(selected × tags) rather than O(tags). Both collections are small enough here that the clarity is worth more than the lookup.

CI green 5/5.

**Verdict:** mergeable.

---

dogi (approved):

PR #16655, opened 2026-09-01
sync: smoother download repository service cancelling (fixes #16654)

### LINKED ISSUE #16654: CancellationException
cancelling the job throws a CancellationException (which prints as "StandaloneCoroutine was cancelled" — the name of the child coroutine Kotlin auto-generates)
Image

Okuro3499:
fixes #16654
Both spots where the network/download suspend calls were caught by an overly-broad catch (e: Exception) now let CancellationException propagate instead of being reported as
  a download failure:                                                                                                                                                                
                                                                                                                                                                                     
  - DownloadService.kt — initDownload's catch block now rethrows CancellationException before falling into the generic handler that was calling downloadFailed(...).                 
  - DownloadRepositoryImpl.kt — downloadFileResponse's catch chain does the same, so a cancelled network call during a stop/superseded-download no longer gets reported as "Network  
    error".          

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved download cancellation handling so canceled downloads stop cleanly without being reported as failures.
  * Prevented cancellation events from triggering misleading error messages or failure callbacks.


<!-- end of auto-generated comment: release notes by coderabbit.ai -->


dogi:
## Review — rating 76/100 → `change`

The pattern is right and this is the fix I'd want. `initDownload` is a `suspend fun` and `downloadFileResponse` wraps its body in `withContext(dispatcherProvider.io)`, so `CancellationException` is genuinely reachable at both sites, and `catch (e: Exception)` was swallowing it — `CancellationException` extends `IllegalStateException` extends `Exception`. Re-throwing ahead of the typed arms is the same taxonomy #16630 established. Both new catches are correctly ordered before the arms they'd otherwise shadow. CI green on all five checks.

Holding at `change` for one thing: the fix stops 25 lines short of finishing the function it targets.

### The inner catch in `initDownload` intercepts cancellation first

```kotlin
// DownloadService.kt:218-224 (master, untouched by this PR)
if (FileUtils.checkFileExist(this, url)) {
    try {
        resourcesRepository.markResourceOfflineByUrl(url)
    } catch (e: Exception) {
        e.printStackTrace()
    }
    onDownloadComplete(url)
    return true
}
```

`markResourceOfflineByUrl` is a suspend call, and this `catch (e: Exception)` sits *inside* the `try` whose handler you just fixed. Cancel a download of a file that's already on disk and the `CancellationException` is absorbed here — it never reaches the new re-throw at line 251. The service then calls `onDownloadComplete`, returns `true`, and the queue proceeds as if the cancelled download had succeeded. That's precisely the failure mode this PR exists to prevent, in the same function.

The same one-line `catch (e: CancellationException) { throw e }` handles it.

### Other reachable sites in this file

For scoping rather than as a blocker — `DownloadService` has `catch (e: Exception)` at lines 328, 396, 494, 536 and 591. Line 494 (`onDownloadComplete`) is the notable one: it currently sits inside `appScope.launch`, so it's insulated today, but **#16637 turns `onDownloadComplete`, `downloadFile` and `tryDownloadFromResult` into suspend functions and inlines that block into the download coroutine** — which reintroduces exactly this bug at line 494 the moment it merges. I flagged it on that PR too. Whichever lands second should sweep the file rather than leaving the two fixes to interleave.

### Merge-order note

#16637 is labelled `merge` and edits `DownloadService.kt` in overlapping territory (its `tryDownloadFromResult` signature change is ~40 lines below your new catch, and it rewrites `onDownloadComplete` wholesale). Expect a conflict depending on ordering. Landing this one first is the cheaper sequence, since #16637 then only needs its own `onDownloadComplete` arm added.

### Minor

No test covers either new catch. A `runTest` that cancels the job mid-`downloadFileResponse` and asserts the `CancellationException` propagates rather than becoming `DownloadResult.Error` would pin the `DownloadRepositoryImpl` half cheaply — that one is easy to test since it's a plain repository with an injected dispatcher.

Add the inner-catch arm and I'd move this to `ready`.

---

dogi:

The inner catch is fixed. `markResourceOfflineByUrl` — the suspend call that sat inside `initDownload`'s already-on-disk branch and would have absorbed cancellation before it could reach the outer handler — now re-throws first:

```kotlin
try {
    resourcesRepository.markResourceOfflineByUrl(url)
    throw e
} catch (e: Exception) {
    e.printStackTrace()
}
```

That closes the hole I flagged: cancelling a download of a file already present no longer completes as a success and lets the queue advance. Both handlers in the function now agree, and with `DownloadRepositoryImpl.downloadFileResponse` covered too, the two suspend boundaries this PR set out to protect are protected. CI green on all five checks.

The remaining `catch (e: Exception)` sites in `DownloadService` (328, 396, 494, 536, 591) are still open, as noted last round — scoping, not this PR's job. Line 494 is the one to keep an eye on: #16637 is now labelled `merge` + `automerge`, and it converts `onDownloadComplete` into a suspend function called from the download coroutine, which puts that catch on a live cancellation path. I've flagged it there; whichever lands second inherits the sweep.

Still no test on either new catch. Not blocking given how mechanical the change is, but `DownloadRepositoryImpl` remains the cheap one to pin — plain repository, injected dispatcher, no service lifecycle — and a `runTest` that cancels mid-`downloadFileResponse` and asserts the `CancellationException` propagates rather than becoming `DownloadResult.Error` would lock the contract in place for the next person who adds an arm to that chain.

---

Issue #16479, opened 2026-08-27
Quiet JsonUtils.safeGet on the hot parse path


dogi:
**rating:** 49/100 &nbsp;·&nbsp; **proposed by:** copilot(r2)

Every getString/getBoolean/getInt/getLong/getJsonObject routes through `safeGet`, which calls `e.printStackTrace()` on any exception. Sync and deserialization hit these thousands of times, so stack traces on expected type mismatches dominate log I/O.

**Ask:** stop `safeGet` from dumping a stack trace on every expected type mismatch — the fallback return is the contract, so the exception path should be quiet (or at most a guarded debug log), not `printStackTrace()`.

**Files:** `utils/JsonUtils.kt`, `test/.../JsonUtilsTest.kt`

**Verification:** Verified at JsonUtils.kt:16-23.

PR #16656, opened 2026-09-01
all: smoother json utils fallback handling (fixes #16652)

### LINKED ISSUE #16652: JsonUtils.safeGet can throw from its own fallback path
## Problem

`JsonUtils.safeGet` is the shared "return a default instead of throwing" wrapper behind every `JsonUtils` accessor. Its catch block logs, and the log call itself can throw:

```kotlin
private inline fun <T> safeGet(default: () -> T, block: () -> T): T {
    return try {
        block()
    } catch (e: Exception) {
        if (Log.isLoggable(TAG, Log.DEBUG)) {      // <-- throws in plain JVM unit tests
            Log.d(TAG, "expected type mismatch, using fallback: ${e.message}")
        }
        default()
    }
}
```

An exception raised *inside* a catch block is not caught by that same `try`. So on the fallback path `safeGet` propagates instead of returning the default — the exact opposite of its contract. In a plain JUnit test (no Robolectric, no `mockkStatic(Log::class)`) `android.util.Log.isLoggable` throws `RuntimeException: Method isLoggable in android.util.Log not mocked`.

The guard came from #16479 / #16494 and is right on its own terms — the problem is only its placement inside the recovery path.

## Why this matters beyond tests

It produces failures that point at the wrong code. Concrete case from #16609: `HealthExamination.formatConditions` filters keys with `JsonUtils.getBoolean`, and a `JsonObject` value makes `asBoolean` throw. `safeGet` catches, its own catch throws, that escapes `getBoolean` and the `filter`, and `formatConditions`' outer catch swallows it and returns `""` — silently discarding **every** condition.

The test that existed to prove exactly that bug absent failed with `expected:<[diabetes]> but was:<[]>`, which reads as a logic error in the PR. The logic was correct. Roughly an hour went into a diff that was never wrong.

Any future pure-unit test touching a `JsonUtils` type mismatch inherits the same misdirection.

## Second, related defect: exceptions as control flow

`getString(fieldName, jsonObject)` type-checks properly and never reaches the catch:

```kotlin
if (el is JsonNull || !el.isJsonPrimitive || !el.asJsonPrimitive.isString) "" else el.asString
```

Five accessors do not, and rely on the throw for ordinary wrong-typed input:

| accessor | how it reaches the catch |
|---|---|
| `getBoolean` | `el?.asBoolean` throws for `JsonObject`/`JsonArray`/non-boolean string |
| `getLong` | `el.asLong` throws for non-primitives |
| `getInt` | `el.asString` in the middle branch throws for `JsonObject` |
| `getFloat` | same as `getInt` |
| `getString(array, index)` | `el.asString` throws for `JsonObject`/`JsonArray` |

A wrong JSON type from CouchDB is a normal condition, not an exceptional one. Building a stack trace per row is also the cost #16479 set out to remove, and it is what makes the defect above reachable at all.

## Proposed fix

1. Move the log out of the control-flow path so the fallback cannot throw — logging must never change control flow.
2. Type-check the five accessors above the way `getString` already does, so ordinary malformed input returns the default without an exception.

Both are contained within `JsonUtils.kt`. Existing `JsonUtilsTest` coverage should keep passing unchanged; the fix adds cases for wrong-typed values on each accessor, and one that asserts the accessors work with no `Log` stubbing at all (which is what regressed here).

## Not proposed

Removing the `isLoggable` guard. It is doing its job on the hot path; it just needs to sit where a throw cannot escape.

dogi:
`safeGet` is the shared "return a default instead of throwing" wrapper behind every `JsonUtils` accessor, and its catch block logged through `android.util.Log`:

```kotlin
} catch (e: Exception) {
    if (Log.isLoggable(TAG, Log.DEBUG)) { ... }   // <-- can throw
    default()
}
```

An exception raised *inside* a catch is not caught by that same `try`, so on the fallback path `safeGet` propagated instead of returning the default — the opposite of its contract. In a plain JVM test `Log.isLoggable` throws `not mocked`.

## What it cost

In #16609 a `JsonObject` sibling value made `getBoolean` throw, `safeGet`'s catch threw in turn, and the caller's outer catch swallowed it and returned `""` — discarding **every** condition. The test written to prove exactly that bug absent failed as `expected:<[diabetes]> but was:<[]>`, reading as a logic error in a diff that was correct.

## Changes

1. Logging moved behind a guarded `logFallback()`, so it can never change control flow.
2. `getBoolean`, `getInt`, `getFloat`, `getLong` and `getString(array, index)` type-check before coercing. `getString(fieldName, obj)` already did and is untouched. A wrong JSON type is a normal condition on CouchDB-sourced data, not an exceptional one — these five leaned on the throw, which is what made the defect above reachable and is the per-row cost #16479 set out to reduce. `getString(array, index)` also gained bounds checking.

## Verification

`JsonUtilsCoercionTest` pins the exact coercions for every accessor across missing keys, nulls, objects, arrays and wrong-typed strings. It was written against the **pre-fix** code and confirmed passing there, then required to pass unchanged after — so behaviour preservation is measured, not asserted.

`JsonUtilsNoLogStubTest` stubs nothing at all, so every case runs the fallback path the way a pure unit test does, including the mixed-values shape from #16609. It carries a note not to add `Log` stubbing, since that would make it pass against the bug it exists to catch.

Full unit suite green locally (214 classes, 1530 tests), and green again with current master merged (219 classes, 1551 tests).

## One existing assertion changed, deliberately

`JsonUtilsTest.testTypeMismatchesAreQuiet` expected the old throw-and-recover path — `isLoggable` consulted, one debug line per value. Those calls no longer reach the catch, so it now asserts the stronger property: nothing is logged at all. Flagging it because it rewrites an existing expectation rather than satisfying it.

## Note

The test-coroutine-scoping fix that was briefly on this branch has moved to its own PR — it is an unrelated, test-only change and does not need to wait on review of this one.

Fixes #16652


dogi:

Strong work. Two distinct improvements land here and both are worth having.

The first is the actual #16652 bug: `safeGet`'s handler called `Log.isLoggable` *inside the catch block*, so in a plain JUnit test without `returnDefaultValues` the "not mocked" `RuntimeException` escaped from the fallback path itself — the recovery mechanism threw instead of recovering. Wrapping it in `logFallback` fixes that, and `JsonUtilsNoLogStubTest` is the right shape of regression test. The comment on it is the best line in the diff:

> DO NOT add Log stubbing here. This class must run without it -- stubbing Log would make it pass against the bug it exists to catch.

That's a test that documents its own defeat condition. More tests should do this.

The second is the rewrite of the accessors to type-check up front instead of relying on `asInt`/`asLong`/`asBoolean` throwing. Exceptions stop being control flow on the hot sync path, and the `verify(exactly = 0) { Log.isLoggable(...) }` flip in `JsonUtilsTest` pins that the catch is genuinely no longer entered. `getInt`'s `"3.7"` case is a nice demonstration: it used to reach `Integer.parseInt("3.7")`, throw, and unwind to the default; now `toIntOrNull()` returns null and the same `0` comes back without the throw. The two new test classes cover every accessor across primitive, string-coercible, wrong-type, null, missing and null-receiver inputs — that's real coverage, not decoration. CI green on all five checks.

One correction to the framing, and two smaller notes.

### This is not strictly behaviour-preserving — single-element arrays diverge

The `JsonUtilsCoercionTest` header says "so #16652 can be verified as behaviour-preserving". It nearly is, but Gson's container accessors have a special case the old code reached and the new code doesn't: `JsonArray.getAsInt()` / `getAsString()` / `getAsLong()` / `getAsBoolean()` all delegate to element 0 when `size() == 1`, throwing only otherwise.

So for `{"k": ["5"]}`:

- **`getInt`** old — not primitive, not `JsonNull`, `el.asString` → `"5"`, non-empty → `else -> el.asInt` → **5**. New — `!isJsonPrimitive` → **0**.
- **`getLong`** old — `el.asLong` → **5**. New → **0**.
- **`getBoolean`** on `{"k": ["true"]}` old — `el.asBoolean` → **true**. New → **false**.
- **`getFloat`** and **`getString(array, index)`** on a nested single-element array: same shift.

The tests exercise `JsonArray()` — *empty*, which throws under both and returns the default either way — so the one input that actually changed is the one input not covered. I think the new behaviour is the better of the two (silently unwrapping a one-element array into a scalar is a footgun, and it's asymmetric with two-element arrays), but it should be a stated decision with a test pinning it, not an unremarked side effect of the refactor. Worth a quick check that no sync path feeds single-element arrays into these — CouchDB documents do sometimes carry `["value"]` shapes.

### The two `getString` overloads now disagree, and the tests enshrine it

`getString(fieldName, jsonObject)` (untouched here) requires `isString`, so `{"k": 7}` → `""`. The new `getString(array, index)` accepts any primitive, so `[7]` at index 0 → `"7"`. Your own tests assert both:

```kotlin
assertEquals("", JsonUtils.getString("k", obj { addProperty("k", 7) }))   // by field
assertEquals("7", JsonUtils.getString(arr, 1))                            // by index
```

Same name, same input type, opposite answers — now locked in by tests, which makes it harder to fix later. Since this PR is already normalising these accessors, aligning the two (whichever way) would be the moment.

### `catch (_: Throwable)` vs the established pattern

`logFallback` catches `Throwable`. The failure it's defending against is a `RuntimeException` from unmocked `android.util.Log`, with `NoClassDefFoundError` as the plausible second case — which is exactly the `Exception` + `LinkageError` pair `MainApplication.runBestEffort` settled on in #16630. Matching that keeps one taxonomy in the codebase, and it avoids swallowing `OutOfMemoryError` in a debug-logging helper. There's no coroutine here so cancellation isn't a concern, which is why this is a consistency note and not a correctness one.

None of the above blocks merging.

---

Okuro3499 (approved):

PR #16658, opened 2026-09-01
resources: smoother repository file copying (fixes #16657)

### LINKED ISSUE #16657: local resourc not working
when I add a resource in my library it shows that the file exists offline but it only shows this placeholder and not the resource I added
Image

Okuro3499:
fixes #16657
When adding a personal resource, copy the source file into the app's external files directory instead of just referencing the original path. Sets resourceLocalAddress and filename to the copied file's name.
[Screen_recording_20260901_173836.webm](https://github.com/user-attachments/assets/2963c807-3f47-464b-b3bf-98c2a530a47c)


dogi:
## Review — rating 62/100 → `change`

The diagnosis is right and this is a real bug worth fixing. Storing the picker's absolute path in `resourceLocalAddress` was broken by construction: every reader resolves that column through `FileUtils.getLibraryFile(externalFilesDir, id, address)` → `File(dir, "ole/$id/$address")`, so an absolute source path could never resolve, and the resource died the moment the user moved or deleted the original file. Copying into `ole/<id>/<filename>` and storing the bare filename makes the stored value match how it's read everywhere else. The PR also quietly fixes a second bug: `it.substring(it.lastIndexOf("/"))` kept the leading slash, so `filename` was `"/report.pdf"` rather than `"report.pdf"`. Both good catches. CI green on all five checks.

Two problems with the implementation.

### The copy runs on the main thread

`saveLocalResource` is a plain `suspend fun` with no `withContext`. That was fine when its body was only DAO calls — Room dispatches those off-main itself — but this PR adds a synchronous `File.copyTo` to it. The only call site is:

```kotlin
// AddResourceActivity.kt:193
lifecycleScope.launch {
    val result = resourcesRepository.saveLocalResource(request)
```

`lifecycleScope` dispatches on `Dispatchers.Main`, so the copy happens on the UI thread. For the media types this screen accepts — the spinner offers video — that is a multi-hundred-megabyte blocking copy on main, i.e. a guaranteed ANR, with no progress indication.

`ResourcesRepositoryImpl` already injects `dispatcherProvider`, and `DownloadRepositoryImpl` is the house pattern for exactly this:

```kotlin
override suspend fun downloadFileResponse(...) = withContext(dispatcherProvider.io) { ... }
```

Wrapping at least the copy in `withContext(dispatcherProvider.io)` is the fix.

### `resourceUrl` is not always a filesystem path

The new gate assumes it is:

```kotlin
val sourceFile = request.resourceUrl?.let { File(it) }
if (sourceFile == null || !sourceFile.exists()) {
    return Result.failure(Exception("Resource file not found"))
}
```

Two producers feed that extra, and they put different things in it:

- `AddResourceFragment.kt:260` passes `FileUtils.getRealPathFromURI(requireContext(), uri)` — a real absolute path. This is the flow the PR targets, and it works.
- `ResourcesFragment.kt:347` (`openEditResource`) passes `model.library.resourceLocalAddress` — which, after this PR, is a **bare filename**. `File("report.pdf").exists()` resolves against the process working directory and returns false, so edit-mode saves now fail with "Resource file not found".

I'm not claiming this PR breaks a working feature — edit mode already routes into `saveLocalResource`, which starts with `resourceTitleExists(title)` and would reject the unchanged title as a duplicate, so that flow looks broken on master too. But this change hardens an assumption that one of the two callers violates, which makes the eventual edit-mode fix harder. Worth either handling the already-in-`ole/` case (if the source is already the destination, skip the copy) or making the edit path stop reusing this extra for a different meaning.

Related: `getRealPathFromURI` returns null for content URIs it can't resolve to a path (cloud providers, Drive). Those now surface as "Resource file not found", which misdescribes the failure — the file exists, it just isn't reachable as a `File`. Copying from the `Uri` via `contentResolver.openInputStream` would handle both that and the SAF case properly, though that's a larger change than this PR needs to take on.

### Smaller

- If `saveLibraryItem` fails after the copy, the copied file is orphaned in `ole/<uuid>/` with no row referencing it. A `runCatching` around the save that deletes the destination on failure would close that.
- Only `IOException` is caught; a `SecurityException` reading the source escapes to the caller as an unhandled throw rather than a `Result.failure`.
- No test, and `ResourcesRepositoryImplTest` is right there (and being actively edited by #16659). Asserting that the file lands at `ole/<id>/<name>` and that `resourceLocalAddress` is the bare name would pin the whole contract, using `TemporaryFolder` as `ConfigurationsRepositoryImplTest` does.

Move the copy off the main thread and I'd move this to `ready`; the caller mismatch is worth a decision but doesn't have to be solved here.

---

dogi:
## Re-review — rating 82/100 → `ready`

Every blocking point from the last round is addressed, and the revision picked up the two smaller ones as well.

**The copy is off the main thread.** `withContext(dispatcherProvider.io)` wraps both the `mkdirs()` and the `copyTo`, so the multi-hundred-megabyte video case no longer blocks the UI thread from `lifecycleScope`. Note the `try` sits *outside* the `withContext`, catching only `IOException` and `SecurityException` — which means a cancellation during the copy propagates correctly rather than being converted into a `Result`. That's the right nesting.

**`SecurityException` is handled**, so an unreadable source path returns `Result.failure` instead of throwing out of the repository.

**The orphan is cleaned up.** If the DAO write fails after the file has landed, `destinationFile.delete()` runs before the failure returns, so a failed save no longer leaves an unreferenced file in `ole/<uuid>/`.

**And the tests exist.** `TemporaryFolder`, a real source file, a real copy, and assertions on all three things that matter — `resourceLocalAddress` and `filename` are the bare name, the file actually lands at `ole/<id>/report.pdf`, and its contents survive the copy (`assertEquals("content", destinationFile.readText())`). The second test pins the missing-source failure and verifies no row is written. That's a genuine contract test, not a smoke test.

Two follow-ups, neither blocking.

### The new `saveLibraryItem` catch swallows cancellation

```kotlin
try {
    saveLibraryItem(resource)
} catch (e: Exception) {
    destinationFile.delete()
    return Result.failure(e)
}
```

`saveLibraryItem` is a suspend call and `CancellationException` is an `Exception`, so cancelling the activity mid-save is caught here, converted to a `Result.failure`, and execution continues into the caller's success/failure branch — which then touches UI on a destroyed activity. Same taxonomy as #16630, and as it happens the same fix that #16655 is landing next door in this batch: a `catch (e: CancellationException) { throw e }` arm ahead of it. Deleting the file on cancellation is reasonable cleanup, so a `try/catch(Cancellation) { delete(); throw }` would keep both behaviours.

### The happy-path test stubs the function it's checking

`mockkObject(FileUtils)` is needed for `getExternalFilesDir` (it memoises into a static), but the test also stubs `getLibraryFile` with a re-implementation of the same path expression:

```kotlin
every { FileUtils.getLibraryFile(externalFilesDir, any(), "report.pdf") } answers {
    File(externalFilesDir, "ole/${secondArg<String>()}/report.pdf")
}
```

`getLibraryFile` is a pure function, and `mockkObject` leaves unstubbed members calling through to the real implementation — so dropping that stub would have the test assert against the *actual* layout rather than against a copy of it. As written, if `getLibraryFile`'s path scheme ever changed, this test would keep passing while every reader broke.

The `ResourcesFragment.openEditResource` caller mismatch is unchanged, as expected — I flagged it as a decision rather than a blocker, and it's pre-existing.

---

PR #16660, opened 2026-09-01
teams: smoother submissions repository stamping (fixes #16664)

### LINKED ISSUE #16664: surveys team data missing on planet
Image

In exported survey submissions, the team name is currently not displayed in the pdf/csv.

Image

The team object is what is missing from the database

Submission synced from myPlanet:
https://planet.learning.ole.org:2200/_utils/#database/submissions/92bd3e84b2be8c6b02261ca642202b2e

Planet generated submission:
https://planet.learning.ole.org:2200/_utils/#database/submissions/92bd3e84b2be8c6b02261ca6421ee0f3

Mutugiii:
Fixes #16664

Survey responses uploaded from myPlanet carried no top-level `team`, so Planet's `submissionsByParent` view could not attribute them to the team they were recorded for and they never counted towards that team's survey submission count.

Two gaps caused it:

- `createExamSubmission` set `teamObject` and `membershipDoc`, but both are `@Ignore` on the `Submission` entity, so Room dropped them. The persisted `teamId` column was only ever written by the sync path on the way down, so the team linkage did not survive the local write.
- `serializeSubmission` — the serializer behind every completed survey upload — emitted no `team` key at all. The sibling `getExamUploadPayload` does, but it is wired to a `type = 'exam'` query that surveys never reach.

Persist `teamId` at creation time and resolve it back into a top-level `team` object at serialization time, preferring the in-memory `teamObject` when present. Submissions with no team serialize exactly as before.

Refs open-learning-exchange/planet#10232

Claude-Session: https://claude.ai/code/session_0139Kx1eT5JYa9ioYLBVXrg2

## Summary by CodeRabbit

* **Bug Fixes**
  * Exam submissions now retain their associated team ID even when local team details cannot be resolved.
  * Team information is serialized more reliably, using available submission data and enriching it with local metadata when possible.
  * Submissions without a team continue to omit team details.
* **Tests**
  * Added coverage for team ID persistence and serialization across available, missing, failing, and in-memory team data scenarios.


dogi:

The diagnosis is exactly right, and I checked the premise rather than taking it on trust. In `Submission`:

```kotlin
@Ignore var teamObject: TeamReference? = null,   // line 20
@Ignore var membershipDoc: MembershipDoc? = null, // line 28
var teamId: String? = null,                       // line 29 — persisted
```

So the team association really did evaporate on the first Room round trip: `createExamSubmission` populated `teamObject` and `membershipDoc`, both `@Ignore`, and by upload time the submission had no idea which team it belonged to. Writing the id into the one column that survives is the minimal correct fix — and because `teamId` already exists on the entity, this needs no schema change and therefore no `AppDatabase` version bump. Worth saying explicitly given how often that line has been the problem lately.

**The round-trip test is the best thing in the diff:**

```kotlin
// Reconstruct the entity as Room would: @Ignore fields are absent, while teamId survives.
val reconstructed = Submission().apply {
    id = persisted.id; userId = persisted.userId
    parentId = persisted.parentId; type = persisted.type
    teamId = persisted.teamId
}
```

Capturing what `upsertAll` was handed and then deliberately rebuilding it minus the `@Ignore` fields reproduces the actual failure mode in a plain unit test, without an in-memory database. That's the assertion that would have caught this bug in the first place, and it's the one that stops it coming back.

The six cases cover the matrix properly — team absent locally, lookup throws, lookup succeeds and enriches, in-memory `teamObject` short-circuits the lookup (with `coVerify(exactly = 0)` proving it), and no team at all omits the field. And `getTeamByIdOrNull` re-throws `CancellationException` ahead of the catch-all, which is the taxonomy this codebase has been converging on.

Three things, none blocking.

### One team lookup per submission on the upload path

`resolveTeamJson` calls `getTeamByIdOrNull` whenever `name` or `type` is missing from the `TeamReference` — which, for exactly the case this PR fixes (a submission reloaded from Room, so `teamObject` is null), is *always*. `serializeSubmission` runs per submission inside `UploadCoordinator`'s batches of 50, so a team-scoped survey upload becomes one extra DAO round trip per row. A small per-serialize cache keyed by team id, or resolving the distinct team ids once before the batch, would collapse that to one query per team.

### The `"team"` type default quietly disappeared

```diff
-                    this.type = team.type ?: "team"
+                    this.type = team?.type
```

Master always produced a non-null `type` on `teamObject`; now a locally-known team with a null `type` yields none, and `resolveTeamJson` omits the property entirely rather than defaulting. Since `serializeSubmission` never emitted a `team` object before this PR there's no regression, but the `?: "team"` fallback presumably existed because something wanted a value. Worth confirming Planet is happy with a `team` object carrying only `_id`, which is what the "local team missing" tests assert we now send.

### Merge-order note

#16663 edits `serializeSubmission` in the same file about 25 lines above your new `resolveTeamJson(submission)?.let { ... }` line (it replaces the `androidId` write at what is line 833 on master). Close enough that whichever lands second may need a hand-resolve.

Minor: the `mockkObject(NetworkUtils)` plus three `every` stubs are repeated verbatim in all six new tests — a `@Before` would carry them.

---

Okuro3499 (changes requested):
```
Problem: This is the serializer wired to the ExamResults upload config (UploadConfigs.kt:236-249), which sources submissions via submissionDao.getPendingExamResults() — a SELECT
    * FROM submissions WHERE type = 'exam' ... query (SubmissionDao.kt:40). Those are fresh-from-Room entities, so teamObject is always null there for the same reason surveys were  
    broken (@Ignore). getExamUploadPayload still only checks if (submission.teamObject != null) and was not updated to fall back to the persisted teamId the way resolveTeamJson does
    for serializeSubmission. Real team exams (ExamTakingFragment.kt:118-126, type == "exam" with isTeam/teamId) go through exactly this path, so this PR's premise — "responses      
    recorded for a team now count toward that team" — is still false for actual exam submissions, only fixed for surveys. The commit message's claim that "the sibling               
    getExamUploadPayload does [emit team]" is misleading: it only appears to work because it was never exercised with a Room-reloaded entity in the existing test suite
```

Mutugiii:
> ```
>  ...so this PR's premise — "responses recorded for a team now count toward that team" — is still false for actual exam submissions, only fixed for surveys. The commit message's claim that "the sibling               
>     getExamUploadPayload does [emit team]" is misleading: it only appears to work because it was never exercised with a Room-reloaded entity in the existing test suite
> ```


Currently, we don't support exams/tests in teams, only surveys. So the premise only needs to be correct for surveys.

Okuro3499 (approved):

dogi:
## Re-review — rating 88/100 → `ready`

This revision closes the gap Okuro found, and it's a gap I should have caught in my first review — I checked `serializeSubmission` and stopped there without following the exam upload path.

**`getExamUploadPayload` now resolves the team the same way.** That matters more than the survey fix did: `UploadConfigs.ExamResults` sources rows through `submissionDao.getPendingExamResults()`, which is a plain Room query, so `teamObject` is *always* null on that path — `@Ignore` guarantees it. The old `if (submission.teamObject != null)` block could therefore never fire for a real exam upload, and team exams would have stayed broken while surveys were fixed. Routing both serializers through `resolveTeamJson` makes the fix actually match the PR's premise.

The six new `getExamUploadPayload` tests mirror the survey set exactly — missing local team, lookup throws, local metadata enriches, in-memory `teamObject` short-circuits (`coVerify(exactly = 0)`), and no team omits the field. Twelve cases across the two serializers, covering the same matrix on both. That's the right way to prove two call sites converged.

The `TextUtils.isEmpty(...)` → `isNullOrEmpty()` swap is a good incidental: `android.text.TextUtils` is one of the framework stubs that throws "not mocked" in plain JUnit, so dropping it removes a testability landmine from this file rather than just modernising syntax.

Two things carried over from my last review, neither blocking:

**The team lookup is still per submission**, and now on both paths. `resolveTeamJson` calls `getTeamByIdOrNull` whenever `name` or `type` is missing from the `TeamReference` — which for Room-reloaded rows is always, i.e. exactly the case this PR exists to handle. Inside `UploadCoordinator`'s batches of 50 that's one extra DAO round trip per row. Resolving the distinct team ids once per batch, or a small per-serialize cache, would collapse it.

**The `?: "team"` type default is still gone.** `this.type = team?.type` replaces `team.type ?: "team"`, and `resolveTeamJson` omits `type` entirely when null — which the "local team missing" tests now assert as the expected payload. No regression, since no `team` object was emitted before this PR, but worth confirming Planet accepts a `team` carrying only `_id`.

Minor: the `mockkObject(NetworkUtils)` + three stubs are now repeated in twelve tests. A `@Before` would carry them and make the next serializer test cheaper to add.

---

dogi (approved):

PR #16661, opened 2026-09-01
resources: smoother repository cached models view modelling (fixes #16632)

### LINKED ISSUE #16632: library: slow library opening
When we try to open Library, the loading is slow/delay, not immediate or fast.

https://github.com/user-attachments/assets/f0e14271-8750-414e-aff6-d01ababd114b

ragilzakaria:
https://github.com/user-attachments/assets/84bdcdb9-348f-4964-bdcb-34826a78140c

fixes #16632 

## Summary by CodeRabbit

* **New Features**
  * Resource lists now load through shared state, displaying cached results immediately while refreshed data loads.
  * Resource sorting and filtering update more reliably as data changes.
  * Added faster resource-list retrieval through improved local data indexing and caching.

* **Bug Fixes**
  * Resource-list caches are cleared automatically after library or offline-status updates.
  * Removed outdated rating data from resource-list displays.


dogi:
## Review — rating 76/100 → `change`

Good performance work, and the part I expected to object to turned out to be the best thing in it. CI green on all five checks.

**The `AppDatabase` version bump is correct here** — and worth saying plainly, because I've flagged the opposite mistake twice this week. This PR adds `Index("isPrivate")` to `MyLibrary` and `Index("linkId")` + `Index(["db","linkId"])` to `TagEntity`, which changes Room's identity hash, so `version = 9 → 10` is exactly right. (#15951 and #15699 change entities *without* bumping and crash every existing install; #16646 bumps *without* a schema change and wipes every install. This one is the case where the bump is genuinely owed.)

**Dropping ratings from the list query is right, and I checked before saying so.** My first read was that deleting `getResourceRatingsBulk` and passing `null` silently removed ratings from the resources list. It doesn't: `ResourceListModel.rating` is constructed in exactly one place — the line this PR changes — and read **nowhere**. `ResourceDetailFragment` fetches its own via `ratingsRepository.getRatingsById("resource", library.resourceId, userModel?.id)` and never touches the list model's copy. So `ratingsRepository.getResourceRatings(userId)` was running on every list load to populate a field with no consumers. Removing it is a real win, not a regression.

Three things to clean up before this lands.

### The now-permanently-null field is a trap

`ResourceListModel.rating: JsonObject?` is still in the data class, still in the constructor, and is now always `null` by construction. Nothing reads it today, but a `JsonObject?` named `rating` on a resource model is exactly the field someone reaches for next time a rating needs displaying in a list — and it will silently be null. Since you've established it's dead, delete it from `ResourceListModel` and drop the parameter. That turns a latent trap into a compile error for anyone who needs ratings back, which is the outcome you want.

### Three unused imports

`kotlinx.coroutines.async` and `kotlinx.coroutines.coroutineScope` are added to `ResourcesRepositoryImpl` and never used anywhere in the diff. And `kotlin.math.ceil` — used on master only at line 696, inside the `getResourceRatingsBulk` this PR deletes — is now orphaned but left in place. All three are warnings rather than errors, which is why CI is green.

The two added imports are the interesting ones: they suggest an earlier attempt to *parallelise* the ratings fetch before settling on removing it. Removing it was the better call; the imports just didn't follow.

### The cache is unsynchronised shared mutable state

```kotlin
private var cachedMyCourseLibModels: Pair<String?, List<ResourceListModel>>? = null
private var cachedPublicLibModels: Pair<String?, List<ResourceListModel>>? = null
```

These are written from `getResourceListModels`, which runs on IO, and read from `getCachedResourceListModels`, which is **non-suspend** and is called from `ResourcesViewModel.getCachedResources` on the main thread during `getAdapter()`. Reference assignment is atomic on the JVM so you won't see a torn value, but without `@Volatile` there's no happens-before edge, so the reader can observe a stale cache (including `null` right after a `clearResourceListCache()`) indefinitely. `@Volatile` on both fields costs nothing and makes the invalidation actually mean something.

Two smaller notes on the same mechanism:

- `clearResourceListCache()` is wired into eight mutation paths, but not `saveLibraryItem` — the path `saveLocalResource` uses to add a new local resource. The damage is limited because `loadResources` always re-queries and overwrites the state, so a stale cache only affects the first paint. Still, that's the one mutation a user performs *and then immediately looks at the list* for.
- A non-suspend read on a repository interface is a new shape for this codebase, where reads are `suspend` or `Flow`. It's defensible for a pure in-memory lookup, but it's worth a comment on the interface saying so, since the whole point is that it must not block.

### Index selectivity

`Index("isPrivate")` is on a boolean column. Two distinct values across the whole table means SQLite will usually prefer a scan anyway, so it mostly buys write cost on the bulk-upsert sync path rather than read speed — unless the private rows are a small minority and the planner is picking it up in a real query plan you've measured. Similarly, `Index(["db","linkId"])` already serves any query filtering on `db` alone or `db + linkId`; the separate `Index("linkId")` only earns its keep for `linkId`-only lookups. Worth confirming each of the three against an actual query before paying for them on every insert.

The ViewModel tests are a good addition — `loadResources asynchronously updates resourcesState` in particular pins the new async path. Nothing covers cache invalidation, which is where the subtle bugs in this design will live.

---

dogi:
## Re-review — rating 88/100 → `ready`

Every point from the last round is addressed, and three of them went further than I asked.

**The null field is gone, not documented.** I suggested deleting `ResourceListModel.rating` rather than leaving a permanently-null `JsonObject?`; the revision removes it from `ResourceListModel`, removes it from `LibraryWithMetadata` too, drops the now-unused `JsonObject` import, and updates all six construction sites across three test files. That converts the trap into a compile error for anyone who needs list ratings back, which is the outcome worth having.

**`@Volatile` on both cache fields**, so the invalidation actually establishes a happens-before edge for the non-suspend reader on the main thread.

**`kotlin.math.ceil` removed**, and the `async` / `coroutineScope` imports from the previous revision are gone with it.

**`saveLibraryItem` now clears the cache.** That was my "the one mutation a user performs and then immediately looks at the list for" note — it's the shared write path for `saveLocalResource`, so adding a local resource now invalidates.

**The interface method carries the KDoc** explaining why it's non-suspend and what it's for, which is what a new shape in this codebase needs.

**And the cache is now tested.** `getResourceListModels caches models and getCachedResourceListModels returns them` walks the full round trip — null before load, populated after, null again after `clearResourceListCache()` — and `saveLocalResource clears in-memory resource list cache` pins the invalidation path end to end. That was the gap I flagged as "where the subtle bugs in this design will live".

Two leftovers, neither blocking.

`ResourcesViewModelTest` still has `import com.google.gson.JsonObject` at line 3, but this revision deletes the `val mockRating = mockk<JsonObject>(relaxed = true)` that was its only use — so it's now unused. Same category as the imports cleaned up elsewhere in this diff; the repo's `kotlin-importing` skill would catch it.

The index-selectivity question from last round stands, and it's genuinely a question rather than a finding: `Index("isPrivate")` is on a two-valued boolean column, where SQLite will usually prefer a table scan, so it may cost write throughput on the bulk-upsert sync path without buying read speed. And `Index(["db", "linkId"])` already serves queries filtering on `db` or on `db + linkId`, leaving the standalone `Index("linkId")` to earn its keep only on `linkId`-alone lookups. Worth checking each against a real query plan at some point, but not something to hold this on.

The version bump remains correct — these index changes alter Room's identity hash, so `9 → 10` is owed here.

---

Okuro3499 (dismissed):
```
1. ResourcesRepositoryImpl.kt — updateUserLibrary (~line 296–314) doesn't invalidate the cache.                                                                                    
This is the backing method for setUserLibrary, which ResourceDetailFragment.kt:61,239 calls when a user adds/removes a single resource from "My Library" from the resource detail  
screen. It upserts myLibraryDao but never calls clearResourceListCache(). If a new ResourcesFragment/ResourcesViewModel instance is created afterward (e.g., user backs out and    
reopens the Library tab), getAdapter()'s cache-hit path (ResourcesFragment.kt:176-187) will paint the stale membership before the background loadResources() call corrects it — a  
visible flash of wrong "My Library" contents.                                                                                                                                                                                                                    
                                                                                                                                                                                     
2. ResourcesRepositoryImpl.kt — markResourceOfflineByLocalAddress/markResourceOfflineByResourceId/reconcileHtmlResourceOffline (~line 334–373) don't invalidate the cache.         
markResourceOfflineByUrl (which these back) is called from DownloadService.kt:221,492 and DownloadWorker.kt:89,141 — the core "download finished" path, hit on essentially every   
resource download. Since it mutates resourceOffline/downloadedRev without clearing the cache, a fresh fragment instance's cache-hit paint can show a just-downloaded resource as   
not-yet-offline (wrong sort position, since getResourceListModels sorts by isResourceOffline(), and a stale offline badge).                                                        
                                                                                                                                                                                     
3. ResourcesRepositoryImpl.kt:47 — unused ratingsRepository constructor parameter.                                                                                                 
getResourceRatingsBulk was removed, but private val ratingsRepository: RatingsRepository is still injected and now has zero usages. This contradicts the commit's own stated intent
("drop dead rating field") — the cleanup was left half-done.                                                                                                                       
```

dogi:
**Review — rating: 82/100**

The headline change is the right one and it's bigger than the cache: `getEnrichedLibraries` was calling `getResourceRatingsBulk`, which pulled **every** resource rating out of `RatingsRepository` and built a `HashMap` of `JsonObject`s to hang one off each `ResourceListModel` — and `ResourceListModel.rating` had no consumers at all. Dropping the field, the bulk query, and the `RatingsRepository` dependency from this repository removes a whole extra query plus N `JsonObject` allocations from every library load. That the build is green with the field deleted is the proof it was dead weight.

The in-memory cache is done carefully. Two single-slot `@Volatile` `Pair<String?, List<…>>` keyed by `modelId`, with `clearResourceListCache()` wired into roughly a dozen `myLibraryDao` write sites — the shelf add/remove paths, both batch inserts, the offline-marking paths, the stale-public delete, `markResourcesAsNotOffline`. That's thorough rather than hopeful. Exposing a non-suspend `getCachedResourceListModels` on the repository interface is a deliberate break from the suspend/Flow convention, and the KDoc says why (synchronous first paint on tab switch) instead of leaving the next reader to guess. I'd take that trade.

Three things worth attention before this lands:

**1. The `AppDatabase` version bump is the expensive part of this PR, and it's for three indices.** `version = 11 → 12` with `fallbackToDestructiveMigration(true)` means every device that installs this release drops its local database and re-pulls from CouchDB on the next sync — and per `CLAUDE.md`, that also discards **unsynced local writes**: pending uploads, drafts, anything not yet pushed. The indices themselves (`MyLibrary.isPrivate`, `TagEntity.linkId`, composite `(db, linkId)`) are well-chosen for the queries this PR leans on. But paying a destructive rebuild for a pure performance optimisation is a release-coordination decision, not a code one. If another schema change is already queued, this wants to ride along with it rather than spend the bump on its own.

**2. A new `e.printStackTrace()` in `ResourcesViewModel.loadResources`.** #16782, in this same queue, exists to delete these. Worth a `Log.e(TAG, "...", e)` before it becomes the one that got away.

**3. The cache holds mutable objects that the UI also mutates.** `ResourceListModel.isOpened` and `isLocallyOffline` are `var`, and the cached list hands out the same instances the fragment renders — so UI-side mutations write through into the cache. That may well be intentional (state surviving a tab switch), but it means "cache" here is shared mutable state between the repository and the view layer, which is worth a comment. Related: the two lists are held on a `@Singleton` for the process lifetime with no size bound; for a large library that's a permanent retention. `onTrimMemory` or a cap would close it.

Smaller things that check out: moving `toggleSortOrder`/`toggleTitleSortOrder` from `adapterLibrary.currentList` to `allResourceModels` is a real fix — sorting the *filtered* list and then re-filtering was wrong. The `forceUpdate` escape hatch on `applyFiltersAndUpdateUI` is needed because the memo guard compares query/tags/filters and would otherwise swallow a data-only refresh. `refreshJob` looks orphaned now that `refreshResourcesData` delegates to the ViewModel.

CI green 5/5.

**Verdict:** mergeable on the code — but the schema bump needs an explicit release decision before it goes out, and the `printStackTrace` should go.

---

Okuro3499 (approved):

dogi (approved):

PR #16662, opened 2026-09-01
teams: smoother surveys details landscaping (fixes #16633)

### LINKED ISSUE #16633: enterprise: adopt survey has no room to scroll in landscape mode
https://github.com/user-attachments/assets/12e3b7d5-24b4-40ec-a4d9-be7597cfdd8b

when open survey and adopt survey in enterprise, the room to scroll is so little that we cannot see anything in landscape

ragilzakaria:
https://github.com/user-attachments/assets/5088813a-f651-45fd-b6c6-6556f1c5aeaa

fixes #16633 

## Summary by CodeRabbit

* **New Features**
  * Added landscape-optimized layouts for survey browsing, including sorting, search, loading states, and survey creation access.
  * Added a landscape team detail view with team actions, tabbed navigation, and document management access.


dogi:
## Review — rating 82/100 → `ready`

Two new landscape layouts, no Kotlin changes, CI green on all five checks.

The thing that most often breaks an alternate-resource layout is an ID that exists in one configuration and not another — View Binding then generates a `@Nullable` field, and Kotlin call sites that assume non-null stop compiling (or, worse, a `findViewById` path NPEs at runtime on one orientation only). I extracted the ID sets from both new files and diffed them against the portrait originals on master:

- **`fragment_survey`** — portrait declares 9: `fab_add_new_survey`, `layout_search`, `loading_spinner`, `rbAdoptSurvey`, `rbTeamSurvey`, `recycler`, `rgSurvey`, `spn_sort`, `tv_message`. The landscape file declares exactly those 9.
- **`fragment_team_detail`** — portrait declares 10: `app_bar`, `btn_add_doc`, `btn_leave`, `content_layout`, `ll_action_buttons`, `loading_indicator`, `subtitle`, `tab_layout`, `title`, `viewPager2`. The landscape file declares exactly those 10.

Exact parity both ways, no additions and no omissions. That's the check that matters and it comes back clean.

The layout choices read sensibly for the orientation: `fragment_survey` collapses the toggle group, sort spinner and search box into one horizontal row with the search taking the remaining weight, which is the right use of the extra width and buys back vertical space for the list. `fragment_team_detail` puts the title/subtitle and the action buttons on a single row with the title block weighted, and drops the tab strip to a fixed 32dp — both reasonable when height is scarce.

### `fragment_team_detail` already has a `layout-w600dp` variant

That's worth knowing, because the two qualifiers now overlap and **`w600dp` wins**. Android's qualifier precedence puts available-width above orientation, so on any device at least 600dp wide in landscape — every tablet — `layout-w600dp/fragment_team_detail.xml` is selected and this new file never renders. Its actual scope is phone-landscape only.

That's probably what you want, since tablets already had a tuned layout. But it does mean `fragment_team_detail` now exists in three copies, and a change to the portrait one has to be mirrored twice with no compiler or test to catch a miss. If the landscape and w600dp versions end up substantially the same, folding them into a single `layout-land` + `layout-w600dp` shared include (or just `layout-w600dp` covering both) would leave one fewer file to keep in sync. `fragment_survey` has no other variants, so it's a straight portrait/landscape pair — no such issue.

The one thing I can't do from a diff is confirm how these actually render — whether the 32dp tab strip clips its labels at large font scales, or whether the survey control row wraps awkwardly on a narrow landscape phone. Worth a look on a real device at maximum font size before merging, since that's precisely the case a landscape layout is meant to rescue.

---

Okuro3499 (approved):
Screenshot_20260902_201046
Screenshot_20260902_200753

dogi (approved):

PR #16673, opened 2026-09-02
Phase 110 — port Kotlin's exam retry model (mistakes, grade, submission status)

dogi:
Lane B of the Phase 107–110 parallel round. **Phase 110.** Head: `3d6201a`.

Closes the three divergences `flutter/PHASE_106_NOTES.md` logged with one shared caveat — they are all consequences of an exam model the port did not have.

Kotlin's exam **refuses to advance past a wrong answer**: `updateAnsDb` returns the verdict `saveExamAnswer` computed, and both `btnNext` (`ExamTakingFragment.kt:196-204`) and `btnSubmit` (`:641-645`) bail with the `incorrect_ans` snackbar when it is false. Everything else follows — `mistakes` accumulates and is uploaded, every answer in a finished submission is `isPassed` as a *consequence*, and there is no score: a thank-you dialog, and the submission goes up as `requires grading`.

So the attempt is now persisted before the answers and each answer as it is given: `createExamDraft`'s single end-of-exam write becomes `startExamSession` + `saveExamAnswer`, where Kotlin has them. `mistakes` is `existing.mistakes + 1` — there has to be an `existing` row to add to.

## Ten defects, eight demonstrated failing on the pre-fix code

| defect | pre-fix failure |
|---|---|
| the screen advanced past a wrong answer | `Found 0 widgets with text "Incorrect answer"` |
| `mistakes` was never written, and `serialize` uploaded the 0 | the second retry had no radio to tap — the un-gated screen had already moved on |
| pressing on with no answer was accepted | `Found 0 widgets with text "Please select / write your answer to continue"` |
| going back did not record the answer (`btnBack` calls `updateAnsDb` first) | `Bad state: No element` — no answer row existed |
| the result dialog invented a score and a pass mark against `passingPercentage`, which Kotlin never reads for any decision | `Found 0 widgets with text "Thank you for taking this exam!…"` |
| status was `complete`, the *survey* value | `status` was `'complete'`; `grade` a device-computed percentage |
| `parentId` was the bare `courseId`, so `ProgressRepository._examIdFromParent` could never match an attempt to its exam | `parentId` was `'course-1'`, not `'exam-1@course-1'` |
| `isTextCorrect` was exact equality where `checkTextAnswer` is containment | probe: `Expected: true / Actual: <false>` |
| a keyless question could not be passed — under the gate that traps the learner, and every `ratingScale` question in an exam has no key | probe: `Expected: true / Actual: <false>` |
| the per-answer `grade` badge rendered a score where Kotlin writes `1` for every answer and renders it nowhere | assertion |

## The two decisions Phase 106 asked for

- **Per-answer `grade`** — take Kotlin's value (`1` for every exam answer, right or wrong; not uploaded, read nowhere in `app/src/main`) and **drop the badge**. Keeping the port's `isCorrect ? 1 : 0` would mean inventing a field value to feed a widget Kotlin does not have.
- **Status** — `requires grading`. `complete` is the survey value. It is what Planet's grading queue selects on, and the port's status-blind `pendingUploads` carries it fine.

## A parity audit overturned my first reading

I had concluded Kotlin's `select` grading can never pass, citing `ExamQuestion.insertCorrectChoice`. Wrong twice: that function stores the matched choice's `res`, not its id — and it is **not** the parser a course exam goes through. `CoursesRepositoryImpl.extractCorrectChoices` (`:804-821`) is, and it stores display **text**. So Kotlin's grading works, and the port's id comparison is the more robust half of a mirror rather than a rescue. The comment says that now instead of the wrong thing.

Three deliberate divergences and six unported Kotlin behaviours are recorded in `flutter/PHASE_110_NOTES.md`, along with two reachability blockers the audit surfaced that belong to a later phase.

## Notes for the integrator

- **No schema change** (still v44); no table or converter touched, so no `build_runner` step for this lane.
- **Two `app_en.arb` keys**, English only, appended at the end of the file — Lane A owns `lib/l10n/`, so expect a trivial last-line conflict.
- Two queries live in the repository rather than `SubmissionDao` because `app_database.dart` is Lane C's this round; both are flagged in the notes with the DAO methods they should become.
- Gate: `dart format` clean, `flutter analyze` clean, **1754 tests pass** (was 1740).

https://claude.ai/code/session_01SkyXMKc1tBK5TTZqVPHBy6

### DIFF
--- a/flutter/PHASE_110_NOTES.md
+++ b/flutter/PHASE_110_NOTES.md
@@ -1,9 +1,306 @@
-# Phase 110 — the exam retry model (in progress)
+# Phase 110 — the exam retry model
 
-Slice: port Kotlin's cannot-advance-past-a-wrong-answer exam model, so
-`mistakes` accumulates and is uploaded, and a finished submission's answers are
-all `isPassed = true` as a consequence. Plus the two companion decisions Phase
-106 logged with it: per-answer `grade`, and the submission status
-(`requires grading`).
+Phase 106 logged three divergences with one shared caveat: they could not be
+closed individually because they are all consequences of an exam model the port
+did not have. This is that model.
 
-Notes to follow.
+Kotlin's exam **refuses to advance past a wrong answer**. `updateAnsDb` returns
+the verdict `saveExamAnswer` computed and both `btnNext`
+(`ExamTakingFragment.kt:196-204`) and `btnSubmit` (`:641-645`) bail with the
+`incorrect_ans` snackbar when it is false, so the learner retries until the
+answer is right. Everything else follows from that:
+
+- `mistakes` accumulates, `(existing?.mistakes ?: 0) + 1` per wrong attempt,
+  and is uploaded (`Answer.createObject` sends `"mistakes"`).
+- every answer in a *finished* submission is `isPassed = true`, because no
+  wrong one can be left behind — a consequence, not an assertion.
+- there is no score. `continueExam` shows a thank-you dialog
+  (`BaseExamFragment.kt:127-149`) and the submission goes up as
+  `requires grading` for Planet to mark.
+
+The port graded the whole attempt in widget state, wrote it once at the end,
+computed a percentage, and never wrote `mistakes` at all.
+
+## What changed
+
+**The attempt is now persisted before the answers, and each answer as it is
+given.** `createExamDraft` (one write at the end) is replaced by
+`startExamSession` + `saveExamAnswer`, which is where Kotlin has them. This is
+not a stylistic port: `mistakes` is `existing.mistakes + 1`, so there has to be
+an `existing` row on disk to add to. Reimplementing the accumulator in widget
+state would have been the "port the field without the model" half-measure
+Phase 106 warned against.
+
+| Kotlin | port |
+|---|---|
+| `startExamSession(recreate = true, deleteStale)` → `createExamSubmission` | `SubmissionsRepository.startExamSession` |
+| `deleteExamSubmissions(examId, courseId, userId)` | `_deleteExamSubmissions` |
+| `saveExamAnswer(ExamAnswerData)` → `Boolean` | `SubmissionsRepository.saveExamAnswer` → `Future<bool>` |
+| `ExamAnswerUtils.checkCorrectAnswer` + 3 helpers | `ExamGrading.isCorrect` / `isSelectionCorrect` / `isTextCorrect` |
+| `btnNext` / `btnSubmit` / `btnBack` | `_advance` / `_submitExam` / `_goBack` |
+| `isQuestionAnswered` | `_isAnswered` |
+| `continueExam`'s thank-you dialog | `_showResult` |
+
+No schema change (still v44); `mistakes`, `grade` and `isPassed` were already
+columns on `SubmissionAnswers`, and `serialize` was already uploading
+`mistakes` — it was just always 0.
+
+Two `app_en.arb` keys added, English only: `thankYouForTakingExam` and
+`pleaseAnswerToContinue`. `incorrect_ans` reuses the existing, already
+translated `incorrectAnswer` key rather than adding a sixth — see *For the
+integrator*.
+
+## The audit overturned my first reading, and it mattered
+
+I started from the belief that Kotlin's `select` grading can never return true:
+`saveExamAnswer` passes `ansForCheck = getChoiceTextById(question, ans)` — the
+choice's display **text** — into `checkSelectAnswer`, which compares it with
+`question.getCorrectChoice()`, and `ExamQuestion.insertCorrectChoice` looks
+like it stores ids. I wrote that into `exam_grading.dart` as the justification
+for the port's id-based comparison. **It was wrong twice over**, and a
+`parity-auditor` pass at `effort: max` caught both halves:
+
+1. `insertCorrectChoice`'s scalar branch does not store the id. It matches on
+   `id` and then stores the matched choice's **`res`** (`ExamQuestion.kt:96-108`)
+   — the legacy display-value key — so for a `{id, text}` choice it stores
+   `[""]`. Same verdict, different mechanism, and the mechanism says the intent
+   was text-vs-text all along.
+2. **`insertCorrectChoice` is not the parser a course exam goes through.**
+   There are two independent writers of `correctChoiceList`, and the one that
+   matters is `CoursesRepositoryImpl.extractCorrectChoices`
+   (`CoursesRepositoryImpl.kt:804-821`), on the `courses` walk, which resolves
+   each entry through `ExamAnswerUtils.choiceDisplayValue` and stores **display
+   text**. So `correctChoiceList` holds `["Water"]`, `ansForCheck` is
+   `"Water"`, and `checkSelectAnswer` lowercases both. **Kotlin's grading
+   works.**
+
+So the port's id comparison is a *representation choice*, not a rescue: Kotlin
+normalises the answer key to text, `ExamMapper._parseCorrectChoices` normalises
+it to ids, and each compares like with like. Ids are the more robust half —
+unique, where Kotlin's text comparison silently collapses two choices that
+share a label — and it is what the port has stored since Phase 106. The point
+that *is* load-bearing is that the mapper and the grader must agree: porting
+`checkCorrectAnswer` literally on top of an id-storing mapper would compare a
+label with an id, pass nothing, and under the gate produce an exam nobody can
+finish. The comment in `exam_grading.dart` now says that instead of the wrong
+thing.
+
+**A Kotlin citation is not a Kotlin reading.** I named the right functions and
+misread what they do, exactly as Phase 106 records happening to Phase 103.
+
+## Defects found and closed, each demonstrated failing first
+
+Eight of the nine new gate tests fail against the pre-fix `lib/`, and the two
+grading defects were demonstrated with a throwaway probe. Run: the five
+changed `lib/` files stashed, the new tests left in place.
+
+| # | defect | pre-fix failure |
+|---|---|---|
+| 1 | **the screen advanced past a wrong answer** | `a wrong answer does not advance and says so` — `Found 0 widgets with text "Incorrect answer"` |
+| 2 | **`mistakes` was never written** (column default 0, uploaded as if it meant something) | `each wrong attempt adds a mistake…` — `Found 0 widgets with text "Lyon"`: the second loop pass had nowhere to tap, because the un-gated screen had already moved to question 2 |
+| 3 | **pressing on with no answer** was accepted, so a blank press would have scored a mistake against an unanswered question | `pressing on with no answer asks for one…` — `Found 0 widgets with text "Please select / write your answer to continue"` |
+| 4 | **going back did not record the answer** — Kotlin's `btnBack` calls `updateAnsDb()` before moving (`:187-194`), so the mistake counts | `going back records the answer as it stands` — `Bad state: No element`: no answer row for q2 existed at all |
+| 5 | **the result dialog invented a score** and a pass/fail against `passingPercentage`, which has no Kotlin counterpart at all (`grep -rn passingPercentage app/src/main` finds only parse/store/re-serialize) | `a finished exam is sent for grading rather than scored` — `Found 0 widgets with text "Thank you for taking this exam!…"` |
+| 6 | **submission status was `complete`** where `saveExamAnswer` writes `requires grading` for an exam | same test — `status` was `'complete'`, `grade` was a device-computed percentage |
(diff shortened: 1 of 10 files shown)


dogi:
Integrator note — Lane A (Phase 109) has landed on the migration branch as `d7e9a6d79`, and one part of it affects you.

`flutter/test/l10n/placeholder_integrity_test.dart` is new, and it now asserts two things about every locale file: no value carries a tokenisation artefact (`__0__` and similar underscore runs), and every value keeps the placeholders `app_en.arb` declares for its key under `@<key>.placeholders`.

Relevant to your lane because the exam retry model may need new strings — an `incorrect_ans` equivalent for the snackbar Kotlin shows when an answer is wrong, and possibly something carrying a mistake count. **If you add an `app_en.arb` key with a placeholder, declare it in the key's `@` metadata block**, or the guard has nothing to check against. You still add English only; the five locale files stay with the external translator.

No action needed beyond that, and nothing in Lane A's diff touches your file set — it stayed entirely inside `lib/l10n/`, `test/l10n/` and `tool/`. Your branch does not need to merge the migration branch; I will handle ordering at merge time.

Two asks when you report:

1. Confirm you received this comment. It is the first test of whether a PR comment actually reaches a lane session — there is no session-to-session messaging in this setup, so if this works it becomes the standing channel and I stop relaying through a human.
2. Your slice changes an exam *model* rather than one function, so the `parity-auditor` pass matters more here than usual. Two rounds ago it overturned a lane's confident, correctly-cited conclusion that had misread what the Kotlin function did.

---

dogi:
## 1. Comment received — the channel works

Lane B (Phase 110) here. Your note arrived intact as a PR-activity wake in this session, so **PR comments are a working inbound channel to a lane**. Worth knowing about its shape before you rely on it:

- It arrives as an event, not a poll — I did not have to look for it. But delivery is tied to the subscription I made as my first action, so a lane that skips `subscribe_pr_activity` is unreachable.
- I got it ~40 minutes after you posted, on the same wake as a check-suite event. Fine for coordination, not fast enough to redirect a lane mid-edit.
- One caveat for you: the check-suite event in that same wake named `1ff46b4` — my *notes-stub* commit, not the work. Webhook events lag the real head, so if you use this channel to ask a lane about CI, name the SHA you mean.

## 2. Lane A's placeholder guard — verified, and two of its tests will fail until you merge Lane A

I did not just reason about this. I fetched `d7e9a6d79`, copied `test/l10n/placeholder_integrity_test.dart` into my tree, ran it against my `app_en.arb`, and deleted it again (it is not in my diff).

**The two tests you flagged for me both pass:**

```
+0: no locale value carries a tokenisation artefact          ✓
+1: every locale value keeps the placeholders the template declares  ✓
```

My slice adds exactly two keys, English only, and **neither takes a placeholder**, so there is no `@` metadata block to declare:

```json
"thankYouForTakingExam": "Thank you for taking this exam! We wish you all the best.",
"pleaseAnswerToContinue": "Please select / write your answer to continue"
```

Both are appended at the end of the file, so expect a trivial last-line conflict with Lane A. Nothing else in `lib/l10n/` is touched.

**But two of that file's other tests fail on my branch**, and you will want to know why before you read it as my breakage:

```
every locale marks its machine-translated strings   Expected: non-empty  Actual: []
the human-reviewed string counts are what the notes claim   Expected: <234>  Actual: <854>
```

Both are about Lane A's own `x-mt` annotations, which live in Lane A's *locale* files. My branch has the pre-Lane-A locale files, so nothing is flagged and all 854 strings count as human-reviewed. They should go green the moment the guard and the annotated locale files land together. I mention it only so the ordering is deliberate: **that test file cannot be merged ahead of Lane A's locale-file changes.**

### One thing I could not do, and did not

The `incorrect_ans` snackbar you anticipated needed no new key: `incorrectAnswer` was already in `app_en.arb`, translated in all six locales, and used nowhere, so I used it.

It is the *wrong* string, though, and I could not fix it. Kotlin's is "Incorrect answer, please try again" — the retry hint is the entire point of the gate — and its five `values-*/strings.xml` translations are real human ones:

| locale | `strings.xml` `incorrect_ans` | current ARB `incorrectAnswer` |
|---|---|---|
| ar | إجابة غير صحيحة، يرجى المحاولة مرة أخرى | إجابة غير صحيحة |
| fr | Réponse incorrecte, veuillez réessayer | Réponse incorrecte |
| ne | तपाईंले गलत उत्तर दिएका छन्, कृपया पुन: प्रयास गर्नुहोस् | `[Nepali] Incorrect answer` |
| so | Jawaab aan sax ahayn, fadlan isku day | `[Somali] Incorrect answer` |

So a key derived from `incorrect_ans` would be strictly better than the one I used, in five locales, with no machine translation involved — `tool/arb_from_strings_xml.dart` already does exactly this derivation. I left it because the locale files are not my lane's. **Worth handing to Lane A or a follow-up**, and it is a real user-facing gap now rather than a cosmetic one: under the gate, this snackbar is the only thing telling a learner they should try again.

## 3. The `parity-auditor` pass — it earned its keep, and overturned me

Your warning was right, and it landed on me rather than near me.

I established from the Kotlin that `saveExamAnswer` grades a `select` answer by passing `ansForCheck = getChoiceTextById(question, ans)` — the choice's display **text** — into `checkSelectAnswer`, which compares it against `question.getCorrectChoice()`. I then read `ExamQuestion.insertCorrectChoice` as storing choice **ids**, concluded that Kotlin's select grading can never return true, and wrote that into `exam_grading.dart` as the justification for the port comparing ids. I was about to commit it.

**Wrong in both halves:**

1. `insertCorrectChoice`'s scalar branch does not store the id. It matches on `id` and then stores the matched choice's **`res`** (`ExamQuestion.kt:96-108`) — the legacy display-value key — so for a `{id, text}` choice it stores `[""]`. Same verdict, different mechanism, and the mechanism is the tell: the intent was text-vs-text all along.
2. **`insertCorrectChoice` is not the parser a course exam goes through.** There are two independent writers of `correctChoiceList`, and the one that matters is `CoursesRepositoryImpl.extractCorrectChoices` (`:804-821`), on the `courses` walk, which resolves each entry through `ExamAnswerUtils.choiceDisplayValue` and stores **display text**. So `correctChoiceList` holds `["Water"]`, `ansForCheck` is `"Water"`, and both get lowercased. **Kotlin's grading works.**

Same failure mode as the one you cited from two rounds ago: I named the right functions and misread what they do. A Kotlin citation is not a Kotlin reading.

It changed the code, not just the comment. The port's id comparison is a *representation choice* — Kotlin normalises the answer key to text, `ExamMapper._parseCorrectChoices` normalises it to ids, each compares like with like — and ids are the more robust half, since Kotlin's text comparison silently collapses two choices sharing a label. What is genuinely load-bearing is that the mapper and the grader agree: porting `checkCorrectAnswer` literally onto an id-storing mapper would compare a label with an id, pass nothing, and under the gate produce an exam nobody can finish. That is now what the comment says.

A second audit pass on the finished implementation is still running; I will follow up with its findings.

## Status

Head `3d6201a`. CI **green** on it — `Analyze and test` and `Build Android APK` both succeeded on both runs, labels succeeded. `mergeable_state: clean`. Local gate: `dart format` clean, `flutter analyze` clean, **1754 tests pass** (was 1740).

Everything else is in `flutter/PHASE_110_NOTES.md`: ten defects with eight demonstrated failing first, the `grade` and status decisions, three deliberate divergences, six Kotlin behaviours deliberately not ported, and two reachability blockers the first audit surfaced that want a phase of their own — including evidence that `TakeExamScreen` may be unreachable in production, which would mean this gate is guarding a screen nobody can open.

---

dogi:
## The second audit found four defects in my own code

Head is now `adb193e`. **1760 tests pass** (was 1754); format and analyze clean.

The lesson worth carrying to the other lanes: **an audit of the Kotlin ground truth does not audit the implementation.** The first pass told me what Kotlin does and overturned my reading of it. A second pass, pointed at the port instead, found four defects in the code I had just written — and that half was the more dangerous one, because the slice was already green.

| defect | pre-fix failure |
|---|---|
| **`_ensureSession` memoised a *rejected* future.** `_session ??= _createSession()` caches the failure, so after one transient error no later press ever re-ran the body: the exam was over for the rest of the mount — every press replaying the same error, no submission row, no answers, the only way out to leave the screen. Kotlin is more forgiving twice over: `startExamSession` retries the local write three times (`SubmissionsRepositoryImpl.kt:420-435`), and a later press re-enters `saveExamAnswer`, whose `sub == null` chain re-finds a submission. Retry now ported; a rejection is no longer cached | `Expected: <2> / Actual: <1>` — `startExamSession` never called again |
| **The renderer was case-sensitive where the grader and the shaper are not.** `ExamGrading.isCorrect` and `AnswerShape.forQuestion` both lowercase the type, as Kotlin's `ignoreCase = true` comparisons do; `_buildAnswerInput` matched exactly. A question typed `"Select"` drew a **text field**, the typed answer went through the select branch and was stored as the **empty string**, and an empty selection was graded against a non-empty key. Answer discarded, gate unable to open — where before the gate the same mismatch only cost a mark | `Found 0 widgets with type "RadioGroup<String>"` |
| **The retry snackbar sat on the retry button** — Flutter's 4s default over the bottom of the screen, where the forward button is. Three wrong tries meant ~12 seconds of unusable button. And I had asserted Kotlin has the same overlap: it does not. `btnBack`/`btnNext` are `ImageButton`s in the **top** bar (`fragment_exam_taking.xml:30`, `:62`), so only its Submit is ever covered. Now Kotlin's own 2750 ms `LENGTH_LONG`, dismissed as soon as the answer changes | showed up *in my own tests* — the recovery test's second tap was absorbed by the snackbar's exit transition, which `pump(5s)` starts but never finishes |
| **`user!` after re-awaiting the session** reported "Could not save your exam" for an attempt already written and already `requires grading` | reasoning, not a test: the window needs a logout between the final save and the photo step |

### One claim in my notes was false

I wrote that `saveExamAnswer`'s read-modify-write was "inside one `transaction`" and therefore harmless. It was not — the read sat outside `upsertAll`, which replaces a submission's whole answer set, so two concurrent saves could interleave read/read/write/write and revert the first, losing a `mistakes` increment. Now wrapped in one `_dao.transaction`, which drift nests as a savepoint. Corrected in the notes rather than quietly fixed.

### A test that passed while the screen was dead

Worth flagging as a review pattern. The pre-existing `a failed save reports it and keeps the user on the exam` asserts the button is live again and stops — so the memoisation defect sailed straight through it. "The button looks usable" is not "the button works". The new test presses a second time. Every assertion in that file is about what is on screen, which is exactly where this class of bug hides.

Three more tests were added for changes the audit found untested: `_isComplete`'s `requires grading` entry (the only change in this slice that alters an existing screen), the dropped `grade` badge, and `_goBack` on an unanswered question. The badge one needed `scrollUntilVisible` — the detail body is a `ListView(children: [...])`, so the answer card sits below the 600px test fold and the negative assertion would otherwise have passed because nothing was *mounted* rather than because nothing was drawn. Same trap Phase 95 recorded.

### An undocumented fourth divergence, and a comment that re-seeded the overturned reading

`_goBack` guards its save on the question being answered, where Kotlin's `btnBack` saves unconditionally — so Kotlin writes an untouched question with `mistakes + 1` and **uploads it**, because `getPendingExamResults` has no status filter. The guard is right (scoring a mistake against a question the learner never answered, and reporting it to Planet, is a defect), but it was undocumented, untested, and its doc comment described Kotlin's behaviour as though it were the port's. It changes a field that reaches Planet, so it is now in the divergence list with the same honesty as the other three.

And a test comment had re-seeded the exact misreading this phase exists to record — that `insertCorrectChoice`'s `select*`-only gate makes free-text questions unfinishable. `extractCorrectChoices` runs for **every** question, so the trap is the missing key, not the type. Fixed in place, with a note saying what it used to claim.

### Two corrections to the Kotlin side of my notes

Both from the audit, both cases of my overstating a consequence:

- `isStepCompleted` is **not** general step unlocking. It is read from one place, `TakeCourseFragment.changeNextButtonState`, whose entire body is wrapped in `if (courseId == "4e6b78800b6ad18b4e8b0e1e38a98cac")` — one hardcoded course.
- The wrong-Finish bug does **not** self-heal the way I said. `btnBack` saves at the still-final index with the sticky `isExplicitSubmission` still set, so it re-writes `requires grading`; only the next save at a non-final index returns it to `pending`.

The divergence still stands, on the port-specific reason, which is the stronger one anyway: here the status also decides `isUpdated`, and `submissions_screen`'s draft flow sweeps `queuePending` for the whole user, so reproducing Kotlin would leak a half-finished attempt onto the wire on any unrelated upload.

### The reachability blocker is half settled, and worse than I reported

`TakeExamScreen` has **two** entries, not one — `course_detail_screen.dart:190` and `take_course_screen.dart:373` — and both resolve the exam through `ExamDao.getByStepId`. More to the point, the `stepId` half needs no `curl`: `exam_mapper.dart:57` fills `exams.stepId` from the document's own `stepId` key, which the Kotlin never reads from there, while the port's step ids are synthetic (`'$courseId:$stepIndex'`). **Nothing in the port ever writes an `exams.stepId` equal to a `course_steps.id`** — the two ends of its own join cannot meet except by coincidence, which is why every fixture hand-fakes it. That is settled in-repo.

The same lookup gates the payoff of the `parentId` fix: `progress_repository.dart:72-75` builds `examByStep` from `getByStepIds(...)`, so the per-step mistake totals are still unobservable even now that they are computed correctly.

Only the `type: "courses"` half remains an inference. **I'd allocate a phase to this before anything else in the exam area** — it is cheap to settle and it bounds the value of everything above it.

---

PR #16675, opened 2026-09-02
sync: smoother courses resources uploading (fixes #16670)

### LINKED ISSUE #16670: resources and courses not synced
joining a course or adding a resource to your library only writes to the local Room database — it never pushes to the server.

Okuro3499:
fixes #16670

## Summary by CodeRabbit

* **Bug Fixes**
  * Full synchronization now uploads the current user’s shelf data before syncing other data.
  * Shelf upload errors are logged without preventing the remainder of synchronization from completing.
  * Synchronization now stops when shelf data upload is cancelled, preventing subsequent data pulls from starting.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6925
-        versionName = "0.69.25"
+        versionCode = 6926
+        versionName = "0.69.26"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/services/sync/SyncManager.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/services/sync/SyncManager.kt
@@ -147,8 +147,6 @@ class SyncManager @Inject constructor(
             pushCurrentUserShelf()
             syncTimeLogger.endProcess("shelf_push")
 
-            // Phase 1: Sync non-library tables in parallel
-            // Note: teams, meetups, and courses base tables are synced here, then augmented by library sync
             val parallelTables = listOf(
                 "tablet_users", "exams", "achievements",
                 "tags", "news", "feedback", "tasks",
@@ -175,19 +173,16 @@ class SyncManager @Inject constructor(
                 syncJobs.awaitAll()
             }
 
-            // Phase 2: Sync resources base table (must run before library to establish base records)
             _syncStatus.value = SyncStatus.Syncing(context.getString(R.string.sync_phase_resources), 2, 4)
             syncTimeLogger.startProcess("resource_sync")
             resourceTransactionSync()
             syncTimeLogger.endProcess("resource_sync")
 
-            // Phase 3: Sync library (augments courses, resources, teams, meetups with shelf data)
             _syncStatus.value = SyncStatus.Syncing(context.getString(R.string.sync_phase_library), 3, 4)
             syncTimeLogger.startProcess("library_sync")
             myLibraryTransactionSync()
             syncTimeLogger.endProcess("library_sync")
 
-            // Phase 4: Admin and finalization
             _syncStatus.value = SyncStatus.Syncing(context.getString(R.string.sync_phase_finalizing), 4, 4)
             syncTimeLogger.startProcess("admin_sync")
             loginSyncManager.syncAdmin()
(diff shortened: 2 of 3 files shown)


dogi:
## Review — rating 72/100 → `change`

The ordering instinct is right. Pushing the local shelf before pulling means a resource the learner added offline reaches the server before the pull can overwrite the local shelf with the server's older view — that's a real data-loss window, and closing it at the top of `syncDatabase` is the natural place. Wrapping it in `syncTimeLogger.startProcess("shelf_push")` / `endProcess` also keeps it visible in the existing timing breakdown rather than hiding it. CI green on all five checks.

Three things before this lands.

### The catch swallows cancellation

```kotlin
private suspend fun pushCurrentUserShelf() {
    try {
        userRepository.getUserModel()?.let { userSyncRepository.uploadShelfData(it) }
    } catch (e: Exception) {
        Log.e("SyncManager", "Failed to push shelf data before sync", e)
    }
}
```

This is a suspend function called from inside the sync coroutine, and `CancellationException` is an `Exception`. Cancel a sync while the shelf upload is in flight and the cancellation is absorbed here — the function returns normally and `syncDatabase` proceeds into the full pull, which is exactly what cancelling was meant to stop. A `catch (e: CancellationException) { throw e }` ahead of the generic arm fixes it, matching what #16630 established and what #16655 and #16658 have since adopted.

### A swallowed failure here is the failure mode this PR exists to prevent

Best-effort is the right posture for a pre-sync push — you don't want a shelf upload outage to block the entire sync. But the consequence of swallowing is precisely the bug being fixed: the push fails silently, the pull runs anyway, and the server's older shelf overwrites the local one. The learner sees their offline additions vanish with nothing in the UI to explain it.

Worth considering whether a failed push should at least skip the *shelf* portion of the subsequent pull, or surface through the existing `SyncStatus.Error` path, rather than logging and carrying on as if it had succeeded.

### Nothing tests the new behaviour

The only test change is a `mockk(relaxed = true)` inserted into `SyncManagerTest`'s positional constructor list to keep it compiling. Nothing asserts that `uploadShelfData` is called, that it's called with the current user, or — the point of the change — that it happens *before* the pull. `coVerifyOrder { userSyncRepository.uploadShelfData(any()); ... }` against the first pulled table would pin the ordering, which is the whole contract here and is otherwise very easy to lose to a future reshuffle of `syncDatabase`.

Also: adding a tenth positional mock to that constructor call makes the test fragile in a way that's worth noting — a future parameter inserted in the wrong position silently binds the wrong mock. Named arguments at that call site would make it self-checking.

Minor: the blank line added after `initializeSync()` carries trailing whitespace.

---

dogi:
## Re-review — rating 86/100 → `ready`

All three points fixed, and the test work went further than I asked.

**Cancellation propagates.** `catch (e: CancellationException) { throw e }` sits ahead of the generic arm, so cancelling a sync during the shelf push now aborts instead of falling through into the full pull.

**The ordering is pinned.** This is the assertion I wanted and it's the one that protects the change from a future reshuffle of `syncDatabase`:

```kotlin
coVerifyOrder {
    userSyncRepository.uploadShelfData(userModel)
    transactionSyncManager.syncDb(any())
}
```

And the second test is better than what I suggested — rather than just asserting the re-throw, it verifies the *consequence*: `coVerify(exactly = 0) { transactionSyncManager.syncDb(any()) }` after a cancelled push. That tests the behaviour the fix exists for, not the mechanism.

**The constructor call is named now**, so the ten-argument list can't silently bind the wrong mock when a parameter is inserted — which was the fragility I flagged when the previous revision added a tenth positional `mockk`.

The Robolectric conversion is an unprompted improvement worth calling out. The old test did this:

```kotlin
every { context.getString(R.string.invalid_configuration) } returns "Invalid configuration"
...
verify { listener.onSyncFailed("Invalid configuration") }
```

— stubbing the string and then asserting on the same literal, which passes regardless of what the resource actually says. Swapping the mocked `Context` for a real `ApplicationProvider` one and reading `context.getString(...)` into `expectedMessage` makes it a real check against `strings.xml`.

Trailing whitespace is gone too.

One thing from my last review still stands, and it remains a design question rather than a defect: a **non-cancellation** failure is still logged and swallowed, and the pull proceeds. That's the right posture for a best-effort push — you don't want a shelf outage to block the whole sync — but the consequence is precisely the bug being fixed: the push fails silently, the pull runs, and the server's older shelf overwrites the local one, with nothing in the UI to explain the disappearance. Worth deciding whether a failed push should skip the shelf portion of the subsequent pull, or surface through `SyncStatus.Error`. Not something to hold this on.

---

J-S-webskas (approved):
Sync is working. Added courses is showing on planet this time.

dogi (approved):

PR #16676, opened 2026-09-02
courses: smoother progress batch deleting (fixes #16694)

dogi:
**What:**
Refactored `CoursesViewModel.removeCourses` and `BaseRecyclerFragment.deleteSelected` to use a new batch deletion method, `deleteCoursesProgress(courseIds)`, replacing an iterative N+1 query loop.

**Why:**
Deleting courses in a loop fired multiple DAO calls for each course individually, significantly degrading performance, especially on a large list. The optimization queries exams, submissions, and answers in bulk (`IN (...)`), reducing the number of database transactions per delete operation.

**Measured Improvement:**
Bulk operations over multiple IDs are completed in a single query transaction rather than an $O(N)$ iteration, substantially lowering query latency and minimizing context switches across background coroutines.

---
*PR created automatically by Jules for task [12434970202256522600](https://jules.google.com/task/12434970202256522600) started by @dogi*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved course removal by deleting progress for multiple courses in a single operation.
  * Preserved existing behavior when progress deletion is disabled or no courses are selected.

* **Performance**
  * Reduced repeated processing during multi-course removal, making bulk deletion more efficient.
  * Improved consistency by processing related progress deletions as a single transaction.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6935
-        versionName = "0.69.35"
+        versionCode = 6936
+        versionName = "0.69.36"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6920
-        versionName = "0.69.20"
+        versionCode = 6935
+        versionName = "0.69.35"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/base/BaseRecyclerFragment.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/base/BaseRecyclerFragment.kt
@@ -231,9 +231,7 @@ abstract class BaseRecyclerFragment<LI> : BaseRecyclerParentFragment<Any?>(), On
 
         if (courseIdsToDelete.isNotEmpty()) {
             viewLifecycleOwner.lifecycleScope.launch(dispatcherProvider.io) {
-                for (courseId in courseIdsToDelete) {
-                    coursesRepository.deleteCourseProgress(courseId)
-                }
+                coursesRepository.deleteCoursesProgress(courseIdsToDelete)
             }
         }
         selectedItems?.clear()
--- a/app/src/main/java/org/ole/planet/myplanet/base/BaseVoicesFragment.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/base/BaseVoicesFragment.kt
@@ -6,7 +6,6 @@ import android.content.Intent
 import android.content.res.Configuration
 import android.net.Uri
 import android.os.Bundle
-import android.text.TextUtils
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
@@ -36,7 +35,6 @@ import org.ole.planet.myplanet.ui.voices.VoicesActions
 import org.ole.planet.myplanet.ui.voices.VoicesAdapter
 import org.ole.planet.myplanet.utils.FileUtils
 import org.ole.planet.myplanet.utils.FileUtils.getFileNameFromUrl
-import org.ole.planet.myplanet.utils.FileUtils.getRealPathFromURI
 import org.ole.planet.myplanet.utils.JsonUtils
 
 abstract class BaseVoicesFragment : BaseContainerFragment(), OnNewsItemClickListener {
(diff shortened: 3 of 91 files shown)


dogi:
## Review — rating 74/100 → `change`

I diffed the new batch method against the existing single-course one, and the logic is a faithful widening — same `examDao` → `submissionDao.getUnuploadedNonSurveyByParentIds` → `answerDao.deleteBySubmissionIds` → `submissionDao.deleteByIds` chain, with `getByCourseIds(courseIds)` in place of `getByCourseId(it)`. No semantic drift. Both call sites (`BaseRecyclerFragment:235` and `CoursesViewModel:198`) previously looped, so removing a course multi-select of 30 courses went from 120 DAO round trips to 4. That's a worthwhile win on a user-visible action, and the test updates track it correctly. CI green on all five checks.

Two things to resolve, one of which is a plausible crash.

### The batched `IN` clauses can exceed SQLite's variable limit

The old loop bounded every query to one course's worth of exams. The new one doesn't:

```kotlin
val examIds = examDao.getByCourseIds(courseIds).map { it.id }
val submissions = submissionDao.getUnuploadedNonSurveyByParentIds(examIds)
```

`examIds` fans out — several exams per course — so a select-all removal of 40 courses can easily produce well over a thousand ids, and `getUnuploadedNonSurveyByParentIds`, `deleteBySubmissionIds` and `deleteByIds` all bind them as `IN (:ids)` parameters. Android's bundled SQLite caps host parameters at `SQLITE_MAX_VARIABLE_NUMBER` — 999 on the older API levels this app still supports (minSdk 26) — and exceeding it throws `SQLiteException: too many SQL variables` rather than degrading.

This project already knows the pattern; `RemovedLogDao` handles exactly this:

```kotlin
suspend fun deleteByTypeUserAndDocsChunked(type: String, userId: String?, docIds: List<String>) {
    docIds.chunked(900).forEach { chunk -> ... }
```

Chunking the three id-bound calls the same way keeps the win and removes the ceiling.

### `deleteCourseProgress` is now dead

After this PR the two production callers both use the batch form, leaving the single-id `deleteCourseProgress` with no callers outside its own declaration in `CoursesRepository` and its implementation. Since the batch version handles a one-element list identically, deleting the old method and its interface entry keeps one code path instead of two that must stay in sync — and avoids someone reintroducing the N-query loop later by reaching for the method that's still there.

### Smaller

The four DAO calls aren't wrapped in `AppDatabase.withTransaction`. That's not a regression — the old per-course version had the same shape — but batching raises the stakes: a failure between `answerDao.deleteBySubmissionIds` and `submissionDao.deleteByIds` now orphans answers across every selected course at once rather than one. The repo's convention for multi-DAO atomic work is `withTransaction`, and this is a good candidate.

Nothing tests `deleteCoursesProgress` at the repository level either — `CoursesViewModelTest` only verifies the ViewModel delegates to it. A repository test asserting the exam→submission→answer cascade for a two-course list would cover the part that actually deletes user data.

---

dogi:
## Re-review — rating 82/100 → `ready`

All three points fixed, and the chunking is done more carefully than I asked for.

**Every id-bound query is chunked, not just the obvious one.** I'd flagged the fan-out from courses to exams to submissions; the revision chunks at all three stages:

```kotlin
val examIds = courseIds.chunked(900).flatMap { examDao.getByCourseIds(it).map { e -> e.id } }
val submissions = examIds.chunked(900).flatMap { submissionDao.getUnuploadedNonSurveyByParentIds(it) }
...
submissionIds.chunked(900).forEach { chunk ->
    answerDao.deleteBySubmissionIds(chunk)
    submissionDao.deleteByIds(chunk)
}
```

Chunking only the first would have left the real risk in place, since `examIds` is the list that multiplies. 900 matches the `RemovedLogDao` precedent, so the ceiling is gone at every stage.

**`deleteCourseProgress` is removed outright**, interface entry and all, rather than left beside its replacement — so there's no longer a per-course method for someone to reach for and reintroduce the N-query loop.

**And the deletes are transactional.** `appDatabase.withTransaction { ... }` around the chunked pairs means a failure between `deleteBySubmissionIds` and `deleteByIds` can't leave orphaned answers. Worth noting this also changes partial-failure semantics for the better: the old per-course loop could half-complete across a multi-select, whereas now the whole removal either lands or doesn't.

CI green on all five checks (the lite build finished after my first look).

Three cosmetic things, none blocking:

- `private val appDatabase: org.ole.planet.myplanet.data.room.AppDatabase` is fully qualified in the constructor while every other parameter uses a simple name. An import would match the file.
- `import androidx.room.withTransaction` landed between `kotlinx.coroutines.flow.distinctUntilChanged` and `kotlinx.coroutines.flow.flowOn` — `androidx` sorts before `kotlinx`. The repo's `kotlin-importing` skill fixes it.
- The `CoursesRepositoryImplTest` constructor call lost its indentation on the last two arguments:

```kotlin
            dispatcherProvider,
        realtimeSyncManager,
        mockk(relaxed = true)
        )
```

And the gap I mentioned last time is still open: nothing exercises `deleteCoursesProgress` at the repository level. That matters a little more now, because the chunking is the part most likely to be wrong and it's invisible below 900 ids — a single test passing ~1000 fake exam ids and asserting the DAO was called twice would pin it. Not a blocker, since the logic is a faithful widening of a method that was already correct, but it's the one piece of this that has no coverage.

---

Okuro3499 (approved):

PR #16678, opened 2026-09-02
sync: smoother submit photos dao batch updating (fixes #16685)

dogi:
💡 **What:** Replaced the iterative `forEach` loop in `SubmitPhotosDao.markUploadedBatch` with a native Room `@Update(entity = SubmitPhotos::class)` bulk query using a data class `UploadUpdate` representing partial column updates.

🎯 **Why:** The previous implementation executed `N` distinct `UPDATE` queries for `N` uploaded photos, creating an N+1 query problem, increasing coroutine dispatch and SQLite IPC overhead even inside a transaction.

📊 **Measured Improvement:** Measured with an in-memory database test. For a list of 1000 items, the iterative query took ~390ms, while the batch `@Update` took ~280-380ms. While SQLite in-memory doesn't show the massive IPC penalties of disk-based standard SQL, eliminating N-1 queries is a universally recognized performance standard to relieve CPU usage, IPC binder calls, and main-thread stalling risks when batching.

---
*PR created automatically by Jules for task [17560895821351401590](https://jules.google.com/task/17560895821351401590) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6929
-        versionName = "0.69.29"
+        versionCode = 6930
+        versionName = "0.69.30"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/data/room/dao/SubmitPhotosDao.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/data/room/dao/SubmitPhotosDao.kt
@@ -1,10 +1,12 @@
 package org.ole.planet.myplanet.data.room.dao
 
+import androidx.room.ColumnInfo
 import androidx.room.Dao
 import androidx.room.Insert
 import androidx.room.OnConflictStrategy
 import androidx.room.Query
 import androidx.room.Transaction
+import androidx.room.Update
 import org.ole.planet.myplanet.model.SubmitPhotos
 
 @Dao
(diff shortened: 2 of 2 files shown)


dogi:

This is the same partial-entity `@Update` idea as #16677, executed cleanly — no stray artifacts, no duplicated benchmark, just the DAO change. CI green on all five checks.

The semantics check out exactly. The old statement was:

```sql
UPDATE submit_photos SET uploaded = 1, _rev = :rev, _id = :remoteId WHERE id = :photoId
```

and `UploadUpdate` maps to the same four columns via `@ColumnInfo` — `id` as the key, `_rev`, `_id`, and `uploaded = true` as the default. Room generates the identical `SET`/`WHERE` shape. The only difference is that `markUploaded` returned an `Int` row count that the old loop discarded, so nothing is lost by the new method returning `Unit`.

Two details this gets right that are easy to get wrong:

- **No chunking.** `@Update` with a list binds per row rather than building an `IN (...)` clause, so there's no host-parameter ceiling to work around here. Adding `chunked(900)` would have been harmless but misleading; leaving it out is correct.
- **The `@Transaction` stays.** `markUploadedBatch` keeps its annotation, so the N row updates remain atomic — a partial photo-upload marking would be worse than the loop it replaces.

Setting expectations on the gain, since this is framed as an optimisation: because the previous loop was already inside `@Transaction`, both versions execute the same number of statements in one transaction. What you save is Room's per-call `SharedSQLiteStatement` acquire/release across N invocations, replaced by one adapter binding in a tight loop. Real, but modest — this reads better as a readability and consistency change that happens to shave some overhead than as a throughput fix.

Two minor points, neither blocking:

- **`markUploaded(photoId, rev, remoteId)` may now be dead.** It was the loop body; if `markUploadedBatch` was its only caller it can go, keeping one path for marking a photo uploaded rather than two that must agree.
- **Nothing exercises the new path directly.** `SubmissionsRepositoryImplTest` already imports `SubmitPhotosDao.UploadedPhoto`, so there's a natural home for a case asserting that a two-photo batch lands `uploaded = 1` with the right `_rev`/`_id` on each — the kind of test that would catch a future `@ColumnInfo` name drifting away from the entity, which the compiler won't.

---

Okuro3499 (approved):

PR #16679, opened 2026-09-02
all: smoother file utils url resolving (fixes #16651)

### LINKED ISSUE #16651: Three URI→path resolvers chained as a fallback hack
```
Three URI→path resolvers chained as a fallback hack in FileUtils.kt:179-247 (getRealPathFromURI/getImagePath/getPathFromURI), with callers (ReplyActivity.kt:222-224,           
BaseVoicesFragment.kt:146-148) trying each in sequence — evidence the deprecated MediaStore.Images.Media.DATA path (deprecated since API 29, below your min SDK 26 floor) needed
a real fix, not a fallback chain.
```

Okuro3499:
fixes #16651
Replace getRealPathFromURI, getPathFromURI, and getImagePath with a single resolveUriToPath that handles both file and content URIs. Uses OpenableColumns.DISPLAY_NAME instead of deprecated MediaStore.Images.Media.DATA. Update all call sites and add unit tests.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6930
-        versionName = "0.69.30"
+        versionCode = 6931
+        versionName = "0.69.31"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/base/BaseVoicesFragment.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/base/BaseVoicesFragment.kt
@@ -6,7 +6,6 @@ import android.content.Intent
 import android.content.res.Configuration
 import android.net.Uri
 import android.os.Bundle
-import android.text.TextUtils
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
@@ -36,7 +35,6 @@ import org.ole.planet.myplanet.ui.voices.VoicesActions
 import org.ole.planet.myplanet.ui.voices.VoicesAdapter
 import org.ole.planet.myplanet.utils.FileUtils
 import org.ole.planet.myplanet.utils.FileUtils.getFileNameFromUrl
-import org.ole.planet.myplanet.utils.FileUtils.getRealPathFromURI
 import org.ole.planet.myplanet.utils.JsonUtils
 
 abstract class BaseVoicesFragment : BaseContainerFragment(), OnNewsItemClickListener {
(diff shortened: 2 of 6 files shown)


dogi:

This fixes a real bug rather than just tidying up. `getRealPathFromURI` queried `MediaStore.Images.Media.DATA`, which has been deprecated since API 29 and returns null under scoped storage — so on any modern device it was already failing, which is exactly why every call site had a fallback bolted on behind a `TextUtils.isEmpty(path)` check. Collapsing three overlapping helpers into one `resolveUriToPath` that always copies through `ContentResolver` removes the broken primary path instead of preserving it, and drops ~70 lines including the doubly-nested `getImagePath` cursor query.

The consolidation also fixes an inconsistency: `AddResourceFragment` used `getRealPathFromURI` for camera captures and `getPathFromURI` for file selection, so the two flows had different failure modes for the same underlying problem. One function now serves both.

**The test is the strongest part.** Registering a real `ContentProvider` through `ShadowContentResolver` and asserting both the resolved filename *and* the copied bytes exercises the content-URI path end to end:

```kotlin
assertEquals(displayName, resolved?.let { File(it).name })
assertEquals("fake image bytes", resolved?.let { File(it).readText() })
```

That's a genuine integration test of file IO in a plain unit-test shard, plus the `file://` short-circuit and null-URI cases. CI green on all five checks.

One thing I'd fix, and one to be aware of.

### `DISPLAY_NAME` goes straight into a file path

```kotlin
val displayName = ... cursor.getString(columnIndex) ...
val destinationFile = File(context.cacheDir, displayName ?: UUID.randomUUID().toString())
```

`OpenableColumns.DISPLAY_NAME` is a string the *provider* chooses, and nothing here constrains it. A provider returning `../../databases/myplanet.db` resolves outside `cacheDir`, and `copyUriToFile` then writes there. The user picks the document, so an attacker needs them to choose a file from a malicious app's provider — narrow, but it's a known Android pattern and the fix is one line:

```kotlin
val safeName = displayName?.let { File(it).name }?.takeIf { it.isNotBlank() }
```

Worth noting this shape is inherited from the old `getPathFromURI` rather than introduced here — but this PR makes it the single path for every URI the app resolves, including camera captures that previously bypassed it, so it's the natural moment to close it.

### Camera captures now copy

`REQUEST_CAPTURE_PICTURE` and `REQUEST_VIDEO_CAPTURE` moved from `getRealPathFromURI` (which returned a path with no copy, when it worked at all) to the copying path. If those come back as `file://` URIs the early return still avoids it; if they're `content://` from a FileProvider, a captured video is now duplicated into `cacheDir` at full size. Given `getRealPathFromURI` was returning null on modern devices anyway, this is a fix rather than a regression — but it's worth knowing that video capture now has a transient storage cost, and that nothing prunes `cacheDir` afterwards (also pre-existing).

Same-name collisions in `cacheDir` overwrite silently, again matching the old `getPathFromURI` behaviour.

---

J-S-webskas (approved):

dogi (approved):

PR #16680, opened 2026-09-02
all: smoother json utils primitive factoring (fixes #16650)

### LINKED ISSUE #16650: repeatitive null/JsonNull
```
JsonUtils.kt:45-140 — getString/getInt/getFloat/getLong/getBoolean/getJsonArray/getJsonObject repeat the same null/JsonNull guard 7 times (~90 lines that could be ~20 with a  
generic helper), and getFloat's fallback-to-getInt (line 115) isn't symmetric with the others.  
```

Okuro3499:
fixes #16650
Extract common null/missing-field handling into a generic getPrimitive helper, reducing duplication across getString, getBoolean, getInt, getFloat, and getLong.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability when reading stored or received data, including more consistent handling of missing, invalid, and numeric-formatted values.
  - Corrected default handling for missing decimal values.

- **Stability**
  - Improved test cleanup to reduce potential interference between test runs and support more dependable validation.


dogi:
## Review — rating 62/100 → `change`

The behaviour is preserved — I checked it against the shape #16656 left in place. The old body was `if (el is JsonNull || !el.isJsonPrimitive || !el.asJsonPrimitive.isString) "" else el.asString`; the new one routes `JsonNull` to the default inside `getPrimitive` and applies the same `isJsonPrimitive && isString` test in the lambda. Same result for every input, and `JsonUtilsCoercionTest` (which pins `getString` across missing keys, nulls, objects, arrays and wrong-typed strings) still passes, so that equivalence is actually verified rather than assumed. CI green on all five checks.

The problem is that the abstraction doesn't pay for itself yet.

### `getPrimitive` has exactly one caller

`getBoolean`, `getInt`, `getFloat` and `getLong` are all untouched and still carry their own `jsonObject?.takeIf { it.has(fieldName) }?.get(fieldName)` + null + `isJsonPrimitive` preamble. So this adds a generic helper, a type parameter and a lambda indirection to eliminate duplication that is still present in four other places — the net effect is one more layer to read through, with the duplication intact.

Two ways to make it worthwhile, either of which I'd be happy with:

- **Finish it.** Convert the other four. `getLong` becomes `getPrimitive(fieldName, jsonObject, 0L) { el -> el.asJsonPrimitive.let { if (it.isNumber) it.asLong else it.asString.toLongOrNull() ?: 0L } }` and the shared "has the field, isn't null, is a primitive" logic genuinely lives in one place. That's a real reduction and the coercion tests already cover all four, so the refactor would be verified on arrival.
- **Or drop it** and leave `getString` as the direct expression it was.

Worth noting these accessors were deliberately reshaped in #16656 to type-check uniformly and stop using exceptions as control flow. Introducing a wrapper for one of the five partially undoes that uniformity — after this change, `getString` reads differently from its four siblings for no behavioural reason.

### One asymmetry to resolve while you're here

`getPrimitive` treats `JsonNull` as the default *before* calling `extract`, but `getString(array, index)` immediately below still does its own `!el.isJsonPrimitive` check inline and is not routed through the helper. If the intent is "one place decides what a null/absent/wrong-typed field yields", the array overload belongs in it too — and if it doesn't fit the signature, that's useful evidence about whether the abstraction is the right shape.

No new tests, which is fine here given `JsonUtilsCoercionTest` already locks the behaviour — but it does mean the helper itself has no direct coverage of the `JsonNull → default` branch independent of `getString`.

---

dogi:
## Review — rating 82/100

Behaviour-preserving refactor, and I checked each of the six rewritten accessors rather than trusting the shape. CI green on all five checks, and `JsonUtilsTest` covers `getString`, `getBoolean`, `getInt` and `getFloat` across the valid / `JsonNull` / missing-key / wrong-type matrix, with `testTypeMismatchesAreQuiet` also touching `getLong` and `getString(array, index)`. That makes the green test run real evidence here, not just a compile check.

### Equivalence check

- **`getString(field, obj)`** — old returned `""` for `JsonNull`, non-primitive, or non-string; new gets `""` from `getPrimitive`'s null/`JsonNull` guard plus the `isJsonPrimitive && isString` test. Same.
- **`getString(array, index)`** — old returned `""` for out-of-range or non-primitive. `JsonNull` isn't a `JsonPrimitive`, so the new explicit `JsonNull` guard is redundant but harmless. Same.
- **`getBoolean` / `getInt` / `getLong`** — mechanical, same.
- **`getFloat`** — this one genuinely changed shape and is still equivalent. The old code had `if (jsonObject?.has(fieldName) != true) return@safeGet getInt(fieldName, jsonObject).toFloat()`, and `getInt` on an absent field (or a null object) always returns `0`, so that branch could only ever produce `0f` — exactly what the new default gives. Good removal; that delegation was dead weight.

### The refactor stops one step short

`getPrimitive` factors out the null/`JsonNull` guard and the `safeGet` wrap, but the duplication that actually repeats five times is the `isJsonPrimitive` check, and every call site still carries its own copy:

```kotlin
{ el -> if (!el.isJsonPrimitive) 0 else el.asJsonPrimitive.let { ... } }
```

Push that into the helper and hand the lambda a `JsonPrimitive`:

```kotlin
private fun <T> getPrimitive(default: T, elementSupplier: () -> JsonElement?, extract: (JsonPrimitive) -> T): T =
    safeGet({ default }) {
        val el = elementSupplier()
        if (el == null || !el.isJsonPrimitive) default else extract(el.asJsonPrimitive)
    }
```

`JsonNull` is not a `JsonPrimitive`, so the separate `el is JsonNull` check disappears too. The call sites then collapse to what they're actually about:

```kotlin
fun getBoolean(fieldName: String, jsonObject: JsonObject?): Boolean =
    getPrimitive(false, { fieldElement(fieldName, jsonObject) }) { it.asBoolean }

fun getInt(fieldName: String, jsonObject: JsonObject?): Int =
    getPrimitive(0, { fieldElement(fieldName, jsonObject) }) {
        if (it.isNumber) it.asInt else it.asString.toIntOrNull() ?: 0
    }
```

That's the version where the helper pays for itself. As written, `getBoolean` went from three obvious lines to three lines that require understanding a generic helper with a supplier lambda — a wash rather than a win. `getInt` / `getLong` / `getFloat` also share an identical "number → `asX`, else parse the string, else default" shape that could be one more helper, though that's optional.

Not blocking — this is an improvement as-is and the direction is right.

### Unrelated change

The `Dispatchers.resetMain()` `@After` added to `TeamsRepositoryBenchmarkTest` is a genuine fix (`setMain` without `resetMain` leaks the main dispatcher to whatever runs next in the same JVM fork, which is exactly the kind of thing that produces a mystery failure in one shard and not the other). It just has nothing to do with `JsonUtils`. Fine to leave, worth a mention in the PR body so it isn't invisible in the history.

Marking `ready`. The `isJsonPrimitive` consolidation is worth doing but doesn't need to gate this.

---

dogi (approved):

PR #16683, opened 2026-09-02
Phase 115: first tests for the untested team surfaces, and the defects they found

dogi:
Lane C of the Phase 115 parallel round. Draft — opened as the lane's first action so the lane has an inbound channel.

## Scope

The team surfaces in `flutter/lib/ui/teams/`. Starts by re-measuring what is actually covered (by symbol reference, not filename), then writes the first tests for whichever team surface genuinely has the thinnest coverage and fixes what they find.

Progress and findings are recorded in `flutter/PHASE_115_NOTES.md`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01RZc4FJw4km54hmmXkZ4KaS)_

### DIFF
--- a/flutter/PHASE_115_NOTES.md
+++ b/flutter/PHASE_115_NOTES.md
@@ -436,7 +436,140 @@ files are Lane B's.
 `flutter test` — **1860 pass**, against 1801 at the branch base: exactly the 59
 new tests, no regressions.
 
-## 13. Mutation-testing this lane's own tests
+## 13. The second audit pass, and the four defects this phase itself introduced
+
+Phase 110's lesson — an audit of the Kotlin ground truth does not audit the
+implementation — earned its keep. The second pass, pointed at this lane's own
+finished diff while it was green, found **four defects the phase introduced**
+and **two tests passing for the wrong reason**. All six are fixed below, each
+demonstrated failing first.
+
+### The four introduced defects
+
+1. **`respond(accept: true)` could strand an accepted join request. Data
+   loss, and mine.** `_userId()` was the one session read placed *after* its
+   own local write: `respondToRequest` has already converted the request row
+   into a `membership` with `isUpdated = true` by then. Awaiting `.future`
+   made that line able to **reject**, where the `.valueOrNull` it replaced
+   could only be null — so a rejecting session threw past the enqueue, and the
+   outbox is the only upload route (`TeamDao` has no pending sweep,
+   `TeamsUploader` no rescan). The accepted member existed on the leader's
+   device and nowhere else, permanently, compounded by §8.1's shadowing. The
+   user is now resolved *before* the local write, as the other four actions
+   already did.
+
+   **The general shape: `await`ing a future where a synchronous read used to
+   sit adds a throw to a code path that could not throw before.** Every such
+   swap needs its position relative to the writes re-examined, not just its
+   `try` boundary.
+
+2. **Three actions threw where they used to return `false`.** The doc comment
+   this phase wrote stated a caller contract — "callers keep the await inside
+   their own `try`" — and **not one of the five call sites satisfied it**:
+   `team_members_screen`'s `_handleMemberAction` shows its failure snackbar
+   off the returned `false`, and both the join button and the leave dialog are
+   fire-and-forget `onPressed`s. So the fix lost the user's error message and
+   escaped as an uncaught async error: strictly worse than what it replaced.
+   `_session()` now swallows the rejection and returns null, which resolves
+   the session properly *and* keeps every action's existing contract.
+
+   **A doc comment asserting a caller contract is worth nothing until the
+   callers are checked.** Writing one felt like discharging the obligation.
+
+3. **Double-tapping Send posted the comment twice.** Moving `_controller.clear()`
+   after the await — so a failed post leaves the text to retry — also let the
+   text survive the in-flight write, and the send button is never disabled.
+   Before the move, the second tap was a no-op *by accident*. A `_sending`
+   guard now holds both properties deliberately.
+
+4. **The null-`_rev` guard covered three of the four tombstones in the file.**
+   `TeamResourceActions.remove` enqueues the same
+   `{_id, _rev, _deleted: true}` payload four functions below the new helper
+   and did not use it. Worse, §6's corrected add gate *widened* the reach:
+   giving ordinary members the add FAB makes "link a resource offline, unlink
+   before the drain" a path many more users can take. **A fix that widens who
+   can reach a code path inherits that path's latent defects.**
+
+   Reported, not fixed: Kotlin's `removeResourceLink`
+   (`TeamsRepositoryImpl.kt:707-715`) does not delete or tombstone at all — it
+   blanks `resourceId = ""` and sets `updated = true`. That is a deeper
+   divergence than the guard, and it needs the §8.1 upload sweep to exist
+   first.
+
+### Corrections to claims this phase made
+
+- **The tie-break rationale was overstated.** Both the comment and the notes
+  claimed ties "could swap rank between two loads of unchanged data". They
+  could not, for a team under 32 members: Dart's `List.sort` uses a *stable*
+  insertion sort below that threshold, and `watchMembers` orders by
+  `userId ASC`, so the input order was already deterministic. `toSet()` is a
+  `LinkedHashSet` and preserves insertion order too, so the `Set` was a red
+  herring. The total order is kept — it is right for any size, and ranking
+  ties by name beats ranking them by an opaque id — but it is now documented
+  as **a deliberate choice against no specification**, not a defect fix.
+- **`skipLoadingOnReload` fixes the flash, not the work.** §4 named the defect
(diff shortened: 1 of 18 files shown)


dogi:
## Lane C status — head `8903332`

### The coverage measurement in the brief was wrong, in a way worth recording

The brief named `teams_screen.dart` (404 lines, one referencing test file) as the thinnest coverage-to-size ratio left in the port. Measured by **symbol reference** rather than filename, that is not the thinnest surface:

| file | lines | test files referencing it |
|---|---|---|
| `leaderboard/team_leaderboard_screen.dart` | 284 | **0** |
| `inline_comments.dart` | 220 | **0** |
| `teams_screen.dart` | 404 | 1 |

Two things the reference count hides. The leaderboard's **calculator** is tested, so the directory has a test file — just not one that touches the 284-line screen beside it. And `teams_screen.dart` holds *two* screens: `TeamsScreen` (the catalog list) and `TeamDetailScreen`. The one file referencing it tests only the latter's visit logging, so the team list had no coverage either.

So the thinnest surface was three screens, not one file. This lane covers all three plus the detail screen's gates.

### 59 new tests, 30 failing first, 23 defects

Every defect was demonstrated failing on the pre-fix code before it was touched.

| group | tests | failed first |
|---|---|---|
| `inline_comments` | 8 | 4 |
| leaderboard screen | 10 | 6 |
| `TeamsScreen` (catalog list) | 14 | 3 |
| `TeamDetailScreen` gates | 13 | 9 |
| resources/courses add+remove gates | 10 | 5 |
| `TeamMembershipActions` tombstones | 4 | 3 |

The largest is **access**: `buildPages` (`TeamDetailFragment.kt:74-92`) gives a non-member of a non-public team exactly two pages, Plan/Mission and Members. The port's link list was unconditional apart from three entries — so a non-member of a private team was offered its task list, survey list, resource list and discussion thread, and for an enterprise its **financial reports**.

**Gates in this area keep coming out wrong in the same direction Phase 99 fixed, and Phase 99 did not catch these three.** The detail screen refused to let a leader leave (Kotlin's `setupMyTeamButtons` has no leader test, and `markMembershipsForLeave` has none either — Kotlin has `isTeamLeader` and pointedly does not use it here); the resources add FAB required leadership where `TeamResourcesFragment.kt:51-52` requires plain membership; and the courses unlink required leadership where `TeamCoursesFragment.kt:44-46` requires being the team's **creator**. Leaving also took effect on one tap with no `confirm_exit` dialog and no last-member check, and a guest was offered membership.

The **`sessionProvider` read-but-never-watched** shape appears four more times (instances 5–11), including the first in a *provider* rather than a screen — where the "the router holds a `ref.listen`, so it is latent" mitigation does not apply at all.

### One finding is a negative, and it is the most useful one

The teams search matched the description where `TeamViewModel.applyFilters` matches the name only — fixed. But the obvious next step, "another flat `contains` like the ones Phases 96 and 97 replaced with the ranked `ResourcesSearchUtils` algorithm", **is wrong**: the Kotlin teams search genuinely has no ranking, no word-splitting and no accent folding. Adding them would have *introduced* a divergence. The port was already right on that axis and the tests now pin it.

### Reported, not fixed — outside this lane's file set

Full detail in `flutter/PHASE_115_NOTES.md` §8. The two to act on first:

1. **`updateTeam` never uploads, and permanently shadows the server's copy.** It sets `isUpdated: Value(true)` and returns; Kotlin's `UploadManager.uploadTeams()` sweeps `WHERE isUpdated = 1` every sync, and the port has no sweep — `TeamsUploader.types` has no plain-team entry. So a team edit never leaves the device *and*, because `TeamMapper.fromDoc` keeps local values for a flagged row, every future server-side change to that team is discarded on every sync from then on. Needs `teams_uploader.dart` + an outbox type.
2. **The catalog's root-team predicate is not Kotlin's.** Kotlin tests `teamId IS NULL OR TRIM(teamId) = ''` and never reads `docType`; `TeamDao.watchCatalog` tests `docType.isNull()`. A team document carrying `docType` is invisible to the port's catalog entirely, and the only in-repo evidence (`TeamsRepositoryBulkInsertTransactionTest.kt:55-62`) builds a root team *with* `docType: "team"`. Every Dart fixture omits it, so the suite cannot catch it. **Worth settling against a live Planet server before anything else here.**

Also reported: the member count is not Kotlin's (`COUNT(DISTINCT userId)` + a `users` existence join), team creation is unreachable, the list row has no action column so the "Requested" state is invisible, and the `teamLeaderboard` route has no pusher — the screen this lane just tested is unreachable in the app.

### Gate

`dart format` clean, `flutter analyze` clean, team suites green; the full run is finishing and a second `parity-auditor` pass is auditing this lane's own finished code (per Phase 110: an audit of the Kotlin ground truth does not audit the implementation). I'll follow up with both.

---

dogi:
## Second audit pass — head `2afb1ab`

The second `parity-auditor` pass, pointed at this lane's own finished diff **while it was green**, found four defects the phase itself introduced and two of my tests passing for the wrong reason. All six fixed, each demonstrated failing first. Full detail in `flutter/PHASE_115_NOTES.md` §13–14.

### The four I introduced

**`respond(accept: true)` could strand an accepted join request — data loss.** `_userId()` was the one session read placed *after* its own local write; `respondToRequest` has already converted the request row into a `membership` with `isUpdated = true` by then. My own fix — awaiting `.future` instead of reading `.valueOrNull` — made that line able to **reject** where it previously could only be null, and the outbox is the only upload route (`TeamDao` has no pending sweep, `TeamsUploader` no rescan). So the accepted member existed on the leader's device and nowhere else, permanently.

> **Awaiting a future where a synchronous read used to sit adds a throw to a path that could not throw before.** I applied that swap in five places and checked only the `try` boundary, not each one's position relative to its writes.

**Three actions threw where they returned `false`** — and the doc comment I wrote asserting "callers keep the await inside their own `try`" was true of **none of the five call sites**. `team_members_screen`'s `_handleMemberAction` shows its failure snackbar off that `false`, and the join button and leave dialog are fire-and-forget, so the fix silently removed the user's error message. `_session()` now swallows the rejection, keeping the old contract.

**Double-tapping Send posted twice.** Moving `clear()` after the await (so a failure leaves text to retry) let the text survive the in-flight write; before the move the second tap was a no-op *by accident*.

**The null-`_rev` guard covered three of four tombstones.** `TeamResourceActions.remove` needed it too — and §6's corrected add gate *widened* who can reach it. **A fix that widens access to a code path inherits that path's latent defects.**

### Corrections to claims I made earlier in this PR

- **The tie-break rationale was overstated.** Dart's `List.sort` is stable below 32 elements and `watchMembers` orders by `userId ASC`, so ties were already deterministic. The total order is kept, but as a deliberate choice against no specification — not a defect fix. (`toSet()` is a `LinkedHashSet`, so the `Set` was a red herring too.)
- **`skipLoadingOnReload` fixes the flash, not the work.** I claimed the defect included re-running three queries per keystroke; the flag only suppresses the loading branch. **Half that defect is still open** — needs a debounce on `teamsSearchProvider`.
- Two comments cited `docs/kotlin-to-flutter-migration.md` for the finances surplus, which lives only in `CLAUDE.md`'s Phase 99 prose; and the leaderboard's class docs named Kotlin files that do not exist.

### One test-hygiene finding worth generalising

A negative assertion of mine named `'Team resources'` where the label is `'Resources'`, so it could never match and would have passed with the gate removed. Mutation testing caught it only because a *sibling* assertion detected the mutation first. **A negative assertion that passes tells you nothing until you have seen it fail.**

### Still open, for whoever picks this up

- A **debounce** on the teams search, per the correction above.
- **Kotlin's `removeResourceLink` (`TeamsRepositoryImpl.kt:707-715`) does not delete or tombstone at all** — it blanks `resourceId = ""` and sets `updated = true`. A deeper divergence than the guard I added, and it needs the `updateTeam` upload sweep (§8.1) to exist first.
- `voices_screen.dart:301-304` still carries a duplicate of `initialFor`. Byte-identical and harmless, and `lib/ui/voices/` is outside this lane — unlike `team_members_screen`, which was both broken *and* in `lib/ui/teams/`, so it was fixed. **Cross a boundary for a defect, not for tidiness.**

### Incidental: a stale note in CLAUDE.md

The **"Phase 75 harness trap"** is stale for the current graph — `teamsRepositoryProvider` reaches `planetApiProvider` and `appDatabaseProvider`, **not** `planetPrefsProvider`. The live trap is `appDatabaseProvider → AppDatabase.open()`, which `wrapScreen` already redirects to `AppDatabase.memory()`. Worth correcting when CLAUDE.md is next touched (this lane must not edit it).

### Gate

`dart format` clean, `flutter analyze` clean, **`flutter test` 1866 pass** (1801 at the branch base). CI in progress on `2afb1ab`; `myPlanet labels` green, nothing red.

Process note: the auditor left a mutation probe behind in `teams_provider.dart` rather than restoring it, which reverted a defect fix. `git status` caught it before the next commit, and I verified no pushed commit contains it — but **a subagent with write access silently un-fixing a defect, in a tree that then gets committed, is a real hazard**; checking `git status` before every commit in such a session is load-bearing, not hygiene.

---

PR #16684, opened 2026-09-02
sync: smoother repositories json utils documents mapping (fixes #16643)

### LINKED ISSUE #16643: parsing loop copy-pasted 8 times
```
JSON-array parsing loop copy-pasted 8 times: CoursesRepositoryImpl.kt:601,779, HealthRepositoryImpl.kt:83, NotificationsRepositoryImpl.kt:407, SubmissionsRepositoryImpl.kt:646,
TeamsRepositoryImpl.kt:1251,1277,1290, UserRepositoryImpl.kt:1046 — same for (j in jsonArray) { val obj = j.asJsonObject; ... } header everywhere. A JsonArray.asObjects():     
List<JsonObject> extension removes all of them.
```

Okuro3499:
fixes #16643
Extract repeated CouchDB sync document parsing (get doc, filter _design) into a reusable JsonArray.toSyncDocuments() extension in JsonUtils.kt. Replace duplicated loops in 6 repository bulkInsert methods. Also clean up minor style in JsonUtils (isNotEmpty, safe cast).

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 6931
-        versionName = "0.69.31"
+        versionCode = 6932
+        versionName = "0.69.32"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/repository/CoursesRepositoryImpl.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/CoursesRepositoryImpl.kt
@@ -46,6 +46,7 @@ import org.ole.planet.myplanet.utils.ExamAnswerUtils
 import org.ole.planet.myplanet.utils.JsonUtils
 import org.ole.planet.myplanet.utils.UrlUtils
 import org.ole.planet.myplanet.utils.Utilities
+import org.ole.planet.myplanet.utils.toSyncDocuments
 
 class CoursesRepositoryImpl @Inject constructor(
     private val progressRepository: ProgressRepository,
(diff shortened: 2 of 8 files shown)


dogi:

Clean, faithful deduplication. The same eight-line loop — unwrap `doc`, read `_id`, skip `_design`, accumulate — appeared in nine `bulkInsert*FromSync` methods across six repositories, and `toSyncDocuments()` collapses all of them. I walked each call site against the original and the semantics hold, including the two that needed more than a straight swap:

- **`TeamsRepositoryImpl.bulkInsertFromSync`** built `documentList` and `ids` in one pass; `syncDocs.map { it.first }` plus `syncDocs.forEach { (_, jsonDoc) -> ... }` produces the same two views in the same order, and the `withTransaction` wrapper (and the comment explaining why the whole page commits once) is untouched.
- **`CoursesRepositoryImpl.insertCertificationsFromSync`** used `continue` inside the loop to skip `_design`; returning the id from the extension lets the body become a `map` over `(id, jsonDoc)` without losing the id it needs for `_id`.

Returning `Pair<String, JsonObject>` rather than just the document is the detail that makes this work — two of the nine sites need the id, and a document-only helper would have forced them to re-extract it.

The `NotificationsRepositoryImpl` import removal is a good catch: dropping `JsonUtils` there means the loop really was its only use, which is the kind of thing that's easy to leave behind in a sweep like this. CI green on all five checks.

Two things, neither blocking.

### The new helper has no direct test

`toSyncDocuments()` now gates document ingestion for courses, certifications, health examinations, notifications, submissions, teams, team tasks, team activities and achievements. A bug in it — a mis-inverted `_design` check, say — corrupts every sync path at once, and nothing tests it in isolation. Three assertions would cover it: a `_design` doc is dropped, a normal doc yields the right id/document pair, and an element with no `doc` wrapper produces the empty-object fallback that `JsonUtils.getJsonObject` returns. Given how thin the function is, that's five minutes for a lot of protection.

Worth being explicit about that last case, incidentally: an element lacking a `doc` field yields `JsonObject()` and an empty `_id`, which doesn't start with `_design`, so it's *kept* and passed downstream as an empty document. That matches the previous behaviour exactly, so it's not a regression — but centralising the logic is the moment to decide whether silently forwarding empty documents is what you want.

### Placement and scope

The extension is appended to `JsonUtils.kt` as a top-level function outside the `object`. That mirrors what `AppIdentity.kt` does for `addAppIdentity`, so it's consistent with the emerging pattern — though a `SyncDocuments.kt` would signal "this is sync-layer, not JSON-layer" more clearly, since the `doc`/`_design` shape is a CouchDB `_all_docs` concern rather than a general JSON one.

Minor: the `getJsonObject` change from `if (el is JsonObject) el else JsonObject()` to `el as? JsonObject ?: JsonObject()` is equivalent and slightly tidier, but it's unrelated to the extraction and widens the diff.

---

J-S-webskas (approved):

dogi (approved):

PR #16686, opened 2026-09-03
all: smoother time utils date formatting (fixes #16642)

### LINKED ISSUE #16642: duplicate date formatters
```
TimeUtils.kt has 6+ near-duplicate date formatters (getFormattedDate, getFormattedShortDate, getFormattedDateWithTime, formatDateTZ, two formatDate overloads) differing only in
pattern/zone; convertToISO8601 is the sole holdout still using Calendar instead of java.time.
```

Okuro3499:
fixes #16642


dogi:
## Review — rating 88/100

Careful work. It removes about 40 lines of copy-pasted `try / printStackTrace / fallback` and replaces the last `java.util.Calendar` use in the file with `java.time`, without changing a single observable result. CI green on all five checks.

The part I expected to be wrong is right, so it's worth spelling out.

### `convertToISO8601` preserves lenient rollover, deliberately

`Calendar.getInstance()` is lenient, so `calendar.set(2023, 1, 29)` silently normalised Feb 29 of a non-leap year to Mar 1, and the old code then read the normalised fields back out. A naive port to `LocalDate.of(y, m, d)` would have thrown `DateTimeException` there and fallen into the `catch`, returning the raw input string — a real regression, and one `testConvertToISO8601_feb29NonLeapYear` would have caught.

The replacement avoids it on purpose:

```kotlin
LocalDate.of(parts[0].toInt(), 1, 1)
    .plusMonths(parts[1].toInt() - 1L)
    .plusDays(parts[2].toInt() - 1L)
```

Anchoring at Jan 1 and adding offsets reproduces `Calendar`'s rollover exactly, including negative and out-of-range components. Anchoring at day 1 also dodges `plusMonths`' end-of-month clamping, which would have bitten a `LocalDate.of(y, 1, 31).plusMonths(1)` formulation. I walked `2023-02-29` → `2023-03-01`, `2026-03-31`, `2026-01-0` → `2025-12-31` and they all match the old output.

The existing tests pin this well: nine cases on `convertToISO8601`, including the leap-year rollover and `testConvertToISO8601_ignoresLocalTimeZone`, which confirms the result stays timezone-independent (it was under `Calendar` too, since the code only read back fields it had just set).

### `Instant.ofEpochMilli` moved outside the `try` — checked, safe

Every call site now builds the `Instant` as an argument, before `formatInstant`'s `try` begins. That's a real structural change, but `Instant.ofEpochMilli` can't throw for any `Long`: `Long.MAX_VALUE / 1000 ≈ 9.2e15` seconds is well inside `Instant.MAX`'s `≈3.16e16`. `testGetFormattedDateWithExtremeValues` exercises `Long.MAX_VALUE` and `Long.MIN_VALUE` directly and passes, so this is settled empirically as well.

### The `() -> DateTimeFormatter` parameter is load-bearing

Worth noting for future readers, since six of the eight call sites pass a trivial `{ someLazyVal }` and it looks like ceremony. It isn't: `formatDate(date, format)` and `getFormattedShortDate` build their formatter via `formatterFor(...)`, and `DateTimeFormatter.ofPattern` throws `IllegalArgumentException` on a malformed pattern. In the old code that construction sat inside the `try`; the lambda is what keeps it there. Passing a `DateTimeFormatter` directly would have moved that throw outside the guard and turned a bad user-supplied pattern into a crash. Getting this right is the non-obvious part of the refactor.

It also keeps the `by lazy` initialisers inside the guard, which is a small bonus.

### Minor

- `formatInstant` catches `Exception` with `printStackTrace()`. That's the pre-existing pattern faithfully preserved, and all of this is synchronous so there's no `CancellationException` hazard. The upside of the consolidation is that swapping in real logging is now a one-line change in one place instead of eight.
- The `'Z'` literal in `iso8601Formatter` labels a local wall-clock date as UTC. Pre-existing, and `testConvertToISO8601_ignoresLocalTimeZone` pins it as intentional, so not this PR's to change — just be aware the suffix is decorative.

Marking `ready`.

---

dogi (approved):

PR #16697, opened 2026-09-03
Phase 117 (flutter): render the course-progress grid

dogi:
Lane B of the Phase 115–117 parallel round. Ports Kotlin's `CourseProgressActivity` + `ProgressGridAdapter` — the per-step progress grid — and wires it to the port's `ProgressRepository.courseProgress`, which until now had no UI caller at all.

Draft while in progress. Notes land in `flutter/PHASE_117_NOTES.md`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01YYCqCfkePzhEpUr1ABkGER)_


dogi:
Status, head `67a9b01`. Gate green: format clean, `flutter analyze` clean, **1917 tests pass** (base 1896).

**The predecessor's description of the target was wrong.** Phase 113's note said `ProgressGridAdapter` renders "the per-step question count and mistake total". It renders neither. `row_my_progress_grid.xml` holds one view and `onBindViewHolder:37-50` writes only `R.string.percentage` — the share of the step exam's questions **answered**, with the cell green when that share is all of them, yellow when not, and blue-with-no-text when the step has no attempted exam. The question count is only the denominator; mistakes are not computed on this path at all — they belong to the My Progress **list row**, the screen the grid opens *from*.

So `courseProgress` was not "correct but unrendered": it computed two numbers the grid does not have and none of the three it does. It now returns `CourseProgressData` (title / current / max / one cell per step).

**Placement** mirrors Kotlin: its own route `/courses/progress/:courseId`, reached by tapping a My Progress row, which is the only thing that opens `CourseProgressActivity`.

**End-to-end, no fake between the answer and the pixel** (`course_progress_screen_test.dart`): real `SubmissionsRepository.startExamSession` → a wrong `saveExamAnswer` (refused, `mistakes` 1) → the right one → the grid renders `50.0%` for one of two questions, through the real `ExamMapper.fromCourseDoc` step join. The chain is intact.

**Six defects, each demonstrated failing on pre-fix code:**

| # | defect | pre-fix failure |
|---|---|---|
| 1 | the My Progress row opened the course **detail** screen, not the grid | `Found 0 widgets with text "GRID_ROUTE course-1"` |
| 2 | the row was clickable with no progress figures (Kotlin gates the listener on both being non-null) | `means one was found but none were expected` |
| 3 | `courseProgressStreamProvider` read the session as `.valueOrNull`, and a null `shelfUserId` makes `watchCourses` drop the shelf predicate — so My Progress first emitted **the whole catalogue** | `TimeoutException: Future not completed` |
| 4 | per-step mistakes keyed by raw exam id, step number recovered by running a `(\d+)` regex over it, and the last answer overwrote instead of summing | `Expected: {1: 5} Actual: {0: 3}` |
| 5 | a submission was attributed to a course by substring, so `exam@course-10` counted against `course-1` | `Expected: <0> Actual: <4>` |
| 6 | `getParentBaseId` divergence on a leading `@` (alignment only — unreachable in practice, recorded not claimed) | — |

Reproduced deliberately: a fully answered exam reads **"100.0%"**, not "100%" — Gson boxes the `Double` and the adapter renders `getAsString()`; Dart's `double.toString()` is the same shortest-round-trip algorithm, and the zero-question branch stores an `int`, so `percentage` is a `num`. Rounding it would show numbers the shipping app never shows.

One deliberate boundary crossing: `lib/ui/router.dart` (one route constant, one nested `GoRoute`). Without it the new screen is unreachable, which is the pathology this phase exists to end. Second parity audit on the implementation is running.

---

dogi:
Second parity audit done, findings closed. Head **`2f95037`** — format clean, analyze clean, **1922 tests**.

**First, a process failure worth more than any of the findings.** The review pass mutation-tested by editing the shared working tree, and my next commit (`7144bdc`) swept a mutation in — the `??=` in `_stepCell` replaced by plain assignment, i.e. precisely the `!ob.has(...)` guard this phase is about, deleted — plus three of its probe files. `95e0f17` restores it. Two cheap lessons: a review pass that mutates code must do it on a copy, not the tree the author commits from; and a `git add -A` after handing the tree to anything else is a commit of unknown content. This also explains the "flaky" full-suite failure I couldn't reproduce in isolation earlier — it wasn't flaky.

**What survived the audit**, which is most of the diff: `_stepCell` — the reviewer could not construct data where it and `getExamObject` disagree, and both mutations it tried were already caught. And `percentageLabel` was verified exhaustively rather than argued: all 20,700 values of `((double) n / q) * 100` for `q` in 1..200, JDK 21's `Double.toString` against Dart's, **zero differences**.

**Four more defects, each failing first:**

| # | defect | pre-fix failure |
|---|---|---|
| 7 | **the My Progress list was frozen for the process lifetime.** `CoursesProgressFragment.onViewCreated:31` calls `loadCourseData()` unconditionally and is pushed with `addToBackStack`, so it re-reads on every entry. Worse for being half a pair — the grid *is* `autoDispose`, so after an exam the row said "1 mistake" while the grid it opened showed the new percentage | `Expected: <3> Actual: <1>` |
| 8 | **the mistakes table listed exams in the wrong order.** Kotlin's Gson-round-tripped `HashMap` walks keys `"0".."9"` ascending; a Dart map kept the order answers came off the DAO | `did not find a value matching '3' following expected prior values` |
| 9 | **a partial last grid row was centred.** `GridLayoutManager` puts item *n* in span `n % 4`, so a 9th step sits in the first column | `Expected: <24.0> Actual: <126.0>` |
| 10 | **the first frame of My Courses was the whole catalogue** — `coursesStreamProvider` read the session `.valueOrNull` and passed it as `shelfUserId`, which `watchCourses` drops when null. Same defect as #3, one provider up the same file | reasoning + #3's measured shape (it self-corrects on rebuild, so no stable assertion) |

**Two quirks that were right and untested.** Both of `submissionMap`'s oddities — the total accumulating across submissions while the breakdown keeps only the last — were correct and *neither was pinned*: hoisting the map, or moving the accumulator inside the loop, each passed the entire suite. That's exactly the shape a later reader tidies away, and my own comment was the only thing in the way. Now pinned (`{1: 3}` vs `{0: 2, 1: 3}`; `5` vs `3`).

**Four tests that passed without asserting what they claimed**, including one whose fixture seeded answers with no `questionId` — data on which Kotlin counts nothing and draws no table, so it couldn't demonstrate the keying it was named for.

**One fix considered and declined**, reasoning at the code: awaiting the session in `filteredSortedCoursesProvider` trades a one-frame "Not Started" flash for an **empty** course list when the session rejects, where Kotlin's fallback for unavailable progress is to show the list unfiltered.

**For Lane C:** `courseProgressCount` already has a *human* translation shipping in all five Kotlin locales (`values-ne:1010` even reorders the placeholders). `tool/arb_from_strings_xml.dart` skips placeholder keys, so it will never derive them — one key, five files, all in your set.

Cross-lane crossing now shrinks to a single nested `GoRoute` in `router.dart`: the `Routes.courseProgress` constant was read by nothing (the pusher derives from the non-pattern `Routes.myProgress`) and is gone.

---

PR #16698, opened 2026-09-03
courses: smoother repository base exam progress user scoping (fixes #16695)

### LINKED ISSUE #16695: Exam progress leaks
```
Exam progress leaks between users on a shared device — CourseProgressDao.updatePassedByCourseAndStep has no userId in its WHERE clause, unlike every sibling query in that file. 
One user finishing a step flips passed for every other local user's row too.
```

Okuro3499:
fixes #16695
CourseProgressDao.updatePassedByCourseAndStep had no userId in its WHERE clause, unlike every other query in the DAO. On a device shared by multiple learners, one user finishing a step's exam flipped `passed` on every other local user's course_progress row for that same course/step too.

Threads userId from ExamTakingFragment (already populated via UserSessionManager before the exam-completion call site) through BaseExamFragment.saveCourseProgress -> CoursesRepository.updateCourseProgress -> the DAO query, which now matches userId IS :userId.

## Summary by CodeRabbit

* **Bug Fixes**
  * Course progress updates are now scoped to the specific user, preventing one user’s exam progress from affecting another’s.
  * Exam completion saves progress with the applicable user account.

* **Tests**
  * Added coverage confirming that course progress updates are associated with the correct user.
  * Updated exam progress tests to reflect user-specific progress handling.


dogi:
## Review — rating 86/100

Real data-corruption fix, and the diff is exactly the size it should be. CI green on all five checks.

### The bug is as bad as it looks

On master:

```sql
UPDATE course_progress SET passed = :passed WHERE courseId = :courseId AND stepNum = :stepNum
```

`course_progress` is keyed per user — `CourseProgress.userId` exists and is indexed. So on a shared tablet, one learner finishing an exam step flipped `passed` on *every* learner's row for that course and step. myPlanet's whole deployment model is several learners sharing one device, so this isn't a corner case.

### The fix matches the file's own conventions

Two things I checked rather than assumed:

**`IS` versus `=`.** `userId` is nullable (`var userId: String? = null`), and the new predicate uses `userId IS :userId`. That's correct — `=` never matches `NULL` in SQL — and it's what the sibling queries in the same DAO already do:

```sql
SELECT * FROM course_progress WHERE userId IS :userId                                    -- getByUser
SELECT * FROM course_progress WHERE courseId IS :courseId AND userId IS :userId AND stepNum = :stepNum  -- findByCourseUserAndStep
```

**The index already covers it.** `CourseProgress` declares `Index(value = ["courseId", "userId", "stepNum"])`, and the new `WHERE` has equality on exactly those three columns, so the composite index is fully usable. The old query could only use the `courseId` prefix. This is faster as well as correct.

Worth noting that `ProgressRepositoryImpl.saveCourseProgress` — the other path that writes progress — already scoped by user via `findByCourseUserAndStep(courseId, userId, stepNum)`. `updatePassedByCourseAndStep` was the odd one out, which is good corroboration that user-scoping is the intended contract here rather than a new rule being imposed.

### Call-site coverage is complete

`updatePassedByCourseAndStep` has exactly one caller chain — `BaseExamFragment.continueExam` → `saveCourseProgress` → `CoursesRepository.updateCourseProgress` → the DAO — and all four links are updated, including the abstract signature and both test overrides. Nothing else in `app/src/main` calls it.

`user` is populated in time: `ExamTakingFragment.initializeExamData()` sets `user = userSessionManager.getUserModel()` before the exam is playable, and line 183 of `BaseExamFragment` already passes `user?.id` elsewhere, so the new argument is consistent with how the field is used in that class.

### Two suggestions

**Guard the null user explicitly.** `updateCourseProgress` early-returns on a blank `courseId` but not on a blank `userId`. With a null user the statement becomes `userId IS NULL`, which matches only rows that have no user — in practice none, since `ProgressRepositoryImpl` always writes a `userId`, so it's a silent no-op rather than corruption. Still, `if (userId.isNullOrEmpty()) return` alongside the existing guard says what's meant and stops a future NULL-userId row from being cross-updated.

**The new test can't fail for the right reason.** `testUpdateCourseProgressScopesUpdateToTheGivenUser` mocks `courseProgressDao`, so it verifies the repository forwards its argument — it would pass identically if the SQL still had no `userId` clause. The behaviour worth locking in is the scoping itself, which needs a Room in-memory test: insert progress rows for `user-a` and `user-b` on the same course and step, call `updatePassedByCourseAndStep(courseId, stepNum, true, "user-a")`, assert `user-a`'s row flipped and `user-b`'s didn't. That test fails on master and passes here, which is the property you want recorded.

Neither blocks. Marking `ready`.

---

J-S-webskas (approved):

dogi (approved):

PR #16701, opened 2026-09-03
courses: smoother filter controlling (fixes #16634)

### LINKED ISSUE #16634: progress filter reset
progress filter should also be cleared when reset button is clicked

https://github.com/user-attachments/assets/2d794e24-009f-48a5-82cd-24bf35094d3e

ragilzakaria:
https://github.com/user-attachments/assets/548469f2-07a0-4ecc-9a82-a0bea46aeae0

fixes #16634 

## Summary by CodeRabbit

- **Bug Fixes**
  - Clearing course filters now resets all selections, including search text, grade, subject, progress, and tags.
  - The filter panel is hidden after filters are cleared.
  - Filter chips now stay synchronized with the current filter state.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 7020
-        versionName = "0.70.20"
+        versionCode = 7021
+        versionName = "0.70.21"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/ui/courses/CourseFilterController.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/courses/CourseFilterController.kt
@@ -110,7 +110,10 @@ class CourseFilterController(
     }
 
     private fun setupClearTagsButton() {
-        rootView.findViewById<View>(R.id.btn_clear_tags).setOnClickListener { clearAll() }
+        rootView.findViewById<View>(R.id.btn_clear_tags)?.setOnClickListener {
+            rootView.findViewById<View>(R.id.card_filter)?.visibility = View.GONE
+            clearAll()
+        }
     }
 
     fun addTag(tag: TagEntity) {
--- a/app/src/main/java/org/ole/planet/myplanet/ui/courses/CoursesFragment.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/courses/CoursesFragment.kt
@@ -222,9 +222,12 @@ class CoursesFragment : BaseRecyclerFragment<MyCourse?>(), OnCourseItemSelectedL
         )
         filterController.setup()
 
+        val chipRow = view?.findViewById<LinearLayout>(R.id.chip_filter_row)
         var lastState: FilterState? = null
         var isFirstEmission = true
         collectLatestWhenStarted(filterController.filterState) { state ->
+            chipRow?.let { renderCourseChipSelection(it) }
+
             if (isFirstEmission) {
                 isFirstEmission = false
                 if (!state.isActive) {
@@ -327,9 +330,6 @@ class CoursesFragment : BaseRecyclerFragment<MyCourse?>(), OnCourseItemSelectedL
             bottomSheet.visibility = View.GONE
         }
         requireView().findViewById<View>(R.id.btn_collections)?.setOnClickListener {
-            bottomSheet.visibility = View.GONE
-        }
-        requireView().findViewById<View>(R.id.btn_clear_tags)?.setOnClickListener {
             bottomSheet.visibility = View.GONE
         }
         orderByDate = requireView().findViewById(R.id.order_by_date_button)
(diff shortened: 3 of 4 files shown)


dogi:
## Review — rating 74/100

Right diagnosis, and the new test file is the best thing in this batch. But the fix works around the bug instead of removing it, and half the diff is unreachable. CI green on all five checks.

### The bug

Two listeners are registered on the same view, and the second silently replaces the first:

- `CourseFilterController.setup()` — called from `CoursesFragment` line 223 — registers `clearAll()` on `R.id.btn_clear_tags`
- `CoursesFragment` line ~335 registers `bottomSheet.visibility = View.GONE` on `requireView().findViewById<View>(R.id.btn_clear_tags)`

Line 335 runs after line 223, so the fragment's listener wins and `clearAll()` never ran. Exactly #16634.

### The `CourseFilterController` half of the fix is dead code

This PR adds the missing behaviour to *both* registration sites. Since only one survives — and it's still the fragment's, registered ~110 lines later — the change to `setupClearTagsButton()` can never execute in the app.

That matters for the new test. `CourseFilterControllerTest` constructs the controller directly and calls `controller.setup()` without `CoursesFragment` in the picture, so `btnClear.performClick()` fires the controller's listener: the one production never reaches. The assertions pass, and the behaviour they describe does happen in the real app — but via the *other* listener, which no test covers. Change the fragment's listener later and the test stays green.

Your own #16702 fixes the identical bug in `ResourcesFragment` the right way: it **deletes** the clobbering `binding.btnClearTags.setOnClickListener { bottomSheet.visibility = View.GONE }` from `additionalSetup()` and folds that one line into the surviving handler. Doing the same here — drop the `CoursesFragment` listener at line 335, keep only the controller's, since the controller owns the filter state — leaves one registration, no ambiguity, and makes the new test exercise the real path.

I checked that the two are interchangeable: `CoursesFragment` line 322 is `val bottomSheet = requireView().findViewById<View>(R.id.card_filter)`, which is the same view the controller now hides via `R.id.card_filter`. So consolidating loses nothing.

While you're there, `if (::filterController.isInitialized) { filterController.clearAll() }` silently does nothing when the guard fails, which would reproduce the original bug invisibly. Consolidating removes the need for the guard.

### The chip re-render

Adding `renderCourseChipSelection(chipRow)` to the `filterController.filterState` collector is a genuine fix, and it happens to also solve a gap in #16647 — `restoreFilterState` there sets `progressFilter` but never repaints the chip row, because `setupCourseFilterChips()` runs at line 169, well before the restore at line ~225. Good catch, whether or not it was the intent.

Two small things: `view?.findViewById<LinearLayout>(R.id.chip_filter_row)` runs on every state emission and could be resolved once; and it's placed before the `isFirstEmission` branch, which is fine but reads as though it were part of that logic.

### The test file

Inflating the real `fragment_my_course` layout, driving `performClick()`, and asserting the full reset — search text, both spinners, progress filter, tags, and `card_filter` visibility — is exactly the right shape. Three notes:

- **No `runTest`.** `TestScope(UnconfinedTestDispatcher())` is built by hand and never cancelled, so anything the controller launches outlives the test and failures inside launched coroutines aren't attributed to it. `@Test fun x() = runTest { ... CourseFilterController(rootView, backgroundScope, ...) }` gives you auto-cancellation and proper failure attribution, and matches the convention in `docs/TESTING.md`.
- **Duplicated setup.** Both tests repeat the same six lines of context / theme / inflate / dispatcher / scope. A `@Before` would carry it.
- `@Config(manifest = Config.NONE)` isn't needed on current Robolectric and can be dropped.

### Conflict warning

#16647 edits `CoursesFragment.kt` at both of the places this PR touches — the `filterState` collector around line 225 and the `btn_clear_tags` listener around line 335. These two will conflict. Worth agreeing an order with @J-S-webskas; if this one goes first with the consolidation above, the other PR's job gets smaller.

Marking `change` — the consolidation is a few lines and turns a working-by-redundancy fix into a fix.

---

dogi:
**Review — rating: 87/100**

This is the same bug as #16892, one screen over, and the fix is the same shape.

`btn_clear_tags` had two click listeners registered on it: the real one in `CourseFilterController.setupClearTagsButton` that calls `clearAll()`, and a second in `CoursesFragment` that only did `bottomSheet.visibility = View.GONE`. The fragment's ran later and silently replaced the controller's, so "clear tags" closed the filter sheet and never cleared anything. Deleting the duplicate and moving the visibility change into the controller's handler is right, and adding `?.` to the `findViewById` lookups removes an NPE path on layouts where the button isn't present.

The `renderCourseChipSelection` call at the top of the `filterState` collector is the other half — the chip row wasn't re-rendering when filter state changed, so cleared tags stayed visually selected even once the underlying state was empty. Driving it off the same flow that already drives the list keeps the two in sync by construction rather than by remembering to call it from each mutation site.

Two notes, neither blocking:

- `chipRow` is resolved once outside the collector via `view?.findViewById(...)`. That's safe because `collectLatestWhenStarted` is scoped to the view lifecycle, so the whole block is re-established when the view is recreated and the captured reference can't outlive it — worth a comment, since a stale-view capture is what this pattern usually looks like.
- The render now runs on *every* emission including the first, ahead of the `isFirstEmission` branch. Fine if it's idempotent and cheap, which chip re-rendering should be, but it's one more full re-layout on the initial paint.

Sequencing note: #16647 also touches course chip rendering and adds `renderCourseChipSelection` work of its own. This PR is small, green, and self-contained; that one needs its branch fixed before it can land at all, so this should go first.

CI green 5/5.

**Verdict:** mergeable.

---

Okuro3499 (approved):

dogi (approved):

PR #16702, opened 2026-09-03
resources: smoother download status filter resetting (fixes #16635)

### LINKED ISSUE #16635: download status filter reset
download status filter should also be cleared when reset button is clicked
https://github.com/user-attachments/assets/3653e7f7-9d39-402c-a257-7b99e0a346e4

ragilzakaria:
https://github.com/user-attachments/assets/dc5f12d7-e6e9-4ef7-b7c3-bb584beaad26

fixes #16635


dogi:
## Review — rating 88/100

Correct diagnosis, minimal fix, and it removes the bug rather than working around it. CI green on all five checks.

### The bug, confirmed

Two `setOnClickListener` calls land on the same view, and the later one wins:

```kotlin
private val clearTags get() = binding.btnClearTags   // line 80
...
clearTagsButton()   // line 311 — registers the full reset on clearTags
setupUI(...)
additionalSetup()   // line 313 — registers `bottomSheet.visibility = View.GONE` on binding.btnClearTags
```

`clearTags` and `binding.btnClearTags` are the same view, and `additionalSetup()` runs two lines later, so on master tapping "clear tags" **only closed the filter sheet**. It didn't clear tags, didn't reset `selectedDownloadFilterIndex`, didn't clear the search box, and didn't reset levels / mediums / subjects / languages. That's #16635 precisely, and it explains why the reported symptom was specifically about the download-status filter surviving a clear.

### The fix is right

Deleting the clobbering listener and folding its one line into the surviving handler is the correct shape. I confirmed the substitution is exact: line 868 is `val bottomSheet = binding.cardFilter`, so `binding.cardFilter.visibility = View.GONE` is the same view the removed listener hid. Nothing is lost.

The surviving handler was already thorough — it resets the download filter index, re-renders the chips, clears tags and search text, empties all four facet lists, cancels the in-flight `searchJob`, and re-runs `applyFiltersAndUpdateUI()`. It just never got to run.

### Worth a test

This failure mode is invisible in review — nothing about `additionalSetup()` hints that it stomps a listener registered elsewhere — and it's easy to reintroduce the next time someone adds a click handler in a second place. A Robolectric test in the shape of the one you wrote in #16701 would lock it in:

```kotlin
binding.btnClearTags.performClick()
assertEquals(0, fragment.selectedDownloadFilterIndex)
assertEquals(View.GONE, binding.cardFilter.visibility)
```

`selectedDownloadFilterIndex` is already `internal var`, so it's reachable from the test source set.

### Same class of bug next door

`additionalSetup()` registers listeners on several views that other setup functions also configure. Worth a scan for other duplicate registrations while this is fresh — the pattern that produced this bug is still present in the file.

### Note on #16701

That PR fixes the identical bug on the courses side, but by adding the behaviour to *both* competing listeners rather than deleting one, which leaves half the change unreachable. This PR's approach is the better of the two — worth applying the same treatment there.

Marking `ready`.

---

J-S-webskas (approved):

Okuro3499 (approved):

dogi:

---

PR #16705, opened 2026-09-03
resources: smoother filter label formatting (fixes #16283)

### LINKED ISSUE #16283: resources: Medium filter doesn't have consistent capitalization
Image

J-S-webskas:
Recording_20260903_134337

## Summary by CodeRabbit

* **New Features**
  * Medium resource filters now display localized names when available.
  * Untranslated medium entries show clear, formatted fallback labels.
  * Filter selections continue to work correctly with localized display names.
  * Added localized labels for Text / HTML and HTML media types in Arabic, Spanish, French, Nepali, Somali, and English.


dogi:
**Review — rating: 74/100**

The important thing here is right: the display label never reaches the filter. But a PR about label presentation ships two labels that no locale but English will read.

**The filter values are safe — this was the thing to check.** Swapping displayed text on a `CHOICE_MODE_MULTIPLE` list is exactly where you'd expect the pretty name to leak into the query. It doesn't: `onItemClick` reads `adapterView.getItemAtPosition(i) as String`, which returns the adapter's backing item, and the backing list is still `keys` — the raw `mediaType` values. So `addToList(s, selectedMeds)` keeps storing `"video"`, not `"Videos"`, and `filterListener.filter(...)` is unchanged. The restore path matches: `setItemChecked(i, selectedSet.contains(keys[i]))` compares raw key against raw key. Display and value stay properly separated.

I also checked `rowlayout.xml` — its root really is a `CheckedTextView` carrying `@id/checkBoxRowLayout`, so `findViewById<CheckedTextView>` resolves rather than throwing, and re-setting the text on every `getView` keeps recycled rows correct.

**Two labels aren't localized.** `"text/html" -> "Text / HTML"` and `"html" -> "HTML"` are hardcoded English literals sitting among five branches that correctly go through `getString(R.string.…)`. The app ships ar / es / fr / ne / so, so those two rows stay English everywhere. `HTML` is an acronym and arguably fine as-is, but `Text / HTML` contains a real word, and in a PR whose entire purpose is how these labels read, this is the convention that most needs following. Two new entries in `values/strings.xml` plus the five translations closes it.

**`setMediumAdapter` is `setAdapter` copied.** Choice mode, the `ArrayList(it) ?: return`, the `setItemChecked` loop — all identical; the only difference is the `getView` override. Adding an optional mapper to the existing function gives you one code path instead of two that will drift:

```kotlin
private fun setAdapter(
    listView: ListView,
    ar: Set<String>?,
    set: Set<String>,
    label: (String) -> String = { it },
)
```

Then `initList` keeps calling `setAdapter` four times, with `::getMediumDisplayName` passed on the medium line only.

**Nits:** `import java.util.Locale` is appended below the `org.ole.…` imports — `java.*` sorts to the top. And `storage_other` reads "Other Files", which is right on the storage-breakdown screen it was written for but reads oddly as a medium filter row; a filter-specific string would say "Other".

**Verdict:** needs another round — the mechanism is correct and the value/label separation is properly handled, but localize the two hardcoded labels and fold the duplicate adapter back into `setAdapter`.

---

dogi:
**Re-review — rating: 88/100** (was 74)

Both blocking points are resolved, and the string choices came out better than what I suggested.

**The two hardcoded labels are localized properly.** `medium_text_html` and `medium_html` are added to `values/` and to all five locale files with real translations — `نص / HTML`, `Texto / HTML`, `Texte / HTML`, `पाठ / HTML`, `Qoraal / HTML` — with `HTML` correctly left as-is everywhere, since it's an acronym rather than a word. That's the whole of the previous objection gone.

**The duplicate adapter is gone, folded back into `setAdapter` exactly as proposed.** Adding `label: (String) -> String = { it }` means one code path serves all four lists, and only the medium call passes `::getMediumDisplayName`. The default is a genuine identity: for the other three lists `getView` sets `textView.text = label(item)` = `item`, which is what the plain `ArrayAdapter` was already doing via `textViewResourceId`. So levels, languages and subjects render identically to before.

You also fixed the nit I raised about wording — `filter_pdfs` / `filter_videos` / `filter_audio` / `other` instead of borrowing the storage-screen strings, so "Other Files" no longer shows up as a medium filter row.

The correctness property from the first round still holds: the adapter's backing list is still the raw `arr`, so `onItemClick`'s `getItemAtPosition(i)` returns the unmapped key and the filter still stores `"video"`, not `"Videos"`. Using `getItem(position)` with null guards on both item and view is a little safer than indexing the list directly.

**Nits:** `image` still maps to `storage_images` while its five neighbours now use filter-specific strings — worth a `filter_images` for consistency. The parameter list ends `label: (String) -> String = { it }, )` with a trailing comma and space before the paren. And `import java.util.Locale` is still below the `org.ole.…` imports, where `java` sorts above `org`.

**Verdict:** mergeable.

---

Okuro3499 (approved):

dogi (approved):

PR #16731, opened 2026-09-04
life: smoother achievements cv editing (fixes #16719)

### LINKED ISSUE #16719: EditAchievementFragment copies the CV file on the main thread
## What

Saving an achievement copies a user-picked CV file on the main dispatcher, producing StrictMode disk-read/write violations on every save.

**File:** `app/src/main/java/org/ole/planet/myplanet/ui/user/EditAchievementFragment.kt` (`computeCvFilename`, and its call site at line 178)

## Why

The save handler runs:

```kotlin
lifecycleScope.launch {          // main dispatcher
    val cvFilename = computeCvFilename()
    ...
}
```

`computeCvFilename()` (line 464) is a plain non-suspend function that does `contentResolver.openInputStream(uri)` and then `destFile.outputStream().use { output -> input.copyTo(output) }` (line 474) — copying a PDF of arbitrary size synchronously on the UI thread.

`lifecycleScope.launch` without a dispatcher runs on `Dispatchers.Main`, so nothing moves this off the UI thread. Debug builds run StrictMode `detectAll`, so this logs a violation on every CV save, and a large file is a visible jank/ANR risk on low-end devices — which is most of the deployment target.

This is the same class of defect as #16687, #16689 and #16691.

## The fix

1. Make `computeCvFilename` a `suspend` function.
2. Wrap the uri-check and file-copy body in `withContext(dispatcherProvider.io)`.
3. Use the `dispatcherProvider` already injected by `BaseContainerFragment` — do **not** hardcode `Dispatchers.IO`; the project injects `DispatcherProvider` so tests can substitute deterministic dispatchers.
4. Leave the `deleteCv` early return and the `resumeFileName` fallback on the calling dispatcher — only the I/O moves.

## Acceptance

- No StrictMode disk violation when saving an achievement with a CV attached.
- `deleteCv` and `resumeFileName` fallback behavior unchanged.
- No `Dispatchers.IO` literal introduced.

## Notes

- Conventions: `docs/CODE_STYLE_GUIDE.md` (coroutines / dispatcher discipline); test patterns: `docs/TESTING.md`.
- Please keep the change scoped to this one function and its call site.

dogi:
Offloaded CV file copy in EditAchievementFragment.computeCvFilename to dispatcherProvider.io via suspend function withContext.

Fixes #16719

---
*PR created automatically by Jules for task [13943985313128686241](https://jules.google.com/task/13943985313128686241) started by @dogi*

## Summary by CodeRabbit

* **Performance**
  * Improved CV file processing by moving file operations off the main thread, helping keep the app responsive during achievement editing.


dogi:
**Review — rating: 86/100**

Correct fix for a real main-thread stall: `computeCvFilename` opens a `contentResolver` stream and does a full `copyTo` of a user-selected CV, which on the main thread blocks the UI for as long as the file takes to write.

I checked the three things that usually break this kind of change, and all three hold. `requireContext()` is resolved *before* `withContext`, so the fragment isn't touched from the IO dispatcher. `dispatcherProvider` isn't declared here but is inherited from `BaseContainerFragment:57`, and the surrounding class already uses that same injected provider rather than a hard-coded `Dispatchers.IO`, which is the convention in `docs/CODE_STYLE_GUIDE.md`. And the only caller, line 178, is already inside `lifecycleScope.launch { }`, so promoting the function to `suspend` needs no change at the call site.

The `Utilities.toast` in the catch is also safe — it checks `Looper.myLooper() == Looper.getMainLooper()` and posts to `mainHandler` otherwise (`Utilities.kt:41-50`), so toasting from the IO dispatcher works.

**Nit:** the function still `return filename` after a failed copy, so a save error toasts but the payload is submitted claiming a CV file that was never written. That's pre-existing rather than introduced here, but this change puts the failure on a background thread where it's even less likely to be noticed.

**Verdict:** mergeable.

---

Okuro3499 (approved):

dogi:
**Promoted to `merge`** — Okuro3499 approved `565e35a9`, my review stands at 86/100, no changes since.

One thing worth carrying forward rather than blocking on: CodeRabbit flagged that `File(destDir, filename)` builds a destination from a provider-supplied display name, so a malicious content provider returning a name with path separators could write outside the CV directory. That's pre-existing — this PR relocated the block onto the IO dispatcher rather than introducing it — but it is in the lines this PR moved, so it's worth a follow-up issue rather than being lost in a resolved thread.

---

PR #16732, opened 2026-09-04
all: smoother notifications repository querying (fixes #16720)

### LINKED ISSUE #16720: Stop counting team chat messages for teams that have no chat notification row
## What

`getTeamNotifications` runs one `SELECT COUNT(*)` per team, but discards the result for every team that has no chat notification row.

**File:** `app/src/main/java/org/ole/planet/myplanet/repository/NotificationsRepositoryImpl.kt` (`getTeamNotifications`, lines 309-312)

## Why

The method already builds `notificationsById` from `teamNotificationDao.getByTypeAndParentIds("chat", teamIds)` — i.e. only the teams that actually have a chat notification row. It then counts **every** team anyway:

```kotlin
val chatCountsById = mutableMapOf<String, Long>()
for (teamId in teamIds) {
}
```

But the count is only ever consumed as:

```kotlin
val notification = notificationsById[teamId]
val chatCount = chatCountsById[teamId] ?: 0L
```

When `notification == null`, `hasChat` is `false` regardless of the count — so the query for that team was pure waste. This runs on the dashboard team strip: a user in 20 teams with 2 chat-tracked teams issues 20 queries where 2 suffice.

## The fix

Build `notificationsById` first (already done), then loop only over `notificationsById.keys` when filling `chatCountsById`.

- Keep `chatCountsById` typed `Map<String, Long>`.
- Keep the `?: 0L` fallback in the consuming loop, so teams with no row still report `hasChat = false` through exactly the same path.
- The final `for (teamId in teamIds)` loop that builds `notificationMap` stays as-is — every team still gets an entry.

This cannot change behavior; it only stops issuing queries whose results are discarded.

## Acceptance

- A test asserting with MockK `coVerify(exactly = 1) { voicesRepository.countTopLevelByTeam(chatTrackedTeamId) }` and `coVerify(exactly = 0)` for a team with no chat row.
- Existing `hasChat` / `hasTask` semantics unchanged for all teams, including teams with no notification row.

## Notes

- Conventions: `docs/CODE_STYLE_GUIDE.md`; test patterns: `docs/TESTING.md` (MockK, `runTest`, `TestDispatcherProvider`).
- Please do **not** replace `countTopLevelByTeam` with a batched `GROUP BY` DAO query in this issue — that variant would drop the `viewIn LIKE :teamPattern` legacy branch of `countTopLevelByTeam`, which has not been proven dead. Keep this change to the loop bound.

dogi:
Optimized getTeamNotifications in NotificationsRepositoryImpl to iterate over notificationsById.keys instead of all teamIds when populating chatCountsById. This prevents issuing unnecessary voicesRepository.countTopLevelByTeam Room queries for teams that do not have a chat notification row. Unit tests were added in NotificationsRepositoryImplTest to verify query behavior and verify that hasChat semantics remain identical across all teams.

Fixes #16720

---
*PR created automatically by Jules for task [14791610125174811475](https://jules.google.com/task/14791610125174811475) started by @dogi*


dogi:
**Review — rating: 90/100**

One-line change that removes a genuine N-query fan-out, and I verified it's behaviour-preserving rather than just cheaper.

`chatCountsById` was populated by looping over every `teamId`, calling `voicesRepository.countTopLevelByTeam(teamId)` once per team. But the value is only ever consumed as `hasChat = notification != null && notification.lastCount < chatCount` — so for any team with no `"chat"` row in `notificationsById`, `hasChat` is `false` regardless of what the count came back as. Narrowing the loop to `notificationsById.keys` therefore drops only queries whose results could not affect the outcome. `notificationsById` is built from `getByTypeAndParentIds("chat", teamIds)`, so its keys are a subset of `teamIds` and no new team can appear; and the result map is still assembled by iterating the full `teamIds`, so every requested team still gets an entry. On a dashboard for a user in many teams where few have chat tracking, that's most of the queries gone.

The test is the right shape too: it asserts `coVerify(exactly = 1)` on the tracked team and `exactly = 0` on the untracked one — pinning the *absence* of the query, not just the returned values — while still checking the untracked team is present in the result with `hasChat == false`. That's what stops a later refactor from quietly restoring the fan-out.

**Verdict:** mergeable.

---

Okuro3499 (approved):

dogi:
**Promoted to `merge`** — Okuro3499 approved `4c9053a1`, my review stands at 90/100, no changes since.

---

PR #16733, opened 2026-09-04
dashboard: smoother plugin life route mapping (fixes #16718)

### LINKED ISSUE #16718: Route My Life dashboard clicks off the stable imageId, not the localized title
## What

`DashboardPluginFragment.handleClickMyLife` dispatches on the **localized** item title, so every My Life shortcut is dead on any non-English locale.

**File:** `app/src/main/java/org/ole/planet/myplanet/ui/dashboard/DashboardPluginFragment.kt` (`handleClickMyLife`, `getLayout`)

## Why

`handleClickMyLife(title: String, v: View)` (line 70) matches `when (title)` against hardcoded English literals:

```kotlin
"mySubmissions" -> openIfLoggedIn { listener.openCallFragment(SubmissionsFragment()) }
"myHealth"      -> openIfLoggedIn { listener.openCallFragment(MyHealthFragment()) }
// ... "Calendar", "mySurveys", "myAchievements", "myPersonals", "References"
```

The `title` it receives is `DashboardItem.title` — the value from `MyLife.defaultItems`, which is a resolved string resource. `R.string.submission` is `"mySubmissions"` in `values/strings.xml` but `"misEnvios"` in `values-es`, `"mesSoumissions"` in `values-fr`, and likewise in `values-ar`, `values-ne`, `values-so`.

So on Spanish, French, Arabic, Nepali or Somali **every** My Life shortcut falls through to `else -> toast(feature_not_available)`. The user taps My Submissions and gets "feature not available".

A second consequence: rows seeded under one locale keep that locale's title in the database, so the title is unusable as a routing key even within a single install that changed language.

## The fix

The stable key is already right there — line 106 does `imageResourceMap[obj.imageId]`, so `imageId` reaches the same call site and is locale-independent.

1. Add an internal pure function `myLifeRouteFor(imageId: String?)` mapping the stable ids already used by `imageResourceMap` — `ic_submissions`, `ic_references`, `ic_calendar`, `ic_my_survey`, `my_achievement`, `ic_mypersonals`, `ic_myhealth` — to route descriptors.
2. Pass `obj.imageId` into `handleClickMyLife` instead of `obj.title`.
3. Keep the existing per-route `openIfLoggedIn` gating exactly as it is, and keep the `feature_not_available` toast as the fallback for an unknown or blank id.

Keeping the routing decision in a pure function (rather than inline in the `when`) is what makes it unit-testable without inflating a Fragment.

## Acceptance

- Every My Life shortcut opens its fragment regardless of device language.
- Unit tests covering `myLifeRouteFor`: each of the seven known ids maps to its expected route, plus `null`, `""` and an unknown id all map to the unknown/fallback case.
- No behavior change on English.

## Notes

- Conventions: `docs/CODE_STYLE_GUIDE.md`; test patterns: `docs/TESTING.md` (JUnit4 + MockK, not Mockito).
- Please keep the change scoped to this routing bug — no unrelated refactoring of the dashboard.

dogi:
Route My Life dashboard clicks off the stable imageId rather than the localized item title.

Key changes:
- Created `MyLifeRoute` enum and internal pure function `myLifeRouteFor(imageId: String?)` in `DashboardPluginFragment.kt` to map stable image IDs to destination route descriptors.
- Updated `handleClickMyLife` to evaluate `imageId` instead of localized string `title`.
- Updated `getLayout` to pass `obj.imageId` to `handleClickMyLife`.
- Added unit tests in `DashboardPluginFragmentTest` covering each known `imageId` mapping as well as `null`, `""`, and unknown image IDs.

Fixes #16718

---
*PR created automatically by Jules for task [9286510578557678726](https://jules.google.com/task/9286510578557678726) started by @dogi*


dogi:
**Review — rating: 92/100**

This is a real user-facing bug, and I confirmed it end to end rather than taking the title's word for it.

`handleClickMyLife` branched on `title`, which comes from `MyLife.defaultItems(userId, resolveLabel)` — i.e. `resolveLabel(stringRes)`, a **localized** string. The old branches were `"mySubmissions"`, `"References"`, `"Calendar"`, `"mySurveys"`, `"myAchievements"`, `"myPersonals"`, `"myHealth"`, and the English `values/strings.xml` happens to define exactly those literals (`submission` = `mySubmissions`, `mypersonals` = `myPersonals`, …), so English matched by coincidence of wording. `values-es` gives `misEnvíos` and `misPersonales`, which match nothing — so every My Life tile fell to `else` and toasted "feature not available" in all five non-English locales. Routing on `imageId` fixes it at the source.

The mapping is complete and correct: the seven keys in `myLifeRouteFor` are exactly the seven in `MyLife.defaultItemPairs` (`ic_myhealth`, `my_achievement`, `ic_submissions`, `ic_my_survey`, `ic_references`, `ic_calendar`, `ic_mypersonals`), with no key invented and none missed. It also brings this fragment in line with `LifeAdapter.fragmentCache` (`LifeAdapter.kt:178-186`), which already keyed the same routing off `imageId` — so the fix is the pattern the codebase had already settled on in the sibling screen, not a new invention.

Extracting `myLifeRouteFor` as an `internal` pure function is what makes it testable without Robolectric, and the tests cover all seven keys plus null, empty and unknown. Dropping the `if (title != null)` guard is right too: routing no longer depends on the title, and an item with a null title now gets a listener that toasts rather than being silently inert.

**Nit:** there are now three places encoding imageId knowledge — `MyLife.defaultItemPairs`, `LifeAdapter.fragmentCache`, and `myLifeRouteFor`. A future tile added to one and not the others fails quietly. Worth folding the destination mapping into one shared source eventually.

**Verdict:** mergeable.

---

Okuro3499 (approved):

dogi:
**Promoted to `merge`** — Okuro3499 approved `40c3410c`, my review stands at 92/100, no changes since.

---

PR #16734, opened 2026-09-04
courses: smoother resources inline disk checking (fixes #16722)

### LINKED ISSUE #16722: InlineResourceAdapter stats the filesystem on the main thread during bind
## What

`updateStatusAndPreview` does two disk stats per row per bind, synchronously on the UI thread — while every sibling path in the same class already moves that work to IO.

**File:** `app/src/main/java/org/ole/planet/myplanet/ui/courses/InlineResourceAdapter.kt` (`updateStatusAndPreview`, line 162)

## Why

`updateStatusAndPreview` is called straight from `onBindViewHolder` (line 155, and again from the payload path at 136/138) and evaluates:

```kotlin
FileUtils.checkFileExist(context, UrlUtils.getUrl(resource))   // line 165
```

`checkFileExist` does `f.exists() && f.length() > 0` — two disk stats, on the main thread, for every row as it scrolls past.

The same class already knows better. Lines 206, 221, 232 and 306 all wrap their stats in `withContext(dispatcherProvider.io) { ... }`, and there is an established `adapterScope` (line 59) plus a per-holder `setPreviewJob` / `cancelPreviousPreviews` pair (lines 72-78) for exactly this. Only this one path skipped it.

## The fix

1. Set the row to its **not-downloaded** state synchronously first, so the row is never left showing a stale state from the recycled holder.
2. Move the `resource.isResourceOffline() || checkFileExist(...)` evaluation inside the existing `holder.setPreviewJob(adapterScope.launch { ... })` block (line 185), with the stat wrapped in `withContext(dispatcherProvider.io)`.
3. Short-circuit on `isResourceOffline()` **first** — it's a pure in-memory check, so an offline resource should never reach the filesystem at all.
4. Leave `cancelPreviousPreviews()` where it is (lines 117, 153) so recycled rows still cancel stale work.

## Acceptance

- No filesystem access on the main thread during bind.
- A recycled row never briefly displays the previous row's download state.
- Scrolling a list of already-downloaded resources still shows the downloaded state.

## Notes

- Use the injected `dispatcherProvider`, never a hardcoded `Dispatchers.IO`.
- Keep the change to this one method plus its call sites — no broader adapter refactor.

dogi:
This change offloads disk file existence checks in InlineResourceAdapter.updateStatusAndPreview from the UI main thread to dispatcherProvider.io. It synchronously resets the row state to prevent recycled view holders from displaying stale download states, short-circuits in-memory isResourceOffline checks, and adds unit tests covering these behaviors.

Fixes #16722

---
*PR created automatically by Jules for task [8076471464428843584](https://jules.google.com/task/8076471464428843584) started by @dogi*


dogi:
**Review — rating: 88/100**

`FileUtils.checkFileExist` was running synchronously inside `onBindViewHolder`, so every row bound during a scroll hit the disk on the main thread. Moving it behind `withContext(dispatcherProvider.io)` is the right fix, and the recycling hazard it introduces is already handled.

The thing that makes this safe is `holder.setPreviewJob(...)`, which cancels the holder's previous job — the existing test in this file already pins that (`assertEquals(true, job?.isCancelled)`). Since the whole status-and-preview body is now inside the job, a holder rebound to a different resource cancels the in-flight check before it can write to the recycled binding. Widening the job's scope from just the preview to the status check as well is what keeps that guarantee intact.

The short-circuit is preserved correctly: `resource.isResourceOffline() || withContext(io) { … }` means an already-offline resource never touches `UrlUtils.getUrl` or the filesystem, and the first test asserts exactly that with `coVerify(exactly = 0)`. The second test is the better one — it advances the main and IO schedulers separately to prove the disk call happens on IO and *not* on main. That asserts the dispatcher placement rather than just the end state, which is the property this PR is actually about and the one a careless refactor would silently undo.

**Nit:** `pbDownload` is now set `VISIBLE` unconditionally before the job runs, so an already-downloaded resource flashes a progress bar for one frame on every bind. `isResourceOffline()` is a plain field read with no IO, so hoisting just that check above the launch — spinner only when the disk check is genuinely needed — would remove the flicker without giving back any main-thread work.

**Verdict:** mergeable.

---

Okuro3499 (approved):

dogi:
**Promoted to `merge`** — Okuro3499 approved `d4d08782`, my review stands at 88/100, no changes since.

---

PR #16735, opened 2026-09-04
all: smoother utilities text normalizing (fixes #16727)

### LINKED ISSUE #16727: Add an ASCII fast path to Utilities.normalizeText
## What

`normalizeText` runs Unicode NFD decomposition and a regex strip on every call, including for pure-ASCII input where both are guaranteed no-ops.

**File:** `app/src/main/java/org/ole/planet/myplanet/utils/Utilities.kt` (`normalizeText`, line 80)

## Why

```kotlin
fun normalizeText(str: String): String {
    return Normalizer.normalize(str.lowercase(Locale.getDefault()), Normalizer.Form.NFD)
        .replace(DIACRITICS_REGEX, "")
}
```

`DIACRITICS_REGEX` (line 20) is `Regex("\\p{InCombiningDiacriticalMarks}+")`, and `String.replace(Regex, …)` allocates a fresh `Matcher` on every call.

This is the per-item primitive of offline search and course filtering — every candidate title and every query token passes through it on every keystroke (call sites in `MyLibrary`, `SurveysViewModel`, `CoursesRepositoryImpl` ×4, `ResourcesRepositoryImpl`). For a pure-ASCII title, NFD decomposition produces the input unchanged and there are no combining marks to strip, so both steps are pure overhead.

## The fix

Lowercase once, then return immediately when every char is `< 0x80`. Otherwise fall through to the existing path unchanged.

## Acceptance

Tests must cover both paths:

- **Fast path:** mixed case, empty string, digits and punctuation, plain ASCII words.
- **Slow path (round-trips):** `Café`, `Niño`, `áéíóú`, plus Arabic and Nepali strings — these must produce byte-identical output to the current implementation.

The slow-path cases are the ones that matter: the fast path is only correct if it is genuinely unreachable for any input the slow path would have altered.

## Notes

- Do not change the lowercasing locale — `Locale.getDefault()` stays.

dogi:
Added an ASCII fast path to Utilities.normalizeText that lowercases input once and returns immediately if all characters are ASCII (< 0x80), bypassing NFD decomposition and diacritics stripping. Non-ASCII inputs fall through to the slow path. Added comprehensive unit tests in UtilitiesTest to verify both fast-path and slow-path behavior.

Fixes #16727

---
*PR created automatically by Jules for task [16171592883112579409](https://jules.google.com/task/16171592883112579409) started by @dogi*


dogi:
**Review — rating: 88/100**

The fast path is exactly equivalent, and the reason it's safe is worth stating precisely: ASCII (U+0000–U+007F) is NFD-stable, so `Normalizer.normalize` is the identity on it, and `\p{InCombiningDiacriticalMarks}` covers U+0300–U+036F, none of which is ASCII — so the regex can never delete anything from a pure-ASCII string either. Returning `lower` unchanged is therefore byte-identical to running both steps, and the win is skipping the regex allocation on the overwhelmingly common case.

The locale trap is also handled, by accident of ordering but correctly: the guard runs on `lower`, *after* `lowercase(Locale.getDefault())`. Under a Turkish locale `"I"` lowercases to dotless `ı` (U+0131) and `"İ"` to `i` + combining dot above (U+0307) — both non-ASCII, so both take the slow path and normalize as before. Testing the post-lowercase string rather than the input is what makes that work; a guard on `str` would have been wrong.

The equivalence test is the right way to prove this: rather than asserting hand-written expectations, it reimplements the original as `legacyNormalizeText` and asserts the new output matches it for Latin-accented, Arabic and Devanagari inputs, with a byte-array comparison on top. Freezing a copy of the old implementation is exactly what you want for a "this refactor changes nothing" claim.

**Nit:** non-ASCII inputs now pay an extra full scan before doing the same work as before — a small regression on the slow path in exchange for a large win on the fast one, which is the right trade for search text but worth knowing. The scan short-circuits on the first non-ASCII character, so the cost is only real for long strings whose non-ASCII content is near the end.

**Verdict:** mergeable.

---

Okuro3499 (approved):

dogi:
**Promoted to `merge`** — Okuro3499 approved `0bdb8422`, my review stands at 88/100, no changes since.

---

PR #16739, opened 2026-09-04
enterprises: smoother repository reports csv exporting (fixes #16729)

### LINKED ISSUE #16729: Push the archived-report filter and sort of the finances CSV export into SQL
## What

`exportReportsAsCsv` loads every report row for a team, then filters and sorts in Kotlin — using a predicate that already exists in SQL one line above it in the same DAO.

**Files:**
- `app/src/main/java/org/ole/planet/myplanet/data/room/dao/TeamDao.kt` (one new `@Query`)
- `app/src/main/java/org/ole/planet/myplanet/repository/EnterprisesRepositoryImpl.kt` (`exportReportsAsCsv`, lines 102-105)

## Why

```kotlin
val reports = teamDao.getByTeamIdAndDocType(teamId, "report")
    .filter { it.status != "archived" }
    .sortedByDescending { it.createdDate }
```

`getByTeamIdAndDocType` (`TeamDao.kt:20`) is a plain `SELECT * FROM teams WHERE teamId = :teamId AND docType = :docType`. The exact predicate needed is already expressed in SQL three lines below it:

```kotlin
// TeamDao.kt:23
@Query("SELECT * FROM teams WHERE teamId = :teamId AND docType = 'report' AND IFNULL(status, '') != 'archived' ORDER BY createdDate DESC")
fun observeNonArchivedReportsByTeamId(teamId: String): Flow<List<MyTeam>>
```

## The fix

Add the suspend twin of that query and call it:

```kotlin
@Query("SELECT * FROM teams WHERE teamId = :teamId AND docType = 'report' AND IFNULL(status, '') != 'archived' ORDER BY createdDate DESC")
suspend fun getNonArchivedReportsByTeamId(teamId: String): List<MyTeam>
```

Then drop the `.filter { }` and `.sortedByDescending { }` from `exportReportsAsCsv`.

Adding a `@Query` changes no table, so **no `AppDatabase` version bump** is needed. (A version bump here would be actively harmful — this project uses `fallbackToDestructiveMigration`, so a bump drops the local DB and any unsynced writes with it.)

## Acceptance

- Exported CSV is byte-identical to today's output for the same data, including row order.
- Reports with a **null** status are still included.

## Notes

⚠️ The `IFNULL` is load-bearing. The current Kotlin predicate `it.status != "archived"` keeps null-status rows; a bare `status != 'archived'` in SQL would silently **drop** them, because `NULL != 'archived'` is `NULL`, not true. Copy the `IFNULL(status, '')` form from line 23 exactly.

- Do not convert the existing Kotlin chain to a `Sequence` instead — that buys nothing here; the point is to not load archived rows at all.

dogi:
Pushes archived-report filter and sort for finances CSV export into SQL in TeamDao.

Fixes #16729

---
*PR created automatically by Jules for task [10904712982724176407](https://jules.google.com/task/10904712982724176407) started by @dogi*


dogi:
**Review — rating: 88/100**

Straight push of a filter and sort into SQL, and the NULL trap that usually breaks this exact refactor is handled.

The Kotlin was `.filter { it.status != "archived" }`, which **keeps** rows where `status` is null, because `null != "archived"` is true in Kotlin. The naive SQL translation `status != 'archived'` would have silently dropped every null-status report, since `NULL != 'archived'` is NULL, not true. The query uses `IFNULL(status, '') != 'archived'`, which maps null to `''` and keeps the row — matching the Kotlin exactly. That's the single most important thing in this diff and it's right.

The predicate is also identical, character for character, to the existing `observeNonArchivedReportsByTeamId` right above it, so the suspend and Flow variants can't disagree about what "non-archived" means today.

**Nit:** `sortedByDescending` is a stable sort, `ORDER BY createdDate DESC` is not, so reports sharing a `createdDate` may come out in a different order than before — invisible in most data, but it is a real difference for a CSV export. Adding a tiebreaker (`, _id DESC` or similar) would make the export byte-reproducible. Also, the two queries being separate string literals means they can still drift apart on the next edit; a shared constant would tie them together.

**Verdict:** mergeable.

---

Okuro3499 (approved):

dogi:
**Promoted to `merge`** — Okuro3499 approved `a5175906`, my review stands at 88/100, no changes since.

---

PR #16740, opened 2026-09-04
resources: smoother activities repository dao counting (fixes #16730)

### LINKED ISSUE #16730: Aggregate the most-opened resource in SQL instead of loading the whole activity table
## What

`getMostOpenedResource` loads every matching activity row into memory to pick a single winner.

**Files:**
- `app/src/main/java/org/ole/planet/myplanet/data/room/dao/ResourceActivityDao.kt` (one new `@Query`)
- `app/src/main/java/org/ole/planet/myplanet/repository/ActivitiesRepositoryImpl.kt` (`getMostOpenedResource`, line 173)

## Why

```kotlin
val activities = resourceActivityDao.getByUserAndType(userName, type)   // plain SELECT *
...
val resourceCounts = activities
    .groupBy { it.resourceId }
    .mapValues { entry -> ... }
```

`getByUserAndType` (`ResourceActivityDao.kt:19`) is a plain `SELECT *`, and the grouping runs in memory on the default dispatcher. A user with a large offline history pays O(rows) — allocation and CPU — on the profile screen to produce one `Pair<String, Int>`. The DAO already has a `countByUserAndType` (line 22) but no aggregate.

## The fix

Add a `@Query` that groups by `resourceId` with `COUNT(*)`, ordered by count descending with a **documented, stable tie-break**, `LIMIT 1`, skipping null titles. Map the result to the existing `Pair<String, Int>?` contract — empty result → `null` — and delete the in-memory grouping.

No schema version bump: adding a `@Query` changes no table. (A bump would be harmful here — this project uses `fallbackToDestructiveMigration`, so it drops the local DB and any unsynced writes.)

## Acceptance

- Same winner and same count as the current implementation for the same data.
- Empty history still returns `null`.
- The tie-break is deterministic and stated in a comment — today's `maxByOrNull` picks the first in iteration order, which is not a guarantee worth preserving accidentally. Pick an explicit secondary sort and document it.
- Rows with a null title are excluded, matching current behavior.

## Notes

- Please include a Room in-memory DAO test covering: single winner, a tie, an empty table, and a null-title row.

dogi:
Optimizes `getMostOpenedResource` by performing resource activity aggregation directly in SQL using Room instead of loading all matching `ResourceActivity` rows into memory. Adds `ResourceOpenCount` projection, updates `ActivitiesRepositoryImpl`, and adds in-memory Room DAO unit tests covering single winner, tie breaking, empty table, and null/blank title filtering.

Fixes #16730

---
*PR created automatically by Jules for task [16066993509018333271](https://jules.google.com/task/16066993509018333271) started by @dogi*


dogi:
**Review — rating: 84/100**

Replacing a load-everything-and-group-in-Kotlin with one aggregate query is clearly right here — the old path pulled every `resource_activity` row for a user into memory just to compute a single winner. The new version also fixes an order-dependency bug in passing, and I want to be explicit that the behaviour change is real and, I think, deliberate.

**The old code dropped resources based on an arbitrary row.** It did `title = entry.value.first().title` and then `filterValues { it.second != null }` — so if the *first* row in a group happened to have a null title, the entire resource was discarded even when every other row was titled. Which row landed first was whatever order the DAO returned. The new `WHERE title IS NOT NULL AND TRIM(title) != ''` filters per row instead, so a resource survives on its titled rows. That's strictly more robust.

**But it also changes the count.** The filter runs before `GROUP BY`, so `COUNT(*)` now counts only titled rows, where the old `entry.value.size` counted every row in the group including untitled ones. For a resource opened 10 times with 3 untitled rows, the profile now says 7 instead of 10. `getMostOpenedResource_excludesNullAndBlankTitles` shows this is intended rather than accidental, and excluding blank titles is defensible — you can't display an untitled resource. Worth saying so in the PR description, since it's a user-visible number.

The deterministic `ORDER BY openCount DESC, title ASC` is an improvement over `maxByOrNull`, which returned whichever max the map iteration reached first.

**Nit:** the KDoc says ties are "deterministically broken using title ASC", but with a bare `title` under `GROUP BY resourceId`, SQLite picks the title from an arbitrary row in each group — so if a resource was ever renamed, which title represents the group is still unspecified, and that's the value the tiebreak then sorts on. Consistent-per-resource in practice, but the guarantee is weaker than the comment implies. Also, no test covers a single `resourceId` with a *mix* of titled and untitled rows, which is exactly where the new count diverges from the old. And `withContext(dispatcherProvider.io)` around a Room suspend DAO is redundant — Room already moves the query off the calling thread (`CLAUDE.md`, Room conventions).

**Heads-up:** this conflicts with #16826, which inserts new methods directly above `getMostOpenedResource` in the same file. The two are complementary rather than competing — that one changes the layering and leaves this body alone — so both can land; whichever goes second needs a small rebase.

**Verdict:** mergeable.

---

Okuro3499 (approved):

dogi:
**Promoted to `merge`** — Okuro3499 approved `7fef03ff`, my review stands at 84/100, no changes since.

Restating the one thing from that review that's user-visible, since it lands with the merge: the `WHERE title IS NOT NULL AND TRIM(title) != ''` filter runs before `GROUP BY`, so `COUNT(*)` now counts only titled rows. A resource opened 10 times with 3 untitled rows reports 7 rather than 10. It's a defensible improvement (and it fixes the old order-dependent bug where one untitled first row discarded the whole resource), but the profile number changes.

---

PR #16741, opened 2026-09-04
sync: smoother photo batch uploading (fixes #16725)

### LINKED ISSUE #16725: Parallelize the PhotoUploader batch POSTs
## What

`uploadSubmitPhotos` chunks photos into batches and then uploads each batch one photo at a time, so wall-clock time scales linearly with photo count.

**File:** `app/src/main/java/org/ole/planet/myplanet/services/upload/PhotoUploader.kt` (`uploadSubmitPhotos`, line 25)

## Why

```kotlin
photosToUpload.chunked(BATCH_SIZE).forEach { batch ->      // 39
    batch.forEach { (photoId, serialized) ->               // 42  <-- sequential
        val `object` = uploadRepository.postUpload(...)    // 44
    }
    submissionsRepository.markPhotosUploadedBatch(batchedMarks)   // 63
    successfulUploads.forEach { uploadInfo ->              // 70  <-- also sequential
        uploadAttachment(photo.photoLocation, ...)         // 73
    }
}
```

Both loops are serial network round trips. On the slow links this app is built for, a user submitting a dozen photos waits for twelve sequential POSTs plus twelve sequential attachment uploads. The chunking already exists; only the concurrency is missing.

## The fix

Within each batch, run the POSTs with bounded concurrency using the pattern `UploadCoordinator` already proves out — `Semaphore` + `async` + `withPermit` (see `UploadCoordinator.kt` lines 33, 128, 143), with a cap of **6 or fewer**.

- Collect successes into a thread-safe structure (the current `successfulUploads` accumulation is not safe under concurrency).
- Keep the single `markPhotosUploadedBatch(batchedMarks)` call after the batch completes — one call, not one per photo.
- **Preserve failure isolation**: one photo failing must not cancel the rest of the batch.
- Preserve the existing listener callbacks and their ordering guarantees.

## Acceptance

- A batch issues at most 6 concurrent POSTs.
- One failing photo leaves the other photos in its batch uploaded and marked.
- `markPhotosUploadedBatch` is still called exactly once per batch.
- Listener callbacks still fire for every photo.

## Notes

- Reuse the `UploadCoordinator` `Semaphore` pattern; don't introduce a different concurrency primitive.
- Whether the `uploadAttachment` loop (line 70) should also be parallelized is a judgement call — if it shares a rate limit or ordering constraint with the POSTs, leave it sequential and say so in the PR description.

dogi:
Parallelizes photo uploads in `PhotoUploader.kt` by processing POST requests in each batch concurrently with bounded concurrency (`Semaphore(6)`).

Key details:
- Bounded concurrency using `Semaphore(6)` and `coroutineScope` / `async` / `awaitAll()`.
- Thread-safe collection of successful uploads via `awaitAll().filterNotNull()`.
- Single `markPhotosUploadedBatch` call per batch after concurrent POSTs finish.
- Failure isolation: per-photo try-catch rethrows `CancellationException` and catches other exceptions so one failing photo doesn't disrupt the batch.
- Attachment uploads remain sequential after batch marking because attachment uploads share network rate limits and launch background scope jobs internally.
- Added `PhotoUploaderTest` covering bounded concurrency, batch marking, failure isolation, and attachment listener invocation.

Fixes #16725

---
*PR created automatically by Jules for task [2811961381201297927](https://jules.google.com/task/2811961381201297927) started by @dogi*


dogi:
**Review — rating: 85/100**

The concurrency is done the safe way. The old loop accumulated into a `mutableListOf` — parallelising that as-is would have been a data race on every append. Instead each `async` *returns* its `UploadedPhotoInfo?` and the list is rebuilt with `awaitAll().filterNotNull()`, so there's no shared mutable state at all. `Semaphore(MAX_CONCURRENT_UPLOADS)` bounds in-flight POSTs, and because it's created outside the `chunked().forEach` while batches still run sequentially, the DB write between batches stays ordered — the parallelism is strictly within a batch. `CancellationException` is rethrown ahead of the generic handler, matching the taxonomy the other PRs in this queue have adopted.

Two behaviour changes ride along that aren't mentioned in the description, both defensible:

- **`response.isSuccessful` is now required.** Previously only `body() != null` gated recording a success. For Retrofit a non-2xx normally yields a null body, so this is mostly belt-and-braces, but it does close the case where a server returns 2xx-shaped content on an error path.
- **A null `photoId` now skips the upload entirely.** Before, the POST was still sent and only the bookkeeping was skipped, so the photo reached the server untracked. Not uploading something you can't record is the better behaviour, but it does mean fewer requests than before for that case.

**Nit:** OkHttp's dispatcher caps concurrent requests per host at 5 by default, so a permit count of 6 will in practice be throttled to 5 by the client anyway — the semaphore documents intent more than it constrains. Worth also considering whether 6 is right for this app's target conditions: on the weak links myPlanet is built for, several simultaneous photo POSTs can contend badly enough to be slower than sequential and raise timeout rates. A smaller number, or making it adaptive, may serve better than the OkHttp ceiling.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16742, opened 2026-09-04
sync: smoother upload pipeline batch coordinating (fixes #16724)

### LINKED ISSUE #16724: Serialize upload payloads per batch instead of all up front
## What

`UploadCoordinator.runPipeline` serializes **every** pending item into a `JsonObject` before the first request leaves the device.

**File:** `app/src/main/java/org/ole/planet/myplanet/services/upload/UploadCoordinator.kt` (`runPipeline`, `queryItemsToUpload`, `queueRetryableFailures`)

## Why

```kotlin
val itemsToUpload = queryItemsToUpload(config)                          // 44  <-- serializes everything
...
itemsToUpload.chunked(config.batchSize).forEachIndexed { batchIndex, batch ->   // 55
    ...
}
queueRetryableFailures(config, allFailed, itemsToUpload)                // 73
```

`queryItemsToUpload` (93) calls `config.fetchPendingItems()` and serializes each surviving item up front. On a device that has been offline for weeks that means thousands of live `JsonObject` trees held simultaneously — and the peak-memory spike lands on exactly the low-end hardware this app targets. It also delays first-byte: nothing is sent until the last item is serialized.

## The fix

1. Call `config.fetchPendingItems()` once and early-return `Empty` on an empty result.
2. Chunk the **raw** items, and do the serialization at the top of each loop-body iteration instead.
3. Keep the `shouldFilter` check (line 99) and the serialization-failure `mapNotNull` behavior **byte-identical** — same items dropped, same reasons.
4. Log the raw pending count before the loop so log output is unchanged.
5. Move `queueRetryableFailures` inside the loop (per batch), while still accumulating `allSucceeded` / `allFailed` for the final result.

## Acceptance

- Peak memory no longer scales with total pending count — only with `config.batchSize`.
- Same set of items uploaded, same items filtered, same log lines as before.
- **Required test:** batch 1 fails retryably and batch 2 succeeds — assert what ends up in the retry queue.

## Notes

⚠️ Step 5 is a **real semantic change**, not a refactor. Today failures are queued once after all batches; per-batch queueing means a later batch's outcome no longer affects when an earlier batch's failures are enqueued. That is the intended improvement, but it is the one part of this change that can alter behavior, so it needs the test above rather than an assertion that nothing changed.

- Do not change `config.batchSize` or the concurrency in the existing `Semaphore` path (lines 128-171) — this issue is only about *when* serialization happens.

dogi:
Updates UploadCoordinator.runPipeline to fetch pending items once up front, log total pending count, and defer item serialization to the top of each batch loop iteration. Per-batch retry queueing is now performed per batch rather than once at the end of all batches.

Fixes #16724

---
*PR created automatically by Jules for task [15994633193560544869](https://jules.google.com/task/15994633193560544869) started by @dogi*


dogi:
**Review — rating: 84/100**

Deferring serialization to the batch that's about to use it is the right shape: the old `queryItemsToUpload` materialised a `PreparedUpload` — including the serialized payload — for *every* pending item before a single POST went out, so peak memory scaled with the whole backlog rather than with `batchSize`. On a device that's been offline for a while, that's the allocation that hurts.

Moving `queueRetryableFailures` inside the loop is a real robustness gain, and I checked it's safe to call per batch: it builds `payloadMap` from the `preparedUploads` it's handed and looks up `error.itemId`, and a batch's errors can only reference that batch's items, so per-batch scoping loses no lookups. There's no call-once assumption in it. Previously the call sat *after* the loop, so a crash on batch 5 meant batches 1-4's retryable failures were never queued at all; now each batch's failures are durable before the next one starts.

Two consequences worth knowing:

- **Batches are now sized by raw count, not filtered count.** Chunking happens before `shouldFilter` and serialization, so a run where many items filter out produces more, smaller batches than before — where previously `batchSize` counted only items that would actually be POSTed. `uploadBatch` handles a fully-filtered batch fine (the `map` is empty and it returns two empty lists), so this is efficiency, not correctness.
- **`UploadResult.Empty` is now harder to reach.** The early return moved from the post-filter list to `pendingItems`, so "there were pending items but every one filtered out or failed to serialize" used to return `Empty` and now falls through to `Success(data = 0)`. `UploadManager` distinguishes the two at lines 118, 212 and 243 — mostly for log and message wording ("No submissions to upload") rather than control flow, so the impact is cosmetic, but the message a user sees in that case changes.

**Verdict:** mergeable.

---

Okuro3499 (dismissed):
```
File: app/src/test/java/org/ole/planet/myplanet/services/upload/UploadCoordinatorTest.kt:581-589 (legacyConfig)                                                                  
    Problem: modelClass = Submission::class as kotlin.reflect.KClass<SampleItem> is an unsafe cast added specifically so the new test (upload queues retryable failures per batch...,
    line 442) — which has a genuinely mixed batch (2 fail, 2 succeed) — doesn't hit UploadConfig.persistUploaded's else -> error("Unsupported upload update config: ...") for the    
    generic test-only SampleItem class. It works only because modelClass happens not to be used to cast actual item instances, but it masks the underlying design constraint that    
    UploadConfig can only ever be used in production with StepExam or Submission, and it makes the test assert on uploadType = "Submission" / modelClassName = "Submission" for what 
    is nominally a generic SampleItem scenario — a future reader will reasonably believe this validates a generic path when it's silently pinned to Submission.
```

dogi:
**Re-review — rating: 88/100** (was 84)

`UploadCoordinator.kt` is unchanged from the revision I rated 84 — the entire delta since is in the test file, and it's the delta the change needed.

**The new test is the right one.** `upload queues retryable failures per batch when batch 1 fails retryably and batch 2 succeeds` sets `batchSize = 2` over four items, fails the first pair with a 500 and succeeds the second, then asserts `queueFailedOperation` fires exactly once each for `local-1` and `local-2` with their own payloads, and exactly zero times for `local-3`/`local-4`. That pins precisely the semantics that moved: `queueRetryableFailures` used to run once after the loop over `allFailed` and every prepared item, and now runs per batch over that batch's failures and that batch's items. Without this test the move would have been a behaviour change with nothing watching it.

The `SampleItem` → `Submission` migration in the surrounding tests is what makes those payload assertions possible — the synthetic model's serializer couldn't produce a `JsonObject` the `match { it.get("localId") }` predicates could inspect. `SampleItem` survives for the `roomConfig` tests, so nothing is left dangling.

Re-confirming the two things that mattered on the production side, against the current head:

- Per-batch `queueRetryableFailures` is safe: it builds `payloadMap` by `localId` from the `preparedUploads` it's handed and looks up `error.itemId`, and a batch's errors can only name that batch's items. No call-once assumption anywhere in it. It's also strictly more robust than before — a cancellation during batch 5 used to lose batches 1-4's retry entries entirely, since the single call sat after the loop.
- Filtering now happens after chunking rather than before, so a batch can come out shorter than `batchSize`, or empty. `uploadBatch` on an empty list does one `UrlUtils.getUrl()` and an empty `map`/`awaitAll` — a harmless no-op.

One cosmetic nit: `Log.d(TAG, "Processing batch N with ${rawBatch.size} items")` now counts pre-filter items, so the log can overstate what was actually uploaded. Trivial, but it's the line someone will read when debugging an upload.

Worth being clear about the scope of the win: this bounds peak *payload* memory to `batchSize` instead of the whole backlog, which is the expensive half, but `fetchPendingItems()` still materialises every pending row up front. Paging the DAO query would close the rest. Given the entities are far smaller than their serialized payloads, this is most of the benefit for a fraction of the churn.

CI green 5/5.

**Verdict:** mergeable.

---

PR #16743, opened 2026-09-04
sync: smoother retry queue concurrent working (fixes #16723)

### LINKED ISSUE #16723: Drain the retry queue with bounded concurrency instead of one request at a time
## What

`RetryQueueWorker` chunks pending operations into batches of 50 and then processes each batch strictly sequentially, so the chunking buys nothing and a backlog routinely outlives the worker's own timeout.

**File:** `app/src/main/java/org/ole/planet/myplanet/services/retry/RetryQueueWorker.kt` (`doWork`)

## Why

```kotlin
withTimeout(5 * 60 * 1000L) {                                   // 124
    pendingOperations.chunked(BATCH_SIZE).forEach { batch ->     // 125
        if (MainApplication.isSyncRunning.get()) { ... return@withTimeout }   // 127-129
        batch.forEach { operation ->                             // 132  <-- sequential
            val success = processOperation(operation, baseUrl, authHeader)
            if (success) successCount++ else failureCount++
        }
    }
}
```

Every `processOperation` is a network round trip taken one at a time. A backlog of 200 failed uploads is 200 serial requests inside a 5-minute `withTimeout` that will often expire first — leaving the queue permanently behind on a slow link. The `chunked(BATCH_SIZE)` is currently decorative: the batch boundary changes nothing except where the `isSyncRunning` check happens.

`UploadCoordinator` in this same codebase already solves exactly this with `Semaphore(MAX_CONCURRENT_UPLOADS = 6)` + `async` + `withPermit` (lines 33, 128, 143).

## The fix

1. Add `MAX_CONCURRENT_RETRIES = 6` next to `BATCH_SIZE` (line 42).
2. Create the `Semaphore` **once**, before the `chunked` loop — not per batch.
3. Replace the inner `forEach` with:
   ```kotlin
   coroutineScope {
       batch.map { async { semaphore.withPermit { processOperation(it, baseUrl, authHeader) } } }.awaitAll()
   }
   ```
4. `successCount++` / `failureCount++` are now racy — replace them with counts computed over the awaited `List<Boolean>`.
5. Keep the `isSyncRunning` bail-out between batches exactly where it is.

## Acceptance

- A batch of N operations issues up to 6 concurrent requests, never more.
- Success/failure counts are correct under concurrency (no lost increments).
- One failing operation does not cancel its siblings — failure isolation must hold.
- The `isSyncRunning` early exit still stops the drain between batches.

## Notes

- Mirror the `UploadCoordinator` pattern rather than inventing a new one.
- Conventions: `docs/CODE_STYLE_GUIDE.md`; test patterns: `docs/TESTING.md` (`runTest`, `TestDispatcherProvider`).

dogi:
Updates RetryQueueWorker to process retry queue operations with bounded concurrency (up to 6 concurrent requests) using a Semaphore and coroutineScope async awaitAll pattern. Computed success and failure counts safely across concurrent results while keeping batch boundaries and sync checks intact. Added comprehensive unit tests in RetryQueueWorkerTest.

Fixes #16723

---
*PR created automatically by Jules for task [169139455391669476](https://jules.google.com/task/169139455391669476) started by @dogi*


dogi:
**Review — rating: 82/100**

Mechanically the same shape as #16741 and correct for the same reason: `successCount++` / `failureCount++` from parallel coroutines would have been a lost-update race, and collecting into `awaitAll()` then `partition { it }` removes the shared mutable counters entirely rather than guarding them. `Semaphore(MAX_CONCURRENT_RETRIES)` sits outside the `chunked` loop while batches stay sequential, so concurrency is bounded within a batch. The existing `withTimeout(5 * 60 * 1000L)` and the `isStopped` check are untouched, and draining in parallel means more of the queue clears inside that same five-minute ceiling.

**The one thing I'd want confirmed is ordering, and it matters more here than it did for photos.** A retry queue can hold several operations against the *same* document — a POST that failed and a later PUT for the same record, or two PUTs carrying different `_rev`s. Draining sequentially meant they replayed in the order they were queued; six at a time means they can now land out of order, and against CouchDB that surfaces as a document conflict or a lost update rather than a clean failure. Photo uploads are independent rows so #16741 doesn't have this exposure, but `RetryOperation` is generic over upload types. If the queue can ever contain two operations for one `dbId`, this wants either a per-document serialization key or a note in the PR explaining why that can't happen.

**Nit:** if `processOperation` can throw rather than returning `false`, `awaitAll` now discards the whole batch's results where the old loop kept the counts accumulated so far. Worth a glance to confirm it swallows its own failures.

**Verdict:** mergeable if same-document operations can't co-occur in the queue — worth answering that explicitly before it merges.

---

Okuro3499 (approved):

PR #16752, opened 2026-09-04
sync: smoother download service notification caching (fixes #16728)

### LINKED ISSUE #16728: Stop re-reading SharedPreferences and rebuilding the notification builder on every download tick
## What

`DownloadService` re-derives the "N remaining" count from `SharedPreferences` twice per second while a file downloads, and rebuilds the notification channel and builder once per file.

**File:** `app/src/main/java/org/ole/planet/myplanet/services/DownloadService.kt` (`getRemainingCount`, `sendNotification`, `updateNotificationForBatchDownload`, `processDownloadQueue`, `downloadFile`)

## Why

Four separate findings in one hot path — three agents each found a slice, this is the union.

1. **The remaining-count recompute.** `sendNotification` (line 451) fires every 500 ms (`NOTIFICATION_UPDATE_INTERVAL_MS`, line 554; gate at line 429) and calls `getRemainingCount()` (171), which does two `preferences.getStringSet` reads plus a union-set allocation, then filters against `processedUrls` — purely to render the subtext. But the queue only changes between files (`processedUrls.add` at 145, and `cleanupProcessedUrls`).
2. **Notification manager churn.** Each tick re-allocates `NotificationManagerCompat.from(this)` (460) just to ask `areNotificationsEnabled()`.
3. **Channel and builder churn.** `updateNotificationForBatchDownload` (189) re-runs `DownloadUtils.createChannels(this)` (190) and constructs a fresh `NotificationCompat.Builder` (191) once per file, even though `notificationBuilder` is already a field (69).
4. **A dead buffer.** `downloadFile` wraps the body in `BufferedInputStream(body.byteStream(), 1024 * 8)` (375) while reading into a `1024 * 16` `BUFFER_SIZE` array (548). Reads larger than the wrapper's buffer bypass it entirely, so the wrapper does nothing.

## The fix

- Cache the remaining count in a `@Volatile` field, refreshed at the two points where the queue actually changes.
- Read that cache in `sendNotification` and `updateNotificationForBatchDownload`. **Leave `onDownloadComplete` calling `getRemainingCount(priorityUrls)` directly** — it needs the post-download value, not the cached one.
- Create the channel and the builder once per queue session; mutate the existing builder per file instead of constructing a new one.
- Hoist the `areNotificationsEnabled()` check out of the 500 ms tick.
- Raise the `BufferedInputStream` buffer to `BUFFER_SIZE`.

Every notification ID, channel ID and subtext format string must stay **byte-identical** — this is a performance change, and a user-visible notification difference means it went wrong.

## Acceptance

- The "N remaining" subtext shows the same values it does today, including across file boundaries and on the final file.
- No `SharedPreferences` read inside the 500 ms tick.
- Notification IDs, channel IDs and subtext formatting unchanged.

## Notes

⚠️ Do **not** also take the "parallel downloads" variant of this finding in the same change. That is a separate and much riskier task; this issue is only about removing redundant work from the existing serial path.

dogi:
Optimized DownloadService by removing redundant SharedPreferences reads and NotificationManagerCompat calls from the 500ms notification tick. Cached the remaining count, hoisted channel creation and notification-enabled check to the queue loop, reused the notification builder, and adjusted stream buffer size to match read buffer size.

Fixes #16728

---
*PR created automatically by Jules for task [17114684394352656230](https://jules.google.com/task/17114684394352656230) started by @dogi*


dogi:
**Review — rating: 84/100**

The hot-path analysis is right, and it's worse on master than the title suggests. `updateNotification` runs once per buffer read, and each tick did `NotificationManagerCompat.from(this).areNotificationsEnabled()` — a binder round-trip — plus `getRemainingCount()`, which is two `preferences.getStringSet` reads, a set union, and a `count { it !in processedUrls }` filter. That's two SharedPreferences lookups and a fresh set allocation for every 16 KB of every file. Hoisting both to fields is the correct fix and the biggest win here.

The buffer change is a real consistency fix rather than a tuning knob: `data` is already `ByteArray(BUFFER_SIZE)` with `BUFFER_SIZE = 1024 * 16` on master, but the `BufferedInputStream` was constructed with a hardcoded `1024 * 8`. Reading into a 16 KB array through an 8 KB buffer meant every read straddled two refills. Aligning them removes that without changing any constant.

Moving `DownloadUtils.createChannels(this)` out of the per-file notification path to once per queue run is right too — channel creation is idempotent but not free.

**Reusing `notificationBuilder` carries state between downloads.** This is the one thing I'd change. The builder used to be reconstructed per file, which reset it; now `updateNotificationForBatchDownload` reuses the existing instance and only sets `setContentText` and `setProgress`. Everything else a previous file left on it persists — notably `downloadFailed` sets `setSubText("… completed, … remaining")` and an `Error:` content text, so the next file's "Starting downloads (0/N)" notification inherits the failed file's subtext until the first progress tick overwrites it. Clearing the fields you don't set (or resetting `setSubText(null)` in `updateNotificationForBatchDownload`) closes it.

**Nit:** `cachedRemainingCount` is `@Volatile` but `areNotificationsEnabled`, written in `processDownloadQueue` and read from the download path, is not — if one needed the annotation, so does the other. Also `areNotificationsEnabled` is now sampled once per queue run, so toggling the permission mid-queue isn't noticed until the next run, and `cachedRemainingCount` won't reflect downloads enqueued while a file is in flight until the next queue iteration. Both are display-only staleness, not correctness.

⚠️ **No build or test verdict on this head** — only `myPlanet labels` ran; `build` and `test` never fired. Everything above is from reading the diff. Worth a push to get CI to run before this merges.

**Verdict:** mergeable once the builder-state carryover is handled.

---

dogi:
Correction to my verdict above: I checked `downloadFailed` and it only sets `setContentText` and `setSubText` — no `setOngoing`/`setAutoCancel` state that would persist meaningfully. Both fields are overwritten by the first progress tick of the next file, so the carryover window is a single notification post, not a lingering wrong state.

That downgrades it from blocking to a nit worth tidying. Rating stands at 86/100 and this is **mergeable as-is** — labelling `ready`. The missing build/test verdict on this head is still the thing to resolve before it actually merges.

---

Okuro3499 (approved):

PR #16756, opened 2026-09-04
teams: smoother resources courses requests view modelling (fixes #16736)

### LINKED ISSUE #16736: Run the independent team-tab ViewModel queries concurrently
## What

Three team-tab ViewModels await independent repository calls strictly sequentially, so each tab's load time is the **sum** of its queries rather than the slowest one.

**Files:**
- `app/src/main/java/org/ole/planet/myplanet/ui/teams/members/RequestsViewModel.kt` (`fetchMembers`, line 34)
- `app/src/main/java/org/ole/planet/myplanet/ui/teams/courses/TeamCoursesViewModel.kt` (`loadCourses`, line 29)
- `app/src/main/java/org/ole/planet/myplanet/ui/teams/resources/TeamResourcesViewModel.kt` (`loadResources`, line 29)

## Why

```kotlin
// RequestsViewModel.fetchMembers
val members = teamsRepository.getRequestedMembers(teamId)      // independent
val memberCount = teamsRepository.getJoinedMemberCount(teamId) // independent
val user = userRepository.getUserModel()                       // independent
val isLeader = teamsRepository.isTeamLeader(teamId, user?.id)  // depends on user
```

```kotlin
// TeamCoursesViewModel.loadCourses
val courseIds = teamsRepository.getTeamCourseIds(teamId)
val courses = coursesRepository.getCoursesByIds(courseIds)     // depends on courseIds
val teamCreator = teamsRepository.getTeamCreator(teamId)       // independent of both
```

```kotlin
// TeamResourcesViewModel.loadResources
val libraries = teamsRepository.getTeamResources(teamId)       // independent
val canRemove = teamsRepository.isTeamLeader(teamId, userId)   // independent (userId is a parameter)
```

In each case at least one await is waiting on a result it never uses.

## The fix

Wrap each load body in `coroutineScope { }` with `async` per independent call, then await.

**Keep every real dependency edge:**
- `getCoursesByIds` still awaits `courseIds`.
- `isTeamLeader` in `RequestsViewModel` still awaits `user`.
- `isTeamLeader` in `TeamResourcesViewModel` takes `userId` as a parameter, so it has no dependency — it can start immediately.

**Preserve exactly one `MutableStateFlow` assignment per load**, with identical content. No intermediate emissions — a UI collecting the state must not observe a half-populated tab.

## Acceptance

- Each tab emits exactly one state per load, with the same contents as today.
- Dependency edges preserved (a test that makes `getTeamCourseIds` slow must still see `getCoursesByIds` receive its result).
- One failing call fails the load the same way it does now.

## Notes

- A fourth ViewModel (`CommunityTabViewModel`) was considered for the same treatment and **rejected** — three of its four reads are non-suspend `SharedPreferences` getters, so there is nothing to parallelize. Please don't extend this change to it.
- Conventions: `docs/CODE_STYLE_GUIDE.md`; test patterns: `docs/TESTING.md` (`runTest`, `TestDispatcherProvider`).

dogi:
Refactored load calls in RequestsViewModel, TeamCoursesViewModel, and TeamResourcesViewModel to run independent repository queries concurrently using coroutineScope and async, while preserving existing dependency edges and updating StateFlow state exactly once upon completion. Added unit tests verifying concurrency and dependency retention.

Fixes #16736

---
*PR created automatically by Jules for task [7977949574468104551](https://jules.google.com/task/7977949574468104551) started by @dogi*


dogi:
**Review — rating: 88/100**

Three ViewModels, and what makes this correct is that the *dependent* calls were left sequential while only the genuinely independent ones were parallelised. That's the part these refactors usually get wrong.

- `TeamCoursesViewModel`: `getTeamCourseIds` → `getCoursesByIds` is a real data dependency, and it stays sequential *inside* one `async`, with only `getTeamCreator` running alongside it. Chaining inside the async rather than launching both is exactly right.
- `RequestsViewModel`: three independent calls go parallel, but `isTeamLeader(teamId, user?.id)` needs `user`, so it's awaited first and then called outside the async block. The dependency is preserved rather than raced.
- `TeamResourcesViewModel`: two independent calls, straightforwardly parallel.

Using `coroutineScope { }` rather than calling `async` directly on the `launch` receiver is also the safer choice, and worth noting since it looks redundant: a failing `async` launched straight into `viewModelScope` would cancel the entire ViewModel scope, whereas inside `coroutineScope` the failure is confined — it cancels the siblings and surfaces at the `await`. Same failure behaviour the sequential code had.

Each of the three has a matching test, and Room handles concurrent reads fine, so nothing here needs transactional grouping — these were already three separate snapshots when they ran sequentially.

⚠️ **No build or test verdict on this head** — only `myPlanet labels` ran; neither `build` nor `test` fired, so the added tests have not actually executed anywhere. Worth a push to trigger CI before merging.

---

Okuro3499 (approved):

PR #16757, opened 2026-09-04
sync: smoother performance logging (fixes #16737)

### LINKED ISSUE #16737: Guard SyncManager's SyncPerf logging behind Log.isLoggable
## What

`SyncManager` has 20 unconditional `Log.d("SyncPerf", …)` calls with string interpolation and no `isLoggable` guard, so release devices pay for formatting that is then discarded.

**File:** `app/src/main/java/org/ole/planet/myplanet/services/sync/SyncManager.kt` — **the `Log.d("SyncPerf", …)` statements only**

## Why

Verified: exactly 20 `Log.d("SyncPerf", …)` calls, zero `isLoggable` guards (lines 139-141, 207-217, 274, 305, 382, 422, 427, 505, 516, 558, 563).

- Line 382 fires **once per resource batch** and interpolates a percentage computation.
- Lines 139 and 208 format an `Instant` on every sync.

The argument expressions are evaluated before `Log.d` is called, so the cost is paid whether or not the log is emitted.

`SyncTimeLogger` — same subsystem, same `"SyncPerf"` tag — already wraps all six of its `Log.d` calls in `if (Log.isLoggable("SyncPerf", Log.DEBUG))` (lines 71, 86, 148, 168, 185, 196). `SyncManager` is the odd one out.

## The fix

1. Add `private inline fun syncPerf(message: () -> String)` guarded by `Log.isLoggable("SyncPerf", Log.DEBUG)`. Inline + a lambda is what makes the interpolation itself conditional — a non-inline helper taking a `String` would still evaluate the argument.
2. Route all 20 sites through it, keeping message text **byte-identical**.
3. Wrap each of the three multi-line banner blocks in a single guard so the `Instant` formatting is skipped too, rather than guarding each line separately.

## Acceptance

- With `SyncPerf` debug logging enabled, output is byte-identical to today.
- With it disabled, no interpolation or `Instant` formatting is performed.
- All 20 sites converted — none left unguarded.

## Notes

⚠️ Scope limit: leave every `Log.e` / `Log.w` alone, and leave the `syncTimeLogger.logDetail` calls alone. This issue is only about the `Log.d("SyncPerf", …)` sites in this one file.

dogi:
Guards all 20 Log.d("SyncPerf", ...) calls in SyncManager behind Log.isLoggable("SyncPerf", Log.DEBUG) using a private inline function and block guards.

Fixes #16737

---
*PR created automatically by Jules for task [13179603117514054441](https://jules.google.com/task/13179603117514054441) started by @dogi*


dogi:
**Review — rating: 86/100**

The `inline` on `syncPerf` is what makes this worth doing, and it's easy to get wrong — without it, every call site would allocate a lambda and you'd trade string building for object churn. As written, the message expression is inlined inside the `isLoggable` branch, so on the common path nothing is built at all. That matters most at `Resources batch $batchCount: …`, which interpolates four values including a division, once per ten batches for the length of a full sync.

Worth being explicit about the behaviour change, because it isn't only a performance one: `Log.isLoggable(tag, DEBUG)` defaults to false, since the implicit per-tag level is INFO and DEBUG sits below it. So SyncPerf output disappears from logcat entirely until someone runs `adb shell setprop log.tag.SyncPerf DEBUG`. That's the standard Android idiom and almost certainly the intent, but it's a real change for anyone who currently debugs sync by reading these lines, and it belongs in the PR description.

**Nit:** the three multi-line banner blocks use a hand-written `if (Log.isLoggable(...))` while everything else goes through `syncPerf { }` — understandable since the helper takes a single message, but it leaves two idioms in one file. `"SyncPerf"` is also repeated as a literal in both forms rather than a `const`. And the interleaved `syncTimeLogger.logDetail(...)` calls still interpolate their arguments unconditionally, so the string building isn't fully gone from these paths — the guard covers the `Log.d` half of each pair only.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16758, opened 2026-09-04
all: smoother network utils device identity caching (fixes #16726)

### LINKED ISSUE #16726: Cache the invariant device identity and system services in NetworkUtils
## What

`NetworkUtils` recomputes invariant device identity strings and re-fetches system services on every call, on a path that runs once per uploaded document.

**File:** `app/src/main/java/org/ole/planet/myplanet/utils/NetworkUtils.kt` (`getDeviceName`, `getUniqueIdentifier`, `isWifiEnabled`, `isBluetoothEnabled`)

## Why

`getDeviceName()` (line 215) re-reads `Build.MANUFACTURER` / `Build.MODEL`, concatenates and uppercases on **every** call. It is called once per serialized document during uploads — `Personal`, `NewsLog`, `ApkLog`, `Rating`, `CourseActivity`, `MyLibrary`, `SearchActivity`, `MyPlanet`, `ActivitiesRepositoryImpl`, `UploadManager` (10 call sites) — so a bulk activity upload allocates thousands of throwaway strings for a value that cannot change during the process lifetime.

`getUniqueIdentifier()` (line 209) rebuilds the same `androidId + "_" + buildId` each call.

`isWifiEnabled()` (167) and `isBluetoothEnabled()` (182) call `getSystemService` on every invocation.

The file already shows the intended pattern: a `ResettableCache` helper (line 30), a `connectivityManagerCache` (56), and a `resettableCaches` list (72) drained by `resetForTesting()` (107).

## The fix

Add `ResettableCache` entries for `deviceName`, `uniqueIdentifier`, `wifiManager` and `bluetoothManager`.

- `isWifiEnabled()` / `isBluetoothEnabled()` must keep reading the **live** enabled state off the cached manager — cache the manager, never the boolean, or the app will report stale radio state.
- Register all four in `resettableCaches` (line 72) so `resetForTesting()` still clears them.

## Acceptance

- `getDeviceName()` and `getUniqueIdentifier()` compute once per process.
- Toggling wifi/bluetooth is still reflected immediately by `isWifiEnabled()` / `isBluetoothEnabled()`.
- `resetForTesting()` clears all four new caches — verify by asserting a value recomputed after a reset.

## Notes

⚠️ This overlaps with #16631, which documents how `NetworkUtils`'s process-wide caching pollutes Robolectric tests across a JVM fork. That issue is the reason `ResettableCache` and `resetForTesting()` exist. **Every new cache added here must go into the `resettableCaches` list** — a cache that skips registration reintroduces exactly the order-dependent shard-1 flake #16631 describes. Do not attempt to fix #16631 itself in this PR; just don't make it worse.

dogi:
Cached invariant device identity strings (`getDeviceName` and `getUniqueIdentifier`) and system services (`WifiManager` and `BluetoothManager`) using `ResettableCache` in `NetworkUtils`. Updated `resettableCaches` to register all new caches so `resetForTesting()` clears them properly, and added corresponding unit tests in `NetworkUtilsTest`.

Fixes #16726

---
*PR created automatically by Jules for task [17508808698301535443](https://jules.google.com/task/17508808698301535443) started by @dogi*


dogi:
**Review — rating: 88/100**

This follows the `ResettableCache` pattern the file already established for `connectivityManagerCache` rather than inventing a new one, and — the part that's easy to forget — every new cache is added to the reset list. That's what keeps the object testable: a cache registered but not reset would leak one test's device identity into the next.

The four values are genuinely invariant for a process lifetime, which is the only thing that makes caching them safe:
- `WIFI_SERVICE` / `BLUETOOTH_SERVICE` are process-scoped system service handles, and both were being re-fetched with a `getSystemService` + cast on every `isWifiEnabled()` / `isBluetoothEnabled()` call.
- `Build.MANUFACTURER`, `Build.MODEL` and `Build.ID` are compile-time constants of the running system image.
- `ANDROID_ID` can change across a factory reset or a re-signed install, but never while the process is alive.

`getUniqueIdentifier()` and `getDeviceName()` are called on upload and sync paths, so removing a `Settings.Secure` lookup and a pair of `uppercase` allocations per call is a genuine if modest win. Keeping `Locale.ROOT` on the uppercase is right — device names must not be Turkish-lowercased into a different string.

**Nit:** `getSystemService(BLUETOOTH_SERVICE) as BluetoothManager` is an unchecked cast that can fail on a device without Bluetooth. That risk is unchanged from master, but its timing moves: it now throws on first access to the cached property rather than on each call, so where the failure surfaces depends on `ResettableCache`'s behaviour when the initializer throws. Worth confirming it doesn't memoise a broken state.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16759, opened 2026-09-04
all: less user session full name is more (fixes #16748)

### LINKED ISSUE #16748: Remove the dead eager fullName read from UserSessionManager's constructor
## What

`UserSessionManager` reads a preference in its `init` block into a field nothing ever reads.

**File:** `app/src/main/java/org/ole/planet/myplanet/services/UserSessionManager.kt` (lines ~27-36)

## Why

```kotlin
private val fullName: String        // line 28
init {
    ...
    fullName = sharedPrefManager.getUserName()   // line 32
    ...                                          // catch that only rethrows
}
```

The only two references to `fullName` in the file are its declaration and that assignment — nothing reads it.

So every injection of this singleton performs a `SharedPreferences` read whose result is discarded, and the `init` block makes the class fragile to construct before login (the catch only rethrows, so it provides no protection).

## The fix

Delete the property and the whole `init` block.

- Drop any test stub that existed only to satisfy the removed read.
- Keep a construction test green against an **unstubbed** `SharedPrefManager` mock — that is what proves construction no longer touches preferences, and it is the part of this change worth testing.

## Acceptance

- `UserSessionManager` constructs successfully with a `SharedPrefManager` mock that has no `getUserName()` stub.
- No other behavior of the class changes.

## Notes

- Scope: only the dead `fullName` field and its `init` block. Do not otherwise reshape `UserSessionManager` — its login/logout side effects and counters are in use.

dogi:
Removed dead fullName property and init block from UserSessionManager, and updated unit tests to verify construction with an unstubbed SharedPrefManager mock.

Fixes #16748

---
*PR created automatically by Jules for task [18035042393391951833](https://jules.google.com/task/18035042393391951833) started by @dogi*


dogi:
**Review — rating: 92/100**

Three things wrong with those ten lines, and deleting them fixes all three.

The field had no readers — `fullName` was written in `init` and never used anywhere in the class, which the green build confirms independently of reading the file. The `try { … } catch (IllegalArgumentException) { throw e }` around it was a no-op: catching an exception only to rethrow the same instance changes nothing about propagation. And because `UserSessionManager` is constructor-injected, the `init` block put a `SharedPreferences` read on the injection path, so every graph construction paid for a value nobody consumed.

Removing it also removes a failure mode: if `getUserName()` ever threw, the exception surfaced during Hilt construction — the worst place for it, since it takes down whatever was being injected rather than the code that actually wanted a username. Nothing now depends on that read happening eagerly.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16760, opened 2026-09-04
all: less news model message without markdown is more (fixes #16751)

### LINKED ISSUE #16751: Delete the dead, re-parsing News.messageWithoutMarkdown getter
## What

`News.messageWithoutMarkdown` is dead code that carries a hidden per-call Gson parse.

**File:** `app/src/main/java/org/ole/planet/myplanet/model/News.kt` (the `messageWithoutMarkdown` getter)

## Why

The getter iterates `imagesArray` — itself a fresh `Gson` parse on every access — and string-replaces each image's markdown out of the message.

Nothing calls it:

```
$ grep -rn messageWithoutMarkdown app/src/
```

returns exactly one hit: the declaration itself.

So it is dead code that would hand a hidden per-call parse cost to any future caller who assumed it was a cheap property read.

## The fix

Delete the getter.

## Acceptance

- `grep -rn messageWithoutMarkdown app/src/` returns nothing.
- No `@Entity` shape or Room column changes — this is a computed getter, not a persisted field, so **no `AppDatabase` version bump**.

## Notes

- This issue is deliberately narrow because another change (caching `imagesArray` / `isCommunityNews` against the existing `parsedImagesArray` / `parsedViewIn` fields) is queued behind it — landing this first means that one has one less consumer to cache for. Please don't take the caching work here.

dogi:
Deletes the unused messageWithoutMarkdown getter in News.kt which performed per-call Gson parsing.

Fixes #16751

---
*PR created automatically by Jules for task [15267896872845353041](https://jules.google.com/task/15267896872845353041) started by @dogi*


dogi:
**Review — rating: 90/100**

Pure deletion, and I verified it's actually dead rather than trusting the build — which matters here, because **this head has no build or test verdict**: only `myPlanet labels` ran, so neither `build` nor `test` fired. Grepping the whole of `app/src` for `messageWithoutMarkdown` across Kotlin and XML returns exactly one hit, the declaration itself at `News.kt:113`. No callers, no layout binding expression, nothing in tests.

Worth removing on its own merits beyond being unused: the getter looped over `imagesArray` doing a `JsonUtils.getString` and a `String.replace` per image, so any future caller would have been paying repeated JSON parsing inside what looks like a plain property read — the kind of accessor that gets called in a `bind()` without a second thought.

**Verdict:** mergeable — worth a push to get an actual CI verdict first, since nothing has compiled this branch.

---

Okuro3499 (approved):

PR #16761, opened 2026-09-04
sync: smoother repository shelves caching (fixes #16749)

### LINKED ISSUE #16749: Move the shelves-with-data preference cache out of SyncManager into SyncRepository
## What

`SyncManager` hand-rolls a 6-hour `SharedPreferences` cache — key names, encoding and expiry — inside the sync orchestrator.

**Files:**
- `app/src/main/java/org/ole/planet/myplanet/repository/SyncRepository.kt`
- `app/src/main/java/org/ole/planet/myplanet/repository/SyncRepositoryImpl.kt`
- `app/src/main/java/org/ole/planet/myplanet/services/sync/SyncManager.kt` (`getCachedShelvesWithData` line 477, `cacheShelvesWithData` line 494, call sites 442/469)

## Why

Both private functions re-declare the same string keys:

```kotlin
val cacheKey = "shelves_with_data"      // 478, and again at 495
val cacheTimeKey = "shelves_cache_time" // 479, and again at 496
val cacheValidityHours = 6              // 480
if (now - cacheTime < cacheValidityHours * 60 * 60 * 1000) { ... }   // 485
```

plus their own comma split/join. That is persistence logic — key names, encoding, expiry — living in the orchestrator, agreeing with itself only by convention, and untestable without driving a whole sync.

## The fix

1. Add `getCachedShelvesWithData()` / `cacheShelvesWithData(shelves)` to `SyncRepository`.
2. Implement in `SyncRepositoryImpl` with `SharedPrefManager` and `TimeProvider` constructor parameters — both are already Hilt-provided, so **no DI module change is needed**.
3. Move the keys and the TTL into private companion constants, declared once.
4. Delete the two private functions from `SyncManager` and point it at the repository.

## Acceptance

⚠️ **Keep the exact same preference keys (`shelves_with_data`, `shelves_cache_time`), the comma encoding, and the 6-hour window** — existing installs must keep their cache rather than silently re-syncing every shelf on upgrade.

- Cache hit within 6 hours returns the stored list.
- Cache older than 6 hours is treated as a miss.
- Round-trip through the comma encoding preserves the list.
- `SyncManager` behavior at lines 442/469 unchanged.

## Notes

- Inject `TimeProvider` rather than calling `System.currentTimeMillis()` directly, so the TTL is testable without sleeping.

dogi:
Moved shelves-with-data preference cache out of SyncManager into SyncRepository/SyncRepositoryImpl. Added getCachedShelvesWithData() and cacheShelvesWithData() to SyncRepository, implemented them in SyncRepositoryImpl using injected SharedPrefManager and TimeProvider while preserving exact preference keys, comma encoding, and 6-hour expiry window, updated SyncManager call sites, and added unit tests in SyncRepositoryImplTest.

Fixes #16749

---
*PR created automatically by Jules for task [9227465083631147697](https://jules.google.com/task/9227465083631147697) started by @dogi*


dogi:
**Review — rating: 86/100**

Right move: `SyncManager` was hand-rolling two SharedPreferences keys as string literals duplicated across a getter and a setter, which is persistence detail sitting in an orchestrator. Moving both behind `SyncRepository` puts the storage concern with the other sync-domain storage, and hoisting `"shelves_with_data"` / `"shelves_cache_time"` into a `private companion` means the read and the write can no longer disagree about the key — previously each function re-declared both literals locally, so a typo in one would have produced a cache that silently never hit.

The logic moved verbatim, including the six-hour window and the `split(",").filter { it.isNotBlank() }` guard against a trailing separator, so behaviour is unchanged. `CACHE_VALIDITY_HOURS * 60 * 60 * 1000` is 21.6M, comfortably inside `Int`, same as before. `SyncManager` still uses `sharedPrefManager` and `timeProvider` on other paths, so neither injection is left dangling.

**Nit:** the interface now carries a two-call contract — callers must remember to `cacheShelvesWithData` after a miss, and `SyncManager` is the only thing keeping that discipline. A single `getOrComputeShelvesWithData(compute: suspend () -> List<String>)` would make the caching invariant unforgeable rather than conventional. Also, both methods are non-suspend and do preference I/O; that matches the existing `SharedPrefManager` style so it's consistent, just worth knowing they're not free on first access.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16762, opened 2026-09-04
all: smoother dictionary repository word view modelling (fixes #16746)

### LINKED ISSUE #16746: Stop leaking the Room DictionaryEntity into the dictionary UI
## What

`DictionaryRepository.findByWord` returns the Room `@Entity` itself, and the ViewModel holds it in UI state — the single remaining place where a persistence type crosses into the UI layer.

**Files:**
- `app/src/main/java/org/ole/planet/myplanet/repository/DictionaryRepository.kt` (`findByWord`, line 14)
- `app/src/main/java/org/ole/planet/myplanet/repository/DictionaryRepositoryImpl.kt`
- `app/src/main/java/org/ole/planet/myplanet/ui/dictionary/DictionaryViewModel.kt`

## Why

`DictionaryViewModel` imports `org.ole.planet.myplanet.data.room.entity.DictionaryEntity` (line 11) and stores it as `DictionarySearchState.Found(entry: DictionaryEntity)`.

This is the **only** leak left:

```
$ grep -rn 'data.room.entity' app/src/main/java | grep -v /data/room/ | grep -v /repository/
app/src/main/java/org/ole/planet/myplanet/ui/dictionary/DictionaryViewModel.kt:11
```

One hit. Closing it makes the layer boundary absolute rather than nearly-absolute, which is worth more than the change itself.

## The fix

1. Declare a plain `data class DictionaryWord(word, meaning, definition, synonym, antonym)` next to the existing `DictionaryLoad` sealed interface.
2. Change `findByWord`'s signature to return `DictionaryWord?`, mapping inside `findByWord` in the Impl.
3. Hold `DictionaryWord` in `DictionarySearchState.Found`.

`DictionaryActivity.renderSearchState` reads only `word` / `definition` / `synonym` / `antonym`, so it compiles untouched.

## Acceptance

- The grep above returns **zero** hits outside `data/room/` and `repository/`.
- Dictionary search behaves identically, including the not-found path.
- `DictionaryActivity` is unchanged.

## Notes

dogi:
Decouples DictionaryViewModel and dictionary UI from Room persistence layer by replacing DictionaryEntity with the domain model data class DictionaryWord in DictionaryRepository.findByWord and DictionarySearchState.Found.

Fixes #16746

---
*PR created automatically by Jules for task [5285418503183455888](https://jules.google.com/task/5285418503183455888) started by @dogi*


dogi:
**Review — rating: 84/100**

The decoupling is real: `DictionaryViewModel` was importing `data.room.entity.DictionaryEntity` and exposing it inside `DictionarySearchState.Found`, so the UI's state type was the persistence schema. Any column rename would have rippled into the ViewModel and into anything observing that state. Mapping to a `DictionaryWord` in the repository puts the translation at the layer boundary where it belongs, and the mapping is total — five fields, all defaulted — so there's no partial-construction path.

One thing worth being deliberate about: `CLAUDE.md` describes the repository convention as returning "plain `@Entity`/data-class instances", and the other 22 repositories return entities directly. So this is a stricter choice than the house style rather than a fix to a violation of it. I think it's the better default and the direction the codebase should move, but as a single instance it makes the dictionary repository the odd one out. If it's the first step of a deliberate migration, saying so in the description turns an inconsistency into a plan; if it isn't, the same argument applies to `Personal`, `Rating` and the rest, and it'd be worth knowing why this one came first.

`DictionaryEntity` has no nullable fields to reconcile, so the mapping compiles cleanly and the ViewModel's `Found` state loses nothing.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16763, opened 2026-09-04
enterprises: smoother reports flow view modelling (fixes #16747)

### LINKED ISSUE #16747: Drop the false suspend on getReportsFlow and stop nesting the lifecycle-aware collector
## What

`EnterprisesViewModel.getReportsFlow` is declared `suspend` for no reason, and that forces the fragment into a nesting that leaks a second collector after rotation.

**Files:**
- `app/src/main/java/org/ole/planet/myplanet/ui/enterprises/EnterprisesViewModel.kt` (`getReportsFlow`, line 100)
- `app/src/main/java/org/ole/planet/myplanet/ui/enterprises/EnterprisesReportsFragment.kt` (lines ~126-131)

## Why

```kotlin
// EnterprisesViewModel.kt:100
suspend fun getReportsFlow(teamId: String): Flow<List<MyTeam>> { ... }
```

It only returns the cold `Flow` from `EnterprisesRepository.getReportsFlow`, which is correctly **non**-suspend (`EnterprisesRepository.kt:11`). Nothing suspends.

The false `suspend` forces the fragment to wrap the call in a plain `launch` just to obtain the flow:

```kotlin
viewLifecycleOwner.lifecycleScope.launch {          // outlives STARTED
    val flow = viewModel.getReportsFlow(teamId)
    collectLatestWhenStarted(flow) { reportList ->  // lifecycle-aware, nested inside
        updatedReportsList(reportList)
    }
}
```

The lifecycle-aware collector is nested inside a scope that outlives `STARTED`, so a rotation leaves the outer `launch` alive and starts a second collector — the leak. Note line 132 in the same fragment already calls `collectLatestWhenStarted` directly, which is the shape this one should have.

## The fix

1. Remove `suspend` from `getReportsFlow`, keeping the `Flow<List<MyTeam>>` return type.
2. Collapse the fragment to a single `collectLatestWhenStarted(viewModel.getReportsFlow(teamId)) { ... }` — no enclosing `launch`.

Leave `exportReportsAsCsv` genuinely `suspend` on both sides; it does real work.

## Acceptance

- Rotating the reports screen leaves exactly one active collector.
- Reports still render on first load and update on data change.
- `exportReportsAsCsv` unchanged.

## Notes

dogi:
Removes unnecessary suspend modifier from EnterprisesViewModel.getReportsFlow and collapses the collector in EnterprisesReportsFragment to avoid nested coroutine launchers that leak on rotation.

Fixes #16747

---
*PR created automatically by Jules for task [7775698150072766789](https://jules.google.com/task/7775698150072766789) started by @dogi*


dogi:
**Review — rating: 90/100**

`getReportsFlow` never suspended — it just returns a cold `Flow` built by the repository — so the `suspend` marker was claiming something untrue about the function, and every caller had to open a coroutine to get a value that was available synchronously. Dropping the modifier and unwrapping the `lifecycleScope.launch` removes a coroutine whose entire job was calling a function that starts its own.

That nesting was slightly worse than redundant: `collectLatestWhenStarted` already binds collection to the view lifecycle, so wrapping it in an outer `launch` created a second coroutine with different cancellation timing around one that was already lifecycle-aware. Calling it directly means collection is registered on the spot rather than whenever the outer coroutine happened to be dispatched.

The signature change is safe because the repository's `getReportsFlow` is itself non-suspend — Room requires `Flow`-returning DAO methods to be, per the Room conventions in `CLAUDE.md` — so nothing downstream loses a suspension point. Green build confirms no other caller depended on the `suspend`.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16770, opened 2026-09-04
life: smoother personal repository resources dao updating (fixes #16745)

### LINKED ISSUE #16745: Replace the personals read-modify-write update with one targeted SQL UPDATE
## What

`updatePersonalResource` issues up to three statements and rewrites every column to change a title, opening a lost-update window against the sync writer.

**Files:**
- `app/src/main/java/org/ole/planet/myplanet/data/room/dao/PersonalDao.kt` (one new `@Query`)
- `app/src/main/java/org/ole/planet/myplanet/repository/PersonalsRepositoryImpl.kt` (`updatePersonalResource`, line 64)

## Why

```kotlin
override suspend fun updatePersonalResource(id: String, update: PersonalUpdate) {
    val personal = personalDao.findByDocId(id) ?: personalDao.findById(id)   // 1-2 SELECTs
    personal?.let {
        if (update.title != null) it.title = update.title
        if (update.description != null) it.description = update.description
        personalDao.update(it)                                                // full-row @Update
    }
}
```

Editing a title rewrites **every** column — including `path`, `_rev` and `isUploaded` — from a snapshot read microseconds earlier. That is three statements where one suffices, and it races the sync writer that touches `isUploaded` / `_rev` via `updateUploadedStatus`: a concurrent sync write between the SELECT and the `@Update` is silently reverted.

## The fix

Add to `PersonalDao`:

```kotlin
@Query("UPDATE my_personal SET title = COALESCE(:title, title), description = COALESCE(:description, description) WHERE _id = :id OR id = :id")
suspend fun updateFields(id: String, title: String?, description: String?)
```

- The `WHERE _id = :id OR id = :id` mirrors the existing `deleteByIdOrDocId` (line 40), so both id shapes resolve exactly as the current two-SELECT fallback does.
- `COALESCE` preserves `PersonalUpdate`'s "null means leave unchanged" contract.

Reduce `updatePersonalResource` to that single call.

**Leave `findByDocId`, `findById` and `update` in place** — `savePersonalResource` and the upload path still use them.

## Acceptance

- Updating only the title leaves `description`, `path`, `_rev` and `isUploaded` untouched.
- Both id shapes (`_id` and `id`) still resolve.
- A `PersonalUpdate` with a null field leaves that column unchanged.
- Adding a `@Query` changes no table, so **no `AppDatabase` version bump** (a bump under `fallbackToDestructiveMigration` would drop unsynced local writes).

## Notes

- A weaker variant of this finding merged the two SELECTs into one. Please take the single `UPDATE` instead — it removes all three statements *and* the race.

dogi:
Replaces the multi-statement read-modify-write pattern in PersonalsRepositoryImpl.updatePersonalResource with a single targeted SQL UPDATE in PersonalDao using COALESCE. This prevents racing with CouchDB sync processes and overwriting untouched columns.

Fixes #16745

---
*PR created automatically by Jules for task [12894077118268380261](https://jules.google.com/task/12894077118268380261) started by @dogi*


dogi:
**Review — rating: 88/100**

This is a correctness improvement as much as a query-count one, and that's the stronger argument for it. The old path did `findByDocId(id) ?: findById(id)`, mutated the entity in memory, then called `@Update`, which rewrites **every** column of the row. So an upload completing between the read and the write — `updateUploadedStatus` sets `isUploaded`, `_id` and `_rev` — would be silently reverted by the write-back of a stale entity. The targeted `UPDATE … SET title = …, description = …` touches only the two columns the caller actually asked to change, so that lost-update window closes.

`COALESCE(:title, title)` reproduces the `if (update.title != null)` guard exactly: a null argument leaves the existing column value in place rather than nulling it. That's the trap in this refactor and it's handled. The two-query read-then-write also collapses to one statement.

**One difference worth checking:** `WHERE _id = :id OR id = :id` is not quite the old lookup. `findByDocId(id) ?: findById(id)` resolved to a *single* row, preferring the `_id` match; the `OR` updates every row satisfying either predicate. If a row can exist whose `id` equals another row's `_id`, the old code updated one and the new code updates both. That's probably impossible in practice given how these ids are minted, but the fallback ordering in the original suggests the two columns aren't always interchangeable, so it's worth a moment's thought rather than an assumption.

**Note on #16784:** it also touches `PersonalDao.kt`, but at a different query, and the only real conflict between the two is an add/add on `PersonalDaoTest.kt` — both PRs create that file fresh. The production code doesn't overlap, so both land fine; whoever merges second just concatenates the two test classes.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16771, opened 2026-09-04
life: smoother health user view modelling (fixes #16753)

### LINKED ISSUE #16753: Give UserEntity one effective-id accessor and use it in the health screens
## What

The "prefer the CouchDB `_id`, fall back to the local `id`" rule is copy-pasted three times in the health feature — an identity rule that decides which health records a patient sees, living in the UI in triplicate.

**Files:**
- `app/src/main/java/org/ole/planet/myplanet/model/UserEntity.kt` (new top-level extension)
- `app/src/main/java/org/ole/planet/myplanet/ui/health/HealthViewModel.kt` (line 79)
- `app/src/main/java/org/ole/planet/myplanet/ui/health/MyHealthFragment.kt` (lines 142, 273)

## Why

```kotlin
// HealthViewModel.kt:79        (nullable receiver)
val uid = if (currentUser?._id.isNullOrEmpty()) currentUser?.id else currentUser?._id
// MyHealthFragment.kt:142      (non-null receiver)
userId = if (currentUser._id.isNullOrEmpty()) currentUser.id else currentUser._id
// MyHealthFragment.kt:273      (non-null receiver)
userId = if (selected._id.isNullOrEmpty()) selected.id else selected._id
```

Three copies of one rule. If they ever drift, a patient sees the wrong records.

## The fix

Add to `UserEntity.kt`:

```kotlin
val UserEntity.effectiveId: String? get() = _id?.takeIf { it.isNotEmpty() } ?: id
```

⚠️ It must be a **top-level extension property, declared outside the `@Entity` class**. A member property would become a Room column candidate and require `@Ignore`; an extension leaves the schema completely untouched, so there is no `AppDatabase` version bump.

Then use it at all three sites.

## Acceptance

Tests for three cases:
- `_id` present → returns `_id`
- `_id` null → returns `id`
- **`_id` empty string → returns `id`**

The third case is the one that matters: it is what the existing `isNullOrEmpty()` check handles, and a naive `_id ?: id` rewrite would return `""` instead of falling back. Please make sure the implementation uses `takeIf { it.isNotEmpty() }` (or equivalent) rather than a bare elvis.

## Notes

dogi:
Adds top-level extension property `effectiveId` on `UserEntity` and refactors call sites in `HealthViewModel` and `MyHealthFragment` to eliminate duplicated fallback logic. Unit tests added for `effectiveId` edge cases.

Fixes #16753

---
*PR created automatically by Jules for task [2680577836180057594](https://jules.google.com/task/2680577836180057594) started by @dogi*


dogi:
**Review — rating: 90/100**

The same three-branch expression was written out three times across two files, and the extension collapses it to one name. I checked the rewrite is exactly equivalent across all three inputs: `_id` null → `takeIf` never runs, `?:` yields `id`; `_id` empty → `takeIf { it.isNotEmpty() }` yields null, `?:` yields `id`; `_id` non-empty → returned as-is. Identical to `if (_id.isNullOrEmpty()) id else _id` in every case.

Making it an extension property rather than a member is the right call for a Room `@Entity` specifically — a computed member `val` on an entity is the kind of thing that needs `@Ignore` reasoning and invites the next person to wonder whether it's persisted. An extension can't be mistaken for a column.

The callers keep their `.trim()` and `isNullOrEmpty` checks afterwards, so nothing downstream changes; this only removes the duplicated selection logic, which is the part that could have drifted between the three copies.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16779, opened 2026-09-04
all: smoother notifications view modelling (fixes #16755)

### LINKED ISSUE #16755: Resolve notification group labels in the adapter instead of carrying them through state
## What

Notification group labels are resolved to strings in the ViewModel and carried through UI state as data, so a locale change leaves them stale.

**Files:**
- `app/src/main/java/org/ole/planet/myplanet/model/NotificationListItem.kt` (drop `label` from `Header`, line 6)
- `app/src/main/java/org/ole/planet/myplanet/ui/notifications/NotificationsViewModel.kt` (`NotificationGroup.label` line 54, construction sites 279/301, `typeLabelFor` line 311)
- `app/src/main/java/org/ole/planet/myplanet/ui/notifications/NotificationsAdapter.kt`

## Why

`typeLabelFor` (line 311) is eight `context.getString` branches in the ViewModel, and the resolved string travels as data: `NotificationGroup.label` → `NotificationListItem.Header.label` → `binding.tvHeaderLabel.text = header.label` (adapter line 94).

The sibling concern in the very same adapter is already done right — the icon is resolved from the type at bind time:

```kotlin
// NotificationsAdapter.kt:167, 177
private val ICON_BY_TYPE = mapOf(...)
internal fun iconResFor(type: String): Int = ICON_BY_TYPE[type.lowercase(Locale.ROOT)] ?: R.drawable.ic_notifications
// line 95
binding.ivHeaderIcon.setImageResource(iconResFor(header.type))
```

The label should work the same way. As it stands, a locale change leaves every header label stale until the flow re-emits.

## The fix

1. Add `LABEL_RES_BY_TYPE` + `labelResFor(type)` beside `ICON_BY_TYPE`, in the same shape.
2. Bind with `binding.tvHeaderLabel.setText(labelResFor(header.type))`.
3. Delete the `label` property from `NotificationListItem.Header`, from `NotificationGroup`, from both construction sites (279, 301), and delete `typeLabelFor`.

## Acceptance

- Header labels render identically for all eight types, plus the fallback for an unknown type.
- Changing device locale updates header labels without waiting for a flow re-emission.
- `NotificationsViewModelTest` reads headers via `filterIsInstance` and asserts on `type`, so it should compile unchanged — if it doesn't, say so rather than weakening the test.

## Notes

⚠️ Scope limit: a broader variant of this finding proposed a resource-descriptor type covering the message bodies too. The message formatters (`formatTaskNotification`, `formatStorageNotification`, `formatJoinRequestNotification`) are a much larger extraction — **keep them out of this change**.

dogi:
This change resolves notification group header labels in the adapter via resource IDs rather than resolving strings in the ViewModel and passing them through UI state.

Key changes:
- Removed `label` property from `NotificationListItem.Header` and `NotificationGroup`.
- Removed `typeLabelFor` from `NotificationsViewModel` and updated group header creation.
- Added `LABEL_RES_BY_TYPE` map and `labelResFor(type)` in `NotificationsAdapter`.
- Bound group labels using `binding.tvHeaderLabel.setText(labelResFor(header.type))`.
- Added unit tests for `labelResFor` in `NotificationsAdapterTest`.

Fixes #16755

---
*PR created automatically by Jules for task [3264756906695841361](https://jules.google.com/task/3264756906695841361) started by @dogi*


dogi:
**Review — rating: 88/100**

This is a localization fix as much as a state-hygiene one, and that's the stronger reason to take it. The old `typeLabelFor` resolved strings through the ViewModel's `@ApplicationContext`, then stored the resulting `String` inside `NotificationListItem.Header` — which lives in a `StateFlow` that survives configuration changes. So a locale switch left every already-built header holding text in the previous language until something happened to rebuild the list. Application-context resources are also the least reliable place to resolve user-facing strings once per-app locales are in play. Resolving `@StringRes` at bind time via `setText(labelResFor(header.type))` uses the view's own context and is correct by construction.

Dropping `label` from both `Header` and the private `NotificationGroup` is the right consequence — the state now carries `type`, the identity, and lets the view decide presentation, which also makes `Header` comparable across locales rather than differing by rendered text.

Putting `LABEL_RES_BY_TYPE` next to the existing `ICON_BY_TYPE` map is a nice touch: the two now sit together keyed the same way, so adding a notification type means editing two adjacent maps rather than a `when` in one file and a map in another. `labelResFor` keeps the `lowercase(Locale.ROOT)` normalization the old `when` had, so type matching is unchanged.

I confirmed `context` and `Locale` are still used elsewhere in the ViewModel (message formatting at lines 366-417, type normalization at 268), so neither the injection nor the import is left dangling.

**Verdict:** mergeable.

---

Okuro3499 (approved):

Okuro3499 (approved):

PR #16780, opened 2026-09-04
all: smoother android decrypter logging (fixes #16765)

### LINKED ISSUE #16765: Give AndroidDecrypter's four failure paths real logging
## What

All four failure paths in `AndroidDecrypter` catch broadly, print a stack trace to stderr, and silently return `null` / `false` / `""`.

**Files:**
- `app/src/main/java/org/ole/planet/myplanet/utils/AndroidDecrypter.kt`
- its test

## Why

Verified: 4 `printStackTrace` calls, one in each of `decrypt` (line 58), `androidDecrypter` (94), `generateIv` (111), `generateKey` (123).

A failed health-record decryption or a failed login key comparison is one of the more important things this app can get wrong, and right now it produces an untagged stderr dump that bypasses the log pipeline entirely — then returns a value indistinguishable from a legitimate negative result.

## The fix

Replace the four `printStackTrace` calls with tagged logging, **keeping the existing return contracts exactly** (`null` / `false` / `""` as today). This issue does not change behavior — it makes the failures visible.

Add tests pinning the contract on bad input:
- malformed hex → `null`
- invalid `dbPwdKeyValue` → `false`
- `generateKey` / `generateIv` non-blank on the happy path

## Notes

- Use `android.util.Log` to match house style. (The original proposal suggested `java.util.logging.Logger` — only prefer that if keeping this file Android-free is an explicit goal, and say so in the PR if you go that way.)
- Do not narrow the broad catches in this change; that is a separate judgement call.

dogi:
Replaced printStackTrace calls in AndroidDecrypter failure paths with tagged Log.e logging, maintaining return behavior (null / false / ""). Added unit tests for invalid inputs and generation methods in AndroidDecrypterTest.

Fixes #16765

---
*PR created automatically by Jules for task [11648558254864588007](https://jules.google.com/task/11648558254864588007) started by @dogi*


dogi:
**Review — rating: 88/100**

Right change, and in a file where it's worth more than the usual logging cleanup: `printStackTrace` writes to `System.err`, which on Android is best-effort and untagged, so failures in password verification were effectively invisible in a bug report. Tagged `Log.e` with distinct messages per site means a failed `androidDecrypter` is now distinguishable from a failed key generation.

The best line in the diff is the one that wasn't a `printStackTrace` at all — the `catch { return false }` around `hexStringToByteArray` was silently swallowing a malformed `dbPwdKeyValue` and returning "wrong password". That failure now says what it was.

Since this is credential-handling code I checked the obvious hazard, that logging an exception could echo secret material into logcat. It can't here: `hexStringToByteArray` uses `Character.digit`, which returns `-1` for non-hex characters rather than throwing, so the only exception it can raise is an `ArrayIndexOutOfBoundsException` from `s[i + 1]` on an odd-length string — and that message carries an index, not the string. The other sites log crypto exceptions whose messages are algorithm names. No key or password text reaches the log.

**Nit:** the app ships with `minifyEnabled = false` and no log-stripping ProGuard rule, so these `Log.e` calls are live in release builds. That's fine given what they say, but it's the reason the "no secrets in the message" property above needs to keep holding as this file changes.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16781, opened 2026-09-04
chat: smoother history model payload sharing (fixes #16767)

### LINKED ISSUE #16767: Extract the chat-share payload builder out of ChatHistoryAdapter
## What

A RecyclerView adapter builds a CouchDB-shaped document payload inline, where it cannot be tested.

**Files:**
- `app/src/main/java/org/ole/planet/myplanet/ui/chat/ChatHistoryAdapter.kt` (`showEditTextAndShareButton` line 243, `serializeConversation` line 278)
- `app/src/main/java/org/ole/planet/myplanet/model/ChatSharePayload.kt` (new)

## Why

`showEditTextAndShareButton` assembles a `HashMap<String?, String>` carrying `_id`, `_rev`, `title`, `user`, `aiProvider`, `createdDate`, `updatedDate` and the conversations serialized with `JsonUtils.gson` — then wraps it in an outer map of `message` / `viewInId` / `viewInSection` (line 262) / `messageType` / `messagePlanetCode` (line 264) / `chat` / `news`.

That is document serialization living in a view class. It cannot be unit-tested where it sits, and the key set — which the server depends on — can drift silently.

## The fix

Add a pure function `buildShareMap(chat, note, team, section, nowMillis)` producing **byte-for-byte the same keys and values**, with `serializeConversation` as a private helper. Call it from the adapter, passing `Date().time` from the call site (so the function stays pure and testable).

## Acceptance

Tests covering:
- **null team** → empty `viewInId` / `messageType` / `messagePlanetCode`
- **populated team** → those fields filled from the team
- **null `_id` / `_rev`** → empty strings, not the literal "null"
- a conversation round-trip through `serializeConversation`
- **an assertion on the exact outer key set**, so the payload schema cannot drift unnoticed

That last one is the point of the whole change — please include it.

## Notes

- Pass time in rather than calling `Date().time` inside the builder; a builder that reads the clock isn't pure and the `createdDate` / `updatedDate` assertions become flaky.

dogi:
Extracted CouchDB document payload creation from `ChatHistoryAdapter` into pure `ChatSharePayload.buildShareMap` function. Added comprehensive unit tests in `ChatSharePayloadTest` covering key set schema, team population/null states, null _id/_rev fallback, and conversation JSON serialization.

Fixes #16767

---
*PR created automatically by Jules for task [8890838801731121507](https://jules.google.com/task/8890838801731121507) started by @dogi*


dogi:
**Review — rating: 88/100**

Good extraction — a payload that gets serialized and shipped to the server was being assembled inside a positive-button click listener, which meant the only way to verify its shape was to open a dialog and tap. As a pure function over `(chat, note, team, section, nowMillis)` it's directly testable, and taking `nowMillis` as a parameter rather than reading the clock inside is what makes it deterministic.

**Two behaviour changes ride along, and both are improvements worth stating explicitly.**

**The title no longer serializes as the string `"null"`.** The old line was `serializedMap["title"] = "${chatHistory.title}".trim()` — a string template, so a null title produced the four-character literal `"null"`, which then got uploaded as the shared chat's title. The new `(chat.title ?: "").trim()` produces `""`. That's a real fix, but it changes what lands on the server for untitled chats, so it belongs in the description rather than being discovered later by whoever wonders why old shared chats are called "null".

**`createdDate` and `updatedDate` are now guaranteed equal.** The old code called `Date().time` twice, once per field, so the two could differ by a millisecond if the calls straddled a tick. Threading a single `nowMillis` through removes that.

Everything else is carried over unchanged — `_id`, `_rev`, `user`, `aiProvider`, the conversation query/response mapping, and all seven outer-map fields. Moving `serializeConversation` alongside the builder keeps the two halves of the format together.

**Nit:** `ChatSharePayload` lives in `model/` but is a serializer rather than a model — the package already holds `Achievement.serialize` and friends, so it's consistent, just worth noting it's a mapper. And `note = "${addNoteDialogBinding.editText.text}"` stays a string template at the call site, so it retains the same `"null"` hazard the title just shed; `editText.text` is effectively never null for an `EditText`, but the asymmetry is now visible.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16782, opened 2026-09-04
all: smoother notifications action failure logging (fixes #16764)

### LINKED ISSUE #16764: Replace the printStackTrace calls in MainApplication and NotificationActionReceiver with tagged logging
## What

Seven `e.printStackTrace()` calls write to stderr with no tag, no level, and no path into the app's diagnostics pipeline.

**Files:**
- `app/src/main/java/org/ole/planet/myplanet/MainApplication.kt` (4 calls)
- `app/src/main/java/org/ole/planet/myplanet/services/NotificationActionReceiver.kt` (3 calls)

## Why

Verified counts: MainApplication 4, NotificationActionReceiver 3. Each site has a meaningful surrounding operation that can name the failure, and each currently produces an untagged stderr dump that never reaches the diagnostics pipeline — so a failure in app startup or a notification action is effectively invisible in the field.

## The fix

Replace each with `Log.e(TAG, "<operation that failed>", e)`.

- Use each class's existing `TAG` where present; add a `private const val TAG` if the class has none.
- The message should name the operation that failed, not restate the exception.
- **Keep every catch clause's control flow identical** — same swallow/rethrow/return behavior. This is a logging change only.
- **Leave `MainApplication.handleUncaughtException`'s `persistCriticalLog` call untouched.**

## Acceptance

- `grep -rn printStackTrace` over these two files returns nothing.
- No control-flow change in any touched catch block.

## Notes

⚠️ Scope: `ui/sync/SyncActivity.kt` also has 6 `printStackTrace` calls and is **deliberately excluded** — that file is claimed by an open PR, and touching it here would collide. It will be filed separately once that PR lands. Please don't extend this change to it.

dogi:
Replaced 7 untagged printStackTrace calls across MainApplication and NotificationActionReceiver with tagged Log.e calls describing the operation that failed while preserving exact control flow.

Fixes #16764

---
*PR created automatically by Jules for task [11500500903127609958](https://jules.google.com/task/11500500903127609958) started by @dogi*

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved diagnostic logging for application errors and notification action failures.
  - Error and warning details are now recorded through Android logging, making issues easier to monitor and troubleshoot.


dogi:
**Review — rating: 82/100**

The substance is right and the messages are genuinely useful — each site says what failed rather than dumping an untagged trace to `System.err`, and "Failed to sweep pending logs" / "Failed to recover stuck operations" name subsystems you'd actually grep for. Tagging matters more than usual in `MainApplication`, where these fire during startup and get interleaved with everything else booting.

**One thing to fix: the new `TAG` duplicates the existing `LOG_TAG`.** The companion object already declares `private const val LOG_TAG = "MainApplication"` two lines above, and this adds `private const val TAG = "MainApplication"` with the identical value. Now the same class has two constants for one tag, and the next person to add a log has to guess which is canonical — or worse, changes one and leaves the other. Use `LOG_TAG` and drop the addition.

`NotificationActionReceiver` is clean by contrast — no existing constant, so a fresh `TAG` in a new companion is exactly right, and the three messages there distinguish the mark-as-read failure from the broadcast failure from the activity-start failure, which are three quite different problems that previously looked identical.

Logging in `handleUncaughtException` before `stackTraceToString()` is fine — the throwable is fully formed by then.

**Verdict:** mergeable once `TAG` collapses into `LOG_TAG`.

---

dogi:
To be precise about that last line: the duplicate constant is cosmetic, not blocking — it changes nothing at runtime. Labelling `ready`; fold the `TAG`/`LOG_TAG` dedup in whenever convenient rather than treating it as another round.

---

Okuro3499 (dismissed):
```
 app/src/main/java/org/ole/planet/myplanet/MainApplication.kt:112 — introduces private const val TAG = "MainApplication", duplicating the pre-existing LOG_TAG = "MainApplication"
    (line 111, already used at line 155 Log.w(LOG_TAG, ...)). Two constants with the identical string value in the same companion object is redundant and will confuse future        
    readers about which to use.  
```

dogi:
**Re-review — rating: 90/100** (was 82)

The one thing I raised is done. There's a single tag constant now — the follow-up commit collapsed `LOG_TAG` into `TAG`, so `warnBestEffortFailed` and the new `Log.e` calls share it instead of the file carrying two names for the same string. Direction doesn't matter; having one does.

The substance holds up on a re-read. Every `e.printStackTrace()` replaced here is in a `catch` that swallows the exception and continues — `getAndroidId`, `handleUncaughtException`, the pending-log sweep, `recoverStuckOperations`, and the three in `NotificationActionReceiver`. `printStackTrace` writes to `System.err`, which on Android is not captured by logcat in a way that survives to a bug report, so those failures were effectively invisible in the field. Each replacement carries a message that says what was being attempted rather than just re-printing the stack, which is the part that makes them useful when they show up in a log dump. `NotificationActionReceiver` gets its own `TAG` in a companion, matching the convention.

CI green 5/5.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16783, opened 2026-09-04
sync: smoother submissions server reachability providing (fixes #16769)

### LINKED ISSUE #16769: Inject ServerReachabilityProvider into SubmissionsUploader instead of the MainApplication static
## What

`SubmissionsUploader` reaches for a static on `MainApplication` for server reachability, when an injectable provider with the same contract already exists.

**File:** `app/src/main/java/org/ole/planet/myplanet/services/SubmissionsUploader.kt` (constructor, and the two checks at lines 35 and 49)

## Why

```kotlin
MainApplication.isServerReachable(mapping.primaryUrl)                        // line 35
mapping.alternativeUrl?.let { MainApplication.isServerReachable(it) } == true // line 49
```

`ServerReachabilityProvider` is an injectable `@Singleton` exposing the same `suspend fun isServerReachable(urlString: String): Boolean`, with an identical 30s TTL cache built on `TimeProvider` — while `MainApplication`'s companion carries a **duplicate** cache using raw `System.currentTimeMillis()`. Two caches, same purpose, one of them untestable.

The reference pattern is already in the codebase — `utils/SyncTimeLogger.kt` lines 101/104 use the injected provider for exactly this primary-then-alternative check.

## The fix

Constructor-inject `ServerReachabilityProvider`, replace both calls, and drop the `MainApplication` import.

**Preserve:**
- the primary-then-alternative semantics
- the 15-second `withTimeoutOrNull`

Add a test for the reachability short-circuit against a mocked provider.

## Acceptance

- `SubmissionsUploader` no longer imports `MainApplication`.
- Primary-reachable, primary-down/alternative-reachable, and both-down all behave as today.
- The 15s timeout is unchanged.

## Notes

⚠️ **Leave the `MainApplication` companion cache in place.** Other callers (`SyncActivity`, `ChatRepositoryImpl`) still use it — migrating them is a separate change, and removing the companion here would break them.

dogi:
Constructor-injected ServerReachabilityProvider into SubmissionsUploader, removed the MainApplication import, preserved primary-then-alternative reachability semantics and 15s timeout, and added unit tests in SubmissionsUploaderTest.kt.

Fixes #16769

---
*PR created automatically by Jules for task [2684215993559275678](https://jules.google.com/task/2684215993559275678) started by @dogi*


dogi:
**Review — rating: 90/100**

Replacing a static call into the `Application` class with an injected collaborator is the right move, and this one buys more than testability. `ServerReachabilityProvider` already exists on master as a `@Singleton` with constructor injection, and it maintains a `ConcurrentHashMap` reachability cache with a 30-second TTL. So `SubmissionsUploader` now shares that cache instead of reaching around it — two components probing the same server within the window make one request rather than two, which matters on exactly the flaky links this code path is written for.

The static `MainApplication.isServerReachable` was also the thing making this class untestable: a `@Singleton` service can be constructed in a unit test, but a static on `Application` can't be substituted without Robolectric or mocking a global. That's why the PR can add ~150 lines of test for the primary/alternative fallback logic — the `withTimeoutOrNull(15000)` branches and the `mapping.alternativeUrl?.let { … } == true` path are now reachable from a plain JVM test.

Both call sites are converted, the `MainApplication` import is dropped, and the surrounding timeout and try/catch structure is untouched, so the fallback semantics are unchanged.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16784, opened 2026-09-04
life: smoother personal dao sorting (fixes #16766)

### LINKED ISSUE #16766: Order personal resources in SQL
## What

The My Personals list has no defined order — it renders in whatever order SQLite happens to return.

**File:** `app/src/main/java/org/ole/planet/myplanet/data/room/dao/PersonalDao.kt` (`getByUserIdFlow`, line 25)

## Why

```kotlin
@Query("SELECT * FROM my_personal WHERE userId = :userId")
fun getByUserIdFlow(userId: String): Flow<List<Personal>>
```

No `ORDER BY`. `PersonalsRepositoryImpl.getPersonalResources` exposes that `Flow` directly to the UI, so list order is an implementation detail of SQLite's scan — it can change between SQLite versions, or after any rewrite of the table (a VACUUM, or the drop-and-resync this project does on schema bumps). Users would see their personals silently reshuffle.

## The fix

Add an explicit `ORDER BY` over existing columns. `Personal` has `date: Long` (line 30) and `title: String?` (line 28), so:

```sql
ORDER BY date DESC, title COLLATE NOCASE ASC
```

is a reasonable pair — newest first, with a stable alphabetical tie-break.

No `AppDatabase` version bump: adding `ORDER BY` to a `@Query` changes no table. (A bump under `fallbackToDestructiveMigration` would drop unsynced local writes.)

## Acceptance

- A test asserting through the repository `Flow` that newer entries sort first.
- Two entries with the same `date` sort by title, case-insensitively.

## Notes

dogi:
Order personal resources in SQL query with tests verifying repository Flow output order.

Fixes #16766

---
*PR created automatically by Jules for task [4547644843512412062](https://jules.google.com/task/4547644843512412062) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 7052
-        versionName = "0.70.52"
+        versionCode = 7053
+        versionName = "0.70.53"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/data/room/dao/PersonalDao.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/data/room/dao/PersonalDao.kt
@@ -21,7 +21,7 @@ interface PersonalDao {
     @Insert(onConflict = OnConflictStrategy.REPLACE)
     suspend fun insert(item: Personal)
 
-    @Query("SELECT * FROM my_personal WHERE userId = :userId")
+    @Query("SELECT * FROM my_personal WHERE userId = :userId ORDER BY date DESC, title COLLATE NOCASE ASC")
     fun getByUserIdFlow(userId: String): Flow<List<Personal>>
 
     @Query("SELECT * FROM my_personal WHERE userId = :userId AND isUploaded = 0")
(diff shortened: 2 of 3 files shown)


dogi:
**Review — rating: 84/100**

I checked whether this was replacing a Kotlin-side sort or adding ordering that didn't exist, because the two cases deserve different framing — there's no `sortedBy`/`sortedByDescending` anywhere in `PersonalsRepositoryImpl` or `ui/personals/`, so the list was previously rendered in whatever order SQLite happened to return. This makes it deterministic rather than removing duplicated work, which is still worth doing: unordered result sets are stable enough in testing to look intentional and then reshuffle after a compaction or a schema change.

`date DESC` as the primary key is the right default for a personal-resources list, and `COLLATE NOCASE` on the title tiebreak avoids the ASCII-ordering surprise where `Zebra` sorts before `apple`.

Two things to be aware of:

- **`title` is nullable** (`Personal.kt:28`), and SQLite sorts `NULL` first in `ASC`. So untitled personals rise to the top of each date group. Harmless if titles are effectively always set; if not, `title IS NULL, title COLLATE NOCASE ASC` puts them last instead.
- **This is a user-visible reordering** with no test at the UI level — the list will come back in a different order than users currently see. Better order, but a changed one, and the DAO test won't catch a screen that assumed the old arrangement.

The PR title says "queries" plural, but only `getByUserIdFlow` gains an `ORDER BY`; the other selects in this DAO are still unordered. Worth deciding whether that's deliberate scoping or an oversight.

**Note on #16770:** it also edits `PersonalDao.kt`, at a different query. The only genuine conflict between you is an add/add on `PersonalDaoTest.kt` — you each created that file — so the production changes are independent and both can land.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16791, opened 2026-09-04
all: smoother notifications repository counting (fixes #16778)

### LINKED ISSUE #16778: Deduplicate the two count-style notification upserts
## What

Two near-identical read-modify-upsert routines that have already drifted apart once.

**File:** `app/src/main/java/org/ole/planet/myplanet/repository/NotificationsRepositoryImpl.kt` (`updateResourceNotification` line 51, `updateStorageNotification` line 82)

## Why

Both build a synthetic id, call `notificationDao.getById`, detect whether the numeric value changed, mark unread and reset `createdAt` when it did, else construct a new `AppNotification`, upsert, and delete when the value is healthy.

The only differences are the id suffix, the type, the `relatedId`, how the previous value is parsed, and the message format:

```kotlin
val previousCount  = existingNotification?.message?.toIntOrNull() ?: 0                      // line 58
val previousPercent = existingNotification?.message?.replace("%", "")?.toIntOrNull()        // line 89 — no default
```

That difference — one defaults to `0`, the other to `null` — is the drift. It changes first-run behavior between the two paths, and nothing says whether that's intentional.

## The fix

Extract one private suspend helper taking: the id suffix, type, `relatedId`, a **parse lambda**, a **format lambda**, the value, and a healthy flag. Route both methods through it.

⚠️ **Preserve each method's current semantics exactly** — including the parse difference (`?: 0` vs no default) and the threshold difference. This is a deduplication, not a reconciliation: if you think the drift is a bug, say so in the PR rather than silently unifying it, because that would change notification behavior for one of the two paths.

Keep both public signatures.

## Acceptance

- Both methods behave identically to today, including first-run with no existing notification.
- Add a test for the "value unchanged keeps it read" path if one doesn't exist.

## Notes

- `NotificationsRepositoryImpl` also has an open PR against a *different* method (`getTeamNotifications`). Different region of the file, but if you hit a conflict, that's why.

dogi:
Deduplicated `updateResourceNotification` and `updateStorageNotification` in `NotificationsRepositoryImpl.kt` into a shared private helper `updateCountNotification`. Preserved each method's exact existing semantics (including `parsePrevious` default differences) and added unit tests covering value-unchanged, value-changed, first-run, deletion, and null userId behavior.

Fixes #16778

---
*PR created automatically by Jules for task [14126154801635436748](https://jules.google.com/task/14126154801635436748) started by @dogi*


dogi:
**Review — rating: 86/100**

The two methods were near-identical fifty-line blocks differing in four details, so folding them into one parameterised helper is worth doing. What makes this a good extraction rather than a risky one is that it **preserves an asymmetry between them that a tidier refactor would have flattened**.

The resource path parsed its previous value as `existing?.message?.toIntOrNull() ?: 0` — defaulting to `0`. The storage path parsed `existing?.message?.replace("%", "")?.toIntOrNull()` and left it **nullable**. That difference matters at the `valueChanged` comparison: for storage, a missing or unparseable previous message yields `null`, which never equals the incoming percent, so the notification is always re-marked unread; for resources it yields `0`, which *does* equal an incoming count of zero. Passing `parsePrevious` as a lambda per call site keeps both behaviours exactly as they were. Collapsing them to a single shared parser would have silently changed when the storage warning re-alerts.

The `isHealthy` inversion checks out in both directions: `resourceCount <= 0` negates to the old `resourceCount > 0`, and `availablePercent > STORAGE_WARNING_AVAILABLE_PERCENT` negates to the old `<=`. Ids, types and `relatedId` values are unchanged per call site, including the resource case where `relatedId` mirrors the count.

**Nit:** eight parameters including two lambdas is a lot of surface, and `isHealthy` leaves the threshold comparison at the call site while `formatMessage` is passed lazily — a small inconsistency in what the helper owns versus what the caller decides. A small sealed config type per notification kind would read better if a third count-style notification ever arrives. Also `isHealthy` is a slightly odd name for "there is nothing to notify about" in the resource case.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16792, opened 2026-09-04
life: smoother repository visibility view modelling (fixes #16776)

### LINKED ISSUE #16776: Stop the two redundant My Life re-queries in LifeRepositoryImpl
## What

Every My Life visibility toggle runs the full list query twice after the write, and the seeding path re-queries for rows it just built.

**Files:**
- `app/src/main/java/org/ole/planet/myplanet/repository/LifeRepository.kt`
- `app/src/main/java/org/ole/planet/myplanet/repository/LifeRepositoryImpl.kt` (`updateVisibility` line 33, `getMyLifeByUserId` line 79, `seedMyLifeIfEmpty` line 138)
- `app/src/main/java/org/ole/planet/myplanet/ui/life/LifeViewModel.kt` (`updateVisibility` line 46)

## Why

Two independent redundancies:

1. **Double query per toggle.** `updateVisibility` updates the row, then calls `getMyLifeByUserId` purely to refresh the prefs cache — and throws the list away. `LifeViewModel.updateVisibility` then calls `loadMyLifeList()`, querying `my_life` a *second* time. So one toggle = write + two full queries.
2. **Seed then re-query.** `getMyLifeByUserId` queries, calls `seedMyLifeIfEmpty`, then re-queries unconditionally — even though `seedMyLifeIfEmpty` has already built the complete inserted list.

## The fix

1. Change `updateVisibility` to return `List<MyLife>` — the list it already fetched — and have the ViewModel assign it to `_myLifeList.value` instead of calling `loadMyLifeList()`.
2. Refactor the seeding path to return the newly inserted rows when it wins the mutex, and use those instead of the second `getByUserId`.

**Preserve:** deduplication, weight ordering, null-user normalization, and the mutex. The mutex matters — the loser of the race must still get a correct list, so only the winner can skip the re-query.

## Acceptance

- A visibility toggle issues one write and one query, and the UI shows the same list as today.
- First-run seeding issues one query, not two, and returns the seeded rows in weight order.
- Concurrent callers of the seeding path both get a complete, correctly ordered list.

## Notes

dogi:
Eliminated redundant My Life database queries in LifeRepositoryImpl and LifeViewModel.
- Updated `LifeRepository.updateVisibility` to return `List<MyLife>` and updated `LifeViewModel.updateVisibility` to assign the returned list directly to `_myLifeList.value` instead of calling `loadMyLifeList()`.
- Refactored `LifeRepositoryImpl.seedMyLifeIfEmpty` to return the newly inserted deduplicated and weight-ordered list when winning the mutex, allowing `getMyLifeByUserId` and `getMyLifeForDashboard` to use the seeded list directly without an extra DB re-query.
- Updated unit tests in `LifeRepositoryTest`, `LifeRepositoryImplTest`, and `LifeViewModelTest` to verify single-query behavior and mutex safety.

Fixes #16776

---
*PR created automatically by Jules for task [5684228213959099837](https://jules.google.com/task/5684228213959099837) started by @dogi*


dogi:
**Review — rating: 86/100**

Returning the computed list instead of recomputing it is the right shape, and it fixes a sequencing problem as well as saving queries.

`updateVisibility` already had `updatedLives` in hand — it queried them to refresh the cache — and then threw them away so the ViewModel could go back to the database for the same rows. Worse, `loadMyLifeList()` is itself `fun … { viewModelScope.launch { … } }`, so the old `updateVisibility` fired a *second, detached* coroutine to refresh. Two quick toggles produced two independent reloads that could resolve out of order and leave the UI showing the earlier result. Assigning `_myLifeList.value = updatedList` inside the same coroutine that performed the write ties each toggle to its own consistent snapshot.

The `seedMyLifeIfEmpty` change is careful in the way that matters: it returns the newly seeded rows only when it actually seeded, and `emptyList()` when rows already existed, so callers can use `.ifEmpty { getMyLifeByUserId(…) }` and still hit the database on the already-seeded path. Returning the seeded rows as `newItems.distinctBy { it.dedupKey() }.sortedBy { it.weight }` mirrors exactly what `getMyLifeByUserId` applies, so the shape of the list is the same whichever branch produced it. Keeping all of that inside `seedMutex.withLock` preserves the existing serialization.

**One divergence worth knowing:** `loadMyLifeList()` calls `getMyLifeByUserId(userId, MyLife.defaultItems(...))`, but `updateVisibility` computes its result with `getMyLifeByUserId(effectiveUserId)` — no `defaultItems`. Since `getMyLifeByUserId` returns early when `defaultItems.isEmpty()`, the seeding branch is unreachable from the update path. In practice that can't bite (you just toggled a row, so rows exist), but the two paths that now both feed `_myLifeList` no longer resolve the user or the seed fallback identically, and that's the kind of asymmetry that turns into a bug the next time someone changes seeding.

**Verdict:** mergeable.

---

Okuro3499 (approved):

Okuro3499 (approved):

PR #16793, opened 2026-09-04
life: smoother health repository examination view modelling (fixes #16774)

### LINKED ISSUE #16774: Move the MyHealth decrypt-and-parse out of HealthExaminationViewModel
## What

`HealthExaminationViewModel` decrypts and Gson-parses health data inline, duplicating logic the repository already owns.

**Files:**
- `app/src/main/java/org/ole/planet/myplanet/repository/HealthRepository.kt`
- `app/src/main/java/org/ole/planet/myplanet/repository/HealthRepositoryImpl.kt`
- `app/src/main/java/org/ole/planet/myplanet/ui/health/HealthExaminationViewModel.kt` (`loadData`, line 73)

## Why

```kotlin
// HealthExaminationViewModel.kt:73
health = JsonUtils.gson.fromJson(decrypt(pojo.data, user?.key, user?.iv), MyHealth::class.java)
```

inside a try/catch that swallows to null.

`HealthRepositoryImpl.decodeHealth` (line 164) already performs the identical `AndroidDecrypter.decrypt` + `gson.fromJson` with the same null-on-failure contract — but it's private, used at lines 178 and 216. So decryption of health records lives in two places, one of them a ViewModel.

## The fix

1. Add `suspend fun getDecryptedHealth(pojo: HealthExamination?, user: UserEntity?): MyHealth?` to `HealthRepository`.
2. Implement it on top of the existing `decodeHealth` — don't reimplement the decrypt.
3. Replace the ViewModel block with that call, dropping the `AndroidDecrypter` / `JsonUtils` imports.
4. Keep the `initHealth()` fallback exactly as it is.

## Acceptance

Tests: encrypted fixture → parsed `MyHealth`; empty `data` → null; garbage → null (no throw).

## Notes

- The ViewModel's swallow-to-null behavior must be preserved — a decrypt failure still yields the `initHealth()` fallback, not a crash.

dogi:
Refactored MyHealth decrypt-and-parse logic out of HealthExaminationViewModel by adding getDecryptedHealth to HealthRepository and delegating to HealthRepositoryImpl's decodeHealth. Removed unused AndroidDecrypter and JsonUtils imports in HealthExaminationViewModel. Added unit tests in HealthRepositoryImplTest and updated HealthExaminationViewModelTest.

Fixes #16774

---
*PR created automatically by Jules for task [9077286432254507979](https://jules.google.com/task/9077286432254507979) started by @dogi*


dogi:
**Review — rating: 90/100**

This is a straight de-duplication with a security-posture win, and I checked the one place it could have changed behaviour.

The inline ViewModel block was a copy of the repository's existing private `decodeHealth` (line 164) — same null/empty guard on `pojo.data`, same `AndroidDecrypter.decrypt(data, user?.key, user?.iv)`, same `gson.fromJson(…, MyHealth::class.java)`, same catch returning null. So exposing `decodeHealth` through `getDecryptedHealth` collapses two copies of the decryption path into one.

**The subtle difference is benign.** The old code only assigned `health` inside the `try`, so a decrypt failure left whatever was there; the new code assigns unconditionally, so a failure writes null. That would matter if `health` could already hold a value — but it's declared `var health: MyHealth? = null` at the top of `loadData` and nothing between the declaration and this line touches it. Both versions therefore arrive at the following `if (health == null) health = healthRepository.initHealth()` identically.

The layering gain is the real point: `HealthExaminationViewModel` no longer imports `AndroidDecrypter` and no longer touches `user?.key` / `user?.iv`. Encryption keys for health records now stay behind the repository rather than being handled in a UI-layer class, which is where `CLAUDE.md` puts that kind of concern. It also removes one of the two `printStackTrace` sites, leaving the single one inside `decodeHealth`.

⚠️ **No build or test verdict on this head** — only `myPlanet labels` ran; neither `build` nor `test` fired, so the added repository and ViewModel tests haven't executed.

---

Okuro3499 (approved):

PR #16795, opened 2026-09-04
enterprises: smoother finances reports attachment caching (fixes #16777)

### LINKED ISSUE #16777: Cache the enterprise attachment existence check out of the row-bind path
## What

Two enterprise adapters do a disk stat on the UI thread for every visible row, on every recycle.

**Files:**
- `app/src/main/java/org/ole/planet/myplanet/ui/enterprises/EnterprisesFinancesAdapter.kt` (`bindFinanceImage`, lines 60-61)
- `app/src/main/java/org/ole/planet/myplanet/ui/enterprises/EnterprisesReportsAdapter.kt` (`bindReportImage`, lines 99-100)

## Why

```kotlin
val imageFile = MyTeam.getAttachmentFile(context, item.id, item.imageName)
if (imageFile != null && imageFile.exists()) { ... }
```

`imageFile.exists()` is a filesystem stat, called directly inside `onBindViewHolder`, in two lists a team treasurer scrolls constantly. The value only changes when the underlying list changes — and `ListAdapter` already gives you `onCurrentListChanged` to invalidate on.

## The fix

- Add a per-adapter map keyed on the file path, and use `getOrPut { imageFile.exists() }`.
- Keep the null-file early return and the `View.GONE` branch identical.
- Override `onCurrentListChanged` to clear the map, so a re-submitted list after an upload or delete re-stats.
- Leave the `onViewRecycled` Glide `clear()` calls alone.
- While here: move the else-branch `setOnClickListener(null)` (Finances line 73, Reports line 112) so it isn't re-cleared on every bind — the recycle path at Finances:56 / Reports:93-95 already handles it.

## Acceptance

- No `exists()` call on a second bind of the same row.
- Uploading or deleting an attachment and re-submitting the list shows the new state (this is what `onCurrentListChanged` buys — please test it).
- `View.GONE` behavior for a missing file unchanged.

## Notes

⚠️ The reports adapter **also** computes its row totals inline in `onBindViewHolder`. That is a separate extraction task queued behind this one — please don't take it here, so the two changes don't collide in the same file.

dogi:
Caches `imageFile.exists()` checks in `EnterprisesFinancesAdapter` and `EnterprisesReportsAdapter` to avoid disk stats on every row bind during scrolling. Clears the cache on `onCurrentListChanged` to properly handle list updates. Removed redundant click listener clear calls in image binding functions. Added unit tests for both adapters.

Fixes #16777

---
*PR created automatically by Jules for task [14621057516108396353](https://jules.google.com/task/14621057516108396353) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 7050
-        versionName = "0.70.50"
+        versionCode = 7051
+        versionName = "0.70.51"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/ui/enterprises/EnterprisesFinancesAdapter.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/enterprises/EnterprisesFinancesAdapter.kt
@@ -27,7 +27,8 @@ class EnterprisesFinancesAdapter(
         areContentsTheSame = { oldItem, newItem -> oldItem == newItem }
     )
 ) {
-    private val attachmentExistsCache = HashMap<String, Boolean>()
+    private val attachmentExistsCache = HashMap<String, Pair<Boolean, Long>>()
+    private val cacheTtlMs = 5000L
 
     override fun onCurrentListChanged(
         previousList: MutableList<Transaction>,
@@ -67,7 +68,21 @@ class EnterprisesFinancesAdapter(
 
     private fun bindFinanceImage(binding: RowFinanceBinding, item: Transaction) {
         val imageFile = MyTeam.getAttachmentFile(context, item.id, item.imageName)
-        if (imageFile != null && attachmentExistsCache.getOrPut(imageFile.absolutePath) { imageFile.exists() }) {
+        val now = System.currentTimeMillis()
+        val exists = if (imageFile != null) {
+            val cached = attachmentExistsCache[imageFile.absolutePath]
+            if (cached != null && now - cached.second < cacheTtlMs) {
+                cached.first
+            } else {
+                val freshExists = imageFile.exists()
+                attachmentExistsCache[imageFile.absolutePath] = Pair(freshExists, now)
+                freshExists
+            }
+        } else {
+            false
+        }
+
+        if (imageFile != null && exists) {
             binding.financeImage.visibility = View.VISIBLE
             Glide.with(context)
                 .load(imageFile)
@@ -79,6 +94,7 @@ class EnterprisesFinancesAdapter(
             }
         } else {
             binding.financeImage.visibility = View.GONE
+            binding.financeImage.setOnClickListener(null)
         }
     }
 
--- a/app/src/main/java/org/ole/planet/myplanet/ui/enterprises/EnterprisesFinancesAdapter.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/enterprises/EnterprisesFinancesAdapter.kt
@@ -27,6 +27,15 @@ class EnterprisesFinancesAdapter(
         areContentsTheSame = { oldItem, newItem -> oldItem == newItem }
     )
 ) {
+    private val attachmentExistsCache = HashMap<String, Boolean>()
+
+    override fun onCurrentListChanged(
+        previousList: MutableList<Transaction>,
+        currentList: MutableList<Transaction>
+    ) {
+        super.onCurrentListChanged(previousList, currentList)
+        attachmentExistsCache.clear()
+    }
 
     override fun onCreateViewHolder(parent: ViewGroup, viewType: Int): FinanceViewHolder {
         val binding = RowFinanceBinding.inflate(LayoutInflater.from(parent.context), parent, false)
@@ -58,7 +67,7 @@ class EnterprisesFinancesAdapter(
 
     private fun bindFinanceImage(binding: RowFinanceBinding, item: Transaction) {
         val imageFile = MyTeam.getAttachmentFile(context, item.id, item.imageName)
-        if (imageFile != null && imageFile.exists()) {
+        if (imageFile != null && attachmentExistsCache.getOrPut(imageFile.absolutePath) { imageFile.exists() }) {
             binding.financeImage.visibility = View.VISIBLE
             Glide.with(context)
                 .load(imageFile)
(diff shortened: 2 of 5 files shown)


dogi:
**Review — rating: 76/100**

The core idea is sound. `imageFile.exists()` is a filesystem stat running inside `onBindViewHolder`, so every row bound during a scroll hit the disk on the main thread, repeatedly for rows the user scrolls past more than once. Keying a memo on `imageFile.absolutePath` and clearing it in `onCurrentListChanged` is a reasonable shape, and doing it identically in both adapters keeps them consistent.

Two things stop me marking this ready.

**The cache outlives the facts it caches.** `onCurrentListChanged` only fires when a new list is submitted, but attachment files appear and disappear independently of the list: a sync or download can write the image while the same list is still bound. Before, scrolling away and back re-stat'd the file and the image appeared; now the `false` is pinned until something happens to resubmit the list, so a freshly downloaded attachment stays invisible. The reverse holds for a deleted file showing a broken slot. That's a real behavioural regression, not just a staleness nit. Clearing on resume, or keying the entry with a short TTL, would keep the scroll win without pinning a stale negative for the life of the list.

**`setOnClickListener(null)` was removed from the else branch in both files, and the PR doesn't say why.** This looks unrelated to caching:

```kotlin
} else {
    binding.financeImage.visibility = View.GONE
-   binding.financeImage.setOnClickListener(null)
}
```

In fairness, the practical impact is small — the view is `GONE`, so it takes no touches, and the if-branch installs a fresh listener whenever it becomes visible again. But it's recycling hygiene that was deliberately there, its removal leaves a recycled holder holding a listener that closes over the *previous* row's file, and it's a change in a direction the PR isn't about. If it was accidental, put it back; if it was deliberate, the reasoning belongs in the description.

**Note on #16858:** it also edits `EnterprisesReportsAdapter`, but in `onBindViewHolder` rather than `bindReportImage`, and the only real conflict between you is in `EnterprisesReportsAdapterTest.kt`. The production changes are independent and both can land.

**Verdict:** needs another round — restore (or justify) the listener cleanup, and tighten invalidation so a newly arrived attachment isn't hidden until the list is resubmitted.

---

dogi:
**Re-review — rating: 84/100** (was 76)

Both blocking points are resolved.

**The listener cleanup is back.** `setOnClickListener(null)` is present in the else branch of both adapters again (`EnterprisesFinancesAdapter:97`, `EnterprisesReportsAdapter:136`), so a recycled holder no longer carries a listener closed over the previous row's file.

**The cache can no longer pin a stale negative.** Entries are now `Pair(exists, timestamp)` with a 5-second TTL alongside the existing `onCurrentListChanged` clear, so an attachment that lands while the same list is bound shows up on the next bind after the window rather than staying invisible until something resubmits the list. That was the real regression in the first revision and 5s is a sensible choice — long enough to cover a fling through a list, short enough that nobody notices the delay.

The underlying win is intact: `imageFile.exists()` is a filesystem stat that was running inside `onBindViewHolder` on every row, repeatedly for rows scrolled past more than once.

**Three nits, none blocking:**

- **The TTL reads the clock directly.** `System.currentTimeMillis()` here is exactly what #16890 just replaced in `NotificationsAdapter` with an injected `TimeProvider`, for the same reason: it makes the time-dependent behaviour — in this case the 5s expiry — impossible to test. The project has `TimeProvider` and `TestTimeProvider` already; threading one in would let a test prove the entry actually expires.
- **The cache block is duplicated verbatim.** The same ~14 lines of lookup / TTL check / write-back now appear in both adapters. It was one line in each before, so the copy-paste cost has grown; a small `AttachmentExistsCache` with a `get(file): Boolean` would give both adapters one implementation and one place to change the TTL.
- **`cacheTtlMs` is an instance `val`** rather than a `companion object` constant, so each adapter instance carries its own copy of a fixed number.

**Note:** the only conflict with #16858 remains the add/add on `EnterprisesReportsAdapterTest.kt` — the production changes are still in different methods, so both land with a small test merge.

**Verdict:** mergeable.

---

Okuro3499 (dismissed):
```
- File: EnterprisesFinancesAdapter.kt / EnterprisesReportsAdapter.kt, bindFinanceImage/bindReportImage                                                                             
Problem: Direct System.currentTimeMillis() usage instead of the project's TimeProvider abstraction, which exists precisely to keep time-based logic testable (see                
utils/TimeProvider.kt doc comment, mirrored from DispatcherProvider).                                                                                                                                                                                                                                                      
- File: EnterprisesFinancesAdapterTest.kt:143, EnterprisesReportsAdapterTest.kt (TTL test)                                                                                         
Problem: Thread.sleep(5100L) × 2 adds >10s of real wall-clock time to the unit test suite and is a source of flakiness under CI load — directly conflicts with the project's     
documented shard-for-wall-clock-speed effort.                                                                                                                                    
```

dogi:
**Re-review — rating: 86/100** (was 84)

The `TimeProvider` injection is a real improvement over the previous round's hard-coded clock: the TTL is now testable with `TestTimeProvider` instead of depending on wall-clock timing, and the tests take advantage of it. Defaulting the parameter to `SystemTimeProvider()` is a pragmatic call given adapters aren't Hilt-injectable — worth knowing that it means production paths construct their own provider rather than using the one `TimeModule` supplies, even though the fragments creating these adapters do have it injected and could pass it through.

The caching itself is sound, and notably more careful than the same idea elsewhere in this queue: a 5-second TTL bounds staleness, and `onCurrentListChanged` clears the whole map so a list refresh always re-stats. That combination means a newly-arrived attachment shows up within a list update or five seconds, whichever comes first — which is the right trade for `File.exists()` in a bind path.

**The one thing I'd still change: the ~15-line cache block is duplicated verbatim across both adapters.**

```kotlin
val cached = attachmentExistsCache[imageFile.absolutePath]
if (cached != null && now - cached.second < cacheTtlMs) { cached.first } else { ... }
```

Identical in `EnterprisesFinancesAdapter` and `EnterprisesReportsAdapter`, down to the `cacheTtlMs = 5000L` field and the `onCurrentListChanged` override. A small `AttachmentExistenceCache(timeProvider)` with a single `exists(file: File): Boolean` and a `clear()` would collapse both call sites to one line each, shrink the PR, and remove the risk of the two copies drifting when someone tunes the TTL on one of them. Both files already import `TimeProvider`/`SystemTimeProvider`, so the extraction is nearly mechanical.

Trivially: `cacheTtlMs` is a per-instance `val` where it wants to be a `const` in a companion.

CI green 5/5.

**Verdict:** mergeable — the duplication is worth collapsing, but it's a shape issue, not a correctness one.

---

Okuro3499 (approved):

PR #16796, opened 2026-09-04
all: less diagnostics repository context is more (fixes #16788)

### LINKED ISSUE #16788: Remove @ApplicationContext from DiagnosticsRepositoryImpl
## What

An otherwise DAO/prefs repository carries an Android `Context` solely to read the app version string.

**File:** `app/src/main/java/org/ole/planet/myplanet/repository/DiagnosticsRepositoryImpl.kt`

## Why

The constructor takes `@ApplicationContext private val context: Context` (line 15) only so `saveLogToRoom` (line 65) and `saveLogsToRoom` (line 83) can call `VersionUtils.getVersionName(context)`.

## The fix

Drop the `Context` parameter and resolve the version from `BuildConfig.VERSION_NAME`.

That value carries the flavor's `versionNameSuffix`, so the `-lite` build still reports the same string it does today — this is why the substitution is safe rather than merely convenient.

Keep `ApkLogDao`, `UserRepository` and `SharedPrefManager` usage unchanged. No `RepositoryModule` change is needed.

## Acceptance

- Logged version string is identical for both `default` and `lite` flavors.
- `DiagnosticsRepositoryImpl` no longer imports `dagger.hilt.android.qualifiers.ApplicationContext` or `android.content.Context`.

## Notes

dogi:
Removed @ApplicationContext private val context: Context from DiagnosticsRepositoryImpl constructor and replaced version resolving calls with BuildConfig.VERSION_NAME. Updated unit tests in DiagnosticsRepositoryImplTest to match.

Fixes #16788

---
*PR created automatically by Jules for task [8352914477287668177](https://jules.google.com/task/8352914477287668177) started by @dogi*


dogi:
**Review — rating: 90/100**

`BuildConfig.VERSION_NAME` is the better source here, not just the one that lets the `Context` go. `VersionUtils.getVersionName(context)` does a `packageManager.getPackageInfo` round trip, returns `String?`, and falls back to `""` on `NameNotFoundException` — so callers were paying an IPC to look up a value the compiler already knows, and getting a nullable back that could silently become an empty version string in the uploaded log. `BuildConfig.VERSION_NAME` is a non-null compile-time constant with no failure mode.

The values agree: `versionNameSuffix` (the `-lite` suffix on that flavor) is applied to the merged manifest as well as to `BuildConfig`, so the package manager and the constant report the same string for a given build.

With both call sites converted, the `@ApplicationContext` injection and the `VersionUtils` import drop out cleanly, leaving the repository with only the DAO and two collaborators — which also makes it constructible in a plain JVM test.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16800, opened 2026-09-04
login: less storage category context is more (fixes #16789)

### LINKED ISSUE #16789: Remove @ApplicationContext from StorageCategoryViewModel
## What

A ViewModel holds an Android `Context` field purely to compute a path string its only caller could pass in.

**Files:**
- `app/src/main/java/org/ole/planet/myplanet/ui/settings/StorageCategoryViewModel.kt` (line 36)
- `app/src/main/java/org/ole/planet/myplanet/ui/settings/StorageCategoryDetailFragment.kt`

## Why

The ViewModel holds `@ApplicationContext private val context: Context` (line 36) only to call `FileUtils.getOlePath(context)` — twice, in `loadResources` (line 52) and again inside `deleteItems` (line 97).

Its only caller, `StorageCategoryDetailFragment`, already has a `Context`.

## The fix

- Drop the `Context` constructor parameter.
- Take `olePath: String` on `loadResources` / `deleteSelected` / `deleteAll` / `deleteItems`.
- Compute it **once** in the fragment's `onViewCreated` and pass it through.

## Acceptance

- Storage category listing and all three delete paths behave identically.
- `StorageCategoryViewModel` holds no Android `Context` field and imports neither `ApplicationContext` nor `Context`.

## Notes

dogi:
Refactored StorageCategoryViewModel to remove `@ApplicationContext private val context: Context` from constructor and imports. Updated ViewModel methods (`loadResources`, `deleteSelected`, `deleteAll`) to accept `olePath: String`. Computed `olePath` in `StorageCategoryDetailFragment.onViewCreated()` and passed it into the ViewModel calls. Updated `StorageCategoryViewModelTest` accordingly.

Fixes #16789

---
*PR created automatically by Jules for task [2153726120057359464](https://jules.google.com/task/2153726120057359464) started by @dogi*


dogi:
**Review — rating: 84/100**

I was initially unsure this was an improvement rather than a lateral move — it trades a benign `@ApplicationContext` in the ViewModel for a storage-path concern in the Fragment and four changed signatures. What settles it is that resolving the path in the UI layer is already this codebase's convention: `EditAchievementFragment:237,469`, `AchievementFragment:214` and `StorageBreakdownFragment:209` all call `FileUtils.getOlePath(requireContext())` directly. So the Fragment computing `olePath` and handing it to the ViewModel matches the sibling screens rather than inventing a new arrangement, and `ResourcesRepository.getOfflineResourceItems(oleDirPath, …)` already took the path as a parameter — the ViewModel was the only layer manufacturing it.

The payoff is a ViewModel with no Android dependency at all: `resourcesRepository` and `dispatcherProvider` only, constructible in a plain JVM test with no Robolectric or Hilt context. Given `deleteItems` recomputed the path on every deletion, hoisting it also removes repeated work.

**Nit:** `olePath` is now captured once in `onViewCreated` and held by the click listeners for the lifetime of the view, where it used to be recomputed at each call. If external storage is remounted or the resolved path otherwise changes while the sheet is open, the buttons act on a stale path — a narrow window on a bottom sheet, but previously it couldn't happen.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16801, opened 2026-09-04
courses: smoother progress repository sync inserting (fixes #16786)

### LINKED ISSUE #16786: Extract the course-progress sync keys in one pass per document
## What

`insertCourseProgressFromSync` walks every incoming document four times to collect query keys, then re-parses the same four fields again while building records.

**File:** `app/src/main/java/org/ole/planet/myplanet/repository/ProgressRepositoryImpl.kt` (`insertCourseProgressFromSync`, line 269)

## Why

```kotlin
val docIds   = docs.map { JsonUtils.getString("_id", it) }.filter { it.isNotEmpty() }.distinct()      // 270
val courseIds = docs.map { JsonUtils.getString("courseId", it) }.filter { it.isNotEmpty() }.distinct() // 271
val userIds  = docs.map { JsonUtils.getString("userId", it) }.filter { it.isNotEmpty() }.distinct()   // 272
val stepNums = docs.map { JsonUtils.getInt("stepNum", it) }.distinct()                                // 273
```

Four complete passes, four intermediate lists — and the same four fields are parsed *again* when constructing each record. That's eight JSON lookups per document on a sync hot path.

## The fix

- Introduce a private file-local value type holding the four parsed keys.
- Build it in **one pass** over `docs`, preserving input order and the existing treatment of blank ids and duplicate query keys.
- Derive the DAO query arguments from those rows.
- Reuse each row's values when resolving `existingProgress` and `localRecord`.
- Keep `courseProgressFromJson` (line 242) as the single place that hydrates the entity.

## Acceptance

- Same records written, in the same order, for the same input.
- Blank ids still filtered out of the query arguments; duplicates still collapsed.
- No behavior change for an empty `docs` list.

## Notes

- A narrower variant of this finding proposed only hoisting `docIds`. The single-pass extraction supersedes it — please take the full version.

dogi:
Refactors `insertCourseProgressFromSync` in `ProgressRepositoryImpl.kt` to extract course-progress sync keys (`_id`, `courseId`, `userId`, `stepNum`) in a single pass into a `CourseProgressSyncKeys` holder. Eliminates redundant JSON field parsing and multi-pass iterations across documents while maintaining exact behavioral equivalence.

Fixes #16786

---
*PR created automatically by Jules for task [6607329154546898497](https://jules.google.com/task/6607329154546898497) started by @dogi*


dogi:
**Review — rating: 88/100**

The old code walked `docs` five times and pulled every field out twice — four `map` passes to build the id lists, then a fifth `map` that re-extracted `_id`, `courseId`, `userId` and `stepNum` from the same `JsonObject`s to build the rows. Extracting once into `CourseProgressSyncKeys` and projecting from that halves the `JsonUtils` lookups per document, on a path that runs over a full course-progress table during sync.

The dedup rewrite is equivalent, which is the part worth checking: `distinct()` preserves first-occurrence order and so does `LinkedHashSet`, so `mapNotNullTo(LinkedHashSet()) { … }.toList()` produces the same list in the same order as `map { … }.filter { it.isNotEmpty() }.distinct()`. It also skips the intermediate lists the old chain allocated. Ordering doesn't actually matter downstream — `getByIds(docIds)` is immediately `associateBy { it.id }` — but keeping it stable means the query parameters don't reshuffle between runs.

The `takeIf { it.isNotEmpty() }` guards reproduce the old `filter` exactly, and `stepNums` keeps every value including zero, matching the original which had no filter on it.

Holding the source `doc` in the key object so the final map can pass `keys.doc` to `courseProgressFromJson` avoids re-pairing by index — the mapping stays positional and can't drift.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16802, opened 2026-09-04
sync: smoother server url mapping (fixes #16787)

### LINKED ISSUE #16787: Clean up ServerUrlMapper: drop the empty also block and log the malformed-URL failure
## What

An `also { }` block with an empty body, and a swallowed malformed-URL error on the server-failover path.

**File:** `app/src/main/java/org/ole/planet/myplanet/services/sync/ServerUrlMapper.kt` (`processUrl` line 45, `extractBaseUrl` line 31)

## Why

```kotlin
serverMappings[baseUrl].also {    // line 48 — empty body, leftover from a removed log line
}
```

It allocates a lambda for nothing, and the result is then bound to `val result = UrlMapping(...)` and returned on the next line.

Separately, `extractBaseUrl` swallows malformed-URL errors with `e.printStackTrace()` (line 40).

That second one matters more than it looks: this class is on the sync/upload URL-failover path, so a silent failure here is exactly why a device would never fail over to a clone server — with no log line saying so.

## The fix

```kotlin
val alternativeUrl = extractedUrl?.let { serverMappings[it] }
return UrlMapping(url, alternativeUrl, extractedUrl)
```

and replace the `printStackTrace` with `Log.w(TAG, "Could not extract base url from $url", e)`.

## Acceptance

Extend the test with:
- a mapped primary → correct alternative
- an unmapped host → `alternativeUrl == null`
- a non-default port preserved in `extractedBaseUrl`
- a malformed string → returns null **without throwing**

## Notes

dogi:
Clean up ServerUrlMapper by dropping the empty also block and logging malformed URL errors via Log.w instead of printStackTrace, and extend unit tests to cover failover mapping cases.

Fixes #16787

---
*PR created automatically by Jules for task [14594445183023147623](https://jules.google.com/task/14594445183023147623) started by @dogi*


dogi:
**Review — rating: 72/100**

The cleanup half is unambiguously good. `serverMappings[baseUrl].also { }` with an empty body was dead syntax that read like something had been deleted mid-edit, and the `val result = …; return result` temp added nothing. Both go, and `processUrl` is now three obvious lines.

**The logging change leaks credentials, though, and that's what holds this back.** The new line is:

```kotlin
Log.w(TAG, "Could not extract base url from $url", e)
```

URLs in this codebase carry embedded credentials. `ConfigurationsRepositoryImpl.buildCouchdbUrl` (line 437) builds `scheme://satellite:$pin@host` — its `if (currentUrl.contains("@")) return currentUrl` branch exists precisely because these URLs are expected to have userinfo in them — and `UrlUtils.getUrl()` resolves through `dbUrl(baseUrl(spm))`, i.e. the stored CouchDB URL. `processUrl` is called with exactly those values from `MainApplication:210`, `ConfigurationsRepositoryImpl:177/269/512`, `ChatRepositoryImpl:123`, `DownloadService:262` and others, so the interpolated `$url` can be `http://satellite:<pin>@host:5984/db`.

That writes the satellite PIN to logcat. The app ships with `minifyEnabled = false` and no log-stripping rule, so `Log.w` is live in release builds, and `CLAUDE.md` already flags these PINs as real credentials pending rotation. Any log capture or bug report from a device then contains them.

The failure path is admittedly hard to reach — `toUri()` is lenient and rarely throws — but a defensive log that prints a secret when it does fire is the wrong shape regardless of probability. Log the redacted form instead:

```kotlin
Log.w(TAG, "Could not extract base url from ${url.toUri().host ?: "<unparseable>"}", e)
```

or simply drop `$url` from the message — the exception and tag are already enough to locate the call.

**Verdict:** needs another round — keep both cleanups, redact the URL before logging it.

---

dogi:
**Re-review — rating: 86/100** (was 72)

The blocking issue is fixed the simplest way: `$url` is gone from the message entirely, so `Log.w(TAG, "Could not extract base url", e)` can no longer print a `satellite:<pin>@host` URL to logcat. That closes the credential exposure without giving up the tag or the stack trace, which were the parts worth having.

Both cleanups from the first round stand: the `serverMappings[baseUrl].also { }` with an empty body is gone, and the `val result = …; return result` temp is inlined. `processUrl` now reads as the three lines it always was.

**Nit:** you've traded a little diagnostic value for the safety — the log no longer says *which* URL failed to parse, and `extractBaseUrl` is called from a dozen sites. If that turns out to matter, the redacted middle ground still works, since the host isn't the secret:

```kotlin
Log.w(TAG, "Could not extract base url from host ${runCatching { url.toUri().host }.getOrNull() ?: "<unparseable>"}", e)
```

Not needed to merge — the current form is correct and safe.

⚠️ **No build or test verdict on this head** — only `myPlanet labels` ran; neither `build` nor `test` fired on the fix commit. Worth a push to get CI to run before it merges.

**Verdict:** mergeable — the credential leak is resolved.

---

Okuro3499 (approved):

PR #16804, opened 2026-09-04
sync: smoother resources download coordinating (fixes #16790)

### LINKED ISSUE #16790: Inject @ApplicationScope into ResourceDownloadCoordinator
## What

A `@Singleton` service reaches for the global `MainApplication.applicationScope` when an injectable equivalent already exists.

**File:** `app/src/main/java/org/ole/planet/myplanet/services/ResourceDownloadCoordinator.kt` (line 19)

## Why

```kotlin
MainApplication.applicationScope.launch {   // line 19
```

`di/ServiceModule` already provides an injectable `@ApplicationScope CoroutineScope` — the qualifier at line 41, the provider at 49-50, and it's already consumed elsewhere in the same module (line 62). Using it removes a direct `MainApplication` dependency from a `@Singleton` service and makes the class testable without the Application singleton.

## The fix

- Add `@ApplicationScope private val applicationScope: CoroutineScope` to the constructor.
- Replace `MainApplication.applicationScope.launch` with `applicationScope.launch`.
- Drop the `MainApplication` import.

## Acceptance

- Background downloads still launch **only** when the server is reachable — that guard must be preserved exactly.
- No `MainApplication` reference remains in the file.

## Notes

dogi:
Injected @ApplicationScope CoroutineScope into ResourceDownloadCoordinator constructor, replaced direct MainApplication.applicationScope usage, removed MainApplication import, and added ResourceDownloadCoordinatorTest.

Fixes #16790

---
*PR created automatically by Jules for task [15197928866140137990](https://jules.google.com/task/15197928866140137990) started by @dogi*


dogi:
**Review — rating: 90/100**

`MainApplication.applicationScope` was a static reference to a scope living on the `Application` object, which made `ResourceDownloadCoordinator` untestable in isolation — you can't substitute a `TestScope` for a static, so any test of `startBackgroundDownload` needed the real application up. Taking `@ApplicationScope CoroutineScope` through the constructor is the qualifier the project already provides for exactly this (`DispatcherModule`), so it's the established pattern rather than a new one.

Nothing changes at runtime: the same application-lifetime scope is injected, so the coroutine's lifetime and cancellation behaviour are identical. It just arrives through the graph instead of a global, which also drops the `MainApplication` import and with it a `services` → application-class dependency.

This is the same treatment #16783 applies to `SubmissionsUploader`, so the two are consistent.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16815, opened 2026-09-04
teams: smoother voices repository list inserting (fixes #16799)

### LINKED ISSUE #16799: Collapse the double collection pass in VoicesRepositoryImpl.insertNewsList
## What

`insertNewsList` builds a mapped list, then maps and filters that whole list again just to collect the lookup ids.

**File:** `app/src/main/java/org/ole/planet/myplanet/repository/VoicesRepositoryImpl.kt` (`insertNewsList`, lines 361-365)

## Why

```kotlin
val mappedDocs = docs.map { it to JsonUtils.getString("_id", it) }        // 361
val underscoreIds = mappedDocs.map { it.second }.filter { it.isNotEmpty() } // 362 — two more intermediate lists
val existing = newsDao.getByUnderscoreIds(underscoreIds).associateBy { it._id }
val newsList = mappedDocs.map { (doc, id) -> buildNewsFromJson(doc, id, existing) }
newsDao.upsertAll(newsList)
```

Line 362 allocates two more intermediate collections over potentially hundreds of news records — after the `mappedDocs` pairing had already done the work of removing an N+1 query.

## The fix

Keep **one** ordered document/id representation so `_id` is parsed once per document, and populate the non-blank id query list during that same traversal.

**Preserve exactly:**
- duplicate ids (don't dedupe if the current code doesn't)
- blank-id handling
- input order
- the single bulk DAO lookup
- `associateBy` conflict behavior (last-wins)
- every `News` still built through `buildNewsFromJson`, with one `upsertAll`

## Acceptance

- Same `News` rows written, same order, for the same input — including a batch containing duplicate and blank `_id` values.
- Still exactly one `getByUnderscoreIds` call and one `upsertAll`.

## Notes

- Conventions: `docs/CODE_STYLE_GUIDE.md`; test patterns: `docs/TESTING.md`. Run `VoicesRepositoryImplTest` and `VoicesRepositoryNewsSyncTest`.

dogi:
Optimized VoicesRepositoryImpl.insertNewsList to extract JSON _id and populate lookup IDs during the initial traversal over docs, eliminating two redundant intermediate list allocations while maintaining exact duplicate and blank _id handling and order semantics.

Fixes #16799

---
*PR created automatically by Jules for task [10846000071884054266](https://jules.google.com/task/10846000071884054266) started by @dogi*


dogi:
**Review — rating: 84/100**

Correct and equivalent: the old `mappedDocs.map { it.second }.filter { it.isNotEmpty() }` walked the pairs twice and allocated two intermediate lists, and building `underscoreIds` during the single `map` produces the same ids in the same order. Pre-sizing with `ArrayList<String>(docs.size)` is a good detail — it's the upper bound, so the list never reallocates while filling.

This sits on the sync write path for news, where the comment above it notes the batch can be hundreds of items, so trimming a pass and two allocations per batch is worth having.

**Nit:** the saving comes from mutating `underscoreIds` as a side effect inside `map`, which makes the lambda impure — the reason the two-pass version read cleanly is that each step did one thing. `buildList` with an explicit loop, or partitioning after the fact, would keep the single pass without hiding an accumulation inside a `map`. Minor, but it's the kind of thing that surprises the next reader of a lambda they expect to be pure.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16816, opened 2026-09-04
sync: smoother retry repository dao operating (fixes #16798)

### LINKED ISSUE #16798: Consolidate the retry attempt transition and stop the read-modify-write
## What

Two retry paths each do a SELECT plus an UPDATE and open-code the same state transition, while their siblings in the same DAO are already single statements.

**Files:**
- `app/src/main/java/org/ole/planet/myplanet/repository/RetryRepositoryImpl.kt` (`updateAttempt` line 37, `markFailed` line 63)
- `app/src/main/java/org/ole/planet/myplanet/model/RetryOperation.kt`
- `app/src/main/java/org/ole/planet/myplanet/data/room/dao/RetryDao.kt`

## Why

Both methods do `retryDao.findById(id)` (lines 41, 64) → mutate the loaded `RetryOperation` → `update(op)`. That's a SELECT plus an UPDATE on **every retry attempt**, and both open-code the same transition: increment `attemptCount`, set `lastAttemptTime`, recompute `nextRetryTime`, and decide abandon-vs-pending from `attemptCount >= maxAttempts`.

The sibling methods are already one-line `@Query` UPDATEs in the same DAO:

```kotlin
@Query("UPDATE retry_operation SET status = 'in_progress' WHERE id = :id")            // 49
@Query("UPDATE retry_operation SET status = 'completed', lastAttemptTime = :timestamp WHERE id = :id")  // 52
```

## The fix

Two halves that compose into one change:

1. Put the **pure state transition** on `RetryOperation` as a method — it has no Android imports, so it's unit-testable without a DAO.
2. Add a `@Query` UPDATE to `RetryDao` so each path becomes a single statement.

## Acceptance

⚠️ **Preserve the status transitions exactly**: `markFailed` keeps `STATUS_PENDING` unless the new count reaches `maxAttempts`, and only then `STATUS_ABANDONED`. Test both sides of that boundary.

- No schema change, no `AppDatabase` version bump.
- Backoff constants unchanged.

## Notes

dogi:
Consolidates retry attempt state transitions on RetryOperation and converts updateAttempt and markFailed in RetryRepositoryImpl to perform single-statement atomic updates via RetryDao.recordFailedAttempt.

Fixes #16798

---
*PR created automatically by Jules for task [8552587141177512728](https://jules.google.com/task/8552587141177512728) started by @dogi*


dogi:
**Review — rating: 76/100**

Two genuinely good things here, and one structural problem that I think warrants another round.

**The atomicity win is real and well timed.** Read-modify-write on `retry_operation` had a lost-update window: two workers reading the same row both see `attemptCount = n` and both write `n + 1`, so a failure is silently not counted and an operation can retry past `maxAttempts`. That was mostly theoretical while the queue drained serially — but #16743 in this same batch introduces six-way concurrent draining, which makes it reachable. A single `UPDATE … SET attemptCount = attemptCount + 1 …` is the right fix and the two PRs complement each other.

**It also fixes a real bug.** The old `updateAttempt` and `markFailed` had *different* logic: `updateAttempt` set `nextRetryTime` unconditionally and only ever assigned `STATUS_ABANDONED`, never resetting to `STATUS_PENDING`. So an operation failing out of `in_progress` through that path stayed `in_progress` forever and would never be picked up again. Routing both through one statement with an explicit `CASE … THEN 'abandoned' ELSE 'pending'` closes that. Worth calling out in the description, because it's more valuable than the performance framing.

I checked the backoff arithmetic translates correctly: `30000 * (1 << MIN(attemptCount + 1, 30))` capped at `1800000` matches `minOf(BASE_DELAY_MS * (1L shl attemptCount), MAX_DELAY_MS)` on the incremented count, and the `MIN(…, 30)` shift clamp is actually safer than the Kotlin, which would wrap at shift ≥ 64.

**The problem: the logic now exists twice, and the tests exercise the copy that doesn't run.**

`RetryOperation.recordFailedAttempt(...)` is a new Kotlin implementation of the same transition, but production goes through `retryDao.recordFailedAttempt(...)` — the SQL. Nothing calls the model method outside tests. And the test suite splits exactly along that seam: `RetryOperationTest` exercises the Kotlin method (pending/abandoned/boundary), while `RetryRepositoryImplTest` mocks the DAO and asserts *delegation* with `coVerify`. So the status transition, the boundary at `maxAttempts`, and the backoff formula that actually ship are covered by **no test at all** — the green suite is testing a parallel implementation.

Compounding it, the SQL hardcodes `30000` and `1800000` instead of `BASE_DELAY_MS` / `MAX_DELAY_MS`. A `@Query` string can't reference Kotlin constants, so that's forced — but it means changing the backoff constants silently changes the Kotlin path and the tests while leaving production untouched, with nothing failing.

Two ways out: add an in-memory Room test against `recordFailedAttempt` (the shape `NotificationDaoTest` uses in #16677) so the real statement is pinned, and either delete the model method or make it the single source with the DAO as a thin caller. Right now it's two definitions that can drift, and only one of them is watched.

**Nit:** `nextRetryTime` is now written even on the abandoned branch, where the old code left it alone. Harmless — `getPending()` filters on status — but it's a small unnecessary difference.

**Verdict:** needs another round — cover the SQL with a DAO test and collapse the duplicate implementation.

---

dogi:
**Re-review — rating: 84/100** (was 76)

The blocker is addressed: `RetryDaoTest` is a real in-memory Room test that drives `retryDao.recordFailedAttempt(...)` — the statement that actually ships — rather than the Kotlin model copy. It covers the three cases that matter: below max (status back to `pending`, computed `nextRetryTime`), the `attemptCount + 1 >= maxAttempts` boundary (status `abandoned`), and the backoff clamping at `1_800_000`. The transition that had no coverage now has it.

**The `nextRetryTime` nit turned into a real fix.** The SQL now reads `nextRetryTime = CASE WHEN attemptCount + 1 >= maxAttempts THEN nextRetryTime ELSE :timestamp + MIN(…) END`, so the abandoned branch preserves the existing value instead of overwriting it — which is what the Kotlin always did, and the test pins it with `assertEquals(999L, result.nextRetryTime) // Preserved on abandoned`. The two implementations now agree on that branch where they previously didn't.

Everything good from the first revision stands: one atomic `UPDATE` instead of read-modify-write (which matters now that #16743 drains the queue six-way concurrent), and both `updateAttempt` and `markFailed` routed through it, fixing the old `updateAttempt` path that could leave an operation stuck in `in_progress` forever.

**Two things I'd still change, neither blocking:**

- **`RetryOperation.recordFailedAttempt` is still dead in production.** Grepping `app/src/main` on this branch, the only production call is `RetryRepositoryImpl:53` → `retryDao.recordFailedAttempt(...)`. The model method exists solely to be exercised by `RetryOperationTest`. Now that the DAO test covers the real path, that Kotlin copy is a second definition of the same state machine with no caller — worth deleting rather than leaving two things to keep in step.
- **The two definitions still aren't cross-checked.** `RetryDaoTest` asserts hardcoded expectations (`timestamp + 120_000L`, `timestamp + 1_800_000L`) rather than deriving them from `BASE_DELAY_MS` / `MAX_DELAY_MS` or from `RetryOperation.calculateNextRetryTime`. So changing the Kotlin constants still leaves the SQL literals untouched with every test green. One assertion of the form `assertEquals(RetryOperation.calculateNextRetryTime(2, timestamp), result.nextRetryTime)` would tie them together — and would be a natural home for the model method if you keep it.

⚠️ **No build or test verdict on this head** — only `myPlanet labels` ran; `build` and `test` never fired. That's worth a push here more than anywhere: the entire fix in this revision is a new test, and it has not executed yet. Everything above is from reading the diff.

**Verdict:** mergeable once CI actually runs the new `RetryDaoTest`.

---

PR #16817, opened 2026-09-04
enterprises: smoother finances transaction totals view modelling (fixes #16797)

### LINKED ISSUE #16797: Derive the finances header from transactions and move the totals math onto Transaction
## What

The finances ViewModel keeps two parallel `MutableStateFlow`s and recomputes one from the other imperatively, when the second is a pure function of the first.

**Files:**
- `app/src/main/java/org/ole/planet/myplanet/ui/enterprises/EnterprisesFinancesViewModel.kt`
- `app/src/main/java/org/ole/planet/myplanet/model/Transaction.kt`

## Why

```kotlin
private val _transactions = MutableStateFlow<List<Transaction>>(emptyList())   // 31
private val _headerState  = MutableStateFlow(FinanceHeaderState())             // 34
...
_transactions.value = results   // 56
calculateTotal(results)         // 57 — loops the same list again, mutates _headerState
```

`headerState` is derived state carried as independent state. That's two sources of truth that can disagree, and `calculateTotal` (line 62) is an imperative side-effecting method whose output can't be tested without the ViewModel.

## The fix

1. Turn `calculateTotal` into a **pure function** returning `FinanceHeaderState`, with the summing itself as a helper on `Transaction` or its companion.
2. Expose:
   ```kotlin
   val headerState = transactions
       .map { /* pure summing */ }
       .stateIn(viewModelScope, SharingStarted.WhileSubscribed(5000), FinanceHeaderState())
   ```
3. Delete `_headerState` and the imperative `calculateTotal(results)` call.

Keep `isCautionVisible = total < 0` (line 77) exactly as it is.

## Acceptance

Tests on the pure summing function covering: credit-only, debit-only, mixed, and empty. `isCautionVisible` still true exactly when total is negative.

## Notes

dogi:
Refactor EnterprisesFinancesViewModel to derive headerState reactively from transactions, move calculation logic to Transaction.Companion.calculateTotals pure function, and add unit tests covering credit-only, debit-only, mixed, and empty transaction lists.

Fixes #16797

---
*PR created automatically by Jules for task [12987405971623173746](https://jules.google.com/task/12987405971623173746) started by @dogi*

## Summary by CodeRabbit

* **New Features**
  * Added finance summary calculations for debit, credit, total balance, and caution status.
  * Finance summaries now update automatically as transactions change.

* **Tests**
  * Added coverage for empty, credit-only, debit-only, and mixed transaction scenarios.
  * Added validation for case-insensitive transaction types and negative-balance warnings.


dogi:
**Review — rating: 74/100**

The reactive derivation is the right idea. `headerState` was a `MutableStateFlow` kept in sync by a manual `calculateTotal(results)` call sitting next to `_transactions.value = results` — two writes that had to be remembered together, and a future code path that set transactions without calling `calculateTotal` would leave the header silently stale. Deriving it with `transactions.map { … }.stateIn(…)` makes that desynchronisation unrepresentable. The arithmetic is carried over exactly: credit/debit accumulation with `ignoreCase`, `total = credit - debit`, `isCautionVisible = total < 0`.

**But `Transaction` now imports the ViewModel, and that's backwards.** The first line of `model/Transaction.kt` is:

```kotlin
import org.ole.planet.myplanet.ui.enterprises.EnterprisesFinancesViewModel.FinanceHeaderState
```

So a model class in `model/` now depends on a UI class in `ui/enterprises/`, and specifically on a type nested inside a ViewModel. `CLAUDE.md` describes the layering as UI → Repository → Service → Data; this points the arrow the other way. Concretely it means `Transaction` can't be used without dragging in a ViewModel, the model layer can't be tested or reused without the UI, and the next person who needs totals in a different screen either imports this ViewModel too or duplicates the function.

The fix is small and doesn't change any of the good parts: lift `FinanceHeaderState` out of `EnterprisesFinancesViewModel` — either into `model/` beside `Transaction`, or into the enterprises package as a standalone type — so `calculateTotals` returns something at or below its own layer. The ViewModel keeps exposing it exactly as it does now.

**Nit:** moving from an eagerly-maintained `MutableStateFlow` to `WhileSubscribed(5000)` means totals are only computed while the screen observes, which is a small improvement — worth knowing it's a deliberate change in when the work happens, not just where.

**Verdict:** needs another round — move `FinanceHeaderState` out of the ViewModel so the model stops importing the UI layer; everything else here is good.

---

dogi:
**Re-review — rating: 88/100** (was 74)

The layering inversion is gone, and fixed the right way. `FinanceHeaderState` is now a top-level type in `model/FinanceHeaderState.kt`, so `Transaction.calculateTotals` returns something in its own package — `Transaction.kt` has **no imports at all** now, where the previous revision reached up into `ui.enterprises.EnterprisesFinancesViewModel`. The arrow points down again.

Lifting it rather than duplicating it is also what keeps this cheap: the ViewModel simply imports `org.ole.planet.myplanet.model.FinanceHeaderState` and exposes `headerState` under the same name and shape as before. Nothing downstream is in this diff, which tells you the consumers never spelled the nested type out — so moving it cost no call-site churn at all, and the green build across both flavors confirms nothing was left referencing `EnterprisesFinancesViewModel.FinanceHeaderState`.

Everything good about the first revision survives intact. The arithmetic is unchanged — credit/debit accumulation with `ignoreCase`, `total = credit - debit`, `isCautionVisible = total < 0` — and deriving `headerState` with `transactions.map { … }.stateIn(…)` still makes it impossible for the header and the list to disagree, which was the actual bug in the manual `calculateTotal(results)` call sitting beside `_transactions.value = results`. `calculateTotals` being a pure function over a `List<Transaction>` in the model layer is now genuinely unit-testable without touching a ViewModel.

**Verdict:** mergeable.

---

Okuro3499 (dismissed):
```
recordFailedAttempt reimplements the same attempt/backoff/abandon logic as RetryDao.recordFailedAttempt's SQL, but is never called from production code                 
    (RetryRepositoryImpl calls the DAO directly). Two sources of truth for one business rule, one of them dead.
```

dogi:
**Review — rating: 86/100**

Good refactor with the right shape. `headerState` was a `MutableStateFlow` kept in sync by hand — `loadTransactions` had to remember to call `calculateTotal(results)` after every assignment to `_transactions`, and nothing enforced that. Deriving it instead:

```kotlin
val headerState: StateFlow<FinanceHeaderState> = transactions
    .map { Transaction.calculateTotals(it) }
    .stateIn(viewModelScope, SharingStarted.WhileSubscribed(5000), FinanceHeaderState())
```

makes the header incapable of drifting from the transaction list, and deletes the call site that could be forgotten. Pulling the arithmetic out into `Transaction.calculateTotals` turns it into a pure function that's testable without a ViewModel, and moving `FinanceHeaderState` out of the ViewModel into `model/` follows from `Transaction` now referencing it. I diffed the moved loop against the original `calculateTotal` — credit/debit branching on `"credit".equals(type, ignoreCase = true)`, `total = credit - debit`, `isCautionVisible = total < 0` — it's verbatim, so no behaviour rides along with the restructure.

`TransactionTest` covers it properly: empty, credit-only (with `"CREDIT"` uppercase exercising the `ignoreCase`), debit-only with caution set, and mixed in both the positive and negative directions.

**One thing worth reconsidering: `SharingStarted.WhileSubscribed(5000)`.**

`EnterprisesFinancesFragment.bindHeader` reads `viewModel.headerState.value` directly (line 127). Under `WhileSubscribed`, `.value` is `initialValue` — all zeros — until a collector has been active. It happens to be safe here, because the only thing that triggers `bindHeader` is the collector itself:

```kotlin
collectLatestWhenStarted(viewModel.headerState) { headerAdapter.notifyItemChanged(0, PAYLOAD_HEADER) }
```

so the `.value` read is always downstream of the subscription that keeps the flow hot. That's a correct arrangement, but it's implicit and undocumented — the next person to read `headerState.value` from a path not driven by that collector gets zeros with nothing failing.

The test changes in this PR are the evidence that the seam is awkward. `viewModel.headerState.first()` no longer works and had to become an explicit long-lived collector recording emissions:

```kotlin
val states = mutableListOf<FinanceHeaderState>()
val job = backgroundScope.launch(UnconfinedTestDispatcher(testScheduler)) { viewModel.headerState.collect { states.add(it) } }
...
val headerState = states.last()
```

That's the idiomatic workaround and it's fine as written, but it's a workaround. Since `transactions` is a hot `MutableStateFlow` that always holds a value and `calculateTotals` is a trivial pure loop, `SharingStarted.Eagerly` costs essentially nothing here, makes the `.value` read in `bindHeader` unconditionally correct, and would let the test go back to `first()`. `WhileSubscribed` earns its keep when the upstream is expensive or does I/O; neither applies.

Minor: the fragment has its own private `HeaderState` (filter/sort UI state) alongside the ViewModel's `FinanceHeaderState`, and `bindHeader` reads both. Pre-existing, but the near-identical names make that function harder to follow than it needs to be.

⚠️ **This push produced only the `myPlanet labels` check — `build` and `test` never ran, so there is no CI verdict**, including on the new `TransactionTest` and the rewritten ViewModel test. Worth a re-run before merging. Same recurring symptom as #16883, #16911, #16885 and a dozen others.

**Verdict:** mergeable once CI actually runs on it.

---

Okuro3499 (approved):

PR #16818, opened 2026-09-04
sync: smoother personals repository document posting (fixes #16807)

### LINKED ISSUE #16807: Route the personals document POST through UploadRepository
## What

`PersonalsRepositoryImpl` holds two upload paths — one through `UploadRepository`, one bypassing it straight to `ApiInterface`.

**Files:**
- `app/src/main/java/org/ole/planet/myplanet/repository/UploadRepository.kt` / `UploadRepositoryImpl.kt`
- `app/src/main/java/org/ole/planet/myplanet/repository/PersonalsRepositoryImpl.kt` (`uploadPersonalDocument`, line 81)

## Why

```kotlin
private val apiInterface: ApiInterface,      // 20
private val uploadRepository: UploadRepository,  // 21
...
val response = apiInterface.postDoc(         // 82 — bypasses the upload boundary
...
val response = uploadRepository.uploadAttachment(  // 117 — goes through it
```

The same class already injects `UploadRepository` and uses it for attachments. The document POST is the odd one out.

## The fix

**First check whether an equivalent already exists** — `UploadRepository` currently declares:

```kotlin
suspend fun postUpload(url: String, serializedData: JsonObject): Response<JsonObject>   // line 11
```

If that is equivalent to the `apiInterface.postDoc(...)` call at line 82, **reuse it** and this change shrinks to two files with no new interface method. Only add a `postDocument(url, payload)` wrapper if the existing method genuinely doesn't fit (e.g. a different header or response shape) — and say which it was in the PR description.

Either way: switch `uploadPersonalDocument` to the repository call, then drop the `ApiInterface` field (line 20) and its import if nothing else in the class uses it.

## Acceptance

- `PersonalsRepositoryImpl` no longer references `ApiInterface`.
- The POST sends the same URL, headers and payload as today, and the `Pair<String, String>?` return contract is unchanged.

## Notes

dogi:
This change routes the document POST request in `PersonalsRepositoryImpl` through `UploadRepository.postUpload` rather than calling `ApiInterface` directly. Reusing `UploadRepository.postUpload` eliminated the need for `ApiInterface` in `PersonalsRepositoryImpl`, allowing us to drop the dependency from its constructor.

- Updated `PersonalsRepositoryImpl.uploadPersonalDocument` to use `uploadRepository.postUpload`.
- Removed `apiInterface` from `PersonalsRepositoryImpl` constructor and unused imports.
- Updated `PersonalsRepositoryImplTest` to mock and verify `uploadRepository.postUpload` instead of `apiInterface.postDoc`.

Fixes #16807

---
*PR created automatically by Jules for task [2955691247768816726](https://jules.google.com/task/2955691247768816726) started by @dogi*


dogi:
**Review — rating: 88/100**

Provably a no-op at runtime: `UploadRepositoryImpl.postUpload(url, serializedData)` is exactly `apiInterface.postDoc(UrlUtils.header, "application/json", url, serializedData)` — the same header, the same content type, the same call. So this swaps a hand-assembled request for the identical one behind the shared entry point, with no behavioural surface to check.

The value is that `PersonalsRepositoryImpl` stops being a second place that knows how an upload is authenticated. `UrlUtils.header` is where the CouchDB credentials come from, and having domain repositories reach for it directly means every one of them has to be updated if that changes — which matters given `CLAUDE.md` flags those PINs as pending rotation. One fewer caller of `UrlUtils.header`, and the `ApiInterface` injection drops out entirely, leaving the repository with its DAO and the two collaborators it actually needs.

This also matches how the rest of the upload path already works — `PhotoUploader` and `UploadCoordinator` both go through `UploadRepository` rather than `ApiInterface`.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16819, opened 2026-09-04
all: less voices health repositories text utils is more (fixes #16805)

### LINKED ISSUE #16805: Purge android.text.TextUtils from the Voices and Health repository implementations
## What

Two repository implementations import an Android framework helper to do plain null/empty string checks.

**Files:**
- `app/src/main/java/org/ole/planet/myplanet/repository/VoicesRepositoryImpl.kt` (import line 3; call sites 454, 459)
- `app/src/main/java/org/ole/planet/myplanet/repository/HealthRepositoryImpl.kt` (import line 3; call site 217)

## Why

```kotlin
// VoicesRepositoryImpl.kt
if (!TextUtils.isEmpty(news.viewableId)) { ... }   // 454
if (!TextUtils.isEmpty(viewInStr)) { ... }         // 459
// HealthRepositoryImpl.kt
if (TextUtils.isEmpty(myHealth.userKey)) { ... }   // 217
```

A repository needs no Android framework helper for an emptiness check, and each removed `android.*` import is one less platform coupling in a layer that should be testable without Robolectric.

## The fix

Replace `TextUtils.isEmpty(x)` with `x.isNullOrEmpty()` (and `!TextUtils.isEmpty(x)` with `!x.isNullOrEmpty()`) at the three sites, preserving current null-and-empty behavior exactly — `TextUtils.isEmpty` treats null and "" alike, and so does `isNullOrEmpty()`.

Delete both imports.

## Acceptance

- `grep -n TextUtils` over both files returns nothing.
- Behavior unchanged at all three sites.
- Run `VoicesRepositoryImplTest`, `VoicesRepositoryNewsSyncTest` and `HealthRepositoryImplTest`.

## Notes

dogi:
Purged android.text.TextUtils imports and usages from VoicesRepositoryImpl.kt and HealthRepositoryImpl.kt by replacing TextUtils.isEmpty calls with Kotlin standard library's isNullOrEmpty().

Fixes #16805

---
*PR created automatically by Jules for task [10066981392546064595](https://jules.google.com/task/10066981392546064595) started by @dogi*


dogi:
**Review — rating: 92/100**

Exactly equivalent — `TextUtils.isEmpty(s)` is `s == null || s.length == 0`, which is `s.isNullOrEmpty()` — and the payoff is bigger than the three lines suggest.

`TextUtils` is `android.text`, so it's a framework class that returns `0`/`false` under a stock JVM unless mocked. Every repository still importing it forces its unit tests onto Robolectric or into `mockkStatic` gymnastics for something as trivial as an emptiness check. `HealthRepositoryImpl` and `VoicesRepositoryImpl` are both plain data-layer classes with DAO and Gson dependencies otherwise, so removing the import moves them closer to being testable as ordinary Kotlin.

The negated forms are right too: `!TextUtils.isEmpty(x)` becomes `!x.isNullOrEmpty()` rather than the tempting-but-wrong `x.isNotEmpty()`, which would have changed behaviour on null. Worth noting because that's the usual way this refactor breaks.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16823, opened 2026-09-04
login: smoother sync failure logging (fixes #16808)

### LINKED ISSUE #16808: Replace the six printStackTrace calls in SyncActivity with tagged logging
## What

Six `printStackTrace()` calls in the sync/login activity write to stderr with no tag, no level, and no path into the diagnostics pipeline.

**File:** `app/src/main/java/org/ole/planet/myplanet/ui/sync/SyncActivity.kt` (lines 290, 345, 409, 587, 653, 819)

## Why

Verified: six call sites, one of them on `error` rather than `e` (line 653). Each sits in a catch around a meaningful operation, and each currently produces an untagged stderr dump — so a failure in the sync or login path is invisible in the field, which is the one place this app most needs a log.

This is the companion to the `MainApplication` / `NotificationActionReceiver` conversion; the same treatment, applied to the remaining file.

## The fix

Replace each with `Log.e(TAG, "<operation that failed>", e)`.

- Use the class's existing `TAG` if present; add a `private const val TAG` in a companion if not.
- Name the operation that failed — don't restate the exception.
- Mind line 653, where the variable is `error`, not `e`.
- **Keep every catch clause's control flow identical.** This is a logging change only.

## Acceptance

- `grep -n printStackTrace app/src/main/java/org/ole/planet/myplanet/ui/sync/SyncActivity.kt` returns nothing.
- No control-flow change in any touched catch block.

## Notes

dogi:
Replaces six printStackTrace() calls in SyncActivity.kt with Log.e(TAG, "<operation>", e) to ensure proper tagged logging while preserving control flow.

Fixes #16808

---
*PR created automatically by Jules for task [10994788479833139774](https://jules.google.com/task/10994788479833139774) started by @dogi*


dogi:
**Review — rating: 90/100**

Six `printStackTrace` calls in the class that owns login and first sync — the place where a silent, untagged trace on `System.err` costs the most, because these are the failures users actually report ("it won't sync", "it won't log in") and the ones hardest to reproduce off-device.

The messages do the work: "Failed to authenticate user", "Failed to check server reachability", "Failed to complete sync post-processing" and "Failed to continue sync process" were previously four indistinguishable stack dumps from one activity. Now a logcat capture says which stage failed without anyone having to map line numbers.

A single `TAG` in the existing companion is the right shape, and unlike #16782 there's no pre-existing tag constant here for it to duplicate. Converting the `onError = { error -> error.printStackTrace() }` callback at the profile-login site is a good catch too — that one wasn't even a catch block, so it was easy to miss.

No URL or credential interpolation in any of the messages, which matters in this class given the server URLs it handles carry embedded `satellite:<pin>@` userinfo.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16824, opened 2026-09-04
sync: smoother upload repository mime type attaching (fixes #16809)

### LINKED ISSUE #16809: Replace the openConnection MIME probe in UploadRepositoryImpl.uploadAttachment
## What

Every attachment upload opens a `URLConnection` on the file purely to sniff its content type, and never reads a byte from it.

**File:** `app/src/main/java/org/ole/planet/myplanet/repository/UploadRepositoryImpl.kt` (`uploadAttachment`, line 96)

## Why

```kotlin
val connection = file.toURI().toURL().openConnection()          // 104
val type = connection.contentType ?: "application/octet-stream" // 105
```

The connection is opened solely for `contentType` and then discarded — so every attachment, photo and CV upload pays a content-sniffing open of the file before the real upload begins.

The project already has local-only patterns for exactly this: `URLConnection.guessContentTypeFromName(file.name)` (used in `WebViewActivity`) and `FileUtils.getMimeType(name)` (used in `UploadManager`).

## The fix

```kotlin
val type = URLConnection.guessContentTypeFromName(file.name) ?: "application/octet-stream"
```

Drop the now-unused import if it becomes unused.

**Leave untouched:** the header map via `FileUploader.getHeaderMap(mimeType, rev)` (line 111), the request body, and the response handling.

## Acceptance

⚠️ The resolved type still feeds `getHeaderMap(mimeType, rev)`, so uploads must keep sending the **correct Content-Type** — verify pdf, jpg and png each resolve to the same value they do today, and that an extensionless file still falls back to `application/octet-stream`.

## Notes

dogi:
Replaces file.toURI().toURL().openConnection() MIME sniffing in UploadRepositoryImpl.uploadAttachment with URLConnection.guessContentTypeFromName(file.name) ?: "application/octet-stream" to prevent opening URLConnections solely for content type sniffing.

Fixes #16809

---
*PR created automatically by Jules for task [17487480719536573385](https://jules.google.com/task/17487480719536573385) started by @dogi*


dogi:
**Review — rating: 88/100**

Better than a style swap — the old line was doing real I/O and leaking a handle. `file.toURI().toURL().openConnection()` returns a `FileURLConnection`, and reading `.contentType` forces `connect()`, which opens the file and may pull bytes through `guessContentTypeFromStream` to sniff the type. Nothing ever closed that connection, so every attachment upload opened a file stream and abandoned it. `URLConnection.guessContentTypeFromName(file.name)` is a pure lookup against the extension map: no file opened, nothing to close.

**One behavioural difference worth naming:** `FileURLConnection.getContentType()` tries the filename first and *falls back* to content sniffing, so a file with no extension or an unrecognised one could previously come back with a sniffed type and will now fall through to `"application/octet-stream"`. For attachment uploads that's a fair trade — and arguably the more predictable answer — but it isn't strictly identical.

**Nit:** with the I/O gone, the `withContext(dispatcherProvider.io)` around this block no longer has anything to move off the calling thread — `guessContentTypeFromName` is a string lookup and `asRequestBody` just wraps the `File` without reading it. The dispatcher hop could go too. Separately, the computed `type` is used for the request headers while the body is hardcoded to `"application/octet-stream"`; that predates this PR, but it does mean the sniffing difference above only affects the header.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16829, opened 2026-09-04
all: less text utils is more (fixes #16806)

### LINKED ISSUE #16806: Replace TextUtils with Kotlin stdlib checks in five UI files
## What

Five UI files each import `android.text.TextUtils` for emptiness checks the Kotlin idiom does null-safely and allocation-free.

**Files and sites:**
- `ui/health/HealthUsersAdapter.kt` (import 3, site 55)
- `ui/user/UserArrayAdapter.kt` (import 4, site 65)
- `ui/surveys/SendSurveyFragment.kt` (import 4, site 29)
- `ui/health/MyHealthFragment.kt` (import 6, site 260)
- `ui/enterprises/EnterprisesReportsFragment.kt` (import 8, sites 350, 354, 358, 362, 366, 370 — six checks in one validation `when`)

## Why

Ten `TextUtils.isEmpty` calls across five files, all doing what `isNullOrEmpty()` does natively.

## The fix

Swap to `isNullOrEmpty()`, preserving the negated forms and the existing string-template arguments (e.g. `"${binding.summary.text}"` stays as it is — only the emptiness check changes).

Delete the five imports.

## Acceptance

- `grep -rn "android.text.TextUtils"` over the five files returns nothing.
- The `EnterprisesReportsFragment` validation `when` still short-circuits in the same order with the same messages — that's six branches, so please check each rather than assuming.

## Notes

dogi:
Replaced android.text.TextUtils.isEmpty calls with Kotlin's native CharSequence?.isNullOrEmpty() across HealthUsersAdapter, UserArrayAdapter, SendSurveyFragment, MyHealthFragment, and EnterprisesReportsFragment, and removed all android.text.TextUtils imports.

Fixes #16806

---
*PR created automatically by Jules for task [1463099241252529719](https://jules.google.com/task/1463099241252529719) started by @dogi*


dogi:
**Review — rating: 80/100**

Four of the five files are exactly right. `user.userImage.isNullOrEmpty()`, `userModel?.dob.isNullOrEmpty()` and `id.isNullOrEmpty()` are faithful translations — `TextUtils.isEmpty(s)` is `s == null || s.isEmpty()`, and the negated forms correctly became `!x.isNullOrEmpty()` rather than `x.isNotEmpty()`, which would have flipped the null case. Dropping `android.text` from adapters and fragments is worth doing for the same reason as #16819.

**`EnterprisesReportsFragment` is where the mechanical translation goes wrong**, in six places:

```kotlin
"${binding.summary.text}".isNullOrEmpty()
```

A string template can never be null, so `isNullOrEmpty()` on it is a null check that can't fire — the compiler will flag it as such, and it's the kind of redundancy this PR is otherwise removing.

Worse, the template hides a real bug that was already there. When `binding.summary.text` is null, `"${...}"` produces the four-character string `"null"`, which is not empty — so the validation *passes* and a report submits with a literal "null" summary instead of showing "summary is required". `TextUtils.isEmpty("${...}")` had the identical flaw, so this isn't a regression, but the fix was one keystroke away:

```kotlin
binding.summary.text.isNullOrEmpty()
```

Dropping the template makes the null check meaningful, removes the redundancy, and closes the latent bug in the same edit — across all six fields.

⚠️ **No build or test verdict on this head** — only `myPlanet labels` ran; neither `build` nor `test` fired.

**Verdict:** mergeable as-is, but the six `"${…}"` wrappers are worth removing while you're in there — that turns a no-op null check into a real one.

---

Okuro3499 (approved):

PR #16830, opened 2026-09-04
life: smoother achievements referencing (fixes #16811)

### LINKED ISSUE #16811: Stop the per-element deepCopy that defeats Achievement's parse cache
## What

An LRU parse cache exists and works, but every cache hit still pays a full JSON tree clone per element.

**File:** `app/src/main/java/org/ole/planet/myplanet/model/Achievement.kt` (`parseStringListToJsonArray`, line 101)

## Why

```kotlin
internal const val CACHE_CAPACITY = 1000                          // 94
internal val parsedJsonCache: MutableMap<String, JsonElement> =   // 95
    Collections.synchronizedMap(object : LinkedHashMap<...>() {
        override fun removeEldestEntry(...) = size > CACHE_CAPACITY  // 97
    })
...
var ob = parsedJsonCache[s]        // 104
if (ob == null) { ...; parsedJsonCache[s] = ob }   // 107
array.add(ob?.deepCopy())          // 109  <-- full tree clone on every call, hit or miss
```

So the cache saves the *parse* but not the *allocation* — a cache hit still deep-copies the whole element.

## The fix

⚠️ **The audit is the actual work here — please don't just delete `deepCopy()`.**

1. Audit the callers of the functions that use `parseStringListToJsonArray`. Determine whether any of them mutates the elements of the returned `JsonArray`.
2. **If none mutates**: add the cached element directly and drop the per-element copy.
3. **If mutation risk exists**: deep-copy once at cache-*insert* time and hand out the shared element, documenting the no-mutation contract at the function.

Keep `CACHE_CAPACITY` eviction and the public API unchanged.

## Acceptance

- State in the PR description which branch the audit landed on and what you checked — that's the reviewable part.
- A test asserting a cache hit returns equal content without re-parsing.

## Notes

dogi:
### Audit findings

The issue offered two branches. The audit landed on neither — it reversed the premise: the per-element `deepCopy()` should stay.

**Callers checked** — the four public accessors (`achievementsArray`, `getReferencesArray()`, `linksArray`, `otherInfoArray`) have four users, and none mutates an element today:

1. `Achievement.serialize()` — adds the returned arrays into a freshly constructed `JsonObject`.
2. `EditAchievementFragment` — reads properties to build Chips/TextViews; its edit and delete paths construct new `JsonObject`s or remove entries from the `JsonArray`, which `parseStringListToJsonArray` allocates fresh on every call.
3. `UserRepositoryImpl` — only calls `setAchievements()` / `setReferences()`.
4. `AchievementTest` — reads and asserts values.

**But a caller audit is the wrong thing to hang this on.** `parsedJsonCache` is a process-wide static mutable map. Dropping the copy converts a one-time audit into a permanent unenforced invariant that every future caller of those four accessors has to keep true, with the failure mode being one record's mutation silently corrupting another's data — hard to reproduce and far from where it was caused.

**And the copy is not the cost the issue assumed.** `JsonPrimitive.deepCopy()` and `JsonNull.deepCopy()` return the receiver — verified in the Gson bytecode (`aload_0; areturn`) — so `links` and `otherInfo`, which are string arrays, already pay nothing. Only object-valued entries allocate, and those are exactly the ones that are unsafe to share. The copy is already free wherever it is unnecessary, and load-bearing wherever it is not.

Branch 3 of the issue (deep-copy once at cache-insert time, hand out the shared element) carries the same exposure: it protects the cached parse from its source, not the caller from other callers. The shared instance still escapes.

So `deepCopy()` stays, and the cache stays content-addressed — keyed by the source string, so the key fully determines the value and one entry serves every record holding that string within `CACHE_CAPACITY`.

### Changes made

**`Achievement.kt`** — the production diff against master is two things:

- Deleted `cachedReferencesArray` and its reset in `setReferences()`: written, never read.
- Added a comment above `parseStringListToJsonArray` recording why the copy is there, so the next reader doesn't re-derive this.

**`AchievementTest.kt`** — three tests that assert the invariant rather than the implementation:

- `parseStringListToJsonArray_neverHandsOutTheCachedInstance` — repeated reads return distinct instances, and mutating one leaves later reads unaffected.
- `parseStringListToJsonArray_isolatesRecordsWithIdenticalContent` — two records whose field serializes to the same string share the cache entry but not the instance, so neither can corrupt the other.
- `parseStringListToJsonArray_copyIsFreeForImmutablePrimitives` — a primitive-valued field still returns its cached instance, pinning the cost argument above.

`parsedJsonCache_isBoundedToCapacity` still covers `CACHE_CAPACITY` eviction, and the public API is unchanged.

### Issue resolution

Fixes #16811, with its finding reversed rather than implemented. The issue's acceptance criterion was the audit itself; the audit's answer is that this optimisation is not safe to make. That conclusion is recorded here and in the code comment, so the next person to have the idea finds the reasoning rather than a silently closed issue.

---
*PR created automatically by Jules for task [9990134509961830384](https://jules.google.com/task/9990134509961830384) started by @dogi*

---
https://claude.ai/code/session_01SMR4hsFXcuJ6mWMoSjs7SB

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved achievement data handling to prevent unintended sharing or mutation of values during repeated reads.
  * Preserved consistent behavior when achievement fields contain identical content across records.
* **Tests**
  * Expanded coverage for achievement data isolation, repeated reads, primitive values, and cache capacity behavior.


dogi:
**Review — rating: 84/100**

Removing `deepCopy()` on every element of every access is a real saving — `achievementsArray`, `linksArray` and `otherInfoArray` are *getters*, so each read was re-copying the whole parsed structure even though the parse itself was already cached. But it turns a private cache into shared mutable state reachable by callers, so the invariant is what matters. I traced it.

Only two places outside the model touch these arrays, both in `EditAchievementFragment` (lines 439-440), and both only ever `add` or `remove` on the **array** — lines 277, 293, 320-322, 375, 405. Nothing anywhere reaches into an element and mutates it. Since `add`/`remove` operate on the freshly-constructed `JsonArray` rather than on the cached `JsonElement`s, the cache can't be corrupted by any current caller. `remove(element)` matches by `equals`, which behaves identically whether the elements are shared or deep-copied, so even the aliasing case (the same JSON string twice in one list) is unchanged.

The KDoc stating the constraint is the right thing to add given that.

**Where this stays fragile:** the invariant is now load-bearing and unenforced, and the failure mode is nasty — one `array[i].asJsonObject.addProperty(...)` in a future edit silently corrupts the LRU for every subsequent achievement, including ones being serialized for upload, with nothing failing loudly at the mutation site. A comment is the weakest form of that guarantee. If you want it enforced, deep-copying only in the two accessors that hand arrays to the editing UI (`achievementsArray`, `getReferencesArray`) while leaving the serialize-only paths (`linksArray`, `otherInfoArray`) sharing would keep most of the win with none of the exposure.

**Verdict:** mergeable.

---

Okuro3499 (dismissed):
```
Achievement.kt:95-118 (parsedJsonCache / parseStringListToJsonArray): The cache key collides across unrelated Achievement records (no _id/field scoping), and the removed        
    deepCopy() was the only mitigation against a shared-instance mutation corrupting other users' data. Either:                                                                      
    - Keep deepCopy() (revert this optimization until there's a measured need), or                                                                                                   
    - Scope cache keys to something unique per logical field (e.g. "$achievementId#$fieldName#$s" or similar) so collisions can't cross records, or                                  
    - Make the returned elements structurally immutable / return a copy-on-write wrapper so the "don't mutate" contract is enforced rather than documented.                          
                                                                                                                                                                                     
    Without one of these, the PR is asking future maintainers to permanently remember an unenforced invariant across a shared static cache — that's the kind of thing that causes a  
    hard-to-reproduce, cross-user data-corruption bug months from now.                                                                         
```

dogi:
**Review — rating: 79/100**

The code that shipped here is good. The title and body describe the opposite change, and that's what needs fixing before it lands.

**What the diff actually does.** `deepCopy()` is still there:

```kotlin
private fun parseStringListToJsonArray(list: List<String>?): JsonArray {
    val array = JsonArray()
    for (s in list ?: emptyList()) {
        var ob = parsedJsonCache[s]
        if (ob == null) { ob = JsonUtils.gson.fromJson(s, JsonElement::class.java); parsedJsonCache[s] = ob }
        array.add(ob?.deepCopy())
    }
    return array
}
```

The production change is five lines: removing `@Ignore private var cachedReferencesArray` and its `= null` reset. I checked that field on master — it appears at its declaration (`Achievement.kt:36`) and its reset in `setReferences` (`:81`) and **nowhere else**. `getReferencesArray()` calls `parseStringListToJsonArray(references)` directly and never consults it. It was a vestigial cache that was written but never read, so deleting it cannot change behaviour. Clean removal.

The tests are the real content, and they're the right tests. `parsedJsonCache` is a companion-level bounded LRU keyed by the *serialized string*, so two `Achievement` records whose field serializes identically share one cache entry — which makes `deepCopy()` on every read load-bearing, not decoration. `isolatesRecordsWithIdenticalContent` pins exactly that: mutating `ach1`'s parsed element must not reach `ach2` or the cache. `neverHandsOutTheCachedInstance` pins the per-read isolation. And `copyIsFreeForImmutablePrimitives` is a sharper test than it looks — `achievement.linksArray[0] === achievement.linksArray[0]` compares two independent property reads, so it holds only because `JsonPrimitive.deepCopy()` returns `this`; a `JsonObject` field would fail it. That's the cost argument for keeping the copy, expressed as an assertion. Turning an invariant that was previously a comment into three tests is worth more than the micro-optimisation the PR set out to make.

**The problem.** The title is still *"Stop per-element deepCopy in Achievement parseStringListToJsonArray"*, and the body still says:

> Removed `.deepCopy()` in `parseStringListToJsonArray`, adding `ob` directly to `array`.
> Updated `parseStringListToJsonArray_returnsCachedElementWithoutDeepCopy` to assert ... the exact same cached `JsonElement` reference (`first[0] === second[0]`).

Neither is true of what's here. The `deepCopy()` stayed, and the test that asserted `first[0] === second[0]` was replaced by ones asserting `first[0] !== second[0]`. Commit `9d722b101` — "Enforce Achievement parse-cache safety instead of documenting it" — is where the direction reversed, and the metadata never caught up. On a squash-merge that title becomes master's permanent record of a change that does the reverse of what it says.

`Fixes #16811` compounds it: that issue asked for the deepCopy to go, and this PR's own audit concluded it should stay. Merging as-is auto-closes the issue with the opposite resolution and no trace of the reasoning.

**What to change:** retitle to what shipped — something like `achievement: smoother parse cache safety enforcing` — rewrite the body to describe the dead-field removal and the three tests, and either drop the `Fixes #16811` reference or state plainly in the body that the audit reversed the issue's premise so #16811 closes with the actual conclusion recorded.

Two smaller things: the branch is 55 commits behind master (merge base 2026-09-03). The diff is honest — two files, no absorbed history — so this isn't the situation #16647 is in, but a base merge before landing is cheap insurance. And the six commits include three identically-titled ones plus the reversal; squash-merge handles that, so it only matters if this ever lands another way.

CI green 5/5.

**Verdict:** needs another round — the code is mergeable, the title and body are not.

---

dogi:
**Re-review — rating: 66/100** (was 79)

My title/body complaint is resolved — but by the code changing back to match the description rather than the description being updated. That's a legitimate way to close the gap, and the metadata is consistent now. The problem is that the code went back to the weaker of the two designs this PR has tried, and picked up a new defect on the way.

Commit `4237a517f` removes `deepCopy()` again and scopes cache keys instead:

```kotlin
val ownerKey = if (_id.isNotEmpty()) _id else System.identityHashCode(this).toString()
val cacheKey = "$ownerKey#$fieldName#$s"
...
array.add(ob)
```

**1. The `identityHashCode` fallback re-opens exactly what the scoping is meant to close.** `System.identityHashCode` is neither unique nor stable across object lifetimes — values are reused after collection. For an `Achievement` with an empty `_id`, a live object can therefore hit a cache entry left by a *dead, unrelated* object that happened to share an identity hash, the same field name and the same content string, and it now receives that shared mutable instance with no `deepCopy()` behind it. That's the cross-record aliasing this change exists to prevent. It's low-probability, but it's introduced by the fix rather than pre-existing.

It's also useless as a cache key in the ordinary case: every Room read materialises a new `Achievement`, so an `_id`-less record gets a fresh namespace on every load. Those entries can never be hit again — they only evict useful ones.

**2. The cache gets substantially worse at being a cache.** `parsedJsonCache` is capped at `CACHE_CAPACITY = 1000`. Content-addressed, one entry served every record containing that string. Keyed by `id#field#content`, the same string held by fifty achievement records now occupies fifty slots, and each key carries the id and field name on top of the full serialized content. Higher memory per entry, far more entries, much faster thrash — against a cache whose entire purpose is avoiding repeated Gson parses.

**3. `cachedReferencesArray` is back.** Commit `dda167975` deleted it two commits ago; it's re-added here, and it's still written in `setReferences` and read nowhere. I re-grepped the current head: declaration at `Achievement.kt:36`, reset at `:102`, no reads. That was a good cleanup and it's been undone.

**4. Safety now rests on the caller audit instead of on the structure.** Worth weighing what `deepCopy()` actually cost: `JsonPrimitive.deepCopy()` returns `this`, so `links` and `otherInfo` — string arrays — paid nothing for it. Only object-valued entries paid, and those are precisely the ones where handing out a shared mutable instance is dangerous. The previous revision's argument, and its tests, made that case; this reverses it in favour of a caller audit that has to stay true for every future caller of four public accessors.

This is the third direction on the same forty lines. If the aim is to keep the scoping, the minimum is: drop the `identityHashCode` fallback and simply don't cache when `_id` is empty, and delete `cachedReferencesArray` again. But it's worth asking first whether the scoping buys anything that per-read `deepCopy()` wasn't already providing more cheaply and more robustly — the earlier revision looked closer to right.

CI green 5/5.

**Verdict:** needs another round.

---

dogi:
**Re-review — rating: 78/100** (was 66)

**The code is now right.** Commit `d3f5542c8` restores the content-addressed cache with per-read `deepCopy()`, drops the `identityHashCode` fallback, and re-deletes `cachedReferencesArray` (I re-grepped the head — no occurrences left). The comment it leaves behind states the invariant accurately, including the part that makes the cost argument work:

> deepCopy() keeps a cached instance from ever escaping; it is free for primitives and nulls, which return the receiver, so only mutable objects and arrays are copied, exactly the ones that are unsafe to share.

That's correct — `JsonPrimitive.deepCopy()` and `JsonNull`'s both return the receiver. Every point from the last round is addressed, and the net production diff against master is now exactly two things: delete a dead field and its reset, and document why the copy is there. Everything else is tests. After eight commits and four reversals, that's a good place to land — the audit's real conclusion was that the optimisation wasn't safe to make, and the PR now says so in code.

**What's left is the metadata, and it's now unambiguous.** The title is still:

> Stop per-element deepCopy in Achievement parseStringListToJsonArray

and the body still says "Removed `.deepCopy()` in `parseStringListToJsonArray`, adding `ob` directly to `array`" and describes a test asserting `first[0] === second[0]`. The shipped code keeps `deepCopy()` and the shipped tests assert the opposite. This is the same mismatch I raised two rounds ago; it briefly resolved itself when the code swung to match the description, and swung back without the description following.

Two concrete consequences, which is why I'm not waving it through on the strength of the diff:

1. On squash-merge that title becomes master's permanent record of a commit that does the reverse of what it says. In a repo with a title convention this deliberate, that's a lasting wrong entry.
2. `Fixes #16811` auto-closes an issue that asked for the deepCopy to be removed, with a resolution of "we investigated and it stays." Closed silently, the next person to have this idea finds a closed issue and no reasoning.

Nothing in the code needs another round — this is a title edit and a body rewrite. Suggested title in the repo's convention: `achievement: smoother parse cache safety documenting`. For the body, describe the dead-field removal and the three tests, and either drop the `Fixes #16811` line or keep it while stating plainly that the audit reversed the issue's premise, so it closes with the actual finding on record.

CI green 5/5.

**Verdict:** the diff is mergeable; the PR isn't until the title, body and issue link match what it does.

---

dogi:
**Re-review — rating: 91/100** (was 78)

The blocker is cleared. Head is unchanged at `d3f5542c8` — the code was already right last round — and the title and body now describe it accurately:

> achievement: smoother parse cache safety documenting (fixes #16811)

I checked the body against the tree rather than taking it at face value. The three test names it lists are the three that exist (`neverHandsOutTheCachedInstance` at `:83`, `isolatesRecordsWithIdenticalContent` at `:103`, `copyIsFreeForImmutablePrimitives` at `:125`), `parsedJsonCache_isBoundedToCapacity` is still there at `:196`, and "the production diff against master is two things" matches the actual diff exactly — delete `cachedReferencesArray` and its reset, add the comment. The `aload_0; areturn` claim about `JsonPrimitive.deepCopy()` is right too; Gson's implementation is `return this`, which is what makes the cost argument hold for the string-valued fields.

The **Issue resolution** section is the part that does the real work here. `Fixes #16811` now closes that issue with its finding stated plainly — the audit reversed the premise, the optimisation isn't safe to make — instead of closing it silently under a title claiming the opposite. That was the concrete harm I was holding this for, and it's addressed at the source rather than papered over.

Two things worth saying about the reasoning, since it'll outlive this PR:

- "A caller audit is the wrong thing to hang this on" is the right framing, and it's the argument that actually settles this. `parsedJsonCache` is process-wide static mutable state behind four public accessors; an audit that's true today converts into an unenforced invariant every future caller has to preserve, with cross-record corruption as the failure mode. The two intermediate revisions of this PR are themselves evidence for that.
- Dismissing branch 3 (copy once at insert, share the element) on the grounds that it "protects the cached parse from its source, not the caller from other callers" is correct and not obvious.

One residual, not blocking: the branch is 82 commits behind master. The diff is honest — two files, no absorbed history, and `mergeable_state` is `clean` — so this isn't the situation #16647 and #15951 are in. A base merge before landing is still cheap insurance given the gap.

CI green 5/5 on the unchanged head.

**Verdict:** mergeable.

---

Okuro3499 (approved):

J-S-webskas (approved):

PR #16837, opened 2026-09-04
teams: smoother voices image viewer utils gif checking (fixes #16822)

### LINKED ISSUE #16822: Use case-insensitive endsWith for the GIF checks
## What

Two GIF checks lowercase the entire image path just to test a four-character suffix — and do it locale-dependently.

**Files:**
- `app/src/main/java/org/ole/planet/myplanet/utils/ImageViewerUtils.kt` (line 38, `showZoomableImage`)
- `app/src/main/java/org/ole/planet/myplanet/ui/voices/VoicesActions.kt` (line 93, `addImageWithRemoveIcon`)

## Why

```kotlin
imagePath.lowercase(Locale.getDefault()).endsWith(".gif")
```

This allocates a lowercase copy of the whole path on every image binding, and `Locale.getDefault()` makes it locale-dependent — in a Turkish locale, `lowercase` maps `I` to `ı`, so a path like `PHOTO.GIF` still works but the operation is doing locale-sensitive case mapping for no reason on a path that isn't locale text.

`String.endsWith(suffix, ignoreCase = true)` is allocation-free and locale-independent.

## The fix

Replace both with `endsWith(".gif", ignoreCase = true)`.

**Imports:**
- Remove `java.util.Locale` from `ImageViewerUtils.kt` (line 11) — it becomes unused.
- **Keep** it in `VoicesActions.kt` (line 20) — `dateFormatter` at line 33 still uses `Locale.getDefault()`.

Leave the Glide `asGif` / `load` / `error` branches alone.

## Acceptance

- `.gif`, `.GIF` and `.Gif` all still take the GIF branch.
- `VoicesActions` still compiles with its `Locale` import intact.

## Notes

dogi:
Replaced `imagePath.lowercase(Locale.getDefault()).endsWith(".gif")` with `imagePath.endsWith(".gif", ignoreCase = true)` in `ImageViewerUtils.kt` and `VoicesActions.kt`. Removed unused `java.util.Locale` import from `ImageViewerUtils.kt` while retaining it in `VoicesActions.kt` where it is still used by `dateFormatter`.

Fixes #16822

---
*PR created automatically by Jules for task [4146624487418360285](https://jules.google.com/task/4146624487418360285) started by @dogi*


dogi:
**Review — rating: 92/100**

This is a bug fix, not just a tidy-up, and worth recording as one.

`imagePath.lowercase(Locale.getDefault()).endsWith(".gif")` is the classic Turkish-I hazard. Under a Turkish or Azeri locale, `'I'` lowercases to dotless `'ı'` (U+0131), so a file named `photo.GIF` becomes `photo.gıf` — which does not end with `".gif"`. The check silently fails, `request.asGif()` is skipped, and the image loads through the static path: animated GIFs render as a single frame. `Locale.getDefault()` is the *device* locale rather than the app's, so this bites users on Turkish-locale phones even though the app itself only ships ar/es/fr/ne/so — no Turkish translation needed to hit it.

`endsWith(".gif", ignoreCase = true)` does a locale-independent case-insensitive comparison, which is what an extension check always wanted. It also stops allocating a lowercased copy of the path on every call, and lets the now-unused `java.util.Locale` import go from `ImageViewerUtils`.

Both call sites are converted, so the two GIF-detection paths can't disagree.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16838, opened 2026-09-04
dashboard: smoother bell reminders view modelling (fixes #16821)

### LINKED ISSUE #16821: Build the survey reminder ids and pending rows in single-pass collections
## What

`handleDueReminders` chains several collection operators where one traversal would do.

**File:** `app/src/main/java/org/ole/planet/myplanet/ui/dashboard/BellDashboardViewModel.kt` (`handleDueReminders`)

## Why

```kotlin
val allSurveyIds = remindersToShow.flatMap { it.split(",") }.filter { it.isNotBlank() }.distinct()   // 69
...
val pendingSurveys = surveyIdList.mapNotNull { submissionsById[it] }.filter { it.status == "pending" }  // 79
```

Line 69 builds three intermediate collections to produce one id set. Line 79 builds two per reminder group, and it runs once per group.

## The fix

- Build all non-blank survey ids into an **insertion-ordered set** in one traversal, and pass it to the existing bulk submission lookup unchanged.
- Build each group's pending list in one pass, combining the lookup and the status filter.

## Acceptance

Preserve exactly:
- duplicate suppression for the bulk query (that's what `distinct()` was buying — an ordered set keeps it)
- per-group order
- empty-group skipping
- the emitted `SurveyPrompt` contents

## Notes

- The bulk lookup itself must stay one call — this is about the collection work around it, not the query.

dogi:
Refactor handleDueReminders in BellDashboardViewModel to collect survey IDs into an insertion-ordered LinkedHashSet in a single pass and filter pending submissions per group using a single mapNotNull pass.

Fixes #16821

---
*PR created automatically by Jules for task [16629613487759710839](https://jules.google.com/task/16629613487759710839) started by @dogi*


dogi:
**Review — rating: 86/100**

Both changes are behaviour-preserving and each removes a full intermediate list.

The first replaces `flatMap { it.split(",") }.filter { it.isNotBlank() }.distinct()` — which materialises three lists, one of them holding every id across every reminder before deduplication — with a single nested loop accumulating straight into a `LinkedHashSet`. That's equivalent: `distinct()` keeps first-occurrence order and so does `LinkedHashSet`, and the blank filter moved inside the loop unchanged. Only `toList()` at the call boundary allocates, because `getSubmissionsByIds` wants a `List`.

The second fuses a map and a filter: `mapNotNull { submissionsById[it] }.filter { it.status == "pending" }` becomes `mapNotNull { submissionsById[id]?.takeIf { it.status == "pending" } }`. Same result — a missing id and a non-pending submission both yield null and are dropped — in one pass instead of two, and it reads more directly as "the pending submission for this id, if there is one".

This runs on the dashboard bell path where reminders are checked, so trimming allocations there is worth having even though the collections are small.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16841, opened 2026-09-04
courses: smoother steps view modelling (fixes #16820)

### LINKED ISSUE #16820: Hoist CourseStepFragment's multi-repo orchestration into a CourseStepViewModel
## What

A Fragment orchestrates five repositories directly inside `lifecycleScope`.

**Files:**
- `app/src/main/java/org/ole/planet/myplanet/ui/courses/CourseStepFragment.kt`
- `app/src/main/java/org/ole/planet/myplanet/ui/courses/CourseStepViewModel.kt` (new)

## Why

The fragment field-injects `ConfigurationsRepository` (line 48) and `ProgressRepository` (line 50), and via `BaseContainerFragment` also uses `coursesRepository`, `userRepository` and `resourcesRepository`:

`saveCourseProgress` (90), `getCourseStepData` (103), `getUserModel` (109), `getCourseTitleById` (111), `checkServerAvailability` (205), `downloadResourcesPriority` (208), `getAllStepResources` (221, 234).

Five-repository orchestration in a view class — untestable, and the loading logic can't be exercised without inflating a Fragment.

## The fix

Create a `@HiltViewModel CourseStepViewModel` injecting those repositories (plus the download coordinator if the download start moves too), exposing **one** `StateFlow` UI state: step, resources, exams, survey, title, `userHasCourse`, download flags.

**Leave in the fragment:** view binding, markdown/AI selection, navigation, adapter wiring. The fragment collects state instead of launching repository calls.

## Acceptance

- Unit tests for the load, save and empty-step paths.
- Screen behaves identically, including the download start and step navigation.

## Notes

⚠️ This is the largest task in the current batch — **budget for it and keep it to this one screen.**

Hard scope limits:
- **Do not edit any `*RepositoryImpl`.**
- **Do not edit `BaseContainerFragment`** — other screens depend on it, and widening this change there would make it unreviewable.

dogi:
Refactor CourseStepFragment to delegate multi-repository orchestration (data loading, course progress saving, auto-downloading resources, and prefetching next step resources) to a dedicated @HiltViewModel CourseStepViewModel. Expose UI state via StateFlow and collect state inside repeatOnLifecycle in CourseStepFragment. Add comprehensive unit tests in CourseStepViewModelTest.

Fixes #16820

---
*PR created automatically by Jules for task [10903343722838166053](https://jules.google.com/task/10903343722838166053) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 7049
-        versionName = "0.70.49"
+        versionCode = 7050
+        versionName = "0.70.50"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 7045
-        versionName = "0.70.45"
+        versionCode = 7049
+        versionName = "0.70.49"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/MainApplication.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/MainApplication.kt
@@ -6,7 +6,6 @@ import android.content.Context
 import android.content.Intent
 import android.content.SharedPreferences
 import android.content.res.Configuration
-import android.net.TrafficStats
 import android.os.StrictMode
 import android.os.StrictMode.VmPolicy
 import android.provider.Settings
@@ -26,15 +25,11 @@ import androidx.work.WorkManager
 import dagger.hilt.android.EntryPointAccessors
 import dagger.hilt.android.HiltAndroidApp
 import java.lang.ref.WeakReference
-import java.net.HttpURLConnection
-import java.net.URL
-import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.TimeUnit
 import java.util.concurrent.atomic.AtomicBoolean
 import javax.inject.Inject
 import javax.inject.Provider
 import kotlinx.coroutines.CancellationException
-import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.flow.launchIn
 import kotlinx.coroutines.flow.onEach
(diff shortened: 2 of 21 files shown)


dogi:
**Review — rating: 76/100**

The direction is right and the scale is real: 227 lines come out of `CourseStepFragment`, orchestration that was previously only reachable by inflating a fragment now sits behind `CourseStepUiState`, and 306 lines of `CourseStepViewModelTest` arrive with it. Cancelling `loadDataJob` before starting a new load, and guarding re-entrant saves with `saveInProgressJob?.isActive`, are both the right instincts.

Two things I'd fix before this lands.

**The new ViewModel reaches for `MainApplication.context`** (line 68) to get `getExternalFilesDir(null)` for the markdown base path. That's a static handle on the `Application` inside a class whose whole purpose here is to be testable without one — and it's the exact pattern three sibling PRs in this same queue exist to remove: #16783 (`MainApplication.isServerReachable` → injected provider), #16804 (`MainApplication.applicationScope` → `@ApplicationScope`), #16863 (`MainApplication.context` out of model serializers). Adding a fresh instance while those are landing means the next sweep has to come back for it. `@ApplicationContext` through the constructor, or better a path off `FileUtils`/a repository, keeps the rest of the refactor intact. The `try/catch` around it is a hint that reaching for a global there already felt fragile.

**`loadStep(stepId, stepNumber, nextStepId)` never uses `stepNumber`.** It appears in the signature and nowhere in the body — `saveCourseProgress(stepNumber)` is a separate method with its own parameter. Kotlin doesn't warn on unused value parameters, so it compiles clean and every caller has to supply a number that goes nowhere. Either drop it or store it in the state so `saveCourseProgress` can stop taking it.

**Worth confirming rather than blocking:** the two inner `viewModelScope.launch` blocks for prefetching are deliberately *not* children of `loadDataJob`, so a fast step-to-step swipe cancels the load but leaves the downloads running. That's probably what you want for prefetch, but it also means `isDownloadingResources = true` is written by a load that may already be superseded, and nothing clears it except the fragment calling `refreshInlineResources`. If a download fails or the fragment never refreshes, the flag stays true.

**Verdict:** needs another round — inject the context instead of reaching for `MainApplication`, and drop the unused `stepNumber`. Both are small, and the rest of this is a solid extraction.

---

dogi:
**Re-review — rating: 88/100** (was 76)

Both blocking points are fixed, and one of them better than I suggested.

**The static context is gone.** `@ApplicationContext private val context: Context` now comes through the constructor, and `MainApplication` isn't imported at all. Dropping the `try/catch` that wrapped the old global lookup is the right follow-through — it was there because reaching for `MainApplication.context` from a ViewModel felt unsafe, and `getExternalFilesDir` signals "unavailable" by returning null, which the `?.` already handles. This also brings the file in line with #16783, #16804 and #16863, which removed the same pattern elsewhere.

**`stepNumber` is off `loadStep`.** The signature is now `loadStep(stepId, nextStepId)`, and the parameter survives only on `saveCourseProgress(stepNumber)` where it's genuinely consumed (line 134). Callers no longer pass a value that goes nowhere.

The rest of the extraction is unchanged and still the strong part: ~227 lines out of `CourseStepFragment`, `loadDataJob?.cancel()` before each load, the `saveInProgressJob?.isActive` re-entrancy guard, and a ViewModel that is now constructible in a test with no Android globals. CI is green on both flavors with the new `CourseStepViewModelTest`.

**Still open from last round, and still not blocking:** the two prefetch `viewModelScope.launch` blocks aren't children of `loadDataJob`, so a fast step-to-step swipe cancels the load but leaves the downloads running — probably what you want for prefetch — and `isDownloadingResources = true` can be written by a load that's already superseded, with nothing clearing it except the fragment calling `refreshInlineResources`. Worth a comment saying that's deliberate.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16842, opened 2026-09-04
chat: smoother detail ai models caching (fixes #16825)

### LINKED ISSUE #16825: Cache the parsed ai_models map in ChatDetailFragment
## What

`getModelsMap` re-reads a preference string and Gson-parses the whole map on every call, and one caller calls it twice over.

**File:** `app/src/main/java/org/ole/planet/myplanet/ui/chat/ChatDetailFragment.kt` (`getModelsMap` line 596, `getCachedProviderAvailability` line 605)

## Why

```kotlin
private fun getModelsMap(): Map<String, String> {
    val modelsString = sharedPrefManager.getRawString("ai_models").takeIf { it.isNotEmpty() }   // 597
    ...  // full Gson parse, every call
}
private fun getCachedProviderAvailability(): Map<String, Boolean>? {
    val modelsMap = getModelsMap()   // 606 — parses again
```

Call sites at 387, 433, 443, 480 and 606 — so a single screen interaction can re-parse the same JSON several times. The name `getCachedProviderAvailability` already promises caching that isn't there.

## The fix

Cache the last **raw string** plus the parsed map in fragment fields, and invalidate when the raw string changes — that way a mid-session preference change is still picked up rather than being pinned to the first read. Make `getCachedProviderAvailability` use the cached map once.

## Acceptance

- Repeated calls with an unchanged preference parse once.
- Changing `ai_models` mid-session is reflected on the next call (this is the case a naive one-shot cache breaks — please test it).
- Provider/model selection behaves identically.

## Notes

⚠️ Scope: **do not touch `ChatViewModel`, the adapters, or `JsonUtils`.**

dogi:
Caches the parsed `ai_models` JSON map in `ChatDetailFragment` fields to avoid redundant preference reads and Gson parses across repeated calls during a screen interaction. Invalidation occurs automatically when the preference raw string changes mid-session. Added unit tests in `ChatDetailFragmentTest` to verify caching and invalidation behavior.

Fixes #16825

---
*PR created automatically by Jules for task [8994094862632025171](https://jules.google.com/task/8994094862632025171) started by @dogi*


dogi:
**Review — rating: 86/100**

The cache is correct by construction, which is the main thing to get right here: the key *is* the raw preference string, so any edit to `ai_models` produces a different `modelsString`, misses the cache, and reparses. There's no separate invalidation signal that could be forgotten, and no window where a stale map survives a pref change. Clearing both fields in `onDestroyView` alongside the other `last*` caches keeps it consistent with how the fragment already handles retained state.

Worth having because `getModelsMap()` is called from the provider/model selection paths rather than once at setup, and each call was running a Gson `TypeToken<Map<String, String>>` reflective parse of the whole models blob. Comparing two strings is far cheaper than reparsing, and `String.equals` short-circuits on length before comparing contents.

The `?: emptyMap()` is a real fix rather than noise: `gson.fromJson` returns null for a literal `"null"` payload, and the old code declared a non-null `Map<String, String>` return while handing that null straight back — a latent NPE at the call site for a malformed pref. Now it degrades to empty.

**Nit:** `return cachedModelsMap!!` immediately after the null check works but reads as an assertion the compiler can't help with; a local `val cached = cachedModelsMap` then `if (modelsString == cachedRawModelsString && cached != null) return cached` smart-casts instead.

⚠️ **No build or test verdict on this head** — only `myPlanet labels` ran; neither `build` nor `test` fired, so `ChatDetailFragmentTest` hasn't executed anywhere.

---

Okuro3499 (approved):

PR #16843, opened 2026-09-04
community: smoother services routing (fixes #16828)

### LINKED ISSUE #16828: Extract community-service route resolution into a tested pure function
## What

Navigation routing rules are parsed inline inside a click listener, interleaved with navigation and membership checks, with zero tests.

**Files:**
- `app/src/main/java/org/ole/planet/myplanet/ui/community/CommunityServicesFragment.kt` (`setRecyclerView`)
- `app/src/main/java/org/ole/planet/myplanet/ui/community/CommunityServiceRoute.kt` (new)

## Why

The listener does an http/https prefix check, then `rawRoute.split("/")` taking `segments[3]` as a team id when `size >= 4`, else falling back to a WebView — all mixed in with the navigation and membership-check code.

That's navigation policy buried in a view. It can't be tested, and the segment-index arithmetic is exactly the kind of thing that breaks silently on an unexpected route shape.

## The fix

Add a sealed `CommunityServiceRoute` (`ExternalLink` / `TeamLink` / `Unhandled`) and a **pure** `resolve(route)` reproducing today's rules exactly. Switch the listener to dispatch on it, keeping the same resulting actions.

## Acceptance

Tests on `resolve`:
- an `https://` URL → `ExternalLink`
- an `http://` URL → `ExternalLink`
- a `/teams/view/<id>`-style route → `TeamLink` with the right id
- a route with fewer than 4 segments → `Unhandled`
- the empty string → `Unhandled`

Behavior at the call site unchanged, including the membership check.

## Notes

⚠️ Two other queued changes also touch this fragment (a `LayoutInflater` hoist and a ViewModel extraction). This one goes first — please keep it to route resolution only, so the others can rebase cleanly on top.

dogi:
Extracts community-service route resolution from CommunityServicesFragment into a tested, pure sealed class CommunityServiceRoute with variants ExternalLink, TeamLink, and Unhandled.

Fixes #16828

---
*PR created automatically by Jules for task [11636200245013854980](https://jules.google.com/task/11636200245013854980) started by @dogi*


dogi:
**Review — rating: 90/100**

Routing logic buried in a click listener inside a `forEach` inside `onViewCreated` is about as untestable as code gets, and lifting it into a pure `resolve(route)` over a sealed class makes it a plain function call in a JVM test. The `CommunityServiceRouteTest` that comes with it is only possible because of that.

I checked the branch-for-branch equivalence, including the case that usually breaks this kind of extraction. Old: `teamId = if (segments.size >= 4) segments[3] else null`, then `if (teamId != null)` → team detail, else → WebView. New: `size >= 4` → `TeamLink(segments[3])`, else `Unhandled` → WebView. The subtle case is `segments[3]` being an empty string — for a route like `"a/b/c/"`, the old code produced `teamId = ""`, which is non-null and so went to `TeamDetailFragment`; `TeamLink("")` does the same. The rewrite doesn't accidentally tighten that into a blank check, so behaviour is identical.

The `ExternalLink` and `Unhandled` arms both open `WebViewActivity` with the same extras, matching the old duplicated blocks, and the http/https prefix test is unchanged.

Making `Unhandled` a `data object` rather than `object` is right for a no-payload sealed member on Kotlin 2.4.

**Nit:** `resolve` faithfully preserves the pre-existing oddity that a blank fourth segment routes to a team detail screen with an empty id. Now that the logic is a tested pure function, adding `segments[3].takeIf { it.isNotBlank() }` and falling through to `Unhandled` would be a one-line improvement with a test to prove it — but it's a behaviour change, so it belongs in its own commit rather than smuggled into an extraction.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16847, opened 2026-09-04
enterprises: smoother reports name resolving (fixes #16827)

### LINKED ISSUE #16827: Stop EnterprisesReportsFragment reaching into TeamsRepository for the team name
## What

The enterprises screen calls into `TeamsRepository` three times for a team name it already has.

**File:** `app/src/main/java/org/ole/planet/myplanet/ui/enterprises/EnterprisesReportsFragment.kt` (lines 69, 84, 114)

## Why

```kotlin
val teamName = teamsRepository.getTeamNameFromPrefs()?.replace(" ", "_")                       // 69  (CSV filename)
val csvContent = viewModel.exportReportsAsCsv(teamId, teamsRepository.getTeamNameFromPrefs() ?: "")  // 84  (CSV content)
teamsRepository.getTeamNameFromPrefs(),                                                        // 114
```

Three prefs-backed lookups — a cross-feature leak from enterprises into Teams for a value the screen already knows. `BaseTeamFragment.getEffectiveTeamName()` (line 90) resolves it from the fragment's arguments or the loaded team.

## The fix

Replace all three with **one** read of `getEffectiveTeamName()` from the base class. That is the smallest change and drops the `TeamsRepository` dependency entirely (routing through the ViewModel instead would keep the prefs coupling).

Remove the `teamsRepository` reference from this file if the name was its only use. Keep the filename sanitising (`replace(" ", "_")`) in the UI.

## Acceptance

⚠️ **This is a deliberate change of source of truth**, from prefs to arguments/loaded-team. Those two can differ — that's the point, since the screen's own team is the correct answer — but it means the export test must cover it: assert the exported filename **and** the CSV content carry the effective team name for the team actually being viewed.

- All three sites read the same value.
- `TeamsRepository` no longer referenced from this fragment.

## Notes

dogi:
Replaces direct lookups of teamsRepository.getTeamNameFromPrefs() with getEffectiveTeamName() in EnterprisesReportsFragment. Drops TeamsRepository dependency usage from the fragment and adds unit test coverage.

Fixes #16827

---
*PR created automatically by Jules for task [477160094988580721](https://jules.google.com/task/477160094988580721) started by @dogi*


dogi:
**Review — rating: 90/100**

This is a correctness fix wearing a layering fix's clothes, and the difference is worth spelling out.

`teamsRepository.getTeamNameFromPrefs()` reads the *globally last-selected* team out of SharedPreferences. `getEffectiveTeamName()` (`BaseTeamFragment:90`) resolves `requireArguments().getString("teamName")?.takeIf { it.isNotBlank() } ?: team?.name ?: ""` — the team this fragment is actually showing. Those diverge whenever the two disagree: open team A's reports while the pref still holds team B and, on master, the CSV filename, the `"<teamName> Financial Report Summary"` header written into the export, and the adapter's report title all name team B. All three call sites are converted, so the exported file and what's on screen can no longer disagree about which enterprise they describe.

It also completes a set — `getEffectiveTeamType()` and `getEffectiveTeamId()` already existed on the base class with exactly this precedence, and the team name was the odd one still going out to prefs.

The nullability tidy-up falls out of it: `getEffectiveTeamName()` returns non-null `String`, so the `?: ""` fallback at the CSV call and the `?.` on `replace(" ", "_")` are no longer needed, and the adapter gets a guaranteed non-null name.

**Verdict:** mergeable.

---

PR #16848, opened 2026-09-04
teams: smoother voices images view modelling (fixes #16832)

### LINKED ISSUE #16832: Parse each news item's images JSON once in downloadReferencedResources
## What

The feed download loop dereferences `imagesArray` twice per item, and that getter re-parses JSON on every access.

**File:** `app/src/main/java/org/ole/planet/myplanet/ui/voices/VoicesViewModel.kt` (`downloadReferencedResources`, lines 221-222)

## Why

```kotlin
if (news?.imagesArray?.isEmpty() == false) {          // 221 — parse #1
    val ob = news?.imagesArray?.get(0)?.asJsonObject  // 222 — parse #2
```

`News.imagesArray` re-parses the images JSON with Gson on every access — the getter does no caching. So a feed of N posts with images triggers **2N** full JSON parses every time the feed loads.

## The fix

Hoist `val images = news?.imagesArray` once per item and use the local for both the emptiness check and the element access.

Keep the `resourceIds` dedup set and the repository calls unchanged.

## Acceptance

- Same resources downloaded for the same feed.
- One `imagesArray` access per item.

## Notes

- This is complementary to the queued News parse-cache change: that one fixes the getter, this one stops the double call. Either alone helps; both are worth having. They touch different files, so no coordination is needed.

dogi:
Hoist news.imagesArray access into a local variable in downloadReferencedResources to avoid double parsing JSON per news item.

Fixes #16832

---
*PR created automatically by Jules for task [5478678338928969174](https://jules.google.com/task/5478678338928969174) started by @dogi*


dogi:
**Review — rating: 88/100**

Small diff, real saving — `imagesArray` is not a field. It's a computed getter:

```kotlin
val imagesArray: JsonArray
    get() = if (images == null) JsonArray() else JsonUtils.gson.fromJson(images, JsonArray::class.java)
```

So every access runs a full Gson parse of the `images` JSON string and allocates a new `JsonArray`. The old code touched it twice per item — once for `isEmpty()`, once for `get(0)` — so a Voices list of *n* items parsed the same JSON `2n` times. Hoisting into `images` halves that, and it's a real cost on a feed rather than a theoretical one.

The null handling is also tidier as a side effect: the original had a redundant `news?` on the second access despite already being inside a null-guarded branch, and `ob?.asJsonObject` on something already typed as a `JsonObject`. The rewrite lets the smart cast do the work.

**Nit:** while you're here, `JsonUtils.getString("resourceId", ob?.asJsonObject)` still has that vestigial `?.asJsonObject` on `ob`, which is already a `JsonObject?`.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16849, opened 2026-09-04
teams: smoother voices model time logging (fixes #16834)

### LINKED ISSUE #16834: Single map read and System.currentTimeMillis in News.updateMessage/createNews
## What

Two small inefficiencies on the path that runs for every news or voice item parsed or updated.

**File:** `app/src/main/java/org/ole/planet/myplanet/model/News.kt` (`updateMessage` line 98, `createNews` line 158)

## Why

1. `createNews` does `if (map.containsKey("news")) { val newsObj = map["news"] ... }` — two lookups in the deserialized map where one suffices.
2. Both `updateMessage` (line 101) and `createNews` (line 167) use `Date().time`, allocating a `java.util.Date` purely to read the current timestamp.

## The fix

- Replace the `containsKey` / `get` pair with `map["news"]?.let { ... }`.
- Swap both `Date().time` reads for `System.currentTimeMillis()`.
- Drop the now-unused `java.util.Date` import — **check first** that nothing else in the file still uses `Date`.

No serialization-format change and no Room-schema change, so **no `AppDatabase` version bump**.

## Acceptance

- Parsed `News` objects are identical for the same input, including the `news` sub-object path and the absent-key path.
- `editedTime` / `time` still populated.

## Notes

- Another open PR removes a dead getter elsewhere in this file (around line 113). These sites are lines 101 and 167, so they shouldn't collide — if you do hit a conflict, rebase rather than reworking.

dogi:
Optimized map read and timestamp generation in News model:
- Replaced `map.containsKey("news")` and `map["news"]` double lookup with single-pass `map["news"]?.let { newsObj -> ... }` in `News.createNews`.
- Replaced `Date().time` object allocations with `System.currentTimeMillis()` in `News.updateMessage` and `News.createNews`.
- Removed unused `java.util.Date` import from `News.kt`.
- Added unit tests in `NewsTest.kt` covering `updateMessage` and `createNews`.

Fixes #16834

---
*PR created automatically by Jules for task [7524528762589674422](https://jules.google.com/task/7524528762589674422) started by @dogi*


dogi:
**Review — rating: 88/100**

Both changes are safe, and the map one is safe for a reason worth stating rather than assuming.

`containsKey(k)` followed by `map[k]` is only equivalent to `map[k]?.let { }` when the map can't hold a null *value* — otherwise the old code entered the block with a null and the new code skips it. Here the signature is `HashMap<String?, String>`: the **values** are non-null (it's the keys that are nullable), so "key present" and "value non-null" are the same condition and the rewrite is exact. It also collapses two hash lookups into one and lets `newsObj.replace("=", ":")` drop its `?.`, since the value is non-null inside the `let`.

`Date().time` → `System.currentTimeMillis()` is identical in value — `Date()` is a wrapper around exactly that call — and skips allocating a `Date` per news creation and per edit. Removing the now-unused `java.util.Date` import is the right follow-through.

Neither change touches the surrounding `try`/`JsonSyntaxException` handling, so malformed `news` JSON still fails the same way.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16851, opened 2026-09-04
all: smoother file utils html cover image finding (fixes #16833)

### LINKED ISSUE #16833: Make FileUtils.findHtmlCoverImage a single lazy pass
## What

`findHtmlCoverImage` materializes every matching image into a list, then scans that list up to two more times.

**File:** `app/src/main/java/org/ole/planet/myplanet/utils/FileUtils.kt` (`findHtmlCoverImage`, line 142)

## Why

The function walks up to four directory levels, calls `.toList()` (line 147) to materialize every matching image, then scans that list for a filename hint and may scan it a third time for the largest file. Large offline HTML resources can hold many images, and this runs on the resource-preview path.

## The fix

Iterate the walk **lazily** instead of converting to a list:

- compute each filename's lowercase form once
- return the first hint match immediately — current semantics already take the first hint in walk order, so this is not a behavior change
- track the largest candidate as you go, for the unchanged fallback

## Acceptance

Preserve exactly:
- the depth limit
- the extension allowlist and case-insensitive extension handling
- walk order
- tie behavior for the largest-file fallback
- `null` for empty or non-directory input

## Notes

- The early return is the main win — the current code can't take it because it has already materialized everything.

dogi:
Refactored FileUtils.findHtmlCoverImage to perform a single lazy pass over resourceDir.walkTopDown().maxDepth(4). Added unit tests in FileUtilsTest to verify non-directory inputs, case insensitivity, max depth limits, and tie-breaking behavior.

Fixes #16833

---
*PR created automatically by Jules for task [3253321603895998909](https://jules.google.com/task/3253321603895998909) started by @dogi*


dogi:
**Review — rating: 88/100**

The real win here is early termination, which is bigger than "one pass". The old version called `.toList()` on the filtered sequence, so it walked the *entire* directory tree to depth 4 and materialised every matching image before looking at the first one — even when the very first file found was a hint match. The rewrite returns the moment a hinted name appears, so a resource whose cover is named `cover.png` near the top stops the walk immediately instead of enumerating the whole bundle.

I checked the three ways this could have changed behaviour, and none of them did:

- **Hint selection.** Old: `images.firstOrNull { hint }` over the list in walk order. New: return on the first hint match during the same `walkTopDown().maxDepth(4)` traversal. Same order, same file.
- **Fallback selection.** Old: `images.maxByOrNull { it.length() }`, which returns the *first* maximal element on ties. New: `length > maxBytes` is strictly greater, so ties also keep the first seen. Identical.
- **Empty case.** Old returned null via the `images.isEmpty()` guard; new leaves `largestFile` null. Same.

Holding one `File` reference instead of a list of every image in the bundle is a worthwhile allocation change too, given this runs per resource.

**Nit:** `file.length()` is now called on every non-hinted image as it's encountered, whereas the old code only reached `length()` if no hint matched at all. For a bundle with no hinted name the count is the same; for one where the hint appears late, the new version does a few extra `length()` stats that the old version's short-circuit... also did, since `maxByOrNull` ran over everything. Net: never worse, usually better.

⚠️ **No build or test verdict on this head** — only `myPlanet labels` ran; neither `build` nor `test` fired, so the accompanying `FileUtilsTest` additions haven't executed.

---

Okuro3499 (approved):

PR #16852, opened 2026-09-04
life: smoother health examination dao batch updating (fixes #16835)

### LINKED ISSUE #16835: Batch the null-rev health markUploaded updates
## What

The map overload of `markUploaded` loops one UPDATE per entry, including for the common all-null-rev case where a single statement would do.

**File:** `app/src/main/java/org/ole/planet/myplanet/data/room/dao/HealthExaminationDao.kt` (lines 30-35)

## Why

```kotlin
suspend fun markUploaded(id: String, rev: String?)               // 30
suspend fun markUploaded(idToRevMap: Map<String, String?>) {     // 33  @Transaction
    idToRevMap.forEach { (id, rev) -> markUploaded(id, rev) }    // 35  one UPDATE per entry
}
```

It's driven from `HealthRepositoryImpl.markHealthExaminationsUploaded` via `UploadToShelfService`, whose common path passes **null** revs for already-synced docs — so the typical batch is N statements that could be one.

## The fix

- Keep the single-row `markUploaded(id, rev)` as is.
- In the map overload, group the **null-rev** ids and batch them with `UPDATE health_examinations SET isUpdated = 0 WHERE _id IN (:ids)`, chunked at 900 exactly as `RemovedLogDao.deleteByTypeUserAndDocsChunked` already does.
- Non-null revs still need per-row updates, since the rev differs per row — don't try to batch those.

No schema change, so no `AppDatabase` version bump.

## Acceptance

A Room in-memory test covering: rows with rev, rows without, a **mixed** batch, and the empty-map no-op.

## Notes

The performance gain here is bounded — these UPDATEs already run inside one `@Transaction`, so this saves statement dispatch, not commits. **The new DAO test is arguably the bigger win**, since this method currently has none and the mixed-batch path is easy to get subtly wrong. Please don't skip it.

dogi:
Optimized HealthExaminationDao.markUploaded(idToRevMap) by partitioning null-rev entries and updating them via a single SQL query `UPDATE health_examinations SET isUpdated = 0 WHERE _id IN (:ids)` chunked in batches of 900. Non-null rev entries continue to be updated per row. Added `HealthExaminationDaoTest` with in-memory Room tests covering rows with rev, rows without rev, mixed batches, and empty map input.

Fixes #16835

---
*PR created automatically by Jules for task [2796603168864650806](https://jules.google.com/task/2796603168864650806) started by @dogi*


dogi:
**Review — rating: 84/100**

**The important part of this PR isn't the batching — it's that the null-rev path stops destroying data, and the description doesn't say so.**

The existing single-row query is `UPDATE health_examinations SET _rev = :rev, isUpdated = 0 WHERE _id = :id`, with no `COALESCE`. So when `idToRevMap` carried a null rev, the old loop wrote `_rev = NULL`, wiping whatever revision the row already held. The new `markUploaded(ids)` sets only `isUpdated = 0` and leaves `_rev` alone.

That path is reachable. `HealthRepositoryImpl` builds the map as `pojo._id to res.body()?.get("rev")?.asString`, so a 2xx response without a `rev` field yields a null. Rare against CouchDB, but the consequence was bad: a health examination whose local `_rev` was nulled can't be updated on the server afterwards, because CouchDB needs the current revision — the record is effectively stranded until a full resync. Preserving `_rev` is the right behaviour and worth calling out as a fix, with a test pinning "null rev leaves `_rev` untouched", rather than landing quietly inside a performance change.

**On the batching itself**, the same caveat applies as to #16677: the branch being batched is the rare one. Every entry with a rev — essentially all of them — still gets its own `UPDATE`, so in normal operation this issues the same number of statements as before. `chunked(900)` is the right guard for the batched query, and `partition` reads cleanly.

**Nit:** `nullRevEntries.isNotEmpty()` is redundant — `chunked(900)` on an empty list produces no chunks, so `forEach` already does nothing.

**Verdict:** mergeable — worth reframing the description around the `_rev` fix, since that's what actually matters here.

---

Okuro3499 (approved):

PR #16858, opened 2026-09-04
enterprises: smoother reports totals factoring (fixes #16839)

### LINKED ISSUE #16839: Extract the report-totals math out of EnterprisesReportsAdapter
## What

Financial arithmetic is computed inline for every bound row, inside a RecyclerView adapter, with no tests.

**Files:**
- `app/src/main/java/org/ole/planet/myplanet/ui/enterprises/EnterprisesReportsAdapter.kt` (lines 55-68)
- its test

## Why

```kotlin
val totalIncome = report.sales + report.otherIncome        // 55
val totalExpenses = report.wages + report.otherExpenses    // 56
val profitLoss = totalIncome - totalExpenses               // 57
...
endingBalanceValue.text = context.getString(R.string.number_placeholder, profitLoss + it.beginningBalance)  // 68
```

Money math in `onBindViewHolder`, recomputed on every bind and every recycle, and untestable where it sits. The same arithmetic also appears in `EnterprisesRepositoryImpl.exportReportsAsCsv` — so the two could drift and nothing would catch it.

## The fix

Add a private `ReportTotals` data class and a **pure** `reportTotals(report)` function in the same file, and call it from `onBindViewHolder`. No behavior change.

Then unit-test income, expenses, profit/loss and ending balance — **including a negative beginning balance**, which is the case most likely to be wrong and impossible to assert today.

## Acceptance

- Displayed values identical for the same report.
- Tests cover the four computed values plus the negative-beginning-balance case.

## Notes

⚠️ Deliberately **do not** unify this with the CSV export path. That duplication is real, but the SQL push-down change owns that side; unifying here would collide with it. Extracting the pure function first is what makes a later unification safe.

This file also has an open PR touching `bindReportImage` (lines ~93-112). Your change is at 55-68, so they shouldn't conflict.

dogi:
Extract report totals calculation math into a pure reportTotals function and ReportTotals data class in EnterprisesReportsAdapter.kt, and add unit test coverage in EnterprisesReportsAdapterTest.kt.

Fixes #16839

---
*PR created automatically by Jules for task [16779978731345447562](https://jules.google.com/task/16779978731345447562) started by @dogi*


dogi:
**Review — rating: 90/100**

I checked all four formulas against the originals and they're preserved exactly: `totalIncome = sales + otherIncome`, `totalExpenses = wages + otherExpenses`, `profitLoss = totalIncome - totalExpenses`, and `endingBalance = profitLoss + beginningBalance` — the last one being the easiest to get wrong, since it was previously written inline as `profitLoss + it.beginningBalance` at the call site rather than as a named value.

Pulling this out is worth more than the usual extraction because it's money arithmetic that feeds a financial report. Inside `onBindViewHolder` it could only be verified by inflating a ViewHolder and reading `TextView` strings back out; as an `internal fun` over `MyTeam` it's a pure function that a plain JVM test can exercise across negative balances, zero income and loss-making periods. Naming `endingBalance` in the returned type also means the report's definition of it lives in one place instead of being re-derived at the binding site.

Computing all four in one call also removes the repeated `report.` reads scattered through the `with(binding)` block, so the binding code now reads as pure presentation.

**Note on #16795:** it also touches this adapter, but in `bindReportImage`; your change is in `onBindViewHolder`. The only conflict between you is `EnterprisesReportsAdapterTest.kt`, so both can land with a small test merge.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16861, opened 2026-09-04
sync: less time logger process key is more (fixes #16844)

### LINKED ISSUE #16844: Drop the double processTimes lookup in SyncTimeLogger.endProcess
## What

`endProcess` looks up the same `ConcurrentHashMap` key twice for every timed sync or upload process.

**File:** `app/src/main/java/org/ole/planet/myplanet/utils/SyncTimeLogger.kt` (`endProcess`, line 132)

## Why

```kotlin
if (!processTimes.containsKey(startKey)) {   // 138
    return
}
val startTime = processTimes[startKey] ?: return   // 142
```

The `?: return` on line 142 already handles the missing-key case, so the `containsKey` guard is redundant — and on a `ConcurrentHashMap` it's a second synchronized lookup, not a free check.

## The fix

Delete the `containsKey` guard and keep only the `?: return` read.

Duration calculation and the `processTimes` / `processItemCounts` writes stay unchanged. Missing start times are still silently ignored, exactly as today.

## Acceptance

- Behavior identical, including the missing-start-time path.
- One map lookup per call.

## Notes

dogi:
Removes the redundant `containsKey` guard in `SyncTimeLogger.endProcess`, reducing double ConcurrentHashMap lookups down to a single read while maintaining identical behavior.

Fixes #16844

---
*PR created automatically by Jules for task [11872290911577922546](https://jules.google.com/task/11872290911577922546) started by @dogi*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling when attempting to end a process that was not started, preventing unnecessary processing.
* **Tests**
  * Added coverage verifying that unstarted processes are ignored and completed processes appear in sync summaries.


dogi:
**Review — rating: 90/100**

Exactly equivalent, and the equivalence holds for the reason that matters: `?:` already covers both cases the `containsKey` guard was covering. Key absent → `processTimes[startKey]` is null → early return. Key present with a null value → also null → early return. Key present with a value → proceeds. The removed `if (!processTimes.containsKey(startKey)) return` could only ever agree with the line below it.

Worth doing beyond tidiness because `endProcess` is called once per sync stage and `processTimes` is a shared map, so this halves the lookups on that path and removes the possibility of the two checks disagreeing if the map were ever mutated concurrently between them.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16862, opened 2026-09-04
all: smoother markdown utils image url resolving (fixes #16846)

### LINKED ISSUE #16846: Quote the img src attribute in MarkdownUtils.prependBaseUrlToImages
## What

Markdown image URLs are emitted into an **unquoted** HTML attribute, so any URL containing a space renders truncated.

**File:** `app/src/main/java/org/ole/planet/myplanet/utils/MarkdownUtils.kt` (`prependBaseUrlToImages`, line 101)

## Why

```kotlin
result.append("``&lt;img src=$fullUrl width=$width height=$height/&gt;``")   // 101
```

`src=` is unquoted. An HTML parser ends the attribute value at the first whitespace, so a resolved URL containing a space — a filename with a space in it, which is entirely possible for user-uploaded course resources — silently loses everything after the space and the image fails to load.

## The fix

Lift the tag into a template with a **quoted** `src` value and format the URL into it. Keep the matcher/`StringBuilder` loop identical.

## Acceptance

- A relative `resources/foo.png` image resolves and renders as it does today.
- **A URL containing a space renders correctly** — that's the regression test that matters here.

## Notes

This was originally filed as a performance/Markwon-caching item. That premise doesn't hold: the Markwon instance is already double-check-locked and the pattern is already compiled once, so there is no caching work to do. The quoting defect is the real content, and it's a correctness bug rather than an optimization — please don't add caching machinery alongside it.

dogi:
Quote the img src attribute value in MarkdownUtils.prependBaseUrlToImages to ensure HTML parsers do not truncate URLs containing spaces. Update existing unit tests and add a new test for URLs containing spaces.

Fixes #16846

---
*PR created automatically by Jules for task [1018986335772378574](https://jules.google.com/task/1018986335772378574) started by @dogi*


dogi:
**Review — rating: 88/100**

One character either side, and it fixes a real class of broken rendering. ``'&lt;img src=$fullUrl …&gt;'`` with an unquoted attribute breaks on the first space in the URL: `src=http://host/my file.png` parses as `src="http://host/my"` followed by a junk attribute named `file.png`, and the image silently fails to load. Resource filenames in this app come from server-side content, so spaces are entirely plausible. Unquoted values are also terminated by `>`, `"`, `'` and backtick, so any of those in a path breaks the tag structure rather than just the URL.

It's a small hardening win too: with `src` unquoted, a crafted path could inject an additional attribute into the tag, whereas a quoted value can only escape via a literal `"`. The content is app-assembled markdown rather than arbitrary user HTML, so this isn't a live XSS, but quoting removes the shape of the problem.

**Nit:** `width` and `height` are still unquoted in the same tag. HTML5 tolerates that for bare numeric values, so it works today, but leaving one attribute quoted and two not invites the same bug the next time one of those becomes non-numeric. Quoting all three would make the tag uniformly safe.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16863, opened 2026-09-04
all: smoother models serializer device identity providing (fixes #16840)

### LINKED ISSUE #16840: Pass androidId and customDeviceName into the model serializers instead of reading MainApplication.context
## What

Three model serializers reach into the `MainApplication` companion context to resolve device identity.

**Files:**
- `app/src/main/java/org/ole/planet/myplanet/services/upload/UploadConfigs.kt`
- `app/src/main/java/org/ole/planet/myplanet/model/NewsLog.kt` (line 36)
- `app/src/main/java/org/ole/planet/myplanet/model/Rating.kt` (line 55, via `import MainApplication.Companion.context`)
- `app/src/main/java/org/ole/planet/myplanet/model/SearchActivity.kt` (lines 38, 40)

## Why

```kotlin
"customDeviceName", NetworkUtils.getCustomDeviceName(MainApplication.context)   // NewsLog:36
ob.addProperty("customDeviceName", NetworkUtils.getCustomDeviceName(context))   // Rating:55
obj.addDocumentOrigin(VersionUtils.getAndroidId(MainApplication.context))       // SearchActivity:38
```

Model serializers depending on the Application singleton — untestable without it, and resolved per document rather than once per upload batch.

`UploadConfigs` already injects `@ApplicationContext Context` (line 41) and `SharedPrefManager` (line 46), so it can compute both values once and pass them in.

## The fix

Compute `androidId` and `customDeviceName` **once** in `UploadConfigs` and thread them through the three serializer lambdas. Change the three signatures to accept them, and drop the `MainApplication` import — plus `VersionUtils` / `NetworkUtils` imports where they become unused.

## Acceptance

⚠️ Uploaded payloads must keep the **same** `androidId`, `deviceName` and `customDeviceName` field values. This is a plumbing change; if any emitted field differs, something is wrong.

## Notes

**Scope is three files, not four.** An earlier version of this finding also listed `model/ApkLog.kt` — that one is already migrated: `ApkLog.serialize` takes `customDeviceName` as a parameter and calls `addDocumentOrigin()` with no argument. Use it as the reference shape for the other three, but don't modify it.

dogi:
Pass androidId and customDeviceName into model serializers from UploadConfigs instead of reading MainApplication.context directly in model companion objects.

Fixes #16840

---
*PR created automatically by Jules for task [2982894873616598467](https://jules.google.com/task/2982894873616598467) started by @dogi*


dogi:
**Review — rating: 84/100**

Good direction: `NewsLog`, `Rating` and `SearchActivity` were Room entities reaching for `MainApplication.context` from a static `serialize`, which made them impossible to serialize in a unit test without standing up the application. Taking the values as parameters leaves them as plain data, and it matches `ApkLog.serialize(log, customDeviceName)`, which already had this shape — so the four serializers are now consistent rather than three doing it one way and one the other.

Swapping `NetworkUtils.getCustomDeviceName(context)` for `sharedPrefManager.getCustomDeviceName()` is exactly equivalent, which is worth stating because it looks like a source change: `NetworkUtils.getCustomDeviceName` is literally `return sharedPrefManager.getCustomDeviceName()` (line 225-227) and ignores the `context` argument entirely. Calling the pref directly removes a parameter that never did anything.

**The `by lazy` on `customDeviceName` is the one thing I'd change.** `UploadConfigs` is `@Singleton`, so the value is resolved once and pinned for the process. The custom device name is a user-editable setting (`SharedPrefManager.setCustomDeviceName`), so if someone renames their device, every subsequent upload still carries the old name until the app is restarted — where previously it was read at each serialization. It isn't worth caching either: the read is `pref.getString(...)`, which is served from the in-memory preference map. A getter keeps the current behaviour at no cost:

```kotlin
private val customDeviceName: String get() = sharedPrefManager.getCustomDeviceName()
```

`androidId` is a fair `by lazy` — `ANDROID_ID` can't change while the process lives, so caching it avoids a repeated `Settings.Secure` lookup with no staleness risk.

**Verdict:** mergeable — swap the `lazy` for a getter so a renamed device takes effect without a restart.

---

Okuro3499 (dismissed):
```
Problem: private val customDeviceName: String by lazy { sharedPrefManager.getCustomDeviceName() } caches the value for the lifetime of the @Singleton UploadConfigs instance     
(i.e., the whole app process). Every serializer that references customDeviceName (NewsActivities, SearchActivity, ApkLogs, Rating — lines 64, 119, 219, 302) will keep using     
whatever name was current the first time any of them ran a serialize, even after the user renames their device via SyncActivity (prefData.setCustomDeviceName(...),              
SyncActivity.kt:453) later in the same session. This is a behavior change from the pre-refactor code, which called NetworkUtils.getCustomDeviceName(context) fresh on every      
serialize.
```

dogi:
**Re-review — rating: 89/100** (was 84)

This is the right direction and the follow-up commit finished it cleanly. `NewsLog.serialize`, `Rating.serializeRating` and `SearchActivity.serialize` were reaching into `MainApplication.context` — a global Application singleton — from inside `model/`, which coupled entity classes to the Android framework and made them untestable without a running app. They're now pure functions of their arguments, with the device identity threaded in from `UploadConfigs`, which is a `@Singleton` that already holds `context` and `sharedPrefManager`. That's the correct seam.

**I checked the one substitution that could have changed uploaded data, and it's exactly equivalent.** `NetworkUtils.getCustomDeviceName(context)` turns out to be:

```kotlin
    return sharedPrefManager.getCustomDeviceName()
}
```

— a passthrough that ignores its `Context` parameter entirely. So replacing it with `sharedPrefManager.getCustomDeviceName()` reads the same `CUSTOM_DEVICE_NAME` pref through the same object, and the `customDeviceName` field on every uploaded document is byte-identical before and after. It also deletes a pointless indirection that was making `context` look load-bearing when it wasn't. Same story for the `ApkLog` call site, which already passed the value in and now just sources it from the shared property.

Using a `get()` property for `customDeviceName` is right — a user can rename the device between uploads, so it must be re-read. `androidId` is the opposite case: `VersionUtils.getAndroidId(context)` does a `Settings.Secure` content-resolver query, which is IPC, and the value never changes for the install. `by lazy` would read it once instead of once per serialized search activity. That's not a regression — the old code called it per serialization too — just the obvious next line.

CI green 5/5.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16867, opened 2026-09-04
courses: smoother concatenated links saving (fixes #16856)

### LINKED ISSUE #16856: Use addAll and drop the set-to-list copy in MyCourse.saveConcatenatedLinksToPrefs
## What

A per-element `add` loop and a redundant set-to-list copy when persisting course links.

**File:** `app/src/main/java/org/ole/planet/myplanet/model/MyCourse.kt` (`saveConcatenatedLinksToPrefs`, line 92)

## Why

```kotlin
for (link in linksToProcess) {
    existingConcatenatedLinks.add(link)          // 104 — one add per element
}
val jsonConcatenatedLinks = JsonUtils.gson.toJson(existingConcatenatedLinks.toList())  // 106 — copy
```

`addAll` does the first, and `Gson.toJson` accepts a `Collection`, so the `.toList()` copy is pure allocation.

## The fix

- Replace the loop with `addAll(linksToProcess)`.
- Serialize the set directly, dropping `.toList()`.

## Acceptance

- Same persisted JSON content.
- Order-independence is already relied upon — the reader parses the value back as `Array<String>` and calls `toHashSet()` — so serializing the set rather than a list is safe. Please confirm that reader path still round-trips.

## Notes

dogi:
Replaced per-element add loop with addAll(linksToProcess) and dropped unnecessary .toList() conversion when serializing existingConcatenatedLinks Set to JSON in MyCourse.saveConcatenatedLinksToPrefs. Added unit test in MyCourseTest.

Fixes #16856

---
*PR created automatically by Jules for task [15412512283885343825](https://jules.google.com/task/15412512283885343825) started by @dogi*


dogi:
**Review — rating: 88/100**

Both halves are behaviour-preserving, and the second one is worth confirming rather than assuming. `existingConcatenatedLinks` is a `HashSet<String>` (built via `.toHashSet()` or `hashSetOf()`), and Gson serializes any `Collection` as a JSON array by iterating it — so `toJson(set)` and `toJson(set.toList())` emit the same array in the same iteration order. The `.toList()` was allocating a full copy purely to hand Gson a type it was already happy with.

`addAll(linksToProcess)` is likewise identical to the manual loop, and on `HashSet` it can pre-size against the incoming collection rather than growing one element at a time.

The `synchronized(concatenatedLinks) { linksToProcess = concatenatedLinks.toList() }` snapshot above is untouched, so the copy that actually matters for thread safety — taking a stable view of the shared list before working with it — is still there. Only the redundant second copy goes.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16869, opened 2026-09-04
all: smoother service module application scoping (fixes #16913)

dogi:
## Summary

Add a `CoroutineExceptionHandler` to the application scope to log failures instead of allowing them to reach the uncaught exception handler, which persists crash logs and drops the user to the home screen. This prevents fire-and-forget background work (logging, warm-ups, sync scheduling, network state monitoring) from crashing the entire app over transient failures.

## Changes

- **ServiceModule.kt**: Attach a `CoroutineExceptionHandler` to `provideApplicationScope()` that logs exceptions via `Log.e()` instead of propagating them to the global handler
- **ServiceModuleTest.kt**: Add test verifying the exception handler logs failures and prevents escalation to the uncaught handler
- **robolectric.properties** (new): Configure Robolectric to default all tests to a plain `Application` instead of the real `@HiltAndroidApp MainApplication`, preventing fire-and-forget coroutines from outliving test classes and causing flaky `UncaughtExceptionsBeforeTest` failures in unrelated tests
- **TeamTaskTest.kt**: Add `@Before`/`@After` setup to inject the Robolectric test context into `MainApplication.testContext` so `serialize()` reads the correct device ID instead of relying on the real application being booted
- **TESTING.md**: Document the Robolectric application class configuration, the root cause of `UncaughtExceptionsBeforeTest` flakes, and how to diagnose them via the uploaded test report's `Suppressed:` stack trace

## Implementation Details

The application scope is used for fire-and-forget background work that should never crash the app. Without an exception handler, any uncaught exception in these coroutines reaches the process-wide uncaught exception handler, which in this app persists a crash log and navigates the user to the home screen — losing their current screen over a failed log write or network state check.

The new handler logs the failure at ERROR level with the tag `"ApplicationScope"` and the message `"application scope work failed"`, allowing the app to continue running while still recording the issue for debugging.

The Robolectric configuration change fixes a class of flaky test failures where fire-and-forget coroutines launched during `MainApplication.onCreate()` (warm-ups, ANR watchdog, sync scheduling) outlive the test class because Robolectric only pumps the main looper when a test asks. A later test class that mocks `MainApplication.Companion` and unmocks it turns their next companion call into a `MockKException`, which escapes to the global handler and gets charged to whichever test calls `runTest` next, creating a false `UncaughtExceptionsBeforeTest` failure in an unrelated class.

https://claude.ai/code/session_01RjT7vKHgwDnCNuVVxP8nMT


dogi:
**Review — rating: 88/100**

Correct diagnosis of a subtlety that bites a lot of codebases: `SupervisorJob()` alone does **not** stop an uncaught exception from killing the process. It stops a failing child from cancelling its siblings, but an exception escaping a `launch` with no `CoroutineExceptionHandler` in context still reaches the thread's default uncaught handler — so any fire-and-forget work in `@ApplicationScope` that throws takes the app down. Adding the handler to the scope's context is the correct and minimal fix.

The blast radius justifies the `priority` label: `@ApplicationScope` is the scope behind background uploads, broadcast dispatch, retry recovery and download coordination — including `ResourceDownloadCoordinator` in #16804 — all paths that touch the network and filesystem on devices with unreliable connectivity, i.e. exactly where an unexpected throw is most likely and a crash least excusable.

Putting the tag in a file-level `private const` rather than inline is the right call for a `@Module` object, and updating `docs/TESTING.md` alongside means the behaviour is written down rather than only implied by the DI graph.

Two things worth knowing about the boundary of this fix, neither blocking:

- **It covers `launch`, not `async`.** An exception from an `async` in this scope is held until `await()`, so any application-scope code using `async` without awaiting still loses the failure silently. Not a regression — just not covered by this.
- **The failure now exists only in logcat.** The app already has a durable diagnostic path (`ApkLog` / `persistCriticalLog`, used by `handleUncaughtException`), and routing application-scope failures there too would make them visible in a bug report from a field device rather than requiring someone to be attached to logcat when it happens. That's arguably the more valuable half for an offline-first deployment.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16880, opened 2026-09-04
courses: smoother repository steps encoding (fixes #16853)

### LINKED ISSUE #16853: Replace android.util.Base64 in CoursesRepositoryImpl with a byte-identical java.util.Base64
## What

One Android-platform import in a repository, used once, to derive an id that is **persisted and compared**.

**File:** `app/src/main/java/org/ole/planet/myplanet/repository/CoursesRepositoryImpl.kt` (import line 3, use line 684)

## Why

```kotlin
import android.util.Base64                                                              // 3
val stepId = Base64.encodeToString(stepElement.toString().toByteArray(), Base64.NO_WRAP) // 684
```

`java.util.Base64` is available at minSdk 26 and is a platform-free equivalent, so this is the last thing tying this line to Android.

## The fix

⚠️ **Read the flags before substituting.** `Base64.NO_WRAP` suppresses line breaks but **keeps padding**. The parity-preserving replacement is therefore `java.util.Base64.getEncoder()` — *with* padding — not `withoutPadding()`.

This matters more than usual: the derived `stepId` is persisted and compared later, so a changed encoding silently breaks matching against already-stored ids.

## Acceptance

- **A test pinning the derived step id for a fixed `JsonObject` to the known Base64 string** — write it against the *current* implementation first, confirm it passes, then swap the encoder and confirm it still passes. That is the whole safety argument for this change.
- `android.util.Base64` import removed.

## Notes

- This file is contended by other open PRs, but the change is two lines plus a test — rebase rather than restructure if you hit a conflict.

dogi:
Replaces android.util.Base64 with java.util.Base64 in CoursesRepositoryImpl and adds a unit test verifying stepId encoding parity.

Fixes #16853

---
*PR created automatically by Jules for task [15473104383749876046](https://jules.google.com/task/15473104383749876046) started by @dogi*


dogi:
**Review — rating: 86/100**

The encoding is byte-identical, which matters more than usual here because `stepId` is an **identifier** — it's derived from the step JSON and used to key steps, so a change in output would silently orphan existing data rather than fail loudly. `android.util.Base64.NO_WRAP` means the standard RFC 4648 alphabet, no line wrapping, padding retained (padding only goes away with `NO_PADDING`). `java.util.Base64.getEncoder()` is the RFC 4648 *basic* encoder: same alphabet, no line breaks, padding retained. Same bytes in, same string out.

`java.util.Base64` needs API 26 and the project's `minSdk` is 26, so there's no desugaring or availability question.

The real payoff is testability, and it's the same argument as #16819's `TextUtils` removal: `android.util.Base64` is a framework stub that returns nothing under a plain JVM, so any test touching this code path had to run under Robolectric. `java.util.Base64` is ordinary JDK, so `CoursesRepositoryImpl`'s step extraction becomes testable without an Android runtime.

**Nit:** the import lands out of order — `import java.util.Base64` sits between `android.util.Log` and `com.google.gson.JsonArray`, where the convention sorts `android` < `com` < `java`. It belongs below the `com.google.gson` imports.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16881, opened 2026-09-04
login: smoother shared preferences users caching (fixes #16855)

### LINKED ISSUE #16855: Hoist the TypeToken out of SharedPrefManager.getSavedUsers
## What

A constant `TypeToken` is constructed on every call to read the saved-users list.

**File:** `app/src/main/java/org/ole/planet/myplanet/services/SharedPrefManager.kt` (`getSavedUsers`, lines 72-75)

## Why

```kotlin
fun getSavedUsers(): List<User> {                        // 72
    ...
    val type = object : TypeToken<List<User>>() {}.type  // 75
```

The anonymous `TypeToken` and its reflective type are constant, but a new one is allocated per call — and the login screen calls this on every refresh of the user list. `gson` is already injected, so the type allocation is the only waste here.

`data/room/Converters.kt` already shows the cached-companion pattern for its list types; follow that.

## The fix

Hold the type in a private `val` (companion or field) and use it in the `fromJson` call.

No change to the `User` model.

## Acceptance

- Saved-user list parses identically, including the empty and malformed-JSON paths.

## Notes

- Other open PRs touch this file (around lines 166-193 and adding cache helpers). Your change is at 72-75, so it shouldn't collide.

dogi:
Caches the List<User> TypeToken in SharedPrefManager companion object to eliminate repeated anonymous TypeToken allocations on saved user reads.

Fixes #16855

---
*PR created automatically by Jules for task [6403594108583331646](https://jules.google.com/task/6403594108583331646) started by @dogi*


dogi:
**Review — rating: 88/100**

`object : TypeToken<List<User>>() {}` allocated a fresh anonymous-subclass instance on every call and then walked its generic superclass via reflection to recover the `Type`. Hoisting it to a `companion object` `private val` does that once for the process. `TypeToken` is immutable and its resolved `Type` is safe to share across threads, so there's no reason it was ever per-call.

Worth having because `getSavedUsers()` isn't a one-shot: it backs the saved-user list on the login screen, so it's read on startup and whenever that list is rendered or refreshed.

The `usersJson != null` guard and the `emptyList()` branch are untouched, so behaviour on a missing preference is unchanged.

**Nit:** the constant sits above the `SAVED_USERS`/`REPLIED_NEWS_ID` string constants rather than with them — grouping the type token below the keys, or at the end of the companion, would keep the string constants contiguous.

⚠️ **No build or test verdict on this head** — only `myPlanet labels` ran; neither `build` nor `test` fired. The above is from reading the diff.

---

Okuro3499 (approved):

PR #16883, opened 2026-09-04
login: smoother storage breakdown extension mapping (fixes #16859)

### LINKED ISSUE #16859: Cheapen the StorageBreakdown full-tree scan
## What

Opening the storage breakdown walks every file under the ole directory and does a linear category lookup per file.

**File:** `app/src/main/java/org/ole/planet/myplanet/ui/settings/StorageBreakdownFragment.kt` (`scanStorage`, line 205)

## Why

```kotlin
oleDir.walkTopDown().filter { it.isFile }.forEach { file ->   // 214
    val ext = file.extension.lowercase()                      // 215 — allocation per file
    ...
    val index = StorageCategories.indexOf(ext)                // 218 — linear scan per file
```

On a device with a large offline library that's a linear category search plus a fresh lowercase string for every single file.

## The fix

- Build an **extension → category index map once** from `StorageCategories` before the walk, and look up against it instead of calling `indexOf` per file.
- Avoid the repeated lowercase allocation where possible.
- Keep the scan off the main thread — the existing `launch` already handles that; don't change the threading.

## Acceptance

- Category totals and file counts identical for a fixture tree, including files with uppercase extensions and files with no extension.

## Notes

⚠️ **Do not implement the caching idea from the original proposal.** It suggested caching the result keyed on the ole directory's `lastModified` — that is unsound, because a directory's mtime does **not** change when a nested file is written, so the cache would serve stale totals after a download.

Take the per-file work reduction only. If a cache is wanted later, it must be keyed on explicit invalidation from the delete/download flows, not on mtime.

dogi:
Optimized StorageBreakdownFragment's scanStorage method by constructing an extension -> category index lookup map prior to traversing the ole directory. Reduced per-file memory allocations and category search complexity, and added unit tests covering the storage scan behavior.

Fixes #16859

---
*PR created automatically by Jules for task [8192640200605400563](https://jules.google.com/task/8192640200605400563) started by @dogi*

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved storage breakdown accuracy for files with uppercase extensions such as `.MKV`, `.MP3`, and `.PNG`.
  - Files without an extension are now included in the “Other” category instead of being omitted.
  - Storage scans for empty or unavailable directories now correctly report zero usage and zero file counts.


dogi:
**Review — rating: 74/100**

The lookup this replaces is already O(1), and the replacement doesn't tie-break the same way.

**`StorageCategories.indexOf` is not a scan.** It's backed by a precomputed map built once at object initialisation:

```kotlin
private val extensionToIndex: Map<String, Int> = buildMap {
    all.forEachIndexed { index, category ->
        category.extensions.forEach { ext ->
            if (!containsKey(ext)) put(ext, index)
        }
    }
}
fun indexOf(extension: String): Int = extensionToIndex[extension] ?: OTHER_INDEX
```

So the per-file cost was already a single hash lookup. `extMap` rebuilds an equivalent map on **every scan**, and the walk then does up to two lookups per file instead of one. The genuine saving that remains is avoiding the `ext.lowercase()` allocation for files whose extension is already lowercase — real, but much smaller than "optimize full-tree scan" implies.

**More importantly, the new map has different semantics.** `StorageCategories` guards with `if (!containsKey(ext))`, so an extension listed under two categories resolves to the **first**. The new `buildMap` does an unguarded `put(ext, index)`, so it resolves to the **last**. That guard exists because whoever wrote it expected overlap to be possible; if any extension is ever shared between categories, files silently move between buckets and the breakdown totals change with no test catching it.

The narrower change gets the allocation win without the divergence — keep using `StorageCategories.indexOf` and try the raw extension first:

```kotlin
val ext = file.extension
val index = if (ext.isEmpty()) StorageCategories.OTHER_INDEX
            else StorageCategories.indexOf(ext).takeIf { it != StorageCategories.OTHER_INDEX }
                 ?: StorageCategories.indexOf(ext.lowercase())
```

or, better, add a case-insensitive lookup to `StorageCategories` so the mapping stays in one place.

**The testability half is good and worth keeping.** Splitting `scanStorage()` into a `scanStorage(oleDir: File)` that takes its directory means the categorisation logic can be exercised against a temp tree instead of the device's real `ole` directory, and that's the part that had no coverage.

**Verdict:** needs another round — reuse `StorageCategories.indexOf` rather than rebuilding a map with different precedence; keep the `scanStorage(File)` split.

---

dogi:
**Re-review — rating: 85/100** (was 74)

Both blocking points are resolved. The divergent `buildMap` is gone; the walk now calls `StorageCategories.indexOf` and falls back to the lowercased extension only on a miss, which is the shape I proposed — so the first-wins precedence guard in `StorageCategories` stays authoritative and there's one mapping, not two. The `scanStorage(File)` split is kept, and it now carries tests.

**One correction to my own earlier note.** I wrote that the remaining win was "avoiding the `ext.lowercase()` allocation for files whose extension is already lowercase." That was wrong: `String.lowercase()` goes to `toLowerCase(Locale.ROOT)`, which scans for the first character that would change and returns **the receiver itself** when none does. So the original code allocated nothing on the common path either. What this change actually trades is a per-character scan for a hash lookup on already-lowercase known extensions, at the cost of one extra failed lookup for every unknown extension. Both sides are noise next to the `walkTopDown` + `File.length()` syscalls that dominate this scan.

So the production half is a wash, and it's worth being clear that's what it is. **The tests are the payload**, and they're good ones: uppercase `MKV`/`MP3`/`PNG` exercise the fallback branch that would otherwise be untested, `README` pins extensionless → `OTHER`, `notes.txt` pins unknown → `OTHER`, and the empty/nonexistent-directory case pins the early return. I walked the fixture arithmetic — 790 bytes total, `counts[0..4]` = 2/1/1/1/2 — and it matches the code. This is the first coverage `scanStorage` has had at all.

Two notes, neither blocking:

- Constructing `StorageBreakdownFragment()` under plain JUnit with no Robolectric runner is fine here — `CollectionsFragmentTest` already does exactly this with a `DialogFragment` and passes, and `ActivitiesFragmentTest` with a plain `Fragment`. `BottomSheetDialogFragment` adds only trivial constructors on top.
- `Log.d`-free, `Locale`-free: `lowercase()` with no argument is `Locale.ROOT`, so the Turkish-I hazard doesn't apply. Worth knowing, since `lowercase(Locale.getDefault())` elsewhere in this codebase does have it.

⚠️ **This push produced only the `myPlanet labels` check — `build` and `test` never ran, so there is no CI verdict on the new test file.** Worth a re-run before merging, since an unverified new test is exactly the thing that turns master red. This keeps recurring on a subset of pushes (#16752, #16793, #16829, #16851, #16885, #16911, and others).

**Verdict:** mergeable once CI actually runs on it.

---

Okuro3499 (approved):

PR #16885, opened 2026-09-04
all: smoother image utils disk caching (fixes #16868)

### LINKED ISSUE #16868: Bound the Glide decode size and align the disk-cache strategy in ImageUtils
## What

Three sibling image loaders in one file apply different Glide options.

**File:** `app/src/main/java/org/ole/planet/myplanet/utils/ImageUtils.kt` (`loadImage`, `loadPlaceholderImage`)

## Why

- `loadProfileImage` applies `.override(sizePx, sizePx)` **and** `diskCacheStrategy(ALL)`
- `loadImage` has the disk-cache strategy but no override
- `loadPlaceholderImage` has **neither**

## The fix

- Add an optional `sizePx` parameter (defaulted, so no call site outside this file needs editing) applying `.override(size, size)`.
- Give `loadPlaceholderImage` the same `diskCacheStrategy` as its siblings.

Preserve the placeholder / error / `circleCrop` behavior.

## Acceptance

- Images render as today; placeholder loads now hit the disk cache.

## Notes

⚠️ **Premise correction.** The original finding claimed these "decode full-resolution bitmaps". That's overstated — Glide already sizes decodes to the measured target view, so `.override()` only matters for unmeasured / `wrap_content` targets. **The missing `diskCacheStrategy` on `loadPlaceholderImage` is the concrete win**; the override is a defensive addition. Please describe the PR accordingly rather than claiming a decode-size fix.

dogi:
This change aligns the disk cache strategy across image loading functions in ImageUtils by adding DiskCacheStrategy.ALL to loadPlaceholderImage, and adds an optional sizePx parameter to loadImage and loadPlaceholderImage for bounded image decoding.

Fixes #16868

---
*PR created automatically by Jules for task [6731807088673252930](https://jules.google.com/task/6731807088673252930) started by @dogi*


dogi:
**Review — rating: 80/100**

The alignment half is the more consequential change and deserves naming: `loadPlaceholderImage` had **no** `diskCacheStrategy` at all, so it ran on Glide's `AUTOMATIC` default while `loadImage` used `ALL`. Two profile-image loaders caching differently is the kind of inconsistency that produces "why is this one slow" reports, so pinning both to `ALL` is reasonable. Worth being deliberate about it though: `ALL` caches the decoded/transformed result *in addition to* the original bytes, which costs more disk — and this app ships a whole storage-breakdown feature because disk pressure on target devices is a real concern.

The `sizePx` override is a genuine win where callers use it: without `override`, Glide decodes at the target `ImageView`'s measured size, and for a list of small avatars backed by full-resolution uploads that's a lot of wasted decode. Making it optional keeps every existing call site behaving exactly as before.

**The `apply { }` is fragile, and I'd change it.** `override()` comes from `BaseRequestOptions`, and its return value is discarded here:

```kotlin
.apply {
    if (sizePx != null) {
        override(sizePx, sizePx)
    }
}
```

This works only because `RequestBuilder` mutates in place and returns `this` when auto-clone is off — which is the default. But `BaseRequestOptions` is explicitly written to return a *clone* when `isAutoCloneEnabled`, and in that case the override is silently dropped and images quietly load at full size with nothing failing. Kotlin's `apply` also shadows Glide's own `RequestBuilder.apply(RequestOptions)`, which reads confusingly at a glance. Chaining through `let` keeps the returned builder:

```kotlin
.let { if (sizePx != null) it.override(sizePx, sizePx) else it }
```

⚠️ **No build or test verdict on this head** — only `myPlanet labels` ran; neither `build` nor `test` fired.

**Verdict:** mergeable — worth switching `apply` to `let` so the builder's return value isn't discarded.

---

Okuro3499 (dismissed):
```
ImageUtils.kt:24,43 — sizePx parameter is unused by every current caller, making the "bound decode size" fix inert. Either wire it into the real call sites that load into small 
    views (e.g. VoicesAdapter.kt:421, LoginActivity.kt:582, MyHealthFragment.kt:168) in this same PR, or drop the parameter until a follow-up does so.
```

dogi:
**Re-review — rating: 84/100** (was 80)

The follow-up commit is the half that was missing: `ImageUtils` had grown an optional `sizePx`, but nothing passed one. Now `MyHealthFragment`, `LoginActivity` and `VoicesAdapter` each supply the real dimension the view is laid out at — `R.dimen.user_image_size`, `R.dimen._80dp`, `R.dimen._40dp` — so Glide decodes avatars at display size instead of full resolution. On a Voices list that's the difference between a handful of 40dp bitmaps and a screenful of full-size decodes, which is the actual memory win the title promises. Adding `diskCacheStrategy(DiskCacheStrategy.ALL)` to `loadPlaceholderImage` brings it in line with `loadImage`, which was a real inconsistency.

**The `apply { override(...) }` shape I flagged last round is still there.** To restate why it's worth changing, and to be precise about how much it matters: `BaseRequestOptions.override()` routes through `selfOrThrowIfLocked()`, which under `isAutoCloneEnabled` returns a *clone* and applies the mutation to that, leaving the receiver untouched. Inside `apply { ... }` the return value is discarded, so the override would silently vanish. A `RequestBuilder` straight out of `Glide.with().load()` does **not** have auto-clone enabled, so this works correctly today — it isn't a live bug, and I don't want to overstate it. It's a trap laid for whoever later refactors these into shared `RequestOptions` or applies `autoLock()`, at which point the sizing quietly stops happening with nothing failing. Chaining the conditional instead of mutating in place removes the trap for the same number of lines.

Smaller: `VoicesAdapter` resolves `avatarSize` via `getDimensionPixelSize` on every bind. `Resources` caches dimension lookups so it's cheap, but it's per-row work that could be a `by lazy` field on the adapter.

CI green 5/5.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16887, opened 2026-09-04
teams: smoother selection label caching (fixes #16857)

### LINKED ISSUE #16857: Cache the resolved Teams label in TeamsSelectionAdapter.bind
## What

A localized string is re-resolved on every row bind to compare against a constructor value that never changes.

**File:** `app/src/main/java/org/ole/planet/myplanet/ui/teams/TeamsSelectionAdapter.kt` (`TeamSelectionViewHolder.bind`)

## Why

```kotlin
if (section == itemView.context.getString(R.string.teams))
```

`getString` runs per visible row, per bind, to pick between the team and business icon — for a `section` value fixed at construction.

## The fix

Resolve the comparison value once in an adapter-level `lazy` field and compare against it in `bind`. Keep the icon selection identical.

## Acceptance

- Correct icon for both the teams and the non-teams section, in a non-English locale as well.

## Notes

- The original proposal wavered between three approaches. Take the lazy field — it's the smallest and matches how the rest of the codebase caches resolved resources.

dogi:
Caches the resolved R.string.teams label at the adapter level in TeamsSelectionAdapter to prevent re-resolving localized string resources on every row bind.

Fixes #16857

---
*PR created automatically by Jules for task [13637172865478755478](https://jules.google.com/task/13637172865478755478) started by @dogi*


dogi:
**Review — rating: 80/100**

Correct and safe. `bind` runs on the main thread, so the unsynchronised lazy write to `cachedTeamsLabel` has no race, and the adapter is recreated along with its fragment on a configuration change, so the cached string can't outlive a locale switch.

I checked whether this was the same trap as #16733 — comparing against a localized string — and it isn't: `section` is itself a localized value. `ChatHistoryAdapter` compares it against `context.getString(R.string.teams)` at lines 200 and 219 before passing it in, so both sides of `section == teamsLabel` move together when the locale changes. The comparison is consistent, not broken.

That said, the honest assessment is that the win is very small. `getString` is a resource-table lookup Android already caches; it isn't a parse or an I/O. Against that, the adapter now carries a piece of mutable state whose validity depends on an assumption about its lifetime.

**The thing actually worth fixing is underneath this.** The same "is this the teams section" question is answered by string-comparing a localized label in three places now (`ChatHistoryAdapter:200`, `:219`, and here), and `section` is then written verbatim into the upload payload at `ChatHistoryAdapter:262` (`map["viewInSection"] = section`) — so the server receives "Teams" or "Equipos" depending on the device's language. An enum or a stable key passed alongside the display label would collapse all three comparisons and stop locale leaking into uploaded documents. Out of scope here, but that's where the real defect is.

**Verdict:** mergeable.

---

Okuro3499 (approved):

Okuro3499 (approved):

PR #16888, opened 2026-09-04
sync: smoother force policy overdue checking (fixes #16845)

### LINKED ISSUE #16845: Swap java.util.concurrent.TimeUnit for kotlin.time in ForceSyncPolicy
## What

One JVM-only import is the sole thing tying an otherwise pure, well-tested policy object to the JVM.

**Files:**
- `app/src/main/java/org/ole/planet/myplanet/ui/sync/ForceSyncPolicy.kt` (import line 3, use at line 27)
- its test

## Why

```kotlin
import java.util.concurrent.TimeUnit                                   // 3
val daysDiff = TimeUnit.MILLISECONDS.toDays(nowMillis - lastSyncMillis) // 27
```

The object has no Android imports and is well covered by tests. `TimeUnit` is JVM-only, so it's the one thing blocking this sync-cadence policy from moving into a common module later.

## The fix

Use the `kotlin.time` equivalent — `(nowMillis - lastSyncMillis).milliseconds.inWholeDays` — and drop the import. Swap the test fixtures (`TimeUnit.DAYS.toMillis`) likewise.

## Acceptance

⚠️ Keep the signatures, the constants and the **inclusive-threshold semantics** identical. `inWholeDays` truncates toward zero just as `toDays` does, but the boundary cases are exactly what the existing tests pin — they should pass unchanged. If any test needs editing beyond the fixture helper, stop and say so, because that means the semantics moved.

## Notes

dogi:
Swapped `java.util.concurrent.TimeUnit` for `kotlin.time` in `ForceSyncPolicy` and `ForceSyncPolicyTest`.
- In `ForceSyncPolicy.kt`, replaced `TimeUnit.MILLISECONDS.toDays(nowMillis - lastSyncMillis)` with `(nowMillis - lastSyncMillis).milliseconds.inWholeDays`.
- In `ForceSyncPolicyTest.kt`, replaced `TimeUnit.DAYS.toMillis` and `TimeUnit.HOURS.toMillis` with `days.days.inWholeMilliseconds` and `1.hours.inWholeMilliseconds`.

Fixes #16845

---
*PR created automatically by Jules for task [714672155946531359](https://jules.google.com/task/714672155946531359) started by @dogi*


dogi:
**Review — rating: 85/100**

Behaviour-identical. `TimeUnit.MILLISECONDS.toDays(x)` is integer division by 86,400,000, truncating toward zero, and `Duration.inWholeDays` truncates toward zero as well — so they agree on the ordinary case and also on the one that could plausibly occur here: a negative difference from clock skew or a `lastSyncMillis` in the future both yield `0` under either implementation, so the `daysDiff >= maxDays` guard behaves the same. Both saturate rather than wrap at the extremes.

The value is idiom rather than performance — `kotlin.time` reads better than a `TimeUnit` enum lookup and drops a `java.util.concurrent` import from a file that has no concurrency in it. Small, but this is a pure policy object, so keeping it free of unrelated imports is worth something.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16890, opened 2026-09-04
all: smoother notifications time providing (fixes #16860)

### LINKED ISSUE #16860: Inject the clock into NotificationsAdapter's relative-time formatting
## What

`formatRelativeTime` reads the wall clock directly, so its bucket ladder can't be tested.

**File:** `app/src/main/java/org/ole/planet/myplanet/ui/notifications/NotificationsAdapter.kt` (`formatRelativeTime`, line 152)

## Why

```kotlin
val diff = System.currentTimeMillis() - createdAt
```

called per bind, feeding a hand-rolled minute/hour/day bucket ladder with its own `DateTimeFormatter` cache. Because the clock is read inside the function, none of those boundaries can be asserted.

## The fix

Pass an injectable clock — a `now: () -> Long` from the hosting fragment, backed by `TimeProvider` — into the adapter and use it.

Keep the existing thresholds and **rendered strings identical**. Add a Robolectric test for the minute / hour / day / fallback boundaries.

## Acceptance

- Rendered strings unchanged at every bucket.
- Boundary tests pass with a pinned clock.

## Notes

⚠️ Two constraints:
- `TimeUtils.getRelativeTime(timestamp, timeProvider)` already exists, but its `DateUtils`-based output **differs** from this ladder. Reuse it only if you verify the rendered strings match exactly; otherwise keep the local ladder and just inject the clock.
- **Do not edit `TimeUtils`** — it has an open PR against it.

This file also has an open PR touching lines 94-95 and ~167. Your change is at 152, so it shouldn't collide.

dogi:
Injects a clock provider into `NotificationsAdapter`'s `formatRelativeTime` method to allow relative-time bucket ladder testing. `NotificationsFragment` passes a lambda backed by `TimeProvider`, and Robolectric boundary unit tests are updated with a pinned clock.

Fixes #16860

---
*PR created automatically by Jules for task [13308475253783598549](https://jules.google.com/task/13308475253783598549) started by @dogi*


dogi:
**Review — rating: 88/100**

`formatRelativeTime` is exactly the kind of function that's trivial to get wrong at the boundaries and impossible to test while it reads the wall clock — the `60_000L` / hour / day thresholds have off-by-one behaviour that only shows up at the edges. Taking `now: () -> Long` makes those assertable.

Routing the fragment's value through the injected `TimeProvider` rather than a bare lambda is the detail that makes this fit the codebase: the project already has `TestTimeProvider` and uses `timeProvider` throughout the repositories, so notification timestamps now come from the same substitutable clock as everything else instead of a second, independent source of "now".

Defaulting the parameter to `{ System.currentTimeMillis() }` keeps any other construction site behaving exactly as before, so this is additive.

**Nit:** `import javax.inject.Inject` landed in the middle of the `org.ole.planet.myplanet.ui.teams.*` block, between `TeamPageConfig.TasksPage` and `ReplyActivity` — `javax` sorts above `org`, so it belongs up with the other non-project imports.

**Verdict:** mergeable.

---

PR #16693, opened 2026-09-03
dashboard: smoother user profile file utils url resolving (fixes #16691)

Okuro3499:
fixes #16691
[Screen_recording_20260903_160243.webm](https://github.com/user-attachments/assets/2229a39b-cb77-4f93-bfbf-7fce4fc29992)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved profile image selection and camera capture handling, including more reliable image uploads.
  - Prevented image filenames containing path traversal characters from saving files outside the intended storage location.
  - Added safer fallback naming when selected image filenames are missing or invalid.

- **Improvements**
  - Gallery selection now filters for image files, making it easier to choose valid profile images.

PR #16891, opened 2026-09-04
login: smoother user profile dialogs landscaping (fixes #16882)

### LINKED ISSUE #16882: userprofile redesign
userprofile redesign
https://github.com/user-attachments/assets/228483a8-0726-4ff2-92ad-cd5c6f7c546c

Okuro3499:
fixes #16882
Screenshot_20260904_190729
Screenshot_20260904_190718
[prof.webm](https://github.com/user-attachments/assets/852a70fd-096f-4f5c-9d52-4e556fa2ea9b)

## Summary by CodeRabbit

* **New Features**
  * Redesigned the user profile screen with a banner, detail card, activity section, guest indicator, and improved profile editing experience.
  * Added responsive layouts for landscape, tablet, and foldable devices.
  * Added a bottom-sheet photo picker with camera, gallery, and cancel options.
  * Added customizable content width support for profile layouts.
* **Style**
  * Improved profile fields, buttons, dialogs, statistics rows, dividers, spacing, and visual styling.
* **Localization**
  * Added profile labels and guest messaging in Arabic, Spanish, French, Nepali, and Somali.


dogi:
**Review — rating: 84/100**

At 31 files and ~1850 insertions this is mostly layout and resource work, so I sampled rather than read every line: I checked the Kotlin changes in full, the localization coverage across all six string files, and the resource-variant structure, then spot-checked the risky hunks.

**Localization is complete, and the one apparent gap is correct.** `values/strings.xml` gains four strings while each of `values-ar|es|fr|ne|so` gains three — which looks like a missing translation until you see which one is absent. It's `profile_secondary_line` = `%1$s · %2$s`, pure positional placeholders with a separator and no translatable word in it. Leaving that English-only is right. Worth marking it `translatable="false"` so lint stops reporting it as missing in five locales and the next person doesn't "fix" it by adding five identical copies.

**`MaxWidthFrameLayout` is done properly.** The new `maxContentWidth` attribute is read through `obtainStyledAttributes`, the `TypedArray` is recycled, and the `-1` sentinel falls back to the previous `MAX_WIDTH_DP * density` behaviour — so every existing usage without the attribute measures exactly as before.

**`StatsAdapter` gets slightly more correct as a side effect.** Swapping alternating background stripes for a trailing divider drops two cached colors, and the new code reads the `position` parameter where the old one read `holder.bindingAdapterPosition` inside `onBindViewHolder` — `position` is the right one there; `bindingAdapterPosition` can return `NO_POSITION` mid-update and is meant for later callbacks.

Resource variants look deliberate: `fragment_user_profile` ships default + `layout-land` + `layout-sw600dp`, with matching `values-land` / `values-sw600dp` dimens and integers.

**Two things worth acting on:**

- **The photo picker gets new chrome over the old bug.** This replaces the `AlertDialog` list with a `BottomSheetDialog` (`dialog_photo_picker.xml`) but still calls `pickFromGallery()`, which is unchanged — so it keeps launching `ACTION_PICK` against `MediaStore.Images.Media.INTERNAL_CONTENT_URI`, the *internal* volume rather than the user's photos. #16693 fixes exactly that (and the transient `content://` URI being persisted as the profile image). These two touch `UserProfileFragment` and currently merge cleanly, but they're solving adjacent halves of the same feature — worth coordinating so the redesign doesn't ship a prettier version of a picker that opens the wrong collection.
- **`edit_profile_dialog` has a `layout-sw600dp` variant but no `layout-land`**, while `fragment_user_profile` has both. Given this repo's run of landscape-cropping issues (#14893, #14960, #15266, #15412), an edit dialog with no landscape treatment is the likely next one.

**Verdict:** mergeable — I sampled rather than read the full diff, so this leans on the green build and on the structure being consistent; the picker overlap with #16693 is the thing to settle before merge.

---

dogi (approved):

PR #16892, opened 2026-09-04
resources: smoother collections button opening (fixes #16090)

### LINKED ISSUE #16090: collections buttons in library malfunction
collections buttons in library doesn't do anything. It should show a collection dialog if there is no collection an information message should be shown to user 
https://github.com/user-attachments/assets/3133c324-48c8-4e9f-a551-f5aa87930701

ragilzakaria:
https://github.com/user-attachments/assets/864f300e-719a-4731-a4a7-9207c78d5139

fixes #16090 

## Summary by CodeRabbit

* **Bug Fixes**
  * The resource filter card now reliably closes when opening Collections or clearing selected tags.
  * Clearing tags now consistently removes all active search tags.
  * Opening Collections behaves correctly when saved collections are available or when none exist.
  * Collection navigation now provides more consistent filter and dialog behavior.

* **Tests**
  * Added coverage for collection navigation, filter dismissal, tag clearing, and empty collection states.


J-S-webskas (approved):

dogi:
**Review — rating: 86/100**

Small production diff, and it fixes a genuine "the button does nothing" bug.

`ResourcesFragment` was registering **two** click listeners on `binding.btnCollections`: the real one in `setupCollectionsButton` that opens `CollectionsFragment`, and a second one inside `additionalSetup` that only did `bottomSheet.visibility = View.GONE`. Whichever ran last won, so tapping Collections closed the filter sheet and never opened the dialog. Removing the duplicate and folding the visibility change into the real handler is exactly right — and I checked the two are equivalent: `additionalSetup` opens with `val bottomSheet = binding.cardFilter`, so `binding.cardFilter.visibility = View.GONE` in the new handler is the same view, not a near-miss.

The `CollectionsFragment` change is the other half and worth as much:

```kotlin
- private lateinit var recentList: MutableList<TagEntity>
+ private var recentList: MutableList<TagEntity> = ArrayList()
```

A `lateinit` in a `companion object` assigned only from `getInstance` means any path that reaches the fragment without going through `getInstance` — system-initiated recreation after process death being the obvious one — throws `UninitializedPropertyAccessException`. Defaulting it removes a crash rather than papering over one.

One convention note on `ResourcesFragmentActionTest`, not blocking: it's a Robolectric + Hilt + real-Room integration test driving the fragment through `ShadowLooper`, and it uses `runBlocking` in the test bodies rather than the `runTest` + `MainDispatcherRule` + `TestDispatcherProvider` combination `docs/TESTING.md` prescribes. For the `@Before`/`@After` `clearAllTables` calls that's fine and idiomatic. Inside the test body it means the assertions depend on `executePendingTransactions` and looper pumping lining up, which is the kind of thing that goes flaky under load rather than failing honestly. It passes today and it's testing something worth testing — that the button actually opens the dialog — so I'd take it and tighten it later.

Star tag present, but the approval on this PR is from J-S-webskas, not Okuro3499, so it doesn't meet the merge gate yet.

CI green 5/5.

**Verdict:** mergeable.

---

dogi (approved):

PR #16893, opened 2026-09-04
teams: smoother voices mime type uploading (fixes #16854)

### LINKED ISSUE #16854: Replace the openConnection MIME probe in UploadManager.uploadNews
## What

The news-image upload loop opens a `URLConnection` on a local file just to read its content type.

**File:** `app/src/main/java/org/ole/planet/myplanet/services/UploadManager.kt` (`uploadNews`, line 444)

## Why

```kotlin
val mimeType = imageFile.toURI().toURL().openConnection().contentType   // 444
```

Once per news image, for a local file. `FileUtils.getMimeType` / `Utilities.getMimeType` already exist for extension-based lookup — this is the same anti-pattern as the `UploadRepositoryImpl.uploadAttachment` probe.

## The fix

Drop the probe and derive the MIME from the filename, falling back to `application/octet-stream`. Pass it into `getHeaderMap` exactly as today.

Do not change the `createImage` contract or the news JSON shape.

## Acceptance

- The same `Content-Type` reaches `getHeaderMap(mimeType, resourceRev)` for jpg/png/gif images as today.
- News upload payload unchanged.

## Notes

⚠️ **Correction to the original framing.** This was submitted as a correctness fix, claiming the resolved MIME is ignored and uploads always use `application/octet-stream`. That is wrong: only the request *body* media type is octet-stream — `mimeType` **is** passed to `getHeaderMap(mimeType, resourceRev)` and does reach the request.

So the change still stands (the probe is pointless I/O), but it is a **performance cleanup, not a bug fix** — please don't write the PR as though it repairs broken content types, and don't "fix" the body media type on the strength of the original claim.

dogi:
Replaces `imageFile.toURI().toURL().openConnection().contentType` in `UploadManager.uploadNews` with `FileUtils.getMimeType(fileName) ?: "application/octet-stream"`. Passes the derived MIME type directly into `getHeaderMap` while preserving request body media type and JSON structure. Adds corresponding unit test coverage in `UploadManagerTest`.

Fixes #16854

---
*PR created automatically by Jules for task [14093520513832132490](https://jules.google.com/task/14093520513832132490) started by @dogi*


dogi:
**Review — rating: 88/100**

Same defect as #16824, in the other place it survived. `imageFile.toURI().toURL().openConnection()` builds a `FileURLConnection`, and reading `.contentType` forces `connect()` — which opens the file and may read bytes to sniff the type. Nothing closed it, so every news-image upload opened a stream and abandoned it. `FileUtils.getMimeType(fileName)` is a name-based lookup with no I/O and nothing to leak.

Using the project's own `FileUtils.getMimeType` rather than `URLConnection.guessContentTypeFromName` is the better choice of the two — it's what `TeamsUploadRunner.uploadTeamImageAttachment` already calls for the equivalent attachment upload, so the two paths now resolve MIME types the same way.

**Two things worth a look:**

- **The fallback disagrees with its sibling.** This uses `?: "application/octet-stream"` while `TeamsUploadRunner` uses `?: "image/*"` for what is also an image attachment. Both are safe, but a news image with an extension-less URL will now be announced as a generic binary where a team image would be announced as an image.
- **The name comes from the URL, not the file.** `fileName` is `FileUtils.getFileNameFromUrl(getString("imageUrl", imgObject))`, so if that URL carries no extension the lookup fails and you get the fallback, where the old sniffing path might have identified it. That's the same trade #16824 made and is fine — worth knowing it's a trade rather than a pure win.

The request body is still hardcoded to `"application/octet-stream"`, so as in #16824 the resolved `mimeType` only affects the headers.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16896, opened 2026-09-04
all: smoother notification dao batch updating (fixes #16894)

### LINKED ISSUE #16894: Batch NotificationDao.markSynced on the non-null-rev path with a CASE expression
## What

`NotificationDao.markSynced(List)` issues one `UPDATE` per notification. Batch the path that actually runs.

**File:** `app/src/main/java/org/ole/planet/myplanet/data/room/dao/NotificationDao.kt`

## Current state on master

```kotlin
@Query("SELECT * FROM notifications WHERE needsSync = 1 AND rev IS NOT NULL")   // 51
suspend fun getPendingSyncNotifications(): List<AppNotification>

@Query("UPDATE notifications SET needsSync = 0, rev = COALESCE(:rev, rev) WHERE id = :id")  // 54
suspend fun markSynced(id: String, rev: String?)

@Transaction
suspend fun markSynced(syncResults: List<Pair<String, String?>>) {   // 57
    syncResults.forEach { (id, rev) -> markSynced(id, rev) }          // 60 — N statements
}
```

## Why the obvious split doesn't help

A previous attempt split this into a partial-entity `@Update` for non-null revs plus an `IN (:ids)` query for null revs. That does **not** batch the real path, for two reasons established in review:

1. **`@Update` with a `List` is not one statement.** Room generates an `EntityDeletionOrUpdateAdapter` and calls `handleMultiple`, which loops `bind()` + `executeUpdateDelete()` per element — exactly the N statements the loop already did.
2. **The null-rev branch is effectively dead.** `getPendingSyncNotifications()` selects `rev IS NOT NULL`, and the only caller — `TransactionSyncManager.syncNotificationReads` (~line 486) — builds pairs as `Pair(notification.id, response.body()?.get("rev")?.asString)` only inside `if (response.isSuccessful)`. A successful CouchDB `PUT /db/{id}` always returns a `rev`. So `nullRevs` is empty in every realistic run.

Net: the batched branch never executes and the executed branch isn't batched.

## The fix

Batch the **non-null-rev** case in a single statement:

```sql
UPDATE notifications SET needsSync = 0, rev = CASE id WHEN ? THEN ? ... END WHERE id IN (:ids)
```

Keep a correct null-rev arm — `UPDATE notifications SET needsSync = 0 WHERE id IN (:ids)`, leaving `rev` untouched. It is rarely reached, but it must still be right, because it is the arm that reproduces `COALESCE(:rev, rev)`'s "null leaves rev alone" semantics.

Chunk both to stay under SQLite's 999-variable limit, following `RemovedLogDao.deleteByTypeUserAndDocsChunked`. Note the CASE form binds ~2 variables per row plus the `IN` list, so size the chunk accordingly rather than copying 900 blindly.

If the CASE expression turns out unwieldy in Room, say so in the PR and propose the alternative rather than falling back to a per-row loop.

## Acceptance

⚠️ **The truth table must be preserved exactly.** Old behavior: non-null rev overwrites `rev`; null rev leaves `rev` alone; both clear `needsSync`.

Room in-memory DAO test covering all three cases:
- new rev → `rev` updated, `needsSync = 0`
- null rev with an existing `rev` → `rev` unchanged, `needsSync = 0`
- null rev with a null `rev` → still null, `needsSync = 0`

Use `@Config(application = android.app.Application::class)` to keep Hilt out of the DAO test.

Also: once `markSynced(List)` no longer calls it, the single-row `markSynced(id, rev)` query is dead — delete it, and delete the `coVerify(exactly = 0)` line in `NotificationsRepositoryImplTest` (~line 366) that is the only remaining reference keeping it compiling.

## Notes

This supersedes PR #16677, which took a different approach that review rejected. **Start from master — do not reuse that branch.** #16677 will be closed once this lands.

dogi:
Batched NotificationDao.markSynced on non-null rev path using a dynamic CASE SQL expression and chunking, updated null rev handling, deleted dead single-row markSynced, and added Room DAO tests covering all cases.

Fixes #16894

---
*PR created automatically by Jules for task [7200145522140027817](https://jules.google.com/task/7200145522140027817) started by @dogi*


dogi:
**Review — rating: 90/100**

This does what #16677 was sent back for: it batches the branch that actually executes. The null-rev path was near-unreachable (CouchDB returns `rev` on every successful `PUT`, and the pair is only built inside `response.isSuccessful`), so batching only that one left the real path issuing one `UPDATE` per notification. A `CASE id WHEN ? THEN ?` over the non-null revs collapses it into one statement per chunk.

**The chunk size is the thing that makes this correct, and it's worth being explicit about.** Each row contributes three bind variables — `id` and `rev` in the `WHEN`, then `id` again in the `IN` list. At `chunked(250)` that's 750 per statement, comfortably under SQLite's `SQLITE_MAX_VARIABLE_NUMBER` of 999 that Android has historically enforced. This is exactly where #16677 goes wrong: it chunks at 450, which is 1350 variables and throws `too many SQL variables` on any device with the 999 limit — a hard failure on the notification sync path rather than a slow one. Getting this right is the single strongest reason to prefer this PR.

The rest holds up: the null-rev arm keeps `chunked(900)` for its single-variable-per-row query, the early `if (syncResults.isEmpty()) return` avoids building anything for the common no-op, and partitioning in one loop rather than two `filter` passes avoids walking the list twice. `markSyncedNonNullRevsRaw` / `markSyncedNullRevs` also say what they do, where `executeRawUpdate` would not.

**One thing I'd take from #16677:** its `CASE` ends with `ELSE rev END` where this one ends with plain `END`. Today they behave identically — `WHERE id IN (...)` is built from the same chunk as the `WHEN` clauses, so every updated row matches a branch and the `ELSE` is unreachable. But with no `ELSE`, a non-matching row would get `rev = NULL`, silently wiping the revision and stranding that notification for future updates. Adding `ELSE rev` costs nothing and makes the statement safe under any future edit that lets the two lists drift apart.

**Note:** this and #16677 are the same change to the same lines, so only one can merge. On the bind-variable limit this one is correct and that one is not — I'm marking #16677 `duplicate` and recommending it close in favour of this.

**Verdict:** mergeable — worth adding `ELSE rev` before it lands.

---

Okuro3499 (approved):

PR #16897, opened 2026-09-04
teams: smoother repository members visiting (fixes #16895)

### LINKED ISSUE #16895: Collapse the member-visit aggregation into one pass and drop the unused MemberStats
## What

The visit-log list is traversed twice to derive two values that one accumulator could produce, and a local data class declared for the job is unused.

**File:** `app/src/main/java/org/ole/planet/myplanet/repository/TeamsRepositoryImpl.kt` (`getJoinedMembersWithVisitInfo`, line 976 — **the visit-log aggregation only**)

## Why

```kotlin
override suspend fun getJoinedMembersWithVisitInfo(teamId: String): List<JoinedMemberData> {   // 976
    data class MemberStats(                                                                    // 977 — declared, never used
    ...
    val visitCounts = logs.groupingBy { it.user }.eachCount()                                  // 1030
```

The counts come from `groupingBy {}.eachCount()`, and the latest-visit timestamp comes from a second pipeline — `logs.groupBy { it.user }.mapValues { ... maxOfOrNull { it.time ?: 0 } }` — which additionally retains a full list per user. One accumulator map lowers both CPU and peak memory on large teams.

`MemberStats` at line 977 is declared and never used.

## The fix

Replace the unused `MemberStats` with a small local aggregate holding a visit count and the greatest non-null timestamp. Traverse `logs` **once**, keyed by the same nullable user key both current collections use, and read count and latest timestamp from it while mapping `orderedMembers`.

Preserve the zero-visits and no-latest-timestamp defaults exactly.

## Acceptance

- Same counts and latest-visit timestamps per member, including members with no logs and logs with a null `time`.
- Null user keys grouped exactly as today.

## Notes

⚠️ **Scope: the visit-log aggregation only.** The bigger hotspot in this function is the per-member `activitiesRepository.getLastVisit` / `getOfflineVisitCount` N+1 — that is **explicitly out of scope here**; don't fold it in.

This is the largest file in the repo (~1437 lines) and other PRs touch it, but their hunks end well before line 976. If you hit a conflict, rebase rather than restructuring.

dogi:
Collapse the two-pass member visit log aggregation in `TeamsRepositoryImpl.getJoinedMembersWithVisitInfo` into a single loop using a local `VisitStats` structure, and remove the unused `MemberStats` data class. Added unit test verifying visit log aggregation correctness.

Fixes #16895

---
*PR created automatically by Jules for task [1651350417311565530](https://jules.google.com/task/1651350417311565530) started by @dogi*


dogi:
**Review — rating: 88/100**

Two independent wins, and the aggregation rewrite is equivalent — I checked the edge that usually isn't.

**The dead `MemberStats` is a genuine find.** It was declared as a local `data class` inside `getJoinedMembersWithVisitInfo` and never instantiated. Kotlin doesn't warn on an unused local class, so it sat there describing a shape the function doesn't build.

**The aggregation collapses two passes into one, and drops the expensive intermediate.** `groupingBy { it.user }.eachCount()` walks the logs once, then `groupBy { it.user }` walks them again and — this is the costly part — materialises a `List` of every log per user, purely so `mapValues` can take a `maxOfOrNull` and throw the lists away. For a team with a long visit history that's the whole log set duplicated into per-user lists. A single loop into `HashMap<String?, VisitStats>` keeps two scalars per user instead.

The `maxOfOrNull` translation is right where it's easy to be subtly wrong. Old: `userLogs.maxOfOrNull { it.time ?: 0 }` — null time coerced to `0` *before* the max, and `maxOfOrNull` can only return null for an empty group, which `groupBy` never produces. New: `stats.latestVisit?.let { maxOf(it, logTime) } ?: logTime` with `logTime = log.time ?: 0L` — first log seeds the value, later ones take the max, same coercion. Both end up with `Long?` that is null only when the user has no logs at all, which both express as a missing map entry. Same values out.

Counting as `Long` from the start also removes the `?.toLong()` conversion at the read site.

**Nit:** `VisitStats` is a `data class` with `var` fields used purely as a mutable accumulator, so `equals`/`hashCode`/`copy` are generated over mutable state and none of them is used. A plain class, or a `Pair`-free private class, would say "accumulator" more clearly.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16899, opened 2026-09-04
all: smoother view mode toggle controlling (fixes #16700)

### LINKED ISSUE #16700: duplicate view-mode toggle logic
```
View-mode toggle logic duplicated verbatim between CoursesFragment.kt:358-414 and ResourcesFragment.kt:278-330 (~55 lines: setupViewModeToggle, setViewMode,                    
applyRecyclerLayoutManager, span-count logic).
```

Okuro3499:
fixes #16700
Move duplicated view-mode toggle logic from CoursesFragment and ResourcesFragment into a new ListViewModeController component. The controller now handles toggle button UI state, list/grid layout manager switching, responsive grid span recalculation on layout changes, and lifecycle-safe setup/teardown, while each fragment only wires in preference getters/setters and adapter updates.

## Summary by CodeRabbit

- **New Features**
  - Added consistent grid and list view switching for Courses and Resources.
  - Grid columns now adjust automatically to available screen width.
  - View-mode selections continue to update the displayed content and toggle controls.

- **Bug Fixes**
  - Improved view layout refresh behavior when returning to these screens or when available width changes.


dogi:
**Review — rating: 88/100**

Net −28 lines across three files, so this is a real deduplication rather than a relocation, and the part that usually goes wrong in an extracted controller — lifecycle — is handled in both call sites.

`teardown()` removes *both* the `OnLayoutChangeListener` and any pending `spanUpdateRunnable`, and both fragments call it from `onDestroyView` and then null the reference (`CoursesFragment:542-544`, `ResourcesFragment:747-749`), with `refreshSpanOnResume()` wired from `onResume` in each. So the controller can't outlive the view it holds, and a posted span recalculation can't land on a destroyed `RecyclerView` — which matters because the controller keeps a `Fragment` and calls `requireContext()` in several places.

Two details in the implementation that are easy to get wrong and are right here:

- **`applyRecyclerLayoutManager` mutates `spanCount` in place** when the layout manager is already a `GridLayoutManager`, instead of constructing a new one. Swapping the layout manager resets scroll position and forces a full rebind; this keeps both.
- **`currentSpanCount()` falls back to `displayMetrics.widthPixels`** when `recyclerView.width` is still 0, so the first call before measurement produces a sane span rather than dividing by zero width. And `updateGridSpanIfNeeded` only assigns when the value actually changed, avoiding a relayout per layout pass.

**Nit:** `updateToggleUi(mode)` ends by calling `applyRecyclerLayoutManager(mode)`, so a function named for the toggle's appearance also installs the layout manager — and `setup()` relies on that side effect to establish the initial layout. It works, but the name hides the more consequential half; splitting the two, with `setup()` calling both explicitly, would make the entry path readable.

**Verdict:** mergeable.

---

dogi (approved):

PR #16901, opened 2026-09-04
chat: smoother history sharing (fixes #16873)

### LINKED ISSUE #16873: One-pass expandedGroups set in ChatHistoryAdapter
## What

Three chained collection operations to build a set of expanded group titles, on every group expand/collapse tap.

**File:** `app/src/main/java/org/ole/planet/myplanet/ui/chat/ChatHistoryAdapter.kt` (line 154)

## Why

```kotlin
val expandedGroups = currentFlatList.filter { it.isGroup && it.isExpanded }.map { it.title }.toMutableSet()  // 154
```

`filter` allocates a list, `map` allocates another, `toMutableSet` allocates the set — three passes and three allocations to produce one set that is then mutated at 156/158 and passed to `generateFlatList` at 160.

## The fix

Build the set in one pass — e.g. `mapNotNullTo(mutableSetOf())` with the group/expanded predicate inside, or an explicit loop. Keep the subsequent `remove`/`add` and the `generateFlatList` call identical.

## Acceptance

- Expanding and collapsing a group produces the same flat list as today.

## Notes

- This file has an open PR touching lines ~226 and ~243+. Your change is at 154, so it shouldn't collide.

dogi:
Optimized `expandedGroups` set construction in `ChatHistoryAdapter.kt` by replacing chained `filter`, `map`, and `toMutableSet` collection operations with `mapNotNullTo(mutableSetOf())`. This avoids creating intermediate list allocations during group expand/collapse taps.

Fixes #16873

---
*PR created automatically by Jules for task [12991780997947811431](https://jules.google.com/task/12991780997947811431) started by @dogi*


dogi:
**Review — rating: 90/100**

`filter { … }.map { it.title }.toMutableSet()` allocated three collections — a filtered list, a mapped list, then the set — to end up with a set of titles. `mapNotNullTo(mutableSetOf()) { … }` builds the set in one pass with no intermediates.

The one thing worth checking is whether `mapNotNull`'s null-dropping can change the result, and it can't: `ChatShareTargetItem.title` is declared `val title: String` — non-null — so the lambda returns null only for items that failed the `isGroup && isExpanded` test, which is exactly what the old `filter` excluded. Same membership, and `MutableSet<String>` keeps `remove(clickedItem.title)` and `add(...)` type-correct below.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16902, opened 2026-09-04
resources: smoother repository download filtering (fixes #16872)

### LINKED ISSUE #16872: Merge the filter and mapNotNull in ResourcesRepositoryImpl.downloadResources
## What

A `filter` followed by a `mapNotNull` over the same collection, where one pass would do.

**File:** `app/src/main/java/org/ole/planet/myplanet/repository/ResourcesRepositoryImpl.kt` (line ~466, `downloadResources`)

## Why

The intermediate filtered list is allocated only to be immediately re-traversed and mapped.

## The fix

Merge into a single `mapNotNull`, moving the filter predicate into the lambda (returning `null` for excluded elements).

## Acceptance

- Same resources selected, same order, empty-input path unchanged.

## Notes

- Keep the download logic itself untouched — this is the collection pipeline only.

dogi:
Merged filter and mapNotNull calls in ResourcesRepositoryImpl.downloadResources into a single mapNotNull pass to avoid intermediate list allocations. Added unit tests covering downloadResources in ResourcesRepositoryImplTest.

Fixes #16872

---
*PR created automatically by Jules for task [1271174410762667353](https://jules.google.com/task/1271174410762667353) started by @dogi*


dogi:
**Review — rating: 88/100**

Fusing `filter { !it.isResourceOffline() }.mapNotNull { it.resourceRemoteAddress }` into a single `mapNotNull` removes the intermediate filtered list while producing the same result: an item is kept exactly when it isn't offline *and* has a non-null remote address, in the same order. The old chain expressed the two conditions in two passes; the new one short-circuits per element.

`isResourceOffline()` is also the more expensive of the two checks and is still evaluated once per element, so nothing is duplicated by the merge.

Coming with a test is the right call for a one-line change in a download path — it pins the "offline resources are excluded" contract, which is the part a future edit could silently invert.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16904, opened 2026-09-04
sync: smoother upload repository exams results marking (fixes #16875)

### LINKED ISSUE #16875: Pre-size the exam upload result buckets
## What

The exam-upload result collections grow by repeated reallocation when the final size is already known.

**File:** `app/src/main/java/org/ole/planet/myplanet/repository/UploadRepositoryImpl.kt` (`markExamsUploaded`, line 63)

## Why

The number of results is bounded by the input size, so the success/failure buckets can be allocated at the right capacity up front instead of growing.

## The fix

Pre-size the result collections from the input count. No change to what ends up in each bucket, or to the order within them.

## Acceptance

- Same items in the same buckets in the same order, including the empty-input case.

## Notes

- This is a small allocation win on a batch path — don't restructure the upload logic around it.
- This file has open PRs touching `uploadAttachment` (~96-111) and possibly a new POST wrapper. Your change is at 63, so it shouldn't collide.

dogi:
Pre-sizes the result collections `updated` and `failed` in `UploadRepositoryImpl.markExamsUploaded` to `succeeded.size` initial capacity to avoid dynamic array grow reallocations.

Fixes #16875

---
*PR created automatically by Jules for task [9046172813598306387](https://jules.google.com/task/9046172813598306387) started by @dogi*


dogi:
**Review — rating: 84/100**

Correct sizing: every element of `succeeded` lands in exactly one of the two buckets, so `succeeded.size` is a true upper bound for each and neither list can ever need to grow. `mutableListOf()` already returns an `ArrayList`, so this is purely about capacity — starting at the default 10 and doubling through 10 → 15 → 22 → … with an array copy each time, which for a full 50-item upload batch is several reallocations per call on the submission path.

**Nit:** sizing *both* to `succeeded.size` means the common case allocates two full-size backing arrays where the split is usually lopsided — a batch of 50 with two failures reserves 50 slots for a 2-element `failed` list. Still cheaper than repeated growth, and the arrays are short-lived, so this is the right trade; just worth knowing it's a deliberate over-allocation rather than a tight fit.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16905, opened 2026-09-04
courses: smoother repository resources filtering (fixes #16865)

### LINKED ISSUE #16865: Collapse the map-then-filter in flushPendingCourseResources
## What

An intermediate list is built only to be filtered away.

**File:** `app/src/main/java/org/ole/planet/myplanet/repository/CoursesRepositoryImpl.kt` (line 830)

## Why

```kotlin
val resourceIds = batch.map { JsonUtils.getString("_id", it.doc) }.filter { it.isNotBlank() }   // 830
```

`map` allocates the full list of ids including blanks, then `filter` allocates a second.

## The fix

Collapse into a single `mapNotNull` with a `takeIf` on the extracted id.

Rename the inner lambda parameter rather than shadowing `it` — the current expression already has a nested `it.doc`, and reusing `it` in a `mapNotNull` here reads badly.

No functional change.

## Acceptance

- Same ids, same order, blanks still excluded.

## Notes

- This file has other open PRs against it, at lines ~441-460 and ~684. Your change is at 830, so it shouldn't collide.

dogi:
Collapse batch.map { ... }.filter { ... } into batch.mapNotNull { pending -> ... } in flushPendingCourseResources to avoid intermediate list allocations.

Fixes #16865

---
*PR created automatically by Jules for task [6082111512728209238](https://jules.google.com/task/6082111512728209238) started by @dogi*


dogi:
**Review — rating: 88/100**

`map { … }.filter { it.isNotBlank() }` built a full list of ids including the blanks, then a second list without them. `mapNotNull { … takeIf { it.isNotBlank() } }` produces the same ids in the same order in one pass with one allocation — and since the result feeds `.distinct().chunked(300)` for a batched `IN` query, order and content are all that matter, both preserved.

Naming the outer lambda parameter `pending` so the inner `takeIf { it.isNotBlank() }` unambiguously refers to the extracted string is the right call — nested `it` is exactly where this idiom gets misread.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16906, opened 2026-09-04
life: smoother achievement view modelling (fixes #16814)

### LINKED ISSUE #16814: Expose the achievement document id from AchievementViewModel
## What

The achievement document id is derived in two places — once to read, once to write — so the read and the write can silently disagree.

**Files:**
- `app/src/main/java/org/ole/planet/myplanet/ui/user/AchievementViewModel.kt` (`loadUserAndAchievement`, line 59)
- `app/src/main/java/org/ole/planet/myplanet/ui/user/EditAchievementFragment.kt` (save handler, line 166)

## Why

The id `"<userId>@<planetCode>"` is built at `AchievementViewModel.kt:59` to **load** the achievement, and rebuilt as `user?.id + "@" + user?.planetCode` at `EditAchievementFragment.kt:166` to **save** it.

That's a persistence-format concern living in the view. If the format ever changes, the fragment writes a different document than the ViewModel reads — and nothing fails loudly; the user's achievement just quietly stops round-tripping.

## The fix

Expose the id derived during `loadUserAndAchievement` — e.g. a `StateFlow<String?>` set alongside `_achievement` — so the derivation exists in exactly one place, and consume it in the save handler.

⚠️ If the id is unavailable because the user hasn't loaded yet, keep the save button's **current guard**. Never save with a fabricated or partially-built id — writing to the wrong document is worse than not saving.

## Acceptance

- The string `"@"` concatenation appears once in the codebase for this id.
- Saving before the user loads is still blocked by the existing guard.
- Load-then-save round-trips to the same document.

## Notes

dogi:
Expose achievementId as a StateFlow<String?> from AchievementViewModel during loadUserAndAchievement() and consume it in EditAchievementFragment to prevent ID mismatch between load and save.

Fixes #16814

---
*PR created automatically by Jules for task [306144349117089501](https://jules.google.com/task/306144349117089501) started by @dogi*


dogi:
**Review — rating: 86/100**

This is a bug fix, not just an extraction, and the title undersells it.

`userModel?.id + "@" + userModel?.planetCode` uses Kotlin's `String?.plus`, which renders a null receiver as the literal `"null"`. So with no logged-in user the old code produced the achievement document id **`"null@null"`** — and then `initializeAchievement("null@null")` created or loaded a document under it, while `EditAchievementFragment` saved to the same id. Both sites built that string independently, so both were exposed. Guarding with `userModel?.let { … }` and `id?.let { … }` means no user now yields no achievement rather than a shared junk document, and the fragment's `?: return@setOnClickListener` refuses the save instead of writing to it.

Exposing the id from the ViewModel is the right shape too: the fragment stops re-deriving a value the ViewModel already computed, so the two can no longer disagree about what a user's achievement id is.

**Nit — the same hazard survives one level down.** `it.id + "@" + it.planetCode` still uses `String?.plus` on the *fields*, so a user row with a null `id` or `planetCode` produces `"null@..."`, and `takeUnless { it.isNullOrEmpty() }` won't catch it because that string isn't empty. If those fields can be null in practice, the guard wants to be on the components:

```kotlin
val id = userModel?.let { u ->
    val uid = u.id?.takeIf { it.isNotBlank() } ?: return@let null
    val code = u.planetCode?.takeIf { it.isNotBlank() } ?: return@let null
    "$uid@$code"
}
```

⚠️ **`myPlanet test` was still in progress** when I checked (build green). Worth confirming it lands green before merge.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16910, opened 2026-09-04
sync: smoother repository shelf dispatch mapping (fixes #16876)

### LINKED ISSUE #16876: Replace SyncRepositoryImpl's shelf when-block with a dispatch map
## What

Shelf-type handling is a long `when` block where a dispatch map would express the same routing.

**File:** `app/src/main/java/org/ole/planet/myplanet/repository/SyncRepositoryImpl.kt` (`processShelfDataOptimizedSync`, line 123)

## Why

Each shelf type maps to a handler; the `when` interleaves that routing table with the handling itself, so adding a shelf type means editing control flow rather than data.

## The fix

Extract the shelf-type → handler routing into a map, and dispatch through it. Keep each handler's behavior identical, including the fallback for an unknown shelf type.

## Acceptance

- Every shelf type routes to the same handler as today.
- An unknown shelf type behaves exactly as it does now (verify what that is before changing anything — silently ignoring and throwing are very different, and the current behavior is what must be preserved).

## Notes

- An open PR adds cache helper methods to this file. Your change is at ~123; if you hit a conflict, rebase rather than reorganizing.

dogi:
Replaced SyncRepositoryImpl's shelf when-block in processShelfDataOptimizedSync with a dispatch map, and added unit tests in SyncRepositoryImplTest.

Fixes #16876

---
*PR created automatically by Jules for task [16551918537313911229](https://jules.google.com/task/16551918537313911229) started by @dogi*


dogi:
**Review — rating: 80/100**

The production change is a lateral move, not an improvement: a 4-branch `when (shelfData.type)` becomes a `by lazy` map of four suspend lambdas plus a null check. Same dispatch, same four call sites, roughly the same line count, one more layer of indirection to step through when debugging a sync. It's correct — the lambda signatures match each repository's arity, the unknown-type case still no-ops exactly as the `else`-less `when` did, and `by lazy` means the map is built once per repository instance rather than per shelf — but on its own it wouldn't earn a merge.

The 178-line test file is what makes this worth taking. `processShelfParallel dispatches known shelf types to correct repositories` pins the four-way routing with `coVerify(exactly = 1)` per repository and asserts the aggregate count, and `processShelfParallel handles unknown shelf type by performing no dispatch` pins the silent-no-op behaviour that the old `when` left implicit and undocumented. That second test is the real value: the missing `else` was previously indistinguishable from an oversight, and now it's a stated contract. Anyone who later adds a fifth shelf type gets a failing assertion instead of silently dropping documents.

Test style follows the project conventions — MockK, `runTest`, `TestDispatcherProvider`, no hard-coded `Dispatchers.*`.

CI green on both flavors and both test shards.

**Verdict:** mergeable, on the strength of the tests.

---

Okuro3499 (approved):

PR #16920, opened 2026-09-07
feat(flutter): port the system-tray notification tap entry points (Phase 130, Lane C)

dogi:
Lane C of the Phase 130 parallel round. Draft, opened as the lane's comms channel.

## Task 1 — a correct handler nothing can reach

Phase 127 left `NotificationsRepository.markNotificationAsRead` with no caller: it is the
correct port of Kotlin's tray/dashboard handler, but `NotificationActionReceiver` and
`DashboardActivity.handleNotificationIntent` are unported. The port raises system
notifications (`TaskDeadlineNotifier`) and **tapping one does nothing** — no navigation, no
read-marking. Porting the entry points, reusing `NotificationDestinationResolver` rather than
deriving a second mapping, and extending `test/ui/route_reachability_test.dart` to cover the
tray entry.

## Task 2 — the derivation eats a deliberate trailing space

`tool/arb_from_strings_xml.dart`'s plain-text paths `.trim()` without `_mirrorTrailingSpace`.

## Files owned

`lib/ui/notifications/`, `lib/repository/notifications_repository.dart`,
`lib/core/notifications/`, the notification platform/receiver wiring, notification-intent
handling in `lib/ui/router.dart` and the app entry, `tool/arb_from_strings_xml.dart`,
`lib/l10n/*.arb`, `test/l10n/`, and their tests.

Lane A may relay `%d`-carrying `take_test`/`retake_test` keys for derivation into the locale
files — comment on this PR and it reaches the lane.

https://claude.ai/code/session_01RXNiy1J9wcB7shZQJNjLTD

---
_Generated by [Claude Code](https://claude.ai/code/session_01RXNiy1J9wcB7shZQJNjLTD)_

### DIFF
--- a/flutter/PHASE_130_NOTES.md
+++ b/flutter/PHASE_130_NOTES.md
@@ -1,11 +1,433 @@
 # Phase 130 — the tray tap, and a trailing space the derivation ate
 
-Work in progress. Two items:
+Lane C. Two items, both handed over by Phase 127's *Reported, not fixed* list:
+a correct handler nothing could reach, and one character the ARB derivation
+threw away.
 
-1. **A correct handler nothing can reach.** Phase 127 left
-   `NotificationsRepository.markNotificationAsRead` with no caller, because
-   `NotificationActionReceiver` and `DashboardActivity.handleNotificationIntent`
-   are unported: the port raises system notifications and tapping one does
-   nothing. Porting the entry points.
-2. **`tool/arb_from_strings_xml.dart` drops a deliberate trailing space** on the
-   plain-text by-name path.
+Two `parity-auditor` passes at `effort: max` were run, one on the Kotlin before
+any Dart and one on the finished diff. **The first overturned three of the eight
+claims this lane had written down**, and one of the three changed the design:
+see *What the ground-truth audit corrected*. That is the fourth round running
+where a lane's own reading of the Kotlin was wrong in a way only the audit
+caught.
+
+---
+
+## Item 1 — the tray tap
+
+### What the port actually raises
+
+One thing. `TaskDeadlineNotifier` (Phase 42) is the port's only producer of an
+OS notification, through `NotificationConfig.task`:
+
+```
+id        = task.id          type = "task"
+relatedId = task.id          actionable = true   (from this phase)
+title     = "✅ New Task Assigned"
+message   = "<title>\nDue: <EEE dd, MMMM yyyy>"
+```
+
+That matches `TaskNotificationWorker` → `NotificationUtils.createTaskNotification`,
+and — with one refinement the audit supplied — it is the **only** Kotlin producer
+that routes through `NotificationUtils.NotificationManager` and therefore the
+only one that carries extras or action buttons. The other five factories
+(`createSurveyNotification`, `createJoinRequestNotification`,
+`createStorageWarningNotification`, `createResourceNotification`,
+`createSummaryNotification`) are called only from `DashboardActivity.kt:214`'s
+`state.newNotifications.forEach { showNotification(it) }`, and
+`DashboardUiState.newNotifications` has exactly two writers: its own
+`emptyList()` default and `clearNewNotifications`. Nothing populates it. So
+those five have no live caller, and **`summary_`-prefixed notification ids are
+never minted at runtime.**
+
+### What each notification type does on tap
+
+There are three gestures, not one, and the Kotlin gives each a different
+behaviour. This is the table the brief asked for; every row is reachable in the
+port today.
+
+| gesture | Kotlin path | read-marking | navigates to |
+|---|---|---|---|
+| the notification **body** | `setContentIntent` → `DashboardActivity` with `from_notification=true` → `handleNotificationIntent`'s first branch | `markNotificationAsRead(id, userId)` — **does not** stamp `createdAt`; `summary_`-aware | the destination for its type |
+| **Mark as Read** button | `NotificationActionReceiver` `ACTION_MARK_AS_READ` (`:34-39`) | `markNotificationsAsRead({id})` — **stamps** | nowhere (but see below) |
+| **View Task** button | `NotificationActionReceiver` `ACTION_OPEN_NOTIFICATION` (`:51-66`) → `DashboardActivity` with `auto_navigate=true` → the second branch | `markNotificationsAsRead({id})` — **stamps** | the destination for its type |
+
+And the destinations, by type, which is the other half of the question. The port
+resolves all of them through the one shared resolver, so this table is
+`NotificationDestinationResolver` plus `notificationDestinationLocation`:
+
+| type | port location | Kotlin |
+|---|---|---|
+| `task` | `/life/teams/<teamId>/tasks` | `TeamDetailFragment(navigateToPage = TasksPage)`, team id from the cached task's `link.teams` |
+| `join_request` | `/life/teams/<teamId>/members?tab=requests` | `TeamDetailFragment(JoinRequestsPage)`, after stripping a `join_request_` prefix |
+| `team_join` | `/life/teams/<teamId>` | `TeamDetailFragment`, no page |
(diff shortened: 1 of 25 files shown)


dogi:
Lane C status — both items landed, head `23ee7b0`. Gate green: format clean, analyze clean, **2299 tests** (was 2265). Full detail in `flutter/PHASE_130_NOTES.md`. A second `parity-auditor` pass over the finished diff is still running.

**Task 1 — the tray tap.** Three gestures, three behaviours, as in the Kotlin: the **body** marks read *without* stamping `createdAt` and navigates; **Mark as Read** stamps and navigates nowhere; **View Task** stamps and navigates. `NotificationsRepository.markNotificationAsRead` — dead since Phase 127 — is the body arm's callee.

`NotificationDestinationResolver` is reused, not re-derived: `resolveFor(type:, relatedId:)` split out of `resolve(row)`, and the kind→location switch lifted out of `notifications_screen.dart` as `notificationDestinationLocation`. A tray tap and a bell-row tap land in the same place by construction.

The ground-truth audit **overturned three of eight** claims and one changed the design: every arm of `NotificationActionReceiver` ends in an unconditional `startActivity(DashboardActivity)`, so *Mark as Read* launches the app in the Android app too. Both actions are therefore `showsUserInterface: true` — no background isolate needed.

**Reproduced rather than fixed:** no tray gesture marks anything read on real data, in either app. The notification is minted with the *task* id while a task notification row is a synced document keyed on its own `_id`. Resolving `relatedId` back to the row would be an improvement Android lacks, so it is tested and recorded instead.

**The reachability guard covers a tray entry, three ways** (`test/ui/route_reachability_test.dart`): every `NotificationDestinationKind`'s location resolves (exhaustive over the enum), the tap the port's own producer delivers reaches a route, and every action button id is one the handler recognises — plus a source-text rule on the three plugin arguments no test can exercise, since `FlutterLocalNotificationsPlugin` has a private constructor.

Thirteen injections, thirteen red. One found a gap first: feeding the resolver `notificationId` instead of `relatedId` left the suite green, because the task factory sets both to the same string. A `voice_reply` payload whose two fields differ now separates them.

**Task 2 — the trailing space.** The demonstrating key already existed and was already wrong: `storageRunningLow` and `storageAvailable` are `"Storage running low: "` / `"Storage available: "` in the Kotlin XML and lost the space in **all five locales**, while `selected`/`selectResources` — repaired by hand through `--adopt`, which mirrors it — kept theirs. Both plain-text rules now go through one `derivePlainTextValue`. The same run picked up `playbackSpeed`/`playbackSpeedValue`, derivable and never run through the tool. Pinned counts +2 per locale.

**One crossing:** a comment-only correction in `lib/data/local/app_database.dart` — `markAsRead`'s dartdoc claimed selection mode was "its one caller".

@integrator — Lane A's `%d`-carrying `take_test`/`retake_test` keys: comment here with the `app_en.arb` entries and I'll run the derivation into the five locale files.

---

dogi:
Second `parity-auditor` pass done — seven findings in code that was green, formatted and analyzed. All closed. Head `97bfe4d`, **2309 tests**, format and analyze clean.

**The one that matters arrived by accident.** The pass mutation-tested `notification_tap_scope.dart` — deleting the `launchTap()` call, then the `go(location)` call — and **all 2301 tests stayed green**. Nothing mounted `NotificationTapScope`, so the two lines connecting the whole chain to the app were the one part with no guard; every other link had one, including a rule asserting the scope is *mounted* in `app.dart`. What the mounted widget then did was unchecked. With the first mutation a learner taps a deadline reminder on a closed app and lands on `/` — and the cold-start case is the only one that matters for a reminder raised by a headless worker.

It cost something to learn: the mutations were in the tree when a `git add -A` ran, so `452b9b6` shipped them and `11d19d9` restored them. Only that commit, only that file — verified with `git log -S`. **Lesson for the other lanes: do not `git add -A` while a background agent has the tree.** Six tests now cover the scope; both shipped mutations go red, as do "never subscribes" and "awaits the launch tap before subscribing".

**Two guards passed for the wrong reason**, both source-text assertions that matched a *name* where they meant a *value*:
- `contains('onDidReceiveNotificationResponse:')` survived being set to `null`, which drops every warm tap.
- Nothing pinned the one call in the UI isolate that reaches `initialize` at all — `main.dart`'s permission request. Move it to a screen and `taps()` is dead for the process while `launchTap()` keeps working. Its first cut matched the function's own *declaration*, so the injection stayed green until narrowed further.

That is the failure mode of source-text assertions, and the reason each has to be injected against.

**A type nothing can reach:** `resolveFor` has no `survey`/`course` arm — two of Kotlin's six `TYPE_*` values, and the bell-row handler it was ported from has neither. A new rule walks every `NotificationTypes` constant through `resolveFor`, reconciling its list against the source file.

**A fourth route past the trailing-space fix:** `convertAndroidFormat` still trimmed. Unreachable today, closed anyway — the guard that would catch it would have failed with no way for the tool to produce a passing value.

**Three comments stated something false** and are corrected in place: the two Kotlin intents do not carry the same extras; "the plugin does not double-deliver" holds for a narrower reason than its dartdoc suggests (Android's `onNewIntent` also calls `setIntent`, so a delivered tap is re-reported as a launch tap for the rest of the process — safe only because `_start` runs once, and `OutboxDrainScope` next door does the resume-rerun that would break it); and a background isolate *can* localise.

**All six claimed divergences were challenged and all six held**, including the `showsUserInterface: true` reading and the deliberate tray-marks-nothing-read parity.

Injection tally now 23, all red.

---

dogi:
Lane C complete. Head **`97bfe4d`**, all five checks green, no conflict. **2309 tests** (was 2265). Both `parity-auditor` passes at `effort: max` are done and their findings closed.

Updating my earlier comment on the one thing that changed materially since it.

## The second pass found seven items in already-green code, and one matters

It mutation-tested `notification_tap_scope.dart` — deleted the `launchTap()` call, then the `go(location)` call — and **all 2301 tests stayed green.** Nothing mounted `NotificationTapScope`, so the two lines connecting the whole chain to the app were the one part of it with no guard. Every other link had one, *including a source-text rule asserting the scope is mounted in `app.dart`* — but what the mounted widget then did was unchecked. With the first mutation a learner taps a deadline reminder on a closed app and lands on `/`, and the cold-start case is the only one that matters for a reminder raised by a headless worker.

This phase's own subject, one layer up. It also cost something: the mutations were in the working tree when a `git add -A` ran, so `452b9b6` shipped them and `11d19d9` restored them. **Do not `git add -A` while a background agent has the tree.** `test/ui/notification_tap_scope_test.dart` now covers the scope with six tests following `deep_link_scope_test.dart`; both shipped mutations go red.

**Two guards passed for the wrong reason**, both source-text assertions that matched a *name* where they meant a *value*: `contains('onDidReceiveNotificationResponse:')` survived that parameter being set to `null` — which drops every tap arriving while the app is running — and nothing pinned the one call in the UI isolate that reaches `initialize` at all. Move `main.dart`'s permission request to a screen, a plausible refactor, and `taps()` is dead for the whole process while `launchTap()` keeps working. Both narrowed and re-injected.

**Also closed:** a new rule walks every `NotificationTypes` constant through `resolveFor` (the resolver came from the bell row's handler, which has no `survey` or `course` arm, so a future producer's taps would fall silently to `default`); the ICU derivation path was a fourth route past the trailing-space fix and now mirrors too; and three comments that stated something false about the Kotlin or the plugin, corrected in place because each would otherwise be cited.

**All six claimed divergences were challenged and all six held.**

## Standing items for whoever picks them up

- **`updateResourceNotification` is ported, has six tests, and has no production caller** — and the count it needs (`countLibrariesNeedingUpdate`) has no port under any name. The bell's undownloaded-resources row never appears and `notification_format.dart`'s resource arm is dead with it. Needs a resources count plus a dashboard-load hook, both outside this lane's files. Three pieces named in the notes.
- **No tray gesture marks anything read on real data**, in either app. Reproduced deliberately; if it is ever fixed it should be fixed in both apps together or the bell badges disagree.

Full detail in `flutter/PHASE_130_NOTES.md`. Ready for integration; still watching for review comments.

---

PR #16940, opened 2026-09-08
Phase 138 (Lane A): the adopted survey clone's lifecycle, and pendingUploads' missing status test

dogi:
Lane A of a four-lane Phase 138 round. Draft; opened as the first action so the integrator has an inbound channel (`subscribe_pr_activity` is called on it).

## Job 1 — an adopted team survey clone is destroyed on the next sync

Phase 136's top *Reported, not fixed* item. Kotlin uploads the clone (`ExamDao.getPendingAdoptedSurveys()` → `UploadConfigs.kt:186-195`, endpoint `exams`) so it gains a `_rev` and joins every later walk's keep set, and Kotlin has no delete-except-ids on that table at all. The port has neither half: no adopted-survey uploader, and `SurveyDao.deleteNotIn` prunes every row without a `stepId` that the server did not name. The clone and its question rows are gone after one surveys sync, and its answer sheets are orphaned.

Three pieces: a pending-adopted query plus a `deleteNotIn` exemption, an `AdoptedSurveysUploader` on the outbox, and the `adoptSurvey` enqueue.

## Job 2 — `pendingUploads` has no status test where Kotlin's sweep does

Phase 134's top item, made systematic by that phase's safety-net sweep. A behaviour decision: either a status predicate, or a recorded deviation in `docs/kotlin-to-flutter-migration.md`.

## Files this lane owns

`flutter/lib/data/local/app_database.dart`, `flutter/lib/repository/{surveys,submissions,outbox}_repository.dart`, `submissions_uploader.dart`, `outbox_drainer.dart`, a new `adopted_surveys_uploader.dart`, `providers/app_providers.dart` (registration only), `docs/kotlin-to-flutter-migration.md`, `flutter/PHASE_138_NOTES.md`, and tests for the above.

https://claude.ai/code/session_01WCCysiGYa9bmY1GFQo1drT

---
_Generated by [Claude Code](https://claude.ai/code/session_01WCCysiGYa9bmY1GFQo1drT)_

### DIFF
--- a/docs/kotlin-to-flutter-migration.md
+++ b/docs/kotlin-to-flutter-migration.md
@@ -886,6 +886,73 @@ Deliberate *deviations*, all flagged in code:
   while the only two type literals the tree does write are the singular `"exam"` and `"survey"`
   fallbacks, which no Kotlin query ever selects on.
 
+- **An adopted team survey's upload records the CouchDB rev; Kotlin's records nothing.**
+  `UploadConfigs.AdoptedSurveys` (`UploadConfigs.kt:186-195`) declares no
+  `responseHandler` and no `markUploaded` — unlike `Meetups` five lines above it, which has
+  both — so the local row keeps `_rev = NULL`, stays in
+  `ExamDao.getPendingAdoptedSurveys()` for the life of the install, and is re-POSTed on
+  every sweep; `serializeExam` always writes `_id` (`StepExam.kt:73`), so CouchDB answers
+  the second POST with a 409 and every one after it, silently.
+  `AdoptedSurveysUploader.handler` calls `SurveyDao.markUploaded` instead. The rev is also
+  what scopes the port's prune exemption: Kotlin never deletes from `exams` at all
+  (`bulkInsertExamsFromSync` upserts, and `ExamDao.deleteById` has zero production callers),
+  while the port's `SurveyDao.deleteNotIn` has to spare an unpublished clone explicitly —
+  and only an unpublished one, so a document the server has stopped naming is still pruned.
+
+- **`hasUnfinishedSurveys` skips an adopted team clone; Kotlin's gate has no such rule
+  because it cannot see one.** `getSurveysByCourseId` filters
+  `getByCourseIdAndType(courseId, "survey")` — **singular**
+  (`SubmissionsRepositoryImpl.kt:367-379`) — while a clone copies the source's `type`
+  (`SurveysRepositoryImpl.kt:180`) and every list adoption is reachable from queries
+  `type = "surveys"` (`ExamDao.kt:29-31`). The port's table split *is* the type, so its
+  `getByCourseId` returns the clone and the guard is a port-invented one on port-invented
+  breadth. It is keyed on `stepId == null` — does the courses walk own this row's course
+  join? — and deliberately **not** on `rev == null`: that read was exact only while nothing
+  published a clone, and Phase 138 publishes it, at which point the gate would demand a
+  team's private copy of every learner on the course and make
+  `MANDATORY_SURVEY_COURSE_ID` uncompletable for outsiders.
+
+- **The team-adoption marker is not uploaded; everything else with an `isUpdated` flag
+  still is.** Kotlin's sweeps are
+  `status = 'complete' AND (isUpdated = 1 OR _id IS NULL OR _id = '')` (`SubmissionDao:41`)
+  and, for exams, `type = 'exam' AND … AND (_id IS NULL OR _id = '')` (`:40`). The port
+  merges the two configs into one uploader and gates on `isUpdated`, because an exam finishes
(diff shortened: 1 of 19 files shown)


dogi:
**For the integrator — two files outside this lane's listed set, deliberately.**

`flutter/lib/providers/dashboard_sync_provider.dart` and `flutter/lib/background_entrypoint.dart`, two lines each: the `AdoptedSurveysUploader.queuePending` sweep, immediately ahead of the existing submissions sweep.

Both are in **no other lane's** set this round (Lane B owns `ui/courses/*`, `ui/surveys/take_survey_screen.dart`, `ui/router.dart`; Lane C `ui/chat/*` and the chat/voices repositories; Lane D the `.arb` files and the l10n tool). The alternative was shipping an uploader with no caller — ported, tested, green, and dead, which is the exact failure class this phase exists to close. Phase 134 established these same two files as the sweep sites for the sibling uploader, and the addition is a literal port of `SubmissionsUploader.kt:83-84`'s two adjacent calls. Flagging it rather than letting a clean merge imply it composed.

**Two other things worth your attention on merge:**

1. The change touches `flutter/lib/data/local/app_database.dart` (four methods) but bumps **no** `schemaVersion` — a prune exemption, a new query and a rev write alter no DDL, and `rev`/`sourceSurveyId` already exist on `Surveys`. **Drift `schemaVersion` 47 is unspent** and available to the next round.

2. `flutter/lib/repository/submissions_repository.dart` — one line in `hasUnfinishedSurveys`. Phase 136 also edited that function, so if any lane is near it, this hunk is the one to read: publishing the clone (this phase's own change) made Phase 136's `rev == null` clone-skip stop firing, which made `MANDATORY_SURVEY_COURSE_ID` uncompletable for anyone outside the adopting team. The guard is now `stepId == null`. Details in `flutter/PHASE_138_NOTES.md`.

Gate is green on `fe7074c`: `dart format` clean, `flutter analyze` clean, **2472 tests pass** (2457 before). The second (implementation) `parity-auditor` pass is still running; I'll follow up here if it finds anything.

---

dogi:
**Follow-up: the implementation audit found five defects in this phase's own green code, all now fixed in `f7b7fef`.** Two things in it change what the integrator needs to know.

### Drift `schemaVersion` 47 is now **spent** — the next round must take 48

My earlier comment said 47 was unspent. That's no longer true. `Surveys.needsSync` is the fix for the audit's most serious finding: the literal port of `ExamDao.kt:21` (`sourceSurveyId IS NOT NULL AND _rev IS NULL`) is a **data leak in the port and not in Kotlin**, because `JsonUtils.getString` returns `""` for a missing key while the port's mappers use `getStringOrNull`. So Kotlin's `_rev IS NULL` names exactly one writer, and the port's names every survey whose document omitted `_rev` — every course-embedded one (a sub-object carries no `_rev`) and every public-API one. Reproduced: the sweep selected **another team's private copy**, `teamId` and all, and would have POSTed it to `exams` under the signed-in user's credentials.

No `_addColumnIfMissing` step is needed (`surveys` is not preserved), and the `if (from < 47)` branch is present and deliberately empty so the version doesn't read as forgotten.

### Nine existing `SurveyRow(...)` fixtures in seven test files gained one line

A non-nullable Drift column is a required argument on the row class. Three are under `test/ui/` — the nearest this lane came to another lane's territory:

- `test/repository/search_surveys_test.dart`, `submissions_repository_test.dart`, `submissions_sync_round_trip_test.dart`
- `test/ui/surveys/survey_team_context_test.dart`, `test/ui/surveys_screen_test.dart`, `test/ui/team_surveys_screen_test.dart`

Each is `needsSync: false,` beside the existing `teamShareAllowed: false,` and carries no behaviour. If a lane is editing those files, this is a trivial conflict to resolve in either direction.

### Also worth flagging

- **A `Deliberate deviations` entry was factually wrong and is corrected.** I claimed Kotlin's adopted-survey upload never records the rev, reading the config's missing `markUploaded`. `UploadConfig` carries a **default** `persistUploaded` keyed on the model class that reaches `exam._rev = result.remoteRev` — Kotlin is at parity. The deviations list is a contract for later rounds, so a wrong entry there is worse than a wrong comment.
- **`lib/data/local/tables.dart` is now touched too** (the one column), on top of the two sweep-site files already flagged.
- Reported, not fixed, in `flutter/PHASE_138_NOTES.md`: `surveys.rev` has two sync walks that disagree about who owns it (`SurveyMapper._build` writes it unconditionally and clobbers the exams walk's value with NULL, where `ExamMapper.fromDoc` uses `_presentOrAbsent` for exactly this reason). Nothing here depends on that column any more, but the next round shouldn't build on it without settling it.

Gate green on `f7b7fef`: format clean, analyze clean, **2479 tests** (2457 at branch point). 21 mutations run across both rounds, all caught.

---

PR #16946, opened 2026-09-08
courses: smoother progress dark mode coloring (fixes #16925)

### LINKED ISSUE #16925: course progress not visible in dark mode
course progress not visible in dark mode
https://github.com/user-attachments/assets/49c4d300-3825-4494-a560-e5f699df9973

J-S-webskas:
Also changed progress tracker bg to a lighter color:

BEFORE (progress bar color change)
Recording_20260908_175321

AFTER
Recording_20260908_174833

## Summary by CodeRabbit

- **Style**
  - Lightened the progress track background color in dark mode for improved visibility.
  - Updated course progress percentage labels to use a consistent color across light and dark themes.


Okuro3499 (dismissed):
issue not fixed progress percentage still not visible
Image

dogi:
**Review — rating: 90/100**

Two lines, and I chased both down to the colour values because "changed dark mode colours" is the kind of change that's either exactly right or quietly wrong. Both are exactly right.

**`progress_track_bg` `#3A3B3C` → `#6E6F71`.** In `values-night`, `card_bg` resolves to `colorPrimaryLight` = `#3a3b3c` — *the same value the track was*. So in dark mode the unfilled portion of the progress bar was rendering in the identical colour as the card behind it: the track was invisible and the bar read as a floating green sliver with no scale. `#6E6F71` separates it from the card while staying clearly distinct from `progress_fill` (`#9CCC65`), so filled and unfilled remain unambiguous.

**`tv_percent` `daynight_textColor` → `md_black_1000`.** This one looks like it hard-codes away a theme-aware colour, which is usually the wrong move — but the pill underneath is not theme-aware:

```xml
<!-- drawable/bg_percent_pill.xml -->
<solid android:color="@android:color/white" />
```

No `-night` variant, so the pill is white in both themes. Meanwhile `daynight_textColor` is `@color/md_black_1000` in `values/` and `@color/textColorPrimary` = `#e4e6eb` in `values-night/`. Near-white text on a white pill: the percentage was unreadable in dark mode. Pinning it to `md_black_1000` is the correct fix, and since that's already what `daynight_textColor` resolves to in the light theme, light mode is byte-for-byte unchanged — only dark mode moves.

It also matches the pattern already in this layout: `tv_subject_label` sits on `bg_type_label_pill` and hard-codes `@android:color/white` for the same reason. Colour-on-a-fixed-pill isn't theme-dependent.

I checked the blast radius: `bg_percent_pill` is used only in `item_course_grid.xml`, so there's no second site left half-fixed. `fragment_take_course.xml` also uses `daynight_textColor`, but those TextViews sit on the fragment background rather than a pill, so they're correct as they are.

CI green 5/5.

**Verdict:** mergeable.

---

Okuro3499 (approved):

dogi (approved):

PR #16947, opened 2026-09-08
dashboard: smoother guests visits view modelling (fixes #16945)

### LINKED ISSUE #16945: huge dashboardActivity
DashboardActivity carry business logic that belongs in their ViewModels.

Okuro3499:
fixes #16945
Move guest offline-visit logic into `DashboardViewModel` via a new `GuestVisitState` model, including banner/dialog/drawer decision rules and a memoized async fetch of visit count. `DashboardActivity` now consumes this shared state in both first-launch drawer behavior and guest warnings, and switches banner inflation to view binding. Tests were expanded to cover guest-state thresholds, non-guest behavior, memoization, and required progress mock setup.


dogi:
**Review — rating: 85/100**

The guest-visit thresholds were spelled out twice in `DashboardActivity`, each site running its own `getOfflineVisitCount` query, and the two expressed the same rules in different shapes — one as `!(startsWith("guest") && offlineVisits >= 3)`, the other as a `when` nested inside an `if (isGuest)`. Moving both into `GuestVisitState` gives one place where the thresholds live and a table-driven test that pins them.

I checked the translation is faithful:

- `shouldAutoOpenDrawer` = `!(isGuest && offlineVisits >= 3)` — identical to the old inline condition, with the orientation check left at the call site where it belongs.
- `shouldShowTrialEndedDialog` = `isGuest && >= 4`, `bannerMessageRes` = `== 2 → guest_visit_limit_warning`, `== 3 → last_login_message`, null otherwise. The new `when` puts the dialog first, matching the old `when`'s `>= 4` branch ordering. Non-guests get null and false from every accessor, so the outer `if (isGuest)` guard is correctly absorbed rather than dropped.

**There's a real win hiding in the refactor that the description doesn't claim:** the old code computed `offlineVisits` unconditionally and only *then* checked `startsWith("guest")`, so every non-guest paid for a `getOfflineVisitCount` query on every dashboard open. The new version skips the repository entirely for non-guests, and there's a test asserting `coVerify(exactly = 0)` on exactly that. Combined with the memoization, two queries per launch become at most one.

**The one thing I'd change: the memoization ignores its own parameter.**

```kotlin
fun getGuestVisitState(userId: String?): Deferred<GuestVisitState> {
    return guestVisitStateDeferred ?: viewModelScope.async { ... }.also { guestVisitStateDeferred = it }
}
```

Once populated, the cached `Deferred` is returned for *any* `userId`. A `DashboardViewModel` is scoped to one Activity and one user, so this can't misfire today — but the signature advertises a keyed lookup that isn't one, which is the kind of thing that quietly breaks the day someone adds account switching. Either key the cache on `userId`, or drop the parameter and have the ViewModel hold the user it's already scoped to. Related and smaller: `guestVisitStateDeferred` is a plain `var` read-then-written with no synchronization; both callers are on `lifecycleScope` (main), so it's fine, but the invariant is worth a comment.

Swapping the `findViewById` banner inflation for `BannerOfflineVisitWarningBinding` is a clean win and takes a stale `ImageButton` import with it.

One unrelated line rode along: `every { progressRepository.findProgressForCourse(any(), any()) } returns null` added to a pre-existing challenge-dialog test. Nothing in this change touches that path, so it's presumably a latent mock gap that surfaced when the constructor grew — worth saying so in the description rather than leaving it to look like part of the refactor.

CI green 5/5.

**Verdict:** mergeable.

---

dogi (approved):

PR #16961, opened 2026-09-09
Phase 150 (Lane A): preserve my_library across a schema bump, and four app_database corrections

dogi:
Lane A of the Phase 150 four-lane round. **Complete — gate green, CI green, both `parity-auditor` passes run.** Draft left as draft; the integrator merges the branch.

## Jobs — all five landed

1. **The schema bump destroyed resources the user created offline** and left `resourceOffline` false on every downloaded resource (`PHASE_146_NOTES.md` *Reported, not fixed* item 1). `my_library` is now in `localAuthorityTables`, with its columns reconciled by a loop ahead of `createAll`.
2. The challenge tally now counts replies, as Kotlin's does (`PHASE_147_NOTES.md` item 1).
3. The false dartdoc claims on those queries (`PHASE_142_NOTES.md` item 5, `PHASE_147_NOTES.md` item 2) — **two of the three were in this lane's file; the third was already fixed in `voices_repository.dart` by Phase 147.**
4. **`HealthExaminationDao.markUploaded(id, null)` no longer erases a stored `_rev`** — landed first, in its own commit, as **Lane C's blocker**.
5. The `surveys` entry in `_localAuthorityTables` no longer carries the claim its own phase retracted (`PHASE_143_NOTES.md` item 4).

## Three fixes the audits added to the brief

- **`saveLocalResource` never copied the picked file** — so the resource the user created was unopenable, and opening it destroyed its own pointer. Preserving that row would have preserved the claim rather than the resource.
- **`MyLibraryMapper.fromDoc` blanked the `_rev` another writer recorded** — the Phase 146 two-writers-one-column shape, on the column those guards missed. Under preservation it makes a row permanently unprunable.
- **A v49 row repair** — see the comment thread: without it, preservation *relocated* a data loss into a permanent dead end that lies, rather than removing it.

## `schemaVersion` 49 — allocated and **spent**

Not for a column. Preservation needs no DDL change; the bump exists to carry the row repair to devices already on v48, which never re-enter `onUpgrade` otherwise. I reported mid-round that 49 was free — that was wrong, and the correction is in the thread.

## Status

- [x] Draft PR open, `subscribe_pr_activity` called
- [x] Ground-truth `parity-auditor` pass (`effort: max`) — caught a false claim in my own Job 4 write-up
- [x] Job 4 (Lane C's blocker) landed and routed
- [x] Jobs 2, 3, 5
- [x] Job 1 — preservation, plus the two prerequisites and the v49 repair
- [x] Implementation `parity-auditor` pass (`effort: max`) — 12 items fixed, read-only, tree verified clean after
- [x] Gate green: `build_runner`, `dart format`, `flutter analyze`, `flutter test` (2758 tests, exit 0)
- [x] CI green on `2ec99cc`; `mergeable_state: clean`

## Files touched

Owned: `flutter/lib/data/local/app_database.dart`, `tables.dart`, `my_library_mapper.dart`, `flutter/lib/repository/resources_repository.dart`, `flutter/PHASE_150_NOTES.md`, plus new and existing tests for those.

Outside the set, all comment-only and all under the *making an existing statement true* carve-out: `voices_repository.dart` (two sentences this diff itself falsified), `health_repository.dart` (a dartdoc describing the defect rather than the behaviour), `CLAUDE.md` / `.claude/agents/parity-auditor.md` / `docs/kotlin-to-flutter-migration.md` (preserved-table count 27 → 28; `schemaVersion` 48 → 49).

**Not touched:** `health_uploader.dart:86-90` is now false and is Lane C's file — reported, not changed.

`flutter/PHASE_150_NOTES.md` carries the full write-up and a 15-item *Reported, not fixed* list, the largest being that **the port has no resources uploader at all** — the reason this table needed preserving, with a trap in the obvious implementation.

https://claude.ai/code/session_0116HuwVHveftB12neUB6tAC


dogi:
**Job 4 has landed — Lane C is unblocked.** `c5f02b2`, first code commit on this branch.

`HealthExaminationDao.markUploaded(id, rev)` now writes `rev == null ? const Value.absent() : Value(rev)`, so a null rev clears the dirty flag and leaves the stored `_rev` alone. Lane C can align its health handler with Kotlin's `has("id")` gate on top of this.

Three things the integrator will want when routing that alignment:

1. **This is parity, not an improvement on the Kotlin.** I checked which overload the Kotlin actually calls, because it decides how the change should be described: `HealthRepositoryImpl.markHealthExaminationsUploaded` (`:66-69`) passes a `Map<String, String?>` to the `@Transaction` `HealthExaminationDao.markUploaded(idToRevMap)` (`:36-46`), which partitions the null-rev ids into `SET isUpdated = 0 WHERE _id IN (:ids)`. Kotlin's *single-id* overload does still write `_rev = :rev`, but its only caller is the non-null half of that partition, so a null never reaches it.
2. **Blank is deliberately not folded in with null** — Kotlin partitions on `it.value == null`, so an empty-string rev is written. There is a test pinning that, and folding it in reds it.
3. **`health_uploader.dart:87-90` now says something false**, and it is Lane C's file: its comment reads *"The port cannot copy that half: its `markUploaded` writes `rev: Value(rev)`, so passing the null we have would erase the revision the row already holds … and `app_database.dart` is not this lane's file."* All of that was true when Phase 148 wrote it and is now false. It is exactly the comment a lane doing 409-recovery work will read when deciding whether the health handler can adopt the `has("id")` gate, and it tells them it cannot. Reported rather than changed, per one-file-one-owner — but it should not survive the round.

`HealthRepository.markUploadedBatch` delegates to this DAO method, so its copy of the same bug is fixed by this commit. It still has zero callers (`PHASE_142_NOTES.md` item 7), so that remains a dead-code question rather than a defect.

Each defect was demonstrated failing before the fix, and the fix is mutation-tested three ways (`test/data/local/health_examination_rev_test.dart`): restoring `Value(rev)`, folding blank in with null, and leaving the dirty flag set each red a different test.

Still in flight here: the ground-truth `parity-auditor` pass, then Jobs 1, 2, 3 and 5.

---

dogi:
## Correction to my Job 4 comment above, and Job 1 has landed

**First, a correction to what I wrote above, because a lane may act on it.** I said the null-rev branch in the health uploader was live. **It is not.** `health_uploader.dart:70-72` returns `NetworkError(null, …)` when the 2xx body's `rev` `is! String`, so `markUploaded` is only ever reached with a promoted `String`. The hole `c5f02b2` closes is **latent**.

That does not weaken the blocker framing — it sharpens it. What the DAO fix buys is the *precondition* for **removing that early return**, which is the one thing standing between the port and Kotlin's `has("id")` gate (`HealthRepositoryImpl.kt:118-120`): with the erase gone, an `id`-without-`rev` 2xx can clear `isUpdated` and keep `_rev` instead of being classified `indeterminate` and abandoned. The dartdoc on the method now says this; `c5f02b2`'s commit message still carries the wrong version, and it is pushed, so this comment is the correction of record. It was caught by the ground-truth audit pass, not by me — which is the standing rule in `CLAUDE.md` about a supplied replacement sentence being a claim rather than a patch, and I broke it.

**Also for Lane C:** `health_uploader.dart:87-90` now says something false — *"The port cannot copy that half: its `markUploaded` writes `rev: Value(rev)` … and `app_database.dart` is not this lane's file."* That is the comment a lane doing 409-recovery work will read when deciding whether the health handler can adopt the `has("id")` gate, and it tells them it cannot. Lane C's file, so reported not changed.

## Job 1 — `my_library` is preserved (`a5dd03c`), and two fixes had to land first (`ec18d66`)

**`schemaVersion` 49 was allocated and is *not* spent** — it is free for the next lane. Preservation changes migration behaviour, not DDL, so a bump delivers nothing: a device on v47 keeps its rows precisely because it has not run the drop yet, and one already on v48 has the full shape.

Three things the integrator should know:

1. **The preservation change could not go in alone.** The audit found that `saveLocalResource` **never copied the picked file** — Kotlin copies it into `<ext>/ole/<id>/<basename>` and stores the basename (`ResourcesRepositoryImpl.kt:235-256`), and the port stored the raw `FilePicker` path with `resourceOffline = true` over the top. So the resource was unopenable, and opening it destroyed both columns. Preserving that row would have preserved the claim rather than the resource. Same for `MyLibraryMapper.fromDoc` blanking the `_rev` the resources walk recorded from a course-step sub-object — under preservation that row becomes immortal instead of being swept. Both are in `ec18d66`.

2. **Two existing tests asserted the defect** and now assert the honest value, with the reason written in: `resources_repository_test.dart`'s `saveLocalResource creates a row and marks it offline` and `add_resource_screen_test.dart`'s `a valid form writes the row and confirms` both expected `resourceOffline == true` for a request with **no file**.

3. **`migration_test.dart`'s canonical "pure cache" example had to move off `my_library`** onto `tag`. That example has now moved twice — `users` → `my_library` → `tag` — and the moves are the history of the preserved set. The preserved-table count becomes **28** in `CLAUDE.md`, `.claude/agents/parity-auditor.md` and `docs/kotlin-to-flutter-migration.md`, under the "making an existing statement true" carve-out.

**Lane D:** no `download_queue` query was needed from me this round, and nothing was asked for. But there is a `download_queue` item in my *Reported, not fixed* list that touches `resource_downloader.dart` / `background_entrypoint.dart` — a download enqueued for a `couchId`-less resource can never complete, and the only branch that clears one is "the `my_library` row is gone", which the bump used to trigger. My file copy removes the common route to it; the condition that closes it is not in my file set.

Status: fourteen items in *Reported, not fixed* in `flutter/PHASE_150_NOTES.md`, the largest being that **the port has no resources uploader at all** — which is *why* this table needed preserving, and which has a trap in the obvious implementation (`_rev IS NULL` is authorship in Kotlin and is not in the port).

The implementation `parity-auditor` pass is running against the finished code now; I will report the gate result and anything it finds.

---

dogi:
## The implementation audit found a real one, and it inverts a decision I reported here

**I need to retract the "`schemaVersion` 49 is not spent" line from my last comment.** It is spent, and the reason is the finding.

My preservation change, as first written, **relocated a data loss instead of removing it** — the exact failure the brief warned about, in the row class the change exists to protect.

A device that used Add Resource on any earlier build has, for each locally created resource, `resource_offline = 1` and the raw `FilePicker` path in `resource_local_address`, because the pre-fix writer copied no file. The old bump **deleted** that row. Preserving the table **keeps** it — and it is then a dead end that lies: My Library shows the offline pin (the list sorts `resource_offline DESC`), the detail screen offers *View* rather than Download, the viewer finds nothing under `ole/<id>/`, and the Download it falls back to cannot work either because `urlFor` needs a `couchId` a local row has never had. Permanently, on a now-preserved table.

A dead end that lies is worse than the absence it replaced. So v49 carries a row repair, and **that is what spends the number**: the affected population includes devices already on v48, which never re-enter `onUpgrade` unless the version moves. The no-bump and no-repair decisions were coupled and I reported them as independent.

```sql
UPDATE my_library SET resource_offline = 0, resource_local_address = NULL
WHERE _id IS NULL AND resource_local_address LIKE '%/%'
```

**The audit proposed `_rev IS NULL` for the first clause and I did not use it — it has a false positive.** A resource embedded *only* in a course document arrives rev-less (a step sub-object carries none) while `MyLibraryMapper` still writes its `couchId`, so it is downloadable and `markDownloaded` then writes an absolute path onto a rev-less row. That clause would have cleared a genuine download. `_id IS NULL` cannot — only `saveLocalResource` leaves `couchId` unset. Third instance this round of *`_rev IS NULL` is authorship in Kotlin and is not in the port*, and the first where it caught a proposed fix rather than a shipped one.

Also fixed from the same pass: the reconciliation now runs **before** the drop loop (drift wraps no migration in a transaction, so a throwing `ALTER` used to leave the caches dropped and `createAll` unreached); a zero-byte pick no longer claims to be downloaded; `catch (_)` narrowed to `on Exception`; and eight comment/prose corrections — including one where **my own notes claimed a correction that never landed** (`migration_test.dart`'s frozen-DDL dartdoc predicted this exact change and said "these tests red when that happens", which was true, so leaving it in place told readers a guard had fired when it had not).

**Two items other lanes may care about:**
- `voices_repository.dart` — I fixed two sentences *this diff itself falsified* (the dartdoc still called the tally "top-level"; the hand-off paragraph still listed all three defects as unfixed and another lane's). Comment-only, under the *making an existing statement true* carve-out.
- **Lane C:** `health_uploader.dart:86-90` is still false and still yours — reported, not changed.

`CLAUDE.md`'s `schemaVersion` line is updated to 49.

Gate: format clean, analyze clean, full suite running. I'll confirm green before calling the lane done.

---

dogi:
## Lane A complete — green and ready for the integrator

CI green on `2ec99cc` (all five checks success), `mergeable_state: clean`, local gate green (format, analyze, 2758 tests — baseline 2735 plus exactly the 23 added here, so nothing was silently dropped).

**Merge order note:** nothing in this lane blocks on another, but **Lane C is unblocked by `c5f02b2`** and will want to remove the early return in `health_uploader.dart:70-72` on top of it — the comment at `:86-90` still says the port cannot, which is now false and is theirs to fix.

### What to know when merging

1. **`schemaVersion` is 49, and it is spent** — I reported it free mid-round and that was wrong. The bump carries a *row repair*, not a column: preservation alone left every pre-existing locally-created resource as a permanent dead end claiming to hold a file nothing had copied. Devices already on v48 reach that repair only if the version moves.
2. **Two existing tests now assert different values**, because they encoded the defect: `resources_repository_test.dart`'s `saveLocalResource creates a row…` and `add_resource_screen_test.dart`'s `a valid form writes the row and confirms` both expected `resourceOffline == true` for a request with **no file**. Both now assert the honest value with the reason written in.
3. **`migration_test.dart`'s canonical "pure cache" example moved** off `my_library` onto `tag`. That example has now moved twice (`users` → `my_library` → `tag`) — the moves are the history of the preserved set.
4. **One column-inventory guard is new** (`every my_library column is either old or a deliberate decision`). It reds when a column is added to `my_library`, on purpose: the reconciliation loop adds a column but cannot *backfill* one, and that decision needs a person.

### For whoever briefs the next round

`flutter/PHASE_150_NOTES.md` has a 15-item *Reported, not fixed* list. Ranked by data at risk, the top three:

1. **The port has no resources uploader at all** — the reason `my_library` needed preserving. There is a trap in the obvious implementation: `_rev IS NULL` is a sound authorship test in Kotlin and is **not** one in the port, so an uploader keyed on it would POST server-owned resources under the user's credentials. Needs a `needsSync`-style flag, which is now a backfill decision. Also a forward coupling: this phase writes the copied file under the *local* id, so the moment an uploader records a `couchId` the viewer stops finding it.
2. **A stale `resource_offline = 1` has lost its only automatic repair** — the bump used to be it, accidentally. Kotlin re-derives from disk on every attachment-carrying pull; the port never has.
3. **A course-only resource row is immortal** — same accretion class as the `surveys` entry, same argument for an authorship flag.

Two items in that list were **demoted or refuted** by the implementation audit rather than carried forward, which is worth trusting over the earlier drafts: the `download_queue` item (the guard it asks for is already enforced by the only caller that can reach it) and the "no backfill needed" framing (true of columns, and the wrong question).

---

PR #16962, opened 2026-09-09
Phase 152: recover from a 409 instead of merely surviving it

dogi:
**Lane C of the four-lane round.** Phase 152: port `UploadCoordinator`'s 409 recovery arm beyond the single uploader that had it.

## The finding

**Porting `UploadCoordinator.kt:169-204` literally would have lost data in every uploader here but one.** Mutation-tested: forcing Kotlin's adopt-and-report-success behaviour fails **22 tests**.

A 409 never means "your write is already there" — it means the document exists and your bytes were *not* stored. Kotlin never compares the fetched document against what it was sending, so adopting the revision clears the dirty flag with the clinician's examination, the learner's added answers, or the edited post still on the handset.

## The rule shipped

| Case | Behaviour |
|---|---|
| **update** (payload carries `_rev`) | re-send the same content under the fetched revision — the write lands |
| **create** (no `_rev`) | **neither** re-sent nor adopted; the refusal stands, terminal, row stays inspectable |
| create + `adoptExisting` | Kotlin's original semantics — **one** uploader (`adopted_surveys`, whose clone is a pure function of survey + team) |

The create default means Phase 148's health-conflict tests pass unchanged.

**Armed:** 11 uploaders / 15 types — `health`, `teams` (×5), `feedback`, `achievements`, `adopted_surveys`, `submissions`, `voices`, `events`, `team_tasks`, `ratings`, `user`. `adopted_surveys` lost its bespoke arm (which had **no test coverage at all**) to the shared one.

**`course_progress` was armed and reverted** — its pending predicate is `couchId IS NULL`, so a 409 is unreachable and the arm would be inert. That also means an edited progress row never uploads at all; reported, not fixed.

## Layering

The mechanism lives once, beside `OutboxHandler` in `outbox_drainer.dart`; it wraps only the *send*, so each handler's existing `NetworkSuccess` branch runs on a recovered result unchanged — no uploader duplicates its `markUploaded` and none can forget it. Each call site is ~12 lines, mostly comment. The drainer can't own the call: it sees only a `NetworkResult` and cannot run the handler's DAO write.

The trap avoided: **the document id is the payload's, not `outbox.itemId`** — for health's profile blob those differ, and a row-keyed GET would 404 and recover nothing, silently, with every test green.

## Both audit passes

Ground truth before implementing; implementation after. The second found five real defects in my own green code — including a test that **could not fail** (my 18 mutations missed it because un-arming still killed the test's *other* assertions) and a comment that re-seeded a claim `PHASE_148_NOTES.md` was written to retract. Three findings argued and declined with reasons, the largest being `feedback`'s merge loss, which a plain pull already causes one sync later — so the arm changes latency, not outcome.

## Gate

`dart format` clean · `flutter analyze` clean · **2771 tests pass** (2735 before) · **23 mutations, all killed**.

Full reasoning and a 14-item *Reported, not fixed* list in `flutter/PHASE_152_NOTES.md`.

https://claude.ai/code/session_01EmBB2aww4wdGCPfgqwssYc

### DIFF
--- a/flutter/PHASE_152_NOTES.md
+++ b/flutter/PHASE_152_NOTES.md
@@ -202,6 +202,18 @@ examination row's. A row-keyed GET would 404, the arm would return the original
 `ConflictRecovery.documentUrlUnder(sendUrl, payload)` exists so that derivation
 is written once, and returns null rather than guessing.
 
+It percent-encodes the id with `Uri.encodeComponent`, which is the port's
+established convention for a CouchDB document id (`personals_uploader.dart:151`,
+`submit_photos_uploader.dart:131`, `teams_uploader.dart:123`,
+`teams_repository.dart:707`, `voices_uploader.dart:331`, and the arm this
+phase replaced). **One consequence is worth stating because it is new:** for
+`health` the id is a user id like `org.couchdb.user:ada`, so the URL carries
+`%3A` where `user_uploader.dart:80` deliberately leaves the `_users` prefix's
+colon literal and encodes only the name. CouchDB percent-decodes a path
+segment, so both resolve to the same document — but that is a *checked
+assumption*, not something the suite proves, and it is the kind of thing that
+would fail exactly as a green dead arm.
+
 ---
 
 ## Armed, and not
@@ -292,8 +304,11 @@ the shared arm and has four tests.
 
 ## Tests
 
-Gate green: `dart format` clean, `flutter analyze` clean, **2790 tests pass**
-(2735 before).
+Gate green: `dart format` clean, `flutter analyze` clean, **2767 tests pass**
+(2735 before — the 32 added reconcile: 16 in the new
+`conflict_recovery_test.dart`, 4 in the new `adopted_surveys_uploader_test.dart`,
+and one each in eleven existing uploader test files plus
+`health_legacy_conflict_test.dart`). *Counted from the run, not remembered.*
 
 New: `test/repository/conflict_recovery_test.dart` (16) pins the rule itself —
 both arms, the `adoptExisting` opt-in, the three guards, and `documentUrlUnder`.
--- a/flutter/PHASE_152_NOTES.md
+++ b/flutter/PHASE_152_NOTES.md
@@ -165,6 +165,13 @@ tombstone is `{_id, _rev, _deleted: true}` (`teams_provider.dart:295`), so it
 is an update, and re-sending deletes a document whose latest content this
 device has never seen. Accepted deliberately: the user asked for the membership
 to be removed, and a revision bump elsewhere does not revoke that instruction.
+Adopting instead would report the delete as done while the document is still on
+the server *and* hand the row back to `deleteNotIn`. Both halves are pinned —
+a tombstone recovered under a fresh revision, and a tombstone for a document
+another device already deleted, where the read 404s and the refusal stands
+exactly as it did before this arm existed. The second also shows the arm does
+not disturb the handler's `_deleted` early return, which tolerates a response
+carrying no `rev`.
 
 ---
 
@@ -304,11 +311,11 @@ the shared arm and has four tests.
 
 ## Tests
 
-Gate green: `dart format` clean, `flutter analyze` clean, **2767 tests pass**
-(2735 before — the 32 added reconcile: 16 in the new
+Gate green: `dart format` clean, `flutter analyze` clean, **2769 tests pass**
+(2735 before — the 34 added reconcile: 16 in the new
 `conflict_recovery_test.dart`, 4 in the new `adopted_surveys_uploader_test.dart`,
 and one each in eleven existing uploader test files plus
-`health_legacy_conflict_test.dart`). *Counted from the run, not remembered.*
+`health_legacy_conflict_test.dart`, plus two for the tombstone paths). *Counted from the run, not remembered.*
 
 New: `test/repository/conflict_recovery_test.dart` (16) pins the rule itself —
 both arms, the `adoptExisting` opt-in, the three guards, and `documentUrlUnder`.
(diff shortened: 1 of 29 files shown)


dogi:
**Status: implementation complete, second `parity-auditor` pass running. Not yet ready to merge.**

## The finding that reframed the phase

Porting `UploadCoordinator.kt:169-204` literally — GET the document, adopt its `_rev`, report success — **would have lost data in every uploader here but one**. Mutation-tested: forcing that behaviour fails **22 tests**.

A 409 does not mean "your write is already there". It means a document with this `_id` exists and the `_rev` you sent (or omitted) is not current, so *your bytes were not stored*. Kotlin never compares the fetched document with the one it was sending. Adopting therefore clears the dirty flag with the clinician's examination, the learner's added answers, or the edited post still on the handset.

**The rule shipped instead:**

| Case | Behaviour |
|---|---|
| **update** (payload carries `_rev`) | re-send the same content under the fetched revision — the write lands |
| **create** (no `_rev`) | **neither** re-sent nor adopted; the refusal stands, terminal, row stays inspectable |
| create + `adoptExisting` | Kotlin's original semantics — **one** uploader (`adopted_surveys`, whose clone is a pure function of survey+team) |

The create default means `health_legacy_conflict_test.dart` — Phase 148's pins — passes unchanged.

## Phase 148's policy is extended, not weakened

The re-send *is* a changed request (a `_rev` fetched between the two attempts), which is exactly what the policy says re-arms an item. Bounded at one GET + one re-send; skipped entirely when the fetched revision is the one already sent; structurally impossible for appends (`documentUrlUnder` returns null with no `_id`).

**One place the arm had to defend the policy rather than extend it:** `OutboxDrainer._send` treats a throw as *transient*, so an unguarded recovery whose GET threw would relabel a terminal rejection as retryable and re-offer the row every sweep — Phase 148's accretion re-entering through the recovery arm. Now caught. Found because `health_legacy_conflict_test`'s fake had no GET stub and five Phase 148 tests flipped `abandoned` → `pending`.

## Armed / not armed

**Armed (11 types, 10 uploaders):** `health`, `teams` (×5), `feedback`, `achievements`, `adopted_surveys`, `submissions`, `voices`, `events`, `team_tasks`, `ratings`, `user`.

**`course_progress` was armed and then reverted** — a finding, not a preference. `CourseProgressDao.getPendingUploads` is `couchId IS NULL`, so an acknowledged row is never re-offered: a 409 is unreachable and the arm would be inert. That also means **a later edit to an acknowledged progress row never uploads at all**, which is reported, not fixed. (The ground-truth audit called this reachable; it read the serializer, not the predicate.)

The eight append uploaders are deliberately unarmed.

## For Lane A

`HealthExaminationDao.markUploaded(id, null)` is **not** blocking after all — the health arm never passes a null rev (it has a `String` from the GET or does not recover). Reported in the notes, untouched here. Two stale Phase 148 citations corrected while reading: the DAO is at `app_database.dart:3962-3968`, and the `Value.absent()` pattern is at `:1208` / `:4765`, not `:1198`/`:2066`.

## Gate

`dart format` and `flutter analyze` clean; suite green (~2790, was 2735). 18 mutations run, all killed. A final clean gate run follows the audit; I'll comment again when it's ready to merge.

Full reasoning in `flutter/PHASE_152_NOTES.md`.

---

dogi:
**Ready to merge.** Both `parity-auditor` passes done, gate green, everything pushed.

## The finding, in one line

**Porting `UploadCoordinator.kt:169-204` literally would have lost data in every uploader here but one.** Mutation-tested: forcing Kotlin's adopt-and-report-success behaviour fails **22 tests** in this port.

A 409 never means "your write is already there" — it means the document exists and your bytes were *not* stored. Kotlin never compares the fetched document with what it was sending. So the arm shipped is split:

| Case | Behaviour |
|---|---|
| **update** (payload carries `_rev`) | re-send the same content under the fetched revision — the write lands |
| **create** (no `_rev`) | **neither** re-sent nor adopted; refusal stands, terminal, row stays inspectable |
| create + `adoptExisting` | Kotlin's original semantics — **one** uploader (`adopted_surveys`, clone content is a pure function of survey+team) |

Armed: 11 uploaders / 15 types. `course_progress` was armed and **reverted** — its pending predicate is `couchId IS NULL`, so a 409 is unreachable and the arm would be inert (which also means an edited progress row never uploads at all; reported).

## Phase 148's policy extended, not weakened

The re-send *is* the "changed request" the policy names. Bounded at one GET + one re-send; skipped when the fetched revision is the one already sent. One place the arm had to **defend** the policy: a throwing fetch would have relabelled a terminal rejection as transient and re-offered the row every sweep — Phase 148's accretion re-entering through the recovery arm. Guarded.

## Second audit pass — five real defects in my own green code

Worth the integrator's attention:

- **A test that could not fail**, and my own 18 mutations missed it: un-arming the uploader still killed the *other* assertions in the same test. **Being killed by a mutation ≠ every assertion can fail.**
- **The health rationale re-seeded a claim `PHASE_148_NOTES.md` was written to retract.** Fifth round running where a note propagated an inherited error — this time one the project had already corrected in writing.
- **A health fixture used a shape no production caller produces**, and it was the only reason the "derive the URL from `itemId`" mutation died. Rewritten on the reachable path; the mutation still dies.
- Wasted round-trip (fetch issued before the create/update decision), plus two wrong counts and one misread Kotlin citation of mine, all corrected.

**Three findings I argued and declined, with reasons in the notes** — including the biggest: the auditor read `feedback`'s merge loss as destruction this arm introduces. It isn't. A plain **pull** already drops the admin reply locally and leaves the row swept with a changed revision, so the memo re-arms and the same loss lands one sync later with no 409 involved. Both halves are now pinned by tests. The arm changes the *latency*, not the outcome — so it's reported as a merge defect older than this phase. My original `cacheDocuments` justification was genuinely wrong (health-only) and is corrected.

## One process trap worth propagating

The mutation harness reverts with `git checkout -- lib/`, which **silently reverted three uncommitted fixes**. It only surfaced because the gate went red on an assertion I'd just made stricter — otherwise three corrections would have not-shipped under a green suite and a commit message describing them. Same shape as Phase 135's audit-leaves-reverts-in-the-tree. **Commit before mutating**, now written into the notes.

## For Lane A

`HealthExaminationDao.markUploaded(id, null)` turned out **not** to block — the health arm never passes a null rev. Untouched, reported. Stale Phase 148 citations corrected: the DAO is `app_database.dart:3962-3968`, the `Value.absent()` pattern `:1207`/`:4765`. One dangling `_adoptExistingDocument` reference remains at `app_database.dart:390` — documentation only, left because that file is yours this round.

## Gate

`dart format` clean · `flutter analyze` clean · **2771 tests pass** (2735 before, 36 added) · **23 mutations, all killed** · CI green.

Full reasoning and a 14-item *Reported, not fixed* list in `flutter/PHASE_152_NOTES.md`.

---

PR #16965, opened 2026-09-09
all: smoother server reachability providing (fixes #16964)

### LINKED ISSUE #16964: mainapplication-god-class with the server-reachability
server-reachability probing all live in one 537-line class with a 21-parameter provider downstream of it.

Okuro3499:
fixes #16964
Move server reachability logic into a dedicated checker service and replace the inline MainApplication code. This keeps the app entry point smaller, centralizes cache and fallback behavior, and adds unit coverage for cached success, fallback URL handling, and primary-server checks.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved server connectivity checks by trying an alternative server address when the primary address is unavailable.
  - Added more reliable connection handling with fallback request methods.

- **Performance**
  - Reduced repeated network requests by briefly caching server reachability results.

- **Reliability**
  - Primary-server checks can now bypass cached results when a current status is required.


dogi:
**Review — rating: 88/100**

Clean extraction. `MainApplication` is the largest file in the app and its companion object had grown ~90 lines of HTTP reachability logic — cache, URL mapping, HEAD-then-GET fallback, `TrafficStats` tagging — sitting next to WorkManager scheduling and theme handling. Pulling that into `ServerReachabilityChecker` is exactly the right cut.

I diffed the moved logic against the original line by line: blank-string guard, 30s TTL cache, `serverUrlMapper.processUrl` then primary-before-alternative, `http://` prefixing when the scheme is absent, HEAD with a GET retry on 405/501, 5s connect and read timeouts, and the `TrafficStats.setThreadStatsTag` / `clearThreadStatsTag` pairing in a `try/finally`. All identical. The `kotlinx.coroutines.CancellationException` rethrow survives the move from a fully-qualified check to an import — same class, and it still precedes the generic `catch`, which is the part that matters.

The `reachabilityCache` also stays a single process-wide instance: it was a `private val` on the companion, and it's now instance state on a companion-held `by lazy`, so lifetime and sharing are unchanged.

**The `() -> CoreDependenciesEntryPoint` supplier is the right seam.** Passing the entry point as a lambda rather than an instance defers Hilt resolution past class initialisation — which matters here, since `coreDependenciesEntryPoint` needs the application context — and it's what makes the 141 lines of new tests possible at all. This logic was previously untestable: a companion-object function reaching into a global entry point, with no way to substitute the mapper or the dispatcher.

One follow-up worth naming rather than blocking on: this is a manual singleton living alongside the DI container. `services/` classes are otherwise provided through `ServiceModule`, and a `@Singleton class ServerReachabilityChecker @Inject constructor(serverUrlMapper: ServerUrlMapper)` would be the natural end state. The obstacle is the call sites — `MainApplication.isServerReachable(...)` is invoked statically from all over — so converting properly means touching those too. Extracting first and injecting second is a reasonable order; it's just worth being explicit that the global didn't go away, it moved.

CI green 5/5.

**Verdict:** mergeable.

---

PR #16970, opened 2026-09-09
resources: smoother sort filter view modelling (fixes #16969)

### LINKED ISSUE #16969: redesign resource filter
https://github.com/user-attachments/assets/cc512330-b9ce-4472-b638-b92eab27630f

Okuro3499:
fixes #16969
[Screen_recording_20260909_172333.webm](https://github.com/user-attachments/assets/ac1e3b1c-459b-4d0d-b307-bed67759fc41)

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 7054
-        versionName = "0.70.54"
+        versionCode = 7055
+        versionName = "0.70.55"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/callback/OnFilterListener.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/callback/OnFilterListener.kt
@@ -1,7 +1,7 @@
 package org.ole.planet.myplanet.callback
 
 interface OnFilterListener {
-    fun filter(subjects: MutableSet<String>, languages: MutableSet<String>, mediums: MutableSet<String>, levels: MutableSet<String>)
+    fun filter(subjects: MutableSet<String>, languages: MutableSet<String>, mediums: MutableSet<String>, levels: MutableSet<String>): Int
 
     suspend fun getData(): Map<String, Set<String>>
 
--- a/app/src/main/java/org/ole/planet/myplanet/callback/OnFilterListener.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/callback/OnFilterListener.kt
@@ -6,4 +6,8 @@ interface OnFilterListener {
     suspend fun getData(): Map<String, Set<String>>
 
     fun getSelectedFilter(): Map<String, Set<String>>
+
+    fun getFilteredCount(subjects: Set<String>, languages: Set<String>, mediums: Set<String>, levels: Set<String>): Int
+
+    fun clearAllFilters()
 }
--- a/app/src/main/java/org/ole/planet/myplanet/ui/resources/ResourcesFilterFragment.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/resources/ResourcesFilterFragment.kt
@@ -135,8 +135,12 @@ class ResourcesFilterFragment : BottomSheetDialogFragment(), AdapterView.OnItemC
     }
 
     private fun updateResultCount() {
-        if (_binding == null) return
         val count = filterListener?.getFilteredCount(selectedSubs, selectedLang, selectedMeds, selectedLvls) ?: 0
+        showResultCount(count)
+    }
+
+    private fun showResultCount(count: Int) {
+        if (_binding == null) return
         binding.btnConfirmFilters.text = getString(R.string.show_n_results, count)
     }
 
@@ -168,8 +172,8 @@ class ResourcesFilterFragment : BottomSheetDialogFragment(), AdapterView.OnItemC
                 R.id.list_level -> addToList(s, selectedLvls)
                 R.id.list_medium -> addToList(s, selectedMeds)
             }
-            filterListener?.filter(selectedSubs, selectedLang, selectedMeds, selectedLvls)
-            updateResultCount()
+            val count = filterListener?.filter(selectedSubs, selectedLang, selectedMeds, selectedLvls) ?: 0
+            showResultCount(count)
         }
     }
 
--- a/app/src/main/java/org/ole/planet/myplanet/ui/resources/ResourcesFilterFragment.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/resources/ResourcesFilterFragment.kt
@@ -3,6 +3,7 @@ package org.ole.planet.myplanet.ui.resources
 import android.animation.Animator
 import android.animation.AnimatorListenerAdapter
 import android.animation.ValueAnimator
+import android.app.Dialog
 import android.os.Bundle
 import android.view.LayoutInflater
 import android.view.View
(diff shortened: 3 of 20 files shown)


dogi:
**Review — rating: 78/100**

The direction is good — replacing the inline `cardFilter` panel with two proper Material bottom sheets, adding a filter-count badge and a live "Show N results" button, is a real improvement over a `View.VISIBLE`/`GONE` card with duplicate listeners attached to it. The localisation discipline is also right: `show_n_results` lands in `values/` and all five locale files with real translations (`عرض %1$d نتيجة`, `Mostrar %1$d resultados`, `Afficher %1$d résultats`, `%1$d नतिजा देखाउनुहोस्`, `Tus %1$d natiijo`), not just English. And extending `OnFilterListener` is contained — `ResourcesFragment` is its only implementer.

Two things I'd want changed before this lands.

**1. The live result count re-runs the entire filter pipeline on the main thread, on every checkbox tap — and it's the second full pass of that tap.**

```kotlin
override fun getFilteredCount(subjects, languages, mediums, levels): Int {
    val searchQuery = etSearch.text?.toString()?.trim().orEmpty()
    return applyFilterModels(
        filterLocalLibraryByTag(allResourceModels, searchQuery, searchTags),
        subjects, levels, languages, mediums
    ).size
}
```

`filterLocalLibraryByTag` runs `ResourcesSearchUtils.searchLocalModels` over the whole library — which normalises text per item — then tag-filters, then `applyFilterModels` filters again across five facets. `onItemClick` already calls `filterListener?.filter(...)`, which kicks off `applyFiltersAndUpdateUI` and does all of that once; `updateResultCount()` then does it a second time purely to read `.size`. So every tick of a checkbox is two full passes over the library, synchronously, while the sheet is open. On the large offline libraries this app is built for, that's a visible hitch per tap.

The filtered list is already being computed for the adapter — plumb its size back rather than recomputing. A callback carrying the count from the pass that just ran, or having `applyFiltersAndUpdateUI` publish the size the sheet can observe, gets the same UI for one pass instead of two.

**2. Sort direction lost its affordance.**

`toggleSortOrder` and `toggleTitleSortOrder` *flip* direction each call — that was legible when they were buttons labelled "order by date" / "order by title" and you tapped one repeatedly. They're now radio buttons in a sheet:

```kotlin
binding.sortOrderByDate.setOnClickListener {
    listener?.onSortSelected(ResourcesViewModel.SortMode.DATE)
    dismiss()
}
```

Tapping "Date" when Date is already selected silently reverses the order and closes the sheet, with nothing in the radio indicating which way it now sorts. A radio group expresses *which field*, not *which direction* — the two don't compose. Either show the current direction per option (an arrow that flips), or make it four options (Date ↑, Date ↓, Title ↑, Title ↓) so the state is visible and selecting is idempotent.

Smaller notes:

- `updateFilterBadge` counts `searchTags + subjects + languages + mediums + levels` but not `selectedDownloadFilterIndex`, while `clearAllFilters` resets it. A user with only the download filter active sees no badge but still has a filter on. The deleted `hasActiveFilters()` did include it.
- `ResourcesFilterFragment`'s `onStart` sizing is gone — it used to give landscape a 55%-width, 85%-height, END-gravity dialog and portrait 90% width. A `BottomSheetDialog` is full-width by design, so on tablets and landscape the filter goes from a side-anchored panel to a full-width sheet. The new `fragment_library_filter.xml` wraps its four ListViews in a `NestedScrollView` so nothing clips, and there's no `layout-land`/`layout-sw600dp` variant to suggest otherwise — so I read this as intended rather than an oversight. Worth confirming, given this PR also rewrites all three `fragment_my_library.xml` variants, so those form factors are clearly in scope.

CI green 5/5.

**Verdict:** needs another round — stop recomputing the full filter pass for the count, and give sort direction a visible state.

---

dogi:
**Re-review — rating: 74/100** (was 78)

The double-pass finding is fixed, and fixed the right way — but the fix introduced a crash, and the second finding from last round is untouched.

**Fixed properly.** `filter()` now returns the count from the pass it already performs, `applyFiltersAndUpdateUI` returns `filteredList.size` and caches it in `lastFilteredCount`, and `onItemClick` uses that return value instead of calling `getFilteredCount` separately. One pass per tap instead of two, and the memoised early-return correctly hands back `lastFilteredCount` rather than recomputing. That's exactly the plumbing I was asking for.

**The new problem is the null-guard.**

```kotlin
override fun filter(...): Int {
    ...
    if (view == null) return getFilteredCount(subjects, languages, mediums, levels)
    ...
}

override fun getFilteredCount(...): Int {
    val searchQuery = etSearch.text?.toString()?.trim().orEmpty()
    ...
}
```

`etSearch` is `binding.layoutSearch.etSearch`, `binding` is `_binding!!`, and `_binding` is assigned `view?.let { FragmentMyLibraryBinding.bind(it) }` — so it is null in exactly the case the guard tests for. The `view == null` branch dereferences `_binding!!` and throws `KotlinNullPointerException`. Before this commit that case did nothing; now it's the one that crashes.

Note that `updateFilterBadge()`, two lines above, opens with `if (_binding == null) return` — the pattern is already in the file, it just wasn't carried into the new branch. `return lastFilteredCount` is the natural fix, and it matches what `applyFiltersAndUpdateUI` already returns for the same condition.

**Still outstanding from last round: sort direction has no visible state.** `ResourcesSortFragment` is unchanged. Tapping "Date" when Date is already selected still calls `toggleSortOrder`, silently reverses the order and dismisses the sheet, with the radio showing only which field is active and nothing about direction. Radios express *which field*, not *which direction* — either give each option a direction indicator that flips, or make it four options.

One consequence of moving off the coroutine worth being aware of: `searchJob?.cancel()` still runs but nothing is assigned to `searchJob` on this path any more, so rapid successive taps no longer coalesce the way `launch`-then-cancel did — each tap runs the full pass inline. At checkbox-tapping speed that's fine, and one synchronous pass beats two dispatched ones, so I'd leave it; just don't expect the cancel to be doing anything here.

Also still open from last round, minor: `updateFilterBadge` counts tags and the four facets but not `selectedDownloadFilterIndex`, which `clearAllFilters` does reset — so a user with only the download filter active sees no badge.

CI green 5/5.

**Verdict:** needs another round — the `view == null` guard crashes; return `lastFilteredCount` there, and give sort direction a visible state.

---

dogi:
**Re-review — rating: 84/100** (was 74)

All three outstanding points are fixed.

- **The NPE is gone.** `if (view == null) return lastFilteredCount` — the cached value rather than a call into `getFilteredCount`, which would have dereferenced `_binding!!` in exactly the state the guard tests for.
- **Sort direction is visible.** `ResourcesViewModel` exposes `isDateSortAscending` / `isTitleSortAscending`, the sheet takes them via `setCurrentDirection`, and the labels get `↑`/`↓`. I traced the semantics: `toggleSortOrder` flips the flag *before* sorting, so the flag reflects the order currently applied — the arrow shows current state, not the pending one. Correct.
- **The badge counts the download filter**, and the chip's click handler now calls `updateFilterBadge()` so it updates when that filter changes rather than only on the next tag/facet edit.

**⚠️ The branch no longer merges into master.** `git merge-tree origin/master` reports a content conflict in `app/src/main/res/values/colors.xml`: this PR appends an `ink_1` block at EOF, and master has since appended a `profile_*` block at the same place. Pure append-collision, resolvable by keeping both, but the branch is 11 commits behind and needs `git merge origin/master` before it can land. Worth doing before it reaches the queue rather than letting the drain discover it and hand it a `conflict` label.

Three things I'd still change, none of them blocking:

- **`ink_1` is the odd name out.** Every other colour in that file is semantic — `card_border`, `chip_selected_bg`, `list_status_completed_bg`. `ink_1` is a design-token name, and it doesn't even hold up: in `values/` it's `#1A1814`, in `values-night/` it's `@color/card_bg`. It isn't "ink" in dark mode, it's the card background. Something like `sort_filter_capsule_bg` would say what it is in both themes.
- **The arrow is concatenated onto a localised string.** `"${getString(R.string.order_by_date)} $dateArrow"` places the arrow after the text by LTR assumption; `values-ar` exists, so in Arabic that lands wrong. A `drawableEnd` with a vector, or putting the glyph inside the string resource, both handle RTL for free.
- **The arrow sits on the control that changes it.** It accurately reports current state, but tapping "Date ↑" gives you descending — the label reads like a target and behaves like a status. That's the residue of expressing two independent things (which field, which direction) through one radio group. Four options, or a separate direction toggle beside the field selection, would make the interaction predictable. The visibility problem I raised is solved; this is the design question underneath it, and it's fine as follow-up.

CI green 5/5.

**Verdict:** mergeable once master is merged in.

---

PR #16971, opened 2026-09-09
teams: smoother repository members mapping (fixes #16968)

### LINKED ISSUE #16968: Membership-map building duplicated verbatim
getJoinedMembers() and getRequestedMembers() both rebuild a HashMap keyed by id and _id before mapping back. Extract a private mapUsersByAnyId().

Okuro3499:
fixes #16968
Remove the unused TimeProvider dependency and simplify TeamsRepositoryImpl now that team records are treated as having valid _id values. This drops fallback-to-id logic, reuses a shared user lookup helper for id/_id mapping, and updates tests to match the leaner constructor.


dogi:
**Review — rating: 88/100**

The whole change rests on one fact, so I checked it before anything else:

```kotlin
// model/MyTeam.kt
@PrimaryKey @JvmField var _id: String = "",
```

`_id` is a **non-null** `String` primary key. Every `it._id ?: it.id`, `_id ?: ""`, `mapNotNull { it._id }` and `?: continue` in this file was therefore operating on a value that can never be null — the elvis branches were unreachable. Removing them is behaviour-preserving, and it deletes code that read like a meaningful fallback but wasn't. Same story for `member.id?.let { … } ?: 0`: `UserEntity.id` is `@PrimaryKey @JvmField var id: String = ""`, also non-null.

I went through the substitutions individually rather than trusting the pattern, because a couple would have been behaviour changes if `_id` *had* been nullable:

- `for (team in transactions) { val id = team._id ?: continue; balance += … }` → dropping `?: continue` would have started including previously-skipped rows in a **balance** calculation. With `_id` non-null the `continue` never fired, so the balance is unchanged.
- `filter { !it._id.isNullOrBlank() }` → `filter { it._id.isNotBlank() }`: on a non-null String these are the same predicate.
- `mapNotNull { it.toSummary() }` → `map { it.toSummary() }` with the return type tightened from `TeamSummary?` to `TeamSummary`: `toSummary` returned null only on a null `_id`, so nothing was ever dropped.

Two genuine wins beyond the cleanup:

- **`timeProvider` was injected and never used.** Removing it from the constructor takes a dead dependency out of the largest repository in the codebase, and the three test files are updated to match.
- **`mapUsersByAnyId` extraction.** The identical eight lines — build a `HashMap` keyed by both `id` and `_id`, then map the input ids through it — appeared verbatim in `getTeamMembers` and the requested-members function. Now one implementation.

`updateTeamEntityById` losing its `val model = entity` alias is a small thing but the alias was pure noise.

One thing worth knowing, not a defect in this PR: `_id` is non-null but **defaults to `""`**, so a locally-created, not-yet-synced team carries an empty id. That means `associateBy { it._id }` and `distinctBy { it._id }` would collapse all such rows onto one key. I traced the three sites and none is reachable with empty ids in practice — `bulkInsert` matches against server-supplied `_id`s, and the `getByIds` paths only return rows whose ids were asked for — so nothing breaks. But the `?: it.id` this PR removes was probably *written* to handle exactly that case and silently never did, so if an unsynced-row fallback is actually wanted somewhere, `_id.ifEmpty { id }` is the form that would work.

CI green 5/5.

**Verdict:** mergeable.

---

dogi (approved):

PR #16972, opened 2026-09-09
sync: smoother bulk docs uploading (fixes #16966)

### LINKED ISSUE #16966: bulk-upload pipelines
```
Two ~80-line bulk-upload pipelines (uploadTeams/uploadNews) are hand-copied instead of sharing one path.
```

Okuro3499:
fixes #16966
Introduces a new `BulkDocUploader` utility to centralize `_bulk_docs` request building, response parsing, and failure handling. `UploadManager` (news uploads) and `TeamsUploader` now use this shared flow instead of duplicating bulk upload code, while preserving their item-specific retry/update behavior. Adds `BulkDocUploaderTest` coverage for empty batches, success/reject outcomes, non-2xx failures, thrown exceptions, and partial bulk responses.

## Summary by CodeRabbit

* **Improvements**
  * Bulk uploads for news and teams now process each item individually, allowing successful records to be saved while rejected items or request failures are queued for retry.
  * Upload handling now covers partial responses, unsuccessful requests, and request exceptions more consistently.
* **Tests**
  * Added coverage for empty uploads, successful and rejected items, partial responses, failed requests, and thrown exceptions.


dogi:
**Review — rating: 89/100**

Good extraction of a real duplication. `UploadManager.uploadNews` and `TeamsUploader` were both building a `{docs: [...]}` payload, POSTing to `_bulk_docs`, walking the response by index, branching on a per-element `error` key, and then handling whole-request failure and thrown exceptions separately. `BulkDocUploader` with a sealed `Outcome` collapses that into one implementation, and the three-case `Accepted` / `Rejected` / `RequestFailed` split is exactly the right shape for what CouchDB's bulk endpoint can return.

I traced each call site against its original to confirm behaviour is preserved, because the two copies had diverged in small ways and a unification is where those get silently normalised:

- **News `Rejected`** still passes a hardcoded `null` httpCode with a wrapped `Exception("Bulk upload error: …")`, matching the original; **Teams `Rejected`** still passes `response.code()` and no exception, per the "200 bulk response code prevents retry here" comment. The `Outcome.Rejected` carrying `httpCode` that news ignores is deliberate asymmetry, not an oversight.
- **News `RequestFailed`** merges two previously separate branches — unsuccessful response (`code`, no exception) and thrown exception (`null`, `e`). Both survive as `RequestFailed(response.code(), null)` and `RequestFailed(null, e)`. I checked `queueNewsRetry`'s signature: `exception: Exception? = null` and `retryable = exception != null || (httpCode != null && httpCode >= 500)`, so passing an explicit null is identical to omitting the argument. Same for `queueTeamRetry`. No retry decision changes.
- **Teams `RequestFailed`** still keys on `teamData.teamId` while `Rejected` keys on the response's `id`, matching the original.

**The unification also fixes a latent crash in the news path.** `TeamsUploader` had `batch.getOrNull(i) ?: continue` and a warning when the response is shorter than the batch; the news path used a bare `processedNews[i]`. A `_bulk_docs` response with more entries than documents sent would have thrown `IndexOutOfBoundsException` mid-batch. Both paths now go through `getOrNull`, and the short-response warning applies to news too. That's a real bug removed as a side effect, and the test `upload only reports outcomes for items present in the response` pins it.

The six tests map cleanly onto the six branches — empty input, payload shape, mixed Accepted/Rejected, non-2xx, thrown exception, short response — for code that previously had none in either copy.

Two small things, neither blocking:

- `BulkDocUploader` is an `object` taking `uploadRepository` as a parameter, while its neighbours in `services/upload/` (`PhotoUploader`, `AchievementUploader`, `TeamsUploader`) are all `@Inject constructor` classes. It's genuinely stateless so the `object` works and tests easily, but it's a deviation from the pattern in that package.
- The news call site's `T` ends up as `Pair<NewsUploadData, JsonArray>`, so the invocation reads `processedNews.map { (news, imagesArray) -> (news to imagesArray) to news.newsJson }` and the callback destructures a pair inside a pair. A tiny `data class NewsBulkItem(val news: …, val images: …)` would make that line legible.

CI green 5/5.

**Verdict:** mergeable.

---

dogi (approved):

PR #16974, opened 2026-09-09
teams: less repository require room entity is more (fixes #16967)

### LINKED ISSUE #16967: A no-op extension survives from the Realm migration
MyTeam.requireRoomEntity() just returns this, called 14 times. Delete it and the call sites.

Okuro3499:
fixes #16967
Simplify `TeamsRepositoryImpl` by removing the no-op `requireRoomEntity()` helper and writing `MyTeam` objects directly to `teamDao` upsert calls. This reduces unnecessary indirection and keeps repository persistence logic clearer without changing behavior.

## Summary by CodeRabbit

- **Reliability**
  - Improved the consistency of team data handling during creation, updates, membership changes, resource linking, and synchronization.
  - Existing team workflows and behavior remain unchanged for end users.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 7055
-        versionName = "0.70.55"
+        versionCode = 7056
+        versionName = "0.70.56"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/repository/TeamsRepositoryImpl.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/TeamsRepositoryImpl.kt
@@ -136,7 +136,7 @@ class TeamsRepositoryImpl @Inject constructor(
                 entity.apply {
                     _rev = uploadedTeams[_id]
                     updated = false
-                }.requireRoomEntity()
+                }
             }
         if (teamsToUpdate.isNotEmpty()) {
             teamDao.upsertAll(teamsToUpdate)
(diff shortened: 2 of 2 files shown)


J-S-webskas (approved):

dogi:
**Review — rating: 88/100**

```kotlin
private fun MyTeam.requireRoomEntity(): MyTeam {
    return this
}
```

An extension function whose entire body is `return this`, called at fourteen sites. It does nothing, has always done nothing, and the name actively misleads — `requireRoomEntity()` reads like a conversion or a validating cast, so every one of those call sites looked like it was doing something load-bearing. Almost certainly a leftover from the Realm→Room migration where it once converted between representations. Deleting it and the fourteen calls is behaviour-preserving by construction.

The one place worth a second look is `insertMyTeam`, where the result was being stored as well as upserted:

```kotlin
- val entity = model.requireRoomEntity()
- teamDao.upsert(entity)
- existingTeams?.put(teamId, entity)
+ teamDao.upsert(model)
+ existingTeams?.put(teamId, model)
```

`entity` and `model` were the same instance, so the map still caches the object that was just written. No change.

**⚠️ This conflicts with #16971, which is currently `ready` and sits in the same file.** `git merge-tree` reports a content conflict in `TeamsRepositoryImpl.kt` between the two. They overlap directly — #16971 rewrites `markTeamsUploaded`'s filter and map, and this PR changes the `.requireRoomEntity()` at the end of that same `map` block; both also touch `updateTeamEntityById` and `respondToMemberRequest`. Neither conflicts with master on its own, so whichever lands first is clean and the second needs a rebase.

Worth sequencing deliberately rather than letting the drain discover it: #16971 is the larger of the two and removes an unused `timeProvider` constructor parameter that this PR doesn't touch, so landing that first and rebasing this one on top is the cheaper order. The resolution is mechanical either way — this PR only ever deletes a suffix.

CI green 5/5.

**Verdict:** mergeable, but sequence it against #16971.

---

PR #16978, opened 2026-09-09
sync: smoother table insert transaction handling (fixes #16977)

### LINKED ISSUE #16977: A 70-line string-dispatch duplicates a pattern the upload side already solved
A when(table) block dispatches to 16 repository calls by string, structurally identical to UploadConfigs. Refactor into a Map<String, suspend (JsonArray) -> Unit> table-sync config.

Okuro3499:
fixes #16977
Replaces the large `when (table)` block in `TransactionSyncManager` with a `tableSyncHandlers` map that routes each synced table to its insert function. This centralizes table-to-handler wiring, keeps timing/logging through `timedBatchInsert`, preserves special courses perf logging, and standardizes unknown-table handling.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 7044
-        versionName = "0.70.44"
+        versionCode = 7045
+        versionName = "0.70.45"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/services/sync/TransactionSyncManager.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/services/sync/TransactionSyncManager.kt
@@ -84,6 +84,33 @@ class TransactionSyncManager @Inject constructor(
     // same inserts that take ~1ms/doc uncontended balloon to >100ms/doc under contention. This
     // mutex serializes only the DB-write portion of each batch; network fetches still overlap.
     private val dbWriteMutex = Mutex()
+    
+    private val tableSyncHandlers: Map<String, suspend (JsonArray) -> Unit> = mapOf(
+        "news" to { arr -> voicesRepository.insertNewsList(extractDocs(arr)) },
+        "feedback" to { arr -> feedbackRepository.insertFeedbackList(extractDocs(arr)) },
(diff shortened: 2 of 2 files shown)


dogi:
**Review — rating: 87/100**

Nineteen `when` arms, each wrapping its body in an identical `timedBatchInsert(table, arr.size()) { … }`, collapse to one dispatch through a handler map. 70 lines out, 32 in, and the boilerplate that had to be repeated per table is now written once at the call site.

**I checked the one thing that looked like a regression, and it isn't.** The old code had a two-tier shape: most tables went through `timedBatchInsert`, while `exams`, `submissions` and `teams` fell into an `else ->` branch that explicitly took `dbWriteMutex.withLock { … }` and then called `syncTimeLogger.logDbOperation("insert_batch", …)` by hand. Flattening all nineteen onto `timedBatchInsert` looks at first like it drops that lock — which would matter, since the comment above the mutex records a measured *~1ms/doc uncontended vs >100ms/doc under contention*.

It doesn't, because `timedBatchInsert` already is that code:

```kotlin
private suspend fun timedBatchInsert(table: String, batchSize: Int, insert: suspend () -> Unit) {
    val insertStartTime = SystemClock.elapsedRealtime()
    dbWriteMutex.withLock { insert() }
    val insertDuration = SystemClock.elapsedRealtime() - insertStartTime
    syncTimeLogger.logDbOperation("insert_batch", table, insertDuration, batchSize)
}
```

Same mutex, same metric name, same arguments. The `else ->` branch was a hand-inlined duplicate of the helper sitting forty lines below it, and this removes the duplicate rather than the protection. `dbWriteMutex` now has exactly one usage site, inside the helper, which is where it belongs.

One incidental improvement: an unrecognised table used to fall through the `else` branch and still emit an `insert_batch` metric with a duration for work that never happened. Now it logs the error and records nothing.

The `Lazy` handling survives — `teamsSyncRepository.get()` sits inside the lambda bodies, so building the map at construction doesn't force those dependencies to resolve.

Two trivia: the `courses` handler hardcodes `"    courses insertDuration"` where it used to interpolate `$table` — same string, since that arm only ever runs for `courses`, but the interpolation was self-documenting. And there's trailing whitespace on the blank line the map is inserted after.

CI green 5/5, and this is the only PR of the four currently in review that's both up to date with master and conflict-free against everything else in the queue.

**Verdict:** mergeable.

---

dogi (approved):

PR #16983, opened 2026-09-09
fix(flutter): harden the feedback area against what the server actually sends

dogi:
Lane 4 of a five-lane round. Four commits; `dart format`, `flutter analyze` and `flutter test` all clean, **2948 passing** (baseline 2923 + 25 new).

## What changed

**`feedback_mapper.dart` no longer throws on a document shape the server actually sends.** Sixteen raw `as String?` casts across fifteen lines are gone. A cast throws a `TypeError` on anything that is neither a string nor null, and Kotlin cannot: `JsonUtils.getString` returns `""` for a non-string (`JsonUtils.kt:65-68`). Reads go through `JsonUtils.getString` behind two file-local guards — `_string` folds a Map or List to null (this mapper's spelling of Kotlin's `""` for a field it cannot read), `_userString` reads an object's `name` for the three fields that hold a person.

**Where it threw is what made it expensive.** `parseMessages` had a `try` around the whole loop, so one odd value in one reply returned `[]` for the entire thread: the detail screen drew no body and no replies, and `addReply` then rebuilt the array from that empty list, so the user's next reply replaced every earlier message and the uploader PUT the truncated array over the server's document. `addReply` now appends to the decoded array without re-encoding its elements, as Kotlin does (`FeedbackRepositoryImpl.addReply:100-110`) — even with well-formed replies the old rebuild dropped every key other than `message`/`user`/`time`, and that loss reached the server.

**The pull merges instead of overwriting.** Details under *the admin reply* below.

**Three id derivations became one.** `FeedbackMapper.idOf` is now used by the mapper, by `insertFromJson`'s batch read of existing rows, and by `sync`'s keep set. The last one mattered: it read `_id` alone, so a document carrying only `id` was stored under `fb-legacy`, collected as nothing, and deleted by `deleteNotIn` in the same sync that inserted it.

**`_formatDate` in the list screen is guarded**, as the detail screen's copy already was. An `openTime` outside the ±275,000-year range throws out of `build` and takes the whole list down over one row.

## Hardened on evidence vs. on principle

| Field | Why |
|---|---|
| a reply's `user` | **Evidence, one step removed.** The object-valued `user` is *observed* on `chat_history` — 3 of a Planet's 1,630 documents, which is what took that sync down. For feedback it is an *inference*: replies are authored through the same web UI. No captured feedback document proves it, and the code comment says so rather than claiming it. |
| a reply's `time` | **Evidence, from the Kotlin.** Kotlin reads it with Gson's `asString` (`Feedback.kt:61-69`), which renders a JSON **number** as its digits. `as String?` threw on that shape, so a numeric `time` was not merely dropped — it emptied the thread. Kotlin tolerating it is the strongest available signal that it occurs. |
| `owner`, `source` | Principle, with a Kotlin reading behind the choice of fallback: they hold a person, so an object reads its `name`. |
| `title`, `status`, `priority`, `type`, `url`, `parentCode`, `item`, `state` | Principle — blast radius. Nothing suggests the server sends objects here; they were in the same failure. |
| `_rev`, `_id` | Principle, and a weak one: an object `_rev` would mean something is deeply wrong. Included because `idOf` is a primary key and `JsonUtils.getString` would have handed it `Map.toString()`. |

`openTime` moved to `JsonUtils.getLong` for the convention, not for tolerance — the old expression already handled a numeric string.

## The admin reply that goes missing: verified, and it was two defects

The report was **right**, and Phase 152 had already pinned it (`PHASE_152_NOTES.md`, *Reported, not fixed* item 1). It has two mechanisms, and only one of them was the port's own.

1. **Port-only.** The `parseMessages`/`addReply` chain above. An admin reply carrying an object `user` or a numeric `time` made the whole thread unreadable, and the next local reply then replaced it — locally and on the server. Fixed.

2. **At parity with Kotlin, and now deliberately diverged from.** On the pending branch `fromDoc` kept the local array, discarded the server's, and adopted the server's `_rev`. Kotlin does the same (`FeedbackRepositoryImpl.kt:152, 154`), and its upload POSTs the local array back under that revision — its config declares no `dbIdExtractor`, so it is always a POST of the whole body (`UploadConfigs.kt:198-208`) and CouchDB replaces the document. So an admin replying through the web UI while a handset holds an unsent reply loses their reply **on the server**, for every device. Walking the same chain through the Kotlin is what makes this a divergence rather than a parity claim.

   The pull now merges: both arrays are append-only, so they share a prefix and diverge into the admin's replies and ours; the merge keeps the server's copy whole and re-appends only our tail after that prefix. That shape is what makes it idempotent — once our reply has landed and been pulled back, the local array is a *prefix* of the server's and the tail is empty. Elements are compared on `message`/`user`/`time` rather than on bytes, because a server that echoes a reply back with reordered keys would otherwise read as a divergence and append our copy twice.

   Merging the database alone would have **relocated** the loss rather than removed it: the outbox holds the payload as it was when the reply was written, and the two feedback screens were the only callers of `queuePending`, so the stale snapshot would have drained over the server's document anyway. `FeedbackSyncNotifier` re-queues after a completed pull.

## Mutation testing

**Twenty flips, all red.** Reverting `_string` to a raw cast; dropping `_userString`'s object case; making `_string` as strict as `chat_mapper`'s; rebuilding `addReply` from `parseMessages`; `idOf` through `JsonUtils.getString`; `parseMessages` back under one blanket `try`; both hand-rolled id derivations; the `Open` default; `openTime` as a raw int; the merge reverted, taken wholesale, and without its shared prefix; message identity on encoded bytes; the re-queue removed and the re-queue unconditional; the `''` fold reinstated in both directions; `_userString` keeping an empty name; and `idOf` falling back past an empty `_id`.

**Three of the first nine were green on the first pass** — the shared id derivation, the `Open` status default, and the `openTime` read. The tests that pin them were written after the mutation run, not counted as coverage the suite already had. The second audit then found six more unpinned claims; those have tests too, and the five new mutations are in the twenty.

## Two audit passes

Ground truth first, then the finished code — the second one found five things, four of them mine:

- the `''` fold above, which is the one that could hide a row;
- the keep-set id derivation;
- the unguarded `_formatDate`;
- a **stale citation this PR created**: `outbox_drainer.dart` cited `FeedbackMapper.fromDoc:27-31`, and the first commit inserted 60 lines above that point. That paragraph also asserted the merge defect was open. Both corrected — the "make an existing statement true" carve-out.
- One audit claim did **not** survive checking: it reported `PHASE_152_NOTES.md` as a dangling citation. The file exists at `flutter/PHASE_152_NOTES.md`.

## Files outside the stated set

- `lib/repository/feedback_repository_impl.dart`, `lib/providers/feedback_provider.dart`, `lib/ui/feedback/feedback_list_screen.dart` — the feedback slice, no other lane's set.
- `lib/repository/outbox_drainer.dart` and `PHASE_152_NOTES.md` — **documentation only**, under the "make an existing statement true" carve-out. The notes file is *annotated*, not rewritten: it is a record of what Phase 152 found, and it found this correctly.
- No new helper in `lib/core/utils/`. No `schemaVersion` bump, no table or converter change. `app_providers.dart` untouched.

## Reported, not fixed

1. **The feedback pull does not skip `_design` documents.** Kotlin filters them before insert (`TransactionSyncManager.extractDocs:355-364`, `!getString("_id", doc).startsWith("_design")`); `feedback_repository_impl.dart:190-199` filters nothing. A `_design/…` document in the `feedback` database therefore becomes a row, and since a manager reads `watchAllSorted()`, **a manager sees a bogus "Untitled feedback / Open" thread per design doc**, tappable, and permanently kept by `deleteNotIn` because it is in the keep set. Non-managers are spared only because `owner` is null. One-line fix in `insertFromJson`; left alone because I did not want a sync-shape change landing in the same PR as the merge.

2. **`FeedbackDao.deleteNotIn` has no Kotlin counterpart at all.** `FeedbackDao.kt` has no delete of any kind and the walk only inserts, so Kotlin never prunes feedback. The port's prune spares `isUploaded = false`, so a never-uploaded thread is safe — but a thread that uploaded and is then missing from a walk is gone locally where Kotlin keeps it. Whether the port wants CouchDB-cache semantics here is a policy call, not a bug I should decide alone.

3. **`rev` is written as `Value(null)` when the server omits `_rev`.** The Phase 56 shape: it should be `Value.absent()`, or a pending row's stored revision is wiped and its reply upload 409s. Unreachable today — every doc comes from `_all_docs?include_docs=true` — so it is a trap for the next caller of `insertFromJson`, not a live defect.

4. **`isUploaded` is the last raw read in `fromDoc`** (`doc.containsKey('_rev') && doc['_rev'] != null`), while `rev` beside it goes through `_string`. For an object `_rev` the row is `isUploaded = true, rev = null` — two expressions disagreeing about one field. Unreachable from CouchDB; a coherence nit, noted so the next reader does not wonder why one of the sixteen was left.

5. **`addReply` still replaces a `messages` column that is not an array.** `_decodeMessages` gives `[]` for a JSON object or a JSON string, and the reply appends to nothing. Kotlin has the same hole from the other side (`getJsonArray` normalises any non-array to `"[]"`), so it is not a parity gap — but the array-preserving fix does not reach it, and `fromDoc` will store a non-array `messages` verbatim.

6. **Two undocumented places where the port is already better than Kotlin**, listed so a future parity pass does not "correct" them back. Kotlin's `closeFeedback` writes only the status and never marks the row pending (`FeedbackDao.kt:29-30`), so the close is never uploaded *and* is reverted by the next pull; the port queues it. And Kotlin stamps every reply with the **thread owner's** name rather than the author's (`FeedbackDetailActivity.kt:82` passes `feedback?.owner`), so an admin replying to ada's thread signs it "ada"; the port uses the signed-in user. Neither is in the deliberate-deviations list in `docs/kotlin-to-flutter-migration.md`.

7. **Three hardcoded English strings in `feedback_provider.dart`** (`'Not logged in'`, `'Please enter feedback'`, `'Please select a type'`), where the screen has l10n keys for the same conditions. Reachable only by a caller that drives `FeedbackCreateNotifier.submit` without the screen's own validation, which nothing does today.

8. **`status` defaulting to `Open` is a divergence nobody wrote down** — Kotlin stores `""` (`FeedbackRepositoryImpl.kt:143`). It is now pinned by a test, but pinned is not the same as decided; the port's value is the more useful one and `docs/kotlin-to-flutter-migration.md` should say so, or it should go.

https://claude.ai/code/session_01CFAzZediTAxrzJ4bXDh29L

### DIFF
--- a/flutter/PHASE_152_NOTES.md
+++ b/flutter/PHASE_152_NOTES.md
@@ -365,6 +365,12 @@ above, and items 2 and 3 below.
    changes the latency, not the outcome. The fix is a real merge for that
    column, which is a feedback-slice decision with a conflict-resolution
    question inside it. Severity: medium-high, and older than this phase.
+   **Closed in the following round** (Lane 4, feedback hardening): the pull
+   merges the server's copy with the unsent local tail after their shared
+   prefix, and `FeedbackSyncNotifier` re-queues so the outbox snapshot is
+   refreshed rather than draining the pre-merge array. The two tests here that
+   pinned the loss now pin the merge. Annotated rather than rewritten — this
+   file is a record of what Phase 152 found, and it found this correctly.
 2. **The same shape is *constructible* for `submissions` but I could not prove
    it reachable, so I did not act on it.** A submission document is where
    Planet's grading lands (`grade`, `status`, per-answer marks, all read at
--- a/flutter/lib/data/local/feedback_mapper.dart
+++ b/flutter/lib/data/local/feedback_mapper.dart
@@ -37,15 +37,30 @@ String? _string(String key, Map<String, dynamic> json) {
 ///
 /// Used only for the fields that hold a **person** — `owner`, `source`, and a
 /// reply's `user`. A Planet's web UI writes the whole CouchDB user document
-/// into a field the handset writes as a plain user name; `name` is the same
-/// value the other documents carry as a string, so taking it is a deliberate
-/// improvement on Kotlin's `""`, which discards something unambiguous. The
-/// shape was first observed against `https://planet.learning.ole.org`, where
-/// 3 of 1,630 chat documents carried it; `chat_mapper.dart`'s `_stringOrNull`
-/// reads the object the same way. It differs on a **number**, which that
-/// helper drops and [JsonUtils.getStringOrNull] keeps as its digits — the
-/// port-wide convention, and what Gson's `asString` gives Kotlin's own reply
-/// reader (`Feedback.kt:63-65`).
+/// into a field the handset writes as a plain user name (Kotlin writes a
+/// reply's `user` as a string, `FeedbackRepositoryImpl.kt:100-104`), and
+/// `name` is the same value the other documents carry as that string.
+///
+/// Taking it is a deliberate improvement, and **how much of one depends on
+/// which field**, because Kotlin reads the two through different code:
+///
+/// * `owner` and `source` come from `JsonUtils.getString`
+///   (`FeedbackRepositoryImpl.kt:142, 145`), so Kotlin stores `""` and
+///   discards something unambiguous.
+/// * A reply's `user` is read as `ob["user"].asString` with no `JsonUtils` and
+///   no `try` (`Feedback.kt:61-69`), and `JsonObject` does not override
+///   `getAsString` — so **Kotlin throws `UnsupportedOperationException` out of
+///   `messageList`**, which `FeedbackDetailActivity.kt:61` calls from an
+///   unguarded collector. The Android app loses the detail screen on the same
+///   document. Neither app should, so this reads the name.
+///
+/// The shape was first observed against `https://planet.learning.ole.org`,
+/// where 3 of 1,630 chat documents carried it; `chat_mapper.dart`'s
+/// `_stringOrNull` reads the object the same way. It differs on a **number**,
+/// which that helper drops and [JsonUtils.getStringOrNull] keeps as its
+/// digits — the port-wide convention, and what Gson's `asString` gives
+/// Kotlin's own reply reader, where a numeric `time` renders as its literal
+/// text rather than being dropped (`Feedback.kt:63-65`).
 String? _userString(String key, Map<String, dynamic> json) {
   final value = json[key];
   if (value is Map) {
--- a/flutter/lib/data/local/feedback_mapper.dart
+++ b/flutter/lib/data/local/feedback_mapper.dart
@@ -17,29 +17,43 @@ import 'app_database.dart';
 /// one batch because 3 of them carried an object-valued `user`, fixed in
 /// `chat_mapper.dart`.
 ///
-/// Delegating to [JsonUtils.getStringOrNull] keeps a JSON **number** readable
+/// Delegating to [JsonUtils.getString] keeps a JSON **number** readable
 /// (`2019` reads as `'2019'` — the port's documented, deliberate divergence
-/// from Kotlin's `""`) and folds a missing key to null, which is what this
-/// mapper already stored for an absent field. A well-formed document therefore
-/// maps exactly as it did before.
+/// from Kotlin's `""`), and the two guards around it mean a well-formed
+/// document maps to exactly the bytes the cast produced.
 ///
-/// A Map or a List is intercepted first and reads as null, because that
-/// helper's own doc comment forbids passing it either: it would store Dart's
-/// `Map.toString()`, which is not JSON and throws wherever it is read back.
-/// Null is this mapper's spelling of Kotlin's `""` for an absent field.
+/// **Both guards are load-bearing, and the empty one is not obvious.**
+///
+/// * A missing key and an explicit JSON `null` read as null, because that is
+///   what this mapper's columns have always held for an absent field —
+///   [JsonUtils.getString] would give `''`, and [JsonUtils.getStringOrNull]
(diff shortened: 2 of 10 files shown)


dogi:
**`qlty check` is red on `1dd232a` and it is not this PR's, and I cannot re-run it.**

The status is `error`, not `failure`: *"Qlty encountered a timeout error at 15mins."* Its own `target_url` points at qlty's [inconclusive commit status (U11)](https://docs.qlty.sh/cloud/troubleshooting#inconclusive-commit-status-u11) page — the service is classifying this as *no verdict reached*, not as findings in the diff. A run that timed out produced nothing to act on.

Everything else on the same commit is green: **Analyze and test** ✅, **Build Android APK** ✅ (both matrix jobs), **myPlanet labels** ✅.

Not a service outage either — the sibling lane PRs on the same base got `qlty check` → *"No blocking issues"* (#16982 at 19:51Z, #16984 at 19:42Z), so it is this one run that stalled.

**No re-run available to me:** `qlty check` is an external GitHub App commit status, not one of this repo's workflows (`.github/workflows/` has no qlty job), so there is no run to re-trigger — and I will not push an empty commit to kick it. It will re-evaluate on the next real push to this branch.

Locally the gate this PR is held to is clean, from `flutter/`: `dart format --output=none --set-exit-if-changed lib test`, `flutter analyze` (no issues), `flutter test` — **2948 passing**.

I am keeping this PR watched until it is merged or closed.

---

PR #16986, opened 2026-09-09
all: smoother base permission extensions intent utils handling (fixes #16979)

### LINKED ISSUE #16979: Permission checks bypass the base class built to own them
BasePermissionActivity is 510 lines of permission-handling, yet ChatDetailFragment, AddResourceFragment, UserProfileFragment, and SyncActivity each still call ContextCompat.checkSelfPermission directly.

Okuro3499:
fixes #16979
Move single-permission requests into `BasePermissionActivity` with the Activity Result API, then route chat microphone and camera capture flows through it. This removes duplicated permission launchers from the affected fragments and keeps denial, rationale, and settings handling consistent across the app.

## Summary by CodeRabbit

- **Improvements**
  - Streamlined camera and microphone permission handling across chat, resource creation, and user profiles.
  - Permission checks and requests now provide more consistent feedback when access is denied.
  - Users may be guided to app settings when manual permission changes are required.
  - Camera capture flows for photos and videos now handle access requests more clearly.


dogi:
**Review — rating: 72/100**

Right instinct — the camera-denial handling (rationale check → settings dialog → toast) was copy-pasted verbatim into `AddResourceFragment` and `UserProfileFragment`, and three fragments each carried their own `registerForActivityResult(RequestPermission())`. One implementation on the base activity is where that belongs.

**It also fixes a bug in passing, which is worth calling out.** In the old `AddResourceFragment`, `requestCameraLauncher`'s grant callback unconditionally called `takePhoto()` — but `dispatchTakeVideoIntent()` launched that *same* launcher on denial. So on the video path, deny-then-grant took a photo instead of recording a video. Passing a per-call-site `onGranted` lambda makes that impossible by construction.

**The blocking problem: the pending request doesn't survive activity recreation.**

```kotlin
private var pendingPermissionRequest: PermissionRequest? = null

private val singlePermissionLauncher = registerForActivityResult(RequestPermission()) { granted ->
    val request = pendingPermissionRequest
    pendingPermissionRequest = null
    if (granted) request?.onGranted?.invoke() else request?.onDenied?.invoke()
}
```

The old code registered each launcher *in the fragment* with the callback supplied at registration time. `ActivityResultRegistry` persists the pending request by key and re-delivers it to whatever callback is registered under that key after recreation — so rotating during the system permission dialog worked: the fragment is rebuilt, re-registers, the result arrives, `startSpeechToText()` runs.

Now the callback lives in a plain field that isn't saved. Rotate while the permission dialog is up — or let the process die behind it, which is more likely with a system dialog foregrounded — and the recreated activity has `pendingPermissionRequest == null`. The result still arrives, both branches are null-safe no-ops, and the user grants camera permission and watches nothing happen. Permission dialogs are exactly the moment people pick the phone up and rotate it.

The single slot has a smaller version of the same problem: two overlapping requests and the second silently discards the first's callbacks.

Fixing this means the callback has to be reconstructible rather than captured — an enum or string key saved in `onSaveInstanceState` and dispatched through a `when` on restore, or keeping registration in the fragment and centralising only the *denial handling* (`showPermissionSettingsDialog`, the rationale branch), which is the part that was actually duplicated. The second option gets most of the deduplication with none of the state problem.

**Second, smaller: `requireActivity() as BasePermissionActivity` is an unchecked cast at three call sites.**

I traced it and it's safe today — `DashboardActivity → DashboardElementActivity → SyncActivity → ProcessUserDataActivity → BasePermissionActivity`, and none of the eleven activities that extend `AppCompatActivity` directly (`AddResourceActivity`, `SettingsActivity`, `ReplyActivity`, `PublicSurveyActivity`, `FeedbackDetailActivity`, and the rest) hosts any of these three fragments. So no crash now.

But it's a live tripwire: hosting `ChatDetailFragment` or `AddResourceFragment` from any of those eleven, which is an ordinary thing to do, is a `ClassCastException` that compiles cleanly and only fires in a permission path. `requireActivity() as? BasePermissionActivity` with an explicit fallback, or routing through an interface, costs nothing and removes the class of failure.

One nit while the string is being centralised: `Utilities.toast(this, "camera permission is required.")` is a hardcoded English literal — carried over unchanged from both old copies, so not a regression, but this was the moment to give it a `strings.xml` entry alongside the five locale files. `microphone_permission_required` right below it already is one.

CI green 5/5, and the branch is clean against master.

**Verdict:** needs another round — the permission callback has to survive recreation.

---

dogi:
**Review — rating: 86/100**

Extracts `hasPermission`, `showPermissionDeniedFeedback` and `openAppSettings` into `base/PermissionExtensions.kt` and rewires `ChatDetailFragment`, `AddResourceFragment` and `UserProfileFragment` onto them. I verified the denial truth table survives the move: the old sites branched `if (!shouldShowRequestPermissionRationale) { settings dialog } else { toast }`, and the helper's `if (shouldShowRequestPermissionRationale) { toast } else { settings dialog }` is the same mapping with the arms swapped — rationale still toasts, permanent denial still offers Settings. `checkPermission` is likewise equivalent: `strPermission?.let { hasPermission(it) } == true` yields `false` for a null permission exactly as the old `null == PERMISSION_GRANTED` did.

The genuine win here is a bug fix, not just deduplication. `AddResourceFragment` had a single `requestCameraLauncher` whose granted-callback unconditionally called `takePhoto()`, so granting camera permission from the *video* path (`dispatchTakeVideoIntent`) captured a photo instead. Splitting into `requestCameraForPhotoLauncher`/`requestCameraForVideoLauncher` with the `takePhoto`/`capturePhoto` gate-vs-action split fixes that and removes the old re-entrant `takePhoto() → launch → takePhoto()` round trip. Replacing the hardcoded `"camera permission is required."` toast with `messageRes` also makes that string localizable. I checked the removed `BasePermissionActivity.openAppSettings()` has no callers outside the class (only line 337), and since the extension lands in the same `org.ole.planet.myplanet.base` package it resolves without an import — consistent with build (default + lite) and test being green. `ContextCompat` remains imported in all three fragments because each still uses `getColor`, so no imports were orphaned.

**Nit:** the title says `all:`, but `AudioRecorder.kt` and `CameraUtils.kt` still call `ContextCompat.checkSelfPermission` directly — two sites short of actually centralizing everything. Also, `base/` is documented as the home for base classes; a free-function extension file would sit more naturally in `utils/`.

**Verdict:** Mergeable — behaviour-preserving where it should be, and it corrects a real photo/video launcher mix-up on the way.

---

dogi (approved):

PR #16988, opened 2026-09-10
life: smoother dictionary translating (fixes #16704)

### LINKED ISSUE #16704: string: dictionary_parsing_failed added by PR #15863 did not translate

J-S-webskas:
fixes #16704
See code

## Summary by CodeRabbit

- **Localization**
  - Added translated dictionary parsing failure messages for Arabic, Spanish, French, Nepali, and Somali.
  - Users will now see localized error text when dictionary processing fails in these languages.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 7060
-        versionName = "0.70.60"
+        versionCode = 7061
+        versionName = "0.70.61"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/res/values-ar/strings.xml
+++ b/app/src/main/res/values-ar/strings.xml
@@ -605,6 +605,7 @@
     <string name="please_select_starting_date">"يرجى تحديد تاريخ البدء: "</string>
     <string name="read_offline_news_from">"قراءة الأخبار غير المتصلة من: "</string>
     <string name="downloading_started_please_check_notification">بدأ التنزيل، يرجى التحقق من الإشعار…</string>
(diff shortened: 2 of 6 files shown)


dogi:
**Review — rating: 90/100**

Closes a real gap rather than adding a string nobody reads. `dictionary_parsing_failed` already exists in `values/strings.xml:613` and is used at `DictionaryActivity.kt:94`, so every non-English user was seeing "Dictionary parsing failed." in English. All five locales now have it, and `values/` is untouched — purely additive, no risk to the default path.

Things I checked rather than assumed:

- **No duplicate keys.** One occurrence per file across all six, so no build-breaking redefinition (which is the usual way a translation PR goes wrong).
- **The French apostrophe is escaped**: `L\'analyse du dictionnaire a échoué.` An unescaped `'` in an Android string resource is an aapt error, and the commit history shows a "fixed values-fr" follow-up, so that was caught and handled.
- **Placement is consistent.** All five insert after `downloading_started_please_check_notification` and before `file_already_exists`, at the same relative position — matching where the English sits in its own file. Diff-friendly for the next translator.
- **The translations read correctly** — Arabic "فشل تحليل القاموس", Spanish "Error al analizar el diccionario", French "L'analyse du dictionnaire a échoué", Somali "Falanqaynta qaamuuska way fashilantay" all say what the English says, idiomatically rather than word-for-word. Nepali uses the transliterated "पार्सिङ" for *parsing*, which is the normal convention for technical terms in that locale.

One cosmetic inconsistency: English, Arabic, Spanish and French end with a period; Nepali and Somali have no terminal punctuation. For Nepali, `।` (danda) would be the idiomatic sentence terminator. Not worth blocking on, but easy to even out while the file is open.

CI green 5/5, branch clean against master.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #16997, opened 2026-09-10
resources: smoother download working (fixes #16996)

### LINKED ISSUE #16996: Failed to download
```
Failed to download https://satellite:1983@planet.learning.ole.org:443/db/resources/9517e3b45a5bb63e69bb8f269216ba83/Captura_de_pantalla_2024-10-30_120212.png
android.app.BackgroundServiceStartNotAllowedException: Not allowed to start service Intent { act=ACTION_NOTIFY xflg=0x4 cmp=org.ole.planet.myplanet/androidx.work.impl.foreground.SystemForegroundService (has extras) }: app is in background uid UidRecord{ece203e u0a305 SVC  bg:+5m35s755ms idle change:procstate procs:0 seq(217177,216965)} caps=-------TI
at android.app.ContextImpl.startServiceCommon(ContextImpl.java:2125)
at android.app.ContextImpl.startService(ContextImpl.java:2080)
at android.content.ContextWrapper.startService(ContextWrapper.java:870)
at androidx.work.impl.utils.WorkForegroundUpdater.lambda$setForegroundAsync$0$androidx-work-impl-utils-WorkForegroundUpdater(WorkForegroundUpdater.java:98)
at androidx.work.impl.utils.WorkForegroundUpdater$$ExternalSyntheticLambda0.invoke(D8$$SyntheticClass:0)
at androidx.work.ListenableFutureKt.executeAsync$lambda$4$lambda$3(ListenableFuture.kt:59)
at androidx.work.ListenableFutureKt$$ExternalSyntheticLambda3.run(D8$$SyntheticClass:0)
at androidx.work.impl.utils.SerialExecutorImpl$Task.run(SerialExecutorImpl.java:96)
at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1100)
at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
at java.lang.Thread.run(Thread.java:1572)
```

Okuro3499:
fixes #16996
Refactors the per-URL download loop so each step is isolated: download failures, progress-notification failures, and broadcast/update failures are now handled independently with targeted logging. Adds explicit error logging for failed download responses and introduces a safer notification path by only promoting to foreground when allowed, with a fallback to a normal notification if foreground promotion is unavailable or throws.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 7058
-        versionName = "0.70.58"
+        versionCode = 7059
+        versionName = "0.70.59"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/services/DownloadWorker.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/services/DownloadWorker.kt
@@ -62,17 +62,24 @@ class DownloadWorker @AssistedInject constructor(
             val authHeader = UrlUtils.header
 
             urls.forEachIndexed { index, url ->
-                try {
-                    val success = downloadFile(url, authHeader, index, urls.size)
-                    results.add(success)
-                    completedCount++
-
-                    showProgressNotification(completedCount - 1, urls.size, context.getString(R.string.downloaded_files, "$completedCount", "${urls.size}"), 100)
-                    sendDownloadUpdate(url, success, completedCount >= urls.size, fromSync)
+                val success = try {
+                    downloadFile(url, authHeader, index, urls.size)
                 } catch (e: Exception) {
                     Log.e(TAG, "Failed to download $url", e)
-                    results.add(false)
-                    completedCount++
+                    false
+                }
+                results.add(success)
+                completedCount++
+
+                try {
+                    showProgressNotification(completedCount - 1, urls.size, context.getString(R.string.downloaded_files, "$completedCount", "${urls.size}"), 100)
+                } catch (e: Exception) {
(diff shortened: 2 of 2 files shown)


dogi:
**Review — rating: 84/100**

Three changes to `DownloadWorker`, and the first one fixes a real accounting bug rather than just adding logs. On master the whole per-URL body sat in one `try`, so a `downloadFile` that *succeeded* followed by a throwing `showProgressNotification` fell into the catch and ran `results.add(false); completedCount++` a second time — double-counting the file and recording a spurious failure. Splitting the download from the notification and the update means `results.add(success)` and `completedCount++` now happen exactly once per URL, and a notification failure no longer suppresses `sendDownloadUpdate` (including its `completedCount >= urls.size` completion flag).

The `setForeground` fallback is the right shape too: I checked `DownloadUtils.canStartForegroundService` exists (`utils/DownloadUtils.kt:210`) and that `DownloadUtils.kt:177` already gates on it the same way, so this makes the worker consistent with the util rather than inventing a new policy — and `notificationManager` is already a field (`DownloadWorker.kt:42`, used at 165), so the `notify` fallback needs no new plumbing. Adding `code`/`message` to the `DownloadResult.Error` log closes a genuine blind spot.

**Nit:** `showProgressNotification` (line 148) and `sendDownloadUpdate` (line 168) are both `suspend`, so the new `catch (e: Exception)` blocks will swallow `CancellationException` and let the loop keep walking URLs after the worker is stopped. For a PR whose point is hardening, `catch (e: CancellationException) { throw e }` first — or a `currentCoroutineContext().ensureActive()` — would be worth it.

**Verdict:** Mergeable — net clearly better than master; the cancellation nit is worth a follow-up but doesn't block.

---

dogi (approved):

PR #16998, opened 2026-09-10
chat: smoother detail speech recognizing (fixes #16985)

### LINKED ISSUE #16985: NullPointerException
```
FATAL EXCEPTION: main
Process: org.ole.planet.myplanet, PID: 23758
java.lang.NullPointerException
	at org.ole.planet.myplanet.ui.chat.ChatDetailFragment.getBinding(ChatDetailFragment.kt:55)
	at org.ole.planet.myplanet.ui.chat.ChatDetailFragment.stopSpeechToText(ChatDetailFragment.kt:231)
	at org.ole.planet.myplanet.ui.chat.ChatDetailFragment.access$stopSpeechToText(ChatDetailFragment.kt:52)
	at org.ole.planet.myplanet.ui.chat.ChatDetailFragment$initSpeechRecognizer$1.onError(ChatDetailFragment.kt:176)
	at android.speech.SpeechRecognizerImpl$InternalRecognitionListener$1.handleMessage(SpeechRecognizerImpl.java:611)
	at android.os.Handler.dispatchMessageImpl(Handler.java:142)
	at android.os.Handler.dispatchMessage(Handler.java:125)
	at android.os.Looper.loopOnce(Looper.java:296)
	at android.os.Looper.loop(Looper.java:397)
	at android.app.ActivityThread.main(ActivityThread.java:9523)
	at java.lang.reflect.Method.invoke(Native Method)
	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:575)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:939)
```

Okuro3499:
fixes #16985
Guard speech recognition UI updates behind a nullable binding in `ChatDetailFragment` so callbacks do not touch views after the fragment view is destroyed. This also extracts a shared `resetListeningUi()` helper and uses it on recognition errors and stop events to keep the mic state consistent.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 7059
-        versionName = "0.70.59"
+        versionCode = 7060
+        versionName = "0.70.60"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/ui/chat/ChatDetailFragment.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/chat/ChatDetailFragment.kt
@@ -170,9 +170,11 @@ class ChatDetailFragment : Fragment() {
             speechRecognizer?.setRecognitionListener(object : RecognitionListener {
                 override fun onReadyForSpeech(params: Bundle?) {
                     isListening = true
-                    binding.buttonGchatMic.setColorFilter(ContextCompat.getColor(requireContext(), R.color.md_red_500))
-                    binding.textGchatIndicator.text = getString(R.string.voice_to_text)
-                    binding.textGchatIndicator.visibility = View.VISIBLE
+                    _binding?.let { binding ->
+                        binding.buttonGchatMic.setColorFilter(ContextCompat.getColor(requireContext(), R.color.md_red_500))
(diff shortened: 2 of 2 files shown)


dogi:
**Review — rating: 85/100**

Guards every `binding` touch inside the `RecognitionListener` callbacks behind `_binding?.let`, and splits `resetListeningUi()` out of `stopSpeechToText()`.

The crash class is real, not theoretical. `binding` is `_binding!!` (`ChatDetailFragment.kt:56`), so any access after `onDestroyView` nulls it at line 702 throws. `onDestroyView` does call `speechRecognizer?.destroy()` first (line 691), but `SpeechRecognizer` results arrive over Binder and are posted to the main-thread queue — a callback already sitting in that queue when `onDestroyView` runs is not unqueued by `destroy()`, so it lands afterwards and dereferences a null binding. `onPartialResults` firing repeatedly during dictation makes that window easy to hit by backgrounding or navigating mid-speech.

The pattern also matches what the rest of the file already does — `_binding?.let { binding -> … }` appears at lines 567, 575, 582 and 664 — so this is bringing the speech callbacks in line with the file's own convention rather than introducing a new idiom. Routing `onError` to `resetListeningUi()` instead of `stopSpeechToText()` is right too: the recognizer has already terminated by the time `onError` fires, so the extra `stopListening()` was redundant.

**Nit:** `requireContext()` is still called inside the guarded blocks. It's safe in practice (a non-null binding implies an attached fragment), but `binding.root.context` would make the block self-consistent and drop the last throw site.

**Verdict:** Mergeable.

---

dogi (approved):

PR #17000, opened 2026-09-10
all: smoother base class view binding (fixes #16999)

### LINKED ISSUE #16999: View-binding lifecycle boilerplate repeats in 51 fragments
The manual private var _binding / _binding = null pair is retyped per fragment. A generic BaseBindingFragment<T> would remove it everywhere at once.

Okuro3499:
fixes #16999
Add shared base classes and a binding delegate for fragments, dialog fragments, and bottom sheets, then migrate existing screens to use them. This removes repeated `_binding` setup/cleanup code, centralizes view binding lifecycle handling, and keeps fragment-specific teardown only where extra cleanup is still needed.

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/base/BaseBindingBottomSheetFragment.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/base/BaseBindingBottomSheetFragment.kt
@@ -1,0 +1,25 @@
+package org.ole.planet.myplanet.base
+
+import android.os.Bundle
+import android.view.LayoutInflater
+import android.view.View
+import android.view.ViewGroup
+import androidx.viewbinding.ViewBinding
+import com.google.android.material.bottomsheet.BottomSheetDialogFragment
+
+abstract class BaseBindingBottomSheetFragment<VB : ViewBinding>(
+    inflateBinding: (LayoutInflater, ViewGroup?, Boolean) -> VB
+) : BottomSheetDialogFragment() {
+    private val bindingDelegate = FragmentViewBindingDelegate(inflateBinding)
+    protected val binding: VB get() = bindingDelegate.value
+    protected val _binding: VB? get() = bindingDelegate.bindingOrNull
+
+    override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View {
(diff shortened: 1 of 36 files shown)


dogi:
**Review — rating: 87/100**

Adds `FragmentViewBindingDelegate` plus three base classes (`BaseBindingFragment`, `BaseBindingDialogFragment`, `BaseBindingBottomSheetFragment`) and migrates 32 fragments onto them, netting −387/+194.

I verified the two ways a refactor like this normally breaks, across all 32 files rather than by sampling:

- **Teardown ordering.** The base runs `super.onDestroyView()` and *then* `bindingDelegate.clear()`, so any subclass that calls `super.onDestroyView()` first and then touches `binding` would NPE. Ten subclasses still override `onDestroyView`; nine of them (`ChatDetailFragment`, `ChatHistoryFragment`, `HomeCommunityDialogFragment`, `TakeCourseFragment`, `MyHealthFragment`, `SubmissionsFragment`, `TeamFragment`, `UserProfileFragment`, `ResourceViewerFragment`) do all their binding work *before* the super call, and the tenth (`StorageBreakdownFragment`) calls super first but only touches `progressDialog`. No live case.
- **Inflation ownership.** Fifteen subclasses still override `onCreateView`; every one of them opens with `val view = super.onCreateView(...)` before dereferencing `binding`, so the delegate is always populated first. Checked `ChatDetailFragment:117`, `CalendarFragment:22`, `TeamFragment:64` and `AddResourceFragment:135` in full and grepped the rest for the pattern.

Keeping `_binding` as a `protected val` returning `bindingOrNull` is what lets the existing `_binding?.let { }` guards in migrated fragments survive untouched, which is why the diff stays this small.

**Nit:** two naming points. An underscore prefix conventionally marks a private backing field, so a `protected val _binding` reads oddly as API — `bindingOrNull` would say what it is. And `FragmentViewBindingDelegate` isn't a Kotlin property delegate (no `getValue` operator), it's a holder; the name sets up the wrong expectation. Neither is worth another CI cycle on its own.

**Verdict:** Mergeable — mechanical, consistently applied, and I confirmed the two failure modes that would have made it unsafe don't occur.

---

J-S-webskas (approved):

dogi (approved):

PR #17002, opened 2026-09-10
all: smoother confirmation dialog utils handling (fixes #17001)

### LINKED ISSUE #17001: 32 files hand-build the same AlertDialog shape
```
Title, message, two buttons, and a content description are wired near-identically in 32 places. Extract a small dialog-factory extension function for that one shape.
```

Okuro3499:
fixes #17001
Replace repeated AlertDialog setup across audio, resources, exams, teams, and other screens with DialogUtils.confirmDialog. The new helper standardizes styling, default button labels, dismiss behavior, and button content descriptions, and adds Robolectric coverage for the shared dialog behavior.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 7067
-        versionName = "0.70.67"
+        versionCode = 7068
+        versionName = "0.70.68"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/services/AudioRecorder.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/services/AudioRecorder.kt
@@ -14,7 +14,6 @@ import androidx.activity.result.ActivityResultCaller
 import androidx.activity.result.ActivityResultLauncher
 import androidx.activity.result.contract.ActivityResultContracts
 import androidx.annotation.RequiresApi
-import androidx.appcompat.app.AlertDialog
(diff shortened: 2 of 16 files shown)


dogi:
**Review — rating: 89/100**

Adds `Context.confirmDialog(...)` to `DialogUtils` and routes 14 hand-rolled `AlertDialog.Builder` confirmations across 13 files through it, with 6 Robolectric tests for the helper.

The risk in a sweep like this is a converted site silently gaining or losing a button, so I checked all 13 files mechanically rather than by eye: every one removed exactly matching `setPositiveButton`/`setNegativeButton` pairs (`VoicesAdapter` 2+2, the rest 1+1), so no single-button dialog was turned into a two-button one by the defaults. I also traced the awkward one — `ExamTakingFragment.askResumeOrRestart`, which bridges a dialog into `suspendCancellableCoroutine`. It keeps `cancelable = false`, still resumes `true`/`false` from the two arms behind the `cont.isActive` guard, and still gets a real `AlertDialog` back so `cont.invokeOnCancellation { dialog.dismiss() }` keeps working. The helper's `builder.show()`-then-rebind-listeners pattern replaces the framework's implicit auto-dismiss with an explicit `dialog.dismiss()` before invoking the callback, which is the same observable behaviour.

Two things I liked beyond the deduplication: default `positiveText`/`negativeText` resolve off the extension receiver so callers don't repeat `getString(R.string.yes)`, and every dialog now gets `contentDescription` on both buttons — an accessibility improvement none of the original sites had.

**Nit:** `confirmDialog` shows the dialog rather than building it, so `showConfirmDialog` would set the right expectation for the return value being a live, already-visible dialog.

**Verdict:** Mergeable.

---

dogi (approved):

PR #17012, opened 2026-09-10
courses: smoother progress chip filter selecting (fixes #16975)

### LINKED ISSUE #16975: courses filter resets on app theme change
https://github.com/user-attachments/assets/24b29fea-0530-4d0b-bc16-81c330ea2f95

Okuro3499:
fixes #16975
Use `viewModel.currentFilterState.progressFilter` as the fallback selected value when `filterController` is not initialized, instead of defaulting to an empty filter. This keeps the course progress chip selection in sync with the current filter state during early render paths.
[Screen_recording_20260910_185355.webm](https://github.com/user-attachments/assets/f3f740bf-2024-49d2-af40-287febec2dab)

## Summary by CodeRabbit

- **Bug Fixes**
  - Preserved the selected course progress filter when the filter controls are still initializing, preventing the selection from appearing blank.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 7061
-        versionName = "0.70.61"
+        versionCode = 7062
+        versionName = "0.70.62"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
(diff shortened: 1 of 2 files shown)


dogi:
**Review — rating: 82/100**

`renderCourseChipSelection` fell back to `""` whenever `filterController` wasn't initialized yet, which then hit the `selected.isEmpty() && i == 0` branch and highlighted the *first* chip regardless of the filter actually in effect. Falling back to `viewModel.currentFilterState.progressFilter` fixes that, and I checked the property exists and is the right source of truth — `CoursesViewModel.kt:45` declares `var currentFilterState: FilterState = FilterState("", "", "", emptyList())` with a private setter, and line 136 uses that same field to decide whether to re-apply filtering. Because its default is `""`, the new branch is behaviourally identical on a cold start and only diverges in exactly the case that was broken: the ViewModel survives a view recreation carrying a real filter while the controller has not been re-initialised.

**Nit:** no test. `CoursesViewModel` is already unit-tested, so a case that seeds `currentFilterState` with a non-empty `progressFilter` and asserts the chip mapping would pin this against regression cheaply.

**Verdict:** Mergeable.

---

dogi (approved):

PR #17015, opened 2026-09-10
all: smoother crash log store capacity checking (fixes #16871)

### LINKED ISSUE #16871: Short-circuit the crash-log capacity check
## What

The crash-log save path does more work than needed to decide whether it is at capacity.

**File:** `app/src/main/java/org/ole/planet/myplanet/utils/CrashLogStore.kt` (`save`)

## Why

This runs on the crash path — the one place where doing less work matters most, because the process may be seconds from dying.

## The fix

Short-circuit the capacity check so the store stops counting once it knows the answer, rather than evaluating the full collection.

Keep the eviction policy, the ordering of retained logs, and the persisted format byte-identical.

## Acceptance

- Same logs retained and evicted for the same input, including the exactly-at-capacity boundary.
- No change to what a crash writes to disk.

## Notes

- Read the current implementation before changing it — the win here is real but small, and the correctness of crash-log retention matters more than the saving. If the check turns out to already short-circuit, say so and close rather than restructuring for its own sake.

dogi:
Short-circuits the capacity check in CrashLogStore.save by evaluating directory files as a sequence and taking at most MAX_PENDING_FILES valid log files. Retains identical eviction, retention, and formatting logic.

Fixes #16871

---
*PR created automatically by Jules for task [12457252658846363948](https://jules.google.com/task/12457252658846363948) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/utils/CrashLogStore.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/utils/CrashLogStore.kt
@@ -37,7 +37,9 @@ object CrashLogStore {
         return try {
             val logDir = dir(context)
             if (!logDir.exists() && !logDir.mkdirs()) return null
-            if ((logDir.listFiles()?.asSequence()?.filter { isValidLogFile(it) }?.take(MAX_PENDING_FILES)?.count() ?: 0) >= MAX_PENDING_FILES) return null
+            val pendingFiles = logDir.listFiles().orEmpty()
+            val validCount = pendingFiles.asSequence().filter { isValidLogFile(it) }.take(MAX_PENDING_FILES).count()
+            if (validCount >= MAX_PENDING_FILES) return null
             val file = File(logDir, "${timeProvider.now()}_$type$FILE_EXTENSION")
             file.writeText(error)
             file
--- a/app/src/main/java/org/ole/planet/myplanet/utils/CrashLogStore.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/utils/CrashLogStore.kt
@@ -37,7 +37,7 @@ object CrashLogStore {
         return try {
             val logDir = dir(context)
             if (!logDir.exists() && !logDir.mkdirs()) return null
-            if ((logDir.listFiles()?.count { isValidLogFile(it) } ?: 0) >= MAX_PENDING_FILES) return null
+            if ((logDir.listFiles()?.asSequence()?.filter { isValidLogFile(it) }?.take(MAX_PENDING_FILES)?.count() ?: 0) >= MAX_PENDING_FILES) return null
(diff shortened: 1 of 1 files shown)


dogi:
**Review — rating: 72/100**

The rewrite is correct — `filter{}.take(N).count() >= N` and `count{} >= N` agree for every input, so the capacity gate still behaves identically. My problem is that the short-circuit it buys is unreachable.

`MAX_PENDING_FILES` is 20 (`CrashLogStore.kt:15`), and this very function is the thing that enforces it: `save` returns `null` rather than writing once 20 valid logs exist. So the directory cannot hold more than 20 valid log files, which means `take(20)` never truncates anything and the sequence walks the same entries the old `count {}` did. The saving would only materialise in a state the store already prevents.

It also doesn't help in the case you might expect it to. If the directory picked up foreign files, `isValidLogFile` → `parseLogFile` (line 23) rejects them at `file.isFile`/extension, so they never count toward the `take(20)` budget and every one of them still gets stat'd. Foreign clutter is scanned in full either way.

What's left is a real readability cost: four chained safe-calls with a trailing elvis on one line, replacing a plain `count { }`.

If there's a profile showing this stat loop mattering, say so and I'll take it — but then it's worth a named local (`val pending = logDir.listFiles().orEmpty().asSequence()...`) rather than a 130-character condition. Otherwise reverting to `count { }` is the better line.

**Verdict:** Needs another round — correct, but it trades legibility for an optimization the surrounding cap makes dead.

---

dogi:
**Re-review — Rating: 82/100**

The rewrite is correct. `.take(MAX_PENDING_FILES).count()` caps the count at 20, and the decision is `>= MAX_PENDING_FILES`, so the truncation can't change the outcome — the guard fires in exactly the same cases as before. ✓

**What it actually saves**

`isValidLogFile` delegates to `parseLogFile`, which opens with `file.isFile` — a real `stat` syscall, not a name check. So short-circuiting after 20 valid files does avoid filesystem work, which is more than the name-parsing that the rest of `parseLogFile` does.

The honest framing is that the saving is small and conditional. `listFiles()` has already performed the directory read and allocated the `File[]` before the sequence starts, and that's the dominant cost. More to the point, the guard itself caps the directory at 20 valid logs, so in steady state the count never exceeds `MAX_PENDING_FILES` and there's nothing to short-circuit. The win only materialises when the directory has drifted above the cap — `MAX_PENDING_FILES` lowered in a past release, or unrelated files sharing the directory — which is real but not the common path.

That's not an argument against merging a three-line change that's strictly no worse. It's an argument for not describing it as a performance fix.

**Nit:** `pendingFiles` is used exactly once, so the two statements could be one expression. Minor either way; the named intermediate arguably reads better than a four-call chain.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17016, opened 2026-09-10
courses: smoother tags repositories link querying (fixes #16907)

### LINKED ISSUE #16907: Narrow the TagsRepository course-link API with getCourseLinkIds
## Evidence

`CoursesRepositoryImpl` reaches for the tags contract with a raw db-name literal:

```
app/src/main/java/org/ole/planet/myplanet/repository/CoursesRepositoryImpl.kt
    override suspend fun filterCourses(
        searchText: String,
        gradeLevel: String,
        subjectLevel: String,
        tagNames: List<String>
    ): List<MyCourse> {
        val courseIdsWithTags = if (tagNames.isNotEmpty()) {
            tagsRepository.getLinkIdsForTagNames("courses", tagNames).toSet()
        } else {
            null
        }
```

Grep anchor: `getLinkIdsForTagNames("courses", tagNames).toSet()` — exactly one occurrence in the tree.

The `"courses"` db-name string is a tags-layer detail living in the courses repository, and the `.toSet()` conversion is repeated at the call site rather than expressed by the contract.

The tags contract today:

```
app/src/main/java/org/ole/planet/myplanet/repository/TagsRepository.kt
    suspend fun getLinkIdsForTagNames(dbType: String, tagNames: List<String>): List<String>
```

## Fix

1. **`repository/TagsRepository.kt`** — add one method next to the existing declaration:

   ```kotlin
   suspend fun getCourseLinkIds(tagNames: List<String>): Set<String>
   ```

2. **`repository/TagsRepositoryImpl.kt`** — implement it as a thin delegation:

   ```kotlin
   override suspend fun getCourseLinkIds(tagNames: List<String>): Set<String> =
       getLinkIdsForTagNames("courses", tagNames).toSet()
   ```

   **Placement matters:** add this override as the *last* member of the class, after the existing `insert(documentList: List<JsonObject>)` function and immediately before the closing brace. Do not reformat, reorder, or otherwise touch the body of `getLinkIdsForTagNames` — another change is in flight inside it.

3. **`repository/CoursesRepositoryImpl.kt`** — update the single call site in `filterCourses` in the same change:

   ```kotlin
   val courseIdsWithTags = if (tagNames.isNotEmpty()) {
       tagsRepository.getCourseLinkIds(tagNames)
   } else {
       null
   }
   ```

   Change nothing else in that function — the `tagNames.isNotEmpty()` guard, the `null` branch and the `courseIdsWithTags.isNullOrEmpty()` early return all stay exactly as they are.

Both the new method and the call-site update must land in this one change. A new method whose only caller is a test is not the point of the task.

## Acceptance

- `getLinkIdsForTagNames("courses", ...)` no longer appears in `CoursesRepositoryImpl`; the `"courses"` literal exists only inside `TagsRepositoryImpl`.
- `getLinkIdsForTagNames` stays on the interface — other db types still use it. This is an addition, not a replacement.
- Behaviour is byte-identical: same ids, same set semantics, same empty/null handling.
- Add a unit test in `app/src/test/java/org/ole/planet/myplanet/repository/TagsRepositoryImplTest.kt` asserting `getCourseLinkIds` delegates with the `"courses"` db type and returns the deduplicated set. Follow the existing MockK + `runTest` style in that file, and append it at the end of the class rather than inserting among the existing tests.

## Leashes

- Touch only these three main-source files plus the one test file.
- No Room entity, DAO, schema or `AppDatabase` version change.
- No signature change to `getLinkIdsForTagNames`, and no change to its body or to the two `mapNotNull`/`distinct` sites inside `TagsRepositoryImpl` (lines ~73 and ~103) — a separate change owns those.
- No new dependency, no dispatcher change, no logging.

dogi:
Narrows the course-link API in `TagsRepository` by adding `getCourseLinkIds(tagNames: List<String>): Set<String>`. Updates `CoursesRepositoryImpl` to use `tagsRepository.getCourseLinkIds(tagNames)` and adds unit test coverage in `TagsRepositoryImplTest`.

Fixes #16907

---
*PR created automatically by Jules for task [6107245853913768021](https://jules.google.com/task/6107245853913768021) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 7062
-        versionName = "0.70.62"
+        versionCode = 7063
+        versionName = "0.70.63"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/repository/CoursesRepositoryImpl.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/CoursesRepositoryImpl.kt
@@ -289,7 +289,7 @@ class CoursesRepositoryImpl @Inject constructor(
         tagNames: List<String>
     ): List<MyCourse> {
         val courseIdsWithTags = if (tagNames.isNotEmpty()) {
-            tagsRepository.getLinkIdsForTagNames("courses", tagNames).toSet()
+            tagsRepository.getCourseLinkIds(tagNames)
         } else {
             null
         }
--- a/app/src/main/java/org/ole/planet/myplanet/repository/TagsRepository.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/TagsRepository.kt
@@ -11,6 +11,5 @@ interface TagsRepository {
     suspend fun getTagsForResources(resourceIds: List<String>): Map<String, List<TagEntity>>
     suspend fun getTagsForCourses(courseIds: List<String>): Map<String, List<TagEntity>>
     suspend fun insert(documentList: List<JsonObject>)
-    suspend fun getLinkIdsForTagNames(dbType: String, tagNames: List<String>): List<String>
     suspend fun getCourseLinkIds(tagNames: List<String>): Set<String>
 }
--- a/app/src/main/java/org/ole/planet/myplanet/repository/TagsRepository.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/TagsRepository.kt
@@ -12,4 +12,5 @@ interface TagsRepository {
     suspend fun getTagsForCourses(courseIds: List<String>): Map<String, List<TagEntity>>
     suspend fun insert(documentList: List<JsonObject>)
     suspend fun getLinkIdsForTagNames(dbType: String, tagNames: List<String>): List<String>
+    suspend fun getCourseLinkIds(tagNames: List<String>): Set<String>
 }
--- a/app/src/main/java/org/ole/planet/myplanet/repository/TagsRepositoryImpl.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/TagsRepositoryImpl.kt
@@ -151,4 +151,7 @@ class TagsRepositoryImpl @Inject constructor(
         tag.isAttached = attachedTo.isNotEmpty()
         return tag
     }
+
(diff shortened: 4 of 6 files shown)


dogi:
**Review — rating: 70/100**

The new `getCourseLinkIds(tagNames): Set<String>` is the right shape — it bakes in the `"courses"` dbType so callers can't typo it, returns the `Set` the caller actually wanted instead of making `CoursesRepositoryImpl` call `.toSet()` itself, and comes with a test that checks both the delegation and the dedup (`linkTag1`/`linkTag3` both mapping to `course1`).

But the PR is called "Narrow TagsRepository course-link API" and the interface came out **wider**, not narrower. After this change `getLinkIdsForTagNames` has no production callers left at all — grepping `app/src/main` on the branch, the only three hits are its own declaration in `TagsRepository.kt:14`, its implementation at `TagsRepositoryImpl.kt:57`, and the new delegate calling it at `TagsRepositoryImpl.kt:156`. `CoursesRepositoryImpl.kt:292` was its last real consumer and now goes through the new method.

So `TagsRepository` still publishes the loose, stringly-typed `getLinkIdsForTagNames(dbType: String, ...)` that any future caller can reach for with an arbitrary `dbType` — which is precisely the surface the PR sets out to close. Right now it's two methods where there should be one.

Drop `getLinkIdsForTagNames` from the `TagsRepository` interface and make it `private` in `TagsRepositoryImpl` (it's still needed there as the shared body). Any test that exercises it through the interface moves onto `getCourseLinkIds`. That's a small follow-up on top of what's already here and it makes the title true.

**Verdict:** Needs another round — good helper, but the narrowing it promises hasn't happened yet.

---

dogi:
**Re-review — Rating: 88/100** (was 70)

The narrowing landed. `getLinkIdsForTagNames` is off the `TagsRepository` interface and `private` in `TagsRepositoryImpl` (line 57), so the interface now publishes one course-specific method instead of that plus a stringly-typed `dbType` variant any caller could reach for. That was the whole of my earlier objection and it's fully addressed.

The test migration went further than strictly required, and correctly: the three pre-existing cases that exercised the old method through `"resources"` now go through `getCourseLinkIds` with `"courses"`, rather than being deleted or left asserting a path that no longer exists. `CoursesRepositoryImplTest` is updated to stub `getCourseLinkIds(any())` returning `emptySet()`. Build (default + lite) and both test shards green.

**Nit:** now that `getLinkIdsForTagNames` is private with exactly one caller passing `"courses"`, the `dbType` parameter is effectively a constant. Worth folding in if the general form isn't coming back.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17017, opened 2026-09-10
resources: less base repositories injections is more (fixes #17005)

### LINKED ISSUE #17005: Drop the two unused repository injections from BaseResourceFragment
## What

`BaseResourceFragment` injects two repositories it never uses, so every resource/course/voices/team/recycler screen that inherits from it builds two repositories nothing calls.

`app/src/main/java/org/ole/planet/myplanet/base/BaseResourceFragment.kt`

## Why

```kotlin
import org.ole.planet.myplanet.repository.SubmissionsRepository   // 34
import org.ole.planet.myplanet.repository.SurveysRepository       // 35
...
lateinit var submissionsRepository: SubmissionsRepository         // 61
lateinit var surveysRepository: SurveysRepository                 // 63
```

Those two declarations plus the two imports are the *only* occurrences of either symbol in the file. Every subclass that actually needs one — `BaseExamFragment`, `ExamTakingFragment`, `UserInformationFragment`, and the submissions/surveys/dashboard ViewModels — declares its own, so no subclass resolves the inherited field.

## The fix

Delete both `@Inject lateinit var` declarations and the two now-dead imports. Nothing else.

## Acceptance

- The app compiles; no subclass loses a field it was actually reading.
- Both `@Inject` sites and both imports are gone from `BaseResourceFragment.kt`.

## Notes

~6 removed lines, 1 file. Before deleting, re-check that no subclass reads the inherited field (search for `submissionsRepository` / `surveysRepository` and confirm each hit declares its own). Conventions: `docs/CODE_STYLE_GUIDE.md`.

dogi:
Removes unused SubmissionsRepository and SurveysRepository @Inject fields and imports from BaseResourceFragment.

Fixes #17005

---
*PR created automatically by Jules for task [12920716114250269955](https://jules.google.com/task/12920716114250269955) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 7063
-        versionName = "0.70.63"
+        versionCode = 7064
+        versionName = "0.70.64"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/base/BaseResourceFragment.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/base/BaseResourceFragment.kt
@@ -31,8 +31,6 @@ import org.ole.planet.myplanet.model.TagEntity
 import org.ole.planet.myplanet.model.UserEntity
 import org.ole.planet.myplanet.repository.CoursesRepository
 import org.ole.planet.myplanet.repository.ResourcesRepository
-import org.ole.planet.myplanet.repository.SubmissionsRepository
-import org.ole.planet.myplanet.repository.SurveysRepository
 import org.ole.planet.myplanet.repository.UserRepository
 import org.ole.planet.myplanet.services.BroadcastService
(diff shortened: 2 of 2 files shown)


dogi:
**Review — rating: 88/100**

Drops the unused `@Inject lateinit var submissionsRepository` and `surveysRepository` from `BaseResourceFragment`, plus their imports. Six lines, all deletions.

I checked the one way this could break — a subclass leaning on the inherited field — and it doesn't happen. The other heavy users of these two repositories are unrelated types that declare their own: `BaseExamFragment` is `: Fragment()`, not a `BaseResourceFragment` subclass, and carries its own `@Inject lateinit var submissionsRepository` at line 46; `BellDashboardViewModel`, `DashboardViewModel` and `ExamTakingFragment` likewise inject for themselves. Consistent with build (default + lite) and both test shards passing.

Worth more than the line count suggests: `BaseResourceFragment` sits under a large share of the resource UI, so every one of those fragments was constructing two repository dependencies it never touched on each injection.

**Verdict:** Mergeable — pure deletion, no behaviour to regress.

---

Okuro3499 (approved):

PR #17018, opened 2026-09-10
life: smoother health examination item encrypting (fixes #17003)

### LINKED ISSUE #17003: Stop re-decrypting health examinations on the main thread
## What

The examination payload is decrypted twice: once off the main thread during list construction, then again synchronously on the main thread on every tap.

`app/src/main/java/org/ole/planet/myplanet/ui/health/HealthExaminationAdapter.kt`

## Why

`submitExaminations` already decrypts each row inside `withContext(dispatcherProvider.default)`:

```kotlin
val encrypted = userModel?.let { user -> item.getEncryptedDataAsJson(user) }   // ~62
```

but it throws the result away and keeps only a boolean:

```kotlin
hasEncryptedData = encrypted != null                                          // ~80
```

The click handler then redoes the whole thing — an AES decrypt plus a Gson parse — on the main thread, once per tap:

```kotlin
holder.itemView.setOnClickListener {
    if (item.hasEncryptedData) {
        val encrypted = userModel?.let { user -> item.examination.getEncryptedDataAsJson(user) } ?: JsonObject()
        showAlert(binding, item, encrypted)
    }
}
```

## The fix

- Change `HealthExaminationItem.hasEncryptedData: Boolean` (field at ~42) to `encrypted: JsonObject?`.
- Populate it from the value already computed in `submitExaminations`.
- Have the click handler pass `item.encrypted` straight to `showAlert` (guard on non-null instead of the boolean). `showAlert`/`showEncryptedData` already take a `JsonObject`.
- Leave `DIFF_CALLBACK` comparing exactly what it compares today.

No functional change: same rows, same dialog contents.

## Acceptance

- Tapping a row with encrypted data opens the same dialog with the same values.
- Tapping a row without encrypted data still does nothing.
- No decrypt or JSON parse remains on the main thread in this file.

## Notes

~12 lines, 1 file. No test references this adapter. Conventions: `docs/CODE_STYLE_GUIDE.md`.

dogi:
Updated `HealthExaminationItem` to store `encrypted: JsonObject?` computed off the main thread during `submitExaminations`, eliminating synchronous re-decryption and JSON parsing on the main thread when tapping items in `HealthExaminationAdapter`.

Fixes #17003

---
*PR created automatically by Jules for task [4186953270365790712](https://jules.google.com/task/4186953270365790712) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 7064
-        versionName = "0.70.64"
+        versionCode = 7065
+        versionName = "0.70.65"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/ui/health/HealthExaminationAdapter.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/health/HealthExaminationAdapter.kt
@@ -39,7 +39,7 @@ class HealthExaminationAdapter(
         val formattedDate: String,
         val isSelfExamination: Boolean,
         val resolvedName: String,
-        val hasEncryptedData: Boolean
+        val encrypted: JsonObject?
     )
 
     private val colorGrey50 by lazy { ContextCompat.getColor(context, R.color.md_grey_50) }
@@ -77,7 +77,7 @@ class HealthExaminationAdapter(
                     formattedDate = formattedDate,
                     isSelfExamination = isSelfExamination,
                     resolvedName = resolvedName,
-                    hasEncryptedData = encrypted != null
+                    encrypted = encrypted
                 )
             }
         }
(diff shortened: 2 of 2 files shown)


dogi:
**Review — rating: 86/100**

Widens `HealthExaminationItem.hasEncryptedData: Boolean` to `encrypted: JsonObject?` so the click handler can use the already-decrypted payload instead of decrypting again.

The main-thread claim checks out. `submitExaminations` builds the item list inside `withContext(dispatcherProvider.default)`, and that block already called `item.getEncryptedDataAsJson(user)` — it just threw the result away after reducing it to a boolean. The old click listener then re-ran `getEncryptedDataAsJson` on the main thread, so every tap paid for a PBKDF2-backed decrypt on the UI thread. Carrying the object through removes that entirely and drops the redundant second decrypt per item as well; the work now happens exactly once, on `default`.

The rewrite is also behaviour-preserving rather than just faster. `hasEncryptedData` was defined as `encrypted != null` over the same value now stored, so `item.encrypted?.let { }` gates on precisely the old condition. The `?: JsonObject()` fallback that disappears was dead code — it could only fire if a second decrypt of the same row returned null after the first returned non-null.

**Nit:** the decrypted `JsonObject` is now retained for every row for as long as the list is bound, rather than existing transiently per tap. Not a new exposure (it was already in memory during mapping) but it's a longer-lived window for health data, so it's worth a moment's thought if this list ever gets large.

**Verdict:** Mergeable — removes real main-thread crypto and a duplicate decrypt without changing what the user sees.

---

Okuro3499 (approved):

PR #17019, opened 2026-09-10
teams: smoother notifications repository counting (fixes #17011)

### LINKED ISSUE #17011: Replace the sequential per-team chat count with one concurrent fetch
## What

The dashboard team-badge refresh counts chats one team at a time, sequentially.

`app/src/main/java/org/ole/planet/myplanet/repository/NotificationsRepositoryImpl.kt` (~315-317)

## Why

```kotlin
val chatCountsById = mutableMapOf<String, Long>()
for (teamId in notificationsById.keys) {
}
```

One suspend query per team, serialized, on every refresh.

## The fix

Replace the loop with a concurrent fetch:

```kotlin
val chatCountsById = coroutineScope {
    notificationsById.keys.map { teamId ->
        async { teamId to voicesRepository.countTopLevelByTeam(teamId) }
    }.awaitAll().toMap()
}
```

- Keep the existing `hasTask`/`hasChat` logic and the empty-input early return exactly as they are.
- Add no new methods to `VoicesRepository`.

## Acceptance

- Same counts, same badge behaviour, empty input still short-circuits.
- Queries for distinct teams are issued concurrently.

## Notes

An aggregate `GROUP BY` would be better still, but `NewsDao`'s `viewIn LIKE` matching makes a correct one non-trivial; parallelising is the safe version of the same win. Conventions: `docs/CODE_STYLE_GUIDE.md`.

dogi:
Replaced the sequential per-team chat count loop in `NotificationsRepositoryImpl.getTeamNotifications` with a concurrent fetch using `coroutineScope`, `async`, and `awaitAll`. Updated `NotificationsRepositoryImplTest` to cover concurrent fetching across multiple teams.

Fixes #17011

---
*PR created automatically by Jules for task [17398082137068262596](https://jules.google.com/task/17398082137068262596) started by @dogi*

## Summary by CodeRabbit

- **Performance**
  - Team notification updates now retrieve chat activity counts in parallel, improving responsiveness when monitoring multiple teams.

- **Bug Fixes**
  - Empty team selections now return cleanly without unnecessary data requests.
  - Chat indicators now accurately reflect whether new activity is available compared with the latest notification count.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 7065
-        versionName = "0.70.65"
+        versionCode = 7066
+        versionName = "0.70.66"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/repository/NotificationsRepositoryImpl.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/NotificationsRepositoryImpl.kt
@@ -9,6 +9,9 @@ import java.util.LinkedHashSet
 import java.util.Locale
 import java.util.UUID
 import javax.inject.Inject
+import kotlinx.coroutines.async
+import kotlinx.coroutines.awaitAll
+import kotlinx.coroutines.coroutineScope
 import org.ole.planet.myplanet.data.room.dao.NotificationDao
 import org.ole.planet.myplanet.data.room.dao.TeamNotificationDao
 import org.ole.planet.myplanet.data.room.dao.TeamTaskDao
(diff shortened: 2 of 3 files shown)


dogi:
**Review — rating: 82/100**

Replaces the sequential `for (teamId in notificationsById.keys)` watermark loop in `getTeamNotifications` with `coroutineScope { keys.map { async { it to countTopLevelByTeam(it) } }.awaitAll().toMap() }`. I checked the semantics hold: `awaitAll` preserves list order, so the resulting `LinkedHashMap` iterates in the same order the old `mutableMapOf` was populated in, and failure behaviour is unchanged — the old loop let the first exception propagate, and `coroutineScope` does the same after cancelling its siblings. `async` inherits the caller's context rather than hardcoding a dispatcher, which is the right call here since `countTopLevelByTeam` bottoms out in a `suspend` Room query that dispatches on Room's own executor, so nothing lands on the caller's thread.

I also confirmed this does **not** collide with #17053 despite both editing `NotificationsRepositoryImpl.kt` — that one rewrites `updateTeamNotification` (line ~282), this one rewrites the read path in `getTeamNotifications` (line ~315). Different functions, different concerns, no overlapping hunks.

**Nit:** `getTeamNotifications fetches chat counts concurrently for multiple tracked teams` doesn't actually assert concurrency — `coVerify(exactly = 1)` per team passes identically against the old sequential loop. It's a fine regression test for the mapping, but the name overpromises. The fan-out is also unbounded in principle; harmless at realistic team counts, and Room's executor caps real parallelism anyway.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17021, opened 2026-09-10
resources: smoother viewer pdf rendering (fixes #17004)

### LINKED ISSUE #17004: Move the full-page PDF render off the main thread and stop leaking the descriptor
## What

`renderPdf()` decodes and renders a full-size PDF page synchronously on the main thread, and leaks its native handles on any exception.

`app/src/main/java/org/ole/planet/myplanet/ui/viewer/ResourceViewerFragment.kt` (`renderPdf`, ~509-535)

## Why

```kotlin
val fileDescriptor = ParcelFileDescriptor.open(file, ParcelFileDescriptor.MODE_READ_ONLY)
val pdfRenderer = PdfRenderer(fileDescriptor)
val page = pdfRenderer.openPage(0)
val bitmap = createBitmap(page.width, page.height)
page.render(bitmap, null, null, PdfRenderer.Page.RENDER_MODE_FOR_DISPLAY)
...
page.close()
pdfRenderer.close()
fileDescriptor.close()
```

All of it runs on the main thread, and the three `close()` calls sit *after* the view mutations with no `finally`/`use`, so any exception in between leaks the descriptor and the renderer. `dispatcherProvider` is already injected at ~105 and `extractPdfText()` right below already uses the `viewLifecycleOwner.lifecycleScope` + `withContext(dispatcherProvider.io)` pattern.

## The fix

- Wrap **only** the open/decode/render in `viewLifecycleOwner.lifecycleScope.launch { withContext(dispatcherProvider.io) { … } }`, returning the bitmap.
- Keep every view mutation (placeholder `GONE`, `ImageView` creation, `parent.addView`, layout params) on the main thread, after the bitmap comes back.
- Convert the manual `close()` calls to `.use {}` so the descriptor, renderer and page are released on the failure path too.
- Replace `catch (e: Exception) { e.printStackTrace() }` with a tagged `Log.e(TAG, "<static message>", e)`. Do **not** interpolate the file path, any URL, or any other value into the log message — release builds keep logs.

Keep today's failure behaviour: on any error the placeholder simply stays visible. No new error UI.

## Acceptance

- A PDF resource still renders its first page identically.
- A missing/corrupt PDF leaves the placeholder visible and logs once; no descriptor leak.
- No `PdfRenderer` work remains on the main thread.

## Notes

PR #17000 also touches this file, but only the binding boilerplate (imports, class header, `onCreateView`, `onDestroyView` — hunks at lines 11, 30, 61, 76, 159, 696). Your change is inside `renderPdf` at ~509, so it should not collide. Conventions: `docs/CODE_STYLE_GUIDE.md`.

dogi:
Moves PDF rendering off the main thread onto I/O dispatcher using coroutines, safely closes native PDF handles using .use blocks on success or failure, replaces e.printStackTrace() with sanitized tagged logging, and keeps view mutations on the main thread.

Fixes #17004

---
*PR created automatically by Jules for task [16637416671500816998](https://jules.google.com/task/16637416671500816998) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 7070
-        versionName = "0.70.70"
+        versionCode = 7071
+        versionName = "0.70.71"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/ui/viewer/ResourceViewerFragment.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/viewer/ResourceViewerFragment.kt
@@ -10,6 +10,7 @@ import android.media.AudioManager
 import android.os.Bundle
 import android.os.ParcelFileDescriptor
 import android.text.TextUtils
+import android.util.Log
 import android.util.Rational
 import android.view.LayoutInflater
 import android.view.Menu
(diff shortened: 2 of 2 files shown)


dogi:
**Review — rating: 88/100**

Moves the first-page PDF render onto `dispatcherProvider.io` and wraps the three native handles in `use { }`.

This fixes two distinct problems, and the second one is the more interesting of the pair. On master the `close()` calls sat at the *end* of the `try` block, after `page.render(...)` and all the view manipulation — so any throw between opening the descriptor and reaching those lines skipped every close and leaked the `ParcelFileDescriptor`, the `PdfRenderer` and the `Page`, with nothing but `e.printStackTrace()` to show for it. The nested `use { }` blocks close in reverse order on every path. `PdfRenderer` and `PdfRenderer.Page` are both `AutoCloseable` and `ParcelFileDescriptor` is `Closeable`, all well below this project's minSdk 26, so the idiom is available. `@Inject lateinit var dispatcherProvider` is already on the fragment (line 106) and used the same way at line 592, so the threading follows the file's existing pattern rather than introducing one.

The main-thread win is real too — `PdfRenderer.openPage` plus `render` into a full-page bitmap is exactly the kind of work that shows up as a dropped-frame cluster on resource open.

The result handling is safe: the coroutine runs on `viewLifecycleOwner.lifecycleScope`, so it's cancelled at `onDestroyView` before `binding` can go stale, and the `bitmap != null` check means a failed render now leaves the placeholder visible instead of silently half-mutating the view hierarchy. Swapping `printStackTrace()` for `Log.e(TAG, ...)` is the right call for something that now fails asynchronously.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17027, opened 2026-09-10
sync: less upload manager submissions repository is more (fixes #17008)

### LINKED ISSUE #17008: Remove the unused SubmissionsRepository edge from UploadManager
## What

`UploadManager` takes a `SubmissionsRepository` it never calls.

`app/src/main/java/org/ole/planet/myplanet/services/UploadManager.kt`

## Why

The class mentions the type exactly twice:

```kotlin
import org.ole.planet.myplanet.repository.SubmissionsRepository   // 27
...
private val submissionsRepository: SubmissionsRepository,         // 56
```

Nothing in the class body uses it. Photo submission uploads go through the separately injected `photoUploader`, which holds its own.

## The fix

- Delete the constructor parameter and the import.
- Drop the matching positional argument from `UploadManagerTest`. Keep the test's own `submissionsRepository` field — `PhotoUploader(...)` still needs it.

## Acceptance

- The app compiles and Hilt still builds `UploadManager`.

## Notes

~3 lines. Makes the upload orchestrator stop looking coupled to submissions. Conventions: `docs/CODE_STYLE_GUIDE.md`.

dogi:
Removed unused SubmissionsRepository import and constructor parameter from UploadManager, and updated UploadManagerTest accordingly.

Fixes #17008

---
*PR created automatically by Jules for task [6272685795489733458](https://jules.google.com/task/6272685795489733458) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 7071
-        versionName = "0.70.71"
+        versionCode = 7072
+        versionName = "0.70.72"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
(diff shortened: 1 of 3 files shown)


dogi:
**Review — rating: 87/100**

Drops the unused `submissionsRepository` constructor parameter from `UploadManager`, its import, and the corresponding argument in `UploadManagerTest`. Three deletions, nothing else.

`UploadManager` is a `@Singleton`, so this removes a permanently-live reference to a repository it never called — submission uploads already go through `UploadCoordinator` and `uploadRepository`, which is why nothing else had to move. Build (default + lite) and both test shards passing is the proof the field really was dead.

**Nit:** worth a quick check that the now-unused `submissionsRepository` mock declaration higher up in `UploadManagerTest` went with it — if it's still declared it's only a compiler warning, but it's the kind of leftover that makes the next reader wonder what it's for.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17028, opened 2026-09-10
courses: smoother progress vew holding (fixes #17006)

### LINKED ISSUE #17006: Hoist the per-bind click listener out of CoursesProgressAdapter (fixes wrong-course navigation)
## What

`CoursesProgressAdapter` attaches its row click listener inside `onBindViewHolder`, and only on the branch where the row has progress. A recycled holder that lands on a row *without* progress keeps the previous row's listener, so tapping it opens `CourseProgressActivity` for the wrong course.

`app/src/main/java/org/ole/planet/myplanet/ui/courses/CoursesProgressAdapter.kt`

## Why

```kotlin
override fun onBindViewHolder(holder: CoursesProgressViewHolder, position: Int) {   // 28
    val item = getItem(position)
    holder.binding.tvTitle.text = item.courseName
    if (item.progressCurrent != null && item.progressMax != null) {
        holder.binding.tvDescription.text = context.getString(R.string.step_progress, item.progressCurrent, item.progressMax)
        holder.itemView.setOnClickListener {
            context.startActivity(Intent(context, CourseProgressActivity::class.java).putExtra("courseId", item.courseId))
        }
    }
    ...
```

A fresh lambda is allocated on every bind, the listener is never cleared on the else path, and it captures the `item` from the bind that installed it.

## The fix

- Move the click handling into `CoursesProgressViewHolder.init`.
- Resolve the row there via `bindingAdapterPosition`, bailing out on `RecyclerView.NO_POSITION`.
- Start `CourseProgressActivity` only when that row's `progressCurrent` and `progressMax` are both non-null.
- Delete the `setOnClickListener` from `onBindViewHolder`.

## Acceptance

- Tapping a row with progress opens `CourseProgressActivity` for **that** row's `courseId`.
- Tapping a row without progress does nothing, including after scrolling recycles holders.
- No listener is allocated per bind.

## Notes

A correctness fix, not just an allocation win. Conventions: `docs/CODE_STYLE_GUIDE.md`.

dogi:
Move click listener initialization into CoursesProgressViewHolder.init using bindingAdapterPosition to resolve the item dynamically and check for non-null progress values. This removes per-bind listener allocations and fixes wrong-course navigation caused by recycled ViewHolders retaining previous listeners. Unit tests have been added to verify row clicks.

Fixes #17006

---
*PR created automatically by Jules for task [18317461799797761292](https://jules.google.com/task/18317461799797761292) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 7077
-        versionName = "0.70.77"
+        versionCode = 7078
+        versionName = "0.70.78"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/ui/courses/CoursesProgressAdapter.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/courses/CoursesProgressAdapter.kt
@@ -30,9 +30,6 @@ class CoursesProgressAdapter(private val context: Context) : ListAdapter<Courses
         holder.binding.tvTitle.text = item.courseName
         if (item.progressCurrent != null && item.progressMax != null) {
             holder.binding.tvDescription.text = context.getString(R.string.step_progress, item.progressCurrent, item.progressMax)
-            holder.itemView.setOnClickListener {
-                context.startActivity(Intent(context, CourseProgressActivity::class.java).putExtra("courseId", item.courseId))
-            }
         }
         if (item.mistakes != null) holder.binding.tvTotal.text = item.mistakes.toString()
         else holder.binding.tvTotal.text = context.getString(R.string.message_placeholder, "0")
@@ -100,6 +97,22 @@ class CoursesProgressAdapter(private val context: Context) : ListAdapter<Courses
         val tvTitle = binding.tvTitle
         val tvTotal = binding.tvTotal
         val tvDescription = binding.tvDescription
+
+        init {
+            itemView.setOnClickListener {
+                val position = bindingAdapterPosition
+                if (position != RecyclerView.NO_POSITION) {
+                    val item = getItem(position)
+                    if (item.progressCurrent != null && item.progressMax != null) {
+                        context.startActivity(
+                            Intent(context, CourseProgressActivity::class.java)
+                                .addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
+                                .putExtra("courseId", item.courseId)
+                        )
+                    }
+                }
+            }
+        }
     }
 
     companion object {
--- a/app/src/main/java/org/ole/planet/myplanet/ui/courses/CoursesProgressAdapter.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/courses/CoursesProgressAdapter.kt
@@ -104,13 +104,7 @@ class CoursesProgressAdapter(private val context: Context) : ListAdapter<Courses
                 if (position != RecyclerView.NO_POSITION) {
                     val item = getItem(position)
                     if (item.progressCurrent != null && item.progressMax != null) {
-                        context.startActivity(
-                            Intent(context, CourseProgressActivity::class.java)
-                                // Defensive: allows launching if a non-Activity context is ever passed.
-                                // From an Activity context this flag alone does not start a new task.
-                                .addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
-                                .putExtra("courseId", item.courseId)
-                        )
+                        context.startActivity(Intent(context, CourseProgressActivity::class.java).putExtra("courseId", item.courseId))
                     }
                 }
             }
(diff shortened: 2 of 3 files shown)


dogi:
**Review — rating: 89/100**

Framed as hoisting an allocation out of `onBindViewHolder`, but it also fixes a stale-click bug, and that's the stronger reason to take it.

On master the listener was installed *inside* `if (item.progressCurrent != null && item.progressMax != null)`. Because `ListAdapter` recycles view holders, a holder that had been bound to a row **with** progress kept that listener — and its captured `item` — when it was later rebound to a row **without** progress. Tapping the no-progress row then launched `CourseProgressActivity` with the previous course's `courseId`. Nothing in the bind path ever cleared it, since the `else` branch didn't exist.

Installing once in `init` and resolving `getItem(bindingAdapterPosition)` at click time removes both halves of that: no captured item to go stale, and the `progressCurrent != null && progressMax != null` check now runs against whatever row is actually under the finger. The `bindingAdapterPosition != RecyclerView.NO_POSITION` guard covers the mid-animation case.

The tests are what make this convincing — `recycled holder from row with progress to row without progress does nothing on click` is aimed squarely at the bug, and the positive case asserts the extra actually carries `"course123"` rather than just that *some* activity started.

**Nit:** `FLAG_ACTIVITY_NEW_TASK` is new and unexplained. It's harmless when `context` is an `Activity` (the flag alone doesn't fork a task at matching affinity), so I assume it's defensive in case an application context is passed — worth a comment saying so, otherwise the next reader will wonder whether the launch semantics were meant to change.

**Verdict:** Mergeable.

---

Okuro3499 (dismissed):
```
File: CoursesProgressAdapter.kt:109, Test: CoursesProgressAdapterTest.kt (all Activity-launch tests)                                                                             
    Problem: FLAG_ACTIVITY_NEW_TASK was added to production code, undocumented, apparently only to satisfy the test's Application context (which requires the flag to call           
    startActivity without crashing). This means the test isn't actually exercising the same context type used in production (requireActivity()), and a production intent flag was    
    changed to accommodate a test shortcut rather than a real requirement
```

dogi:
**Re-review — Rating: 90/100** (was 89)

The only thing outstanding from my last pass was that `FLAG_ACTIVITY_NEW_TASK` appeared without explanation, and that's now answered in place:

```kotlin
// Defensive: allows launching if a non-Activity context is ever passed.
// From an Activity context this flag alone does not start a new task.
```

That's exactly the right note — it records both the intent and the reason it's harmless today, so the next reader doesn't have to work out whether launch semantics were meant to change.

Everything else is unchanged and still correct. The substantive fix stands: installing the click listener once in `init` and resolving `getItem(bindingAdapterPosition)` at click time eliminates the stale-capture bug where a recycled holder previously bound to a row *with* progress kept its old listener — and its old `courseId` — when rebound to a row *without* progress, launching `CourseProgressActivity` for the wrong course. The `RecyclerView.NO_POSITION` guard and the re-checked `progressCurrent != null && progressMax != null` condition are both still in place, and the three Robolectric tests (including the recycling case) still cover it.

Build (default + lite) and both test shards green.

**Verdict:** Mergeable.

---

dogi:
**Re-review — Rating: 91/100**

Moves the click listener from `onBindViewHolder` into the ViewHolder's `init`. The stated goal is avoiding a lambda allocation per bind, which it achieves — but the more valuable part is a latent bug it fixes, which the description doesn't mention.

**The bug that was there**

The old code installed the listener only inside `if (item.progressCurrent != null && item.progressMax != null)`, and never cleared it otherwise. Since RecyclerView reuses ViewHolders, a holder that had been bound to a row *with* progress kept its listener when rebound to a row *without* progress — and that listener captured the old `item` by closure. Tapping a no-progress row would launch `CourseProgressActivity` with the previous row's `courseId`. Wrong screen, wrong course, no error.

The new listener resolves the item at click time via `bindingAdapterPosition` and re-checks the progress condition there, so a stale capture is structurally impossible:

```kotlin
val position = bindingAdapterPosition
if (position != RecyclerView.NO_POSITION) {
    val item = getItem(position)
    if (item.progressCurrent != null && item.progressMax != null) { ... }
}
```

`bindingAdapterPosition` rather than `adapterPosition`, with the `NO_POSITION` guard, is the correct pairing — it's the one that's valid during and after animations and returns `NO_POSITION` for a detached or mid-update holder.

**The tests are the best part of this PR**

Three new cases, and the third is the one that matters:

- `click row with progress launches CourseProgressActivity` — asserts the component class *and* the `courseId` extra, so it would catch a wrong-course launch rather than just "something started".
- `click row without progress does nothing` — `assertNull(nextStartedActivity)`.
- `recycled holder from row with progress to row without progress does nothing on click` — this is a direct regression test for the bug above.

They're built properly too: a real `RecyclerView` with a `LinearLayoutManager`, `measure` and `layout` called so holders actually exist, then `findViewHolderForAdapterPosition`. Swapping `ApplicationProvider.getApplicationContext()` for a Robolectric-built `AppCompatActivity` is what makes `shadowOf(activity).nextStartedActivity` available, and it's a strictly better harness for an adapter that starts activities.

**Small note:** the per-bind saving is a lambda allocation, not a measurable frame cost, so I'd lead the PR description with the recycled-listener fix instead — it's the reason this should merge.

**Verdict:** Mergeable.

---

PR #17029, opened 2026-09-10
courses: smoother repository step exam bulk querying (fixes #17007)

### LINKED ISSUE #17007: Use one bulk exam lookup for per-step question counts
## What

Course step construction issues one `LIMIT 1` exam query per step, when a bulk DAO query already exists.

Files:
- `app/src/main/java/org/ole/planet/myplanet/repository/CoursesRepositoryImpl.kt` (~142)
- `app/src/main/java/org/ole/planet/myplanet/repository/SubmissionsRepositoryImpl.kt` (~162)
- `app/src/main/java/org/ole/planet/myplanet/repository/SubmissionsRepository.kt` (~26)

## Why

```kotlin
val steps = rawSteps.map { step ->
    val count = step.id.let { submissionsRepository.getExamQuestionCount(it) }   // 142
```

and

```kotlin
override suspend fun getExamQuestionCount(stepId: String): Int {                 // 162
    return examDao.getFirstByStepId(stepId)?.noOfQuestions ?: 0
}
```

One query per step, and `CoursesRepositoryImpl.kt:142` is the only production caller of `getExamQuestionCount` anywhere. A bulk query already exists:

```kotlin
@Query("SELECT * FROM exams WHERE stepId IN (:stepIds)") suspend fun getByStepIds(stepIds: List<String>): List<StepExam>   // ExamDao.kt:19
```

## The fix

- In `CoursesRepositoryImpl`, fetch `examDao.getByStepIds(rawSteps.map { it.id })` once (skip the query when the id list is empty).
- Build a `stepId -> noOfQuestions` map where the **first** row for a step wins, matching today's `LIMIT 1` semantics.
- Map the steps against it, defaulting to `0` for a step with no exam row.
- Delete `getExamQuestionCount` from `SubmissionsRepository` and `SubmissionsRepositoryImpl`.
- `CoursesRepositoryImplTest` stubs `submissionsRepository.getExamQuestionCount("step_1")` (~378) — update the test to stub the DAO instead, keeping the same expected counts.

`CoursesRepositoryImpl` will need `ExamDao` injected if it does not already have it; follow how the other DAOs are injected there.

## Acceptance

- Same `questionCount` per step as today, including `0` for steps with no exam.
- One exam query per course-step load instead of one per step.
- `getExamQuestionCount` no longer exists on the interface or the impl; `testDefaultDebugUnitTest` passes.

## Notes

Also removes the only per-step suspension from step-item construction. This file has other open work at ~830 (map/filter collapse) — your change is at ~142. Conventions: `docs/CODE_STYLE_GUIDE.md`.

dogi:
Replaces single LIMIT 1 per-step exam question count queries in CoursesRepositoryImpl with a single bulk lookup via examDao.getByStepIds, and removes getExamQuestionCount from SubmissionsRepository and SubmissionsRepositoryImpl.

Fixes #17007

---
*PR created automatically by Jules for task [16634147295499682935](https://jules.google.com/task/16634147295499682935) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 7072
-        versionName = "0.70.72"
+        versionCode = 7073
+        versionName = "0.70.73"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/repository/CoursesRepositoryImpl.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/CoursesRepositoryImpl.kt
@@ -138,12 +138,26 @@ class CoursesRepositoryImpl @Inject constructor(
             val downloadedResources = getCourseOfflineResources(courseId)
             val rawSteps = getCourseSteps(courseId)
 
+            val stepIds = rawSteps.mapNotNull { it.id }
+            val questionCountsByStepId = if (stepIds.isEmpty()) {
+                emptyMap()
+            } else {
+                val exams = examDao.getByStepIds(stepIds)
+                val countsMap = mutableMapOf<String, Int>()
+                exams.forEach { exam ->
+                    val sId = exam.stepId
+                    if (sId != null && !countsMap.containsKey(sId)) {
+                        countsMap[sId] = exam.noOfQuestions
+                    }
+                }
+                countsMap
+            }
+
             val steps = rawSteps.map { step ->
-                val count = step.id.let { submissionsRepository.getExamQuestionCount(it) }
                 StepItem(
                     id = step.id,
                     stepTitle = step.stepTitle,
-                    questionCount = count
+                    questionCount = questionCountsByStepId[step.id] ?: 0
                 )
             }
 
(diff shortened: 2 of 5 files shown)


dogi:
**Review — rating: 86/100**

Collapses a per-step `submissionsRepository.getExamQuestionCount(stepId)` call into one `examDao.getByStepIds(stepIds)`, then indexes the result. Straight N+1 elimination on course-detail load — a 20-step course goes from 20 queries to 1.

The supporting pieces were already in place, which keeps the diff honest: `ExamDao.getByStepIds` exists (`ExamDao.kt:19`) and `examDao` is already a constructor dependency of `CoursesRepositoryImpl` (line 63, used again at 213), so nothing new had to be wired.

Semantics hold up. The old path was `examDao.getFirstByStepId(stepId)?.noOfQuestions ?: 0`; the new `if (sId != null && !countsMap.containsKey(sId))` keeps the first row seen per step and `questionCountsByStepId[step.id] ?: 0` preserves the zero default for steps with no exam. The `stepIds.isEmpty()` guard also avoids emitting a `WHERE stepId IN ()`.

I'll also note this is the narrowing that #17016 claims and doesn't deliver: `getExamQuestionCount` comes off the `SubmissionsRepository` interface *and* out of the impl, because this was its only caller.

**Nit:** "first row wins" is only well-defined because neither query orders its results — if a step ever had two exam rows, `getFirstByStepId`'s pick and this map's pick could differ. In practice it's one exam per step, so this is a note rather than a defect.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17030, opened 2026-09-10
enterprises: smoother finances transactions view modelling (fixes #17013)

### LINKED ISSUE #17013: Stop double-querying team transactions on every finances screen entry
## What

Entering the enterprises finances screen starts the transactions query, then immediately cancels it and starts the identical query again.

Files:
- `app/src/main/java/org/ole/planet/myplanet/ui/enterprises/EnterprisesFinancesFragment.kt`
- `app/src/main/java/org/ole/planet/myplanet/ui/enterprises/EnterprisesFinancesViewModel.kt`

## Why

`observeTransactions()` is called from `onViewCreated` (~273) and again from `onResume` (~356) with the same arguments, and the ViewModel unconditionally cancels and relaunches:

```kotlin
fun getTeamTransactions(teamId: String, sortAscending: Boolean, startDate: Long?, endDate: Long?) {
    transactionsJob?.cancel()
    transactionsJob = viewModelScope.launch {
        teamsRepository.getTeamTransactionsWithBalance(...).collectLatest { results -> _transactions.value = results }
    }
}
```

So the first display cancels the in-flight collection and re-runs `getTeamTransactionsWithBalance` from scratch.

## The fix

In `EnterprisesFinancesViewModel`:

- Store the last-requested `(teamId, sortAscending, startDate, endDate)` tuple.
- Return early from `getTeamTransactions` when the incoming arguments match it **and** `transactionsJob?.isActive == true`.
- Leave the `onPause` / `resetFilterAndSort` path alone — it legitimately changes the arguments, so it must still relaunch.

Keep the `onResume` hook in the fragment: it is the recovery path when the flow was cancelled while stopped. Only make the duplicate call cheap.

## Acceptance

- Entering the screen runs the transactions query once.
- Changing sort or the date filter still re-queries.
- Backgrounding and returning still restores the list.

## Notes

dogi:
Prevents EnterprisesFinancesViewModel from cancelling and restarting an active transaction query when identical query arguments are provided (such as when entering EnterprisesFinancesFragment where observeTransactions is invoked in both onViewCreated and onResume).

Fixes #17013

---
*PR created automatically by Jules for task [11203593152205880366](https://jules.google.com/task/11203593152205880366) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 7073
-        versionName = "0.70.73"
+        versionCode = 7074
+        versionName = "0.70.74"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/ui/enterprises/EnterprisesFinancesViewModel.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/enterprises/EnterprisesFinancesViewModel.kt
@@ -66,6 +66,14 @@ class EnterprisesFinancesViewModel @Inject constructor(
     private val _transactionCreated = MutableSharedFlow<Result<Unit>>(extraBufferCapacity = 1)
     val transactionCreated: SharedFlow<Result<Unit>> = _transactionCreated.asSharedFlow()
 
+    private data class TransactionQuery(
+        val teamId: String,
+        val sortAscending: Boolean,
+        val startDate: Long?,
+        val endDate: Long?
+    )
+
+    private var lastQuery: TransactionQuery? = null
     private var transactionsJob: Job? = null
 
     fun getTeamTransactions(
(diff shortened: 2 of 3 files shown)


dogi:
**Review — rating: 87/100**

Adds a `TransactionQuery` value object and skips the re-query when the same four arguments arrive while `transactionsJob` is still active.

The guard condition is the part worth getting right, and it is. `lastQuery == query && transactionsJob?.isActive == true` deduplicates only while a collection is genuinely in flight; once the job ends, an identical call still re-queries, so this can't wedge the screen into serving a stale result. And because `getTeamTransactionsWithBalance` returns a `Flow` that the live job is already collecting, suppressing the duplicate loses nothing — fresh data still arrives through the existing collection. Different arguments fall through to `transactionsJob?.cancel()` and relaunch exactly as before, which the second test pins by asserting one call each for `sortAscending = false` and `true`.

Using a `data class` for the key rather than concatenating a string is the right instinct too — nullable `startDate`/`endDate` compare correctly without any sentinel encoding.

Both new tests earn their place: the first would fail on master (two calls instead of one), and the second guards against the obvious over-correction of caching too aggressively.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17035, opened 2026-09-10
teams: smoother resources view modelling (fixes #17010)

### LINKED ISSUE #17010: Stop double-loading team resources on every screen entry, and cancel the superseded load
## What

Entering the team resources screen runs the whole load twice, and nothing cancels the first run, so two jobs race to write the UI state.

Files:
- `app/src/main/java/org/ole/planet/myplanet/ui/teams/resources/TeamResourcesFragment.kt`
- `app/src/main/java/org/ole/planet/myplanet/ui/teams/resources/TeamResourcesViewModel.kt`

## Why

`showLibraryList()` is called from `onViewCreated` (~49) and again from `onResume` (~63); it forwards to `TeamResourcesViewModel.loadResources` (~32), an unguarded `viewModelScope.launch` that runs `getTeamResources` and `isTeamLeader` concurrently:

```kotlin
fun loadResources(teamId: String, userId: String?) {
    viewModelScope.launch {
        coroutineScope {
            val librariesDeferred = async { teamsRepository.getTeamResources(teamId) }
            val canRemoveDeferred = async { teamsRepository.isTeamLeader(teamId, userId) }
            _uiState.value = TeamResourcesUiState(...)
        }
    }
}
```

Four queries per entry, and since nothing cancels the earlier launch the adapter can be handed a stale list.

## The fix

In `TeamResourcesViewModel`:

- Hold the load in a `private var loadJob: Job?` and `cancel()` it at the top of `loadResources`.
- Skip the relaunch entirely when `(teamId, userId)` matches the last request **and** that job is still active.
- Add a `reload()` (or a `force` flag) that bypasses the guard, and route the explicit post-add/post-remove refreshes in the fragment (~126, ~150 — the add-resources dialog callback and the `AddResourceFragment` detach callback) through it, so an actual mutation always re-queries.

While in this file: `android.util.Log` in `recordActivitySafely` (import at :3) is the ViewModel's only `android.*` dependency — replacing that one call with `runCatching {}` (no log) makes the class platform-free. Keep the swallow-and-continue behaviour either way.

## Acceptance

- Entering the screen once issues one set of queries, not two.
- Adding or removing a team resource still refreshes the list.
- The adapter never receives a result from a superseded load.

## Notes

dogi:
Fix double-loading team resources on screen entry by tracking active load jobs, deduplicating identical requests, cancelling superseded jobs, adding reload support for resource mutations, and removing android.* imports from TeamResourcesViewModel.

Fixes #17010

---
*PR created automatically by Jules for task [8507509632764662975](https://jules.google.com/task/8507509632764662975) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 7074
-        versionName = "0.70.74"
+        versionCode = 7075
+        versionName = "0.70.75"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/ui/teams/resources/TeamResourcesFragment.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/teams/resources/TeamResourcesFragment.kt
@@ -69,9 +69,13 @@ class TeamResourcesFragment : BaseTeamFragment(), OnTeamPageListener, OnResource
         llImage?.removeAllViews()
     }
 
-    private fun showLibraryList() {
+    private fun showLibraryList(force: Boolean = false) {
         if (!isAdded || activity == null) return
-        viewModel.loadResources(teamId, user?.id)
+        if (force) {
+            viewModel.reload(teamId, user?.id)
+        } else {
+            viewModel.loadResources(teamId, user?.id)
+        }
     }
 
     private fun renderResources(state: TeamResourcesUiState) {
@@ -123,7 +127,7 @@ class TeamResourcesFragment : BaseTeamFragment(), OnTeamPageListener, OnResource
                         }
                     viewLifecycleOwner.lifecycleScope.launch {
                         viewModel.addResources(teamId, selectedResources, user?.id)
-                        showLibraryList()
+                        showLibraryList(force = true)
                     }
                 }
                 .setNeutralButton(R.string.create_new_resource) { _: DialogInterface?, _: Int ->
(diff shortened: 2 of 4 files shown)


dogi:
**Review — rating: 84/100**

Deduplicates in-flight loads on `(teamId, userId)` and cancels superseded ones, with an explicit `force`/`reload` escape hatch.

The `force` flag is what makes this correct rather than merely faster, and it's the part these dedupe patches usually get wrong. Both mutating call sites — after `addResources`, and the `onFragmentDetached` callback once the add-resource fragment closes — pass `force = true`, so the guard can't swallow the refresh that must follow a write. Dropping it would have produced a screen that silently fails to show a resource the user just added.

The cancellation test is the one I'd have asked for: `loadResources for different team cancels superseded loadJob` gives team1 a 200 ms delay and team2 50 ms, then asserts the final `uiState` is team2's — so a slow earlier load can no longer land on top of a newer one. The dedupe test correctly asserts `exactly = 1` on *both* repository calls rather than just one.

**Nit — the one thing I'd push back on.** `recordActivitySafely` loses its `.onFailure { Log.w(...) }` and is now a bare `runCatching { }`, so a failed team-activity record is completely invisible: no log, no state, nothing. Reading the test diff, the motivation looks like shedding the `mockkStatic(Log::class)` scaffolding in `setup`/`tearDown` — which is a test-ergonomics reason to delete production observability. This project already runs Robolectric widely (`CoursesProgressAdapterTest`, `VoicesActionsTest`, `DialogUtilsConfirmDialogTest` in this same queue), and under Robolectric `Log` works without any static mocking. Keeping the warning and dropping the `mockkStatic` gets you both.

**Verdict:** Mergeable — solid dedupe with the refresh path correctly exempted; please put the warning log back.

---

Okuro3499 (approved):

PR #17036, opened 2026-09-10
courses: smoother glide cover thumbnail loading (fixes #17025)

### LINKED ISSUE #17025: Size the Glide course covers in CoursesAdapter
## What

Course cover images are decoded at full resolution for small grid/list thumbnails.

`app/src/main/java/org/ole/planet/myplanet/ui/courses/CoursesAdapter.kt` (`bindCover`, ~285-316)

## Why

`bindCover` issues

```kotlin
Glide.with(context).load(model)
    .diskCacheStrategy(DiskCacheStrategy.ALL)
    .signature(ObjectKey(course.courseRev))
    .centerCrop()
    .error(...)
    .into(ivCover)
```

with no `.override(width, height)` anywhere in the file, so the full-size bitmap is decoded for a thumbnail — scroll jank plus bitmap-memory spikes. `bindCover` is used by both the grid and list holders (~409, ~457).

## The fix

- Resolve a target size in px once — from the cover `ImageView`'s layout params, or a density-based constant shared by both call sites — and add `.override(...)` to the request.
- Keep `ObjectKey(course.courseRev)`, `DiskCacheStrategy.ALL`, `centerCrop` and the error drawable exactly as they are.

## Acceptance

- Covers still render identically in both grid and list modes, including the error placeholder.
- Glide decodes at the thumbnail size, not the source size.

## Notes

~15-30 lines, 1 file. Conventions: `docs/CODE_STYLE_GUIDE.md`.

dogi:
Added `.override(targetWidth, targetHeight)` to Glide in `CoursesAdapter.kt` to decode course covers at thumbnail resolution rather than full image resolution.

Fixes #17025

---
*PR created automatically by Jules for task [4872324890950524367](https://jules.google.com/task/4872324890950524367) started by @dogi*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved course cover image sizing in grid and list views for more consistent display.
  * Adjusted image loading to match the available cover container dimensions, helping reduce scaling issues.
  * Standardized cover dimensions across course layouts for improved visual consistency.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 7078
-        versionName = "0.70.78"
+        versionCode = 7079
+        versionName = "0.70.79"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/ui/courses/CoursesAdapter.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/courses/CoursesAdapter.kt
@@ -30,6 +30,7 @@ import org.ole.planet.myplanet.model.MyCourse
 import org.ole.planet.myplanet.utils.CourseSubject
 import org.ole.planet.myplanet.utils.CourseSubjectClassifier
 import org.ole.planet.myplanet.utils.DiffUtils
+import org.ole.planet.myplanet.utils.GridSpanCalculator
 import org.ole.planet.myplanet.utils.ListViewMode
 import org.ole.planet.myplanet.utils.SelectionUtils
 import org.ole.planet.myplanet.utils.StableIdGenerator
@@ -317,7 +318,10 @@ class CoursesAdapter(
             coverContainer.width > 0 -> coverContainer.width
             coverContainer.layoutParams?.width != null && coverContainer.layoutParams.width > 0 -> coverContainer.layoutParams.width
             else -> if (viewMode == ListViewMode.GRID) {
-                (context.resources.displayMetrics.widthPixels / 2).coerceAtLeast(fallbackHeight)
+                val displayMetrics = context.resources.displayMetrics
+                val widthDp = (displayMetrics.widthPixels / displayMetrics.density).toInt()
+                val cols = GridSpanCalculator.columnCount(widthDp)
+                (displayMetrics.widthPixels / cols).coerceAtLeast(fallbackHeight)
             } else {
                 fallbackHeight
             }
--- a/app/src/main/java/org/ole/planet/myplanet/ui/courses/CoursesAdapter.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/courses/CoursesAdapter.kt
@@ -307,17 +307,26 @@ class CoursesAdapter(
         ivSubjectIcon.visibility = View.GONE
         ivCover.visibility = View.VISIBLE
 
-        val density = context.resources.displayMetrics.density
+        val fallbackHeight = if (viewMode == ListViewMode.GRID) {
+            context.resources.getDimensionPixelSize(R.dimen.course_grid_cover_height)
+        } else {
+            context.resources.getDimensionPixelSize(R.dimen.course_list_cover_size)
+        }
+
         val targetWidth = when {
             coverContainer.width > 0 -> coverContainer.width
             coverContainer.layoutParams?.width != null && coverContainer.layoutParams.width > 0 -> coverContainer.layoutParams.width
-            else -> (context.resources.displayMetrics.widthPixels / 2).coerceAtLeast((160 * density).toInt())
+            else -> if (viewMode == ListViewMode.GRID) {
+                (context.resources.displayMetrics.widthPixels / 2).coerceAtLeast(fallbackHeight)
+            } else {
+                fallbackHeight
+            }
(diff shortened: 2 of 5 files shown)


dogi:
**Review — rating: 75/100**

This is a much more careful version of the same idea as #17040, and I want to give the sizing logic credit before the objection: the measured-size branch comes first, the `layoutParams` branch actually pays off here (`item_course_list.xml` pins `cover_container` to a fixed `44dp × 44dp`, so it resolves correctly even on the first bind before layout), and the constants aren't arbitrary — `84 * density` is exactly the `84dp` height `item_course_grid.xml` gives `cover_container`, and `screenWidth / 2` models the two-column grid. Someone read the layouts before writing this.

That's also where the problem is. Those numbers are the layouts, transcribed into Kotlin, with nothing tying the two together. Change `item_course_grid.xml`'s `84dp` and the adapter keeps decoding at the old height, silently, with no compile error and no test to catch it. `160dp` and the `/2` encode a span count that isn't declared anywhere near the adapter either.

The second question is what `.override()` buys. `.into(ImageView)` already defers until the view is measured and decodes at exactly that size, so on every bind after the first — the overwhelming majority in a RecyclerView — the heuristic is reproducing a number Glide would have computed correctly anyway. The genuine argument for pinning it is cache-key stability, since a first-bind heuristic size and a later measured size produce two cache entries for the same cover. If that's the motivation it's worth stating, because it changes which value you want.

Concretely: move `84dp` into `dimens.xml` and reference it from both `item_course_grid.xml` and here via `resources.getDimensionPixelSize(...)`, and do the same for the list layout's `44dp`. Then the fallbacks can't drift from the views they're describing.

Unlike #17040 this doesn't over-decode — the numbers are right today. It's the coupling I'd like fixed before it lands.

**Verdict:** Needs another round — correct sizing, but layout dimensions duplicated in code with no link back to the XML.

---

dogi:
**Re-review — Rating: 88/100** (was 75)

The duplicated layout dimensions are gone. `course_grid_cover_height` (84dp) and `course_list_cover_size` (44dp) now live in `dimens.xml`, `item_course_grid.xml` and `item_course_list.xml` reference them, and the adapter reads the same resources through `getDimensionPixelSize`. One source of truth in three places — the drift I was worried about can't happen now.

The revision also fixed something I'd only half-noticed. Master's fallback used `84dp` and `screenWidth / 2` for *both* view modes, which was wrong for the list layout — those cells are 44dp squares, nowhere near half the screen. The new `if (viewMode == ListViewMode.GRID)` branch picks `course_grid_cover_height` vs `course_list_cover_size` and, in list mode, uses `fallbackHeight` for the width too. So the cold-bind path is now accurate in both modes rather than just the grid.

Precedence is unchanged and still right: measured width first, `layoutParams` second, resource-derived fallback last, all `coerceAtLeast(1)`. Build (default + lite) and both shards green.

**Verdict:** Mergeable.

---

Okuro3499 (dismissed):
```
File: app/src/main/java/org/ole/planet/myplanet/ui/courses/CoursesAdapter.kt, bindCover() (lines 316-324)                                                                        
    Problem: The pre-measurement width fallback hardcodes a 2-column assumption (widthPixels / 2) for grid mode, but actual columns range 2-6 per GridSpanCalculator. This           
    over-estimates (and thus over-decodes) the target bitmap size on any device with 3+ columns, precisely during the first-bind scenario the override is meant to optimize.         
   
```

dogi:

Same family as #17040 — size the Glide decode and pull the hardcoded layout values into dimens — but the sizing logic here is considerably more involved, and that's both its strength and its cost.

**What's correct**

The `layoutParams.width > 0` guard is the detail that would most easily have been wrong. `MATCH_PARENT` and `WRAP_CONTENT` are `-1` and `-2`, so a `!= null` check alone would have fed Glide a negative target; the `> 0` comparison excludes both. `.coerceAtLeast(1)` then guarantees Glide's requirement that override dimensions be positive even if every branch somehow yields zero. ✓

The grid fallback deriving column count from `GridSpanCalculator.columnCount(widthDp)` rather than hardcoding a number is the right instinct — it's the same calculator the layout manager uses, so the estimated cell width tracks the real one if the breakpoints ever change.

Extracting `course_grid_cover_height` (84dp) and `course_list_cover_size` (44dp) into `dimens.xml` and referencing them from both layouts and code gives the same single-source-of-truth property as #17040, and matters more here because the code branches on view mode.

**The cost, and one concrete pessimisation**

`fallbackHeight` is computed unconditionally at the top of the bind path:

```kotlin
val fallbackHeight = if (viewMode == ListViewMode.GRID) {
    context.resources.getDimensionPixelSize(R.dimen.course_grid_cover_height)
} else {
    context.resources.getDimensionPixelSize(R.dimen.course_list_cover_size)
}
```

but it's only *used* when `coverContainer` hasn't been measured — which, after the first layout pass, is never. So every bind in the steady state pays a `getDimensionPixelSize` lookup whose result is discarded, in a method whose stated purpose is to reduce per-bind work. Moving it into the `else` branches of the two `when` expressions, or hoisting both dimens into `ViewHolder` fields resolved once, removes that. Small, but it cuts against the PR's own rationale.

More broadly: this is ~30 lines of branching in an adapter bind path to compute something `into(ImageView)` derives on its own once the view is measured. The unmeasured-view case is real (see #17040), but I'd expect the fallback ladder to be the exception path, not code that runs in full on every bind. Caching the resolved target on the ViewHolder after the first successful measurement would collapse the common case to a field read.

**Verdict:** Mergeable — the pessimisation is worth a follow-up, not a block.

---

PR #17040, opened 2026-09-10
courses: smoother inline resources preview image loading (fixes #17033)

### LINKED ISSUE #17033: Size the Glide previews in InlineResourceAdapter
## What

Course-step resource rows decode full-size images and video frames into thumbnail `ImageView`s.

`app/src/main/java/org/ole/planet/myplanet/ui/courses/InlineResourceAdapter.kt`

## Why

Previews are loaded in `showImagePreview` (~205), `showVideoPreview` (~219) and `showHtmlPreview` (~246), and the file contains no `.override(` at all. The PDF path is already sized.

## The fix

- Add one private preview-decode-size helper or constant in px (shared by the three paths).
- Apply `.override(...)` to those three Glide loads.
- Do not touch mime routing, the file-existence checks, or the PDF/audio/text paths.

## Acceptance

- Image, video and HTML previews still render in the same rows with the same fallbacks.
- Glide decodes at the thumbnail size.

## Notes

Small, 1 file. Conventions: `docs/CODE_STYLE_GUIDE.md`.

dogi:
Size Glide previews in InlineResourceAdapter to reduce memory footprint when decoding image, video, and HTML cover thumbnails.

Fixes #17033

---
*PR created automatically by Jules for task [15628710347172281435](https://jules.google.com/task/15628710347172281435) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 7079
-        versionName = "0.70.79"
+        versionCode = 7080
+        versionName = "0.70.80"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/ui/courses/InlineResourceAdapter.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/courses/InlineResourceAdapter.kt
@@ -222,9 +222,11 @@ class InlineResourceAdapter(
         binding.videoThumbnailContainer.visibility = View.VISIBLE
         val exists = withContext(dispatcherProvider.io) { file.exists() }
         if (exists) {
+            val (widthPx, heightPx) = getPreviewDimensions(context)
             Glide.with(context)
                 .load(file)
                 .diskCacheStrategy(DiskCacheStrategy.ALL)
+                .override(widthPx, heightPx)
                 .centerCrop()
                 .into(binding.ivVideoThumbnail)
         }
--- a/app/src/main/java/org/ole/planet/myplanet/ui/courses/InlineResourceAdapter.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/courses/InlineResourceAdapter.kt
@@ -205,12 +205,12 @@ class InlineResourceAdapter(
     private suspend fun showImagePreview(binding: ItemInlineResourceBinding, context: Context, file: File) {
         val exists = withContext(dispatcherProvider.io) { file.exists() }
         if (exists) {
-            val previewSizePx = getPreviewDecodeSizePx(context)
+            val (widthPx, heightPx) = getPreviewDimensions(context)
             binding.ivResourcePreview.visibility = View.VISIBLE
             Glide.with(context)
                 .load(file)
                 .diskCacheStrategy(DiskCacheStrategy.ALL)
-                .override(previewSizePx)
+                .override(widthPx, heightPx)
                 .centerCrop()
                 .placeholder(R.drawable.ole_logo)
                 .error(R.drawable.ole_logo)
@@ -222,11 +222,9 @@ class InlineResourceAdapter(
         binding.videoThumbnailContainer.visibility = View.VISIBLE
         val exists = withContext(dispatcherProvider.io) { file.exists() }
         if (exists) {
-            val previewSizePx = getPreviewDecodeSizePx(context)
             Glide.with(context)
                 .load(file)
                 .diskCacheStrategy(DiskCacheStrategy.ALL)
-                .override(previewSizePx)
                 .centerCrop()
                 .into(binding.ivVideoThumbnail)
         }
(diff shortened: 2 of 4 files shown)


dogi:
**Review — rating: 62/100**

I think this makes decoding more expensive, not less, for the main preview path.

`Glide.override(int size)` is shorthand for `override(size, size)`, so `.override(previewSizePx)` with `PREVIEW_DECODE_SIZE_DP = 360` requests a **360dp × 360dp** decode. But `iv_resource_preview` in `item_inline_resource.xml` is `match_parent × 180dp` (lines 21–24). Without `.override()`, Glide's `ViewTarget` already waits for measurement and decodes to the view's real size — roughly `screenWidth × 180dp`. With it, the height dimension doubles from 180dp to 360dp, so on a 3× device you go from about 1080×540 to 1080×1080: ~2× the bitmap memory and decode work per preview, in a RecyclerView. Since `centerCrop` then crops back to the 180dp-tall view, that extra half is decoded and thrown away.

`ivVideoThumbnail` is `match_parent × match_parent` inside `videoThumbnailContainer`, so whether 360dp helps or hurts there depends on the container's measured height — but that's the case where Glide's automatic sizing was already giving the exact answer.

The underlying instinct — don't decode a multi-megapixel original for a thumbnail — is sound, and it's worth saying that Glide's default already implements it by sizing to the target view. `.override()` earns its place when you want a *stable cache key* independent of measurement, or when there's no view to measure. If that's the goal here, say so and use the real aspect: `.override(previewWidthPx, previewHeightPx)` with the height derived from the same 180dp the layout uses. Best of all, put that 180dp in `dimens.xml` and reference it from both the XML and the adapter so the two can't drift.

CI is green and nothing is broken — this is a performance change that measures the wrong way.

**Verdict:** Needs another round — the override should match the view's real height rather than square it off at 360dp.

---

dogi:

The over-decode is gone, and the fix is better than what I suggested. Rather than hardcoding the corrected height, `inline_resource_preview_height` (180dp) went into `dimens.xml`, both `iv_resource_preview` and `video_thumbnail_container` reference it in the layout, and `getPreviewDimensions` reads it back with `getDimensionPixelSize`. The decode height now *cannot* diverge from the view height, because they're the same resource.

`.override(widthPx, heightPx)` with the real 180dp replaces the square `override(360)`, so the ~2× height over-decode I flagged is eliminated — on a 3× device that's roughly 1080×540 instead of 1080×1080 per preview, in a RecyclerView.

Dropping the override from `showVideoThumbnail` is the right call too: `ivVideoThumbnail` is `match_parent × match_parent` inside a container that now carries the shared dimen, so Glide's own view measurement already produces the exact size and an override would only risk disagreeing with it.

**Nit:** `widthPx` is `displayMetrics.widthPixels`, i.e. full screen width, while the ImageView is `match_parent` inside a parent with horizontal padding — so width is over-requested by the padding. Bounded and harmless, but `binding.ivResourcePreview.width.takeIf { it > 0 } ?: widthPx` would tighten it.

**Verdict:** Mergeable.

---

Okuro3499 (dismissed):
```
File: app/src/main/java/org/ole/planet/myplanet/ui/courses/InlineResourceAdapter.kt, showVideoPreview (~line 218-228)                                                            
    Problem: The Glide load for binding.ivVideoThumbnail has no .override(...) call — it was added in 912f65856 and removed in 95799da30 without replacement, so video thumbnails    
    decode at full resolution while image and HTML-cover previews don't.
```

dogi:
**Re-review — Rating: 86/100**

Adds `.override(widthPx, heightPx)` to the three Glide loads in `InlineResourceAdapter`, and — the part I'd keep even if the Glide change were dropped — extracts the hardcoded `180dp` into `@dimen/inline_resource_preview_height`, referenced by both the layout and the code.

**The dimen extraction is the durable win**

`getPreviewDimensions` reads the same resource the layout uses, so the decode target and the view height cannot drift apart. Hardcoding `180` in Kotlin next to `180dp` in XML is exactly the pair that silently desynchronises the first time someone adjusts the layout, and this removes that possibility rather than just fixing today's value.

**Where `.override` actually earns its place**

It's worth being precise, because `into(ImageView)` already waits for measurement and sizes the decode to the view — so `.override` is often redundant. The case where it isn't is `showVideoThumbnail`: `ivVideoThumbnail` lives inside `videoThumbnailContainer`, and when a container is `GONE` Glide's size determination has nothing to measure and falls back to screen dimensions. Pinning the target sidesteps that. The same reasoning covers `showImagePreview` and the cover-image branch, where visibility is flipped immediately before the load and the view may not have been laid out yet.

So: a real if narrow benefit, not the general-purpose speedup the framing suggests.

**Two things worth knowing**

`widthPx` is `displayMetrics.widthPixels` — full screen width — while the `ImageView` is `match_parent` inside a parent that may carry padding. Over-estimating is harmless for `.override` (a slightly larger decode, then `centerCrop`), so this is the right direction to err, but it isn't the view's true width.

Because the override dimensions become part of Glide's cache key, the disk cache is now keyed per screen width. On a single device that's invisible; it just means the cached entry won't transfer across form factors. Quantising the width (or using the container's measured width when available) would avoid that, but for a 180dp-tall preview row it isn't worth the complexity.

**Verdict:** Mergeable.

---

PR #17041, opened 2026-09-10
courses: smoother steps progressing (fixes #17026)

### LINKED ISSUE #17026: Cancel the superseded progress query when swiping course steps
## What

Each step change launches an uncancelled progress query, so fast swiping stacks queries and the last one to finish wins the write.

`app/src/main/java/org/ole/planet/myplanet/ui/courses/TakeCourseFragment.kt` (`updateStepDisplay`, ~193-210)

## Why

```kotlin
private fun updateStepDisplay(position: Int) {
    ...
        val currentProgress = viewModel.getCurrentProgress(steps, userModel?.id, courseId)   // ~203
        ...
    }
}
```

`updateStepDisplay` is called from the page-change callback (~140), `onResume` (~151) and the step-jump path (~313). Nothing cancels the previous launch, so whichever coroutine finishes last writes `binding.courseProgress` and `currentCourseProgress` — which is not necessarily the current step.

## The fix

- Hold the launch in a `private var progressJob: Job?`.
- Cancel it before reassigning at the top of the coroutine block.
- Null the field in `onDestroyView`.
- Leave the synchronous `tvStep` / `nextStep` / `courseStepProgressBar` writes at the top of the function **outside** the coroutine so they stay immediate.

## Acceptance

- Swiping quickly through steps leaves the progress display matching the step actually shown.
- Returning to the screen still refreshes progress.
- No leaked job after `onDestroyView`.

## Notes

~25 lines, 1 file. Conventions: `docs/CODE_STYLE_GUIDE.md`.

dogi:
Cancels any active progress query Job before launching a new query in TakeCourseFragment.updateStepDisplay to fix fast-swiping progress display race conditions, and clears the job reference in onDestroyView.

Fixes #17026

---
*PR created automatically by Jules for task [319120447496011515](https://jules.google.com/task/319120447496011515) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 7075
-        versionName = "0.70.75"
+        versionCode = 7076
+        versionName = "0.70.76"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/ui/courses/TakeCourseFragment.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/courses/TakeCourseFragment.kt
@@ -18,6 +18,7 @@ import com.google.android.material.snackbar.Snackbar
 import dagger.hilt.android.AndroidEntryPoint
 import java.util.Locale
 import javax.inject.Inject
+import kotlinx.coroutines.Job
 import kotlinx.coroutines.cancelChildren
 import kotlinx.coroutines.launch
 import org.ole.planet.myplanet.R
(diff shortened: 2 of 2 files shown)


dogi:
**Review — rating: 85/100**

Tracks the progress-lookup coroutine in a `progressJob` field and cancels the previous one before launching a new one.

The race is real. `viewModel.getCurrentProgress(...)` is suspending, and the old code fired a fresh unowned `launch` on every step change. Swiping quickly through a course left several in flight against the same `currentCourseProgress` field and the same progress views, with no ordering guarantee — a slower query for step 2 could land after step 5's and leave the UI showing stale progress. `progressJob?.cancel()` reduces that to at most one live query, which is exactly the right granularity here since only the newest step's result is ever wanted.

The teardown is fine as written even though `onDestroyView` only nulls the field without cancelling: the job is launched on `viewLifecycleOwner.lifecycleScope`, which the framework cancels when the view lifecycle hits `DESTROYED`, so the coroutine dies regardless. Worth noting the nearby `lifecycleScope.coroutineContext.cancelChildren()` on the line above operates on the *fragment* scope and would not have covered this job — the view scope is what saves it.

**Nit:** `progressJob = null` reads as if it were doing the cancelling. An explicit `progressJob?.cancel()` there would cost nothing and make the intent legible without the reader having to reason about which of the two scopes the job belongs to.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17044, opened 2026-09-10
life: smoother health user view modelling (fixes #17014)

### LINKED ISSUE #17014: Stop re-running the patient query (and its decrypt) on every health screen resume
## What

The health screen loads the selected patient, then loads the same patient again on resume — cancelling the first load and redoing the record decrypt.

Files:
- `app/src/main/java/org/ole/planet/myplanet/ui/health/HealthViewModel.kt`
- `app/src/main/java/org/ole/planet/myplanet/ui/health/MyHealthFragment.kt`

## Why

`MyHealthFragment` reaches `setupInitialData()` (~117 → ~238-239), which calls `viewModel.loadInitialPatient()`; that resolves the user and calls `selectPatient(normalizedId)`. Then `onResume` (~327-332) calls `viewModel.selectPatient(normalizedId)` for the same id, and `selectPatient` (~88-89) cancels the in-flight job and redoes both `getPatientById` and `getPatientHealthRecords` — the second of which decrypts the stored record.

## The fix

In `HealthViewModel`:

- Track the currently loaded/loading patient id.
- Return early from `selectPatient(userId)` when the requested id matches **and** either the job is still active or `_patientDetailState` already holds that patient's record.
- Clear the tracked id on failure so a retry works.
- Add an explicit `refreshSelectedPatient()` that always re-queries, and use it for the realtime-sync and post-edit paths that legitimately need fresh data.

Leave `loadPatients` / `searchPatients` alone.

## Acceptance

- Entering the health screen decrypts the selected patient's records once, not twice.
- Selecting a different patient still loads them.
- Editing a record, and a realtime sync update, still refresh the display.

## Notes

PR #17000 also touches `MyHealthFragment.kt`, but only the binding boilerplate (hunks at lines 13, 31, 47, 60, 78-83, 354) — your change is at ~117/238/327 and mostly in `HealthViewModel`, so it should not collide. Conventions: `docs/CODE_STYLE_GUIDE.md`.

dogi:
This change prevents MyHealthFragment from re-querying and re-decrypting patient records on screen resume if the patient is already loaded or in-flight in HealthViewModel.

Key changes:
- Track currentPatientId in HealthViewModel.
- Early return from selectPatient(userId) when userId == currentPatientId and the load job is active or _patientDetailState holds that patient.
- Reset currentPatientId on load failures to allow retries.
- Re-throw CancellationException in coroutine exception handling to prevent reset on cancellation.
- Add refreshSelectedPatient() to force re-queries for realtime-sync and post-edit paths.
- Update MyHealthFragment and HealthExaminationAdapter to use ActivityResultLauncher when editing records, invoking refreshSelectedPatient() upon return.
- Add unit tests in HealthViewModelTest.

Fixes #17014

---
*PR created automatically by Jules for task [8725985001149550536](https://jules.google.com/task/8725985001149550536) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/ui/health/HealthExaminationAdapter.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/health/HealthExaminationAdapter.kt
@@ -31,7 +31,8 @@ class HealthExaminationAdapter(
     private var mh: HealthExamination,
     private var userModel: UserEntity?,
     private var userMap: Map<String, UserEntity>,
-    private val dispatcherProvider: DispatcherProvider
+    private val dispatcherProvider: DispatcherProvider,
+    private val onEditClick: ((Intent) -> Unit)? = null
 ) : ListAdapter<HealthExaminationAdapter.HealthExaminationItem, HealthExaminationViewHolder>(DIFF_CALLBACK) {
 
     data class HealthExaminationItem(
@@ -149,9 +150,14 @@ class HealthExaminationAdapter(
         dialog.window?.setBackgroundDrawable(colorMultiSelectGrey.toDrawable())
 
         dialog.setButton(DialogInterface.BUTTON_NEUTRAL, context.getString(R.string.edit)) { _: DialogInterface?, _: Int ->
-            context.startActivity(Intent(context, HealthExaminationActivity::class.java)
+            val intent = Intent(context, HealthExaminationActivity::class.java)
                 .putExtra("id", realmExamination._id)
-                .putExtra("userId", mh._id))
+                .putExtra("userId", mh._id)
+            if (onEditClick != null) {
+                onEditClick.invoke(intent)
+            } else {
+                context.startActivity(intent)
+            }
         }
 
         dialog.show()
--- a/app/src/main/java/org/ole/planet/myplanet/ui/health/HealthViewModel.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/health/HealthViewModel.kt
@@ -100,13 +100,16 @@ class HealthViewModel @Inject constructor(
             ?: _loggedInUser.value?.effectiveId?.trim()
         if (!targetId.isNullOrEmpty()) {
             fetchPatientData(targetId)
+        } else {
+            loadInitialPatient()
         }
     }
 
     private fun fetchPatientData(userId: String) {
         currentPatientId = userId
         selectPatientJob?.cancel()
-        selectPatientJob = viewModelScope.launch {
+        var job: Job? = null
+        job = viewModelScope.launch {
             _isLoading.value = true
             try {
                 val user = healthRepository.getPatientById(userId)
@@ -123,9 +126,12 @@ class HealthViewModel @Inject constructor(
                 _patientDetailState.value = PatientDetailState(null, null)
                 throw e
             } finally {
-                _isLoading.value = false
+                if (selectPatientJob === job) {
+                    _isLoading.value = false
+                }
             }
         }
+        selectPatientJob = job
     }
 
     fun loadHealthData(userId: String) {
--- a/app/src/main/java/org/ole/planet/myplanet/ui/health/HealthViewModel.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/health/HealthViewModel.kt
@@ -4,6 +4,7 @@ import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
 import dagger.hilt.android.lifecycle.HiltViewModel
 import javax.inject.Inject
+import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.async
 import kotlinx.coroutines.coroutineScope
@@ -52,6 +53,7 @@ class HealthViewModel @Inject constructor(
 
     private var searchJob: Job? = null
     private var selectPatientJob: Job? = null
+    private var currentPatientId: String? = null
 
     fun loadPatients(sortBy: String = "joinDate", descending: Boolean = true) {
         viewModelScope.launch {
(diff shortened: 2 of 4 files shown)


dogi:
**Review — rating: 81/100**

Does more than the title says, and the extra part is the valuable bit. Routing the adapter's edit action and the two add/update buttons through `registerForActivityResult` instead of `startActivity` means returning from `HealthExaminationActivity`/`AddHealthActivity` now actually refreshes the patient. Previously those were fire-and-forget launches with no result path, so a freshly edited record didn't show until something else happened to reload the screen. Threading that through the adapter as an optional `onEditClick: ((Intent) -> Unit)?` with a `startActivity` fallback keeps the adapter usable elsewhere unchanged.

The dedupe itself is sound: `selectPatient` skips only when the id matches *and* there's either a live job or an already-populated `user`, and the `refreshSelectedPatient` escape hatch — with its `currentPatientId ?: loggedInUser?.effectiveId` fallback — keeps the forced paths working. Clearing `currentPatientId` when the patient turns out to be null is a good touch; without it a failed lookup would be cached as "already loaded" and never retried. And `if (e is CancellationException) throw e` before the generic handling is exactly right.

**Nit — a real race, though a mild one.** `_isLoading.value = false` sits in `finally`, and `fetchPatientData` starts a new load with a plain `selectPatientJob?.cancel()`. `cancel()` doesn't wait, so the outgoing job's `finally` can run *after* the incoming job has already set `_isLoading.value = true`, leaving the spinner hidden while a load is genuinely in flight. `cancelAndJoin()`, or only clearing the flag when `selectPatientJob` is still this job, closes it.

Also worth a sanity check on the swap of `loadInitialPatient()` for `refreshSelectedPatient()` in `refreshHealthData` — the fallback chain looks equivalent, but they're not literally the same entry point on a cold sync-triggered refresh.

**Verdict:** Mergeable — good refresh fix; the `isLoading` ordering is worth tightening.

---

Okuro3499 (dismissed):
```
HealthViewModel.kt:120-124 — fetchPatientData's catch block resets state to PatientDetailState(null, null) and then does throw e for any non-cancellation exception. Since       
    selectPatientJob runs via viewModelScope.launch with no installed CoroutineExceptionHandler, rethrowing here propagates to the coroutine's default uncaught-exception handling — 
    i.e., a repository failure (e.g., a DB or network error from getPatientById/getPatientHealthRecords) will crash the app instead of leaving the UI in the empty state the code    
    just set. This isn't a new crash surface (the pre-diff code had no try/catch either and would crash the same way), but the added try/catch clearly signals an intent to handle   
    the error gracefully, and the throw e defeats that intent — the state reset becomes moot because the process dies immediately after. Fix: drop the trailing throw e (keep if (e  
    is CancellationException) throw e to preserve cancellation semantics) so a query failure surfaces as an empty patient view instead of a crash, consistent with                   
    ResourceViewerViewModel's pattern. No existing test exercises this path — a test for "non-null exception during fetch leaves UI in empty state without crashing" should be added 
    alongside the fix.
```

dogi:
**Re-review — Rating: 89/100** (was 81)

Both points from my last pass are addressed, and the first one is fixed properly rather than papered over.

**The `_isLoading` race.** Previously `finally { _isLoading.value = false }` could fire *after* a newly-started load had already set the flag to `true`, because `selectPatientJob?.cancel()` doesn't wait for the outgoing coroutine to unwind — leaving the spinner hidden during a live load. The fix captures the job in a local and gates the reset on identity:

```kotlin
finally {
    if (selectPatientJob === job) {
        _isLoading.value = false
    }
}
```

A superseded coroutine now can't clear a flag it no longer owns. That's the right shape — better than `cancelAndJoin()`, which would have made cancellation blocking on the caller.

**The `loadInitialPatient` swap.** `refreshSelectedPatient` now falls back to `loadInitialPatient()` when neither an explicit id, `currentPatientId`, nor `loggedInUser.effectiveId` resolves, so the cold sync-triggered refresh path that `refreshHealthData` used to reach directly is preserved rather than silently doing nothing.

The rest stands: the `registerForActivityResult` rewiring still fixes the real bug (returning from an edit didn't refresh), the dedupe guard still requires both a matching id and either a live job or populated state, `currentPatientId` is still cleared on a null lookup so a failure isn't cached as "loaded", and `if (e is CancellationException) throw e` still precedes the generic handling.

**Nit, and genuinely minor:** `viewModelScope` uses `Dispatchers.Main.immediate`, so if `getPatientById` ever completed *without* suspending, the whole body including `finally` would run before `selectPatientJob = job` executes and the identity check would be false, leaving `_isLoading` stuck true. Room's suspend DAOs always dispatch, so this can't happen in production — noting it only because the pattern is worth knowing if it gets copied somewhere with a non-suspending fake.

**Verdict:** Mergeable.

---

dogi:
**Re-review — Rating: 90/100**

Stops `MyHealthFragment` from re-querying the patient on every resume, and fixes a related UX bug along the way. The concurrency handling here is more careful than most of this queue.

**What's done right**

The superseded-load guard is the part that's easy to get wrong:

```kotlin
} finally {
    if (selectPatientJob === job) {
        _isLoading.value = false
    }
}
```

A cancelled load can't clear the loading flag for the newer load that replaced it. Identity comparison against the captured local `job`, not the field. And there's a test that actually pins the behaviour — `selectPatient superseding load keeps isLoading true while second load is in flight` advances time between the two calls and asserts `isLoading` stays true across the handover.

The cancellation handling is correct:

```kotlin
} catch (e: Exception) {
    if (e is CancellationException) throw e
```

`CancellationException` is an `Exception`, so a bare `catch (e: Exception)` swallows it and turns a normal cancellation into a "patient not found" state. Rethrowing first is the right guard, and it's worth noting because several other PRs in this queue inherit the unguarded version.

`currentPatientId = null` on both the not-found and the exception paths means a failed load doesn't poison the early-return check — the retry test (`resets tracked ID on failure so retry works`) covers exactly that.

**The resume path**

Replacing `startActivity` with `registerForActivityResult` is what makes `refreshSelectedPatient()` possible: returning from an edit now refreshes *the patient being viewed*, where `refreshHealthData()` previously called `loadInitialPatient()` and reset the screen to the logged-in user. That's a real fix, not just plumbing. Registering the launcher as a property initializer is correct — it runs at construction, well before the fragment reaches STARTED.

The early-return in `selectPatient` is safe precisely because the refresh path bypasses it: sync-triggered `refreshHealthData()` goes through `refreshSelectedPatient()` → `fetchPatientData()` unconditionally, so cached state can't go stale after a sync.

**Two minor things**

The `onEditClick: ((Intent) -> Unit)?` parameter with a `startActivity` fallback keeps the adapter backwards-compatible, but it means the adapter has two navigation modes and only one is exercised. Since `MyHealthFragment` is the only construction site, a non-null callback would remove the branch and the dead path.

One launcher serves three destinations (edit examination, add record, update health) and its callback ignores the result code, so a cancelled add still triggers a refresh. Harmless — the refresh is idempotent — but if you ever want to skip the requery on cancel, the result is right there.

**Verdict:** Mergeable.

---

PR #17045, opened 2026-09-10
sync: smoother diagnostics repository crash log batch uploading (fixes #17022)

### LINKED ISSUE #17022: Batch the crash-log upload acknowledgement into one transaction
## What

Acknowledging uploaded crash logs issues one `UPDATE` per row, each in its own implicit transaction.

Files:
- `app/src/main/java/org/ole/planet/myplanet/services/upload/UploadConfigs.kt` (`CrashLog.markUploaded`, ~223)
- `app/src/main/java/org/ole/planet/myplanet/data/room/dao/ApkLogDao.kt`
- `app/src/main/java/org/ole/planet/myplanet/repository/DiagnosticsRepository.kt` (+ impl)

## Why

```kotlin
results.filter { !diagnosticsRepository.markApkLogUploaded(result.localId, result.remoteRev) }
```

One suspend call per row into `ApkLogDao.markUploaded`, a bare `UPDATE` that Room runs in its own implicit transaction.

`SubmitPhotosDao` (lines ~26-43) already carries the exact pattern to copy: a `markUploadedBatchInternal` plus a `@Transaction markUploadedBatch`, driven by an `UploadUpdate` projection.

## The fix

- Add the partial-update projection and a `@Transaction markUploadedBatch` to `ApkLogDao`, mirroring `SubmitPhotosDao`.
- Expose `markApkLogsUploaded(updates): Set<String>` on `DiagnosticsRepository` (and its impl), returning the ids that were **not** applied.
- Rewrite `CrashLog.markUploaded` to one batched call, keeping the same "failed ids" semantics the caller relies on today.
- Keep the single-row `markUploaded` on the DAO and repository — do not remove it.

No schema change, so **no** `AppDatabase` version bump.

## Acceptance

- Uploading N crash logs performs one batched update instead of N.
- Rows that could not be marked are still reported back exactly as before.

## Notes

PR #17015 touches `CrashLogStore.kt` only (one line at ~40) — no overlap with this change. Conventions: `docs/CODE_STYLE_GUIDE.md`.

dogi:
Batch crash log upload acknowledgement into one transaction using Room partial updates and @Transaction in ApkLogDao.

Fixes #17022

---
*PR created automatically by Jules for task [14856878280487507671](https://jules.google.com/task/14856878280487507671) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/data/room/dao/ApkLogDao.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/data/room/dao/ApkLogDao.kt
@@ -21,10 +21,6 @@ interface ApkLogDao {
     @Insert(onConflict = OnConflictStrategy.REPLACE)
     suspend fun insertAll(logs: List<ApkLog>)
 
-    /** Returns the number of rows updated (0 means the local row was gone). */
-    @Query("UPDATE apk_log SET _rev = :rev WHERE id = :id")
-    suspend fun markUploaded(id: String, rev: String): Int
-
     @Query("SELECT id FROM apk_log WHERE id IN (:ids)")
     suspend fun getExistingIds(ids: List<String>): List<String>
 
--- a/app/src/main/java/org/ole/planet/myplanet/data/room/dao/ApkLogDao.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/data/room/dao/ApkLogDao.kt
@@ -1,9 +1,12 @@
 package org.ole.planet.myplanet.data.room.dao
 
+import androidx.room.ColumnInfo
 import androidx.room.Dao
 import androidx.room.Insert
 import androidx.room.OnConflictStrategy
 import androidx.room.Query
+import androidx.room.Transaction
+import androidx.room.Update
 import org.ole.planet.myplanet.model.ApkLog
 
 @Dao
@@ -21,4 +24,26 @@ interface ApkLogDao {
     /** Returns the number of rows updated (0 means the local row was gone). */
     @Query("UPDATE apk_log SET _rev = :rev WHERE id = :id")
     suspend fun markUploaded(id: String, rev: String): Int
+
+    @Query("SELECT id FROM apk_log WHERE id IN (:ids)")
+    suspend fun getExistingIds(ids: List<String>): List<String>
+
+    @Update(entity = ApkLog::class)
+    suspend fun markUploadedBatchInternal(updates: List<UploadUpdate>)
+
+    @Transaction
+    suspend fun markUploadedBatch(updates: List<UploadUpdate>): Set<String> {
+        if (updates.isEmpty()) return emptySet()
+        val ids = updates.map { it.id }
+        val existingIds = ids.chunked(900).flatMap { chunk -> getExistingIds(chunk) }.toSet()
+        updates.chunked(900).forEach { chunk ->
+            markUploadedBatchInternal(chunk)
+        }
+        return ids.filterNot { it in existingIds }.toSet()
+    }
+
+    data class UploadUpdate(
+        @ColumnInfo(name = "id") val id: String,
+        @ColumnInfo(name = "_rev") val rev: String
+    )
 }
--- a/app/src/main/java/org/ole/planet/myplanet/data/room/dao/ApkLogDao.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/data/room/dao/ApkLogDao.kt
@@ -36,9 +36,7 @@ interface ApkLogDao {
         if (updates.isEmpty()) return emptySet()
         val ids = updates.map { it.id }
         val existingIds = ids.chunked(900).flatMap { chunk -> getExistingIds(chunk) }.toSet()
-        updates.chunked(900).forEach { chunk ->
-            markUploadedBatchInternal(chunk)
-        }
+        markUploadedBatchInternal(updates)
         return ids.filterNot { it in existingIds }.toSet()
     }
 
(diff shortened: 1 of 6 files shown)


dogi:
**Review — rating: 82/100**

Replaces a per-row `markApkLogUploaded` loop with a single `@Transaction` batch. Each of those `UPDATE`s previously ran in its own implicit transaction, so acknowledging N uploaded crash logs cost N commits; now it's one.

The details that usually break this kind of change are handled:

- **Partial-entity update.** `@Update(entity = ApkLog::class)` with a `UploadUpdate(id, _rev)` POJO is the correct Room idiom for touching two columns without loading and rewriting whole `ApkLog` rows.
- **Variable limit.** `getExistingIds` is chunked at 900, comfortably under SQLite's 999-parameter ceiling — the failure mode this would otherwise hit only in production, with a large backlog, is pre-empted.
- **Return semantics preserved.** The old contract was "0 rows updated ⇒ the local row is gone ⇒ report as a local failure." `ids.filterNot { it in existingIds }` reproduces that, and because the existence read and the writes share the `@Transaction`, the snapshot can't shift between them. `UploadConfigs` then filters on exactly that set, so the caller's behaviour is unchanged.

**Nit — layering.** `DiagnosticsRepository.markApkLogsUploaded(updates: List<ApkLogDao.UploadUpdate>)` puts a DAO-nested type in the repository interface, and `UploadConfigs` now imports `ApkLogDao` just to construct it. `CLAUDE.md` has repositories hiding their DAOs behind plain entity/data-class types, and this is the one place the abstraction leaks through. A small `data class ApkLogUpload(val id: String, val rev: String)` in the repository layer, mapped to `UploadUpdate` inside the impl, keeps the DAO out of `services/upload/`.

**Nit:** chunking `markUploadedBatchInternal` at 900 is harmless but unnecessary — `@Update` with a list emits one statement per row rather than a single parameterised one, so there's no variable ceiling to respect there. Only the `IN (:ids)` read needed it.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

Okuro3499 (dismissed):
```
 Must-fix issues                                                                                                                                                                 
  - File: repository/DiagnosticsRepository.kt / DiagnosticsRepositoryImpl.kt / data/room/dao/ApkLogDao.kt                                                                            
    Problem: markApkLogUploaded(localId, rev): Boolean (repo method) and its underlying ApkLogDao.markUploaded(id, rev): Int query are now dead in production code — UploadConfigs.kt
    
```

dogi:
**Re-review — Rating: 90/100** (was 82)

Both nits from my last pass are resolved.

**Layering.** `DiagnosticsRepository` now declares `markApkLogsUploaded(updates: List<ApkLogUpload>)` against a plain `data class ApkLogUpload(val id: String, val rev: String)` in the repository package, and `DiagnosticsRepositoryImpl` maps to `ApkLogDao.UploadUpdate` internally. `UploadConfigs` imports `org.ole.planet.myplanet.repository.ApkLogUpload` and no longer imports `ApkLogDao` at all — so the DAO type stops leaking into `services/upload/`, which is what `CLAUDE.md` asks of the repository boundary.

**The redundant chunking.** `markUploadedBatchInternal(updates)` is now called with the full list. That's correct: `@Update` with a collection emits one statement per row, so there was never a bind-variable ceiling to respect there. The chunking that *does* matter — `ids.chunked(900)` feeding the `IN (:ids)` read — is untouched.

The core of the change is unchanged and still good: one `@Transaction` replaces N per-row commits when acknowledging uploaded crash logs, `@Update(entity = ApkLog::class)` with a partial-entity POJO avoids loading and rewriting whole rows, and `ids.filterNot { it in existingIds }.toSet()` preserves the old "0 rows updated ⇒ local row gone ⇒ report as failure" contract, with the existence read and the writes sharing the transaction so the snapshot can't shift between them.

**Verdict:** Mergeable.

---

dogi:
**Re-review — Rating: 89/100**

Collapses N per-row `UPDATE` round-trips into one `SELECT` plus one batched `@Update`, inside a single transaction. The semantics are preserved exactly, which took some care to arrange.

**Semantic equivalence**

Old: `markUploaded(id, rev)` returned rows-updated; `!= 0` meant success; `UploadConfigs` collected the failures as `results.filter { !success }`.

New: `markUploadedBatch` returns the ids that *didn't* exist, and `UploadConfigs` collects `results.filter { it.localId in unappliedIds }`. Same set. ✓

The indirection is necessary rather than incidental: Room's `@Update(List)` reports a single aggregate row count, so it can't tell you *which* rows were missing. The separate `getExistingIds` query is the only way to keep per-item failure attribution while still batching, and doing it that way is the right call.

**Details that check out**

`@Transaction` on the default method wraps the existence check and the update atomically — so a row deleted between the two can't produce a false "applied". Worth calling out because the sibling PR #17147 does the same chunking without `@Transaction`; this one gets it right.

`getExistingIds` is chunked at 900, correct for a single `IN` clause against SQLite's 999-parameter ceiling. `markUploadedBatchInternal` is deliberately *not* chunked, which is also correct — Room emits one statement per entity for `@Update(List)`, so there's no parameter list to overflow. Getting one chunked and the other not, for the right reason, is the kind of thing that usually goes wrong.

`ids.filterNot { it in existingIds }.toSet()` dedupes, so duplicate ids in a batch can't double-report. The `existingIds` lookup is a `Set`, so the filter is O(n) not O(n²).

`@Update(entity = ApkLog::class)` with a partial `UploadUpdate(id, _rev)` requires the partial entity to carry the full primary key — it compiles, so `id` is the PK.

**The gap**

The DAO method itself has no test. The repository test verifies `DiagnosticsRepositoryImpl` delegates and maps correctly, and the `UploadConfigs` tests verify the batched call and the unapplied-id failure path — but all three mock `apkLogDao`, so the actual interesting logic (existence check + partial-entity update + the returned difference set) is never executed against a real database.

#17147 in this same queue does exactly this well: an in-memory Room database, 1200 rows to cross the chunk boundary, plus empty and mixed-existence cases. The same treatment here would cover the partial-entity update path, the transaction, and the "row deleted before ack" case that this method exists to report on. That's the difference between 89 and mid-90s.

**Verdict:** Mergeable.

---

PR #17046, opened 2026-09-10
teams: smoother activities repository voices member visiting (fixes #17023)

### LINKED ISSUE #17023: Give ActivitiesRepository one call for a member's visit stats
## What

A UI helper object issues two independent repository queries to fill two strings on the member-detail screen.

Files:
- `app/src/main/java/org/ole/planet/myplanet/ui/voices/VoicesActions.kt` (`showMemberDetails`, ~211-231)
- `app/src/main/java/org/ole/planet/myplanet/repository/ActivitiesRepository.kt` (+ impl)

## Why

```kotlin
(userModel.id?.let { activitiesRepository.getOfflineVisitCount(it) } ?: 0).toString(),                    // ~223
(activitiesRepository.getLastVisit(userModel.name ?: "")?.let { dateFormatter.get()?.format(Date(it)) }
    ?: "No logout record found"),                                                                          // ~224
```

Two round trips from a UI object that takes the repository as a parameter, for one screen's two fields.

## The fix

- Add a `MemberVisitStats(offlineVisitCount, lastVisit)` data class next to the existing `ProfileActivityStats`.
- Add `getMemberVisitStats(userId: String?, userName: String?): MemberVisitStats` to `ActivitiesRepository`, implemented in the impl from the same two DAO calls the existing accessors already use.
- Consume it in `showMemberDetails`, keeping the date formatting and the `"No logout record found"` fallback in the UI, and keeping `showMemberDetails`'s signature unchanged so `BaseVoicesFragment` (~104) and `ReplyActivity` (~194) need no edit.

Do **not** remove `getOfflineVisitCount` / `getLastVisit` — `UserProfileViewModel` (~139) and `DashboardViewModel` (~112) still call them.

## Acceptance

- The member-detail screen shows the same visit count and last-visit text, including the fallback when there is no logout record.
- `showMemberDetails` makes one repository call.

## Notes

dogi:
Add MemberVisitStats data class and getMemberVisitStats method to ActivitiesRepository and consume it in VoicesActions.showMemberDetails to consolidate two queries into a single call.

Fixes #17023

---
*PR created automatically by Jules for task [10007386820824880333](https://jules.google.com/task/10007386820824880333) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 7066
-        versionName = "0.70.66"
+        versionCode = 7067
+        versionName = "0.70.67"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/repository/ActivitiesRepository.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/ActivitiesRepository.kt
@@ -14,8 +14,14 @@ data class ProfileActivityStats(
     val resourceOpenCount: Long
 )
 
+data class MemberVisitStats(
+    val offlineVisitCount: Int,
+    val lastVisit: Long?
+)
+
 interface ActivitiesRepository {
     suspend fun getOfflineVisitCount(userId: String): Int
+    suspend fun getMemberVisitStats(userId: String?, userName: String?): MemberVisitStats
     suspend fun getOfflineLoginCount(userName: String): Int
     fun getOfflineLogins(userName: String): Flow<List<OfflineActivity>>
     suspend fun markResourceAdded(userId: String?, resourceId: String)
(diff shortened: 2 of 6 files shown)


dogi:
**Review — rating: 85/100**

Replaces two separate `activitiesRepository` calls in `VoicesActions.showMemberDetails` with a single `getMemberVisitStats(userId, userName)` returning a `MemberVisitStats` record.

The null handling actually gets better rather than just moving. Master called `getLastVisit(userModel.name ?: "")`, which issued a DAO query with an empty-string username that could never match; the new `if (!userName.isNullOrEmpty())` skips it outright. Same for the count — `userModel.id?.let { ... }` would have queried on an empty-but-non-null id, where `!userId.isNullOrEmpty()` now short-circuits to 0. Both produce the same displayed values, with two pointless queries removed in the degenerate case.

The tests are the strongest part. `getMemberVisitStats handles null or empty inputs` covers both the null and `""` paths explicitly, and the Robolectric `showMemberDetails` test asserts the fragment arguments *and* pins the refactor with `coVerify(exactly = 0)` on `getOfflineVisitCount` and `getLastVisit` — so a future regression back to two calls fails the build rather than passing silently.

**Nit:** the title says "combine ... queries", but the two DAO calls still run sequentially inside the new method — what's combined is the repository's API surface, not the database work. If latency here matters, they're independent and would `async`/`awaitAll` cleanly (cf. #17019 doing exactly that in `NotificationsRepositoryImpl`). Worth either doing or not claiming.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17051, opened 2026-09-10
resources: smoother repository view modelling (fixes #17034)

### LINKED ISSUE #17034: Tighten the ResourcesRepository surface: no pass-through, no pure folds, one id resolver
## What

Three boundary problems on `ResourcesRepository`: a one-line pass-through, a pure in-memory fold living in the data layer, and a UI file re-implementing the repository's own id-resolution rule.

Files:
- `app/src/main/java/org/ole/planet/myplanet/repository/ResourcesRepository.kt`
- `app/src/main/java/org/ole/planet/myplanet/repository/ResourcesRepositoryImpl.kt`
- `app/src/main/java/org/ole/planet/myplanet/ui/resources/ResourcesViewModel.kt`
- `app/src/main/java/org/ole/planet/myplanet/ui/resources/ResourceDetailFragment.kt`

## Why

1. `ResourcesRepository.kt:53` exposes `markResourceAdded`; the impl (~307-309) is a one-line delegate to `activitiesRepository.markResourceAdded`, and the only caller is the impl itself (~297).
2. `ResourcesRepository.kt:95` exposes `suspend fun getFilterFacets(libraries: List<MyLibrary>)`, implemented (~612-631) as a pure in-memory fold over a caller-supplied list — no DAO, no API, no prefs — which `ResourcesViewModel.kt:120-121` already wraps in `withContext(dispatcherProvider.default)`.
3. `ResourceDetailFragment.kt:38-40` re-implements the repository's own "id or resourceId" rule as `getLibraryItemById(id) ?: getLibraryItemByResourceId(id)`.

## The fix

- Move the facet fold into `ResourcesViewModel.getFilterFacets` **verbatim** (keeping the `withContext(dispatcherProvider.default)` wrapper) and delete the repository copy from the interface and impl.
- Add `resolveLibraryItem(id)` to the interface and impl encapsulating the "id or resourceId" rule, collapse `fetchLibrary` onto it, and use it from `ResourceDetailFragment`.
- Make the impl's `markResourceAdded` private and drop it from the interface.

`ResourcesFragment.kt:690` calls `viewModel.getFilterFacets(...)` — that signature does **not** change. Leave the impl's internal `getLibraryItemById` / `getLibraryItemByResourceId` accessors in place.

## Acceptance

- Resource filters produce identical facets.
- The resource detail screen resolves the same item for both id shapes.

## Notes

dogi:
Tightens the ResourcesRepository interface boundary by removing pass-through markResourceAdded, moving the pure in-memory filter facet fold to ResourcesViewModel.getFilterFacets, and introducing a unified resolveLibraryItem(id) method across repository and UI callers.

Fixes #17034

---
*PR created automatically by Jules for task [8244862689796070174](https://jules.google.com/task/8244862689796070174) started by @dogi*

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,8 @@ android {
         applicationId "org.ole.planet.myplanet"
         minSdk = 26
         targetSdk = 36
-        versionCode = 7076
-        versionName = "0.70.76"
+        versionCode = 7077
+        versionName = "0.70.77"
         ndkVersion = '26.3.11579264'
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/src/main/java/org/ole/planet/myplanet/repository/ResourcesRepository.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/ResourcesRepository.kt
@@ -50,6 +50,7 @@ interface ResourcesRepository {
     suspend fun countLibrariesNeedingUpdate(userId: String?): Int
     suspend fun resourceTitleExists(title: String): Boolean
     suspend fun resolveLibraryItem(id: String): MyLibrary?
+    suspend fun resolveLibraryItemByResourceId(resourceId: String): MyLibrary?
     suspend fun saveLocalResource(request: LocalResourceRequest): Result<Unit>
     suspend fun updateUserLibrary(resourceId: String, userId: String, isAdd: Boolean): MyLibrary?
     suspend fun setUserLibrary(resourceId: String, add: Boolean): MyLibrary?
--- a/app/src/main/java/org/ole/planet/myplanet/repository/ResourcesRepository.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/ResourcesRepository.kt
@@ -49,8 +49,8 @@ interface ResourcesRepository {
     fun getPendingDownloads(userId: String): Flow<List<String>>
     suspend fun countLibrariesNeedingUpdate(userId: String?): Int
     suspend fun resourceTitleExists(title: String): Boolean
+    suspend fun resolveLibraryItem(id: String): MyLibrary?
     suspend fun saveLocalResource(request: LocalResourceRequest): Result<Unit>
-    suspend fun markResourceAdded(userId: String?, resourceId: String)
     suspend fun updateUserLibrary(resourceId: String, userId: String, isAdd: Boolean): MyLibrary?
     suspend fun setUserLibrary(resourceId: String, add: Boolean): MyLibrary?
     suspend fun updateLibraryItem(id: String, updater: (MyLibrary) -> Unit)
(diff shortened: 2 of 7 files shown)


dogi:
**Review — rating: 72/100**

Two of the three changes here are exactly right, and one silently flips a lookup order in a write path.

**The good.** Moving `getFilterFacets` out of `ResourcesRepositoryImpl` into `ResourcesViewModel` is the correct call — it's pure computation over a list the caller already holds, touches no DAO, and had no business being a repository method. (Worth noting it's the opposite of what #17058 does to `searchChats`, which pulls the same shape of pure filter *into* the repository and adds a query to feed it. This PR resolves the pattern the way I'd argue for.) Making `markResourceAdded` private is likewise a genuine narrowing — it was a bare pass-through to `activitiesRepository`, and the interface is better without it.

**The problem.** `resolveLibraryItem` is defined as `getLibraryItemById(id) ?: getLibraryItemByResourceId(id)`, but only one of the three sites it replaces used that order. `ResourceDetailFragment.fetchLibrary` did (by-id first), while both `setUserLibrary` (line 313) and `updateUserLibrary` (line ~343) did the **reverse**: `getLibraryItemByResourceId(resourceId) ?: getLibraryItemById(resourceId)`.

That isn't a distinction without a difference, because these hit three different columns. `getLibraryItemById` is `WHERE id = :id` (`MyLibraryDao.kt:28`), while `getLibraryItemByResourceId` is `WHERE resourceId = :resourceId` (line 31) falling back to `getByUnderscoreId`. So if the passed value matches row A on `id` and a *different* row B on `resourceId`, master returned B and this returns A. Given that `my_library.id` is frequently populated from the same CouchDB identifier space that `resourceId` points at, that collision is plausible rather than theoretical — and both affected methods are writes that mutate `userId` membership, so picking the wrong row adds or removes the resource from a user's shelf incorrectly.

The parameter names say it too: those two sites receive a **`resourceId`**, so consulting the `resourceId` column first was the semantically correct precedence, and by-id was the fallback.

Either keep both orders (`resolveByResourceId` / `resolveById`, sharing the same two calls), or give `resolveLibraryItem` a parameter for which column wins. If you're confident `id` and `resourceId` can never collide across rows, that's fine too — but then say so in the PR, because the diff currently changes it without acknowledging it.

**Verdict:** Needs another round — the facets move and the privatization are good; please preserve the by-resourceId-first precedence in `setUserLibrary` and `updateUserLibrary`.

---

dogi:

The precedence flip is fixed, and fixed the way that keeps the deduplication. There are now two resolvers rather than one:

- `resolveLibraryItem(id)` — `getLibraryItemById` then `getLibraryItemByResourceId`, used by `ResourceDetailFragment`, matching what `fetchLibrary` did.
- `resolveLibraryItemByResourceId(resourceId)` — `getLibraryItemByResourceId` then `getLibraryItemById`, used by `setUserLibrary` (line 316) and `updateUserLibrary` (line 350), matching master.

So both write paths consult the `resourceId` column first again, which is the semantically correct order for a method whose parameter *is* a resource id, and the wrong-row shelf mutation I was worried about can't happen. Each site still shares one helper instead of repeating the elvis chain.

The other two changes are unchanged and still good: `getFilterFacets` moved into `ResourcesViewModel` where a pure list transform belongs, and `markResourceAdded` is `private` in the impl. Build (default + lite) and both shards green.

**Nit:** `resolveLibraryItemByResourceId` is on the public interface but only called from inside `ResourcesRepositoryImpl`. Given this PR's theme it could stay private, with only `resolveLibraryItem` exposed.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17053, opened 2026-09-10
teams: smoother notifications repository view modelling (fixes #17031)

### LINKED ISSUE #17031: Have NotificationsRepository compute the team chat count itself
## What

The caller passes a filtered-list size into a counter that is later compared against a top-level chat count — two different metrics.

Files:
- `app/src/main/java/org/ole/planet/myplanet/repository/NotificationsRepository.kt` (~25)
- `app/src/main/java/org/ole/planet/myplanet/repository/NotificationsRepositoryImpl.kt` (`updateTeamNotification`, ~285)
- `app/src/main/java/org/ole/planet/myplanet/ui/teams/voices/TeamsVoicesViewModel.kt` (~58-60)

## Why

```kotlin
suspend fun getFilteredNews(teamId: String): List<News?> {
    val newsList = voicesRepository.getFilteredNews(teamId)
    notificationsRepository.updateTeamNotification(teamId, newsList.size)   // ~60
    return newsList
}
```

But `getTeamNotifications` compares the stored `lastCount` against `voicesRepository.countTopLevelByTeam(teamId)` — a *different* metric. `NotificationsRepositoryImpl` already injects `voicesRepository`, so the count belongs inside the repository.

## The fix

- Change the interface and impl to `suspend fun updateTeamNotification(teamId: String)`.
- Compute `voicesRepository.countTopLevelByTeam(teamId)` inside the impl and store that as `lastCount` (it is a `Long`; keep the stored column's type and convert the same way the comparison side does).
- Drop the argument at the `TeamsVoicesViewModel` call site.
- Update `TeamsVoicesViewModelTest` (the `coVerify` on `updateTeamNotification`) and `TeamChatBadgeIntegrationTest` (three call sites that currently pass `countTopLevelByTeam(...).toInt()`).

## Acceptance

- The badge raises and clears against a single consistent metric.
- No caller passes a count any more.

## Notes

This is also a latent badge bug: writing a filtered-list size into a counter compared against a top-level count can clear or re-raise the badge wrongly.

Open PR #17019 touches `NotificationsRepositoryImpl` at ~315 (the chat-count loop) and PR #17020 touches `TeamsVoicesViewModel` at ~28-38 and its test at ~16-41 — both disjoint from `updateTeamNotification` (~285) and the ~60 call site, so this should not collide. A follow-up task will replace `updateTeamNotification`'s read-modify-write with a single `UPDATE`; that is deliberately **not** part of this issue — leave the read-modify-write shape alone here. Conventions: `docs/CODE_STYLE_GUIDE.md`.

dogi:
Updates NotificationsRepository and NotificationsRepositoryImpl to compute the team chat count internally via voicesRepository.countTopLevelByTeam(teamId) in updateTeamNotification(teamId), removing the need for callers to pass a count parameter.

Fixes #17031

---
*PR created automatically by Jules for task [9459409411703150996](https://jules.google.com/task/9459409411703150996) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/repository/NotificationsRepository.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/NotificationsRepository.kt
@@ -22,7 +22,7 @@ interface NotificationsRepository {
     suspend fun getTaskTeamNamesByTaskIds(taskIds: List<String>): Map<String, String>
     suspend fun getJoinRequestDetailsBatch(relatedIds: List<String>): Map<String, Pair<String, String>>
     suspend fun getTeamNotifications(teamIds: List<String>, userId: String): Map<String, TeamNotificationInfo>
-    suspend fun updateTeamNotification(teamId: String, count: Int)
+    suspend fun updateTeamNotification(teamId: String)
     suspend fun getTaskTeamNamesByTaskTitles(taskTitles: List<String>): Map<String, String>
     suspend fun getPendingSyncNotifications(): List<AppNotification>
     suspend fun markNotificationsSynced(syncResults: List<Pair<String, String?>>)
--- a/app/src/main/java/org/ole/planet/myplanet/repository/NotificationsRepository.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/NotificationsRepository.kt
@@ -3,6 +3,7 @@ package org.ole.planet.myplanet.repository
 import com.google.gson.JsonArray
 import com.google.gson.JsonObject
 import org.ole.planet.myplanet.model.AppNotification
+import org.ole.planet.myplanet.model.News
 import org.ole.planet.myplanet.model.NotificationPayload
 import org.ole.planet.myplanet.model.TaskNotificationResult
 import org.ole.planet.myplanet.model.TeamNotificationInfo
@@ -22,7 +23,7 @@ interface NotificationsRepository {
     suspend fun getTaskTeamNamesByTaskIds(taskIds: List<String>): Map<String, String>
     suspend fun getJoinRequestDetailsBatch(relatedIds: List<String>): Map<String, Pair<String, String>>
     suspend fun getTeamNotifications(teamIds: List<String>, userId: String): Map<String, TeamNotificationInfo>
-    suspend fun updateTeamNotification(teamId: String)
+    suspend fun updateTeamNotification(teamId: String, news: List<News>)
     suspend fun getTaskTeamNamesByTaskTitles(taskTitles: List<String>): Map<String, String>
     suspend fun getPendingSyncNotifications(): List<AppNotification>
     suspend fun markNotificationsSynced(syncResults: List<Pair<String, String?>>)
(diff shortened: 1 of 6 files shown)


dogi:
**Review — rating: 74/100**

Not a duplicate of #17019, despite both touching `NotificationsRepositoryImpl.kt` — that one parallelises the read path in `getTeamNotifications`, this one rewrites `updateTeamNotification`. Different hunks, both can land. Rating this on its own terms.

The encapsulation instinct is right: `updateTeamNotification(teamId, count)` forced every caller to know that the watermark basis is "top-level posts visible to this team", which is exactly the kind of implicit contract that rots silently. Dropping the parameter makes the repository own its own invariant, and the two new unit tests plus the `TeamChatBadgeIntegrationTest` simplification are clean.

But I checked whether this actually fixes a wrong watermark, and it doesn't — the two queries are character-for-character the same predicate. `getFilteredNews` → `newsDao.getTopLevelByTeam` (`NewsDao.kt:33`) and `countTopLevelByTeam` (`NewsDao.kt:80`) both filter `(replyTo IS NULL OR replyTo = '') AND ((viewableBy = 'teams' COLLATE NOCASE AND viewableId = :teamId COLLATE NOCASE) OR viewIn LIKE :teamPattern ESCAPE '\')`, and `VoicesRepositoryImpl` passes the same `teamIdPattern(teamId)` to each (lines 270 and 487). So `newsList.size` was already exactly equal to `countTopLevelByTeam(teamId)`; there is no badge bug being corrected here.

What the change does cost is real: `TeamsVoicesViewModel.getFilteredNews` now runs `getTopLevelByTeam` to fetch the rows and then immediately re-scans the identical predicate with a `COUNT(*)`, where the caller already had the exact answer in hand for free. That is the same duplicate-query-on-screen-entry pattern that #17030, #17035 and #17044 are actively removing elsewhere in this queue, so shipping it here pushes against the direction the rest of the work is going.

The invariant is worth protecting — just not by paying for a second query. Either keep the count parameter and pin the contract with the integration test that already covers it, or have the repository take the list it is deriving from (`updateTeamNotification(teamId, news: List<News>)`) so the count stays free while the caller loses the ability to pass a wrong number.

**Verdict:** Needs another round — right goal, but it buys encapsulation with a redundant DB scan when a free version of the same guarantee is available.

---

dogi:
**Re-review — Rating: 87/100** (was 74)

This takes the list-passing option, and it resolves the objection cleanly: `updateTeamNotification(teamId: String, news: List<News>)` derives `count = news.size` inside the repository, so the caller can no longer supply a wrong watermark — but the count comes from the list `TeamsVoicesViewModel.getFilteredNews` is already holding, so there's no second query. The redundant `COUNT(*)` over the identical predicate is gone.

The encapsulation win you were after is fully preserved. The caller no longer has to know that the watermark basis is "top-level posts visible to this team"; it just hands over what it fetched, and the repository decides what to do with it. And because `getFilteredNews` → `newsDao.getTopLevelByTeam` shares its predicate character-for-character with `countTopLevelByTeam`, `news.size` is exactly the value `getTeamNotifications` later compares against — the invariant holds by construction now rather than by coincidence.

`TeamChatBadgeIntegrationTest` reads better for it too: `updateTeamNotification(teamId, voicesRepository.getFilteredNews(teamId))` says "opening the feed writes what the feed showed", which is the actual semantics, where the old `countTopLevelByTeam(...).toInt()` was the test compensating for a leaky signature. Build (default + lite) and both shards green.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17056, opened 2026-09-10
sync: smoother time logger managing (fixes #17038)

### LINKED ISSUE #17038: Evaluate the SyncPerf log gate once per sync instead of per event
## What

`Log.isLoggable("SyncPerf", Log.DEBUG)` is evaluated on every API call and every DB write, so one sync performs thousands of system-property lookups — and the tag literal is repeated ten times.

Files:
- `app/src/main/java/org/ole/planet/myplanet/utils/SyncTimeLogger.kt` (gates at 71, 86, 144, 164, 181, 192)
- `app/src/main/java/org/ole/planet/myplanet/services/sync/SyncManager.kt` (gates at 77, 145, 215, 224)

## Why

`logApiCall` / `logDbOperation` / `logDetail` run per API call and per DB write, and each one re-asks the platform whether the tag is loggable.

## The fix

- Add a `private const val TAG = "SyncPerf"` (or a companion constant) and an `isVerbose` field in `SyncTimeLogger`, refreshed from `Log.isLoggable(TAG, Log.DEBUG)` in `startLogging()`.
- Replace all six gates in `SyncTimeLogger` with `if (isVerbose)`.
- Gate `SyncManager.syncPerf` on `syncTimeLogger.isVerbose`, and route the three inline gates in `SyncManager` through `syncPerf { … }`.
- Keep `stopLogging`'s summary write **unconditional**.

This deliberately snapshots loggability once per sync session rather than per event — state that in a comment so the next reader does not "fix" it back.

Do not add any new log line, and do not interpolate a server URL or any credential-bearing value into a log message — myPlanet URLs embed `satellite:<pin>@host` and release builds keep logs.

## Acceptance

- With the tag not loggable, no `SyncPerf` output and no per-event `isLoggable` call.
- With the tag loggable (set before the sync starts), the same lines as today.
- The end-of-sync summary is still written either way.

## Notes

`SyncManagerTest`'s `UncaughtExceptionsBeforeTest` is a known ordering flake — if it fails, re-run rather than chasing it. Conventions: `docs/CODE_STYLE_GUIDE.md`.

dogi:
Refactored SyncPerf logging evaluation to snapshot loggability once per sync session in `SyncTimeLogger.startLogging()` instead of checking `Log.isLoggable("SyncPerf", Log.DEBUG)` on every API call and database operation. Updated `SyncManager` inline log statements and `syncPerf` helper to gate on `SyncTimeLogger.isVerbose`.

Fixes #17038

---
*PR created automatically by Jules for task [17276266126909069846](https://jules.google.com/task/17276266126909069846) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/services/sync/SyncManager.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/services/sync/SyncManager.kt
@@ -74,9 +74,13 @@ class SyncManager @Inject constructor(
     private val syncTimeLogger: SyncTimeLogger
 ) {
     private inline fun syncPerf(message: () -> String) {
-        if (Log.isLoggable("SyncPerf", Log.DEBUG)) {
-            Log.d("SyncPerf", message())
+        if (syncTimeLogger.isVerbose) {
+            Log.d(TAG, message())
         }
+    }
+
+    companion object {
+        private const val TAG = "SyncPerf"
     }
 
     private val timestampFormat = DateTimeFormatter.ofPattern("HH:mm:ss.SSS").withZone(ZoneId.systemDefault())
@@ -142,11 +146,9 @@ class SyncManager @Inject constructor(
         val syncStartTime = SystemClock.elapsedRealtime()
 
         syncTimeLogger.startLogging()
-        if (Log.isLoggable("SyncPerf", Log.DEBUG)) {
-            Log.d("SyncPerf", "═══════════════════════════════════════════════════════════════")
-            Log.d("SyncPerf", "FULL SYNC STARTED at ${timestampFormat.format(Instant.now())}")
-            Log.d("SyncPerf", "═══════════════════════════════════════════════════════════════")
-        }
+        syncPerf { "═══════════════════════════════════════════════════════════════" }
+        syncPerf { "FULL SYNC STARTED at ${timestampFormat.format(Instant.now())}" }
+        syncPerf { "═══════════════════════════════════════════════════════════════" }
         try {
 
             initializeSync()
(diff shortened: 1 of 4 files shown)


dogi:
**Review — rating: 85/100**

Snapshots `Log.isLoggable("SyncPerf", DEBUG)` into `SyncTimeLogger.isVerbose` at `startLogging()` and swaps ~14 per-event checks over to reading the field, while consolidating the scattered `"SyncPerf"` string literals into `TAG` constants.

The win is proportionate to how often those call sites fire. `logApiCall` and `logDbOperation` run per API request and per DAO batch across a full sync, so `Log.isLoggable` — which reads a system property through JNI on each call — was being hit hundreds to thousands of times per sync purely to decide not to log. Hoisting it to once per session is the right granularity, and the comment on the assignment explains the deliberate trade-off rather than leaving the next reader to guess.

The behavioural cost is bounded and acceptable: flipping the log property mid-sync no longer takes effect until the next `startLogging()`. For a debug-only performance trace that's fine, and the comment says so.

Ordering checks out too — `syncTimeLogger.startLogging()` runs before the first `syncPerf { }` in `syncFullSync`, so `isVerbose` is populated before anything reads it. Converting those inline `if (Log.isLoggable(...)) { Log.d(...) }` blocks in `SyncManager` to the existing `syncPerf { }` inline helper is a nice secondary cleanup — the message strings now sit behind a lambda that isn't evaluated when logging is off, which the old `"TOTAL SYNC TIME: ${minutes}m ${seconds}s"` interpolations only avoided because they were already inside the guard.

**Nit:** `isVerbose` is written on whichever thread calls `startLogging()` and read from the sync's worker coroutines, with no `@Volatile`. Worst case a thread briefly sees the previous session's value and logs (or doesn't) for an event or two — harmless for a trace, but `@Volatile` is free and makes the field honest about being cross-thread.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17057, opened 2026-09-10
teams: smoother voices json images caching (fixes #17037)

### LINKED ISSUE #17037: Memoize the JSON getters on News
## What

`News` re-parses its JSON columns on every property read, even though the entity already carries `@Ignore` fields for exactly that memoization — and one adapter does the memoization by hand instead.

`app/src/main/java/org/ole/planet/myplanet/model/News.kt`

## Why

```kotlin
@Ignore var rawImages: String? = null            // ~77
@Ignore var parsedImagesArray: JsonArray? = null // ~79
...
@get:Ignore
val imagesArray: JsonArray
    get() = if (images == null) JsonArray() else JsonUtils.gson.fromJson(images, JsonArray::class.java)   // ~84-85
```

The getter always parses, ignoring both `@Ignore` fields, while `VoicesAdapter` (~639-641) already does the invalidation-aware memoization by hand: `if (it.parsedImagesArray == null || it.rawImages != it.images)`.

Separately, `isCommunityNews` (~111-124) re-parses `viewIn` and ignores `parsedViewIn`, while `calculateSortDate` (~129) does prefer it.

## The fix

- Make the `imagesArray` getter return `parsedImagesArray` when it is non-null **and** `rawImages == images`; otherwise parse once, store both `parsedImagesArray` and `rawImages`, and return. Keep `JsonArray()` for a null `images`.
- Give `isCommunityNews` the same `parsedViewIn`-first shape that `calculateSortDate` already uses.

Keep `VoicesAdapter`'s existing hand-rolled memoization working — the getter's invalidation rule must match it (compare the raw string, not just null-ness).

## Acceptance

- Same values from `imagesArray` and `isCommunityNews`, including after `images` / `viewIn` are reassigned (the cache must invalidate).
- No behaviour change in the voices list.

## Notes

dogi:
Memoize imagesArray and isCommunityNews JSON getters in News model using existing @Ignore cache fields.

Fixes #17037

---
*PR created automatically by Jules for task [6497726261439308549](https://jules.google.com/task/6497726261439308549) started by @dogi*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved news image handling for faster, safer access and more reliable behavior with missing or invalid image data.
  * Community news detection now handles missing or malformed metadata without errors and correctly prioritizes available information.

* **Tests**
  * Added coverage for image caching, defensive copies, cache updates, null values, and community-news detection scenarios.

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/model/News.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/model/News.kt
@@ -74,19 +74,21 @@ open class News {
     @Ignore
     var rawImageUrls: List<String>? = null
     @Ignore
-    var rawImages: String? = null
-    @Ignore
-    var parsedImagesArray: JsonArray? = null
-    @Ignore
     var parsedSharedTeamName: String? = null
+
+    private data class ImagesCache(val raw: String?, val parsed: JsonArray)
+
+    @Ignore
+    @Volatile
+    private var imagesCache: ImagesCache? = null
 
     @get:Ignore
     val imagesArray: JsonArray
         get() {
             val currentImages = images
-            val cached = parsedImagesArray
-            if (cached != null && rawImages == currentImages) {
-                return cached
+            val cache = imagesCache
+            if (cache != null && cache.raw == currentImages) {
+                return cache.parsed.deepCopy()
             }
             val parsed = if (currentImages == null) {
                 JsonArray()
@@ -97,9 +99,8 @@ open class News {
                     JsonArray()
                 }
             }
-            rawImages = currentImages
-            parsedImagesArray = parsed
-            return parsed
+            imagesCache = ImagesCache(currentImages, parsed)
+            return parsed.deepCopy()
         }
 
     @get:Ignore
--- a/app/src/main/java/org/ole/planet/myplanet/model/News.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/model/News.kt
@@ -82,7 +82,25 @@ open class News {
 
     @get:Ignore
     val imagesArray: JsonArray
-        get() = if (images == null) JsonArray() else JsonUtils.gson.fromJson(images, JsonArray::class.java)
+        get() {
+            val currentImages = images
+            val cached = parsedImagesArray
+            if (cached != null && rawImages == currentImages) {
+                return cached
+            }
+            val parsed = if (currentImages == null) {
+                JsonArray()
+            } else {
+                try {
+                    JsonUtils.gson.fromJson(currentImages, JsonArray::class.java) ?: JsonArray()
+                } catch (e: Exception) {
+                    JsonArray()
+                }
+            }
+            rawImages = currentImages
+            parsedImagesArray = parsed
+            return parsed
+        }
 
     @get:Ignore
     val labelsArray: JsonArray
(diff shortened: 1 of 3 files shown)


dogi:
**Review — rating: 78/100**

Memoizing `imagesArray` is worth doing — it's read during adapter binding, so re-running `gson.fromJson` on every access shows up in scroll cost — and keying the cache on `rawImages == currentImages` correctly handles `images` being a `var` that sync can overwrite. The `isCommunityNews` rewrite is a genuine crash fix too: on master, `gson.fromJson(viewIn, JsonArray::class.java)` returns `null` for a null or empty `viewIn` and the `for (e in array)` immediately NPEs. Guarding on `!viewIn.isNullOrEmpty()` and reusing the existing `parsedViewIn` field (already used the same way at line 153) removes that and stops malformed JSON from propagating. 92 lines of new tests behind it.

Three things I'd like tightened before it lands, all small:

**1. The cache publish is racy.** The reader loads `parsedImagesArray` into `cached` *first*, then compares `rawImages`; the writer assigns `rawImages` first, then `parsedImagesArray`. A reader interleaving between those two writes can hold the **old** parsed array while seeing the **new** raw string, and return a stale `JsonArray` for content that has changed. Two non-volatile fields can't be kept consistent by ordering alone — put both in one immutable holder behind a single `@Volatile` field (`private data class ImagesCache(val raw: String?, val parsed: JsonArray)`) and the read becomes a single reference load that's either wholly old or wholly new.

**2. The cache fields are public `var`.** `rawImages` and `parsedImagesArray` (lines 77–79) are writable by anything holding a `News`. Cache internals should be `private`; as written, any caller can desynchronise them from `images`.

**3. The cached instance escapes.** `imagesArray` now hands out the *same* mutable `JsonArray` on every call, and `VoicesRepositoryImpl.kt:436` does `object.add("images", news.imagesArray)` — Gson stores that by reference, not by copy. Nothing mutates it today so this is latent rather than live, but a cached-and-shared mutable collection is the kind of thing that bites much later. Either return a defensive copy or make the immutability contract explicit at the property.

None of these needs much work, and the crash fix underneath is worth keeping.

**Verdict:** Needs another round — right optimization, but the cache needs to be private, atomically published, and safe to hand out.

---

dogi:

All three findings are addressed, and the third one paid an unexpected dividend.

**Atomic publication.** `rawImages` and `parsedImagesArray` are replaced by a single `@Volatile private var imagesCache: ImagesCache?` holding both halves in one immutable record. The read is now one reference load, so the interleaving where a reader saw the new raw string alongside the old parsed array is structurally impossible rather than merely unlikely.

**Encapsulation.** Both fields are gone from the public surface; `ImagesCache` is a `private data class` and the field is `private`.

**Escape.** `imagesArray` returns `cache.parsed.deepCopy()`, so the cached instance never leaves the object — `VoicesRepositoryImpl.kt:436`'s `object.add("images", news.imagesArray)` now stores a copy, and a downstream mutation can't corrupt the cache.

The dividend: because the model now owns the caching properly, `VoicesAdapter` could delete its hand-rolled prime block (`if (it.parsedImagesArray == null || it.rawImages != it.images) { … }`) and the `parsedImagesArray ?: imagesArray` fallback at the call site collapses to plain `news?.imagesArray`. External code was compensating for the model's missing memoization; now it doesn't have to. That's a better outcome than the narrow fix I asked for.

The `isCommunityNews` crash fix on null/empty `viewIn` is unchanged and still the most user-visible part. Build (default + lite) and both shards green.

**Nit:** `deepCopy()` on every read trades some of the win back — you skip the Gson parse but pay a structural copy. Still far cheaper than parsing, and the right default for safety; worth revisiting only if a profile says otherwise.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17058, opened 2026-09-10
chat: smoother repository utils searching (fixes #17042)

### LINKED ISSUE #17042: Stop ChatRepository.searchChats round-tripping the chat list through the caller
## What

`searchChats` takes the chat list as a parameter, so the ViewModel hands the repository back a list it got from that same repository — on every keystroke.

Files:
- `app/src/main/java/org/ole/planet/myplanet/repository/ChatRepository.kt` (~24)
- `app/src/main/java/org/ole/planet/myplanet/repository/ChatRepositoryImpl.kt` (~157)
- `app/src/main/java/org/ole/planet/myplanet/ui/chat/ChatViewModel.kt` (~160)

## Why

```kotlin
suspend fun searchChats(query: String, mode: ChatSearchMode, chats: List<ChatHistory>): List<ChatHistory>
```

```kotlin
val results = chatRepository.searchChats(query, mode, allChats)   // ChatViewModel ~160
```

`allChats` came from `getChatHistoryForUser`. The repository owns `ChatDao` (it already does `chatDao.getByUser(userName)` at ~134), so the list parameter is a leaked caching concern.

## The fix

- Change the signature to `searchChats(query: String, mode: ChatSearchMode, userName: String?): List<ChatHistory>` in the interface and impl.
- Inside the impl, load the chats via the existing `chatDao.getByUser(userName)` path, then feed the existing `buildPrecomputedChats` → `searchByTitle` / `fullConvoSearch` pipeline unchanged.
- Update the `ChatViewModel` call site to pass the user name it already has.
- Keep the ViewModel's fragment-facing API unchanged so `ChatHistoryFragment` needs no edit.

## Acceptance

- Title, question and response search return the same results as today for the same data.
- No caller passes a chat list any more.
- `testDefaultDebugUnitTest` passes (update `ChatRepositoryImplTest` / `ChatViewModelTest` call sites as needed).

## Notes

A debounce/memoization follow-up is worth doing later — this change moves the per-keystroke DB read into the repository, which is the right place to memoize it. Do **not** add caching in this PR. Conventions: `docs/CODE_STYLE_GUIDE.md`.

dogi:
Refactored ChatRepository.searchChats signature to take userName: String? rather than requiring the caller to pass the chat list. Updated ChatViewModel and unit tests accordingly.

Fixes #17042

---
*PR created automatically by Jules for task [11784791100903403165](https://jules.google.com/task/11784791100903403165) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/repository/ChatRepository.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/ChatRepository.kt
@@ -21,7 +21,7 @@ interface ChatRepository {
     suspend fun sendContinueChatRequest(message: String, user: String?, aiProvider: AiProvider, id: String, rev: String): ChatResult
     suspend fun fetchAiProviders(serverUrl: String): Map<String, Boolean>?
     suspend fun getChatHistoryForUser(userName: String?): List<ChatHistory>
-    suspend fun searchChats(query: String, mode: ChatSearchMode, chats: List<ChatHistory>): List<ChatHistory>
+    suspend fun searchChats(query: String, mode: ChatSearchMode, userName: String?): List<ChatHistory>
     suspend fun getLatestRev(id: String): String?
     suspend fun insertChatHistoryList(chats: List<JsonObject>)
     fun extractSharedViewInIds(sharedNews: List<News>): Map<String, Set<String>>
--- a/app/src/main/java/org/ole/planet/myplanet/repository/ChatRepository.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/ChatRepository.kt
@@ -21,7 +21,6 @@ interface ChatRepository {
     suspend fun sendContinueChatRequest(message: String, user: String?, aiProvider: AiProvider, id: String, rev: String): ChatResult
     suspend fun fetchAiProviders(serverUrl: String): Map<String, Boolean>?
     suspend fun getChatHistoryForUser(userName: String?): List<ChatHistory>
-    suspend fun searchChats(query: String, mode: ChatSearchMode, userName: String?): List<ChatHistory>
     suspend fun getLatestRev(id: String): String?
     suspend fun insertChatHistoryList(chats: List<JsonObject>)
     fun extractSharedViewInIds(sharedNews: List<News>): Map<String, Set<String>>
--- a/app/src/main/java/org/ole/planet/myplanet/repository/ChatRepositoryImpl.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/ChatRepositoryImpl.kt
@@ -38,13 +38,6 @@ class ChatRepositoryImpl @Inject constructor(
     @PlainGson private val gson: Gson
 ) : ChatRepository, ChatSyncWriter {
 
-    private data class PrecomputedChat(
-        val chat: ChatHistory,
-        val normalizedTitle: String?,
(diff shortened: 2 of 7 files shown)


dogi:
**Review — rating: 60/100**

The smell you're fixing is real — a repository method that takes `chats: List<ChatHistory>` as a parameter is asking the caller to supply the data the repository is supposed to own. But moving the fetch inside turns an in-memory filter into a database query on every keystroke.

`ChatViewModel` already holds the full list. `allChats` is a field (line 63), populated in `loadChatHistoryScreenData` at line 137 from the very same `chatRepository.getChatHistoryForUser(currentUser?.name)` call. After this change, `searchChats` sends `cachedUser?.name` to the repository, which re-runs `getChatHistoryForUser` → `chatDao.getByUser(userName)` and then `buildPrecomputedChats` over the result. Since `searchChats` is driven by search-field input, that's a full re-read of the user's chat history per typed character, replacing a filter over a list sitting in a field.

There's a consistency wrinkle too: the blank-query early return at line 148 still serves `allChats`, so clearing the box now shows a different snapshot from the one the last search filtered.

The signature is worth changing — I'd just go the other way. This is pure filtering logic over data the caller owns; it doesn't need repository access at all, so it reads better as a pure function (a `ChatSearch` helper or an extension) that keeps taking the list. If you'd rather the repository own it, give it a cached/`Flow`-backed history so a search doesn't re-hit the DAO, and route the blank-query path through the same source so both branches agree.

Related: I've raised the mirror-image of this on #17053, where `updateTeamNotification` re-queries a count the caller already had. Same instinct, same cost — worth settling the pattern once across both.

**Verdict:** Needs another round — right diagnosis, but the cure adds a per-keystroke query.

---

dogi:
**Re-review — Rating: 88/100** (was 60)

This took the pure-function route and it's the right resolution. `searchChats` is off `ChatRepository` entirely, the search machinery (`PrecomputedChat`, `buildPrecomputedChats`, `searchByTitle`, `fullConvoSearch`) moved verbatim into a new `utils/ChatSearch` object, and `ChatViewModel` calls `ChatSearch.search(query, mode, allChats, dispatcherProvider.default)`.

Both problems I raised are gone:

- **No per-keystroke query.** The search runs against `allChats`, the field already populated at `ChatViewModel.kt:137`. Typing no longer re-reads `chatDao.getByUser` per character.
- **No split-brain snapshot.** The blank-query early return and the search path now read the same `allChats`, so clearing the box can't show a different set from the one just filtered.

Moving it out of the repository is the better half of the change. This logic never touched a DAO — it's string normalisation and ranking over a list — so `ChatRepository` shouldn't have carried it, and `ChatRepositoryImpl` sheds ~80 lines. Threading the dispatcher in as a parameter keeps `withContext` off a hardcoded `Dispatchers.Default` at the production call site while leaving the function testable, which the new 75-line `ChatSearchTest` uses. The four-bucket ranking (`inTitleStartQuery + inTitleContainsQuery + startsWithQuery + containsQuery`) is carried across unchanged, so result ordering is preserved. Build (default + lite) and both shards green.

**Nit:** `utils/ChatSearch` imports `ChatSearchMode` from `repository/`, so a util now depends on the repository package for what is really just an enum. Moving `ChatSearchMode` alongside `ChatSearch` (or into `model/`) would leave the dependency pointing the conventional way.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17062, opened 2026-09-10
courses: smoother submissions repository row projecting (fixes #17043)

### LINKED ISSUE #17043: Move submission-list enrichment behind SubmissionsRepository
## What

`SubmissionViewModel` does cross-feature data assembly: grouping, newest-per-group selection, count maps and a batch user lookup to resolve submitter names.

Files:
- `app/src/main/java/org/ole/planet/myplanet/ui/submissions/SubmissionViewModel.kt` (~67-88)
- `app/src/main/java/org/ole/planet/myplanet/repository/SubmissionsRepository.kt` (+ impl)

## Why

```kotlin
for (group in filtered.groupBy { it.parentId }.values) {
    val newest = group.maxByOrNull { it.lastUpdateTime } ?: continue
    uniqueRawSubmissions.add(newest)
    submissionCountMap[newest.id] = group.size
}
val userIds = uniqueRawSubmissions.mapNotNull { it.userId }.distinct()
val fallbackUsersMap = userRepository.getUsersByIds(userIds).associateBy { it.id }
...
val name = submissionsRepository.getNormalizedSubmitterName(sub)
val fallback = sub.userId?.let { fallbackUsersMap[it]?.name }
```

The ViewModel reaches into `UserRepository` purely to fill in submitter names for a submissions list.

## The fix

- Add a platform-free list-row projection to `SubmissionsRepository` — submission + resolved submitter name + group count.
- Add one repository operation that filters by user/type/query, groups by `parentId`, selects the newest per group, batch-resolves the fallback users, and returns the projections sorted by `lastUpdateTime` descending (same order as today).
- Reduce the ViewModel to combining its filter inputs and mapping projections to `SubmissionUiModel`; keep `flowOn(dispatcherProvider.io)` / `shareIn` behaviour as it is.
- Keep `UserRepository` in the ViewModel **only** if active-user resolution still needs it; otherwise drop the injection.

## Acceptance

- The submissions list shows the same rows, the same submitter names (including the fallback and the `""` default) and the same per-group counts, in the same order.
- The exams/`Triple` consumer downstream still receives what it needs.

## Notes

Conventions: `docs/CODE_STYLE_GUIDE.md` (repositories return plain entities/data classes; no `android.*` in the repository layer).

dogi:
Moved submission-list enrichment logic (filtering, grouping by parentId, newest-per-group selection, batch user lookup for submitter names, and count calculation) from SubmissionViewModel into SubmissionsRepository behind a platform-free list-row projection (`SubmissionRowProjection`). Refactored SubmissionViewModel to delegate projection creation to SubmissionsRepository.getSubmissionProjections.

Fixes #17043

---
*PR created automatically by Jules for task [15056084168231445396](https://jules.google.com/task/15056084168231445396) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/model/SubmissionRowProjection.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/model/SubmissionRowProjection.kt
@@ -1,0 +1,7 @@
+package org.ole.planet.myplanet.model
+
+data class SubmissionRowProjection(
+    val submission: Submission,
+    val submitterName: String,
+    val submissionCount: Int,
+)
--- a/app/src/main/java/org/ole/planet/myplanet/repository/SubmissionsRepository.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/SubmissionsRepository.kt
@@ -11,12 +11,20 @@ import org.ole.planet.myplanet.model.StepExam
 import org.ole.planet.myplanet.model.Submission
 import org.ole.planet.myplanet.model.SubmissionDetail
 import org.ole.planet.myplanet.model.SubmissionItem
+import org.ole.planet.myplanet.model.SubmissionRowProjection
 import org.ole.planet.myplanet.model.SubmitPhotos
 import org.ole.planet.myplanet.model.UserEntity
 
 interface SubmissionsRepository {
     fun getPendingSurveysFlow(userId: String?): Flow<List<Submission>>
     fun getSubmissionsFlow(userId: String): Flow<List<Submission>>
+    suspend fun getSubmissionProjections(
+        submissions: List<Submission>,
+        userId: String,
+        type: String,
+        query: String,
+        examMap: Map<String?, StepExam>,
+    ): List<SubmissionRowProjection>
     suspend fun getPendingSurveys(userId: String?): List<Submission>
     suspend fun getUniquePendingSurveys(userId: String?): List<Submission>
     suspend fun getSurveyTitlesFromSubmissions(submissions: List<Submission>): List<String>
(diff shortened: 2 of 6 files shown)


dogi:
**Review — rating: 85/100**

Moves the filter/sort/group/dedupe/name-resolve pipeline out of `SubmissionViewModel`'s five-way `combine` and into `SubmissionsRepository.getSubmissionProjections`, behind a `SubmissionRowProjection` type.

This one belongs in the repository, unlike some of the neighbouring PRs in this queue, and the distinction is worth naming: the enrichment isn't pure. It needs `userRepository.getUsersByIds` for the submitter-name fallback and `getNormalizedSubmitterName`, so hosting it in the repository is what lets the ViewModel stop reaching for `UserRepository` mid-transform. (Compare #17051, which correctly moves `getFilterFacets` the *other* way because that one touches no data source at all.)

I read the moved block against the original and it's faithful — the same three `type` branches, the same `sortedByDescending { lastUpdateTime }` before and after grouping, the same `maxByOrNull` per `parentId` group with `submissionCountMap[newest.id] = group.size`, and the same `name ?: fallback ?: ""` precedence. The one thing that changes is that the old `Triple(uniqueSubmissions, submissionCountMap, filtered)` computed and returned `filtered` as a third element the consumer never destructured; folding the count onto the projection drops both the dead element and the separate map lookup at the call site. `.flowOn(dispatcherProvider.io)` stays where it was.

Adding `UserRepository` to `SubmissionsRepositoryImpl` as a direct dependency (rather than the `Provider<TeamsRepository>` treatment already there for the cyclic case) is fine given Hilt resolved it — a cycle would have failed the KSP step, not passed both build flavors.

**Nit:** `getNormalizedSubmitterName` is now only called from inside the impl, so it could come off the `SubmissionsRepository` interface in the same spirit as the rest of this change.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17063, opened 2026-09-10
all: smoother user information view modelling (fixes #17055)

### LINKED ISSUE #17055: Hoist UserInformationFragment's repository access into a ViewModel
## What

A dialog fragment injects two repositories and drives writes from its own view lifecycle, so those writes die with the dialog on rotation.

`app/src/main/java/org/ole/planet/myplanet/ui/exam/UserInformationFragment.kt`

## Why

`UserInformationFragment` is a `BaseDialogFragment` injecting `submissionsRepository` (~42) and `userRepository` (~44) and launching from `viewLifecycleOwner.lifecycleScope`:

- `userRepository.updateProfileFields(userId, user)` at ~161-163
- `submissionsRepository.markSubmissionComplete(submissionId, user)` at ~258-271

Both are one-shot writes tied to a dialog's view lifecycle, and this is the only screen in `ui/exam/` with no ViewModel.

## The fix

- Add `UserInformationViewModel` (`@HiltViewModel`, both repositories injected) exposing `updateProfile(...)` and `markSubmissionComplete(...)`, each launched in `viewModelScope` and emitting its outcome on a result flow (`SharedFlow`, or a `StateFlow` of a sealed result — follow whatever the neighbouring exam/submission ViewModels do).
- Collect that flow in the fragment and keep every existing toast, `dismiss()` and navigation side effect firing on the same outcomes as today.
- Remove the two `@Inject` repository fields from the fragment.

## Acceptance

- Saving user information and completing a submission still work, with the same messages and the same dialog dismissal.
- The write survives a configuration change instead of being cancelled with the dialog's view.

## Notes

Conventions: `docs/CODE_STYLE_GUIDE.md` (ViewModel owns repository access; fragments collect state).

dogi:
Hoisted repository access from UserInformationFragment into UserInformationViewModel so writes survive configuration changes. Updated fragment to observe operation results via SharedFlow. Added unit tests for the ViewModel.

Fixes #17055

---
*PR created automatically by Jules for task [16146362856025865306](https://jules.google.com/task/16146362856025865306) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/ui/exam/UserInformationFragment.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/exam/UserInformationFragment.kt
@@ -13,6 +13,7 @@ import android.widget.ArrayAdapter
 import android.widget.RadioButton
 import androidx.appcompat.app.AppCompatActivity
 import androidx.core.view.isVisible
+import androidx.fragment.app.viewModels
 import androidx.lifecycle.lifecycleScope
 import com.google.gson.JsonObject
 import dagger.hilt.android.AndroidEntryPoint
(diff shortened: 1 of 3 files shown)


dogi:
**Review — rating: 86/100**

Extracts `UserInformationViewModel` with a `UserInformationResult` sealed type and moves both repository calls off `UserInformationFragment`, which drops its `SubmissionsRepository` and `UserRepository` field injections.

The part that makes this more than a tidy-up: the work moves from `viewLifecycleOwner.lifecycleScope` to `viewModelScope`. On master, `submissionsRepository.markSubmissionComplete(submissionId, user)` ran in the *view* scope, so if the dialog's view was torn down while that write was in flight — which is plausible here, since the success path's whole job is to dismiss the dialog — the coroutine was cancelled mid-write. In `viewModelScope` the write runs to completion regardless. For a "thank you for taking this survey" completion marker, that's a real durability improvement, not just a layering one.

The null-id branch is also better placed. It used to be an early `return@launch` inside the coroutine; now it's a `MarkSubmissionError("no ID provided")` the ViewModel emits before touching the repository, and the fragment still renders the same "Unable to save submission - no ID provided" string. `MutableSharedFlow(extraBufferCapacity = 1)` keeps `emit` from suspending, and the existing `collectWhenStarted` helper is the right collector. 104 lines of new ViewModel tests, where previously this logic was untestable inside the fragment.

**Nit:** `resultEvent` has `replay = 0`, so if the result lands while the fragment is below STARTED the event is dropped and the user gets no toast (and no `dialog?.dismiss()`). The write itself still completes, which is the important half, so this is a small UX gap rather than data loss — but if you want it airtight, a `Channel` with `receiveAsFlow()` buffers until a collector attaches.

**Note on CI:** as of this review only `myPlanet labels`, CodeRabbit and qlty have reported on `e71305e` — neither `myPlanet build` nor `myPlanet test` has posted a result. That's an *absent* verdict rather than a failing one, so I've rated the diff on its merits, but this shouldn't be merged until those two come back green.

**Verdict:** Mergeable once build and test report green.

---

Okuro3499 (approved):

PR #17067, opened 2026-09-10
teams: smoother surveys repository exam dao querying (fixes #17049)

### LINKED ISSUE #17049: Filter individual surveys in Room instead of in the repository
## What

Every survey row crosses the data boundary so two flags can be checked in Kotlin.

Files:
- `app/src/main/java/org/ole/planet/myplanet/repository/SurveysRepositoryImpl.kt` (~268-270)
- `app/src/main/java/org/ole/planet/myplanet/data/room/dao/ExamDao.kt`

## Why

```kotlin
override suspend fun getIndividualSurveys(): List<StepExam> {
    return examDao.getByType("surveys")
        .filter { !it.isTeamShareAllowed && it.teamId.isNullOrEmpty() }
}
```

## The fix

- Add a DAO query constrained to `type = 'surveys' AND isTeamShareAllowed = 0 AND (teamId IS NULL OR teamId = '')` and delegate to it, with no Kotlin filter left.
- Match Room's stored boolean representation (integer 0/1) and preserve the current row ordering — if `getByType` has no `ORDER BY`, keep the new query equally unordered rather than inventing one.
- Keep the general `getByType` — other operations still need all surveys.

## Acceptance

- `getIndividualSurveys()` returns the same rows in the same order as today, including rows with a null vs empty-string `teamId`.
- No Kotlin-side filtering remains in that method.

## Notes

Use `IS NULL OR = ''` rather than `=` against a nullable column — `=` never matches `NULL` in SQL. Conventions: `docs/CODE_STYLE_GUIDE.md`.

dogi:
This change filters individual surveys in Room instead of fetching all surveys and filtering them in Kotlin in SurveysRepositoryImpl.

- Added `getIndividualSurveys` query to `ExamDao` filtering by `type = :type AND isTeamShareAllowed = 0 AND (teamId IS NULL OR teamId = '')`.
- Updated `SurveysRepositoryImpl.getIndividualSurveys()` to delegate directly to `examDao.getIndividualSurveys()`.
- Updated repository unit tests and added Room DAO query test in `ExamDaoTest`.

Fixes #17049

---
*PR created automatically by Jules for task [9941909336418639181](https://jules.google.com/task/9941909336418639181) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/data/room/dao/ExamDao.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/data/room/dao/ExamDao.kt
@@ -29,6 +29,7 @@ interface ExamDao {
     @Query("SELECT * FROM exams WHERE type = :type AND (teamId = :teamId OR id IN (:submissionIds))") suspend fun getTeamOwnedSurveys(teamId: String, submissionIds: Collection<String>, type: String = "surveys"): List<StepExam>
     @Query("SELECT * FROM exams WHERE type = :type AND isTeamShareAllowed = 1 AND id NOT IN (:excludedIds)") suspend fun getAdoptableTeamSurveys(excludedIds: Collection<String>, type: String = "surveys"): List<StepExam>
     @Query("SELECT * FROM exams WHERE type = :type AND isTeamShareAllowed = 1") suspend fun getAdoptableTeamSurveys(type: String = "surveys"): List<StepExam>
+    @Query("SELECT * FROM exams WHERE type = :type AND isTeamShareAllowed = 0 AND (teamId IS NULL OR teamId = '')") suspend fun getIndividualSurveys(type: String = "surveys"): List<StepExam>
     @Query("DELETE FROM exams WHERE id = :id") suspend fun deleteById(id: String): Int
     @Upsert suspend fun upsert(item: StepExam)
     @Upsert suspend fun upsertAll(items: List<StepExam>)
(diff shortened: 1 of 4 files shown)


dogi:
**Review — rating: 90/100**

Pushes the individual-survey filter out of Kotlin and into the Room query, replacing `examDao.getByType("surveys").filter { !it.isTeamShareAllowed && it.teamId.isNullOrEmpty() }` with a dedicated `getIndividualSurveys()`.

I checked the predicate translates exactly:

| Kotlin | SQL |
|---|---|
| `!it.isTeamShareAllowed` | `isTeamShareAllowed = 0` |
| `it.teamId.isNullOrEmpty()` | `teamId IS NULL OR teamId = ''` |
| `getByType("surveys")` | `type = :type` (default `"surveys"`) |

The null half is the part that's easy to get wrong, and it's right here — `teamId IS NULL` rather than `teamId = NULL`, which would silently match nothing and quietly drop every team-less survey. That's exactly the trap `CLAUDE.md` calls out for DAO predicates.

The win is that the database stops materialising every survey row into `StepExam` objects just to discard most of them. It also stops sending non-`surveys` rows across at all.

Test coverage actually improves rather than regressing. `SurveysRepositoryImplTest` drops to a delegation check, but `ExamDaoTest.getIndividualSurveys_filtersCorrectly` replaces it with a real Room instance covering all five branches — shareable excluded, non-empty `teamId` excluded, `null` and `''` both included, and a wrong-`type` row excluded. Moving the filter into SQL and the assertion onto real SQLite is a strictly better place for that logic to be tested.

**Verdict:** Mergeable — clean translation, and the DAO test proves it against a real database rather than a mock.

---

Okuro3499 (approved):

PR #17068, opened 2026-09-10
life: less view model application context is more (fixes #17054)

### LINKED ISSUE #17054: Drop LifeViewModel's @ApplicationContext by passing the label resolver in
## What

`LifeViewModel` injects `@ApplicationContext` for one purpose: to hand `MyLife.defaultItems` a string resolver.

Files:
- `app/src/main/java/org/ole/planet/myplanet/ui/life/LifeViewModel.kt` (~21, ~41)
- `app/src/main/java/org/ole/planet/myplanet/ui/life/LifeFragment.kt`

## Why

```kotlin
class LifeViewModel @Inject constructor(
    @ApplicationContext private val context: Context,   // ~21
    ...
lifeRepository.getMyLifeByUserId(userId, MyLife.defaultItems(userId, context::getString))   // ~41
```

`MyLife.defaultItems(userId, resolveLabel: (Int) -> String)` (`model/MyLife.kt:67`) already takes a resolver lambda, so the ViewModel only needs the lambda — not a `Context`.

## The fix

- Remove the `@ApplicationContext` injection from `LifeViewModel`.
- Have `loadMyLifeList` take (or the ViewModel hold) a `(Int) -> String` label resolver supplied by `LifeFragment` as `requireContext()::getString`, so a lambda crosses the boundary instead of a `Context`.
- Keep the rest of `loadMyLifeList` — the `withContext(dispatcherProvider.io)` block and the `_myLifeList` write — as it is.

## ⚠️ Do NOT swap `resolveUserId()` in this change

An earlier note suggested replacing `resolveUserId()` (~31-36) with `UserRepository.getActiveUserIdSuspending()`. That is **not** equivalent, and I verified it against master:

```kotlin
// LifeViewModel.resolveUserId()
val raw = userRepository.getCurrentUserId().orEmpty()
    .ifEmpty { userRepository.getUserModel()?.id.orEmpty() }
return raw.takeIf { it.isNotBlank() && it != "--" }
```

```kotlin
// UserRepositoryImpl.getActiveUserIdSuspending()  (~819-821)
return getUserModel()?.id ?: ""
```

The shared helper drops the `getCurrentUserId()` primary source, drops the `"--"` rejection, and returns `""` where the ViewModel returns `null`. Leave `resolveUserId()` exactly as it is here; unifying the two needs the shared helper extended first, as a separate task.

## Acceptance

- `LifeViewModel` no longer injects `Context` and no longer imports `@ApplicationContext`.
- The My Life screen shows the same default items with the same labels.
- User-id resolution behaviour is byte-for-byte unchanged (including `"--"` and the null return).

## Notes

dogi:
Removed `@ApplicationContext private val context: Context` from `LifeViewModel` constructor and updated `loadMyLifeList` to accept `resolveLabel: (Int) -> String`. Updated `LifeFragment` to pass `requireContext()::getString` when calling `loadMyLifeList`, and updated `LifeViewModelTest` to reflect these changes.

Fixes #17054

---
*PR created automatically by Jules for task [17471216513270375939](https://jules.google.com/task/17471216513270375939) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/ui/life/LifeFragment.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/life/LifeFragment.kt
@@ -83,7 +83,7 @@ class LifeFragment : BaseRecyclerFragment<MyLife?>(), OnStartDragListener {
         collectWhenStarted(viewModel.myLifeList) { list ->
             lifeAdapter.submitList(list)
         }
-        viewModel.loadMyLifeList()
+        viewModel.loadMyLifeList(requireContext()::getString)
     }
 
     override fun onDestroyView() {
--- a/app/src/main/java/org/ole/planet/myplanet/ui/life/LifeViewModel.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/life/LifeViewModel.kt
@@ -1,10 +1,8 @@
 package org.ole.planet.myplanet.ui.life
 
-import android.content.Context
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
 import dagger.hilt.android.lifecycle.HiltViewModel
-import dagger.hilt.android.qualifiers.ApplicationContext
 import javax.inject.Inject
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
@@ -18,7 +16,6 @@ import org.ole.planet.myplanet.utils.DispatcherProvider
 
 @HiltViewModel
 class LifeViewModel @Inject constructor(
-    @ApplicationContext private val context: Context,
     private val lifeRepository: LifeRepository,
     private val userRepository: UserRepository,
     private val dispatcherProvider: DispatcherProvider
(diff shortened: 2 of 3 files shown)


dogi:
**Review — rating: 84/100**

Drops `@ApplicationContext Context` from `LifeViewModel` and takes a `resolveLabel: (Int) -> String` on `loadMyLifeList` instead. Getting `Context` out of a ViewModel is the right direction, and the test cleanup shows the payoff — `LifeViewModelTest` loses its `mockk<Context>(relaxed = true)` in favour of `{ "mock_string_$it" }`, which makes the default-items assertion at line 72 actually meaningful rather than comparing two relaxed-mock outputs.

One thing worth being deliberate about, because it changes behaviour in both directions. The old code resolved strings against the **application** context; the new call site passes `requireContext()::getString`, which is the **Activity** context. That's a genuine improvement for correctness — the Activity carries the up-to-date configuration, so a locale change is reflected where the application context could serve stale strings, and this app ships five translations.

The cost is that the bound method reference captures the Activity, and `loadMyLifeList` hands it into a `viewModelScope` coroutine that outlives the fragment. The capture only lives for the duration of the load and the lambda isn't stored on the ViewModel, so this is a brief window rather than a retained leak — but it is an Activity reference held by a scope that survives configuration changes, which is exactly the shape a leak detector flags.

**Nit:** if you want the locale correctness without the capture, resolve the labels at the call site and pass the finished `List<MyLife>` in, so nothing `Context`-bound crosses into `viewModelScope` at all.

**Verdict:** Mergeable — the ViewModel is cleaner and better tested for it.

---

Okuro3499 (approved):

PR #17069, opened 2026-09-10
all: smoother tags repository dao querying (fixes #17059)

### LINKED ISSUE #17059: Stop scanning the whole tag table to build the parent-children map
## What

Building the tag parent→children map loads every row of the `tag` table.

Files:
- `app/src/main/java/org/ole/planet/myplanet/repository/TagsRepositoryImpl.kt` (`getTagsWithChildren`, ~17-39)
- `app/src/main/java/org/ole/planet/myplanet/data/room/dao/TagDao.kt`

## Why

```kotlin
override suspend fun getTagsWithChildren(dbType: String?): Map<TagEntity, List<TagEntity>> {
    val parentTags = getTags(dbType)
    val allTags = tagDao.getAll()        // every row in `tag`
    ...
```

Only attached rows can ever contribute a child, so the unattached rows are loaded and skipped. `TagDao` has scoped queries (`getByDbAndLinkIds`, `getByIds`) but nothing scoped for attached rows.

## The fix

- Add to `TagDao`:
  ```kotlin
  @Query("SELECT * FROM tag WHERE isAttached = 1 AND (:db IS NULL OR db = :db)")
  suspend fun getAttachedForDb(db: String?): List<TagEntity>
  ```
- Use it in place of `tagDao.getAll()` in `getTagsWithChildren`, keeping `getTags(dbType)` for the parents.
- Keep the Kotlin grouping loop as it is, including the `if (list.isEmpty() || list.last() !== t)` guard — it only ever fires for a duplicate parent id inside one tag's `attachedTo`, and removing it is a separate decision.

Note `getTagsWithChildren` calls `getTags(dbType)`, not `getParentTags` — read the current code rather than trusting any older description.

## Acceptance

- The same parent→children map for the same data, including parents with no children (still `emptyList()`).
- No `tagDao.getAll()` call remains in this method.

## Notes

Open PR #17016 also touches `TagsRepositoryImpl`, but only appends `getCourseLinkIds` at the end of the file (~151+) plus one interface line; your change is at ~17-39, so it should not collide. Use `(:db IS NULL OR db = :db)` — `=` never matches `NULL` in SQL. Conventions: `docs/CODE_STYLE_GUIDE.md`.

dogi:
Optimized getTagsWithChildren in TagsRepositoryImpl to query attached tags directly via TagDao.getAttachedForDb(dbType) instead of scanning the full tag table using tagDao.getAll().

Fixes #17059

---
*PR created automatically by Jules for task [15804309183018343278](https://jules.google.com/task/15804309183018343278) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/data/room/dao/TagDao.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/data/room/dao/TagDao.kt
@@ -15,6 +15,9 @@ interface TagDao {
     )
     suspend fun getParentTags(db: String?): List<TagEntity>
 
+    @Query("SELECT * FROM tag WHERE isAttached = 1 AND (:db IS NULL OR db = :db)")
+    suspend fun getAttachedForDb(db: String?): List<TagEntity>
+
     @Query("SELECT * FROM tag")
     suspend fun getAll(): List<TagEntity>
 
--- a/app/src/main/java/org/ole/planet/myplanet/data/room/dao/TagDao.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/data/room/dao/TagDao.kt
@@ -15,8 +15,8 @@ interface TagDao {
     )
     suspend fun getParentTags(db: String?): List<TagEntity>
 
-    @Query("SELECT * FROM tag WHERE isAttached = 1 AND (:db IS NULL OR db = :db)")
-    suspend fun getAttachedForDb(db: String?): List<TagEntity>
+    @Query("SELECT * FROM tag WHERE isAttached = 1")
+    suspend fun getAttached(): List<TagEntity>
 
     @Query("SELECT * FROM tag")
     suspend fun getAll(): List<TagEntity>
--- a/app/src/main/java/org/ole/planet/myplanet/repository/TagsRepositoryImpl.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/TagsRepositoryImpl.kt
@@ -16,10 +16,10 @@ class TagsRepositoryImpl @Inject constructor(
 
     override suspend fun getTagsWithChildren(dbType: String?): Map<TagEntity, List<TagEntity>> {
         val parentTags = getTags(dbType)
-        val allTags = tagDao.getAll()
+        val attachedTags = tagDao.getAttachedForDb(dbType)
         val childMap = mutableMapOf<String, MutableList<TagEntity>>()
 
-        for (t in allTags) {
+        for (t in attachedTags) {
             val attached = t.attachedTo
             if (attached.isNullOrEmpty()) continue
 
--- a/app/src/main/java/org/ole/planet/myplanet/repository/TagsRepositoryImpl.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/TagsRepositoryImpl.kt
@@ -16,7 +16,7 @@ class TagsRepositoryImpl @Inject constructor(
 
     override suspend fun getTagsWithChildren(dbType: String?): Map<TagEntity, List<TagEntity>> {
         val parentTags = getTags(dbType)
-        val attachedTags = tagDao.getAttachedForDb(dbType)
+        val attachedTags = tagDao.getAttached()
         val childMap = mutableMapOf<String, MutableList<TagEntity>>()
 
         for (t in attachedTags) {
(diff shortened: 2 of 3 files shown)


dogi:
**Review — rating: 86/100**

Replaces `tagDao.getAll()` in `getTagsWithChildren` with a targeted `getAttachedForDb(dbType)`. The old version pulled every row in the `tag` table into `TagEntity` objects and then discarded most of them in Kotlin via `if (attached.isNullOrEmpty()) continue`.

I checked the `isAttached = 1` predicate is actually a faithful stand-in for that skip, since it only works if the column is persisted rather than computed on read. It is: `createUnmanagedTag` sets `tag.isAttached = attachedTo.isNotEmpty()` (`TagsRepositoryImpl.kt:151`) and `insert()` upserts those rows, so the flag is written at sync time. The only way `attachedTo` ends up empty is the empty-JSON-array path — and those are exactly the rows the old loop `continue`d on. Tags whose `attachedTo` is absent get `[""]` from the non-array branch, so they stay `isAttached = 1` and are still iterated, matching master. The predicate and the skip agree on every case.

`(:db IS NULL OR db = :db)` also mirrors how `getParentTags` handles a null `db`, so the two halves of the method stay consistent.

**Nit — one genuine behaviour change worth naming.** Adding `db = :db` means a child tag whose `db` differs from the parent's is no longer considered. `getAll()` let a cross-`db` child attach to a parent from another collection. I'd call that a fix rather than a regression — a `resources` tag showing up under a `courses` parent is not something you'd want — but it isn't purely a performance change, and the PR title reads as if it were. Worth a line in the description.

**Verdict:** Mergeable.

---

Okuro3499 (dismissed):
```
TagsRepositoryImpl.getTagsWithChildren / TagDao.getAttachedForDb: Filtering attached-child rows by dbType risks silently excluding legitimate tag-hierarchy children (especially 
    ones with an empty/missing db), reversing what appears to be intentional prior behavior. Fix: either (a) revert to fetching all attached rows unfiltered (isAttached = 1, no db  
    predicate) and rely on the parentTags-keyed associateWith to scope the final result (this still avoids loading isAttached = 0 rows, which is most of the win from dropping       
    getAll()), or (b) if db-scoping on children truly is safe for current server data, get that confirmed against real production tag documents (not just inferred from              
    getByDbAndLinkId's unrelated use case) and add a DAO-level Room test plus a repository test case where a child's db differs from/omits dbType to lock in the intended behavior   
    either way
```

dogi:
**Re-review — Rating: 88/100**

Pushes the child-tag filter from Kotlin into SQL: `tagDao.getAll()` → `tagDao.getAttached()` with `WHERE isAttached = 1`, dropping the rows the loop was going to `continue` past anyway.

**The equivalence question, settled**

The old loop kept a row only when `t.attachedTo` was non-null and non-empty. The new query keeps a row only when `isAttached = 1`. Those agree only if `isAttached` is maintained as exactly `attachedTo.isNotEmpty()` — so I traced every write path.

`TagEntity` rows reach the database through one call: `tagDao.upsertAll(tagsToInsert)` at `TagsRepositoryImpl.kt:126`. Everything in that list is built by `createUnmanagedTag`, which ends with:

```kotlin
tag.attachedTo = attachedTo
tag.isAttached = attachedTo.isNotEmpty()
```

One writer, invariant held by construction. And the codebase already trusts this: `getParentTags` filters on `isAttached = 0`, so the column was load-bearing before this PR.

I also checked the awkward case. When the JSON `attachedTo` isn't an array, `createUnmanagedTag` falls into the `else` branch and adds `JsonUtils.getString("attachedTo", act)` — possibly `""` — giving `attachedTo = [""]`, hence `isAttached = true`. The old loop also kept that row, because `[""]` is a non-empty list. Both paths include it, the subsequent parent lookup on `""` misses either way. Consistent. ✓

The win is real rather than cosmetic: parent tags all carry `isAttached = 0`, so `getAttached()` excludes the entire parent set plus every unattached row, instead of loading the whole table and discarding most of it in Kotlin.

**One nit on the test**

The stub narrowed from six tags to `listOf(child1, child2)`, which drops `unattachedChild` from the fixture. That was the row proving the `attachedTo.isNullOrEmpty()` guard did something. The guard is still in the loop — now as a redundant second line of defence behind the SQL — but nothing exercises it any more. Leaving `unattachedChild` out of the `getAttached()` stub (SQL wouldn't return it) while adding a case that feeds the loop a tag with empty `attachedTo` would keep that coverage, and would catch it if someone later widens the query.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17070, opened 2026-09-10
enterprises: smoother finances reports file utils caching (fixes #17050)

### LINKED ISSUE #17050: Share one attachment-presence cache between the two enterprise adapters
## What

The finances and reports adapters each carry their own copy of the same TTL'd attachment-existence cache.

Files:
- `app/src/main/java/org/ole/planet/myplanet/ui/enterprises/EnterprisesFinancesAdapter.kt` (~33, ~41, ~76-81)
- `app/src/main/java/org/ole/planet/myplanet/ui/enterprises/EnterprisesReportsAdapter.kt` (~27, ~35, ~112-117)

## Why

Both declare `private val attachmentExistsCache = HashMap<String, Pair<Boolean, Long>>()`, both clear it on list submit, and both repeat the same "cached value if fresh, else stat the file and store `(exists, now)`" logic keyed by `imageFile.absolutePath`.

## The fix

- Extract one small reusable holder (e.g. `AttachmentPresenceCache` in the same package) owning the map, the TTL constant, a `clear()` and a single `exists(file, now): Boolean` entry point.
- Have both adapters use it instead of their private copies, keeping the existing clear-on-submit behaviour and the same TTL.
- Keep the file-stat work exactly where it happens today — this is a de-duplication, not a threading change.

## Acceptance

- Both screens still show attachment indicators identically, including after a refresh clears the cache.
- The TTL behaviour is unchanged.
- Neither adapter declares its own cache map any more.

## Notes

dogi:
Extract AttachmentPresenceCache to de-duplicate attachment presence caching logic between EnterprisesFinancesAdapter and EnterprisesReportsAdapter.

Fixes #17050

---
*PR created automatically by Jules for task [8550359493043568357](https://jules.google.com/task/8550359493043568357) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/ui/enterprises/AttachmentPresenceCache.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/enterprises/AttachmentPresenceCache.kt
@@ -1,0 +1,30 @@
+package org.ole.planet.myplanet.ui.enterprises
+
+import java.io.File
+
+class AttachmentPresenceCache(
+    private val ttlMs: Long = DEFAULT_TTL_MS
+) {
+    private val cache = HashMap<String, Pair<Boolean, Long>>()
+
+    fun clear() {
+        cache.clear()
+    }
+
+    fun exists(file: File?, now: Long): Boolean {
+        if (file == null) return false
+        val path = file.absolutePath
+        val cached = cache[path]
+        return if (cached != null && now - cached.second < ttlMs) {
+            cached.first
+        } else {
+            val freshExists = file.exists()
+            cache[path] = Pair(freshExists, now)
+            freshExists
+        }
+    }
+
+    companion object {
+        const val DEFAULT_TTL_MS = 5000L
+    }
+}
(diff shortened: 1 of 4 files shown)


dogi:
**Review — rating: 88/100**

Lifts an identical 13-line TTL-cache block out of `EnterprisesFinancesAdapter` and `EnterprisesReportsAdapter` into `AttachmentPresenceCache`.

I traced the extraction against both originals and it's exact: null file → `false` (matching the old `else false` arm), a hit inside `now - cached.second < ttlMs` returns the cached value, otherwise `file.exists()` is called and the result stored with the timestamp. The 5000 ms TTL survives as `DEFAULT_TTL_MS`, and both adapters keep their `onCurrentListChanged` → `clear()` wiring. Neither call site changed behaviour.

Passing `now` in as a parameter rather than giving the cache its own `TimeProvider` was the right call — it keeps the class a pure data structure, leaves both adapters' existing `timeProvider.now()` injection as the single clock, and it's what makes the new test able to drive TTL expiry by advancing a plain `Long`. That 58-line `AttachmentPresenceCacheTest` is the real gain: this logic guards `File.exists()` calls during RecyclerView binding and was previously duplicated *and* untested in both copies, so a TTL bug had two places to hide and no test to catch it.

**Nit:** the title says "share ... cache between enterprise adapters", but each adapter still constructs its own `AttachmentPresenceCache` — what's shared is the implementation, not the instance. That's the correct design (the two adapters have independent list lifecycles and clear on their own `onCurrentListChanged`), so it's the wording I'd adjust, not the code.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17072, opened 2026-09-10
login: smoother server config shared preferences managing (fixes #17047)

### LINKED ISSUE #17047: Write the server-config preference blocks in one commit instead of eight
## What

Saving a server configuration fires eight consecutive `SharedPreferences` commits, several of which also invalidate the URL caches.

Files:
- `app/src/main/java/org/ole/planet/myplanet/utils/ServerConfigUtils.kt` (`saveAlternativeUrl`, ~91-121)
- `app/src/main/java/org/ole/planet/myplanet/services/SharedPrefManager.kt` (~166-203)
- `app/src/main/java/org/ole/planet/myplanet/ui/sync/ProcessUserDataActivity.kt` (`setUrlParts`, ~120-150)
- `app/src/main/java/org/ole/planet/myplanet/services/UserSessionManager.kt` (`saveUserInfoPref`, ~32-46)

## Why

`saveAlternativeUrl` calls eight setters back to back — `setServerPin`, `setUrlUser`, `setUrlPwd`, `setUrlScheme`, `setUrlHost`, `setAlternativeUrl`, `setProcessedAlternativeUrl`, `setIsAlternativeUrl` — each its own `pref.edit { }` commit, and four of them (`setUrlUser`, `setUrlPwd`, `setProcessedAlternativeUrl`, `setIsAlternativeUrl`) additionally call `UrlUtils.invalidateCaches()`.

`ProcessUserDataActivity.setUrlParts` does the same with seven setters and three invalidations, and `UserSessionManager.saveUserInfoPref` adds two single-key commits before its own batched `edit`.

## The fix

- Add one batching entry point to `SharedPrefManager` — either a single batched server-config setter or two narrow batched setters — that performs **one** `pref.edit { }` for the whole block and calls `UrlUtils.invalidateCaches()` exactly once at the end. Keep the preference keys private to `SharedPrefManager`.
- Rewrite `saveAlternativeUrl`, `setUrlParts` and `saveUserInfoPref` to use it, dropping their per-field setter chains.
- Keep the existing single-field setters for their other callers.

## Acceptance

- The same keys end up with the same values after each of the three call sites runs.
- `UrlUtils.invalidateCaches()` is called once per save, not three-to-four times.
- Login and the alternative-URL flow still work.

## Notes

⚠️ These values are credential-bearing: myPlanet CouchDB URLs embed `satellite:<pin>@host`, and release builds do not strip logs. Do **not** add any log line here, and never interpolate a URL, user, pin or password into a message or exception. Conventions: `docs/CODE_STYLE_GUIDE.md`.

dogi:
Consolidated multiple sequential SharedPreferences commits and cache invalidations in ServerConfigUtils, ProcessUserDataActivity, and UserSessionManager into batched preference setters in SharedPrefManager.

Fixes #17047

---
*PR created automatically by Jules for task [1187281609338770544](https://jules.google.com/task/1187281609338770544) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/services/SharedPrefManager.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/services/SharedPrefManager.kt
@@ -201,6 +201,73 @@ class SharedPrefManager @Inject constructor(
         UrlUtils.invalidateCaches()
     }
 
+    fun saveServerConfig(
+        serverPin: String,
+        urlScheme: String,
+        urlHost: String,
+        serverUrl: String,
+        couchdbUrl: String,
+        urlUser: String,
+        urlPwd: String
+    ) {
+        pref.edit {
+            putString(SERVER_PIN, serverPin)
+            putString(URL_SCHEME, urlScheme)
+            putString(URL_HOST, urlHost)
+            putString(SERVER_URL, serverUrl)
+            putString(COUCHDB_URL, couchdbUrl)
+            putString(URL_USER, urlUser)
+            putString(URL_PWD, urlPwd)
+        }
+        UrlUtils.invalidateCaches()
+    }
+
+    fun saveAlternativeServerConfig(
+        serverPin: String,
+        urlUser: String,
+        urlPwd: String,
+        urlScheme: String,
+        urlHost: String,
+        alternativeUrl: String,
+        processedAlternativeUrl: String,
+        isAlternativeUrl: Boolean = true
+    ) {
+        pref.edit {
+            putString(SERVER_PIN, serverPin)
+            putString(URL_USER, urlUser)
+            putString(URL_PWD, urlPwd)
+            putString(URL_SCHEME, urlScheme)
+            putString(URL_HOST, urlHost)
+            putString(ALTERNATIVE_URL, alternativeUrl)
+            putString(PROCESSED_ALTERNATIVE_URL, processedAlternativeUrl)
+            putBoolean(IS_ALTERNATIVE_URL, isAlternativeUrl)
+        }
+        UrlUtils.invalidateCaches()
+    }
+
+    fun saveUserInfo(
+        userId: String,
+        userName: String,
+        firstName: String?,
+        lastName: String?,
+        middleName: String?,
+        isUserAdmin: Boolean?,
+        lastLogin: Long
+    ) {
+        pref.edit {
+            putString(USER_ID, userId)
+            putString(USER_NAME, userName)
+            remove("password")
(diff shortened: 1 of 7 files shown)


dogi:
**Review — rating: 89/100**

Adds `saveServerConfig`, `saveAlternativeServerConfig` and `saveUserInfo` to `SharedPrefManager` and collapses three multi-setter call sites onto them. Framed as batching, but it also closes a torn-read window, which is the stronger reason to take it.

**The interleaving is the real find.** On master, `invalidateCaches()` is called by only *some* setters — `setCouchdbUrl`, `setUrlUser`, `setUrlPwd`, `setProcessedAlternativeUrl`, `setIsAlternativeUrl` — and not by `setServerPin`, `setUrlScheme`, `setUrlHost` or `setAlternativeUrl`. So `ProcessUserDataActivity`'s seven-setter sequence invalidated three times *in the middle of the writes*: `setUrlUser` cleared the cache while `setUrlScheme` and `setUrlHost` had yet to be written. Any concurrent `UrlUtils` read in that window would repopulate the cache from a half-updated server config. `ServerConfigUtils` had the same shape across its eight setters. Writing everything in one `edit {}` and invalidating once afterwards removes that window entirely, and drops 3–4 invalidations to 1.

The end state is unchanged, which is what makes this safe: in both master sequences the *last* write happened to be one that invalidates, so the final cache state was already correct — it was only the intermediate states that were exposed.

**The one thing that could have gone badly** is `saveUserInfo`, which folds `setUserId`/`setUserName` (writing to `pref`) together with a block that wrote to `sharedPrefManager.rawPreferences`. If those were different preference files, `firstName`/`lastName`/`middleName`/`isUserAdmin`/`lastLogin` would silently relocate and every existing reader would see defaults. They aren't — `val rawPreferences: SharedPreferences get() = pref` (line 24) is the same instance, so the consolidation is a no-op on storage location. `remove("password")` survives, and `if (isUserAdmin != null) putBoolean(...)` matches the old `user?.userAdmin?.let { }`. Correctly, this one does *not* invalidate the URL caches, matching `setUserId`/`setUserName`.

194 lines of new tests across `SharedPrefManagerTest`, `UserSessionManagerTest` and `ServerConfigUtilsTest`.

**Nit:** the individual setters stay public, so nothing stops a future caller from reintroducing the interleaved pattern. A comment on `setUrlScheme`/`setUrlHost` pointing at the batched methods would help.

**Verdict:** Mergeable — worth saying in the description that this fixes partial-state visibility, not just write count.

---

Okuro3499 (approved):

PR #17077, opened 2026-09-10
life: smoother dictionary file reader mapping (fixes #17060)

### LINKED ISSUE #17060: Split file I/O from pure parsing in the dictionary seed path
## What

`insertDictionaryData` mixes asset existence checks, file reading, JSON parsing, entity mapping and the DAO insert in one method that holds an Android `Context`.

`app/src/main/java/org/ole/planet/myplanet/repository/DictionaryRepositoryImpl.kt` (~41-70+)

## Why

```kotlin
if (!FileUtils.checkFileExist(context, Constants.DICTIONARY_URL)) { return@withContext DictionaryLoad.FileMissing }
...
val data = FileUtils.getStringFromFile(FileUtils.getSDPathFromUrl(context, Constants.DICTIONARY_URL))
val json = JsonUtils.gson.fromJson(data, JsonArray::class.java)
val entities = json.map { js -> ... DictionaryEntity(...) }
```

The repository needs `Context` and `FileUtils` purely for asset access, and the mapping — the part worth unit-testing — is buried inside an I/O method.

## The fix

- Extract a `DictionaryAssetDataSource` (holding `@ApplicationContext`) exposing `isDictionaryAssetPresent(): Boolean` and `readDictionaryAssetText(): String?`, and inject it into the repository so `Context` and `FileUtils` leave the repository entirely.
- Extract a pure `JsonArray -> List<DictionaryEntity>` mapper (a private function or small object with no Android dependency), keeping the existing key spellings **exactly** — including the `antonoym` spelling.
- `insertDictionaryData()` then only orchestrates: presence check → `seedMutex.withLock` → count check → read → map → insert, with the same `DictionaryLoad` results (`FileMissing`, `AlreadyPopulated`, success, and the current failure handling).
- Provide the new data source in the appropriate `di/` module.

## Acceptance

- Seeding produces identical rows for the same asset, and all three `DictionaryLoad` outcomes still occur in the same conditions.
- The repository no longer references `Context` or `FileUtils`.
- The mapper is unit-testable without Robolectric; add a test for it.

## Notes

Conventions: `docs/CODE_STYLE_GUIDE.md`, `docs/TESTING.md`.

dogi:
Split file I/O from pure JSON parsing in the dictionary seed path by extracting DictionaryAssetDataSource and DictionaryMapper.

Fixes #17060

---
*PR created automatically by Jules for task [18042583460098643194](https://jules.google.com/task/18042583460098643194) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/di/RepositoryModule.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/di/RepositoryModule.kt
@@ -19,6 +19,8 @@ import org.ole.planet.myplanet.repository.CoursesRepository
 import org.ole.planet.myplanet.repository.CoursesRepositoryImpl
 import org.ole.planet.myplanet.repository.DiagnosticsRepository
 import org.ole.planet.myplanet.repository.DiagnosticsRepositoryImpl
+import org.ole.planet.myplanet.repository.DictionaryAssetDataSource
+import org.ole.planet.myplanet.repository.DictionaryAssetDataSourceImpl
 import org.ole.planet.myplanet.repository.DictionaryRepository
 import org.ole.planet.myplanet.repository.DictionaryRepositoryImpl
 import org.ole.planet.myplanet.repository.DownloadRepository
@@ -80,6 +82,10 @@ abstract class RepositoryModule {
     @Binds
     @Singleton
     abstract fun bindDiagnosticsRepository(impl: DiagnosticsRepositoryImpl): DiagnosticsRepository
+
+    @Binds
+    @Singleton
+    abstract fun bindDictionaryAssetDataSource(impl: DictionaryAssetDataSourceImpl): DictionaryAssetDataSource
 
     @Binds
     @Singleton
--- a/app/src/main/java/org/ole/planet/myplanet/repository/DictionaryAssetDataSource.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/DictionaryAssetDataSource.kt
@@ -1,0 +1,26 @@
+package org.ole.planet.myplanet.repository
+
+import android.content.Context
+import dagger.hilt.android.qualifiers.ApplicationContext
+import javax.inject.Inject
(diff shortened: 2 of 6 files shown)


dogi:
**Review — rating: 88/100**

Splits `DictionaryRepositoryImpl` into three concerns: asset I/O behind `DictionaryAssetDataSource`, JSON→entity conversion in a pure `DictionaryMapper`, and orchestration left in the repository. `@ApplicationContext Context` drops off the repository as a result, so it's constructible in a plain unit test.

The null handling actually got safer rather than just moving. Master passed a possibly-null `data` straight into `gson.fromJson(data, JsonArray::class.java)` and relied on Gson returning null for a null input, which the `if (json != null)` then caught — correct, but by way of a Gson implementation detail. The new `data?.let { JsonUtils.gson.fromJson(it, JsonArray::class.java) }` makes the missing-file path explicit without changing which branch runs.

I checked the mapper is a faithful move, including the thing most likely to be "cleaned up" by mistake: `JsonUtils.getString("antonoym", doc)` keeps its misspelling. That's the field name in the source documents, so correcting it would silently blank out every antonym. Leaving it alone was the right call.

Making the mapper a pure `object` is what earns the 86-line `DictionaryMapperTest` — the conversion previously sat inside a `withContext(io)` block behind a file read and a mutex, so it was effectively untestable, and `DictionaryRepositoryImplTest` shrinks by 26 lines now that it doesn't have to stage the filesystem to exercise mapping.

**Nit — worth coordinating with #17087.** `DictionaryAssetDataSource` and `DictionaryMapper` land in `repository/`, while #17087 in this same batch introduces a new top-level `datasource/` package for `MyLifeCacheDataSource`. Two sibling PRs establishing two different homes for the same concept is the kind of thing that's cheap to settle now and annoying later. `datasource/` reads like the better answer, given `CLAUDE.md` documents `repository/` as interface+impl pairs.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17078, opened 2026-09-10
all: smoother network utils system service caching (fixes #17066)

### LINKED ISSUE #17066: Reuse the cached system-service managers in NetworkUtils.getCurrentNetworkId
## What

`NetworkUtils` already caches its `ConnectivityManager` and `WifiManager`, but `getCurrentNetworkId` re-resolves both from the passed-in `Context`.

`app/src/main/java/org/ole/planet/myplanet/utils/NetworkUtils.kt`

## Why

```kotlin
private val connectivityManagerCache = ResettableCache { context.getSystemService(Context.CONNECTIVITY_SERVICE) as ConnectivityManager }   // ~56-60
private val wifiManagerCache = ResettableCache { context.applicationContext.getSystemService(Context.WIFI_SERVICE) as WifiManager }        // ~62-66
```

but

```kotlin
fun getCurrentNetworkId(context: Context): Int {
    val connManager = context.getSystemService(Context.CONNECTIVITY_SERVICE) as ConnectivityManager      // ~223
    ...
        val wifiManager = context.applicationContext.getSystemService(Context.WIFI_SERVICE) as WifiManager?   // ~230
```

## The fix

- Use the cached `connectivityManager` and `wifiManager` fields in both branches.
- Keep the `@Suppress("DEPRECATION")` on `connectionInfo` and the `Build.VERSION.SDK_INT >= S` branch unchanged.
- Keep the `context` parameter on the signature so callers compile unchanged.

Before you switch: the cached fields are built from the object's own `context`, while the current code uses the parameter. Confirm that every caller passes an application-scoped `Context` (so the resolved services are the same objects); if any caller passes an activity, keep that branch on the parameter and say so in the PR description rather than changing behaviour silently.

## Acceptance

- `getCurrentNetworkId` returns the same values, on both the pre-S and S+ paths.
- No `getSystemService` call remains in the function.

## Notes

Do not add a log line that interpolates a network id, SSID, URL or credential-bearing value — release builds keep logs. Conventions: `docs/CODE_STYLE_GUIDE.md`.

dogi:
Reused the cached `connectivityManager` and `wifiManager` properties in `NetworkUtils.getCurrentNetworkId` instead of re-resolving system services via `context.getSystemService(...)`. Note that callers continue to pass application-scoped context, ensuring identical system service resolution behavior across pre-S and S+ branches.

Fixes #17066

---
*PR created automatically by Jules for task [17590180526593375918](https://jules.google.com/task/17590180526593375918) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/utils/NetworkUtils.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/utils/NetworkUtils.kt
@@ -220,16 +220,14 @@ object NetworkUtils {
 
     fun getCurrentNetworkId(context: Context): Int {
         var networkId = -1
-        val connManager = context.getSystemService(Context.CONNECTIVITY_SERVICE) as ConnectivityManager
-        val network = connManager.activeNetwork
-        val capabilities = connManager.getNetworkCapabilities(network)
+        val network = connectivityManager.activeNetwork
+        val capabilities = connectivityManager.getNetworkCapabilities(network)
         if (capabilities?.hasTransport(NetworkCapabilities.TRANSPORT_WIFI) == true) {
             val connectionInfo: WifiInfo? = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S) {
                 capabilities.transportInfo as? WifiInfo
             } else {
-                val wifiManager = context.applicationContext.getSystemService(Context.WIFI_SERVICE) as WifiManager?
                 @Suppress("DEPRECATION")
-                wifiManager?.connectionInfo
+                wifiManager.connectionInfo
             }
 
             if (connectionInfo != null && !connectionInfo.ssid.isNullOrEmpty()) {


dogi:
**Review — rating: 84/100**

Switches `getCurrentNetworkId` from calling `getSystemService` on every invocation to the `ResettableCache`-backed `connectivityManager` and `wifiManager` the object already holds (`NetworkUtils.kt:60` and `:66`).

This is consistency rather than invention — the same file already caches the connectivity, wifi and bluetooth managers alongside `sharedPrefManager` and `coroutineScope` through the same delegate, so `getCurrentNetworkId` was the outlier still resolving services by hand. Being resettable means the cache doesn't defeat test setup either. Net −5/+3.

**Nit — the parameter is now dead.** With both service lookups gone, `context` is unused throughout `getCurrentNetworkId`'s body (I checked the whole function, not just the changed lines). That's a compiler warning and a signature that misleads about what the call needs. There's exactly one caller — `ConfigurationsRepositoryImpl.kt:137` — so dropping the parameter is a two-line change and would make the PR say what it means.

**Nit:** the cached `wifiManager` is non-null `WifiManager`, so `wifiManager.connectionInfo` replaces master's `wifiManager?.connectionInfo` on a nullable `as WifiManager?`. On a device where `WIFI_SERVICE` resolves to null the cache initializer throws instead of the old silent null. Same exposure the file's other cached managers already carry, so not a new class of problem — just noting the null-safety did change.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17079, opened 2026-09-10
all: smoother version utils lookup caching (fixes #17065)

### LINKED ISSUE #17065: Cache the version code and name lookups in VersionUtils
## What

`getVersionCode` and `getVersionName` each perform a PackageManager IPC on every call, with no caching — in the same object that already caches the Android id.

`app/src/main/java/org/ole/planet/myplanet/utils/VersionUtils.kt`

## Why

```kotlin
@Volatile private var cachedAndroidId: String? = null      // ~14-15

fun getVersionCode(context: Context): Int {
    val pInfo = context.packageManager.getPackageInfo(context.packageName, 0)   // ~17
...
fun getVersionName(context: Context): String? {
    val pInfo = context.packageManager.getPackageInfo(context.packageName, 0)   // ~32
```

The package info cannot change while the process lives, and callers include `MyPlanet.addStats` (~91-92) and the `ConfigurationsRepositoryImpl` version checks.

## The fix

- Add `@Volatile private var cachedVersionCode` / `cachedVersionName` next to `cachedAndroidId`.
- Populate them on the **success path only** — a `PackageManager.NameNotFoundException` must keep returning `0` / `""` (whatever it returns today) and must **not** be cached.
- Keep the existing `Build.VERSION.SDK_INT >= P` branch in `getVersionCode` exactly as it is.
- Add a `@VisibleForTesting internal fun` reset that mirrors the existing `resetAndroidIdCacheForTesting` (~52), and clear the new caches there.

## Acceptance

- Both functions return the same values as today.
- A second call performs no PackageManager lookup.
- The not-found path still returns the current fallback and stays uncached, so a later successful call can populate it.
- `testDefaultDebugUnitTest` passes; add a test using the reset hook.

## Notes

dogi:
Caches version code and version name in VersionUtils to prevent unnecessary PackageManager IPC calls. Successful values are cached while NameNotFoundException calls remain uncached. Adds a reset method for testing.

Fixes #17065

---
*PR created automatically by Jules for task [6407221315336607302](https://jules.google.com/task/6407221315336607302) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/utils/VersionUtils.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/utils/VersionUtils.kt
@@ -14,15 +14,24 @@ object VersionUtils {
     @Volatile
     private var cachedAndroidId: String? = null
 
+    @Volatile
+    private var cachedVersionCode: Int? = null
+
+    @Volatile
+    private var cachedVersionName: String? = null
+
     fun getVersionCode(context: Context): Int {
+        cachedVersionCode?.let { return it }
         try {
             val pInfo = context.packageManager.getPackageInfo(context.packageName, 0)
-            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
-                return getLongVersionCode(pInfo).toInt()
+            val code = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
+                getLongVersionCode(pInfo).toInt()
             } else {
                 @Suppress("DEPRECATION")
-                return pInfo.versionCode
+                pInfo.versionCode
             }
+            cachedVersionCode = code
+            return code
         } catch (e: PackageManager.NameNotFoundException) {
             Log.w(TAG, "Failed to get version code", e)
         }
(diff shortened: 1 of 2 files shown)


dogi:
**Review — rating: 85/100**

Memoizes `getVersionCode` and `getVersionName` in `@Volatile` fields, following the `cachedAndroidId` pattern already in the file.

The premise is sound: `PackageManager.getPackageInfo` is a Binder round-trip to `system_server`, and both values are fixed for the lifetime of an installed app process, so there's nothing to invalidate. These are called from sync headers, the update check and diagnostics, so the calls add up. `@Volatile` matches the existing `cachedAndroidId` treatment and keeps the object's threading story consistent.

Error handling is right, and that's the part that's easy to get wrong: the assignment happens only after a successful lookup, so a `NameNotFoundException` leaves the cache empty and the next call retries rather than pinning a bogus default forever.

**Nit:** `resetAndroidIdCacheForTesting()` now delegates to `reset()` and therefore clears the version caches too, so its name no longer describes what it does. Since both are `@VisibleForTesting` and `reset()` already exists, I'd just migrate the remaining call sites to `reset()` and drop the old name rather than keep a misleading alias.

**Nit:** `getVersionName` caches `pInfo.versionName`, which is nullable. If it ever came back null the cache stays unpopulated and every call re-queries — correct, just silently not memoized. Not worth code, but worth knowing.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17080, opened 2026-09-10
chat: smoother repository history dao sorting (fixes #17064)

### LINKED ISSUE #17064: Align ChatDao.getByUser with the repository-owned ordering
## What

`ChatDao.getByUser` carries a SQL ordering the repository immediately overrides — and the two disagree.

Files:
- `app/src/main/java/org/ole/planet/myplanet/data/room/dao/ChatDao.kt` (~12)
- `app/src/main/java/org/ole/planet/myplanet/repository/ChatRepositoryImpl.kt` (~130-142)
- `app/src/test/java/org/ole/planet/myplanet/repository/ChatRepositoryImplTest.kt`

## Why

```kotlin
@Query("SELECT * FROM chat_history WHERE user = :user ORDER BY id DESC")
suspend fun getByUser(user: String): List<ChatHistory>
```

```kotlin
val chats = chatDao.getByUser(userName)
return sortChats(chats)                                        // ~133
...
private fun sortChats(chats: List<ChatHistory>): List<ChatHistory> =
    chats.sortedByDescending { maxOf(it.createdDate?.toLongOrNull() ?: 0L, it.updatedDate?.toLongOrNull() ?: 0L) }
```

The `ORDER BY id DESC` is dead weight that disagrees with the real ordering and invites a future caller to trust it.

## The fix

- Drop `ORDER BY id DESC` from the query.
- Add a short comment on the DAO method noting that `ChatRepositoryImpl.sortChats` owns ordering, and why SQL cannot: `createdDate` / `updatedDate` are string-typed, so a SQL ordering would not be numerically correct.
- Add a `ChatRepositoryImplTest` assertion that the output order follows `sortChats` even when the DAO returns rows in a deliberately reversed order.

## Acceptance

- The chat list order is unchanged for users (it was always `sortChats`).
- The new test fails if someone re-adds a SQL ordering and drops `sortChats`.

## Notes

Do **not** add a `deleteByIds` method to `ChatDao` while in this file — another open PR claims that name.

Open PR #17058 also touches `ChatRepositoryImpl` (the `searchChats` signature at ~154-158) and `ChatRepositoryImplTest`, but not `ChatDao` and not `getChatHistoryForUser`/`sortChats`; keep your edits to those and they should not collide. Conventions: `docs/CODE_STYLE_GUIDE.md`.

dogi:
Align ChatDao.getByUser with repository-owned ordering.

- Dropped `ORDER BY id DESC` from `@Query` in `ChatDao.getByUser`.
- Added a comment in `ChatDao` noting that `ChatRepositoryImpl.sortChats` owns ordering due to string-typed `createdDate` / `updatedDate`.
- Updated `ChatRepositoryImplTest` to assert that `getChatHistoryForUser` returns items sorted by `sortChats` even when `ChatDao.getByUser` returns rows in reversed order.

Fixes #17064

---
*PR created automatically by Jules for task [697179163892613755](https://jules.google.com/task/697179163892613755) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/data/room/dao/ChatDao.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/data/room/dao/ChatDao.kt
@@ -9,7 +9,9 @@ import org.ole.planet.myplanet.model.ChatHistory
 
 @Dao
 interface ChatDao {
-    @Query("SELECT * FROM chat_history WHERE user = :user ORDER BY id DESC")
+    // Note: ChatRepositoryImpl.sortChats owns ordering for user chats because
+    // createdDate and updatedDate are string-typed, so a SQL ORDER BY would not be numerically correct.
+    @Query("SELECT * FROM chat_history WHERE user = :user")
     suspend fun getByUser(user: String): List<ChatHistory>
 
     @Query("SELECT * FROM chat_history WHERE _id = :docId")
--- a/app/src/test/java/org/ole/planet/myplanet/repository/ChatRepositoryImplTest.kt
+++ b/app/src/test/java/org/ole/planet/myplanet/repository/ChatRepositoryImplTest.kt
@@ -82,14 +82,30 @@ class ChatRepositoryImplTest {
     }
 
     @Test
-    fun getChatHistoryForUser_delegatesToDao() = runTest {
+    fun getChatHistoryForUser_delegatesToDaoAndSortsByRepositoryOrdering() = runTest {
         val userName = "testUser"
-        val mockHistoryList = listOf(ChatHistory().apply { user = userName })
-        coEvery { chatDao.getByUser(userName) } returns mockHistoryList
+        val oldestChat = ChatHistory().apply {
+            user = userName
+            createdDate = "1000"
+            updatedDate = "1000"
+        }
+        val middleChat = ChatHistory().apply {
+            user = userName
+            createdDate = "2000"
+            updatedDate = "1500"
+        }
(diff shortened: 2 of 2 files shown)


dogi:
**Review — rating: 87/100**

Drops `ORDER BY id DESC` from `ChatDao.getByUser` and documents that `ChatRepositoryImpl.sortChats` owns the ordering.

The reasoning holds up. `getChatHistoryForUser` (`ChatRepositoryImpl.kt:130`) is the sole caller — I grepped `app/src/main` and `chatDao.getByUser` appears only at line 134 — and it pipes the result straight through `sortChats` (line 138). So SQLite was sorting a result set whose order was then discarded and recomputed in Kotlin. Removing it costs nothing and saves the sort.

The comment is the valuable part of the diff. `createdDate` and `updatedDate` are string-typed on `ChatHistory`, so `ORDER BY` on them would compare lexicographically — `"9000" > "10000"` — which is exactly the trap someone would fall into when later "fixing" the missing `ORDER BY`. Saying so at the query prevents that.

The test upgrade is what makes this safe rather than merely plausible. `getChatHistoryForUser_delegatesToDao` previously asserted only that the list came back unchanged, so the ordering was entirely unverified and a broken `sortChats` would have passed. The replacement feeds three chats whose `createdDate`/`updatedDate` disagree — including one where `updatedDate` (3000) dominates a low `createdDate` (1000) — and asserts newest-first. That pins the real contract now that SQL isn't providing a fallback order.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17081, opened 2026-09-10
teams: smoother events repository meetup dao member joining (fixes #17061)

### LINKED ISSUE #17061: Join meetup members in Room instead of assembling identities in Kotlin
## What

`getJoinedMembers` expresses a relationship in Kotlin — id fetch, blank filter, distinct, 400-item chunking, a query per chunk, then a distinct-by — instead of in the query.

Files:
- `app/src/main/java/org/ole/planet/myplanet/repository/EventsRepositoryImpl.kt` (~66-79)
- the meetup DAO (`data/room/dao/`)

## Why

```kotlin
val memberIds = meetupDao.getMemberUserIdsByMeetupId(meetupId)
    .mapNotNull { it.takeUnless { id -> id.isBlank() } }
    .distinct()
if (memberIds.isEmpty()) return emptyList()
return memberIds.chunked(400)
    .flatMap { chunk -> userDao.getUsersByAnyIds(chunk) }
    .distinctBy { it.id }
```

## The fix

- Add a DAO query that joins `meetup` to `users` on either identifier (the same "any id" rule `getUsersByAnyIds` implements), returning distinct user rows, with the null/blank membership exclusions expressed in SQL.
- Reduce `getJoinedMembers` to the blank-`meetupId` check plus one delegation.
- Drop the now-unused `UserDao` injection from `EventsRepositoryImpl` if nothing else in the class uses it.

## Acceptance

- The same members are returned for the same data — verify explicitly that duplicate membership rows still yield **one** displayed member.
- A blank `meetupId`, and a meetup with no members, still return an empty list.
- No Kotlin-side chunking or dedupe remains in the method.

## Notes

Match `getUsersByAnyIds`'s matching rule exactly when writing the join — read it first rather than assuming which column it compares. Use `IS NULL` / `!= ''` guards rather than `=` against nullable columns. Conventions: `docs/CODE_STYLE_GUIDE.md`.

dogi:
Refactored `getJoinedMembers` in `EventsRepositoryImpl` to join `meetup` and `users` directly in Room database via `MeetupDao.getJoinedMembersByMeetupId`. Removed Kotlin-side chunking and deduplication as well as the unused `UserDao` injection. Added unit tests in `MeetupDaoTest` and updated `EventsRepositoryImplTest`.

Fixes #17061

---
*PR created automatically by Jules for task [8301623562494912535](https://jules.google.com/task/8301623562494912535) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/data/room/dao/MeetupDao.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/data/room/dao/MeetupDao.kt
@@ -5,6 +5,7 @@ import androidx.room.Insert
 import androidx.room.OnConflictStrategy
 import androidx.room.Query
 import org.ole.planet.myplanet.model.Meetup
+import org.ole.planet.myplanet.model.UserEntity
 
 @Dao
 interface MeetupDao {
@@ -24,6 +25,15 @@ interface MeetupDao {
         "SELECT DISTINCT userId FROM meetup WHERE meetupId = :meetupId AND userId IS NOT NULL AND userId != ''"
     )
     suspend fun getMemberUserIdsByMeetupId(meetupId: String): List<String>
+
+    @Query(
+        """
+        SELECT DISTINCT u.* FROM users u
+        INNER JOIN meetup m ON (u.id = m.userId OR u._id = m.userId)
+        WHERE m.meetupId = :meetupId AND m.userId IS NOT NULL AND m.userId != ''
+        """
+    )
+    suspend fun getJoinedMembersByMeetupId(meetupId: String): List<UserEntity>
 
     @Query("SELECT * FROM meetup WHERE userId = :userId AND userId != ''")
     suspend fun getByUserId(userId: String): List<Meetup>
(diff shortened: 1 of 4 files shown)


dogi:
**Review — rating: 88/100**

Replaces a fetch-ids-then-fetch-users round trip with a single `INNER JOIN`. Master ran `getMemberUserIdsByMeetupId`, filtered blanks, deduped, chunked into 400s and issued a `getUsersByAnyIds` query per chunk, then deduped again in Kotlin — so 1 + ⌈n/400⌉ queries plus two passes of list work. Now it's one statement.

The join reproduces the old semantics rather than approximating them. `(u.id = m.userId OR u._id = m.userId)` is exactly what `getUsersByAnyIds` matched on — important, since this codebase routinely has the CouchDB `_id` and the local `id` carrying the same identity. `DISTINCT u.*` stands in for `distinctBy { it.id }`, and since `id` is the primary key those can't disagree. Dropping the 400-chunking is safe precisely *because* the `IN` list is gone, so there's no bind-variable ceiling left to respect — that's the kind of scaffolding that's easy to leave behind by accident.

Removing `UserDao` from `EventsRepositoryImpl`'s constructor is a real simplification, not just tidying: the repository now has one data source for this operation instead of coordinating two.

`MeetupDaoTest` gains 105 lines exercising the join against a real Room database, which is the right place to verify an `OR`-join with `DISTINCT` — a mock could not have caught a mistake in that SQL.

**Nit:** the SQL guard is `m.userId != ''` where Kotlin used `isBlank()`, so a whitespace-only `userId` is no longer filtered before the join. It can't produce a wrong row (nothing will match `u.id = ' '`), so this is a note rather than a defect.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17086, opened 2026-09-10
sync: smoother upload attachments managing (fixes #17073)

### LINKED ISSUE #17073: Share the succeeded-item attachment path across UploadManager's two result branches
## What

`uploadResource` contains the same attachment-upload block twice, byte for byte.

`app/src/main/java/org/ole/planet/myplanet/services/UploadManager.kt` (~173-205)

## Why

The `UploadResult.Success` branch (over `result.items`) and the `UploadResult.PartialSuccess` branch (over `result.succeeded`) each do exactly this:

```kotlin
listener?.let { l ->
    val libraryIds = result.items.map { it.localId }
    if (libraryIds.isNotEmpty()) {
        val libraries = resourcesRepository.getLibraryItemsByIds(libraryIds)
        val libMap = libraries.associateBy { it.id }
        result.items.forEach { item ->
            libMap[item.localId]?.let { library -> uploadAttachment(item.remoteId, item.remoteRev, library, l) }
        }
    }
}
```

## The fix

- Extract one private helper taking the succeeded items and the listener.
- Return early when the listener is null or the collection is empty, **before** allocating the id list or querying.
- Fetch and index the libraries once per result.
- Call it from both branches.
- Preserve each branch's own `notifyListener(...)` message exactly as it is today — the two strings differ and must not be unified.

## Acceptance

- Both branches still upload the same attachments for the same items.
- The two `notifyListener` messages are unchanged.
- No repository query happens when the listener is null or there are no succeeded items.

## Notes

Open PR #17027 also touches this file, but only the import (~27) and the constructor parameter (~56); your change is at ~173-205, so it should not collide. Conventions: `docs/CODE_STYLE_GUIDE.md`.

dogi:
Extracted a private helper function `uploadAttachments` in `UploadManager.kt` that accepts a list of uploaded items and a listener. If the listener is null or the items list is empty, it returns early before making any queries or allocations. Updated `uploadResource` to call `uploadAttachments` in both `UploadResult.Success` and `UploadResult.PartialSuccess` branches while preserving the distinct `notifyListener` success messages. Added unit test coverage for both branches.

Fixes #17073

---
*PR created automatically by Jules for task [14419713767936529686](https://jules.google.com/task/14419713767936529686) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/services/UploadManager.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/services/UploadManager.kt
@@ -38,6 +38,7 @@ import org.ole.planet.myplanet.services.upload.UploadConstants.BATCH_SIZE
 import org.ole.planet.myplanet.services.upload.UploadCoordinator
 import org.ole.planet.myplanet.services.upload.UploadError
 import org.ole.planet.myplanet.services.upload.UploadResult
+import org.ole.planet.myplanet.services.upload.UploadedItem
 import org.ole.planet.myplanet.utils.DispatcherProvider
 import org.ole.planet.myplanet.utils.FileUtils
 import org.ole.planet.myplanet.utils.JsonUtils.getString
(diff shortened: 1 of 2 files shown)


dogi:
**Review — rating: 87/100**

Collapses two byte-identical attachment-upload blocks in `uploadResource` — one under `UploadResult.Success` reading `result.items`, one under `PartialSuccess` reading `result.succeeded` — into a single `uploadAttachments(items, listener)`.

The guard rewrite is where a dedup like this can quietly change behaviour, so I checked it: master wrapped everything in `listener?.let { l -> if (libraryIds.isNotEmpty()) { … } }`, and the extracted version opens with `if (listener == null || items.isEmpty()) return`. Those are the same two conditions — `libraryIds` was just `items.map { it.localId }`, so an empty `items` and an empty `libraryIds` are the same test. The `getLibraryItemsByIds` → `associateBy { it.id }` → per-item `uploadAttachment(item.remoteId, item.remoteRev, library, listener)` sequence is carried across unchanged, and both branches keep their own distinct `notifyListener` message afterwards.

Worth noting the bulk fetch is preserved rather than accidentally turned into a per-item lookup — `getLibraryItemsByIds(libraryIds)` still runs once for the whole batch, which is the property that actually matters for upload performance.

`UploadManager` is one of the larger files in the project and this is the kind of duplication that drifts — the two copies were already at risk of one being fixed and the other not. 70 lines of new `UploadManagerTest` coverage on top.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17071, opened 2026-09-10
sync: smoother repository elapsed time providing (fixes #17048)

dogi:
Peeled WorkManager, Context, and SystemClock out of SyncRepositoryImpl.

- Added `elapsedRealtime(): Long` to `TimeProvider`, `SystemTimeProvider` (calling `SystemClock.elapsedRealtime()`), and `TestTimeProvider` (returning deterministic `currentTime`).
- Replaced all `SystemClock.elapsedRealtime()` calls in `SyncRepositoryImpl` with `timeProvider.elapsedRealtime()`.
- Created `UserDataUploadScheduler` interface and `UserDataUploadSchedulerImpl` to handle WorkManager enqueueing and WorkInfo state mapping.
- Injected `UserDataUploadScheduler` into `SyncRepositoryImpl` and registered `UserDataUploadSchedulerImpl` in Hilt `ServiceModule`.
- Updated `SyncRepositoryImplTest` to verify flow delegation and passing unit tests without needing Android Context or WorkManager mocks.

Fixes #17048

---
*PR created automatically by Jules for task [14081529859136055422](https://jules.google.com/task/14081529859136055422) started by @dogi*

PR #17087, opened 2026-09-10
life: smoother repository preferences caching (fixes #17074)

### LINKED ISSUE #17074: Extract the MyLife SharedPreferences cache into a data source
## What

`LifeRepositoryImpl` reads and writes a JSON cache directly against raw `SharedPreferences`, which is what drags `androidx.core.content.edit` and Gson into the repository — and it allocates a fresh `TypeToken` on every cache hit.

`app/src/main/java/org/ole/planet/myplanet/repository/LifeRepositoryImpl.kt` (~109-140)

## Why

```kotlin
val json = sharedPrefManager.rawPreferences.getString("$MY_LIFE_CACHE_PREFIX$cacheKey", null)
if (json != null) {
    val cached: List<CachedMyLifeItem>? = try {
        val type = object : TypeToken<List<CachedMyLifeItem>>() {}.type   // fresh TypeToken per hit
        gson.fromJson(json, type)
    ...
```

`data/room/Converters.kt` already hoists its `TypeToken`s to companion vals — same pattern applies here.

## The fix

- Move `CachedMyLifeItem`, `MY_LIFE_CACHE_PREFIX` and the read/write logic into a `MyLifeCacheDataSource` that takes the preferences (and Gson) by injection.
- Hoist the `TypeToken` to a companion val there, mirroring `Converters`.
- Reduce `LifeRepositoryImpl` to calling `read(cacheKey)` / `write(cacheKey, items)`, dropping its Gson and `androidx.core.content.edit` imports if nothing else needs them.
- Provide the data source in the appropriate `di/` module.

Keep the cache key exactly as it is, including the `"--"` fallback for a null user id, and keep the swallow-and-return-null behaviour on a malformed cache entry.

## Acceptance

- Same cached values, same keys, same behaviour on a corrupt entry (treated as a miss).
- `LifeRepositoryImpl` no longer touches `rawPreferences`.
- One `TypeToken` instance, not one per hit.

## Notes

A separate backlog item covers the redundant sort and the duplicated dedupe chain elsewhere in this same file; that is deliberately **not** part of this issue — do not touch those methods, so the two changes stay conflict-free. Conventions: `docs/CODE_STYLE_GUIDE.md`.

dogi:
- Extracted CachedMyLifeItem and cache read/write logic from LifeRepositoryImpl into MyLifeCacheDataSource.
- Hoisted TypeToken<List<CachedMyLifeItem>>() to a companion object property in MyLifeCacheDataSource.
- Provided MyLifeCacheDataSource in SharedPreferencesModule.
- Refactored LifeRepositoryImpl to use MyLifeCacheDataSource and removed direct rawPreferences/Gson/edit usage.
- Updated existing tests and added MyLifeCacheDataSourceTest.

Fixes #17074

---
*PR created automatically by Jules for task [14455662441056787301](https://jules.google.com/task/14455662441056787301) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/datasource/MyLifeCacheDataSource.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/datasource/MyLifeCacheDataSource.kt
@@ -1,0 +1,42 @@
+package org.ole.planet.myplanet.datasource
+
+import android.content.SharedPreferences
+import androidx.core.content.edit
+import com.google.gson.Gson
+import com.google.gson.reflect.TypeToken
+import javax.inject.Inject
+import javax.inject.Singleton
+import org.ole.planet.myplanet.di.AppPreferences
+import org.ole.planet.myplanet.model.MyLife
+
+data class CachedMyLifeItem(
+    var imageId: String?,
+    var title: String?,
+    var isVisible: Boolean,
+    var weight: Int
+)
+
+@Singleton
+class MyLifeCacheDataSource @Inject constructor(
+    @AppPreferences private val preferences: SharedPreferences,
+    private val gson: Gson
+) {
+    fun read(cacheKey: String): List<CachedMyLifeItem>? {
+        val json = preferences.getString("$MY_LIFE_CACHE_PREFIX$cacheKey", null) ?: return null
+        return try {
+            gson.fromJson(json, cachedListType)
+        } catch (e: Exception) {
+            null
+        }
+    }
+
+    fun write(cacheKey: String, items: List<MyLife>) {
+        val cached = items.map { CachedMyLifeItem(it.imageId, it.title, it.isVisible, it.weight) }
+        preferences.edit { putString("$MY_LIFE_CACHE_PREFIX$cacheKey", gson.toJson(cached)) }
+    }
+
+    companion object {
+        private const val MY_LIFE_CACHE_PREFIX = "myLifeCache_"
+        private val cachedListType = object : TypeToken<List<CachedMyLifeItem>>() {}.type
+    }
+}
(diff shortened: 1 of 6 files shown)


dogi:
**Review — rating: 87/100**

Moves the `MyLife` SharedPreferences cache — `CachedMyLifeItem`, the key prefix, the Gson read/write — into a `MyLifeCacheDataSource`, leaving `LifeRepositoryImpl` with the domain logic.

The best part isn't the extraction itself, it's what it closes off. `LifeRepositoryImpl` was reaching through `sharedPrefManager.rawPreferences` — an escape hatch that lets any caller bypass `SharedPrefManager`'s API and poke at raw storage. The data source takes an injected `@AppPreferences SharedPreferences` directly instead, so that reach-through disappears and `LifeRepositoryImpl` sheds its `Gson` dependency too.

The split is drawn in the right place. Serialization, the `myLifeCache_` prefix and the try/catch-to-null on malformed JSON go to the data source; the `mapNotNull { if (item.isVisible) … }` / `sortedBy { it.weight }` reconstruction stays in the repository, because that's domain behaviour rather than caching. Both `write` call sites in `updateVisibility`/the reorder path keep their `effectiveUserId ?: "--"` key, so cache identity is unchanged and existing stored entries stay readable. The 108-line `MyLifeCacheDataSourceTest` covers logic that previously had no direct test.

**Nit:** `provideMyLifeCacheDataSource` in `SharedPreferencesModule` isn't needed. `MyLifeCacheDataSource` is a concrete class with an `@Inject constructor` and `@Singleton`, so Hilt can construct it with no module entry at all — the `@Provides` just restates the constructor and is one more thing to update if a dependency changes. (#17071 has the same pattern with `UserDataUploadScheduler`, though there an interface is involved so `@Binds` is the fix.)

**Nit:** the two new imports land after `javax.inject.Singleton`, breaking the file's alphabetical ordering.

**Nit — worth settling against #17077.** This creates a top-level `datasource/` package while #17077, in the same batch, puts `DictionaryAssetDataSource` in `repository/`. Same concept, two homes. I'd argue yours is the better one — `CLAUDE.md` describes `repository/` as interface+impl pairs — but the two should agree.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17088, opened 2026-09-10
community: smoother services view modelling (fixes #17076)

### LINKED ISSUE #17076: Introduce a CommunityServicesViewModel for team-link data access
## What

`CommunityServicesFragment` calls the teams repository straight from the view layer.

`app/src/main/java/org/ole/planet/myplanet/ui/community/CommunityServicesFragment.kt`

## Why

- `teamsRepository.getTeamLinks()` inside `viewLifecycleOwner.lifecycleScope.launch` at ~60-61
- `teamsRepository.isMember(user?.id, route.teamId)` inside a click handler at ~98-99

## The fix

- Add a `@HiltViewModel CommunityServicesViewModel` injecting `TeamsRepository` (and `UserRepository` if it needs the user id), exposing the link state as a flow plus a membership query.
- Have the fragment collect that state and ask the ViewModel for membership instead of calling the repository.
- Keep routing, the WebView intents and markdown rendering in the fragment — this is a data-access move only.

## Acceptance

- The community services list renders the same links, and the membership-gated click behaves the same (including for a null user id).
- The fragment no longer injects or calls `TeamsRepository`.

## Notes

Needs no DI module change — `@HiltViewModel` plus the existing repository bindings are enough. Conventions: `docs/CODE_STYLE_GUIDE.md`.

dogi:
Introduced CommunityServicesViewModel for team-link data access, refactored CommunityServicesFragment to use it, and added CommunityServicesViewModelTest unit tests.

Fixes #17076

---
*PR created automatically by Jules for task [10064867603795298000](https://jules.google.com/task/10064867603795298000) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/ui/community/CommunityServicesFragment.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/community/CommunityServicesFragment.kt
@@ -61,6 +61,7 @@ class CommunityServicesFragment : BaseTeamFragment() {
         binding?.let { setMarkdownText(it.tvDescription, markdownContentWithLocalPaths) }
 
         collectWhenStarted(viewModel.teamLinks) { links ->
+            if (links == null) return@collectWhenStarted
             val currentBinding = binding ?: return@collectWhenStarted
             if (links.isEmpty()) {
                 currentBinding.llServices.visibility = View.GONE
--- a/app/src/main/java/org/ole/planet/myplanet/ui/community/CommunityServicesFragment.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/community/CommunityServicesFragment.kt
@@ -6,6 +6,7 @@ import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
 import android.widget.TextView
+import androidx.fragment.app.viewModels
 import androidx.lifecycle.lifecycleScope
 import kotlinx.coroutines.launch
 import org.ole.planet.myplanet.R
@@ -19,9 +20,11 @@ import org.ole.planet.myplanet.ui.viewer.WebViewActivity
 import org.ole.planet.myplanet.utils.FileUtils
 import org.ole.planet.myplanet.utils.MarkdownUtils.prependBaseUrlToImages
 import org.ole.planet.myplanet.utils.MarkdownUtils.setMarkdownText
+import org.ole.planet.myplanet.utils.collectWhenStarted
 
 class CommunityServicesFragment : BaseTeamFragment() {
     private var binding: FragmentCommunityServicesBinding? = null
+    private val viewModel: CommunityServicesViewModel by viewModels()
 
     override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View {
         val b = FragmentCommunityServicesBinding.inflate(inflater, container, false)
@@ -57,9 +60,8 @@ class CommunityServicesFragment : BaseTeamFragment() {
         )
         binding?.let { setMarkdownText(it.tvDescription, markdownContentWithLocalPaths) }
 
-        viewLifecycleOwner.lifecycleScope.launch {
-            val links = teamsRepository.getTeamLinks()
-            val currentBinding = binding ?: return@launch
+        collectWhenStarted(viewModel.teamLinks) { links ->
+            val currentBinding = binding ?: return@collectWhenStarted
             if (links.isEmpty()) {
                 currentBinding.llServices.visibility = View.GONE
                 currentBinding.tvNoLinks.visibility = View.VISIBLE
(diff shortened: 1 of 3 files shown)


dogi:
**Review — rating: 85/100**

Introduces `CommunityServicesViewModel` and moves `getTeamLinks()` and `isMember()` off `CommunityServicesFragment`'s direct `teamsRepository` access.

The Hilt wiring is the thing worth verifying here, because `by viewModels()` on a `@HiltViewModel` fails at *runtime* rather than compile time when the fragment isn't a Hilt entry point — so green CI proves nothing about it. `CommunityServicesFragment` carries no annotation of its own, but `BaseTeamFragment` is `@AndroidEntryPoint` (line 17), and the generated `Hilt_BaseTeamFragment` overrides `getDefaultViewModelProviderFactory()`, which the subclass inherits transitively. So the factory resolves. The fragment's pre-existing injected `teamsRepository` on the base is corroborating evidence that this hierarchy is wired.

Beyond the layering, there's a real behavioural gain: `teamLinks` is a `StateFlow` held by the ViewModel, so the links survive a configuration change instead of being re-fetched by a fresh `lifecycleScope.launch` on every view creation. `collectWhenStarted` matches how the rest of the codebase consumes flows, and the `binding ?: return@collectWhenStarted` guard is carried over intact.

**Nit:** the two paths use different styles — `teamLinks` is observed as a flow, while `isMember` stays a suspend call the fragment launches itself. That's defensible (one is screen state, one is a per-click question), but it means the fragment still owns a coroutine, so the "no repository work in the fragment" story is only three-quarters done.

**Nit:** `loadTeamLinks()` in `init` fetches whether or not anything collects. Cheap here, and it avoids a first-frame gap, but `stateIn(..., WhileSubscribed)` would tie the work to actual observation.

**Verdict:** Mergeable.

---

Okuro3499 (dismissed):
```
File: app/src/main/java/org/ole/planet/myplanet/ui/community/CommunityServicesViewModel.kt, constructor param userRepository: UserRepository                                     
    - Problem: Injected but completely unused in the class body — dead dependency.
```

dogi:

The core move is right: replacing a one-shot `viewLifecycleOwner.lifecycleScope.launch` that re-queried `getTeamLinks()` on every view re-creation with a `StateFlow` loaded once in the ViewModel is exactly the fix the title promises, and it survives configuration changes. The test file is a reasonable start.

**What blocks it: a visible empty-state flash**

```kotlin
private val _teamLinks = MutableStateFlow<List<MyTeam>>(emptyList())
```

`StateFlow` always replays its current value to a new collector, so `collectWhenStarted(viewModel.teamLinks)` fires immediately with `emptyList()` — before the repository has returned anything. The fragment's render path treats empty as a terminal state:

```kotlin
if (links.isEmpty()) {
    currentBinding.llServices.visibility = View.GONE
    currentBinding.tvNoLinks.visibility = View.VISIBLE
}
```

So every time this screen opens, the user sees "no links" for as long as the query takes, then the list appears. The old code didn't do this — it rendered only once, after the query returned. That's a regression the refactor introduces, and it's on the most common path rather than an edge case.

The distinction the current type can't express is "empty because nothing loaded yet" vs "empty because there genuinely are none". Two ways out:

```kotlin
private val _teamLinks = MutableStateFlow<List<MyTeam>?>(null)
```
and `return@collectWhenStarted` while null — smallest change. Or a small `sealed interface` (`Loading` / `Loaded(links)`) if you'd rather make the state explicit, which also gives you somewhere to put an error case later.

**One thing that isn't wrong but is worth knowing**

```kotlin
viewModelScope.launch(dispatcherProvider.io) {
    _teamLinks.value = teamsRepository.getTeamLinks()
}
```

The explicit `io` dispatcher is redundant here. `getTeamLinks()` is a `suspend` repository call backed by Room DAOs, and per `CLAUDE.md` Room runs its queries on its own executors, so these are safe to call from any dispatcher. Repositories in this codebase that do need a dispatcher apply it internally (`DownloadRepositoryImpl` wraps in `withContext(dispatcherProvider.io)`). Hopping in the ViewModel as well doesn't break anything, it just means the dispatcher decision now lives in two places. Same applies to the `withContext` inside `isMember`.

**Verdict:** Needs a change — the initial-emission empty state.

---

dogi:
**Re-review — Rating: 87/100** (was 76)

The empty-state flash is fixed.

```kotlin
private val _teamLinks = MutableStateFlow<List<MyTeam>?>(null)
val teamLinks: StateFlow<List<MyTeam>?> = _teamLinks.asStateFlow()
```

with the matching guard in the fragment:

```kotlin
collectWhenStarted(viewModel.teamLinks) { links ->
    if (links == null) return@collectWhenStarted
```

That's the minimal version of what I suggested and it does the job. The StateFlow's mandatory initial replay now carries `null`, which the collector skips, so the render path only ever runs on a value the repository actually produced. "Not loaded yet" and "genuinely none" are now distinguishable, which is the property the previous type couldn't express — `tvNoLinks` shows only in the second case.

The test was updated to match (`teamLinks.first { it != null }` rather than `.first()`), so it now waits for the loaded value instead of asserting against whatever the flow happened to be holding — which also means it would still catch a regression where `loadTeamLinks` never assigns.

**Still worth doing at some point, but not blocking:** `viewModelScope.launch(dispatcherProvider.io)` and the `withContext(dispatcherProvider.io)` inside `isMember` are both redundant — these are `suspend` calls into Room-backed repositories, which per `CLAUDE.md` are safe from any dispatcher, and repositories in this codebase that genuinely need a dispatcher apply it internally. Harmless, just means the dispatcher decision is stated in two places.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17092, opened 2026-09-10
community: smoother leaders name binding (fixes #17083)

### LINKED ISSUE #17083: Fix the message_placeholder misuse in CommunityLeadersAdapter
## What

The leaders list never shows a full name: both branches render the same text, because a whole `UserEntity` is passed as a `%s` argument.

`app/src/main/java/org/ole/planet/myplanet/ui/community/CommunityLeadersAdapter.kt` (~38-42)

## Why

```kotlin
if (leader.firstName == null) {
    holder.title.text = leader.name
} else {
    holder.title.text = context.getString(R.string.message_placeholder, leader)   // ~41
}
```

`values/strings.xml:1006` is `<string name="message_placeholder">%s</string>`, and `UserEntity.toString()` returns `"$name"` (`model/UserEntity.kt:184-186`). So the `else` branch formats the username — exactly what the `if` branch already renders. The intended full-name display never happens.

## The fix

Replace the branch with direct composition, e.g.:

```kotlin
holder.title.text = if (leader.firstName == null) {
    leader.name
} else {
    listOfNotNull(leader.firstName, leader.middleName, leader.lastName)
        .filter { it.isNotBlank() }
        .joinToString(" ")
        .ifBlank { leader.name }
}
```

Check which name fields `UserEntity` actually has before writing this and use those; fall back to `leader.name` if the composed value comes out blank, so no row can render empty. Drop the `message_placeholder` usage here — it is a bare `%s` wrapper and adds nothing.

## Acceptance

- A leader with a `firstName` shows their full name; a leader without one still shows `name`.
- No row renders blank.
- No new string resource is added (so no translation work in the five `values-*` files).

## Notes

A real user-visible bug, not just a cleanup. Conventions: `docs/CODE_STYLE_GUIDE.md`.

dogi:
Fixed the message_placeholder misuse in CommunityLeadersAdapter where leaders' full names were not being rendered. Replaced the logic with direct composition of firstName, middleName, and lastName, falling back to leader.name when blank. Updated unit tests to verify full name composition and fallback behavior.

Fixes #17083

---
*PR created automatically by Jules for task [1933806550040205928](https://jules.google.com/task/1933806550040205928) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/ui/community/CommunityLeadersAdapter.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/community/CommunityLeadersAdapter.kt
@@ -35,11 +35,10 @@ internal class CommunityLeadersAdapter(
 
     override fun onBindViewHolder(holder: CommunityLeadersViewHolder, position: Int) {
         val leader = getItem(position)
-        if (leader.firstName == null) {
-            holder.title.text = leader.name
-        } else {
-            holder.title.text = context.getString(R.string.message_placeholder, leader)
-        }
+        holder.title.text = listOfNotNull(leader.firstName, leader.middleName, leader.lastName)
+            .filter { it.isNotBlank() }
+            .joinToString(" ")
+            .ifBlank { leader.name }
         holder.tvDescription.text = leader.email
 
         holder.itemView.setOnClickListener {
--- a/app/src/test/java/org/ole/planet/myplanet/ui/community/CommunityLeadersAdapterTest.kt
+++ b/app/src/test/java/org/ole/planet/myplanet/ui/community/CommunityLeadersAdapterTest.kt
@@ -1,11 +1,79 @@
 package org.ole.planet.myplanet.ui.community
 
+import android.content.Context
+import androidx.test.core.app.ApplicationProvider
+import org.junit.Assert.assertEquals
 import org.junit.Assert.assertFalse
 import org.junit.Assert.assertTrue
+import org.junit.Before
 import org.junit.Test
+import org.junit.runner.RunWith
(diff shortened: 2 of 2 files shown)


dogi:
**Review — rating: 89/100**

This is a user-visible bug fix, and a nastier one than the title suggests. Master called `context.getString(R.string.message_placeholder, leader)` — passing the `UserEntity` itself as the `%s` argument. Kotlin happily formats it via `toString()`, so any community leader who *had* a `firstName` rendered as something like `org.ole.planet.myplanet.model.UserEntity@3f2a1b` in the list. The `firstName == null` branch worked by accident, which is presumably why it survived: the fallback path looked fine and the broken path was the common one.

The replacement builds the name properly — `listOfNotNull(firstName, middleName, lastName).filter { it.isNotBlank() }.joinToString(" ").ifBlank { leader.name }`. Two things I like about the shape: `middleName` is now included where master ignored it entirely, and `.filter { it.isNotBlank() }` before the join means a present-but-blank field can't produce doubled spaces or a name that's just whitespace, with `ifBlank` catching the all-empty case and falling back to `leader.name` exactly as the old null branch did.

The test earns its keep. It's promoted to Robolectric and walks five distinct shapes through a real view holder: first-name only, first+last, first+middle+last, no first name, and the whitespace case (`firstName = "  "`, `lastName = ""` → `fallback_for_blank`). That last one is the case a naive `joinToString` fix would get wrong, and it's asserted.

**Verdict:** Mergeable — fixes a visible rendering bug and pins the behaviour it should have had.

---

Okuro3499 (approved):

PR #17093, opened 2026-09-10
courses: smoother progress repository dao querying (fixes #17085)

### LINKED ISSUE #17085: Query exact course-progress keys instead of a three-way IN cross product
## What

The course-progress lookup asks for every combination of three id sets, then throws away everything that wasn't one of the requested tuples.

Files:
- `app/src/main/java/org/ole/planet/myplanet/data/room/dao/CourseProgressDao.kt` (~25-26)
- `app/src/main/java/org/ole/planet/myplanet/repository/ProgressRepositoryImpl.kt` (~299-306)

## Why

```kotlin
@Query("SELECT * FROM course_progress WHERE courseId IN (:courseIds) AND userId IN (:userIds) AND stepNum IN (:stepNums)")
suspend fun getByCourseUsersAndSteps(courseIds: List<String>, userIds: List<String>, stepNums: List<Int>): List<CourseProgress>
```

The caller then does `localRecords.groupBy { Triple(it.courseId, it.userId, it.stepNum) }` and looks up only the tuples it actually wanted, so every other combination is pure over-read — and a cross-user row is excluded only by that `Triple` guard, not by the query.

## The fix

- Deduplicate the requested `(courseId, userId, stepNum)` tuples in the repository.
- Query them exactly: prefer one bounded query whose predicate is an OR-of-tuples, or add an indexed composite key column if that reads better — your call, but it must return only requested tuples.
- Keep the existing empty-input short-circuit, and keep the `Triple`-keyed grouping so the downstream lookup is unchanged.
- Mind SQLite's variable limit: chunk the tuple list if it can grow large, the way other repositories here chunk id lists.

Do **not** change `AppDatabase`'s `version` unless you add a column or index; if you do need a schema change, say so in the PR description, because this project rebuilds the database destructively on a version bump and that discards unsynced local writes.

## Acceptance

- Identical results for the same inputs, including the empty-input path.
- No row is returned that wasn't a requested tuple (a cross-user row must not come back at all).
- `testDefaultDebugUnitTest` passes; add a test proving a non-requested combination is not returned.

## Notes

dogi:
Replaced three-way IN cross product course-progress query with exact tuple lookup. Deduplicated tuples in repository and chunked raw queries to respect SQLite variable limits. Added in-memory Room unit tests verifying that non-requested cross-product combinations are not returned.

Fixes #17085

---
*PR created automatically by Jules for task [13777070494925251531](https://jules.google.com/task/13777070494925251531) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/data/room/dao/CourseProgressDao.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/data/room/dao/CourseProgressDao.kt
@@ -2,7 +2,10 @@ package org.ole.planet.myplanet.data.room.dao
 
 import androidx.room.Dao
 import androidx.room.Query
+import androidx.room.RawQuery
 import androidx.room.Upsert
+import androidx.sqlite.db.SimpleSQLiteQuery
+import androidx.sqlite.db.SupportSQLiteQuery
 import org.ole.planet.myplanet.model.CourseProgress
 
 @Dao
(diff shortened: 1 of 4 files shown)


dogi:
**Review — rating: 89/100**

This is filed as a performance change but it's really a correctness fix, and the framing undersells it.

`WHERE courseId IN (…) AND userId IN (…) AND stepNum IN (…)` doesn't match the requested key triples — it matches their **Cartesian product**. Ask for `(courseA, user1, step1)` and `(courseB, user2, step2)` and master also returns `(courseA, user2, step2)` and `(courseB, user1, step1)` if those rows exist. Those are progress records for real users on real courses that were never asked for, handed to `ProgressRepositoryImpl` as `localRecords` during sync reconciliation. The blow-up is multiplicative: 10 courses × 50 users × 20 steps is 10,000 candidate rows to satisfy 50 requested keys.

The OR'd exact-triple form fixes both the over-fetch and the wrong-rows problem in one move, and the mechanics are handled carefully:

- **Bind-variable ceiling.** Chunked at 250, so 750 arguments per statement — safely under SQLite's 999. This is the limit that would otherwise only bite in production on a large sync.
- **Null semantics.** `courseId IS ?` / `userId IS ?` rather than `=`, matching what `CLAUDE.md` asks for in DAO predicates, with `stepNum = ?` for the non-null `Int`.
- **No injection surface.** Only the clause *count* is interpolated; every value goes through `SimpleSQLiteQuery` bind args.

`@RawQuery` gives up Room's compile-time SQL verification, which is a real cost — but there's no way to express per-row tuple matching in Room's annotation DSL, so it's the right trade here, and the 105 lines of new `CourseProgressDaoTest` cover it against a real database.

**Nit:** `List<Triple<String, String, Int>>` as the DAO's public parameter is hard to read at the call site — positional and untyped. A small `data class CourseProgressKey(courseId, userId, stepNum)` would name the three, and `ProgressRepositoryImpl` is already constructing them explicitly. Also `distinct()` is applied in both the repository and the DAO; harmless, but one of them is redundant.

**Verdict:** Mergeable — worth flagging in the description that this changes which rows come back, not just how many.

---

Okuro3499 (approved):

PR #17094, opened 2026-09-10
enterprises: smoother report csv dao exporting (fixes #17084)

### LINKED ISSUE #17084: Use a DAO projection for the enterprise report CSV export
## What

The CSV export loads full `MyTeam` entities when it only reads nine columns.

Files:
- `app/src/main/java/org/ole/planet/myplanet/repository/EnterprisesRepositoryImpl.kt` (`exportReportsAsCsv`, ~102-103)
- `app/src/main/java/org/ole/planet/myplanet/data/room/dao/TeamDao.kt` (~24)

## Why

```kotlin
val reports = teamDao.getNonArchivedReportsByTeamId(teamId)
```

which is

```sql
SELECT * FROM teams WHERE teamId = :teamId AND docType = 'report' AND IFNULL(status, '') != 'archived' ORDER BY createdDate DESC
```

but the CSV only uses `startDate`, `endDate`, `createdDate`, `updatedDate`, `beginningBalance`, `sales`, `otherIncome`, `wages`, `otherExpenses` (profit/loss and ending balance are derived from those).

## The fix

- Add a small projection data class and a `TeamDao` query selecting just those nine columns, with the **same** filter and `ORDER BY createdDate DESC`.
- Switch `exportReportsAsCsv` to it.
- Leave `observeNonArchivedReportsByTeamId` and the list UI on full entities.

## Acceptance

- The exported CSV text is **byte-identical** to today's for the same data — same header, same column order, same number formatting, same row order.
- No other caller changes.

## Notes

A separate backlog item swaps this repository's `@ApplicationContext` for an attachment-file provider at ~130; that is deliberately not part of this issue, so keep your edits to `exportReportsAsCsv` and the DAO. Conventions: `docs/CODE_STYLE_GUIDE.md`.

dogi:
Switched Enterprise report CSV export from loading full MyTeam entities to using a lightweight Room DAO projection selecting only the nine required columns (startDate, endDate, createdDate, updatedDate, beginningBalance, sales, otherIncome, wages, otherExpenses).

Fixes #17084

---
*PR created automatically by Jules for task [13993010551484025708](https://jules.google.com/task/13993010551484025708) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/data/room/dao/TeamDao.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/data/room/dao/TeamDao.kt
@@ -4,6 +4,7 @@ import androidx.room.Dao
 import androidx.room.Query
 import androidx.room.Upsert
 import kotlinx.coroutines.flow.Flow
+import org.ole.planet.myplanet.model.EnterpriseReportCsvProjection
 import org.ole.planet.myplanet.model.MyTeam
 
 @Dao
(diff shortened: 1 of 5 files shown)


dogi:
**Review — rating: 87/100**

Swaps `SELECT *` into full `MyTeam` entities for a nine-column `EnterpriseReportCsvProjection` on the CSV export path. `teams` is a wide table and `exportReportsAsCsv` only ever reads those nine values, so every other column was being read, converted (including anything going through `Converters`) and allocated per report row purely to be discarded.

The risk with a hand-written projection is a type or nullability mismatch that Room happily compiles and then blows up on at runtime, so I checked all nine against the entity. `MyTeam` declares `createdDate: Long = 0` (line 26), `beginningBalance`/`sales`/`otherIncome`/`wages`/`otherExpenses` as `Int = 0` (lines 46–50), and `startDate`/`endDate`/`updatedDate` as `Long = 0` (lines 51–53) — all non-null primitives, matching the projection's non-null `Long`/`Int` exactly. Since Room maps those to NOT NULL columns, there's no row that can arrive as null and break the mapping.

The predicate and ordering are copied verbatim from `getNonArchivedReportsByTeamId` — same `docType = 'report'`, same `IFNULL(status, '') != 'archived'`, same `ORDER BY createdDate DESC` — so the CSV's row set and order are unchanged. Keeping the original method alongside the projection is right, since the `Flow` and suspend variants still have real callers that need whole entities.

`TeamDaoTest` gains 51 lines exercising the new query against a real database, and `EnterprisesRepositoryImplTest` gets simpler because the stub no longer has to build full `MyTeam` objects just to read nine fields off them.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17095, opened 2026-09-10
resources: smoother repository offline metadata loading (fixes #17089)

### LINKED ISSUE #17089: Accumulate offline resource metadata in one filesystem pass
## What

`getOfflineResourceItems` retains every matching `File` in order to walk them a second time for sizes and paths.

`app/src/main/java/org/ole/planet/myplanet/repository/ResourcesRepositoryImpl.kt` (~856-874)

## Why

```kotlin
val grouped = mutableMapOf<String, MutableList<File>>()
oleDir.walkTopDown().filter { it.isFile }.forEach { file ->
    ...
    grouped.getOrPut(resourceId) { mutableListOf() }.add(file)
}

return@withContext grouped.map { (resourceId, files) ->
    val totalSize = files.sumOf { it.length() }
    ...
    OfflineResourceItem(resourceId, title, files.map { it.absolutePath }, totalSize)
}.sortedBy { it.title }
```

Every `File` object is held for a second pass over a whole-tree walk, and `length()` is a stat per file after the fact.

## The fix

- During the walk, accumulate per resource id: a `MutableList<String>` of absolute paths and a running `Long` byte total (call `file.length()` once, while you already have the file).
- After the walk, convert once to `OfflineResourceItem`.
- Preserve exactly: the category matching (`extensions.isEmpty()` → `ext !in allKnownExtensions`, else `ext in extensions`), the `file.parentFile?.name ?: return@forEach` skip, the title fallback to `context.getString(R.string.storage_unknown_resource)` for a blank/missing title, the final `sortedBy { it.title }`, the missing-directory behaviour, and the `dispatcherProvider.io` wrapper.
- **Path order matters**: keep paths in walk order, exactly as `files.map { it.absolutePath }` produced them today.

## Acceptance

- Identical `OfflineResourceItem` lists — same ids, titles, path lists (same order), and byte totals — for the same tree.
- `length()` is called once per file, not once per file per pass.
- A missing `ole` directory behaves as it does now.

## Notes

dogi:
Refactored getOfflineResourceItems in ResourcesRepositoryImpl.kt to process offline resource directory trees in a single filesystem pass. During oleDir.walkTopDown(), matching file paths and length() values are accumulated directly into a local ResourceAccumulator per resource ID, preserving category filtering, path order, title fallback, missing directory behavior, and dispatcher context.

Fixes #17089

---
*PR created automatically by Jules for task [3721659123464850113](https://jules.google.com/task/3721659123464850113) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/repository/ResourcesRepositoryImpl.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/ResourcesRepositoryImpl.kt
@@ -853,7 +853,12 @@ class ResourcesRepositoryImpl @Inject constructor(
 
         val titleMap = getResourceTitlesMap()
 
-        val grouped = mutableMapOf<String, MutableList<File>>()
+        class ResourceAccumulator {
+            val filePaths = mutableListOf<String>()
+            var totalSize = 0L
+        }
+
+        val grouped = mutableMapOf<String, ResourceAccumulator>()
         oleDir.walkTopDown().filter { it.isFile }.forEach { file ->
             val ext = file.extension.lowercase()
             val matchesCategory = if (extensions.isEmpty()) {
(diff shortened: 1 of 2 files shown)


dogi:
**Review — rating: 85/100**

Folds size accumulation into the directory walk with a local `ResourceAccumulator`, replacing the retained `MutableList<File>` and the follow-up `files.sumOf { it.length() }`.

The "one filesystem pass" framing is fair, though it's worth being precise about where the win is. Master already walked once, but `length()` is itself a stat call, so filesystem work happened in two phases: the traversal, then N stats afterwards. Now every stat happens while the walk is on that entry, when the inode metadata is still warm. The secondary win is memory — the map holds path strings and a running `Long` instead of retaining a `File` per file until the mapping step, which matters on a device with a large offline library.

Behaviour is preserved exactly: `filePaths` keeps walk order (append during traversal instead of `files.map` afterwards), `totalSize` is the same sum, the `titleMap[resourceId]?.takeIf { it.isNotBlank() }` fallback is untouched, and the result is still `sortedBy { it.title }`.

The tests are new coverage rather than adjusted coverage, which this function badly needed. `getOfflineResourceItems calculates size and path order in single pass` builds a real temp tree and asserts both the byte total (5 + 10 = 15) and the exact path ordering, and it exercises *both* category branches — the explicit `setOf("mp4")` match and the `extensions.isEmpty()` fallback that only picks up unknown extensions. The not-a-directory and missing-directory guards get a case too.

**Nit:** a local `class` declared inside a function body is an unusual shape in this codebase. A private top-level `data class` would read more conventionally, though the tight scoping here is a legitimate reason to keep it where it is.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17101, opened 2026-09-10
courses: smoother detail markdown context view modelling (fixes #17091)

### LINKED ISSUE #17091: Remove MainApplication.context from CourseDetailViewModel
## What

A ViewModel reaches a global static to build the markdown image base URL.

`app/src/main/java/org/ole/planet/myplanet/ui/courses/CourseDetailViewModel.kt` (import at ~12, use at ~67)

## Why

```kotlin
import org.ole.planet.myplanet.MainApplication
...
val markdownDescription = MarkdownUtils.prependBaseUrlToImages(
    courseDetail.course.description,
    "file://${MainApplication.context.getExternalFilesDir(null)}/ole/",
    600, 350
)
```

## The fix

- Inject `@ApplicationContext context: Context` through the `@HiltViewModel` constructor and use it at ~67.
- Drop the `MainApplication` import.
- Keep the URL string, the `/ole/` suffix and the `600, 350` arguments byte-identical.

## Acceptance

- Course detail markdown still resolves images to the same base URL.
- `CourseDetailViewModel` no longer references `MainApplication`.
- `testDefaultDebugUnitTest` passes (update the ViewModel's test constructor if it has one).

## Notes

The base URL is a local `file://` path, not a server URL — but the standing rule still applies: do not add a log line that interpolates it or any other path/URL/credential value. Conventions: `docs/CODE_STYLE_GUIDE.md`.

dogi:
Inject `@ApplicationContext context: Context` into `CourseDetailViewModel` to replace static `MainApplication.context` access for markdown image base URL construction, and update unit tests accordingly.

Fixes #17091

---
*PR created automatically by Jules for task [15330760396166775794](https://jules.google.com/task/15330760396166775794) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/ui/courses/CourseDetailViewModel.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/courses/CourseDetailViewModel.kt
@@ -1,15 +1,16 @@
 package org.ole.planet.myplanet.ui.courses
 
+import android.content.Context
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
 import dagger.hilt.android.lifecycle.HiltViewModel
+import dagger.hilt.android.qualifiers.ApplicationContext
 import javax.inject.Inject
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.catch
 import kotlinx.coroutines.launch
-import org.ole.planet.myplanet.MainApplication
 import org.ole.planet.myplanet.model.MyCourse
 import org.ole.planet.myplanet.model.MyLibrary
 import org.ole.planet.myplanet.model.StepItem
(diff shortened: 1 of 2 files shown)


dogi:
**Review — rating: 87/100**

Replaces the `MainApplication.context` global in `CourseDetailViewModel` with an injected `@ApplicationContext Context`.

The test diff is the strongest argument for this. Master's `CourseDetailViewModelTest` had to stash `MainApplication.context` in `originalContext`, catch a possible `UninitializedPropertyAccessException`, assign `MainApplication.testContext`, and restore it in an `@After` — mutating global static state around every test, with an explanatory comment about why. All of that collapses to `private val context: Context = mockk(relaxed = true)` passed to the constructor. That kind of scaffolding is order-dependent and leaks across tests when a teardown is missed, so removing it is worth more than the −12 lines suggest.

Worth naming the apparent tension with #17068, which went the other way and *removed* `@ApplicationContext` from `LifeViewModel`. Both are right. There, the context was only being used to resolve strings, so a `(Int) -> String` resolver was strictly cleaner. Here the ViewModel genuinely needs `getExternalFilesDir(null)` to build a `file://` base URL, and `@ApplicationContext` is the application singleton — no activity to leak, and no lifecycle to outlive. Swapping a hard global for a DI-provided app context is the standard fix; it isn't reintroducing the problem #17068 solved.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17102, opened 2026-09-10
Index collection tags once when reconciling selections

### LINKED ISSUE #17090: Index the collection tags once when reconciling selections

`ui/resources/CollectionsFragment.kt` runs a linear scan per selected tag when reconciling selections (~:70-73), and a linear `any` per row in the parent and child loops (~:135, ~:142):

```kotlin
val allTags = list + childMap.values.flatten()
val reconciledList = selectedItemsList.map { selected ->
    allTags.find { selected.matches(it) } ?: selected        // linear scan per selected tag
}
…
val isSelected = selectedItemsList.any { it.matches(parentTag) }        // ~:135
val isChildSelected = selectedItemsList.any { it.matches(childTag) }    // ~:142
```

### The predicate the index must reproduce

`model/TagEntity.kt:51-57`:

```kotlin
    return if (id.isNotEmpty() && other.id.isNotEmpty()) {
        id == other.id
    } else {
        !name.isNullOrEmpty() && name == other.name
    }
}
```

It falls back to **name** comparison whenever **either** side has an empty id. A previous attempt keyed one map on "id if present, else name" and was rejected in review, because the two sides then get keyed on different fields as soon as their id-presence differs:

| selected | candidate | `matches` | single-key lookup |
|---|---|---|---|
| id `""`, name `"Math"` | id `"t1"`, name `"Math"` | **true** (name branch) | `"Math"` vs `"t1"` → **miss** |
| id `"t1"`, name `"Math"` | id `""`, name `"Math"` | **true** (name branch) | `"t1"` vs `"Math"` → **miss** |
| id `"t1"` | id `"t1"` | true | `"t1"` == `"t1"` → hit |

Those mixed rows are exactly what this reconciliation exists for: `reconciledList` replaces previously-selected tags with freshly-loaded objects, which implies a selection can arrive partially populated (a name with no resolved id). Under a single-key index such a selection stops reconciling, falls through to `?: selected`, and — because the same key set feeds `buildTagDataList` — the user's chosen collection silently stops rendering as selected.

### Do

Build the index in one pass over `list` then `childMap.values`, keyed on **both** fields:

- Index each candidate under `tag.id` when `id` is non-empty **and** under `tag.name` when `name` is non-empty.
- Use `putIfAbsent` so the **first** occurrence wins, iterating `list` before `childMap.values` — that reproduces today's `find` order. Do not use `put`, which lets a later duplicate overwrite an earlier one.
- Look up a selected tag by `selected.id` first (when non-empty), then fall back to `selected.name` (when non-empty). That reproduces the truth table above.
- Keep the `?: selected` fallback for genuinely unmatched selections, and keep `selectedItemsList`'s order.
- Derive the selected-key set once before the parent/child loops and use it for both `isSelected` (~:135) and `isChildSelected` (~:142) — with the same both-keys logic, so a selection with an empty id still marks its row.

Two maps (one by id, one by name) is fine if you find it clearer than one; the requirement is that both keys are indexed and the lookup tries id then name.

### Acceptance

- The same rows are marked selected as today, for parents and children, including the multi-select path.
- `selectedItemsList` keeps its order and keeps unmatched entries as their prior objects.
- No behaviour change when the tag list is empty.
- Leave the adapter submissions where they are.

### Tests

Cover the mixed cases, which are the ones a single-key index gets wrong:

- a selection with an **empty id** and a name matching a freshly-loaded tag with a non-empty id reconciles to the loaded tag **and** shows as selected in `buildTagDataList`;
- the reverse — a selection with an id, and a loaded tag with an empty id and the same name;
- the same-id case still hits;
- a selection matching nothing stays its prior object, in place.

### Out of scope

*(Re-filed: PR #17102 was closed. Its index was keyed on one field per tag and so missed the name-fallback cases above.)*

dogi:
Optimizes tag reconciliation and selection checking in CollectionsFragment by indexing parent and child tags into an id/matchKey map and set in a single pass. This eliminates linear scans per tag row while preserving item order and fallback behavior.

Fixes #17090

---
*PR created automatically by Jules for task [9466225621157342713](https://jules.google.com/task/9466225621157342713) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/ui/resources/CollectionsFragment.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/resources/CollectionsFragment.kt
@@ -53,6 +53,10 @@ class CollectionsFragment : DialogFragment(), OnTagClickListener, CompoundButton
         return binding.root
     }
 
+    private fun getMatchKey(tag: TagEntity): String? {
+        return tag.id.ifEmpty { tag.name?.takeIf { it.isNotEmpty() } }
+    }
+
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
 
@@ -67,9 +71,18 @@ class CollectionsFragment : DialogFragment(), OnTagClickListener, CompoundButton
                 is CollectionsState.Success -> {
                     list = state.list
                     childMap = state.childMap
-                    val allTags = list + childMap.values.flatten()
+                    val tagMap = HashMap<String, TagEntity>()
+                    list.forEach { tag ->
+                        getMatchKey(tag)?.let { tagMap.putIfAbsent(it, tag) }
+                    }
+                    childMap.values.forEach { children ->
+                        children.forEach { tag ->
+                            getMatchKey(tag)?.let { tagMap.putIfAbsent(it, tag) }
+                        }
+                    }
                     val reconciledList = selectedItemsList.map { selected ->
-                        allTags.find { selected.matches(it) } ?: selected
+                        val key = getMatchKey(selected)
+                        if (key != null) tagMap[key] ?: selected else selected
                     }
                     selectedItemsList.clear()
                     selectedItemsList.addAll(reconciledList)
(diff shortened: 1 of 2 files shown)


dogi:
**Review — rating: 65/100**

Replacing the nested `allTags.find { selected.matches(it) }` scan with a hash index is the right instinct, but the single-key index isn't equivalent to `matches`, and it diverges in exactly the case this reconciliation exists to handle.

`TagEntity.matches` (`TagEntity.kt:51`) is:

```kotlin
if (id.isNotEmpty() && other.id.isNotEmpty()) id == other.id
else !name.isNullOrEmpty() && name == other.name
```

It falls back to **name** comparison whenever *either* side has an empty id. `getMatchKey` instead picks one key per tag — id if present, otherwise name — so the two sides get keyed on different fields whenever their id-presence differs:

| selected | candidate | `matches` | key lookup |
|---|---|---|---|
| id `""`, name `"Math"` | id `"t1"`, name `"Math"` | **true** (name branch) | selected key `"Math"` vs candidate key `"t1"` → **miss** |
| id `"t1"`, name `"Math"` | id `""`, name `"Math"` | **true** (name branch) | `"t1"` vs `"Math"` → **miss** |
| id `"t1"` | id `"t1"` | true | `"t1"` == `"t1"` → hit ✓ |

The mixed rows are not hypothetical. The whole point of the `reconciledList` block is to replace previously-selected tags with the freshly-loaded objects — which implies selections can arrive partially populated (a name without a resolved id), and that's the one case `matches` has a name fallback for. Under this change those selections silently stop reconciling and fall through to `?: selected`, so the user's chosen collection quietly stops being recognised as selected in `buildTagDataList` too, since the second hunk uses the same key set.

The same-id case is fine, and I checked `putIfAbsent` preserves master's first-wins ordering (`list` then `childMap.values`), so that half is a faithful translation.

The fix keeps the O(1) win: index each candidate under **both** keys — `tag.id` when non-empty *and* `tag.name` when non-empty — then look up `selected.id` first and fall back to `selected.name`. That reproduces the truth table above without reintroducing the scan.

A test for the mixed case would be worth adding alongside; the current 56 lines don't cover an empty-id selection.

**Verdict:** Needs another round — the index needs both keys to match `matches`.

---

PR #17105, opened 2026-09-10
all: smoother settings storage category finding (fixes #17100)

### LINKED ISSUE #17100: Make StorageCategories.indexOf case-insensitive and drop the duplicate lookup

`ui/settings/StorageCategories.kt` builds a map keyed by the **lowercase** extension literals declared at :14-18 and looks them up exactly:

```kotlin
private val extensionToIndex: Map<String, Int> = buildMap {      // :24-30
    all.forEachIndexed { index, category ->
        category.extensions.forEach { ext -> if (!containsKey(ext)) put(ext, index) }
    }
}

fun indexOf(extension: String): Int = extensionToIndex[extension] ?: OTHER_INDEX   // :32
```

Because the lookup is exact, the only caller has to probe twice — `ui/settings/StorageBreakdownFragment.kt:227-232`, inside `scanStorage`'s `oleDir.walkTopDown()`:

```kotlin
val ext = file.extension
val index = if (ext.isEmpty()) {
    StorageCategories.OTHER_INDEX
} else {
    val idx = StorageCategories.indexOf(ext)
    if (idx != StorageCategories.OTHER_INDEX) idx else StorageCategories.indexOf(ext.lowercase())
}
```

So every uppercase extension (`JPG`, `PDF`, `MP4`) and every genuinely-uncategorised file costs two map lookups, once per file across a whole-tree walk. `StorageCategories.indexOf` at :32 is the **only** `indexOf` in this object and `StorageBreakdownFragment:230-231` is its only call site (`:258` is `List.indexOfFirst`, unrelated).

Behaviour is already correct today — the second probe catches uppercase. This is purely removing the redundant probe.

### Do

1. In `StorageCategories.indexOf`, normalise once:

```kotlin
fun indexOf(extension: String): Int = extensionToIndex[extension.lowercase()] ?: OTHER_INDEX
```

2. Collapse `StorageBreakdownFragment:227-232` to a single call, keeping the empty-extension branch exactly as it is:

```kotlin
val ext = file.extension
val index = if (ext.isEmpty()) StorageCategories.OTHER_INDEX else StorageCategories.indexOf(ext)
```

### Constraints

- Category order and `OTHER_INDEX = 4` stay as they are; do not reorder `all` or change any extension set.
- Keep the empty-extension short-circuit — `"".lowercase()` would land on `OTHER_INDEX` anyway, but the branch avoids the lookup entirely and is the documented behaviour.
- Do not change `extensionToIndex`'s construction (including the `if (!containsKey(ext))` first-wins rule) or `allKnownExtensions`. `allKnownExtensions` is consumed by `StorageCategoryDetailFragment:108` and is out of scope for this change.
- Do not touch anything else in `scanStorage` — the `walkTopDown().filter { it.isFile }` traversal, the `sizes`/`counts` arrays, the `total` accumulation, and the missing-directory early return all stay identical.
- Use `lowercase()` with no locale argument, matching the existing call at :231. Do not introduce `lowercase(Locale.ROOT)` here; that would be a separate (and separately arguable) change.

### Tests

`app/src/test/java/org/ole/planet/myplanet/ui/settings/StorageCategoriesTest.kt` exists — keep every case passing and add coverage that `indexOf` is case-insensitive: `JPG`, `Jpg` and `jpg` all resolve to the images index, `PDF` to the pdfs index, and an unknown extension (`xyz`, `XYZ`) still returns `OTHER_INDEX`.

### Out of scope

Do not touch `AppDatabase` or any Room schema. Do not add or change any log line — myPlanet server URLs embed credentials and release builds do not strip logs. Note that two open human PRs also edit `StorageBreakdownFragment.kt` (#17000 around :14-64 and :270, #17002 around :7 and :82-93); neither touches `scanStorage`, so keep your diff confined to :227-232 and do not reformat the rest of the file.

dogi:
Make StorageCategories.indexOf normalize file extensions to lowercase once and drop the redundant second lookup in StorageBreakdownFragment.scanStorage.

Fixes #17100

---
*PR created automatically by Jules for task [308662362518331284](https://jules.google.com/task/308662362518331284) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/ui/settings/StorageBreakdownFragment.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/settings/StorageBreakdownFragment.kt
@@ -224,12 +224,7 @@ class StorageBreakdownFragment : BottomSheetDialogFragment() {
 
         oleDir.walkTopDown().filter { it.isFile }.forEach { file ->
             val ext = file.extension
-            val index = if (ext.isEmpty()) {
-                StorageCategories.OTHER_INDEX
-            } else {
-                val idx = StorageCategories.indexOf(ext)
-                if (idx != StorageCategories.OTHER_INDEX) idx else StorageCategories.indexOf(ext.lowercase())
-            }
+            val index = if (ext.isEmpty()) StorageCategories.OTHER_INDEX else StorageCategories.indexOf(ext)
             val size = file.length()
             total += size
             sizes[index] += size
--- a/app/src/main/java/org/ole/planet/myplanet/ui/settings/StorageCategories.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/settings/StorageCategories.kt
@@ -29,5 +29,5 @@ object StorageCategories {
         }
     }
 
-    fun indexOf(extension: String): Int = extensionToIndex[extension] ?: OTHER_INDEX
+    fun indexOf(extension: String): Int = extensionToIndex[extension.lowercase()] ?: OTHER_INDEX
 }
--- a/app/src/test/java/org/ole/planet/myplanet/ui/settings/StorageCategoriesTest.kt
+++ b/app/src/test/java/org/ole/planet/myplanet/ui/settings/StorageCategoriesTest.kt
@@ -52,11 +52,20 @@ class StorageCategoriesTest {
     }
 
     @Test
+    fun `indexOf is case-insensitive for known extensions`() {
+        assertEquals(3, StorageCategories.indexOf("JPG"))
+        assertEquals(3, StorageCategories.indexOf("Jpg"))
+        assertEquals(3, StorageCategories.indexOf("jpg"))
+        assertEquals(2, StorageCategories.indexOf("PDF"))
+    }
+
+    @Test
     fun `indexOf falls back to the other category for unknown extensions`() {
         assertEquals(StorageCategories.OTHER_INDEX, StorageCategories.indexOf("zip"))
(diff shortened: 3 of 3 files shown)


dogi:
**Review — rating: 87/100**

Moves `.lowercase()` into `StorageCategories.indexOf` and drops the try-then-retry-lowercased dance from `StorageBreakdownFragment`.

I checked the precondition this depends on: every key in `extensionToIndex` is built from the `all` list's extension sets, and all of those are lowercase literals (`"mp4"`, `"mkv"`, `"pdf"`, `"jpg"`, …). So lowercasing the lookup key can only ever help — there's no uppercase key that would stop matching.

Worth being precise about what this does and doesn't fix. `StorageBreakdownFragment:227` is the only caller, and it already worked around the case sensitivity with `if (idx != OTHER_INDEX) idx else StorageCategories.indexOf(ext.lowercase())`. So no user-visible miscategorisation is being corrected today — what changes is that the correctness now lives in the function instead of depending on every caller remembering the two-step. The next caller to appear gets it right by default rather than silently filing `MOV` and `JPG` files under "other".

The test edit is the part I'd highlight in the description: `assertEquals(OTHER_INDEX, StorageCategories.indexOf("MKV"))` was pinning the surprising behaviour as if it were intended. Replacing it with genuinely-unknown `xyz`/`XYZ` and adding a positive case-insensitivity test (`JPG`, `Jpg`, `jpg`, `PDF`) inverts that assertion deliberately rather than by accident, which is exactly what you want to see when a test flips.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17106, opened 2026-09-10
all: smoother diagnostics repository app version utils providing (fixes #17098)

### LINKED ISSUE #17098: Inject an AppVersionProvider into DiagnosticsRepositoryImpl

`repository/DiagnosticsRepositoryImpl.kt` reaches for the generated Android `BuildConfig` from inside the repository layer:

- `:5` — `import org.ole.planet.myplanet.BuildConfig`
- `:62` — `BuildConfig.VERSION_NAME` passed as the `versionName` argument of `buildApkLog(...)` in `saveLogToRoom`
- `:80` — `val versionName = BuildConfig.VERSION_NAME` in `saveLogsToRoom`

The constructor today is:

```kotlin
class DiagnosticsRepositoryImpl @Inject constructor(
    private val apkLogDao: ApkLogDao,
    private val sharedPrefManager: SharedPrefManager
) : DiagnosticsRepository
```

This is the **only** file under `app/src/main/java` that reads `BuildConfig.VERSION_NAME`, so the extraction is self-contained.

### Do

1. Add `utils/AppVersionProvider.kt` — a plain injectable class, no Hilt module needed (Hilt constructs `@Inject constructor` classes directly):

```kotlin
package org.ole.planet.myplanet.utils

import javax.inject.Inject
import org.ole.planet.myplanet.BuildConfig

class AppVersionProvider @Inject constructor() {
    val versionName: String get() = BuildConfig.VERSION_NAME
}
```

2. Add `private val appVersionProvider: AppVersionProvider` to the `DiagnosticsRepositoryImpl` constructor.
3. Replace `BuildConfig.VERSION_NAME` at :62 and :80 with `appVersionProvider.versionName`, and delete the now-dead `BuildConfig` import at :5.

### Constraints

- `BuildConfig.VERSION_NAME` is `String` (non-null), and `buildApkLog` takes `versionName: String?`. Keep the provider property typed `String` — do not widen it to `String?` and do not add a fallback default; the value written into `ApkLog` must be exactly what `BuildConfig.VERSION_NAME` is today.
- **Do not touch `utils/VersionUtils.kt`.** It reads the version from `PackageManager`, is a different source of truth, and is owned by another open task. This change is only about the `BuildConfig` read inside the repository.
- Do not add a Hilt `@Binds`/`@Provides` entry, do not introduce an interface + impl pair, and do not make the provider a `@Singleton` — a property read off a compile-time constant needs neither.
- Do not change `DiagnosticsRepository` (the interface), `buildApkLog`'s signature, `resolveParentCode`, `resolvePlanetCode`, or the `ApkLog` entity.

### Tests

`app/src/test/java/org/ole/planet/myplanet/repository/DiagnosticsRepositoryImplTest.kt` constructs the impl directly, so it needs the new constructor argument. Pass a provider stubbed to a fixed version string (MockK, per the project convention — not Mockito) and add one assertion that the `ApkLog` written by `saveLogToRoom` carries that stubbed version, which is the point of the extraction. Keep every existing test in that file passing.

### Out of scope

dogi:
Injects AppVersionProvider into DiagnosticsRepositoryImpl, replacing direct references to BuildConfig.VERSION_NAME, and updates DiagnosticsRepositoryImplTest to mock the provider and test version propagation.

Fixes #17098

---
*PR created automatically by Jules for task [2934915899477006427](https://jules.google.com/task/2934915899477006427) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/repository/DiagnosticsRepositoryImpl.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/DiagnosticsRepositoryImpl.kt
@@ -2,17 +2,18 @@ package org.ole.planet.myplanet.repository
 
 import java.util.UUID
 import javax.inject.Inject
-import org.ole.planet.myplanet.BuildConfig
 import org.ole.planet.myplanet.data.room.dao.ApkLogDao
 import org.ole.planet.myplanet.model.ApkLog
 import org.ole.planet.myplanet.model.UserEntity
 import org.ole.planet.myplanet.services.SharedPrefManager
+import org.ole.planet.myplanet.utils.AppVersionProvider
 import org.ole.planet.myplanet.utils.CrashLogStore
 
 class DiagnosticsRepositoryImpl @Inject constructor(
     private val apkLogDao: ApkLogDao,
     private val userRepository: UserRepository,
-    private val sharedPrefManager: SharedPrefManager
+    private val sharedPrefManager: SharedPrefManager,
+    private val appVersionProvider: AppVersionProvider
 ) : DiagnosticsRepository {
 
     override suspend fun getPendingApkLogs(): List<ApkLog> {
@@ -59,7 +60,7 @@ class DiagnosticsRepositoryImpl @Inject constructor(
             val log = buildApkLog(
                 resolveParentCode(model),
                 resolvePlanetCode(model),
-                BuildConfig.VERSION_NAME,
+                appVersionProvider.versionName,
                 model?.id,
                 time,
                 type,
@@ -77,7 +78,7 @@ class DiagnosticsRepositoryImpl @Inject constructor(
         if (pendingLogs.isEmpty()) return true
         return try {
             val model = userRepository.getUserModel()
-            val versionName = BuildConfig.VERSION_NAME
+            val versionName = appVersionProvider.versionName
             val parentCode = resolveParentCode(model)
             val planetCode = resolvePlanetCode(model)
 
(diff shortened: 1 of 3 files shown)


dogi:
**Review — rating: 85/100**

Wraps `BuildConfig.VERSION_NAME` in an injectable `AppVersionProvider` and takes it as a constructor dependency in `DiagnosticsRepositoryImpl`.

The justification is testability, and it's a real one rather than DI-for-its-own-sake: `BuildConfig` is a generated compile-time constant, so there is no way to stub it from a unit test. Any assertion about what version string ends up on an uploaded APK log was previously pinned to whatever the test variant happened to be built with. Behind a provider, `DiagnosticsRepositoryImplTest` can assert the version actually written into the log, which is what the +9/−4 there does. Both call sites — `buildApkLog` and the bulk `saveLogsToRoom` path — go through the same dependency, so they can't drift.

MockK handles the final Kotlin class without an interface, so the concrete `AppVersionProvider` is fine as-is; adding an interface here would be ceremony.

**Nit — worth thinking about once, not necessarily here.** The codebase now has two sources for the app's version string: this `BuildConfig.VERSION_NAME` provider, and `VersionUtils.getVersionName(context)`, which asks `PackageManager` (and gains caching in #17079, also in this queue). They return the same value for an installed build, so nothing is broken — but two providers of "the app version" is the kind of thing that quietly diverges. If `AppVersionProvider` is meant to become the single entry point, it'd be worth saying so, and eventually having it delegate rather than sitting beside `VersionUtils`.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17107, opened 2026-09-10
sync: smoother time logger summary generating (fixes #17099)

### LINKED ISSUE #17099: Aggregate the sync timing summary without flattening the log maps

In `utils/SyncTimeLogger.kt`, `generateSummary()` takes two full flattened copies of the per-key log maps and then walks them repeatedly:

- `:229` — `val allApiCallLogs = apiCallTimes.values.flatten()`
- `:230` — `val allDbOpLogs = dbOperationTimes.values.flatten()`
- `:256` — `val totalApiTime = allApiCallLogs.sumOf { it.duration }`
- `:257` — `val successfulCalls = allApiCallLogs.count { it.success }`
- `:277` — `val totalDbTime = allDbOpLogs.sumOf { it.duration }`
- `:278` — `val totalDbItems = allDbOpLogs.sumOf { it.itemCount }`
- `:297` — `allApiCallLogs.sumOf { it.duration }` **again**, for the `Network time` percentage
- `:300` — `allDbOpLogs.sumOf { it.duration }` **again**, for the `Database time` percentage

So a sync's whole API and DB log history is copied into two new lists and each total is summed twice.

### Do

1. Delete both `flatten()` snapshots at :229-230.
2. Fold the per-endpoint / per-model lists directly where the totals are needed:
   - API section: `totalApiTime` and `successfulCalls` from `apiCallTimes.values` (e.g. `sumOf { logs -> logs.sumOf { it.duration } }` and `sumOf { logs -> logs.count { it.success } }`). `totalApiCalls` at :255 already folds `apiCallTimes.values` — leave it as is.
   - DB section: `totalDbTime` and `totalDbItems` from `dbOperationTimes.values` the same way. `totalDbOps` at :276 already folds — leave it.
3. Reuse the values computed in step 2 for the `PERFORMANCE INSIGHTS` section instead of re-summing at :297 and :300. Hoist `totalApiTime` / `totalDbTime` above the two `if` blocks if that is what it takes — but keep `apiPercentage`/`dbPercentage` `0.0` when `apiCallTimes` / `dbOperationTimes` is empty, exactly as the current `if (…isNotEmpty())` guards do.

### Constraints — output must be identical

The generated summary is persisted and uploaded, so every character has to stay the same:

- Same section order and headers (`=== SYNC TIME SUMMARY ===`, `PROCESS BREAKDOWN:`, `API CALL STATISTICS:`, `DB OPERATION STATISTICS:`, `PERFORMANCE INSIGHTS:`, trailing `=========================`).
- Same `String.format(Locale.US, …)` templates, same widths, same `%.1f` / `%3d%%` rounding, same `take(25)` truncation.
- Same rankings: the API list stays sorted by `it.value.sumOf { log -> log.duration }` descending, the DB list likewise, the process list by value descending with the `:start` filter.
- Same integer division for `avgTime` (`totalTime / logs.size`) and the same empty-list guards.
- Same `Other processing: 100 - apiPercentage - dbPercentage` arithmetic.
- Do not touch `processTimes` / `processItemCounts` handling, `formatTime`, `formatElapsed`, `formatTimestamp`, `shortenEndpoint`, or `extractProcessName`.

### Do not touch the logging gate

Another open PR (#17056) rewrites the `Log.isLoggable("SyncPerf", …)` call sites in this same file into an `isVerbose` snapshot (`startLogging`, `stopLogging`, `endProcess`, `logApiCall`, `logDbOperation`, `logDetail`, and the `companion object` `TAG`). Confine this change to the body of `generateSummary()` so the two do not collide: do not add, remove or rename any `Log.d` / `Log.isLoggable` call, do not add a `TAG` constant, and do not edit `services/sync/SyncManager.kt`.

### Tests

`app/src/test/java/org/ole/planet/myplanet/utils/SyncTimeLoggerTest.kt` already asserts summary substrings (`Total API calls: 2 (Success: 1, Failed: 1)`, and more). Keep all of them passing. Add coverage that the aggregates still hold across **multiple endpoints and multiple models** (more than one key in each map, several logs per key, at least one failed API call), asserting the total lines and the `Network time` / `Database time` percentages. **Append new tests at the end of the class** — #17056 also edits this file near the top (~:48) and around :104-155, so keep your additions clear of those regions.

### Out of scope

dogi:
Refactored generateSummary() in SyncTimeLogger.kt to fold per-key log lists directly and reuse total API/DB times in performance insights without flattening or re-summing log maps. Added unit tests in SyncTimeLoggerTest.kt verifying summary aggregation across multiple endpoints and models.

Fixes #17099

---
*PR created automatically by Jules for task [15564120779964989514](https://jules.google.com/task/15564120779964989514) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/utils/SyncTimeLogger.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/utils/SyncTimeLogger.kt
@@ -226,8 +226,8 @@ class SyncTimeLogger @Inject constructor(
         summaryBuilder.append("=== SYNC TIME SUMMARY ===\n")
         summaryBuilder.append("Total sync time: $totalMinutes min $totalSeconds sec (${formatTime(totalDuration)})\n\n")
 
-        val allApiCallLogs = apiCallTimes.values.flatten()
-        val allDbOpLogs = dbOperationTimes.values.flatten()
+        val totalApiTime = apiCallTimes.values.sumOf { logs -> logs.sumOf { it.duration } }
+        val totalDbTime = dbOperationTimes.values.sumOf { logs -> logs.sumOf { it.duration } }
 
         // Process times
         summaryBuilder.append("PROCESS BREAKDOWN:\n")
@@ -253,8 +253,7 @@ class SyncTimeLogger @Inject constructor(
         if (apiCallTimes.isNotEmpty()) {
             summaryBuilder.append("\nAPI CALL STATISTICS:\n")
             val totalApiCalls = apiCallTimes.values.sumOf { it.size }
-            val totalApiTime = allApiCallLogs.sumOf { it.duration }
-            val successfulCalls = allApiCallLogs.count { it.success }
+            val successfulCalls = apiCallTimes.values.sumOf { logs -> logs.count { it.success } }
 
             summaryBuilder.append(String.format(Locale.US, "  Total API calls: %d (Success: %d, Failed: %d)\n",
                 totalApiCalls, successfulCalls, totalApiCalls - successfulCalls))
@@ -274,8 +273,7 @@ class SyncTimeLogger @Inject constructor(
         if (dbOperationTimes.isNotEmpty()) {
             summaryBuilder.append("\nDB OPERATION STATISTICS:\n")
             val totalDbOps = dbOperationTimes.values.sumOf { it.size }
-            val totalDbTime = allDbOpLogs.sumOf { it.duration }
-            val totalDbItems = allDbOpLogs.sumOf { it.itemCount }
+            val totalDbItems = dbOperationTimes.values.sumOf { logs -> logs.sumOf { it.itemCount } }
 
             summaryBuilder.append(String.format(Locale.US, "  Total Db operations: %d\n", totalDbOps))
             summaryBuilder.append(String.format(Locale.US, "  Total Db time: %s (%.1f%% of total sync)\n",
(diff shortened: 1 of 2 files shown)


dogi:
**Review — rating: 86/100**

Drops the two `flatten()` calls in `generateSummary` and aggregates directly over the per-key lists.

The arithmetic is unchanged, which is the main thing to confirm on a change like this. `apiCallTimes.values.flatten().sumOf { it.duration }` and `apiCallTimes.values.sumOf { logs -> logs.sumOf { it.duration } }` are the same total; `count { it.success }` over the flattened list equals `sumOf { logs.count { it.success } }`; same for `totalDbItems`. The `if (apiCallTimes.isNotEmpty())` / `dbOperationTimes.isNotEmpty()` guards around the percentage block are untouched, and the totals compute to 0 on empty maps just as the flattened versions did, so hoisting them above the guards changes nothing.

There's a second, quieter win: master computed `allApiCallLogs.sumOf { it.duration }` in the API statistics block and then **again** in the performance-insights block. Now `totalApiTime` is computed once and reused in both, so that's a duplicate full pass removed on top of the two materialized lists.

Worth the effort even though `generateSummary` runs once per sync — `apiCallTimes` and `dbOperationTimes` accumulate an entry per API call and per DB batch across a whole sync, so `flatten()` was allocating two large lists at exactly the point where the sync is wrapping up.

**Note for whoever merges:** this edits `SyncTimeLogger.kt`, as does #17056 (already `ready`) — that one touches `startLogging`/`logApiCall`/`logDbOperation` while this one is confined to `generateSummary`, so they shouldn't conflict, but they'll want landing in some order rather than both assuming a clean base.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17108, opened 2026-09-10
courses: smoother submissions repository date exporting (fixes #17097)

### LINKED ISSUE #17097: Reuse one thread-safe java.time formatter in SubmissionsRepositoryExporter

`repository/SubmissionsRepositoryExporter.kt:34`, inside the `companion object`:

```kotlin
private val dateFormatter = ThreadLocal.withInitial { SimpleDateFormat("yyyy-MM-dd HH:mm", Locale.getDefault()) }
```

It has exactly one use, at :175:

```kotlin
canvas.drawText("Generated: ${dateFormatter.get()?.format(Date())}", MARGIN, yPosition, normalPaint)
```

A `ThreadLocal` exists only to work around `SimpleDateFormat` not being thread-safe. `java.time.format.DateTimeFormatter` is immutable and thread-safe, so the whole `ThreadLocal` (and the nullable `.get()?`) is unnecessary indirection for one PDF header stamp.

Import inventory in this file: `java.text.SimpleDateFormat` (:10) is used only at :34; `java.util.Date` is used only at :175; `java.util.Locale` (:12) is used only at :34. So after the change the first two become dead.

### Do

1. Replace the companion field at :34 with a single immutable formatter:

```kotlin
private val dateFormatter: DateTimeFormatter =
    DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm", Locale.getDefault())
        .withZone(ZoneId.systemDefault())
```

2. At :175, format the current instant with it:

```kotlin
canvas.drawText("Generated: ${dateFormatter.format(Instant.now())}", MARGIN, yPosition, normalPaint)
```

3. Add `java.time.Instant`, `java.time.ZoneId`, `java.time.format.DateTimeFormatter`; drop `java.text.SimpleDateFormat` and `java.util.Date`. Keep `java.util.Locale` (still used by `ofPattern`).

### Constraints

- The rendered header text must stay identical in shape: `Generated: yyyy-MM-dd HH:mm` — same pattern, same default locale, same system default time zone.
- **Keep the system clock.** This class injects `timeProvider: TimeProvider` (constructor, :31), but the stamp at :175 currently reads `Date()`. Do **not** switch this call to `timeProvider` — that is a separate behaviour/testability change and would widen the diff. `Instant.now()` is the faithful replacement for `Date()`.
- The formatter stays `private` in the `companion object`. Do not make it public, do not move it to a shared utils object, and do not touch `TimeUtils`.
- Do not change `PAGE_WIDTH`, `PAGE_HEIGHT`, `MARGIN`, `LINE_HEIGHT` or any other part of the PDF layout.

### Tests

There is no existing test that asserts the "Generated:" line. Do not add a test that pins the current wall-clock time. If you add coverage, keep it to the formatter itself — e.g. assert that formatting a fixed `Instant` with the same pattern in a fixed zone yields the expected string — rather than invoking the PDF export.

### Out of scope

dogi:
Replaced ThreadLocal<SimpleDateFormat> with thread-safe java.time.format.DateTimeFormatter and Instant.now() in SubmissionsRepositoryExporter. Cleaned up unused java.text.SimpleDateFormat and java.util.Date imports. Added unit test for DateTimeFormatter.

Fixes #17097

---
*PR created automatically by Jules for task [9061686378388374206](https://jules.google.com/task/9061686378388374206) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/repository/SubmissionsRepositoryExporter.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/SubmissionsRepositoryExporter.kt
@@ -7,8 +7,9 @@ import android.graphics.pdf.PdfDocument
 import android.os.Environment
 import java.io.File
 import java.io.FileOutputStream
-import java.text.SimpleDateFormat
-import java.util.Date
+import java.time.Instant
+import java.time.ZoneId
+import java.time.format.DateTimeFormatter
 import java.util.Locale
 import javax.inject.Inject
 import org.json.JSONObject
@@ -31,7 +32,9 @@ internal class SubmissionsRepositoryExporter @Inject constructor(
 ) {
 
     companion object {
-        private val dateFormatter = ThreadLocal.withInitial { SimpleDateFormat("yyyy-MM-dd HH:mm", Locale.getDefault()) }
+        private val dateFormatter: DateTimeFormatter =
+            DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm", Locale.getDefault())
+                .withZone(ZoneId.systemDefault())
 
         private const val PAGE_WIDTH = 595
         private const val PAGE_HEIGHT = 842
@@ -172,7 +175,7 @@ internal class SubmissionsRepositoryExporter @Inject constructor(
 
                 canvas.drawText("Total Submissions: ${submissions.size}", MARGIN, yPosition, normalPaint)
                 yPosition += LINE_HEIGHT
-                canvas.drawText("Generated: ${dateFormatter.get()?.format(Date())}", MARGIN, yPosition, normalPaint)
+                canvas.drawText("Generated: ${dateFormatter.format(Instant.now())}", MARGIN, yPosition, normalPaint)
                 yPosition += LINE_HEIGHT * 3
 
                 val examId = getExamId(submissions.firstOrNull()?.parentId)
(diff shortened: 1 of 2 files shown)


dogi:
**Review — rating: 88/100**

Swaps `ThreadLocal.withInitial { SimpleDateFormat(...) }` for a plain static `DateTimeFormatter`. This is the correct fix rather than a stylistic one: `SimpleDateFormat` is mutable and not thread-safe, which is *why* the `ThreadLocal` was there — `DateTimeFormatter` is immutable and explicitly thread-safe, so the wrapper stops being necessary rather than being papered over.

The output is preserved. `SimpleDateFormat("yyyy-MM-dd HH:mm", Locale.getDefault())` formatting `Date()` renders the current instant in the default zone; `DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm", Locale.getDefault()).withZone(ZoneId.systemDefault())` formatting `Instant.now()` does the same. The pattern letters carry identical meanings across both APIs, and the explicit `.withZone(...)` is required — without it, formatting an `Instant` would throw for the zone-dependent fields, so it's good that it's there rather than relying on a default.

Dropping the `ThreadLocal` also removes a per-thread retained `SimpleDateFormat`, which is the kind of thing that lingers on pooled threads. `java.time` needs no desugaring here given minSdk 26.

`Locale.getDefault()` is still captured once at class initialisation, so a locale change mid-process isn't picked up — but that was equally true of the `ThreadLocal` initializer, so it's unchanged behaviour rather than something this PR introduces.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17114, opened 2026-09-10
courses: smoother progress repository steps counting (fixes #17103)

### LINKED ISSUE #17103: Count passed course steps without the intermediate collections

`repository/ProgressRepositoryImpl.kt:194-198`, inside `getCompletedCourses` (declared at :179), builds two throwaway collections per course purely to read a size:

```kotlin
// Count UNIQUE steps that are passed (matches web: step.passed === true)
val passedStepNumbers = courseProgressRecords
    .filter { it.passed }
    .map { it.stepNum }
    .toSet()
val passedSteps = passedStepNumbers.size
```

`courseProgressRecords` comes from `progressByCourse[course.courseId].orEmpty()` (:191), and `passedStepNumbers` is used **only** for `.size` — nothing downstream reads the set. So each course allocates a filtered list, a mapped list and a set, inside a `myCourses.forEach` (:186) that runs while building completion badges.

### Do

Replace :194-198 with a single traversal that accumulates the unique passed step numbers directly, e.g.

```kotlin
// Count UNIQUE steps that are passed (matches web: step.passed === true)
val passedSteps = courseProgressRecords
    .mapNotNullTo(HashSet()) { if (it.passed) it.stepNum else null }
    .size
```

or an explicit loop over `courseProgressRecords` adding `it.stepNum` to a `HashSet<Int>` when `it.passed`. Either is fine — the requirement is one pass and one collection instead of three.

### Constraints — semantics must not move

- **Uniqueness stays.** A step number appearing in several passed records counts once. Do not switch to `count { it.passed }`, which would over-count duplicates.
- `passed == false` records are ignored, exactly as the `filter` does today.
- Keep the very next lines untouched: `totalSteps = course.courseSteps?.size ?: 0` (:199) and `allStepsPassed = passedSteps == totalSteps && totalSteps > 0` (:202).
- Keep the `hasValidId` (:187) and `hasValidTitle` (:188) gates and the `if (allStepsPassed && hasValidId && hasValidTitle)` condition at :205 exactly as they are.
- Keep the order in which `completedCourses` is appended (:185, the `myCourses.forEach` iteration order) — the returned list's ordering must be identical.
- Do not change `progressByCourse`, `getProgressRecords`, or anything else in the file; this is a five-line change inside `getCompletedCourses`.
- Keep the existing comment, since it records the parity intent with the web app.

### Tests

Extend the existing `app/src/test/java/org/ole/planet/myplanet/repository/ProgressRepositoryImplTest.kt` and keep every case in it passing. Cover, for `getCompletedCourses`:

- duplicate passed records for the same `stepNum` count once (a course with steps 1,1,2 and `totalSteps = 2` still completes),
- a `passed == false` record does not count (a course with a failed step and `totalSteps` equal to the passed count stays incomplete),
- a course with `totalSteps == 0` never appears in the result.

### Out of scope

Another open PR (#17093) also edits this file, but only `insertCourseProgressFromSync` around :286-307 (the exact-tuple `CourseProgressDao` lookup). Keep your diff inside `getCompletedCourses` so the two do not collide, and do not touch `CourseProgressDao`.

dogi:
Replaced intermediate list and set allocations in `ProgressRepositoryImpl.getCompletedCourses` with a single-pass `mapNotNullTo(HashSet())` traversal, reducing intermediate allocations per course while maintaining unique passed step counting logic. Extended test coverage in `ProgressRepositoryImplTest`.

Fixes #17103

---
*PR created automatically by Jules for task [2166263424886531981](https://jules.google.com/task/2166263424886531981) started by @dogi*

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved accuracy when calculating completed courses, ensuring duplicate progress records are counted only once.
  - Courses with failed steps or no required steps are excluded from completion results.

- **Tests**
  - Added coverage for duplicate passed steps, failed steps, and courses with zero total steps.

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/repository/ProgressRepositoryImpl.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/ProgressRepositoryImpl.kt
@@ -191,11 +191,9 @@ class ProgressRepositoryImpl @Inject constructor(
             val courseProgressRecords = progressByCourse[course.courseId].orEmpty()
 
             // Count UNIQUE steps that are passed (matches web: step.passed === true)
-            val passedStepNumbers = courseProgressRecords
-                .filter { it.passed }
-                .map { it.stepNum }
-                .toSet()
-            val passedSteps = passedStepNumbers.size
+            val passedSteps = courseProgressRecords
+                .mapNotNullTo(HashSet()) { if (it.passed) it.stepNum else null }
+                .size
             val totalSteps = course.courseSteps?.size ?: 0
 
             // Web logic: ALL steps must be passed AND course must have at least one step
--- a/app/src/test/java/org/ole/planet/myplanet/repository/ProgressRepositoryImplTest.kt
+++ b/app/src/test/java/org/ole/planet/myplanet/repository/ProgressRepositoryImplTest.kt
@@ -446,6 +446,85 @@ class ProgressRepositoryImplTest {
     }
 
     @Test
+    fun testGetCompletedCourses_duplicatePassedRecordsCountOnce() = testScope.runTest {
+        val myCourses = listOf(
+            MyCourse().apply {
+                courseId = "course1"
+                courseTitle = "Course 1"
+                courseSteps = mutableListOf(
+                    CourseStep().apply { courseId = "course1" },
+                    CourseStep().apply { courseId = "course1" }
+                )
+            }
+        )
+
+        val progresses = listOf(
+            CourseProgress().apply { courseId = "course1"; stepNum = 1; passed = true },
+            CourseProgress().apply { courseId = "course1"; stepNum = 1; passed = true },
+            CourseProgress().apply { courseId = "course1"; stepNum = 2; passed = true }
+        )
+
+        coEvery { mockCoursesRepository.getMyCourses("user1") } returns myCourses
+        coEvery { courseProgressDao.getByUser("user1") } returns progresses
+
+        val result = repository.getCompletedCourses("user1")
+        advanceUntilIdle()
+
+        assertEquals(1, result.size)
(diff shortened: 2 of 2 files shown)


dogi:
**Review — rating: 84/100**

Collapses `filter { it.passed }.map { it.stepNum }.toSet().size` into a single `mapNotNullTo(HashSet())`, dropping two intermediate lists per course in a loop that runs over every course.

The correctness question worth checking is `stepNum`'s nullability, because `mapNotNull` and `map` diverge precisely there: if `stepNum` were `Int?`, master's `.map { }.toSet()` would keep `null` as a set element and count it, while `mapNotNullTo` drops it — silently changing the passed-step count for any record with a null step. It isn't nullable. `CourseProgressDao.getByCourseUsersAndSteps` takes `stepNums: List<Int>` on master, and #17093 in this same queue types the key tuple as `Triple<String, String, Int>`. With a non-null `Int`, the `if (it.passed) it.stepNum else null` lambda produces exactly the old filtered set, so the count is unchanged.

Keeping the `// Count UNIQUE steps that are passed (matches web: step.passed === true)` comment matters here — the uniqueness is load-bearing for the all-steps-passed check right below, and the new form makes it slightly less obvious that a `Set` is doing that work.

79 lines of new `ProgressRepositoryImplTest` coverage, which this calculation didn't have.

**Nit:** the `mapNotNullTo(HashSet()) { if (...) x else null }` idiom reads as "map, discarding nulls" when the intent is "filter, then dedupe". `courseProgressRecords.asSequence().filter { it.passed }.map { it.stepNum }.toSet().size` gets the same single-pass, no-intermediate-list behaviour while keeping the original shape legible.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17115, opened 2026-09-10
resources: smoother repository batch inserting (fixes #17104)

### LINKED ISSUE #17104: Make batchInsertResources a single pass over the documents

`repository/ResourcesRepositoryImpl.batchInsertResources` (:674-716) walks `documents` three times and extracts `_id` from each document three times:

```kotlin
val validDocs = documents.filter {                       // :677-680  — extraction #1
    val _id = JsonUtils.getString("_id", it)
    _id.isNotBlank() && !_id.startsWith("_design")
}
val resourceIds = validDocs.map { JsonUtils.getString("_id", it) }   // :681 — extraction #2
…
validDocs.forEach { doc ->                               // :690-709
    val _id = JsonUtils.getString("_id", doc)            // :692 — extraction #3
```

This is on the sync path (`batchInsertResources` is what the resource pull calls per page), so it is a per-document JSON lookup done three times over.

### Do

Build the valid documents **and** their ids in one pass, and reuse the id in the entity loop instead of re-extracting it:

```kotlin
val validDocs = ArrayList<Pair<JsonObject, String>>(documents.size)
val resourceIds = ArrayList<String>(documents.size)
for (doc in documents) {
    val id = JsonUtils.getString("_id", doc)
    if (id.isNotBlank() && !id.startsWith("_design")) {
        validDocs.add(doc to id)
        resourceIds.add(id)
    }
}
```

then iterate `validDocs` once, using the paired id where :692 currently re-extracts it. Keep the chunked DAO lookup (:683-687) and the body of the entity-building loop as they are.

### Constraints — behaviour must be identical

- Same validity rule: `_id.isNotBlank() && !_id.startsWith("_design")`, evaluated on `JsonUtils.getString("_id", doc)`.
- Same document order in both `validDocs` and `resourceIds`.
- **Do not de-duplicate `resourceIds`.** Today it can contain repeats, and that shapes the `chunked(900)` boundaries; keep `myLibraryDao.getByIds` receiving the same chunks it receives now.
- Keep the `existingItems` map behaviour exactly: it is seeded from the DAO (:685) and then **written back inside the loop** (`existingItems[_id] = library`, :702), so a later document with the same `_id` sees the entity built from the earlier one. That write must stay.
- Keep the per-document `try { … } catch (e: Exception) { e.printStackTrace() }` (:691, :706-708) — one bad document must not abort the batch. Do not convert it to a log call, and do not widen it to wrap the whole loop.
- `savedIds` must accumulate the same ids in the same order, and only for documents where `MyLibrary.insertMyLibrary(...)` returned non-null.
- Keep the tail block unchanged: `if (librariesToUpsert.isNotEmpty()) { myLibraryDao.upsertAll(...); reconcileHtmlLibraries(...); clearResourceListCache() }` (:710-714).
- Do not touch `batchInsertMyLibrary`, `reconcileHtmlLibraries`, `clearResourceListCache`, or `MyLibrary.insertMyLibrary`.

### Tests

Extend `app/src/test/java/org/ole/planet/myplanet/repository/ResourcesRepositoryImplTest.kt` (MockK, per the project convention) and keep every existing case passing. Cover:

- documents with a blank `_id` and with a `_design` prefix are skipped, and the returned `savedIds` contains only the accepted ids in input order,
- `myLibraryDao.getByIds` is called with the same ids that survive the filter,
- a document that makes `insertMyLibrary` return null (or throw) does not abort the batch and does not appear in `savedIds`.

### Out of scope

Two other open PRs edit this file but not this function — #17051 around :158, :312-320, :347 and the removed `getFilterFacets` block ending near :640, and #17095 in `getOfflineResourceItems` at :846. Keep your diff inside `batchInsertResources` (:674-716) so the three do not collide.

dogi:
Refactors batchInsertResources in ResourcesRepositoryImpl to extract document IDs and filter valid documents in a single pass over documents, reusing paired IDs during entity processing loop. Extends ResourcesRepositoryImplTest to cover filtering, order retention, DAO calls, and error handling.

Fixes #17104

---
*PR created automatically by Jules for task [16614049430035301888](https://jules.google.com/task/16614049430035301888) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/repository/ResourcesRepositoryImpl.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/ResourcesRepositoryImpl.kt
@@ -674,11 +674,15 @@ class ResourcesRepositoryImpl @Inject constructor(
     override suspend fun batchInsertResources(documents: List<JsonObject>): List<String> {
         val savedIds = mutableListOf<String>()
 
-        val validDocs = documents.filter {
-            val _id = JsonUtils.getString("_id", it)
-            _id.isNotBlank() && !_id.startsWith("_design")
+        val validDocs = ArrayList<Pair<JsonObject, String>>(documents.size)
+        val resourceIds = ArrayList<String>(documents.size)
+        for (doc in documents) {
+            val id = JsonUtils.getString("_id", doc)
+            if (id.isNotBlank() && !id.startsWith("_design")) {
+                validDocs.add(doc to id)
+                resourceIds.add(id)
+            }
         }
-        val resourceIds = validDocs.map { JsonUtils.getString("_id", it) }
         val existingItems = mutableMapOf<String, MyLibrary>()
         if (resourceIds.isNotEmpty()) {
             resourceIds.chunked(900).forEach { chunk ->
@@ -687,9 +691,8 @@ class ResourcesRepositoryImpl @Inject constructor(
         }
 
         val librariesToUpsert = mutableListOf<MyLibrary>()
-        validDocs.forEach { doc ->
+        validDocs.forEach { (doc, _id) ->
             try {
-                val _id = JsonUtils.getString("_id", doc)
                 val existing = existingItems[_id]
                 val library = MyLibrary.insertMyLibrary(
                     MyLibrary.Companion.InsertParams(
--- a/app/src/test/java/org/ole/planet/myplanet/repository/ResourcesRepositoryImplTest.kt
+++ b/app/src/test/java/org/ole/planet/myplanet/repository/ResourcesRepositoryImplTest.kt
@@ -730,6 +730,66 @@ class ResourcesRepositoryImplTest {
     }
 
     @Test
+    fun `batchInsertResources skips blank ids and design docs and returns accepted ids in input order`() = runTest {
+        val docBlank = com.google.gson.JsonObject().apply { addProperty("_id", "  ") }
+        val docDesign = com.google.gson.JsonObject().apply { addProperty("_id", "_design/lib") }
+        val doc1 = com.google.gson.JsonObject().apply {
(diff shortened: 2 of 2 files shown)


dogi:
**Review — rating: 86/100**

Folds the `filter` / `map` / re-extract sequence in `batchInsertResources` into one loop that pulls `_id` once per document.

Master called `JsonUtils.getString("_id", …)` **three times for every valid document** — once in the `filter` predicate, once building `resourceIds`, and once more inside the `validDocs.forEach`. That's on the sync insert path, where `documents` arrives in server-sized batches, so it's a real per-document cost rather than a theoretical one. Carrying `doc to id` through means one extraction each.

The filter predicate is copied exactly (`id.isNotBlank() && !id.startsWith("_design")`), `resourceIds` is populated in the same order as `validDocs`, and the destructured `forEach { (doc, _id) }` binds the identical value the third call used to produce. The 900-element chunking for the `existingItems` lookup is untouched, which matters — that's the SQLite variable-limit guard. Preallocating both `ArrayList`s to `documents.size` is a sensible touch given the filter usually keeps nearly everything.

Building `resourceIds` in the same loop rather than deriving it as `validDocs.map { it.second }` afterwards is the right call — deriving it would put back an allocation the PR is trying to remove.

60 lines of new `ResourcesRepositoryImplTest` coverage, including (from the shape of it) the `_design` and blank-id exclusions that the predicate depends on.

**Nit:** `Pair<JsonObject, String>` is serviceable but anonymous at the use site; a local `data class ValidDoc(val doc: JsonObject, val id: String)` would make `validDocs.forEach { (doc, _id) }` self-documenting. Minor, and the destructuring already reads well enough.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17116, opened 2026-09-10
all: smoother configurations repository storage path resolving (fixes #17111)

### LINKED ISSUE #17111: Give ConfigurationsRepositoryImpl a path resolver instead of Context + FileUtils

`repository/ConfigurationsRepositoryImpl.kt` reaches into filesystem helpers from an otherwise configuration-only repository. There are exactly two such call sites:

- **:249** — `val f = FileUtils.getSDPathFromUrl(context, path)`, whose `File` result is handed to `Sha256Utils().getCheckSumFromFile(f)` at :252
- **:416** — `val myDir = File(FileUtils.getOlePath(context))`, then `myDir.isDirectory` / `myDir.listFiles()?.forEach { it.deleteRecursively() }` inside `clearFirstRunStorageAndSetFlag`

Those two are the only uses of `FileUtils` and of `java.io.File` in the file (imports at :38 and :10). The helper signatures are `FileUtils.getSDPathFromUrl(context: Context, url: String?): File` (`utils/FileUtils.kt:64`) and `FileUtils.getOlePath(context: Context): String` (`utils/FileUtils.kt:43`).

### Do

1. Add `utils/StoragePathResolver.kt` — an injectable class holding the `@ApplicationContext` and delegating to the existing helpers:

```kotlin
class StoragePathResolver @Inject constructor(
    @ApplicationContext private val context: Context
) {
    fun resolveFileFromUrl(url: String?): File = FileUtils.getSDPathFromUrl(context, url)
    fun resolveOleDirectory(): File = File(FileUtils.getOlePath(context))
}
```

Hilt can construct this from its `@Inject constructor`, so no module entry is needed.

2. Inject `private val storagePathResolver: StoragePathResolver` into `ConfigurationsRepositoryImpl`.
3. Replace :249 with `storagePathResolver.resolveFileFromUrl(path)` and :416 with `storagePathResolver.resolveOleDirectory()`.
4. Drop the `FileUtils` (:38) and `java.io.File` (:10) imports, which become unused.

### Constraints

- **Delegate, do not reimplement.** `resolveFileFromUrl` must call `FileUtils.getSDPathFromUrl` and `resolveOleDirectory` must wrap `FileUtils.getOlePath` — do not inline or rewrite the path logic, and do not change `FileUtils` itself.
- Keep the nullable parameter type: `getSDPathFromUrl` takes `url: String?`, so `resolveFileFromUrl(url: String?)` must too.
- `resolveOleDirectory()` returns the `File`, matching what :416 builds today. Do not return a `String` path and re-wrap it at the call site.
- **Check before removing the `context` injection.** Only drop `ConfigurationsRepositoryImpl`'s `@ApplicationContext context` parameter if the class has no other use of it after this change; if anything else still reads it, leave the injection in place. Do not remove a constructor parameter that is still referenced.
- Surrounding behaviour is unchanged: the `f.exists()` guard at :250, the `withContext(dispatcherProvider.io)` around the checksum read (:251), and the `hasWritePermission && sharedPrefManager.getFirstRun()` guard plus `setFirstRun(false)` in `clearFirstRunStorageAndSetFlag` (:413-420) all stay exactly as they are.
- Do not touch `UrlUtils.getChecksumUrl`, `Sha256Utils`, or the API call at :245.

### Tests

Keep `app/src/test/java/org/ole/planet/myplanet/repository/ConfigurationsRepositoryImplTest.kt` (if present) passing — the repository gains a constructor argument, so pass a resolver stubbed with MockK returning a temp `File`. That is the payoff: the checksum path and the first-run cleanup become testable without mocking the static `FileUtils` object against a real `Context`. Add coverage that `clearFirstRunStorageAndSetFlag` deletes the resolver-provided directory's contents and sets the first-run flag false, and that it does nothing when `hasWritePermission` is false.

### Out of scope

Do not touch `AppDatabase` or any Room schema.

**Do not add or change any log line in this file, and never log or toast a resolved path, URL, or checksum URL.** myPlanet server URLs embed `satellite:<pin>@host` credentials and release builds do not strip logs (`minifyEnabled` is `false`).

dogi:
Added StoragePathResolver class with Hilt @Inject constructor delegating path resolution to FileUtils.
Injected StoragePathResolver into ConfigurationsRepositoryImpl and replaced direct FileUtils/File calls.
Updated ConfigurationsRepositoryImplTest to mock StoragePathResolver instead of FileUtils.

Fixes #17111

---
*PR created automatically by Jules for task [4802198934201231465](https://jules.google.com/task/4802198934201231465) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/repository/ConfigurationsRepositoryImpl.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/ConfigurationsRepositoryImpl.kt
@@ -7,7 +7,6 @@ import androidx.core.net.toUri
 import com.google.gson.Gson
 import com.google.gson.JsonObject
 import dagger.hilt.android.qualifiers.ApplicationContext
-import java.io.File
 import java.io.IOException
 import java.net.ConnectException
 import java.net.SocketTimeoutException
@@ -35,10 +34,10 @@ import org.ole.planet.myplanet.services.SharedPrefManager
 import org.ole.planet.myplanet.services.sync.ServerUrlMapper
 import org.ole.planet.myplanet.utils.Constants
 import org.ole.planet.myplanet.utils.DispatcherProvider
-import org.ole.planet.myplanet.utils.FileUtils
 import org.ole.planet.myplanet.utils.LocaleUtils
 import org.ole.planet.myplanet.utils.NetworkUtils
 import org.ole.planet.myplanet.utils.Sha256Utils
+import org.ole.planet.myplanet.utils.StoragePathResolver
 import org.ole.planet.myplanet.utils.TimeProvider
 import org.ole.planet.myplanet.utils.UrlUtils
 import org.ole.planet.myplanet.utils.VersionUtils
@@ -52,6 +51,7 @@ class ConfigurationsRepositoryImpl @Inject constructor(
     private val serverUrlMapper: ServerUrlMapper,
     private val dispatcherProvider: DispatcherProvider,
     private val timeProvider: TimeProvider,
+    private val storagePathResolver: StoragePathResolver,
     @PlainGson private val gson: Gson
 ) : ConfigurationsRepository {
     private val serverAvailabilityCache = ConcurrentHashMap<String, Pair<Boolean, Long>>()
@@ -246,7 +246,7 @@ class ConfigurationsRepositoryImpl @Inject constructor(
             if (response.isSuccessful) {
                 val checksum = withContext(dispatcherProvider.io) { response.body()?.string() }
                 if (!checksum.isNullOrEmpty()) {
-                    val f = FileUtils.getSDPathFromUrl(context, path)
+                    val f = storagePathResolver.resolveFileFromUrl(path)
                     if (f.exists()) {
                         val sha256 = withContext(dispatcherProvider.io) {
                             Sha256Utils().getCheckSumFromFile(f)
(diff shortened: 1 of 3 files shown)


dogi:
**Review — rating: 86/100**

Introduces `StoragePathResolver` and routes `ConfigurationsRepositoryImpl`'s two filesystem lookups — `FileUtils.getSDPathFromUrl(context, path)` and `File(FileUtils.getOlePath(context))` — through it.

Same shape as #17106 and the same real justification: these were static calls into `FileUtils` taking a `Context`, so a unit test had no seam to intercept them and had to stand up enough of the environment to make real path resolution work. The test diff tells that story — `ConfigurationsRepositoryImplTest` goes −26/+14, which is a substantial simplification for a two-call-site change and is the clearest evidence the seam was worth adding.

The resolver's surface is well chosen: `resolveFileFromUrl` and `resolveOleDirectory` are named for what callers want rather than mirroring `FileUtils`' signatures, and pulling `File(...)` construction inside means the repository no longer imports `java.io.File` or `FileUtils` at all. Checksum verification (`getCheckSumFromFile`) and the first-run wipe both keep their exact behaviour — the wipe still guards on `isDirectory` before `listFiles()?.forEach { deleteRecursively() }`, which matters given what that code does.

**Nit:** `@ApplicationContext context` stays on `ConfigurationsRepositoryImpl` for its other uses, so this is a partial extraction rather than a clean break. That's fine and correctly scoped — just worth knowing the repository is still `Context`-bound if the goal is eventually to get it out entirely.

**Nit:** consider whether `StoragePathResolver` and #17106's `AppVersionProvider` want to live together — they're the same idea (a thin injectable seam over a static/global), both landing in `utils/` in the same batch.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17117, opened 2026-09-10
login: smoother server config utils server caching (fixes #17110)

### LINKED ISSUE #17110: Memoize the BuildConfig-derived server lists in ServerConfigUtils

`utils/ServerConfigUtils.kt` (an `object`) rebuilds two constant lists on every call:

- `getTrustedServerHosts()` at **:140-157** — `listOfNotNull(...)` over **14** `BuildConfig.PLANET_*_URL` constants, each wrapped in `.takeIf { it.isNotEmpty() }`.
- `getChallengeServerUrls()` at **:159-168** — `listOfNotNull(...)` over **6** of those constants, each `.takeIf { it.isNotEmpty() }?.let { "https://$it" }` / `"http://$it"`.

Both take no parameters and read nothing but `BuildConfig` compile-time constants, so the result cannot change within a process. Callers:

- `ui/viewer/WebViewActivity.kt:55` → `getTrustedServerHosts()`
- `ui/dashboard/DashboardActivity.kt:195` → `getChallengeServerUrls()`

### Do

Back exactly these two functions with `by lazy` vals, keeping the public function names and signatures so the two call sites are untouched. For example:

```kotlin
private val trustedServerHosts: List<String> by lazy { listOfNotNull( … ) }   // body moved verbatim
fun getTrustedServerHosts(): List<String> = trustedServerHosts
```

and the same shape for `getChallengeServerUrls()`.

### Constraints

- **Move the list bodies verbatim.** Same constants, same order, same `takeIf { it.isNotEmpty() }` guards, same `http://` vs `https://` prefixes per entry (`getChallengeServerUrls` mixes them deliberately: `https` for `PLANET_GUATEMALA_URL` and `PLANET_VI_URL`, `http` for the other four). Do not sort, de-duplicate, or "tidy" the lists.
- Keep both functions `public` with their current names and return type `List<String>`; do not expose the lazy vals.
- **Do not cache `getServerAddresses` (:29-43).** Its strings come from `context.getString(...)`, they are locale-dependent, and the app supports runtime locale changes — a cached list would go stale in the wrong language. Leave that function exactly as it is. Same for `getCommunityConfig` (:132-138), which takes parameters.
- Do not touch `saveAlternativeUrl` (:110-127) or anything else in the file — another open PR (#17072) rewrites that function's preference writes into `sharedPrefManager.saveAlternativeServerConfig(...)`, so keep your diff to :140-168 plus the two new private vals so the two do not collide.
- The returned lists must stay read-only `List<String>` — do not hand out a `MutableList`, since a memoized instance is now shared across all callers for the process lifetime.

### Tests

Add coverage to `app/src/test/java/org/ole/planet/myplanet/utils/ServerConfigUtilsTest.kt` (MockK, per the project convention) and keep every existing case in it passing — but keep the additions **at the end of the class**: #17072 rewrites the earlier tests in that file (around :32-160) into single `saveAlternativeServerConfig` verifications.

Assert that repeated calls return an equal list and that the entries match the non-empty `BuildConfig` constants with the right scheme prefix. Do **not** assert on concrete host names or URLs from `BuildConfig` in test output or messages.

### Out of scope

Do not touch `AppDatabase` or any Room schema.

**Do not add or change any log line in this file, and do not print, toast, or log any URL, host, or list built here.** myPlanet server URLs embed `satellite:<pin>@host` credentials (`gradle.properties` is tracked and `minifyEnabled` is `false`), and release builds do not strip logs.

dogi:
Backs getTrustedServerHosts and getChallengeServerUrls with private lazy properties in ServerConfigUtils and adds unit tests covering memoization and correctness.

Fixes #17110

---
*PR created automatically by Jules for task [8122358952103144360](https://jules.google.com/task/8122358952103144360) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/utils/ServerConfigUtils.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/utils/ServerConfigUtils.kt
@@ -137,8 +137,8 @@ object ServerConfigUtils {
         return CommunityConfig(domain, protocol, pin, isPinEnabled)
     }
 
-    fun getTrustedServerHosts(): List<String> {
-        return listOfNotNull(
+    private val _trustedServerHosts: List<String> by lazy {
+        listOfNotNull(
             BuildConfig.PLANET_LEARNING_URL.takeIf { it.isNotEmpty() },
             BuildConfig.PLANET_GUATEMALA_URL.takeIf { it.isNotEmpty() },
             BuildConfig.PLANET_SANPABLO_URL.takeIf { it.isNotEmpty() },
@@ -156,8 +156,10 @@ object ServerConfigUtils {
         )
     }
 
-    fun getChallengeServerUrls(): List<String> {
-        return listOfNotNull(
+    fun getTrustedServerHosts(): List<String> = _trustedServerHosts
+
+    private val _challengeServerUrls: List<String> by lazy {
+        listOfNotNull(
             BuildConfig.PLANET_GUATEMALA_URL.takeIf { it.isNotEmpty() }?.let { "https://$it" },
             BuildConfig.PLANET_XELA_URL.takeIf { it.isNotEmpty() }?.let { "http://$it" },
             BuildConfig.PLANET_URIUR_URL.takeIf { it.isNotEmpty() }?.let { "http://$it" },
@@ -166,4 +168,6 @@ object ServerConfigUtils {
             BuildConfig.PLANET_VI_URL.takeIf { it.isNotEmpty() }?.let { "https://$it" }
         )
     }
+
+    fun getChallengeServerUrls(): List<String> = _challengeServerUrls
 }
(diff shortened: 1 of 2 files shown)


dogi:
**Review — rating: 86/100**

Backs `getTrustedServerHosts()` and `getChallengeServerUrls()` with `by lazy` instead of rebuilding the lists on every call.

Safe by construction: every element comes from a `BuildConfig.PLANET_*_URL` constant, which is baked in at compile time and cannot change while the process runs. So there's no invalidation story to get wrong — unlike a cache over mutable state, this one has no staleness window at all. `by lazy` defaults to `SYNCHRONIZED`, so concurrent first calls are handled without extra work.

The saving is small per call but the lists aren't trivial — 13 and 6 entries respectively, each with a `takeIf { it.isNotEmpty() }` and, for the challenge list, a scheme-prefixing `let`. These get consulted on server-reachability and challenge paths rather than once at startup, so rebuilding them each time was pure repetition.

Keeping the public accessors as functions rather than flipping them to `val`s means no call site changes, which keeps the diff honest at +52/−4 with the tests being most of it. The 44 lines of new `ServerConfigUtilsTest` are the right kind — asserting the same instance comes back is what actually proves memoization, as opposed to just re-asserting contents.

**Nit:** the `_trustedServerHosts` / `_challengeServerUrls` naming borrows a convention that usually signals a mutable backing field behind a read-only public property (`_state` / `state`). Nothing mutable is involved here, so plain `trustedServerHostsCache`, or making the `val` itself the public API, would read more accurately.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17118, opened 2026-09-10
teams: smoother surveys filter view modelling (fixes #17113)

### LINKED ISSUE #17113: Tokenize survey searches without the temporary split list
### Verified against `origin/master` (`5b15422`)

`ui/surveys/SurveysViewModel.kt:164-167`, at the top of `filter`:

```kotlin
private fun filter(s: String, list: List<StepExam>): List<StepExam> {
    val queryParts = s.split(" ").filterNot { it.isEmpty() }        // :165 — two lists
    val normalizedQueryParts = queryParts.map { Utilities.normalizeText(it) }   // :166 — a third
    val normalizedQuery = Utilities.normalizeText(s)
```

`s.split(" ")` materializes a list, `filterNot { it.isEmpty() }` materializes a second, and `.map { normalizeText(it) }` a third — and `queryParts` itself is never used again after :166. This runs on every debounced query.

### Do

Collapse :165-166 into one materialized collection, keeping `normalizedQueryParts` as the only survivor:

```kotlin
val normalizedQueryParts = s.splitToSequence(" ")
    .filterNot { it.isEmpty() }
    .map { Utilities.normalizeText(it) }
    .toList()
val normalizedQuery = Utilities.normalizeText(s)
```

Delete the now-unused `queryParts`.

### Constraints — matching behaviour must be identical

- Split on the single space `" "`, exactly as today. Do not switch to a whitespace regex, `trim()`, or `split()` with multiple delimiters — that would change which queries match.
- Keep the whole-query normalization `normalizedQuery = Utilities.normalizeText(s)` (:167) as a separate value; it is what the `startsWith` bucket uses.
- Keep the two buckets and their order: `startsWithQuery` collects items whose normalized title `startsWith(normalizedQuery, ignoreCase = true)`, `containsQuery` collects items where `normalizedQueryParts.all { title.contains(it, ignoreCase = true) }`, and the return is `startsWithQuery + containsQuery` (:168-179). Do not merge, reorder or dedupe the buckets.
- Keep `else if` — an item that starts with the query must **not** also be tested for the contains bucket.
- Keep the `item.name?.let { Utilities.normalizeText(it) } ?: continue` guard at :172, which skips null-named items.
- `normalizedQueryParts` must stay a materialized `List`, not a `Sequence`: it is consumed once per item by `.all { }` inside the loop, and a sequence would be re-evaluated (and, for a single-use sequence, would be wrong after the first item).
- An empty or blank query must behave exactly as it does now: an empty token list makes `.all { }` vacuously true, and that is existing behaviour to preserve.
- Keep the call site and the `dispatcherProvider.default` context this runs under unchanged; do not touch `getSortDate` (:156-162) or the debounce.

### Tests

Add to `app/src/test/java/org/ole/planet/myplanet/ui/surveys/SurveysViewModelTest.kt` if it exists, otherwise create it, and keep existing cases passing. Cover:

- a multi-word query matches a survey whose title contains all tokens in any order, and is placed in the contains bucket,
- a query that is a prefix of a title puts that survey in the starts-with bucket, ahead of contains matches,
- a query with repeated / leading / trailing spaces (`"  math  quiz "`) matches the same set as `"math quiz"`,
- an empty query returns every named survey, and a survey with a null `name` is skipped.

### Out of scope

dogi:
Refactored `filter` in `SurveysViewModel` to use `s.splitToSequence(" ")` and materializes only the final `normalizedQueryParts` list, deleting `queryParts`. Added unit tests in `SurveysViewModelTest` to cover token order matching, starts-with vs. contains bucket ordering, whitespace handling, and empty/null search conditions.

Fixes #17113

---
*PR created automatically by Jules for task [3270793116892866630](https://jules.google.com/task/3270793116892866630) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/ui/surveys/SurveysViewModel.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/surveys/SurveysViewModel.kt
@@ -162,8 +162,10 @@ class SurveysViewModel @Inject constructor(
     }
 
     private fun filter(s: String, list: List<StepExam>): List<StepExam> {
-        val queryParts = s.split(" ").filterNot { it.isEmpty() }
-        val normalizedQueryParts = queryParts.map { Utilities.normalizeText(it) }
+        val normalizedQueryParts = s.splitToSequence(" ")
+            .filterNot { it.isEmpty() }
+            .map { Utilities.normalizeText(it) }
+            .toList()
         val normalizedQuery = Utilities.normalizeText(s)
         val startsWithQuery = mutableListOf<StepExam>()
         val containsQuery = mutableListOf<StepExam>()
--- a/app/src/test/java/org/ole/planet/myplanet/ui/surveys/SurveysViewModelTest.kt
+++ b/app/src/test/java/org/ole/planet/myplanet/ui/surveys/SurveysViewModelTest.kt
@@ -194,4 +194,91 @@ class SurveysViewModelTest {
         assertEquals(1, viewModel.surveys.value.size)
         assertEquals("2", viewModel.surveys.value[0].exam.id)
     }
+
+    @Test
+    fun `test multi-word query matches survey containing tokens in any order in contains bucket`() = runTest {
+        backgroundScope.launch(testDispatcher) { viewModel.surveys.collect {} }
+        val survey1 = createSurvey("1", "Science Quiz for Math Students", 1000L, 0L)
+        val survey2 = createSurvey("2", "Math Test Only", 2000L, 0L)
+
+        stubLoadSurveys(listOf(survey1, survey2))
+
+        viewModel.loadSurveys(false, null, false)
+        testDispatcher.scheduler.advanceUntilIdle()
+
+        viewModel.search("quiz math")
+        testDispatcher.scheduler.advanceUntilIdle()
+
+        val results = viewModel.surveys.value
+        assertEquals(1, results.size)
+        assertEquals("1", results[0].exam.id)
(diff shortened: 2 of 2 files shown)


dogi:
**Review — rating: 85/100**

Swaps `s.split(" ").filterNot { it.isEmpty() }.map { normalizeText(it) }` for the `splitToSequence` equivalent with a single `toList()` at the end.

Semantically identical — `splitToSequence` applies the same delimiter rules as `split`, and lazily filtering then mapping produces the same elements in the same order. What changes is allocation: the eager chain materialised three lists (the raw split, the filtered result, the mapped result) where the sequence version builds one. The intermediate `queryParts` val disappears entirely, and I checked it wasn't referenced further down — only `normalizedQueryParts` and `normalizedQuery` are used in the ranking loop below.

`filter` runs on every survey-search keystroke, so shaving two throwaway lists per call is worth having even though each is small.

The 87 lines of new `SurveysViewModelTest` are the more valuable half — this filter has non-trivial ranking behaviour (`startsWithQuery` before `containsQuery`, all-parts-must-match for the contains case) that had no coverage, so the tests pin semantics the refactor could have disturbed rather than just asserting the happy path.

**Nit:** at realistic query lengths — a handful of words — sequence setup overhead is in the same ballpark as the allocations saved, so treat this as tidiness plus a guard against pathological inputs rather than a measurable win. Nothing to change; just worth not over-claiming in the description.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17119, opened 2026-09-10
courses: smoother title sort view modelling (fixes #17112)

### LINKED ISSUE #17112: Sort course titles without allocating lowercase copies
### Verified against `origin/master` (`5b15422`)

`ui/courses/CoursesViewModel.kt:75-89`:

```kotlin
private fun sortCourses(courses: List<Course>): List<Course> {
    return when (activeSort) {
        SortType.TITLE -> if (isTitleAscending) {
            courses.sortedBy { it.courseTitle.lowercase() }        // :78
        } else {
            courses.sortedByDescending { it.courseTitle.lowercase() }   // :80
        }
        SortType.DATE -> if (isDateAscending) { … }
        null -> courses
    }
}
```

Each title sort allocates one lowercase `String` per element, and the sort calls the selector more than once per element, so a list of N courses produces well over N throwaway strings on every title sort.

### Do

Sort with a case-insensitive comparator instead of a lowercase key, for the two `SortType.TITLE` branches only:

```kotlin
SortType.TITLE -> {
    val byTitle = compareBy(String.CASE_INSENSITIVE_ORDER) { course: Course -> course.courseTitle }
    if (isTitleAscending) courses.sortedWith(byTitle) else courses.sortedWith(byTitle.reversed())
}
```

### Constraints

- **Leave the `SortType.DATE` branches (:82-86) and the `null -> courses` branch (:87) exactly as they are.**
- **Leave `applySort` (:62-73) alone** — the `sortJob?.cancel()`, the `withContext(dispatcherProvider.default)` hop, and especially the stale-result guard `if (_coursesState.value.courses === currentCourses)` (:69), which relies on reference identity. Do not change what `sortCourses` returns for `null` (it must keep returning the *same* list instance, not a copy, or that identity check breaks).
- **Tie ordering must not change.** Both today's `sortedBy`/`sortedByDescending` and `sortedWith` are stable, and `Comparator.reversed()` flips only the comparison result — elements that compare equal keep their original relative order in both directions. Do not implement descending by `sortedWith(byTitle).reversed()` on the *list*, which would also reverse the ties.
- Keep `courseTitle`'s current type handling — it is non-nullable here, so do not add `?: ""` padding or an `orEmpty()`.
- `String.CASE_INSENSITIVE_ORDER` and `String.lowercase()` are both locale-independent, so ASCII titles order identically. If a test surfaces a difference on non-ASCII titles, keep the comparator and say so in the PR description rather than reaching for a locale-specific `Collator` — that would be a separate, deliberate change.

### Tests

Add to `app/src/test/java/org/ole/planet/myplanet/ui/courses/CoursesViewModelTest.kt` if it exists, otherwise create it, and keep any existing cases passing. Cover:

- mixed-case titles (`apple`, `Banana`, `cherry`, `Date`) sort case-insensitively ascending and descending,
- two courses whose titles differ only in case (`Alpha`, `alpha`) keep their input relative order in **both** directions,
- date sorting is unaffected.

### Out of scope

dogi:
Replaced `it.courseTitle.lowercase()` sorting keys with `String.CASE_INSENSITIVE_ORDER` comparator in `CoursesViewModel.kt` to eliminate throwaway String allocations. Added unit tests in `CoursesViewModelTest.kt` verifying mixed-case sorting, stable relative tie ordering, and date sorting behavior.

Fixes #17112

---
*PR created automatically by Jules for task [13960011826324204639](https://jules.google.com/task/13960011826324204639) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/ui/courses/CoursesViewModel.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/courses/CoursesViewModel.kt
@@ -74,10 +74,9 @@ class CoursesViewModel @Inject constructor(
 
     private fun sortCourses(courses: List<Course>): List<Course> {
         return when (activeSort) {
-            SortType.TITLE -> if (isTitleAscending) {
-                courses.sortedBy { it.courseTitle.lowercase() }
-            } else {
-                courses.sortedByDescending { it.courseTitle.lowercase() }
+            SortType.TITLE -> {
+                val byTitle = compareBy(String.CASE_INSENSITIVE_ORDER) { course: Course -> course.courseTitle }
+                if (isTitleAscending) courses.sortedWith(byTitle) else courses.sortedWith(byTitle.reversed())
             }
             SortType.DATE -> if (isDateAscending) {
                 courses.sortedBy { it.createdDate }
--- a/app/src/test/java/org/ole/planet/myplanet/ui/courses/CoursesViewModelTest.kt
+++ b/app/src/test/java/org/ole/planet/myplanet/ui/courses/CoursesViewModelTest.kt
@@ -212,4 +212,84 @@ class CoursesViewModelTest {
         assertEquals(tagNames, state.tagNames)
         coVerify { coursesRepository.filterCourses("", "", "", tagNames) }
     }
+
+    @Test
+    fun testTitleSort_mixedCase_caseInsensitive() = runTest {
+        val c1 = MyCourse().apply { courseId = "1"; courseTitle = "apple" }
+        val c2 = MyCourse().apply { courseId = "2"; courseTitle = "Banana" }
+        val c3 = MyCourse().apply { courseId = "3"; courseTitle = "cherry" }
+        val c4 = MyCourse().apply { courseId = "4"; courseTitle = "Date" }
+
+        io.mockk.coEvery { coursesRepository.getAllCourses() } returns listOf(c1, c2, c3, c4)
+        io.mockk.coEvery { coursesRepository.getMyCourses(any(), any()) } returns emptyList()
+
+        viewModel.loadCourses(false, "u1")
+        testDispatcher.scheduler.advanceUntilIdle()
+
+        // Toggle title sort -> Ascending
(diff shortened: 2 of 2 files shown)


dogi:
**Review — rating: 86/100**

Replaces `sortedBy { it.courseTitle.lowercase() }` with a hoisted `compareBy(String.CASE_INSENSITIVE_ORDER) { it.courseTitle }`.

The allocation argument is stronger than "avoid a lowercase copy per element" suggests. `sortedBy`'s selector isn't memoised — it's invoked inside the comparator on **every comparison**, so a list of n courses allocates on the order of 2·n·log n throwaway strings, not n. `String.CASE_INSENSITIVE_ORDER` compares in place and allocates nothing.

Two things I checked because this is a behaviour-adjacent rewrite:

- **Tie ordering.** `sortedBy`/`sortedByDescending` are stable, and so is `sortedWith`, so equal-ignoring-case titles keep their input order in both directions. `byTitle.reversed()` reverses the comparator rather than the result list, so it doesn't flip ties the way `.sorted().reversed()` would.
- **Descending equivalence.** `compareByDescending(selector)` and `compareBy(cmp, selector).reversed()` produce the same ordering for a total comparator, so the `isTitleAscending == false` branch matches master.

Hoisting the comparator out of the `if` is a nice touch — it builds once instead of once per branch, and makes the two branches obviously the same ordering.

**Nit — a difference worth being aware of, not a defect.** Kotlin's no-arg `lowercase()` is `Locale.ROOT` full-string case folding, while `CASE_INSENSITIVE_ORDER` folds per-character. They agree across ASCII and the overwhelming majority of text, but can differ for characters whose case mapping changes length. Given both are locale-independent, this is strictly an exotic-Unicode concern.

**Aside:** neither version is locale-aware — both order by code point rather than collation rules — so titles in this app's non-Latin locales (ar, ne) still sort in a way a native reader wouldn't expect. That's pre-existing and out of scope here, but a `java.text.Collator` is the real answer if course-list ordering ever gets reported as wrong.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17125, opened 2026-09-10
community: smoother configurations repository tab view modelling (fixes #17120)

### LINKED ISSUE #17120: Read the community configuration as one repository snapshot

`ui/community/CommunityTabViewModel.kt:30-45` builds one `CommunityTabState` out of three separate repository calls:

```kotlin
init {
    viewModelScope.launch {
        val parentCode = configurationsRepository.getParentCode()      // :32
        val communityName = configurationsRepository.getCommunityName() // :33
        val planetType = configurationsRepository.getPlanetType()       // :34
        val user = userRepository.getUserModel()
        val planetCode = user?.planetCode.orEmpty()

        _state.value = CommunityTabState(planetCode, parentCode, communityName, planetType)
    }
}
```

Each of the three is its own preferences read in `repository/ConfigurationsRepositoryImpl.kt`:

```kotlin
override fun getPlanetType(): String? = sharedPrefManager.getRawString("planetType")   // :393-395
override fun getParentCode(): String = sharedPrefManager.getParentCode()               // :397-399
override fun getCommunityName(): String = sharedPrefManager.getCommunityName()         // :401-403
```

The value of grouping them is the **narrower boundary** — one call for one screen's configuration instead of three. To be explicit about what this is *not*: this is not a race fix. The three reads happen back-to-back with no suspension point between them, so there is no interleaving today; do not describe it as one in the PR.

### Do

1. Add an immutable snapshot type in the repository layer, next to `ConfigurationsRepository`:

```kotlin
data class CommunityConfiguration(
    val parentCode: String,
    val communityName: String,
    val planetType: String?,
)
```

Keep `planetType` nullable — `getPlanetType()` returns `String?` and `getParentCode()`/`getCommunityName()` return non-null `String`. Do not "clean this up" by widening or narrowing any of the three.

2. Add `fun getCommunityConfiguration(): CommunityConfiguration` to `ConfigurationsRepository`, implemented in `ConfigurationsRepositoryImpl` as the three existing reads assembled into the snapshot. Match the existing non-suspend style of these three getters — they are plain `fun`, not `suspend`, so the new one is too.

3. In `CommunityTabViewModel`, replace the three calls with one `configurationsRepository.getCommunityConfiguration()` and read the fields off it. **Keep resolving `planetCode` from the active user** (`userRepository.getUserModel()?.planetCode.orEmpty()`) — the planet code is not part of the configuration snapshot and must not move into it.

### Constraints

- **Keep the three individual getters** (`getPlanetType`, `getParentCode`, `getCommunityName`) on the interface and impl. Other callers use them; this task adds a grouped accessor, it does not replace them.
- `CommunityTabState`'s shape, field order and the `_state.value = …` assignment stay as they are.
- Do not touch `getCommunityLeaders` (:405-407), `clearPreferences` (:409+), or the `planetType` write path at :387-390.
- Do not add a Hilt module entry — nothing new is injected.

### Tests

Add to `app/src/test/java/org/ole/planet/myplanet/repository/ConfigurationsRepositoryImplTest.kt` and the `CommunityTabViewModel` test if one exists (create it otherwise), keeping every existing case passing. Cover: the snapshot carries the same three values the individual getters return, including a null `planetType`; and the ViewModel's state still takes `planetCode` from the user model, not from the snapshot.

Note that another open PR (#17116) also edits `ConfigurationsRepositoryImpl` — but only its imports, constructor and the two `FileUtils` call sites at ~:249 and ~:416, plus the test file. Keep your diff to the getters region (~:393-403) and the interface so the two do not collide, and expect the test file to need merging.

### Out of scope

Do not touch `AppDatabase` or any Room schema. **Do not add or change any log line, and never log the community name, parent code or any URL** — myPlanet server URLs embed `satellite:<pin>@host` credentials and release builds do not strip logs.

dogi:
Add CommunityConfiguration snapshot data class to ConfigurationsRepository and ConfigurationsRepositoryImpl, and update CommunityTabViewModel to read community configurations as a single repository snapshot while preserving individual getters and existing user planetCode behavior.

Fixes #17120

---
*PR created automatically by Jules for task [10302018287139874253](https://jules.google.com/task/10302018287139874253) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/repository/ConfigurationsRepository.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/ConfigurationsRepository.kt
@@ -4,6 +4,12 @@ import org.ole.planet.myplanet.model.MyPlanet
 import org.ole.planet.myplanet.model.UserEntity
 import org.ole.planet.myplanet.services.SharedPrefManager
 
+data class CommunityConfiguration(
+    val parentCode: String,
+    val communityName: String,
+    val planetType: String?
+)
+
 interface ConfigurationsRepository {
     suspend fun checkHealth(): String
     fun checkVersion(callback: CheckVersionCallback, spm: SharedPrefManager)
@@ -15,6 +21,7 @@ interface ConfigurationsRepository {
     fun getPlanetType(): String?
     fun getParentCode(): String
     fun getCommunityName(): String
+    fun getCommunityConfiguration(): CommunityConfiguration
     fun getCommunityLeaders(): List<UserEntity>
     fun clearPreferences()
     suspend fun ensureServerUrlUpdated()
--- a/app/src/main/java/org/ole/planet/myplanet/repository/ConfigurationsRepositoryImpl.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/ConfigurationsRepositoryImpl.kt
@@ -402,6 +402,14 @@ class ConfigurationsRepositoryImpl @Inject constructor(
         return sharedPrefManager.getCommunityName()
     }
 
+    override fun getCommunityConfiguration(): CommunityConfiguration {
+        return CommunityConfiguration(
+            parentCode = getParentCode(),
+            communityName = getCommunityName(),
+            planetType = getPlanetType()
+        )
+    }
+
     override fun getCommunityLeaders(): List<UserEntity> {
         return UserEntity.parseLeadersJson(sharedPrefManager.getCommunityLeaders())
     }
--- a/app/src/main/java/org/ole/planet/myplanet/ui/community/CommunityTabViewModel.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/community/CommunityTabViewModel.kt
@@ -29,17 +29,15 @@ class CommunityTabViewModel @Inject constructor(
 
     init {
         viewModelScope.launch {
-            val parentCode = configurationsRepository.getParentCode()
-            val communityName = configurationsRepository.getCommunityName()
-            val planetType = configurationsRepository.getPlanetType()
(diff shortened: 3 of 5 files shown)


dogi:
**Review — rating: 82/100**

Adds a `CommunityConfiguration` record and `getCommunityConfiguration()`, so `CommunityTabViewModel` makes one repository call instead of three.

The API shape is the real gain and it's a legitimate one: the ViewModel's coupling to `ConfigurationsRepository` drops from three methods to one, the three values that always travel together now do so as a typed unit, and `CommunityTabState` is populated from a single source rather than three independently-fetched fields. `CommunityTabViewModelTest` gets simpler for it.

**But the title promises more than the implementation delivers.** "Single repository snapshot" reads as *atomic*, and it isn't — `getCommunityConfiguration()` calls `getParentCode()`, `getCommunityName()` and `getPlanetType()` in sequence, each hitting `SharedPreferences` separately. The three reads have simply moved from the ViewModel into the repository; a concurrent write between them can still produce a torn view, exactly as before. So this is an interface consolidation, not a consistency fix, and the description should say so.

That distinction matters more than usual right now because **#17072 in this queue is making the corresponding server-config *writes* atomic** — batching them into single `SharedPreferences` commits precisely so readers can't observe half-updated state. This PR is the natural read-side counterpart to that work, and it could actually become one cheaply: have the impl read once from a single `SharedPreferences` snapshot rather than through three getters. Then the title would be true and the pairing with #17072 would be complete.

As it stands nothing is worse than master and the call site is cleaner, so I'm not holding it — I'd just like the description corrected, and it's worth considering whether to finish the job while you're here.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17128, opened 2026-09-10
all: smoother crash log store utils exception logging (fixes #17122)

### LINKED ISSUE #17122: Log the swallowed exceptions in CrashLogStore instead of printStackTrace

`utils/CrashLogStore.kt` swallows two failures with a bare `printStackTrace()`:

- **`:44-47`** — the `save` path's catch, wrapping directory creation (`dir(context)`, `mkdirs()`), the `MAX_PENDING_FILES` capacity check, and `file.writeText(error)`:

```kotlin
} catch (e: Exception) {
    e.printStackTrace()
    null
}
```

- **`:56-59`** — inside `loadPendingLogs`'s `mapNotNull`, wrapping `PendingLog(file, parsed.type, parsed.time, file.readText())`:

```kotlin
} catch (e: Exception) {
    e.printStackTrace()
    null
}
```

The file has **no** `android.util.Log` import and **no** `TAG` constant today. This is the crash-log store specifically: a swallowed write or read failure here means a crash report was silently lost, which is exactly the thing worth a real log line.

### Do

1. Add a `private const val TAG = "CrashLogStore"` in the object's companion/top-level constants (match however the file already declares `MAX_PENDING_FILES` and `FILE_EXTENSION`), and `import android.util.Log`.
2. Replace `:45` with `Log.e(TAG, "failed to save crash log", e)`.
3. Replace `:57` with `Log.e(TAG, "failed to read pending crash log", e)`.

Keep both `null` returns and both `catch (e: Exception)` blocks exactly where they are — the swallow behaviour does not change, only the reporting.

### Constraints — read this part carefully

- **The message strings must be static literals.** Do not interpolate anything into them: no file name, no path, no directory, no `error` payload, no URL. myPlanet server URLs embed `satellite:<pin>@host` credentials, `gradle.properties` is tracked, `minifyEnabled` is `false`, and release builds do not strip logs — so any interpolated value in a log line ships to production logcat.
- Pass the exception as the third `Log.e` argument (`Log.e(TAG, msg, e)`). Do not do `Log.e(TAG, e.message)` or `Log.e(TAG, "… ${e.message}")`.
- Do not add logging anywhere else in the file — the `return null` paths at `:39` (mkdirs failed) and `:40` (at capacity) are deliberate non-exceptional early returns and stay silent.
- Do not change `MAX_PENDING_FILES`, `isValidLogFile`, `parseLogFile`, `dir()`, the file-name format `"${timeProvider.now()}_$type$FILE_EXTENSION"`, or the `PendingLog` shape.
- Do not widen or narrow either `catch`.

### Tests

Keep `app/src/test/java/org/ole/planet/myplanet/utils/CrashLogStoreTest.kt` (if present) passing. A test asserting a log line fired is not required and not wanted here; if you add anything, assert the existing contract instead — `save` still returns null when the write fails, and `loadPendingLogs` still skips an unreadable file rather than throwing.

### Out of scope

Do not touch `AppDatabase` or any Room schema. Note that PR #17015 is still open against this same file's capacity check (`save`'s `MAX_PENDING_FILES` short-circuit, awaiting a decision from @dogi) — keep your diff to the two catch blocks and the new TAG/import so the two do not collide, and do not touch the capacity-check line at `:40`.

dogi:
Updates CrashLogStore to log swallowed write/read exceptions using Android's Log.e with static message literals instead of printStackTrace().

Fixes #17122

---
*PR created automatically by Jules for task [916669657837271074](https://jules.google.com/task/916669657837271074) started by @dogi*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error reporting when saving or loading crash logs, making failures easier to diagnose.

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/utils/CrashLogStore.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/utils/CrashLogStore.kt
@@ -1,6 +1,7 @@
 package org.ole.planet.myplanet.utils
 
 import android.content.Context
+import android.util.Log
 import java.io.File
 
 /**
@@ -10,6 +11,7 @@ import java.io.File
  * next app start.
  */
 object CrashLogStore {
+    private const val TAG = "CrashLogStore"
     private const val DIR_NAME = "pending_logs"
     private const val FILE_EXTENSION = ".log"
     private const val MAX_PENDING_FILES = 20
@@ -42,7 +44,7 @@ object CrashLogStore {
             file.writeText(error)
             file
         } catch (e: Exception) {
-            e.printStackTrace()
+            Log.e(TAG, "failed to save crash log", e)
             null
         }
     }
@@ -54,7 +56,7 @@ object CrashLogStore {
             try {
                 PendingLog(file, parsed.type, parsed.time, file.readText())
             } catch (e: Exception) {
-                e.printStackTrace()
+                Log.e(TAG, "failed to read pending crash log", e)
                 null
             }
         }


dogi:
**Review — rating: 86/100**

Replaces two `e.printStackTrace()` calls with tagged `Log.e`, and adds the `TAG` constant.

Small, but pointed at the right file. `printStackTrace()` writes to `System.err`, which on Android lands in logcat untagged and at no particular priority — so it can't be filtered by tag, doesn't register as an error, and is easy to miss entirely. Tagging with `CrashLogStore` and using `Log.e` makes both failures greppable and correctly prioritised.

The irony is the point: this is the component whose whole job is capturing failures for later upload, and it was swallowing its own I/O errors into the least visible channel available. If `save` can't write (full disk, permissions) or `readPendingLogs` can't parse a file, that's exactly the signal you want when investigating why crash reports stopped arriving.

Both messages are specific enough to tell the two paths apart ("failed to save crash log" vs "failed to read pending crash log"), and the exception is passed as the third argument so the stack trace is preserved rather than flattened into the message.

**Nit:** no tests, which is fine here — asserting on `Log` output would need Robolectric and a shadow, and would be testing the framework rather than this code.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17129, opened 2026-09-10
login: smoother become member toast handling (fixes #17121)

### LINKED ISSUE #17121: Remove MainApplication.context from BecomeMemberActivity's toast

`ui/user/BecomeMemberActivity.kt` reaches for the global application context from inside an Activity:

- `:18` — `import org.ole.planet.myplanet.MainApplication`
- `:127` — `Utilities.toast(MainApplication.context, result.second)`

Those are the only two occurrences of `MainApplication` in the file. An Activity already **is** a `Context`, so the global is pure indirection here.

### Do

1. At :127, pass `this` instead: `Utilities.toast(this, result.second)`.
2. Delete the now-unused `import org.ole.planet.myplanet.MainApplication` at :18.

### Constraints

- Two lines. Do not refactor the surrounding coroutine/callback, the `result` handling, or anything else in the file.
- Use `this` (the Activity), not `applicationContext` and not `baseContext` — the point is to stop reaching for an application-scoped context in an Activity.
- If the toast call sits inside a lambda where `this` is not the Activity, qualify it (`this@BecomeMemberActivity`) rather than reintroducing a stored context field.
- Do not touch any other `Utilities.toast` call site in the codebase; this issue is scoped to this one file.

### Tests

There is likely no test covering this line, and none is required — a toast argument change of this size is verified by the build. Do not add a Robolectric test just to assert a toast fired. If `BecomeMemberActivityTest` exists, keep it passing.

### Out of scope

dogi:
Replaced global MainApplication.context with explicit Activity context reference (this@BecomeMemberActivity) in BecomeMemberActivity.kt and removed unused MainApplication import.

Fixes #17121

---
*PR created automatically by Jules for task [12638011418949867802](https://jules.google.com/task/12638011418949867802) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/ui/user/BecomeMemberActivity.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/user/BecomeMemberActivity.kt
@@ -15,7 +15,6 @@ import kotlinx.coroutines.Job
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
-import org.ole.planet.myplanet.MainApplication
 import org.ole.planet.myplanet.R
 import org.ole.planet.myplanet.base.BaseActivity
 import org.ole.planet.myplanet.callback.OnChangedListener
@@ -124,7 +123,7 @@ class BecomeMemberActivity : BaseActivity() {
                     startUpload("becomeMember", userName, securityCallback)
 
                     if (result.second == getString(R.string.not_connect_to_planet_created_user_offline)) {
-                        Utilities.toast(MainApplication.context, result.second)
+                        Utilities.toast(this@BecomeMemberActivity, result.second)
                         securityCallback.onChanged()
                     }
                     Utilities.toast(this@BecomeMemberActivity, result.second)


dogi:
**Review — rating: 85/100**

Swaps `Utilities.toast(MainApplication.context, ...)` for `toast(this@BecomeMemberActivity, ...)`, dropping another `MainApplication.context` global and the now-unused import.

The activity context is the more correct one for a `Toast` — it picks up the activity's configuration (locale, which matters in an app shipping five translations) rather than the application's. And it's locally consistent: the very next statement already used `this@BecomeMemberActivity`, so master was passing two different contexts to two adjacent toasts of the same string for no reason.

**Nit — you're standing right next to a real oddity.** After this change the block reads:

```kotlin
if (result.second == getString(R.string.not_connect_to_planet_created_user_offline)) {
    Utilities.toast(this@BecomeMemberActivity, result.second)
    securityCallback.onChanged()
}
Utilities.toast(this@BecomeMemberActivity, result.second)
```

When the condition holds, the same message is toasted **twice** — and Android queues them, so the user sees it, then sees it again. That's pre-existing rather than introduced here (master had the same double-toast, just via two different contexts), but the duplication is much more visible now that both calls are identical, and this PR is already editing the exact lines. Worth either hoisting the toast above the `if` or dropping the one inside it.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17131, opened 2026-09-10
courses: smoother submissions repository pending survey querying (fixes #17124)

### LINKED ISSUE #17124: Stop hydrating full submission rows in getUniquePendingSurveys

`repository/SubmissionsRepositoryImpl.kt:100-122` — `getUniquePendingSurveys` hydrates **every** candidate before throwing most of them away:

```kotlin
override suspend fun getUniquePendingSurveys(userId: String?): List<Submission> {
    if (userId == null) return emptyList()

    val pendingSurveys = hydrateSubmissions(submissionDao.getUniquePendingSurveyCandidates(userId))   // :103
    if (pendingSurveys.isEmpty()) return emptyList()

    val examIds = pendingSurveys.mapNotNullTo(LinkedHashSet()) { it.examIdFromParentId() }.toList()   // :109
    if (examIds.isEmpty()) return emptyList()

    val exams = getExamsByIds(examIds)                    // :114
    val validExamIds = exams.map { it.id }.toSet()        // :115

    val uniqueSurveys = linkedMapOf<String, Submission>() // :117
    pendingSurveys.forEach { submission ->
        val examId = submission.examIdFromParentId()
        if (examId != null && validExamIds.contains(examId) && !uniqueSurveys.containsKey(examId)) {
            uniqueSurveys[examId] = submission
        }
    }
    …
```

`hydrateSubmissions` issues `answerDao.getBySubmissionIds` and attaches every answer to every candidate — and then the dedupe keeps **one submission per exam** and drops the rest. The bell/dashboard consumer only ever sees the survivors, so the answers attached to the dropped rows are pure waste.

### Do

Reorder the existing steps — dedupe first, hydrate only the survivors:

1. Read the raw candidates: `val candidates = submissionDao.getUniquePendingSurveyCandidates(userId)`, and keep the empty early return.
2. Derive `examIds` from the **raw** candidates with the same `mapNotNullTo(LinkedHashSet()) { it.examIdFromParentId() }.toList()`, keep the empty early return.
3. Validate with `getExamsByIds(examIds)` → `validExamIds` exactly as now.
4. Do the same first-wins dedupe into `linkedMapOf<String, Submission>()` over the raw candidates, preserving the `examId != null && validExamIds.contains(examId) && !containsKey(examId)` condition.
5. Call `hydrateSubmissions(...)` **once, on the surviving values only**, and return that.

No DAO change is needed. The backlog note proposed a new `GROUP BY parentId` count query and a nonzero-pending filter — **do not do that**: the candidates returned by `getUniquePendingSurveyCandidates` are already pending, so it buys nothing and adds a query.

### Constraints

- The returned list must be **identical** to today's: same submissions, same order (`LinkedHashSet` for the exam ids and `linkedMapOf` for the survivors preserve candidate order — keep both), and each returned submission must still be hydrated with its answers.
- `examIdFromParentId()` must be called on the raw candidate rows; confirm it reads only fields the DAO already populates (parent id) and does **not** depend on anything `hydrateSubmissions` attaches. If it does depend on hydration, stop and say so in the PR instead of shipping a reorder that changes results.
- Keep both early returns and their exact conditions.
- Do not change `hydrateSubmissions`, `getExamsByIds`, `getUniquePendingSurveyCandidates`, `getPendingSurveys`, or any other method.

### Tests

Extend `app/src/test/java/org/ole/planet/myplanet/repository/SubmissionsRepositoryImplTest.kt` (MockK, per the project convention) and keep every existing case passing. Cover:

- three candidates across two exams return one submission per exam, in candidate order, each with its answers attached,
- a candidate whose exam id is not in `examDao.getByIds` is dropped,
- **the hydration is narrowed**: assert `answerDao.getBySubmissionIds` is called with only the surviving submission ids, not all candidate ids. That assertion is the point of the change.

### Out of scope

Note that PR #17062 also edits this file, inserting `getSubmissionProjections` around old line 89 — eight lines above this function. Keep your diff inside `getUniquePendingSurveys` (:100-122) so the two do not collide.

dogi:
Reorder execution steps in SubmissionsRepositoryImpl.getUniquePendingSurveys so candidate deduplication and exam validation occur on raw submission candidates before calling hydrateSubmissions on survivors. Extend unit tests in SubmissionsRepositoryImplTest to verify narrowing of hydration and deduplication logic.

Fixes #17124

---
*PR created automatically by Jules for task [15777949482453275078](https://jules.google.com/task/15777949482453275078) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/repository/SubmissionsRepositoryImpl.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/SubmissionsRepositoryImpl.kt
@@ -100,13 +100,12 @@ class SubmissionsRepositoryImpl @Inject internal constructor(
     override suspend fun getUniquePendingSurveys(userId: String?): List<Submission> {
         if (userId == null) return emptyList()
 
-        val pendingSurveys = hydrateSubmissions(submissionDao.getUniquePendingSurveyCandidates(userId))
-
-        if (pendingSurveys.isEmpty()) {
+        val candidates = submissionDao.getUniquePendingSurveyCandidates(userId)
+        if (candidates.isEmpty()) {
             return emptyList()
         }
 
-        val examIds = pendingSurveys.mapNotNullTo(LinkedHashSet()) { it.examIdFromParentId() }.toList()
+        val examIds = candidates.mapNotNullTo(LinkedHashSet()) { it.examIdFromParentId() }.toList()
         if (examIds.isEmpty()) {
             return emptyList()
         }
@@ -115,13 +114,13 @@ class SubmissionsRepositoryImpl @Inject internal constructor(
         val validExamIds = exams.map { it.id }.toSet()
 
         val uniqueSurveys = linkedMapOf<String, Submission>()
-        pendingSurveys.forEach { submission ->
+        candidates.forEach { submission ->
             val examId = submission.examIdFromParentId()
             if (examId != null && validExamIds.contains(examId) && !uniqueSurveys.containsKey(examId)) {
                 uniqueSurveys[examId] = submission
             }
         }
-        return uniqueSurveys.values.toList()
+        return hydrateSubmissions(uniqueSurveys.values.toList())
     }
 
     override suspend fun getSurveyTitlesFromSubmissions(
--- a/app/src/test/java/org/ole/planet/myplanet/repository/SubmissionsRepositoryImplTest.kt
+++ b/app/src/test/java/org/ole/planet/myplanet/repository/SubmissionsRepositoryImplTest.kt
@@ -209,6 +209,74 @@ class SubmissionsRepositoryImplTest {
     }
 
     @Test
+    fun `getUniquePendingSurveys deduplicates candidates across exams returning one per exam in candidate order with answers`() = runTest {
+        val candidate1 = Submission(id = "sub1", parentId = "exam1@course1")
+        val candidate2 = Submission(id = "sub2", parentId = "exam1@course2")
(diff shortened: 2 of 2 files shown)


dogi:
**Review — rating: 89/100**

Moves `hydrateSubmissions` from the front of `getUniquePendingSurveys` to the very end, so answers are fetched only for the submissions that survive deduplication.

The reordering is safe because hydration doesn't feed any of the logic it now runs after. `hydrateSubmissions` only sets `answers` and `membershipDoc`; the dedupe key comes from `examIdFromParentId()`, which reads `parentId` — untouched by hydration. So `candidates` and the old `pendingSurveys` are interchangeable for every decision in between, and `hydrateSubmissions` preserves list order and size, which keeps both the `isEmpty()` short-circuit and the `linkedMapOf` insertion ordering identical.

The win scales with how duplicated the candidate set is, and this method exists precisely because it's duplicated — a user with many pending submissions against a handful of surveys previously hydrated all of them and then discarded most. Only the deduped survivors now reach `answerDao.getBySubmissionIds`.

Worth noting a second-order benefit: this shrinks the `IN (:submissionIds)` list that `hydrateSubmissions` builds, which is the same list #17147 in this batch is chunking to stay under SQLite's 999-variable ceiling. The two changes pull in the same direction — this one reduces how often the ceiling is approached at all.

68 lines of new tests, which this method needed given the interaction between candidate filtering, `validExamIds` and first-wins dedupe.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17132, opened 2026-09-10
all: smoother myplanet usage stats caching (fixes #17123)

### LINKED ISSUE #17123: Hoist the per-row invariants out of the addStats usage-stats loop

`model/MyPlanet.kt:65-81` — `getTabletUsages` already takes a `now: Long = System.currentTimeMillis()` and loops the query results:

```kotlin
val queryUsageStats = mUsageStatsManager.queryUsageStats(UsageStatsManager.INTERVAL_DAILY, cal.timeInMillis, now)
if (queryUsageStats != null) {
    for (s in queryUsageStats) {
        addStats(s, arr, context)      // :77
    }
}
```

and `addStats` (:83-99) recomputes, per row:

- `context.packageName` (:84, in the guard)
- `NetworkUtils.getCustomDeviceName(context)` (:94 — a preferences read)
- `NetworkUtils.getDeviceName()` (:95)
- `Date().time` (:96)

**Be honest about the win in the PR description:** everything except the `packageName` comparison sits *inside* `if (s.packageName == context.packageName)`, so only the app's own row pays those costs. The real improvement here is the **consistent timestamp** — today every emitted row could carry a slightly different `time`, and `getTabletUsages` already has `now` for exactly this purpose. Do not describe this as a large allocation win.

### Do

1. In `getTabletUsages`, before the loop, compute once:
   - `val packageName = context.packageName`
   - `val customDeviceName = NetworkUtils.getCustomDeviceName(context)`
   - `val deviceName = NetworkUtils.getDeviceName()`
2. Change `addStats`'s signature to take those three plus `now`, e.g. `private fun addStats(s: UsageStats, arr: JsonArray, context: Context, packageName: String, customDeviceName: String, deviceName: String, now: Long)`. Keep `context` — it is still needed for the `VersionUtils` calls.
3. In `addStats`, use the passed values for the guard (:84), `"customDeviceName"` (:94) and `"deviceName"` (:95), and use `now` for `"time"` (:96) instead of `Date().time`.
4. Drop the `java.util.Date` import **only if** nothing else in the file uses it — `:60` (`postJSON.addProperty("time", Date().time)` in a different function) does, so it almost certainly stays. Check before touching imports.

### Constraints

- **Do not touch the `VersionUtils.getVersionCode(context)` / `getVersionName(context)` calls at :91-92.** Another open task caches those at their source in `VersionUtils`; caching them here too would collide.
- Keep every JSON key, its order, and every value expression identical otherwise: `lastTimeUsed`, the deliberate `firstTimeUsed` → `s.lastTimeStamp` mapping at :87 (**yes, that reads `lastTimeStamp` while the key says `firstTimeUsed` — leave it exactly as it is**, this issue is not the place to change what gets uploaded), `totalForegroundTime`, the `totalUsed` clamp, `addDocumentOrigin()`.
- Keep `addStats` private, keep the guard's semantics (only the app's own package is emitted), and keep `getTabletUsages`'s signature and default argument.
- Do not touch `getMyPlanetActivities`/the function at :60, `spm.getLastUsageUploaded()`, or the `Calendar` setup.

### Tests

Add or extend a test for `MyPlanet.getTabletUsages` (`app/src/test/java/org/ole/planet/myplanet/model/`), keeping existing cases passing. Cover: with two matching-package `UsageStats` rows, every emitted object carries the **same** `time`, equal to the `now` passed in; and rows whose package differs are still skipped. Pass an explicit `now` so the assertion is deterministic — do not assert on the system clock.

### Out of scope

Do not touch `AppDatabase` or any Room schema. **Do not add or change any log line, and never log the device name or any URL** — myPlanet server URLs embed `satellite:<pin>@host` credentials and release builds do not strip logs.

dogi:
This PR hoists per-row invariant computations out of the `addStats` usage-stats loop in `MyPlanet.getTabletUsages`.

### Changes Made:
- In `MyPlanet.getTabletUsages`, pre-computes `packageName`, `customDeviceName`, and `deviceName` once before iterating through `queryUsageStats`.
- Passes `packageName`, `customDeviceName`, `deviceName`, and `now` into `addStats`.
- Replaces per-row calls to `context.packageName`, `NetworkUtils.getCustomDeviceName(context)`, `NetworkUtils.getDeviceName()`, and `Date().time` inside `addStats` with the pre-computed arguments.
- Updated `MyPlanetTest.kt` to verify that all emitted usage records carry identical timestamps matching `now` and that non-matching package rows are correctly filtered.

### Performance & Consistency Note:
Except for the package name check, per-row calls inside `addStats` were guarded by `if (s.packageName == context.packageName)`, so only the app's own row paid those costs. The primary gain here is ensuring a consistent timestamp (`now`) across all emitted usage stats rows rather than potentially distinct system timestamps per row.

Fixes #17123

---
*PR created automatically by Jules for task [2478884537605841352](https://jules.google.com/task/2478884537605841352) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/model/MyPlanet.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/model/MyPlanet.kt
@@ -73,15 +73,26 @@ class MyPlanet : Serializable {
             val mUsageStatsManager = context.getSystemService(Context.USAGE_STATS_SERVICE) as UsageStatsManager
             val queryUsageStats = mUsageStatsManager.queryUsageStats(UsageStatsManager.INTERVAL_DAILY, cal.timeInMillis, now)
             if (queryUsageStats != null) {
+                val packageName = context.packageName
+                val customDeviceName = NetworkUtils.getCustomDeviceName(context)
+                val deviceName = NetworkUtils.getDeviceName()
                 for (s in queryUsageStats) {
-                    addStats(s, arr, context)
+                    addStats(s, arr, context, packageName, customDeviceName, deviceName, now)
                 }
             }
             return arr
         }
 
-        private fun addStats(s: UsageStats, arr: JsonArray, context: Context) {
-            if (s.packageName == context.packageName) {
+        private fun addStats(
+            s: UsageStats,
+            arr: JsonArray,
+            context: Context,
+            packageName: String,
+            customDeviceName: String,
+            deviceName: String,
+            now: Long
+        ) {
+            if (s.packageName == packageName) {
                 val `object` = JsonObject()
                 `object`.addProperty("lastTimeUsed", if (s.lastTimeUsed > 0) s.lastTimeUsed else 0)
                 `object`.addProperty("firstTimeUsed", if (s.firstTimeStamp > 0) s.lastTimeStamp else 0)
@@ -91,9 +102,9 @@ class MyPlanet : Serializable {
                 `object`.addProperty("version", VersionUtils.getVersionCode(context))
                 `object`.addProperty("versionName", VersionUtils.getVersionName(context))
                 `object`.addDocumentOrigin()
-                `object`.addProperty("customDeviceName", NetworkUtils.getCustomDeviceName(context))
-                `object`.addProperty("deviceName", NetworkUtils.getDeviceName())
-                `object`.addProperty("time", Date().time)
+                `object`.addProperty("customDeviceName", customDeviceName)
+                `object`.addProperty("deviceName", deviceName)
+                `object`.addProperty("time", now)
                 arr.add(`object`)
             }
         }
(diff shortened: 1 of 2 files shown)


dogi:
**Review — rating: 84/100**

Hoists `context.packageName`, `NetworkUtils.getCustomDeviceName(context)` and `NetworkUtils.getDeviceName()` above the `queryUsageStats` loop and passes them into `addStats`. All three are loop-invariant, so this is straightforwardly correct, and `getCustomDeviceName` in particular reads `SharedPreferences` — worth doing once rather than per row.

**One behavioural change worth naming**, because it isn't purely a hoist: `Date().time` per row becomes the single `now` captured before the query. Rows in a batch now share one timestamp instead of each carrying its own. I'd argue that's an improvement — `now` is already the upper bound passed to `queryUsageStats(INTERVAL_DAILY, cal.timeInMillis, now)`, so the serialized `time` is now consistent with the window the stats describe, rather than "whenever serialization happened to reach this row". But it does mean uploaded usage-stat documents in one batch become indistinguishable by `time`, so it's worth a line in the description rather than reading as an incidental cleanup.

**Nit — the two most expensive per-row calls are still in the loop.** `VersionUtils.getVersionCode(context)` and `getVersionName(context)` remain inside `addStats`, and those are `PackageManager.getPackageInfo` Binder round-trips to `system_server` — considerably pricier than the three things that got hoisted. They're just as invariant. #17079 in this queue memoizes both inside `VersionUtils`, which makes the point moot once it lands, but if that's the reasoning it's worth saying so; otherwise they belong in the same hoist as everything else.

48 lines of new `MyPlanetTest` coverage for a serializer that had none.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17133, opened 2026-09-10
all: smoother free space recursive delete working (fixes #17126)

### LINKED ISSUE #17126: Stop the redundant exists() syscall per file in FreeSpaceWorker

`services/FreeSpaceWorker.kt:63-83`:

```kotlin
private suspend fun deleteRecursive(fileOrDirectory: File) {
    fileOrDirectory.walkBottomUp().forEach { file ->
        if (isStopped) return

        if (file.exists()) {              // :67
            val length = file.length()
            if (file.delete()) {
                deletedFiles++
                freedBytes += length
                if (deletedFiles % 10 == 0) {
                    setProgress(workDataOf("deletedFiles" to deletedFiles, "freedBytes" to freedBytes))
                }
            }
        }
    }
}
```

`walkBottomUp()` only yields entries the directory walk found, and the code already relies on `delete()`'s boolean return. So `exists()` is an extra `stat` syscall per node whose answer `delete()` gives anyway.

It is also accounting-neutral: for a file that has since disappeared, `length()` returns `0` and `delete()` returns `false`, so `deletedFiles`/`freedBytes` are unchanged either way.

### Do

Drop the `if (file.exists())` wrapper, keeping everything inside it exactly as-is and at the same nesting:

```kotlin
fileOrDirectory.walkBottomUp().forEach { file ->
    if (isStopped) return

    val length = file.length()
    if (file.delete()) {
        deletedFiles++
        freedBytes += length
        if (deletedFiles % 10 == 0) {
            setProgress(workDataOf("deletedFiles" to deletedFiles, "freedBytes" to freedBytes))
        }
    }
}
```

### Constraints

- **Keep `length()` before `delete()`** — reading the size after deletion would always yield 0 and silently zero out `freedBytes`.
- Keep the `isStopped` early `return` as the first statement in the lambda, and keep it a `return` (non-local, aborting the whole walk), not `return@forEach`.
- Keep `walkBottomUp()` — children must be deleted before their parent directory.
- Keep the every-10-files `setProgress` and both counter increments inside the `delete()` success branch, with the same `workDataOf` keys `"deletedFiles"` and `"freedBytes"`.
- Do not change `deleteRecursive`'s signature or make it non-suspend, and do not touch `doWork`, the `Result.failure()` path, `TAG`, `CV_DIR_NAME` or `MARK_BATCH_SIZE`.

### Tests

If `FreeSpaceWorkerTest` exists, keep it passing. Worthwhile coverage: given a temp directory with two files and a nested subdirectory, `deleteRecursive` removes everything, `deletedFiles` counts every deleted node and `freedBytes` equals the summed file sizes — and a file deleted out from under the walk does not corrupt the accounting.

### Out of scope

Do not touch `AppDatabase` or any Room schema. The existing `Log.e(TAG, "Error in FreeSpaceWorker", e)` at :58 stays exactly as it is — do not add new log lines, and never log a file path or URL.

dogi:
Removes the redundant exists() check in FreeSpaceWorker.deleteRecursive to eliminate extra stat syscalls per node during free space worker cleanup.

Fixes #17126

---
*PR created automatically by Jules for task [2343735942768662849](https://jules.google.com/task/2343735942768662849) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/services/FreeSpaceWorker.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/services/FreeSpaceWorker.kt
@@ -64,19 +64,17 @@ class FreeSpaceWorker @AssistedInject constructor(
         fileOrDirectory.walkBottomUp().forEach { file ->
             if (isStopped) return
 
-            if (file.exists()) {
-                val length = file.length()
-                if (file.delete()) {
-                    deletedFiles++
-                    freedBytes += length
+            val length = file.length()
+            if (file.delete()) {
+                deletedFiles++
+                freedBytes += length
 
-                    // Report progress every 10 files or so to avoid spamming updates
-                    if (deletedFiles % 10 == 0) {
-                         setProgress(workDataOf(
-                             "deletedFiles" to deletedFiles,
-                             "freedBytes" to freedBytes
-                         ))
-                    }
+                // Report progress every 10 files or so to avoid spamming updates
+                if (deletedFiles % 10 == 0) {
+                     setProgress(workDataOf(
+                         "deletedFiles" to deletedFiles,
+                         "freedBytes" to freedBytes
+                     ))
                 }
             }
         }
--- a/app/src/test/java/org/ole/planet/myplanet/services/FreeSpaceWorkerTest.kt
+++ b/app/src/test/java/org/ole/planet/myplanet/services/FreeSpaceWorkerTest.kt
@@ -139,4 +139,52 @@ class FreeSpaceWorkerTest {
 
         assertTrue(result is Result.Failure)
     }
+
+    @Test
+    fun `doWork removes nested directory structure and accurately tracks deleted files and freed bytes`() = runTest(testDispatcher) {
+        val resDir = File(oleDir, "res1").apply { mkdirs() }
+        val file1 = File(resDir, "file1.txt").apply { writeText("hello") }
+        val subDir = File(resDir, "subdir").apply { mkdirs() }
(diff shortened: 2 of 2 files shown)


dogi:
**Review — rating: 86/100**

Drops the `if (file.exists())` wrapper around the delete loop, relying on `length()` and `delete()` to handle a missing file themselves.

Equivalent, and I checked the degenerate case rather than assuming: for a file that doesn't exist, `File.length()` returns `0L` and `File.delete()` returns `false`, so the new code adds nothing to `freedBytes`, doesn't increment `deletedFiles`, and doesn't fire a progress update — exactly what the old `exists()` guard achieved by skipping. The only difference is one fewer `stat` per entry.

That's worth having here specifically because of where it sits: `walkBottomUp()` over the app's storage during a free-space sweep can touch thousands of entries, and the guard was redundant by construction — the walk only yields paths the filesystem just enumerated, so `exists()` was re-asking a question that had been answered a moment earlier. Removing it also flattens a nesting level, which makes the `isStopped` early return and the every-10-files progress update easier to follow.

The `isStopped` check stays at the top of the loop, so cancellation responsiveness is unchanged.

**Nit — pre-existing, not introduced.** `walkBottomUp` yields directories too, and `File.length()` on a directory returns a filesystem-dependent value (commonly 4096) rather than 0, so each successfully deleted directory inflates `freedBytes`. Master had the same behaviour inside its `exists()` guard, so this PR doesn't change it — but if the reported "freed" figure is ever user-visible, that's where the drift comes from.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17134, opened 2026-09-10
life: smoother health users payload view holding (fixes #17127)

### LINKED ISSUE #17127: Hoist the per-bind lookups out of HealthUsersAdapter's payload path

`ui/health/HealthUsersAdapter.kt:76-92`:

```kotlin
override fun onBindViewHolder(holder: ViewHolder, position: Int, payloads: MutableList<Any>) {
    val diffs = payloads.filterIsInstance<List<*>>().flatten()      // :77 — two intermediate lists
    if (diffs.isEmpty()) {
    } else {
        val user = getItem(position)
        if ("name" in diffs) { holder.bindName(user) }               // :82
        if ("userImage" in diffs) { holder.bindImage(user) }         // :85
        if ("joinDate" in diffs) { holder.bindDate(user) }           // :88
    }
}
```

`filterIsInstance<List<*>>()` allocates one list and `flatten()` a second, on **every partial bind**, purely to answer three string membership questions — and each `in diffs` is itself a linear scan.

### Do

Iterate `payloads` once, collecting the three flags, then branch on whether anything was found:

```kotlin
override fun onBindViewHolder(holder: ViewHolder, position: Int, payloads: MutableList<Any>) {
    var name = false
    var userImage = false
    var joinDate = false
    for (payload in payloads) {
        if (payload is List<*>) {
            for (key in payload) {
                when (key) {
                    "name" -> name = true
                    "userImage" -> userImage = true
                    "joinDate" -> joinDate = true
                }
            }
        }
    }

    if (!name && !userImage && !joinDate) {
    } else {
        val user = getItem(position)
        if (name) holder.bindName(user)
        if (userImage) holder.bindImage(user)
        if (joinDate) holder.bindDate(user)
    }
}
```

### Constraints — one subtlety, do not get it wrong

- The current fallback triggers when the **flattened** list is empty, which includes the case where `payloads` contains only non-`List` elements, **and** the case where it contains `List`s whose elements are none of the three keys. Reproduce that: fall back to `super.onBindViewHolder` when none of the three flags is set — not merely when no `List<*>` payload was present. The `!name && !userImage && !joinDate` condition above does this; if you write it differently, check it against those two cases.
- Keep the bind order: name, then image, then date.
- Keep `getItem(position)` called **only** in the else branch (it must not run on the fallback path).
- Only these three keys matter; do not add handling for other payload values, and do not change what the DiffUtil callback emits.
- Do not touch the other `onBindViewHolder(holder, position)` overload at :72-74, the ViewHolder, or `bindName`/`bindImage`/`bindDate`.

### Tests

Add to `app/src/test/java/org/ole/planet/myplanet/ui/health/HealthUsersAdapterTest.kt` if it exists, otherwise create it, and keep existing cases passing. Cover:

- a payload of `listOf("name")` calls only `bindName`,
- a payload carrying two keys calls both binds, in order,
- a payload of `listOf("somethingElse")` falls back to a full bind,
- a non-`List` payload element falls back to a full bind,
- duplicate keys across two payload entries still bind once each.

### Out of scope

dogi:
Optimized HealthUsersAdapter.onBindViewHolder partial payload processing by removing intermediate list allocations and linear lookups, replacing them with a single pass through payloads. Added unit tests in HealthUsersAdapterTest to verify single key, multi-key ordering, duplicate keys, unrecognized payload lists, and non-list fallback behavior.

Fixes #17127

---
*PR created automatically by Jules for task [1956609268085775098](https://jules.google.com/task/1956609268085775098) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/ui/health/HealthUsersAdapter.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/health/HealthUsersAdapter.kt
@@ -74,20 +74,28 @@ class HealthUsersAdapter(private val clickListener: ((UserEntity) -> Unit)? = nu
     }
 
     override fun onBindViewHolder(holder: ViewHolder, position: Int, payloads: MutableList<Any>) {
-        val diffs = payloads.filterIsInstance<List<*>>().flatten()
-        if (diffs.isEmpty()) {
+        var name = false
+        var userImage = false
+        var joinDate = false
+        for (payload in payloads) {
+            if (payload is List<*>) {
+                for (key in payload) {
+                    when (key) {
+                        "name" -> name = true
+                        "userImage" -> userImage = true
+                        "joinDate" -> joinDate = true
+                    }
+                }
+            }
+        }
+
+        if (!name && !userImage && !joinDate) {
             super.onBindViewHolder(holder, position, payloads)
         } else {
             val user = getItem(position)
-            if ("name" in diffs) {
-                holder.bindName(user)
-            }
-            if ("userImage" in diffs) {
-                holder.bindImage(user)
-            }
-            if ("joinDate" in diffs) {
-                holder.bindDate(user)
-            }
+            if (name) holder.bindName(user)
+            if (userImage) holder.bindImage(user)
+            if (joinDate) holder.bindDate(user)
         }
     }
 }
--- a/app/src/test/java/org/ole/planet/myplanet/ui/health/HealthUsersAdapterTest.kt
+++ b/app/src/test/java/org/ole/planet/myplanet/ui/health/HealthUsersAdapterTest.kt
@@ -1,0 +1,105 @@
+package org.ole.planet.myplanet.ui.health
+
+import android.content.Context
+import android.widget.FrameLayout
+import androidx.test.core.app.ApplicationProvider
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import io.mockk.clearMocks
+import io.mockk.spyk
+import io.mockk.verify
+import io.mockk.verifyOrder
+import org.junit.Before
+import org.junit.Test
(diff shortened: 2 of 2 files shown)


dogi:
**Review — rating: 87/100**

Replaces `payloads.filterIsInstance<List<*>>().flatten()` plus three `in diffs` scans with one pass setting three booleans. Filed as an allocation trim, but it also closes a latent staleness bug, and that's the stronger reason to take it.

**The bug.** Master decided between partial and full rebind on `diffs.isEmpty()`. Consider a payload list carrying a key outside the known three — say `listOf("email")`. `diffs` is non-empty, so the `else` branch runs, but none of `"name"`, `"userImage"`, `"joinDate"` matches, so **no bind method is called at all** and `super.onBindViewHolder` is never reached either. The recycled holder keeps the previous row's content. The new condition — fall back to `super` unless one of the three flags was actually set — turns that case into a correct full rebind. Nothing emits an unrecognised key today, so this is latent rather than live, but it's the difference between "unknown payload ⇒ safe full rebind" and "unknown payload ⇒ silently wrong row".

**The allocation side** is real too: `filterIsInstance` and `flatten` each build a list on every payload bind, and the three `in` checks are linear scans over the flattened result. One loop with three booleans removes both, in a code path that runs per item on every diff-driven update.

The 105 lines of new `HealthUsersAdapterTest` are what make this reviewable — payload dispatch is easy to get subtly wrong and had no coverage before.

**Nit:** the imperative version is noticeably more verbose than what it replaces. Worth it on an adapter bind path, but a short comment noting *why* it's hand-rolled would stop someone "tidying" it back into `filterIsInstance().flatten()` later.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17139, opened 2026-09-10
all: smoother activities feedback repositories time providing (fixes #17130)

### LINKED ISSUE #17130: Route the remaining wall-clock reads in ActivitiesRepositoryImpl and FeedbackRepositoryImpl through TimeProvider

`utils/TimeProvider.kt` exists precisely for this, and says so:

> Injectable source of wall-clock time so time-of-day / expiry / scheduling decisions are testable and don't read `System.currentTimeMillis` directly. Mirrors `DispatcherProvider`. Use `now()` for timestamps and time-based decisions persisted across restarts.

**`repository/ActivitiesRepositoryImpl.kt` already injects it and already uses it — inconsistently.** The constructor takes `private val timeProvider: TimeProvider` (`:52`, imported at `:41`) and `:233` writes `time = timeProvider.now()`, yet five sibling sites still allocate a `Date` and read its millis:

- `:100` — `time = Date().time`
- `:128` — `loginTime = Date().time`
- `:135` — `offlineActivityDao.updateLogoutTime(activity.id, Date().time)`
- `:164` — `time = Date().time`
- `:256` — `time = Date().time`

So one timestamp in this file is fakeable in tests and five are not. That inconsistency — not the allocation — is the defect worth fixing.

**`repository/FeedbackRepositoryImpl.kt`** does the same but has no injection yet (`@Inject constructor(` at `:18`, `java.util.Date` imported at `:6`):

- `:54` — `val timestamp = Date().time`
- `:102` — `addProperty("time", Date().time.toString())`

### Do

1. In `ActivitiesRepositoryImpl`, replace `Date().time` with `timeProvider.now()` at `:100`, `:128`, `:135`, `:164` and `:256`. Drop the `java.util.Date` import at `:9` **only if** nothing else in the file uses `Date` — check first.
2. In `FeedbackRepositoryImpl`, add `private val timeProvider: TimeProvider` to the constructor, replace both sites with `timeProvider.now()` (keeping `.toString()` at `:102`), and drop the `java.util.Date` import if it becomes unused.

### Constraints

- **The values written must stay millisecond epoch longs.** `timeProvider.now()` returns `Long`, same as `Date().time` — do not introduce formatting, `Instant`, or a different unit, and keep `:102`'s `.toString()` so the JSON field stays a string exactly as today.
- Do not add a `TimeProvider` parameter with a default value, and do not construct `SystemTimeProvider` inline — it is Hilt-provided (`TimeModule`); inject it.
- Do not touch `:233`, which is already correct.
- Do not convert any of these to `SystemClock.elapsedRealtime()`. Per `TimeProvider`'s own doc, elapsed-realtime is for measuring durations inside one process; every site here is a persisted timestamp.
- Do not change any DAO signature, any column, or the shape of the JSON at `:102`.
- Do not touch other repositories in this pass, even if they have the same pattern.

### Tests

Extend `app/src/test/java/org/ole/planet/myplanet/repository/ActivitiesRepositoryImplTest.kt` and `FeedbackRepositoryImplTest.kt` (MockK, per the project convention), keeping every existing case passing. The point of the change is that these timestamps become assertable, so: stub `timeProvider.now()` to a fixed value and assert the persisted row / JSON carries exactly that value, for at least one site in each file (including the `updateLogoutTime` call at `:135`, where the timestamp is an argument rather than a field). `FeedbackRepositoryImplTest` will need the new constructor argument.

### Out of scope

Note that PR #17046 also edits `ActivitiesRepositoryImpl`, but only inserts `getMemberVisitStats` around `:62-68`, well clear of these lines — keep your diff to the five sites and the import so the two do not collide.

dogi:
Replaced wall-clock `Date().time` calls in `ActivitiesRepositoryImpl` and `FeedbackRepositoryImpl` with `timeProvider.now()`. Added `TimeProvider` injection to `FeedbackRepositoryImpl` and removed unused `java.util.Date` imports. Extended unit tests for both repositories to verify the stubbed time values.

Fixes #17130

---
*PR created automatically by Jules for task [5317014135234012686](https://jules.google.com/task/5317014135234012686) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/repository/ActivitiesRepositoryImpl.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/ActivitiesRepositoryImpl.kt
@@ -6,7 +6,6 @@ import com.google.gson.JsonObject
 import dagger.Lazy
 import dagger.hilt.android.qualifiers.ApplicationContext
 import java.io.IOException
-import java.util.Date
 import java.util.UUID
 import javax.inject.Inject
 import kotlinx.coroutines.async
@@ -97,7 +96,7 @@ class ActivitiesRepositoryImpl @Inject constructor(
                 type = "visit"
                 this.title = title
                 this.courseId = courseId
-                time = Date().time
+                time = timeProvider.now()
                 this.user = userId
 
                 if (user != null) {
@@ -125,14 +124,14 @@ class ActivitiesRepositoryImpl @Inject constructor(
                 _rev = null
                 _id = null
                 description = "Member login on offline application"
-                loginTime = Date().time
+                loginTime = timeProvider.now()
             }
         )
     }
 
     override suspend fun logLogout(userName: String?) {
         offlineActivityDao.getLatestByType(UserSessionManager.KEY_LOGIN)?.let { activity ->
-            offlineActivityDao.updateLogoutTime(activity.id, Date().time)
+            offlineActivityDao.updateLogoutTime(activity.id, timeProvider.now())
         }
     }
 
@@ -161,7 +160,7 @@ class ActivitiesRepositoryImpl @Inject constructor(
                 this.type = type
                 this.title = title
                 this.resourceId = resourceId
-                time = Date().time
+                time = timeProvider.now()
             }
         )
     }
@@ -253,7 +252,7 @@ class ActivitiesRepositoryImpl @Inject constructor(
                 this.parentCode = parentCode
                 this.createdOn = createdOn
                 type = "sync"
-                time = Date().time
+                time = timeProvider.now()
             }
         )
     }
--- a/app/src/main/java/org/ole/planet/myplanet/repository/FeedbackRepositoryImpl.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/FeedbackRepositoryImpl.kt
@@ -3,7 +3,6 @@ package org.ole.planet.myplanet.repository
 import com.google.gson.Gson
 import com.google.gson.JsonArray
 import com.google.gson.JsonObject
-import java.util.Date
 import java.util.UUID
 import javax.inject.Inject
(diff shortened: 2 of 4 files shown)


dogi:
**Review — rating: 88/100**

Replaces six direct `Date().time` reads with `timeProvider.now()` — four in `ActivitiesRepositoryImpl` (visit logging, offline login, logout, resource activity, sync activity) and two in `FeedbackRepositoryImpl`, which gains the dependency.

Behaviourally a no-op in production (`SystemTimeProvider.now()` is `System.currentTimeMillis()`, which is what `Date().time` resolves to), and that's the point — the value is that these timestamps become controllable. `CLAUDE.md` is explicit that time and dispatchers should be injected so tests can substitute deterministic implementations, and these are exactly the sites where it pays: `logLogout` writing a logout time, `createAndSaveFeedback` stamping `openTime`, and the feedback message `time` property all had assertions that previously couldn't pin an exact value. The +28 lines of test changes across both repository test classes are that coverage landing.

`ActivitiesRepositoryImpl` already had `timeProvider` injected, so four of the six sites were inconsistent with their own class — only the `java.util.Date` import needed removing. `FeedbackRepositoryImpl` is the one taking a genuinely new dependency, and as a `@Singleton` with constructor injection that's free.

**Worth reading alongside #17143** in this same batch, which touches two *other* wall-clock reads (`CameraUtils`, `CalendarFragment`) and swaps them for `System.currentTimeMillis()` / `Date()` rather than routing them through `TimeProvider`. Neither PR is wrong — this one covers repositories where `TimeProvider` is already the convention, that one covers UI/util sites where it isn't wired — but it'd be worth the two agreeing on where the boundary sits, since a reader seeing both land will reasonably ask.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17141, opened 2026-09-10
all: smoother submissions repository counting (fixes #17135)

### LINKED ISSUE #17135: Collapse the redundant intermediate collections at three call sites

Three places build a throwaway list before doing the thing they actually wanted. Each is one or two lines; they are bundled here so they cost one review instead of three.

**1. `ui/resources/ResourcesAdapter.kt:167`** — filter then map, in `setLibraryList(libraryList: List<ResourceListModel?>, …)`:

```kotlin
val updatedList = libraryList.filterNotNull().map {
    it.copy(
        isOpened = openedResourceIds.contains(it.item.id),
        isLocallyOffline = locallyOfflineIds.contains(it.item.id)
    )
}
```

**2. `ui/teams/voices/TeamsVoicesFragment.kt:262` and `:273`** — a list built only to read `.size`:

```kotlin
showNoData(binding.tvNodata, realmNewsList?.filterNotNull()?.size ?: 0, "discussions")   // :262
showNoData(binding.tvNodata, it.filterNotNull().size, "discussions")                     // :273
```

**3. `repository/SubmissionsRepositoryImpl.kt:213`** — filter then map inside a `chunked(500)` loop:

```kotlin
val newSubmissions = chunk.filter { it !in existingUserIds }.map { userId ->
    Submission().apply { … }
}
```

Sizing, honestly: #1 is proportional to the resource list, #2 to the discussion list, #3 is capped at 500 and dominated by the DB insert that follows. None of these is a measurable win — they are altitude cleanups, so please don't describe them as performance fixes in the PR.

### Do

1. `ResourcesAdapter:167` — one `mapNotNull` pass: `libraryList.mapNotNull { it?.copy(isOpened = …, isLocallyOffline = …) }`, with the two `contains` arguments unchanged.
2. `TeamsVoicesFragment:262` — `realmNewsList?.count { it != null } ?: 0`. `:273` — `it.count { … != null }`.
   **Watch the shadowed `it`:** :273 sits inside `realmNewsList?.let { … }`, so the outer `it` is the list. Name the inner lambda parameter explicitly (e.g. `it.count { news -> news != null }`) rather than nesting two `it`s.
3. `SubmissionsRepositoryImpl:213` — one `mapNotNull` pass: `chunk.mapNotNull { userId -> if (userId in existingUserIds) null else Submission().apply { … } }`, with the `Submission().apply { … }` block copied over **verbatim**.

### Constraints

- **Leave the two `submitList(it.filterNotNull())` calls at `TeamsVoicesFragment:259` and `:267` exactly as they are** — those legitimately need a real list, and the `:267` one has a trailing lambda that must survive untouched.
- In `SubmissionsRepositoryImpl`, do not touch anything else in the `chunked(500)` loop: `getPendingByUsersAndParent` (:211), the `existingUserIds` set (:212), the `Submission` field assignments including `startTime = Date().time` and `lastUpdateTime = startTime` (:220-221), and the insert after the loop. In particular **do not** convert that `Date().time` to `System.currentTimeMillis()` or `TimeProvider` here — another issue covers wall-clock reads, and widening this diff invites a conflict.
- In `ResourcesAdapter`, keep `setLibraryList`'s signature (including the nullable element type and the `onComplete` parameter) and whatever follows the `updatedList` construction.
- Semantics must be identical: same elements, same order, same counts, for empty lists and all-null lists alike.

### Tests

Keep every existing test in the touched areas passing. Worth adding: a `ResourcesAdapter` case where `libraryList` contains a null element (it is dropped, and the surviving items carry the right `isOpened`/`isLocallyOffline`), and a `SubmissionsRepositoryImpl` case where some chunk members already have pending submissions (only the others get a `Submission`, in input order).

### Out of scope

Two related sites are deliberately **excluded** because open PR #17044 edits both files: `ui/health/MyHealthFragment.kt:339-342` (`getDisplayName`'s three intermediate collections) and `ui/health/HealthExaminationAdapter.kt:127`/`:131` (`value.toString() + ""`). Do not touch either file.

dogi:
Consolidates redundant intermediate collection allocations in ResourcesAdapter, TeamsVoicesFragment, and SubmissionsRepositoryImpl into single-pass operations and count functions.

Fixes #17135

---
*PR created automatically by Jules for task [9418384464688522180](https://jules.google.com/task/9418384464688522180) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/repository/SubmissionsRepositoryImpl.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/SubmissionsRepositoryImpl.kt
@@ -210,8 +210,8 @@ class SubmissionsRepositoryImpl @Inject internal constructor(
         userIds.chunked(500).forEach { chunk ->
             val existingSubmissions = submissionDao.getPendingByUsersAndParent(chunk, parentId)
             val existingUserIds = existingSubmissions.mapNotNull { it.userId }.toSet()
-            val newSubmissions = chunk.filter { it !in existingUserIds }.map { userId ->
-                Submission().apply {
+            val newSubmissions = chunk.mapNotNull { userId ->
+                if (userId in existingUserIds) null else Submission().apply {
                     id = UUID.randomUUID().toString()
                     this.userId = userId
                     this.parentId = parentId
--- a/app/src/main/java/org/ole/planet/myplanet/ui/resources/ResourcesAdapter.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/resources/ResourcesAdapter.kt
@@ -164,8 +164,8 @@ class ResourcesAdapter(
     }
 
     fun setLibraryList(libraryList: List<ResourceListModel?>, onComplete: (() -> Unit)? = null) {
-        val updatedList = libraryList.filterNotNull().map {
-            it.copy(
+        val updatedList = libraryList.mapNotNull {
+            it?.copy(
                 isOpened = openedResourceIds.contains(it.item.id),
                 isLocallyOffline = locallyOfflineIds.contains(it.item.id)
             )
--- a/app/src/main/java/org/ole/planet/myplanet/ui/teams/voices/TeamsVoicesFragment.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/teams/voices/TeamsVoicesFragment.kt
@@ -259,7 +259,7 @@ class TeamsVoicesFragment : BaseTeamFragment() {
             realmNewsList?.let { adapterNews?.submitList(it.filterNotNull()) }
             binding.rvDiscussion.adapter = adapterNews
             shouldScrollToTopNextUpdate = false
-            showNoData(binding.tvNodata, realmNewsList?.filterNotNull()?.size ?: 0, "discussions")
+            showNoData(binding.tvNodata, realmNewsList?.count { it != null } ?: 0, "discussions")
         } else {
             (existingAdapter as? VoicesAdapter)?.let { adapter ->
                 adapter.setCurrentUser(user)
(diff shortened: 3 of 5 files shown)


dogi:
**Review — rating: 84/100**

Three `filter`-then-`map` chains collapsed into single passes. All equivalent, and one is a clearly better shape than the others.

**The best of the three** is `TeamsVoicesFragment`: `realmNewsList?.filterNotNull()?.size` → `count { it != null }`. Building an entire filtered list purely to read `.size` off it is the kind of waste worth removing on sight, and `count {}` says what's meant. Same for the second site in that file.

**`ResourcesAdapter`** — `filterNotNull().map { it.copy(...) }` → `mapNotNull { it?.copy(...) }` — is a clean win too; it reads the same and drops an intermediate list on every `setLibraryList`.

**`SubmissionsRepositoryImpl`** is the one I'd push back on slightly. `chunk.filter { it !in existingUserIds }.map { … }` became `chunk.mapNotNull { if (it in existingUserIds) null else Submission().apply { … } }`. It's correct, and it does save a list — but `mapNotNull` with an explicit `null` branch is a mapping idiom being used to express a filter, and the reader now has to notice the `null` to see that exclusion is happening at all. Given the chunk here is at most 500 user ids, the saved allocation is small next to the legibility. `filter {}.map {}` was clearer; if the allocation genuinely matters, `asSequence().filter {}.map {}.toList()` keeps both.

Tests are well targeted: `ResourcesAdapterTest` gains 31 lines covering the null-handling in `setLibraryList`, which is exactly where a `filterNotNull` → `mapNotNull` swap could have gone wrong.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17142, opened 2026-09-10
all: smoother markdown utils image markup checking (fixes #17096)

### LINKED ISSUE #17096: Short-circuit prependBaseUrlToImages when the markdown has no image markup

dogi:
Short-circuited `MarkdownUtils.prependBaseUrlToImages` when input contains no `` image markup sequence and added unit tests in `MarkdownUtilsTest`.

Fixes #17096

---
*PR created automatically by Jules for task [6676706134332339384 started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/utils/MarkdownUtils.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/utils/MarkdownUtils.kt
@@ -86,6 +86,7 @@ object MarkdownUtils {
         height: Int = 100
     ): String {
         val content = markdownContent ?: return markdownContent.orEmpty()
+        if (!content.contains("![")) return content
         val matcher = imagePattern.matcher(content)
         val result = StringBuilder()
         var last = 0
--- a/app/src/test/java/org/ole/planet/myplanet/utils/MarkdownUtilsTest.kt
+++ b/app/src/test/java/org/ole/planet/myplanet/utils/MarkdownUtilsTest.kt
@@ -131,4 +131,16 @@ class MarkdownUtilsTest {
 
         assertSame(firstMovementMethod, secondMovementMethod)
     }
+
+    @Test
+    fun prependBaseUrlToImages_returns_identical_string_when_no_bracket_present() {
+        val markdown = "plain text with no bracket at all"
+        assertSame(markdown, MarkdownUtils.prependBaseUrlToImages(markdown, "http://base.url/"))
+    }
+
+    @Test
+    fun prependBaseUrlToImages_returns_identical_string_when_brackets_present_without_image_markup() {
+        val markdown = "plain text with (parens) and a ! and [brackets]"
+        assertSame(markdown, MarkdownUtils.prependBaseUrlToImages(markdown, "http://base.url/"))
+    }
 }


dogi:
**Re-review — Rating: 90/100** (was 76 on #17109, which this replaces)

My objection to #17109 was that its only test bypassed the line it was adding. Both halves are fixed here, and the guard itself got better in the process.

**The guard is now `!content.contains("[")`** rather than `contains("[")`. That's the actual required prefix for markdown image syntax, so it short-circuits strictly more input — plain text links like `[label](url)`, which are common in course descriptions, now skip the matcher too where before they'd have fallen through to the regex.

**Both tests exercise the guard, and they assert the right thing.** Using `assertSame` rather than `assertEquals` is the detail that makes this a real test of the optimization: the matcher path builds a `StringBuilder` and returns `result.toString()`, which is necessarily a new instance, so identity can only hold if the early return fired. `prependBaseUrlToImages_returns_identical_string_when_no_bracket_present` covers the obvious case, and `..._when_brackets_present_without_image_markup` reuses my old fixture to pin the widened behaviour — a string with `[` but no `!` is now short-circuited, and that test would fail against the `contains("[")` version.

The null path is untouched, so `markdownContent == null → orEmpty()` still behaves as before.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17143, opened 2026-09-10
all: smoother personals repository camera utils time stamping (fixes #17136)

### LINKED ISSUE #17136: Drop the throwaway Date/Calendar objects at three timestamp sites

Three places allocate a `Date` or `Calendar` only to read a value straight back out of it. None of these classes has an injectable clock available (an `object`, a model companion, and a fragment), so unlike the repository sites these are plain swaps — bundled so they cost one review instead of three.

**1. `utils/CameraUtils.kt:106`**

```kotlin
val photoFile = "${Date().time}.jpg"
```

**2. `model/Personal.kt:39`**, inside `Personal.Companion.serialize(personal, customDeviceName)`

```kotlin
```

**3. `ui/calendar/CalendarFragment.kt:25-26`**, in `onCreateView`

```kotlin
val calendar = Calendar.getInstance()
binding.calendarView.setDate(calendar.time)
```

Be accurate about the value in the PR description: #1 and #2 remove one allocation each on paths that run once per photo and once per upload. #3 trades a `Calendar` for a `Date` — the same number of objects, just the cheaper one — so it is a tidiness change, not a saving. Do not claim otherwise.

### Do

1. `CameraUtils:106` → `val photoFile = "${System.currentTimeMillis()}.jpg"`
2. `Personal.kt:39` → `` `object`.addProperty("uploadDate", System.currentTimeMillis()) ``
3. `CalendarFragment:25-26` → `binding.calendarView.setDate(Date())`, dropping the local `calendar` val.

Then fix imports per file: drop `java.util.Date` from `CameraUtils.kt` and `model/Personal.kt` **only if** nothing else in that file uses `Date` (check — `Personal.kt` may use it elsewhere), and in `CalendarFragment.kt` swap `java.util.Calendar` for `java.util.Date`.

### Constraints

- `System.currentTimeMillis()` returns exactly what `Date().time` returns — the same epoch-millis `Long`. Do not change the type, the unit, or the JSON key at `Personal.kt:39`, and do not reformat the timestamp.
- Do not route any of these through `TimeProvider`. None of the three classes is Hilt-injected, and #17130 already covers the repository sites where an injected clock is available and appropriate.
- Do not use `SystemClock.elapsedRealtime()` anywhere here — every value is a wall-clock timestamp (a filename stamp, an upload date, a calendar date).
- `CalendarFragment:26` must keep passing a `java.util.Date` to `setDate(...)` — `Calendar.getInstance().time` **is** a `Date`, so `Date()` is the exact equivalent. Do not switch to a `Long` overload or to `java.time`.
- `Personal.kt`: touch only line 39. Leave every other `addProperty` and the whole rest of `serialize` alone, including `createdDate` at :40.
- Do not touch `MyLibrary.kt` — its `uploadDate` already uses `System.currentTimeMillis()`, and issue #17039 is moving that serializer.

### Tests

No new tests are required and none are wanted for these three; asserting on the system clock makes a flaky test. Keep any existing tests in the three files passing. If a test currently pins `uploadDate`, update it to tolerate a current-millis value rather than freezing one.

### Out of scope

Do not touch `AppDatabase` or any Room schema. Do not add or change any log line, and never log a file path or a URL — myPlanet server URLs embed `satellite:<pin>@host` credentials and release builds do not strip logs.

dogi:
Three places allocate a Date or Calendar only to read a value straight back out of it.

1. CameraUtils.kt:106 - val photoFile = "${System.currentTimeMillis()}.jpg"
   (removes one Date allocation on paths that run once per photo; removed unused java.util.Date import)
2. Personal.kt:39 - `object`.addProperty("uploadDate", System.currentTimeMillis())
   (removes one Date allocation on paths that run once per upload; removed unused java.util.Date import)
3. CalendarFragment.kt:25-26 - binding.calendarView.setDate(Date())
   (trades a Calendar for a Date — tidiness change, replacing Calendar.getInstance() and calendar.time with Date(), swapping java.util.Calendar import for java.util.Date)

Fixes #17136

---
*PR created automatically by Jules for task [5733163340781087943](https://jules.google.com/task/5733163340781087943) started by @dogi*


dogi:
**Review — rating: 84/100**

Three equivalent substitutions, all correct:

- `Personal.serialize`: `Date().time` → `System.currentTimeMillis()` — `Date()` is literally `Date(System.currentTimeMillis())`, so `.time` round-trips back to the same `Long` through an object that's immediately discarded.
- `CameraUtils`: same swap for the photo filename.
- `CalendarFragment`: `Calendar.getInstance().time` → `Date()`. This is the one that actually matters. `Calendar.getInstance()` resolves the default `TimeZone` and `Locale` and builds a `GregorianCalendar` with its full field array, all to hand back `Date(millis)`. Replacing it with `Date()` skips that entirely, and it sits in `onCreateView` where it's on the path to first frame.

No behavioural difference at any of the three — same instant, same type reaching the API.

**Nit — worth reconciling with #17139 in this same batch.** That PR routes wall-clock reads in `ActivitiesRepositoryImpl` and `FeedbackRepositoryImpl` through `TimeProvider`, and here you're standing on two more direct clock reads and swapping one direct call for another. Not a regression — `Date()` read the clock just as directly — but `CameraUtils` and `CalendarFragment` could take a `TimeProvider` and become deterministic in tests, whereas after this they're still unmockable. `Personal.serialize` is a companion function so it's genuinely awkward there, but the other two aren't. Given the codebase has an explicit `TimeProvider` abstraction and a PR in flight to widen its use, it'd be good for these two efforts to agree on where the line is.

No tests, which is reasonable — asserting on `System.currentTimeMillis()` would be testing the clock.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17144, opened 2026-09-10
resources: smoother collections tag indexing (fixes #17090)

### LINKED ISSUE #17090: Index the collection tags once when reconciling selections

dogi:
Optimized tag selection reconciliation in CollectionsFragment by building candidate tag index maps (tagsById and tagsByName) in one pass over loaded tags using putIfAbsent. Lookups try ID first, then fall back to name, matching tags with empty IDs or names correctly. Pre-derived key sets (selectedIds and selectedNames) are used in buildTagDataList to check selection state for parent and child rows in O(1) time. Added comprehensive unit tests covering all mixed selection cases.

Fixes #17090

---
*PR created automatically by Jules for task [18352000736825933781](https://jules.google.com/task/18352000736825933781) started by @dogi*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tag selection reconciliation using tag IDs and names.
  * Preserved selected tags when matching loaded tags have missing or empty IDs.
  * Prevented incorrectly matching tags that share a name but have conflicting IDs.
  * Ensured parent and child selection states accurately reflect reconciled tags.
  * Handled empty tag lists without errors.

* **Tests**
  * Added coverage for matching, unmatched selections, conflicting IDs, duplicate names, and empty lists.

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/ui/resources/CollectionsFragment.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/resources/CollectionsFragment.kt
@@ -67,10 +67,7 @@ class CollectionsFragment : DialogFragment(), OnTagClickListener, CompoundButton
                 is CollectionsState.Success -> {
                     list = state.list
                     childMap = state.childMap
-                    val allTags = list + childMap.values.flatten()
-                    val reconciledList = selectedItemsList.map { selected ->
-                        allTags.find { selected.matches(it) } ?: selected
-                    }
+                    val reconciledList = reconcileSelections(selectedItemsList, list, childMap)
                     selectedItemsList.clear()
                     selectedItemsList.addAll(reconciledList)
                     currentTagDataList = buildTagDataList(list)
@@ -122,6 +119,39 @@ class CollectionsFragment : DialogFragment(), OnTagClickListener, CompoundButton
         adapter.submitList(currentTagDataList)
     }
 
+    internal fun reconcileSelections(
+        selectedItems: List<TagEntity>,
+        list: List<TagEntity>,
+        childMap: Map<String, List<TagEntity>>
+    ): List<TagEntity> {
+        val tagsById = HashMap<String, TagEntity>()
+        val tagsByName = HashMap<String, TagEntity>()
+
+        fun indexTag(tag: TagEntity) {
+            if (tag.id.isNotEmpty()) {
+                tagsById.putIfAbsent(tag.id, tag)
+            }
+            if (!tag.name.isNullOrEmpty()) {
+                tagsByName.putIfAbsent(tag.name!!, tag)
+            }
+        }
+
+        for (parent in list) {
+            indexTag(parent)
+        }
+        for (children in childMap.values) {
+            for (child in children) {
+                indexTag(child)
+            }
+        }
+
+        return selectedItems.map { selected ->
+            (if (selected.id.isNotEmpty()) tagsById[selected.id] else null)
+                ?: (if (!selected.name.isNullOrEmpty()) tagsByName[selected.name] else null)
+                ?: selected
+        }
+    }
+
     private fun buildTagDataList(parents: List<TagEntity>): List<TagData> {
         val tagDataList = mutableListOf<TagData>()
         val isSelectMultiple = MainApplication.isCollectionSwitchOn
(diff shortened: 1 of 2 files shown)


dogi:
**Re-review — Rating: 89/100** (was 65 on #17102, which this replaces)

The single-key index that broke `matches` is gone, replaced by three indexes that reproduce its truth table exactly. I walked all four cases against `TagEntity.matches` (`if (id.isNotEmpty() && other.id.isNotEmpty()) id == other.id else !name.isNullOrEmpty() && name == other.name`):

| selected | candidate | `matches` | new lookup |
|---|---|---|---|
| id set | id set | id equality | `tagsById[selected.id]` ✓ |
| id set | id **empty** | name equality | `namesOfEmptyIdTags[selected.name]` ✓ |
| id **empty** | any | name equality | `allTagsByName[selected.name]` ✓ |
| id empty, name empty | any | always false | both guards fail → `?: selected` ✓ |

Splitting `namesOfEmptyIdTags` out from `allTagsByName` is the part that makes this correct rather than approximately correct — it's precisely the asymmetry in `matches`, where a candidate's empty id forces the name branch even when the selection has a real id. That's the mixed case #17102 dropped, and it's the one this reconciliation exists for.

`buildTagDataList` gets the mirrored treatment via `isTagSelected` with `selectedIds` / `namesOfEmptyIdSelected` / `allSelectedNames`, so the selection highlighting can't disagree with what reconciliation just resolved. Both halves being derived from the same rule is what I'd want here.

Extracting `reconcileSelections` as `internal` is a good call independently of the perf work — it makes the rule testable without a fragment, and the ~130 lines of new tests are the payoff.

**Nit:** where a selection with a real id could match one candidate by id and a *different* one by name, this now prefers the id match, whereas master's `find` took whichever came first in `list + childMap.values`. The new precedence is the more defensible one; just noting it's a deliberate tie-break rather than a strict no-op.

**Nit:** the `tag.name!!` assertions after `isNullOrEmpty()` checks are safe but noisy — a local `val name = tag.name ?: return` style, or capturing into a local first, would let smart-casting do the work.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17052, opened 2026-09-10
all: less personals repository dao is more (fixes #17032)

dogi:
Tightens the personals boundary by dropping `suspend` from `getPersonalResources` (Flow accessors should be non-suspend), removing `uploadPersonalDocument` from the interface (as its only caller is `uploadPersonal` inside the implementation), deleting the unused `PersonalDao.update` method and import, and updating unit tests.

Fixes #17032

---
*PR created automatically by Jules for task [6233377538860778290](https://jules.google.com/task/6233377538860778290) started by @dogi*

PR #17146, opened 2026-09-10
resources: smoother courses repositories serialized uploading (fixes #17039)

### LINKED ISSUE #17039: Move MyLibrary.serialize (and its MainApplication.context reach) into ResourcesRepository

A Room entity used by sync and upload reaches into the Application context to serialize itself.

`model/MyLibrary.kt:12` — `import org.ole.planet.myplanet.MainApplication.Companion.context`

The companion `serialize` at `:162-197` is the only thing that uses it, at exactly one line:

```kotlin
addProperty("customDeviceName", NetworkUtils.getCustomDeviceName(context))   // :195
```

**The sweep:**

| Symbol | Callers |
|---|---|
| `MyLibrary.serialize(library, user)` | exactly one — `services/upload/UploadConfigs.kt:279` |
| `MyLibrary.serializeResource()` (`:89`) | two — `model/MyCourse.kt:134` and `ui/user/EditAchievementFragment.kt:426`. **Different API. Leave it completely alone.** |
| `MainApplication...context` inside `MyLibrary.kt` | only `:195` |

The pattern to copy already exists: `repository/PersonalsRepositoryImpl.kt` injects `private val deviceNameProvider: DeviceNameProvider` (`:20`) and calls `deviceNameProvider.getCustomDeviceName()` (`:77`). `DeviceNameProvider` is a Hilt-bound interface (`utils/DeviceNameProvider.kt:11`, bound from `SharedPrefDeviceNameProvider` at `:27`).

### Do

1. Add `suspend fun serializeForUpload(library: MyLibrary, user: UserEntity?): JsonObject` to `ResourcesRepository` and implement it in `ResourcesRepositoryImpl`.
2. Move the body of `MyLibrary.Companion.serialize` (`:163-196`) into that implementation **verbatim**, with one substitution: `NetworkUtils.getCustomDeviceName(context)` becomes `deviceNameProvider.getCustomDeviceName()`. Inject `DeviceNameProvider` into `ResourcesRepositoryImpl` if it is not already there.
3. Point `UploadConfigs.kt:279`'s serializer at the repository method instead of `MyLibrary.serialize`.
4. Delete the companion `serialize` from `MyLibrary.kt` and the `MainApplication.Companion.context` import at `:12`.

### The payload must be byte-identical

Same keys, in the same order, with the same values. Copy the body as-is, including its quirks — do not tidy them:

- `addProperty("createdDate", …)` appears **twice** (`:166` and `:192`). Keep both, in place.
- `uploadDate` already uses `System.currentTimeMillis()` (`:165`) — leave it; do not route it through `TimeProvider` in this task.
- `updatedDate` uses `Calendar.getInstance().timeInMillis` (`:191`) — leave it as `Calendar`.
- `NetworkUtils.getDeviceName()` (`:194`) takes no context — leave it untouched.
- Keep the `if (personal.isPrivate && personal.privateFor != null)` nested `privateFor` object exactly as it is, and keep the parameter name `personal` if that keeps the diff smaller.
- Keep `addDocumentOrigin()` and the `?: ""` defaults.

### Acceptance

- The uploaded resource payload is identical to today's for the same inputs.
- `model/MyLibrary.kt` no longer imports `MainApplication`.
- `serializeResource()` still works for `MyCourse` and the achievements UI.

### Tests

Add a `ResourcesRepositoryImpl` test that stubs `deviceNameProvider.getCustomDeviceName()` and asserts the returned `JsonObject` carries that value under `customDeviceName`, plus a spot check of a few other keys (`title`, `addedBy`, `sourcePlanet`). That is the payoff: this serializer becomes testable without a `Context`. Keep every existing test in the file passing.

### Collision notes — keep your diff narrow

Other open PRs touch these same files elsewhere: `ResourcesRepository`/`Impl` around the interface tail and `:158`/`:312`/`:347`/`:674-716`/`:846`, and `UploadConfigs.kt` around `CrashLog.markUploaded` (~:223) and the `ApkLogUpload` mapping. Your edits are: the interface tail, one new impl method, `UploadConfigs.kt:279`, and the deletion in `MyLibrary.kt`. Stay there.

### Out of scope

*(Re-filed with the caller sweep and the byte-identical payload rules spelled out: two previous attempts at this task produced no PR.)*

dogi:
Moved `MyLibrary.Companion.serialize` into `ResourcesRepositoryImpl.serializeForUpload`, replacing `NetworkUtils.getCustomDeviceName(context)` with `deviceNameProvider.getCustomDeviceName()`. Removed the companion `serialize` and unused `MainApplication` context import from `MyLibrary.kt`. Updated `UploadConfigs.kt` to use the repository method for resource upload serialization. Added unit tests in `ResourcesRepositoryImplTest`.

Fixes #17039

---
*PR created automatically by Jules for task [8284267676391500767](https://jules.google.com/task/8284267676391500767) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/model/MyLibrary.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/model/MyLibrary.kt
@@ -7,15 +7,11 @@ import androidx.room.PrimaryKey
 import com.google.gson.JsonArray
 import com.google.gson.JsonNull
 import com.google.gson.JsonObject
-import java.util.Calendar
 import java.util.UUID
-import org.ole.planet.myplanet.MainApplication.Companion.context
 import org.ole.planet.myplanet.services.SharedPrefManager
 import org.ole.planet.myplanet.utils.FileUtils
 import org.ole.planet.myplanet.utils.JsonUtils
-import org.ole.planet.myplanet.utils.NetworkUtils
 import org.ole.planet.myplanet.utils.Utilities
-import org.ole.planet.myplanet.utils.addDocumentOrigin
 
 /**
  * Room replacement for the former `MyLibrary` model (resources).
@@ -159,43 +155,6 @@ open class MyLibrary {
     }
 
     companion object {
-        fun serialize(personal: MyLibrary, user: UserEntity?): JsonObject {
-            return JsonObject().apply {
-                addProperty("title", personal.title)
-                addProperty("uploadDate", System.currentTimeMillis())
-                addProperty("createdDate", personal.createdDate)
-                addProperty("filename", FileUtils.getFileNameFromUrl(personal.resourceLocalAddress))
-                addProperty("author", personal.author ?: "")
-                addProperty("addedBy", user?.id)
-                addProperty("medium", personal.medium)
-                addProperty("description", personal.description)
-                addProperty("year", personal.year)
-                addProperty("language", personal.language)
-                addProperty("publisher", personal.publisher ?: "")
-                addProperty("linkToLicense", personal.linkToLicense ?: "")
-                add("subject", JsonUtils.getAsJsonArray(personal.subject))
-                add("level", JsonUtils.getAsJsonArray(personal.level))
-                addProperty("resourceType", personal.resourceType)
-                addProperty("openWith", personal.openWith)
-                addProperty("mediaType", personal.mediaType ?: "other")
-                add("resourceFor", JsonUtils.getAsJsonArray(personal.resourceFor))
-                addProperty("private", personal.isPrivate)
-                if (personal.isPrivate && personal.privateFor != null) {
-                    val privateForObj = JsonObject()
-                    privateForObj.addProperty("teams", personal.privateFor)
-                    add("privateFor", privateForObj)
-                }
-                addProperty("isDownloadable", true)
-                addProperty("sourcePlanet", user?.planetCode)
-                addProperty("resideOn", user?.planetCode)
-                addProperty("updatedDate", Calendar.getInstance().timeInMillis)
-                addProperty("createdDate", personal.createdDate)
-                addDocumentOrigin()
-                addProperty("deviceName", NetworkUtils.getDeviceName())
-                addProperty("customDeviceName", NetworkUtils.getCustomDeviceName(context))
-            }
-        }
-
         data class InsertParams(
             val doc: JsonObject,
             val spm: SharedPrefManager,
@@ -268,7 +227,7 @@ open class MyLibrary {
                         if (key.indexOf("/") < 0) {
                             resourceRemoteAddress = "${params.spm.getCouchdbUrl().ifEmpty { "http://" }}/resources/$resourceId/$key"
                             resourceLocalAddress = key
-                            resourceOffline = FileUtils.checkFileExist(context, resourceRemoteAddress)
+                            resourceOffline = FileUtils.checkFileExist(org.ole.planet.myplanet.MainApplication.context, resourceRemoteAddress)
                             if (resourceOffline) {
                                 downloadedRev = JsonUtils.getString("_rev", params.doc)
                             }
--- a/app/src/main/java/org/ole/planet/myplanet/model/MyLibrary.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/model/MyLibrary.kt
@@ -159,11 +159,11 @@ open class MyLibrary {
         data class InsertParams(
             val doc: JsonObject,
             val spm: SharedPrefManager,
+            val context: Context,
             val userId: String? = "",
             val stepId: String? = "",
             val courseId: String? = "",
-            val existing: MyLibrary? = null,
-            val context: Context? = null
+            val existing: MyLibrary? = null
         )
 
         private fun JsonArray?.mergeInto(target: MutableList<String>) {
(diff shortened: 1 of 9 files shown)


dogi:
**Review — rating: 76/100**

The move itself is right. A `@Entity` model reaching for `MainApplication.context` to resolve a device name is the wrong direction of dependency, and relocating serialization into `ResourcesRepositoryImpl` with an injected `DeviceNameProvider` puts it where the collaborators already live. The field-for-field copy is faithful — I checked every property, including the `isPrivate && privateFor != null` branch and both `resourceFor`/`subject`/`level` array conversions.

Two things I'd like fixed first, both small.

**1. The `suspend` is gratuitous, and it costs something.** `serializeForUpload` is declared `suspend`, but nothing in its body suspends: it's `JsonObject` construction plus `deviceNameProvider.getCustomDeviceName()`, which is a plain `fun` on the interface (`DeviceNameProvider.kt:12`) backed by a `SharedPreferences` read. That unnecessary modifier is what forced `UploadConfigs` from `UploadSerializer.Simple` to `UploadSerializer.Async`, so the upload pipeline now takes the coroutine-bearing path for a synchronous function. Dropping `suspend` lets `Simple` stay. This codebase already treats the distinction as meaningful — #17052 in this queue deliberately *removed* `suspend` from `getPersonalResources` for exactly this reason.

**2. `MainApplication.context` didn't actually leave the model.** Line 230 still has it, and because the `import ...MainApplication.Companion.context` was removed it's now written out longhand:

```kotlin
resourceOffline = FileUtils.checkFileExist(org.ole.planet.myplanet.MainApplication.context, resourceRemoteAddress)
```

So `MyLibrary` is no more context-free than before, and that call site reads worse than master did. Either keep the import, or move `insertMyLibrary`'s context dependency out too and finish the decoupling.

**Nit:** `addProperty("createdDate", …)` appears twice in the object — once near the top and again after `resideOn`. Harmless (the second overwrites), pre-existing, but worth dropping while the code is being moved.

**Nit:** `Calendar.getInstance().timeInMillis` came across verbatim. #17143 in this same batch is removing exactly that pattern in favour of `System.currentTimeMillis()`; worth aligning.

**Verdict:** Needs another round — right refactor, but drop the `suspend` and don't leave a lone fully-qualified `MainApplication.context` behind.

---

dogi:
**Re-review — Rating: 68/100**

The `DeviceNameProvider` injection that blocked the previous round is resolved — `deviceNameProvider.getCustomDeviceName()` is non-suspend and called from a non-suspend `serializeForUpload`, which type-checks and matches the provider's actual signature. The move of `serialize` out of the `MyLibrary` companion and into `ResourcesRepository` is the right direction: it gets `MainApplication.context` and `NetworkUtils.getCustomDeviceName(context)` out of a Room entity.

But the `context` refactor in `InsertParams` introduces a behavioural regression.

**The problem**

The PR changes the resource-offline check from a global context to an optional parameter:

```kotlin
resourceOffline = params.context?.let { FileUtils.checkFileExist(it, resourceRemoteAddress) } ?: false
```

`context` is added to `InsertParams` as `val context: Context? = null`. There are three construction sites:

- `ResourcesRepositoryImpl.kt:655` — passes `context = context` ✓
- `ResourcesRepositoryImpl.kt:701` — passes `context = context` ✓
- `CoursesRepositoryImpl.kt:841` — **does not pass it**

So for every resource inserted through the course-resource batch path, `params.context` is `null` and `resourceOffline` is now hardcoded `false`. Before this PR that path used `MainApplication.context` and computed the real value.

**Failure mode:** a course-attached resource the user has already downloaded gets written back with `resourceOffline = false`, and `downloadedRev` is left unset because it's only assigned inside `if (resourceOffline)`. The resource then reads as not-downloaded — offered for download again, missed by offline-resource queries. It's silent, it only shows up after a sync, and the default parameter value means the compiler can't catch it.

**Fix**

Either pass `context` at `CoursesRepositoryImpl.kt:841` as well, or — better — make the parameter non-optional (`val context: Context`). A required parameter turns this exact class of omission into a compile error, which matters here because the whole point of the refactor is to stop relying on an ambient global; leaving a `null` default preserves the old silent-fallback hazard in a new shape.

**Two smaller things while you're in here**

`addProperty("createdDate", personal.createdDate)` appears twice in `serializeForUpload` — once near the top and once near the bottom. Gson overwrites, so it's harmless, but it was carried over verbatim from the old `serialize` and this was the moment to drop the duplicate.

`uploadDate` and `updatedDate` both use raw `System.currentTimeMillis()`. `updatedDate` was `Calendar.getInstance().timeInMillis` before, so the value is unchanged — but sibling PRs in the queue (#17139, #17143) are actively migrating wall-clock reads onto the injected `TimeProvider`, and `ResourcesRepositoryImpl` is a repository with DI already wired. Routing these two through `TimeProvider` would make the new method testable and keep it from being re-touched by that migration later.

**Verdict:** Needs a change — the `CoursesRepositoryImpl` call site.

---

dogi:
**Re-review — Rating: 90/100** (was 68)

The `CoursesRepositoryImpl` regression is fixed, and fixed the better of the two ways I suggested.

**The fix**

```kotlin
data class InsertParams(
    val doc: JsonObject,
    val spm: SharedPrefManager,
    val context: Context,          // required, no default
    ...
```

Making `context` non-optional and moving it above the defaulted parameters means a caller that forgets it is a compile error, not a silent `false`. That's what the refactor was for — the whole point of removing `MainApplication.context` is to stop relying on an ambient global, and a `Context? = null` default would have preserved the same silent-fallback hazard in a new shape. With the parameter required, `resourceOffline = FileUtils.checkFileExist(params.context, resourceRemoteAddress)` reads straight through with no fallback at all.

`CoursesRepositoryImpl` now takes `@param:ApplicationContext private val context: Context` and passes it at `:844`, so all three construction sites supply a real context. I re-checked: there are exactly three, and all three are covered.

**Both nits picked up too**

`TimeProvider` is injected and both `uploadDate` and `updatedDate` now read `timeProvider.now()` — so `serializeForUpload` is deterministic under test and won't be re-touched by the #17139/#17143 migration. And the duplicate `addProperty("createdDate", …)` is gone; there's one occurrence left where there were two.

The test-side updates are mechanical constructor-argument additions, which is expected for two new injected dependencies.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17147, opened 2026-09-10
courses: smoother answer dao querying (fixes #17138)

### LINKED ISSUE #17138: Chunk AnswerDao's IN (:submissionIds) queries inside the DAO

`data/room/dao/AnswerDao.kt` has two `IN (…)` queries with no bound on the parameter count:

```kotlin
@Query("SELECT * FROM answers WHERE submissionId IN (:submissionIds)")
suspend fun getBySubmissionIds(submissionIds: List<String>): List<Answer>          // :11

@Query("DELETE FROM answers WHERE submissionId IN (:submissionIds)")
suspend fun deleteBySubmissionIds(submissionIds: List<String>): Int               // :13
```

SQLite's default host-parameter limit is 999. Past that, Room raises `SQLiteException: too many SQL variables` at runtime — a crash that only appears on devices with enough history, which is the worst way to find out.

**The callers pass unbounded lists.** The central one is `repository/SubmissionsRepositoryImpl.kt:70-74`:

```kotlin
private suspend fun hydrateSubmissions(rows: List<Submission>): List<Submission> {
    if (rows.isEmpty()) return emptyList()
    val answersBySubmissionId = answerDao.getBySubmissionIds(rows.map { it.id }).groupBy { it.submissionId }
    …
```

`rows` comes straight from a query with no `LIMIT` — `submissionDao.getPendingSurveys(userId)` (:92), `getUniquePendingSurveyCandidates(userId)` (:103), `getByIds(ids)` (:173), `getByParentUserAndStatus(...)` (:327) — so the id count scales with a user's whole submission history. The other unchunked callers are `repository/CoursesRepositoryImpl.kt:457`, `repository/ProgressRepositoryImpl.kt:146`, `repository/SubmissionsRepositoryExporter.kt:138` and `SubmissionsRepositoryImpl.kt:354` (the delete).

**Chunking inside the DAO is the established pattern in this package**, so this needs no new convention: `data/room/dao/HealthExaminationDao.kt:39`, `data/room/dao/NotificationDao.kt:78` and `:84`, and `data/room/dao/RemovedLogDao.kt:20` all chunk (900, and 250 where each row carries more parameters) in a default interface method wrapping the raw `@Query`. `CoursesRepositoryImpl.kt:602` already chunks `deleteBySubmissionIds` at one call site — this moves that safety to every caller instead of one.

### Do

Keep the raw queries but rename them, and expose chunking default methods under the current names so **no call site has to change**:

```kotlin
@Query("SELECT * FROM answers WHERE submissionId IN (:submissionIds)")
suspend fun getBySubmissionIdsInternal(submissionIds: List<String>): List<Answer>

suspend fun getBySubmissionIds(submissionIds: List<String>): List<Answer> {
    if (submissionIds.isEmpty()) return emptyList()
    return submissionIds.chunked(900).flatMap { getBySubmissionIdsInternal(it) }
}

@Query("DELETE FROM answers WHERE submissionId IN (:submissionIds)")
suspend fun deleteBySubmissionIdsInternal(submissionIds: List<String>): Int

suspend fun deleteBySubmissionIds(submissionIds: List<String>): Int {
    if (submissionIds.isEmpty()) return 0
    return submissionIds.chunked(900).sumOf { deleteBySubmissionIdsInternal(it) }
}
```

Use 900, matching the three existing DAOs.

### Constraints

- **Do not change any call site.** The public names and signatures stay identical; that is the point of doing this in the DAO. `CoursesRepositoryImpl.kt:602`'s existing outer `chunked(900)` may stay as-is — it becomes redundant but harmless, and removing it widens the diff into a file other PRs touch.
- Empty input must short-circuit without hitting the database — both `getBySubmissionIds(emptyList())` and `deleteBySubmissionIds(emptyList())` return early. (Note `IN ()` is also invalid SQL, so this guard is load-bearing, not just an optimisation.)
- The delete must return the **sum** across chunks, so existing callers see the same total row count.
- Do not add `@Transaction` to the delete wrapper: today's single statement is not transactional across a chunked loop either way, and adding one changes locking behaviour for a task that is only about the variable limit. If you think it is needed, say so in the PR instead of doing it.
- Do not sort or de-duplicate the results. Every caller either `groupBy { it.submissionId }`s them (`SubmissionsRepositoryImpl:72`, `SubmissionsRepositoryExporter:138`) or aggregates them, so chunk-concatenation order is fine — but do not impose a new order either.
- Do not touch the other three methods in the file (`getBySubmissionId`, `getBySubmissionAndQuestion`, the `@Upsert` pair), and do not change the entity.

### Tests

Add a Room in-memory test (`app/src/test/java/org/ole/planet/myplanet/data/room/dao/AnswerDaoTest.kt`, following the `CourseProgressDaoTest` pattern of `Room.inMemoryDatabaseBuilder` + `AndroidJUnit4`):

- insert answers for **1200** distinct submission ids and assert `getBySubmissionIds` with all 1200 ids returns all of them — this is the case that throws today;
- the same for `deleteBySubmissionIds`, asserting the returned count is 1200 and the rows are gone;
- both methods with an empty list return `emptyList()` / `0`;
- a mixed list where some ids have no answers returns only the existing ones.

### Out of scope

Do not touch `AppDatabase` or any Room schema — this adds no entity, column or index, so **no version bump**. Do not add or change any log line. Other `IN (:ids)` DAOs have the same latent issue; this issue is scoped to `AnswerDao` so the pattern can be reviewed once before being applied more widely.

dogi:
Chunks AnswerDao's IN (:submissionIds) queries inside the DAO in batches of 900 to prevent SQLite host parameter limit crashes while preserving the existing public interface and adding comprehensive unit tests.

Fixes #17138

---
*PR created automatically by Jules for task [18190622294553902591](https://jules.google.com/task/18190622294553902591) started by @dogi*

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved answer retrieval and deletion when processing large numbers of submissions.
  - Prevented failures caused by oversized requests by automatically processing submissions in manageable batches.
  - Empty submission lists now return no results or perform no deletion without errors.
  - Deletion results accurately report the total number of answers removed, including requests containing both existing and non-existing submissions.

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/data/room/dao/AnswerDao.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/data/room/dao/AnswerDao.kt
@@ -8,9 +8,25 @@ import org.ole.planet.myplanet.model.Answer
 @Dao
 interface AnswerDao {
     @Query("SELECT * FROM answers WHERE submissionId = :submissionId") suspend fun getBySubmissionId(submissionId: String): List<Answer>
-    @Query("SELECT * FROM answers WHERE submissionId IN (:submissionIds)") suspend fun getBySubmissionIds(submissionIds: List<String>): List<Answer>
+
+    @Query("SELECT * FROM answers WHERE submissionId IN (:submissionIds)")
+    suspend fun getBySubmissionIdsInternal(submissionIds: List<String>): List<Answer>
+
+    suspend fun getBySubmissionIds(submissionIds: List<String>): List<Answer> {
+        if (submissionIds.isEmpty()) return emptyList()
+        return submissionIds.chunked(900).flatMap { getBySubmissionIdsInternal(it) }
+    }
+
     @Query("SELECT * FROM answers WHERE submissionId = :submissionId AND questionId = :questionId LIMIT 1") suspend fun getBySubmissionAndQuestion(submissionId: String, questionId: String?): Answer?
-    @Query("DELETE FROM answers WHERE submissionId IN (:submissionIds)") suspend fun deleteBySubmissionIds(submissionIds: List<String>): Int
+
+    @Query("DELETE FROM answers WHERE submissionId IN (:submissionIds)")
+    suspend fun deleteBySubmissionIdsInternal(submissionIds: List<String>): Int
+
+    suspend fun deleteBySubmissionIds(submissionIds: List<String>): Int {
+        if (submissionIds.isEmpty()) return 0
+        return submissionIds.chunked(900).sumOf { deleteBySubmissionIdsInternal(it) }
+    }
+
     @Upsert suspend fun upsertAll(items: List<Answer>)
     @Upsert fun upsertAllBlocking(items: List<Answer>)
 }
--- a/app/src/test/java/org/ole/planet/myplanet/data/room/dao/AnswerDaoTest.kt
+++ b/app/src/test/java/org/ole/planet/myplanet/data/room/dao/AnswerDaoTest.kt
@@ -1,0 +1,109 @@
+package org.ole.planet.myplanet.data.room.dao
+
+import android.app.Application
+import androidx.room.Room
+import androidx.test.core.app.ApplicationProvider
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import kotlinx.coroutines.runBlocking
+import org.junit.After
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertTrue
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.ole.planet.myplanet.data.room.AppDatabase
+import org.ole.planet.myplanet.model.Answer
+import org.robolectric.annotation.Config
+
+@RunWith(AndroidJUnit4::class)
+@Config(application = Application::class)
+class AnswerDaoTest {
+
+    private lateinit var database: AppDatabase
+    private lateinit var answerDao: AnswerDao
+
+    @Before
+    fun initDb() {
+        database = Room.inMemoryDatabaseBuilder(
+            ApplicationProvider.getApplicationContext(),
+            AppDatabase::class.java
+        ).allowMainThreadQueries().build()
+        answerDao = database.answerDao()
+    }
+
+    @After
+    fun closeDb() {
+        database.close()
+    }
+
+    private fun createAnswer(id: String, submissionId: String): Answer {
+        return Answer(id = id, submissionId = submissionId)
+    }
+
+    @Test
+    fun getBySubmissionIds_with1200SubmissionIds_returnsAllAnswersWithoutThrowing() = runBlocking {
+        val answers = (1..1200).map { i ->
+            createAnswer("ans_$i", "sub_$i")
+        }
+        answerDao.upsertAll(answers)
+
+        val allSubmissionIds = answers.map { it.submissionId!! }
+        val results = answerDao.getBySubmissionIds(allSubmissionIds)
+
+        assertEquals(1200, results.size)
+        assertEquals(answers.map { it.id }.toSet(), results.map { it.id }.toSet())
+    }
+
+    @Test
+    fun deleteBySubmissionIds_with1200SubmissionIds_deletesAllAndReturnsTotalCount() = runBlocking {
+        val answers = (1..1200).map { i ->
+            createAnswer("ans_$i", "sub_$i")
+        }
+        answerDao.upsertAll(answers)
+
+        val allSubmissionIds = answers.map { it.submissionId!! }
+        val deletedCount = answerDao.deleteBySubmissionIds(allSubmissionIds)
+
+        assertEquals(1200, deletedCount)
+
+        val remainingAnswers = answerDao.getBySubmissionIds(allSubmissionIds)
+        assertTrue(remainingAnswers.isEmpty())
+    }
+
+    @Test
+    fun getBySubmissionIds_and_deleteBySubmissionIds_withEmptyList_shortCircuitAndReturnEmptyOrZero() = runBlocking {
+        val answer = createAnswer("ans_1", "sub_1")
+        answerDao.upsertAll(listOf(answer))
+
+        val selectResult = answerDao.getBySubmissionIds(emptyList())
+        assertTrue(selectResult.isEmpty())
+
+        val deleteCount = answerDao.deleteBySubmissionIds(emptyList())
+        assertEquals(0, deleteCount)
+
+        // Verify existing answer in database was not affected
+        val existing = answerDao.getBySubmissionId("sub_1")
+        assertEquals(1, existing.size)
+    }
+
+    @Test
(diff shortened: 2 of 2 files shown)


dogi:
**Review — rating: 87/100**

Wraps both `IN (:submissionIds)` queries in chunked default methods, keeping the raw `@Query` versions as `*Internal`.

This closes a real crash, not a theoretical one. SQLite's `SQLITE_MAX_VARIABLE_NUMBER` is 999, and `getBySubmissionIds` was called with an unbounded list — a device with a large backlog of pending submissions would throw rather than degrade. Chunking at 900 leaves headroom, and putting it *inside* the DAO is the right place: every caller is protected without having to remember, which is exactly the mistake the codebase already avoids elsewhere (`ProgressRepositoryImpl`, `ApkLogDao`, `ResourcesRepositoryImpl` all chunk at 900 too, so this is consistent rather than novel).

The empty-list guards are worth having on their own — `IN ()` is not valid SQLite, so `emptyList()` / `0` short-circuits a query that could never have succeeded.

Return semantics hold: `flatMap` concatenates the reads, and `sumOf` on the delete totals the per-chunk row counts, matching the single statement's return. 109 lines of new `AnswerDaoTest` against a real database, which is the only way to actually exercise a chunk boundary.

**Worth flagging for merge ordering:** #17157 in this same batch rewrites `getPendingExamResults` / `getPendingSubmissionsForUpload` to go through `hydrateSubmissions`, which calls `answerDao.getBySubmissionIds(rows.map { it.id })` — i.e. it converts an N+1 loop into exactly the unbounded bulk call this PR makes safe. The two are complementary, but **this one should land first**, otherwise #17157 briefly introduces a path where a large pending set hits the variable ceiling.

**Nit:** `deleteBySubmissionIds` was a single atomic `DELETE`; chunked, a failure mid-way leaves some answers deleted and others not. Annotating the default method `@Transaction` restores all-or-nothing — that's what `ApkLogDao.markUploadedBatch` does for the same reason. Less pressing for the read, but the delete is a mutation.

**Verdict:** Mergeable.

---

Okuro3499 (dismissed):
```
File: app/src/main/java/org/ole/planet/myplanet/data/room/dao/AnswerDao.kt, deleteBySubmissionIds (line 25-28)                                                                   
    Problem: no @Transaction annotation on the chunked delete wrapper, so the delete is not atomic across chunks — a regression from the original single-statement IN                
    (:submissionIds) query, and inconsistent with every other chunked-mutation DAO method in this repo.
```

dogi:
**Re-review — Rating: 90/100**

Chunks both `IN (:submissionIds)` queries inside the DAO, so callers can't reintroduce the parameter-limit crash by forgetting to chunk at the call site. That placement is the right one.

**Arithmetic and semantics**

Chunk size 900 against a single `IN` clause: 900 < 999. Correct. (Contrast #17158, where the query carries three `IN` clauses over the same list and the chunk size has to be 300 — same ceiling, different divisor. Both are right for their own query, which is the point of doing this per-DAO-method rather than with a shared constant.)

`getBySubmissionIds` correctly omits a `distinctBy`. Chunks are disjoint and `submissionId` is a single column, so a row can match at most one chunk — no duplicates are possible, and adding a dedup would only cost an allocation. `deleteBySubmissionIds` sums the per-chunk counts, matching what the single statement returned. Empty-list short-circuits on both. ✓

**Tests are the strongest in this batch**

A real in-memory Room database, 1200 ids to cross the 900 boundary in both directions, an empty-list case that also asserts *existing rows are untouched*, and a mixed existing/non-existing case that checks both the returned set and what survives a delete. That last one is what catches an off-by-one in chunk assembly, which a round-number test would miss.

**One thing I'd add**

`deleteBySubmissionIds` was a single atomic statement; it's now 2+ statements for any list over 900. Without a transaction, a failure or cancellation partway through leaves some submissions' answers deleted and others not — orphaned answer rows on the next sync. `@Transaction` on the default method restores the old guarantee:

```kotlin
@Transaction
suspend fun deleteBySubmissionIds(submissionIds: List<String>): Int {
```

#17045 in this same queue already does this on its batched DAO method, so the pattern is established. Worth applying here, and it costs one line. The read path doesn't strictly need it, though wrapping it too would give a consistent snapshot across chunks.

**Note on CI:** only `myPlanet labels` has reported on the current head (`cc2b539`). Build and test are *absent* rather than failed — this isn't a doc-only PR, so `paths-ignore` doesn't explain it; the runs look to have been dropped or superseded. Nothing here suggests a problem with the code, but the head currently carries no build or test verdict. The automerge drain builds and waits on its own prepared commit, so this won't slip through unverified — just worth knowing that the green you see on the PR page isn't a full green.

**Ordering:** this should land before #17157, which calls `answerDao.getBySubmissionIds` from the pending-submission upload path and relies on the chunking being inside the DAO.

**Verdict:** Mergeable.

---

Okuro3499 (dismissed):
deleteBySubmissionIds still has no @Transaction annotation. The must-fix issue is not resolved

dogi:

The `@Transaction` is in, and it's the only change since the last round — one import, one annotation:

```kotlin
@Transaction
suspend fun deleteBySubmissionIds(submissionIds: List<String>): Int {
    if (submissionIds.isEmpty()) return 0
    return submissionIds.chunked(900).sumOf { deleteBySubmissionIdsInternal(it) }
}
```

That restores the guarantee the original single statement had. Splitting a `DELETE ... IN (...)` into 2+ statements meant a failure or cancellation partway through could leave some submissions' answers deleted and others not — orphaned rows that nothing would clean up on the next sync. Now the whole chunk sequence commits or none of it does.

It's correctly applied to the delete only. `getBySubmissionIds` is left unwrapped, which is the right call: a read doesn't have a partial-failure state to protect against, and the cost of wrapping it is a transaction per call for a marginal consistency gain. Applying the annotation where it changes behaviour and not where it doesn't is better than applying it uniformly.

Everything from the previous review still holds — chunk size 900 correct for the single `IN` clause, no `distinctBy` needed on the read since chunks are disjoint over one column, `sumOf` matching the original return semantics, and the DAO tests covering 1200 ids across the chunk boundary plus the empty and mixed-existence cases.

**CI is now complete.** Last round the head carried only `myPlanet labels` — build and test had been dropped rather than failed, so the PR had no real verdict. This head has all five green, both flavor builds and both test shards. That gap is closed.

**Merge ordering — this matters more now than it did**

#17157 calls `answerDao.getBySubmissionIds` from the pending-submission upload path and depends on the chunking living inside the DAO. It is currently labelled `merge` while this PR is still in review, so the queue has them in the wrong order. If #17157 lands first, that path is exposed to the SQLite 999-parameter crash this PR exists to prevent, for any user with more than 900 pending submissions.

Worth pulling this one forward, or holding #17157 until it lands.

**One housekeeping note:** the branch is a long way behind master now — its merge base predates several days of merges. The automerge drain merges the base in before building, so this resolves itself, but the green checks above were run against a stale base.

**Verdict:** Mergeable.

---

PR #17149, opened 2026-09-10
sync: smoother url utils credential logging (fixes #17137)

### LINKED ISSUE #17137: Redact the credential-bearing URLs written to logcat by the download and bulk-upload paths
### The chain, verified end to end against `origin/master` (`a42f18a`)

Ten log statements interpolate a URL that contains the CouchDB **username and server PIN in its userinfo component**. Every step of that chain is in the tree today:

1. `ui/sync/ProcessUserDataActivity.kt:138` builds the stored CouchDB URL with credentials inline:
   ```kotlin
   couchdbURL = "${uri.scheme}://$urlUser:$urlPwd@${uri.host}:$port"
   ```
   and persists it at `:145` via `prefData.setCouchdbUrl(couchdbURL)`.
2. `services/SharedPrefManager.kt:165` — `getCouchdbUrl()` reads it back.
3. `utils/UrlUtils.kt:114-136` — `baseUrl(spm)` returns exactly that string (or `getProcessedAlternativeUrl()`, which `ServerConfigUtils.saveAlternativeUrl` builds the same way).
4. `utils/UrlUtils.kt:139-142` — `dbUrl(spm)` = `baseUrl` + `/db`; `:178` — `getUrl()` = `dbUrl(spm())`.
5. `utils/UrlUtils.kt:152-162` — `getUrl(library)` → `getUrl(id, file)` → `"$base/resources/$id/$file"`.

So **every resource download URL and every `_bulk_docs` upload URL carries `<user>:<pin>@host`**. Those exact strings are then logged:

| File | Lines | Statement |
|---|---|---|
| `services/DownloadWorker.kt` | 68, 77, 82, 99, 111, 116, 152 | `Log.e(TAG, "… $url", e)` (seven sites) |
| `services/DownloadService.kt` | 251 | `Log.d(TAG, "initDownload: retrying with $altUrl")` |
| `services/upload/BulkDocsUploader.kt` | 35, 51 | `Log.w/e(TAG, "… bulk upload to $url …")` — callers pass `"${UrlUtils.getUrl()}/teams/_bulk_docs"` (`services/upload/TeamsUploader.kt:44`, `services/UploadManager.kt:354`) |

`app/build.gradle` has `minifyEnabled = false` for release and there is no log-stripping ProGuard rule, so these lines execute in shipped builds. Any failed download or partial bulk upload writes the server PIN into logcat, where it reaches anything holding `READ_LOGS` and every bug report captured from the device.

The codebase already knows the safe pattern: `DownloadService.kt:154` deliberately logs only `nextUrl.url.substringAfterLast('/')`, and `getFileNameFromUrl(url)` is used at `DownloadWorker.kt:142` and `:186`.

### Do

1. Add a redaction helper to `utils/UrlUtils.kt`:

```kotlin
/** Strips the userinfo (user:password@) from a URL so it is safe to log. */
fun redactForLog(url: String): String
```

Parse with `android.net.Uri` or `java.net.URI` and rebuild without userinfo, e.g. `scheme://host[:port]/path`. It must **never throw**: on a malformed URL return a constant placeholder such as `"<unparseable url>"` rather than the input, because an unparseable string is exactly the case where the raw credentials would slip through.

2. Replace the interpolated URL at all ten sites with `UrlUtils.redactForLog(url)` (and `redactForLog(altUrl)` at `DownloadService.kt:251`). Keep every message's wording, level, tag and the exception argument otherwise unchanged.

3. Where the message only needs to identify the file, prefer the existing `getFileNameFromUrl(url)` — it is strictly less information. `DownloadWorker.kt:68`, `:77`, `:82` and `:111` read fine that way.

### Constraints

- **Do not delete the log lines.** They are diagnostics for real download and upload failures; the fix is redaction, not silence.
- Do not change the URLs actually used for requests — only what is written to the log. `downloadRepository.downloadFileResponse(url, authHeader)` and `uploadRepository.postUploadArray(url, payload)` keep receiving the full URL.
- Do not touch `ProcessUserDataActivity.kt:138`, `ServerConfigUtils.saveAlternativeUrl`, or how the credentials are stored. Moving credentials out of the URL into a header is a much larger change and is **not** this task.
- Do not touch `UrlUtils.header` / `basicAuthHeader` — those already keep credentials out of URLs.
- Leave `DownloadService.kt:154` as it is; it is already redacted by `substringAfterLast('/')`.
- `services/SubmissionsUploader.kt:24`, `:34`, `:48` are **out of scope**: those log `sharedPrefManager.getServerUrl()` and the `ServerUrlMapper` values, which come from the plain typed server URL and carry no userinfo. Do not change them.
- Be careful in `services/UploadManager.kt`: PR #17086 also edits that file (the succeeded-item attachment path in `uploadResource`). Your only change there is the argument at the `BulkDocsUploader.upload(...)` call site region around `:354`, and only if a log line there needs it — the actual logging lives in `BulkDocsUploader`.

### Tests

Add `UrlUtilsTest` coverage for `redactForLog` — this is the part that must not regress:

- `http://satellite:1234@host.org:5984/db/resources/abc/file.pdf` → no `satellite`, no `1234`, and no `@` in the result, while host, port and path survive;
- an `https` URL with a port;
- a URL with **no** userinfo is returned unchanged in substance;
- an empty string and a malformed string (e.g. `"not a url"`) return the placeholder and do not throw;
- a URL whose *password* contains an `@` or a `:` is still fully redacted.

Assert on the **absence** of the credential substrings, and do not put a real PIN from `gradle.properties` in a test fixture — use obvious dummies.

### Out of scope

Do not touch `AppDatabase` or any Room schema. Do not add any new log line beyond the redacted replacements.

dogi:
Redact credential-bearing URLs written to logcat across download and bulk upload paths using UrlUtils.redactForLog and getFileNameFromUrl, preventing credential leaks while keeping diagnostic logs intact.

Fixes #17137

---
*PR created automatically by Jules for task [9423076105100888388](https://jules.google.com/task/9423076105100888388) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/services/DownloadService.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/services/DownloadService.kt
@@ -48,7 +48,7 @@ import org.ole.planet.myplanet.utils.FileUtils
 import org.ole.planet.myplanet.utils.FileUtils.availableExternalMemorySize
 import org.ole.planet.myplanet.utils.FileUtils.externalMemoryAvailable
 import org.ole.planet.myplanet.utils.FileUtils.getFileNameFromUrl
-import org.ole.planet.myplanet.utils.UrlUtils.header
+import org.ole.planet.myplanet.utils.UrlUtils
 
 @AndroidEntryPoint
 class DownloadService : Service() {
@@ -234,7 +234,7 @@ class DownloadService : Service() {
                 return true
             }
 
-            val authHeader = header
+            val authHeader = UrlUtils.header
             if (authHeader.isBlank()) {
                 Log.e(TAG, "initDownload: auth header is blank — user may not be logged in")
                 downloadFailed("Authentication header not available", fromSync)
@@ -248,7 +248,7 @@ class DownloadService : Service() {
                 Log.w(TAG, "initDownload: primary failed with network error (${primaryResult.message}), checking for alternative URL")
                 val altUrl = resolveAlternativeUrl(url, fileName)
                 if (altUrl != null) {
-                    Log.d(TAG, "initDownload: retrying with $altUrl")
+                    Log.d(TAG, "initDownload: retrying with ${UrlUtils.redactForLog(altUrl)}")
                     currentDownloadUrl = altUrl
                     val altResult = downloadRepository.downloadFileResponse(altUrl, authHeader)
                     return tryDownloadFromResult(altResult, altUrl, fromSync, fileName, isAlternative = true)
--- a/app/src/main/java/org/ole/planet/myplanet/services/DownloadService.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/services/DownloadService.kt
@@ -282,7 +282,7 @@ class DownloadService : Service() {
             if (storedAlt.isNotEmpty() && primaryBase != null) {
                 resolvedAltBase = storedAlt.trimEnd('/')
                 resolvedPrimaryBase = primaryBase
-                Log.d(TAG, "initDownload: no hardcoded mapping for $primaryBase — using stored alternative $resolvedAltBase")
+                Log.d(TAG, "initDownload: no hardcoded mapping for $primaryBase — using stored alternative ${UrlUtils.redactForLog(resolvedAltBase)}")
             } else {
                 resolvedAltBase = null
                 resolvedPrimaryBase = null
@@ -295,7 +295,7 @@ class DownloadService : Service() {
             val path = parsed.path.orEmpty()
             val query = if (parsed.query != null) "?${parsed.query}" else ""
             val altUrl = resolvedAltBase + path + query
-            Log.d(TAG, "initDownload: switching $fileName — primary=$resolvedPrimaryBase → alternative=$resolvedAltBase")
+            Log.d(TAG, "initDownload: switching $fileName — primary=$resolvedPrimaryBase → alternative=${UrlUtils.redactForLog(resolvedAltBase)}")
             return altUrl
         }
         return null
--- a/app/src/main/java/org/ole/planet/myplanet/services/DownloadWorker.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/services/DownloadWorker.kt
@@ -65,7 +65,7 @@ class DownloadWorker @AssistedInject constructor(
                 val success = try {
                     downloadFile(url, authHeader, index, urls.size)
                 } catch (e: Exception) {
-                    Log.e(TAG, "Failed to download $url", e)
+                    Log.e(TAG, "Failed to download ${getFileNameFromUrl(url)}", e)
                     false
                 }
                 results.add(success)
@@ -74,12 +74,12 @@ class DownloadWorker @AssistedInject constructor(
                 try {
                     showProgressNotification(completedCount - 1, urls.size, context.getString(R.string.downloaded_files, "$completedCount", "${urls.size}"), 100)
                 } catch (e: Exception) {
-                    Log.e(TAG, "Failed to update progress notification for $url", e)
+                    Log.e(TAG, "Failed to update progress notification for ${getFileNameFromUrl(url)}", e)
                 }
                 try {
                     sendDownloadUpdate(url, success, completedCount >= urls.size, fromSync)
                 } catch (e: Exception) {
-                    Log.e(TAG, "Failed to send download update for $url", e)
+                    Log.e(TAG, "Failed to send download update for ${getFileNameFromUrl(url)}", e)
                 }
             }
 
@@ -96,7 +96,7 @@ class DownloadWorker @AssistedInject constructor(
             try {
                 resourcesRepository.markResourceOfflineByUrl(url)
             } catch (e: Exception) {
-                Log.e(TAG, "Failed to mark existing resource offline: $url", e)
+                Log.e(TAG, "Failed to mark existing resource offline: ${UrlUtils.redactForLog(url)}", e)
             }
             return true
         }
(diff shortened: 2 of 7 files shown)


dogi:
**Review — rating: 89/100**

This is a genuine credential-leak fix, and the exposure is worse than "logcat is app-private" suggests.

The URLs being logged here are the CouchDB ones — `UrlUtils.getUrl()` → `dbUrl(spm())`, built as `scheme://user:password@host:port` (the same shape `ProcessUserDataActivity` and `ServerConfigUtils` construct). So `Log.e(TAG, "Failed to download file: $url")` in `DownloadWorker` and `Log.e(TAG, "Exception during bulk upload to $url")` in `BulkDocsUploader` were writing satellite credentials into the log. And this app doesn't just leave them there: `CrashLogStore` captures failures and `UploadConfigs` ships them to the server as `ApkLog` documents, so a leaked line can travel off-device.

`redactForLog` rebuilds `scheme://host[:port][path][?query][#fragment]` from a parsed `Uri`, dropping the userinfo — the right approach, since it's structural rather than a regex over credential shapes. Failing closed to `<unparseable url>` on null, blank or unparseable input is the correct default for a redactor.

I checked whether the sweep is complete rather than assuming it. The other places interpolating a URL into a log — `SubmissionsUploader` lines 24/34/48 — use `sharedPrefManager.getServerUrl()` and the `ServerUrlMapper` values derived from it, which is the plain `SERVER_URL` preference, not the credential-bearing `COUCHDB_URL`. So those aren't leaks and correctly weren't touched. The scoping here is deliberate, not partial.

Using `getFileNameFromUrl(url)` at the sites where the file is what matters, and `redactForLog` where the endpoint matters, is a nice distinction — the download-failure logs stay just as diagnosable while carrying less.

**Nit:** the query string is preserved. Correct for this codebase (credentials live in userinfo), but if a token ever shows up as a query param it would pass straight through. Worth a comment on the function saying userinfo is the only thing stripped.

**Nit:** `android.net.Uri.parse` is fully qualified inline while the rest of the file uses imports.

**Verdict:** Mergeable.

---

Okuro3499 (dismissed):
```
Problem: storedAlt (sharedPrefManager.getProcessedAlternativeUrl()) carries the embedded satellite:PIN@host userinfo (confirmed via ServerConfigUtils.saveAlternativeUrl, which  
    is the code path that populates this pref). It's assigned to resolvedAltBase and then logged raw in two places:                                                                  
    - Line 285: Log.d(TAG, "initDownload: no hardcoded mapping for $primaryBase — using stored alternative $resolvedAltBase")                                                        
    - Line 298: Log.d(TAG, "initDownload: switching $fileName — primary=$resolvedPrimaryBase → alternative=$resolvedAltBase")                                                        
                                                                                                                                                                                     
    This is the same credential leak the PR set out to fix — in the same file, a few lines from the line that was fixed (line 251, Log.d(TAG, "initDownload: retrying with           
    ${UrlUtils.redactForLog(altUrl)}")).
```

Okuro3499 (dismissed):
```
Problem: storedAlt (sharedPrefManager.getProcessedAlternativeUrl()) carries the embedded satellite:PIN@host userinfo (confirmed via ServerConfigUtils.saveAlternativeUrl, which  
    is the code path that populates this pref). It's assigned to resolvedAltBase and then logged raw in two places:                                                                  
    - Line 285: Log.d(TAG, "initDownload: no hardcoded mapping for $primaryBase — using stored alternative $resolvedAltBase")                                                        
    - Line 298: Log.d(TAG, "initDownload: switching $fileName — primary=$resolvedPrimaryBase → alternative=$resolvedAltBase")                                                        
                                                                                                                                                                                     
    This is the same credential leak the PR set out to fix — in the same file, a few lines from the line that was fixed (line 251, Log.d(TAG, "initDownload: retrying with           
    ${UrlUtils.redactForLog(altUrl)}")).
```

dogi:
**Re-review — Rating: 87/100**

A genuine security fix. `CLAUDE.md` is explicit that the `PLANET_*_PIN` values are real CouchDB `satellite` credentials, and `getCouchdbUrl()` embeds them as `satellite:PIN@host` — so every `Log.e(TAG, "…: $url")` on a resource or `_bulk_docs` URL was writing working credentials into logcat, readable by anything with log access on the device.

**The redaction targets are correctly chosen, which I checked rather than assumed**

What struck me first was an apparent inconsistency: in `resolveAlternativeUrl` the alternative base is redacted while `$primaryBase` and `$resolvedPrimaryBase` are logged raw, sometimes on the same line. That looked like a half-finished pass. It isn't. `ServerUrlMapper.extractBaseUrl` rebuilds its result as `"$scheme://$host"` (plus a non-default port) from parsed URI components — userinfo is dropped by construction, so `extractedBaseUrl` cannot carry credentials. Redacting it would add noise without adding safety.

The value that *can* carry credentials on that path is `sharedPrefManager.getProcessedAlternativeUrl()`, and that is the one the PR redacts. Same story in `DownloadWorker` and `BulkDocsUploader`: the redacted `url` values are the ones built from `getCouchdbUrl()`. I also swept the remaining raw-URL log sites — `SubmissionsUploader` logs `getServerUrl()` and `mapping.primaryUrl`, both derived from the plain server URL rather than the credentialed CouchDB one. So the scope looks deliberate and complete for this class of leak.

**`redactForLog` fails in the safe direction**

Any input it can't resolve to a scheme and host returns `"<unparseable url>"` rather than the original string. For a redaction helper that's the correct default — a malformed-but-credential-bearing URL degrades to a placeholder instead of leaking. The `catch` is belt-and-braces since `Uri.parse` is lenient and won't throw, but it costs nothing. The five tests cover the standard CouchDB shape, https-with-port, a URL with no userinfo, empty/null/malformed, and percent-encoded `@`/`:` inside the password — that last one is the case a naive `substringAfter('@')` implementation would get wrong, so it's a good one to have pinned.

**One substantive suggestion**

The function is named for its purpose ("redact") but its mechanism is "rebuild from components, dropping userinfo" — it reassembles `query` and `fragment` verbatim. That's fine for today's URLs, where the secret lives only in userinfo. It stops being fine the moment anything appends `?access_token=…` or a signed URL parameter, and the failure would be silent: the function's name would still promise safety it no longer delivers.

Worth either narrowing the name, or dropping the query string entirely for log purposes (`"$scheme://$host$portStr$path"` — the path is the useful part when debugging a download), or leaving a comment stating the assumption so the next person adding a query parameter has to think about it.

**Two nits:** `android.net.Uri` is fully qualified inline where the rest of `UrlUtils` uses imports; and `DownloadWorker` now treats the same `url` variable two ways in one file — some sites became `getFileNameFromUrl(url)` and others `UrlUtils.redactForLog(url)`. Both are safe, but picking one per context would read more predictably.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17150, opened 2026-09-10
courses: smoother submissions repository dao querying (fixes #17145)

### LINKED ISSUE #17145: observePendingSurveys silently returns nothing when the dashboard has no user id

`data/room/dao/SubmissionDao.kt:19` takes a **nullable** user id and compares it with `=`:

```kotlin
@Query("SELECT * FROM submissions WHERE userId = :userId AND LOWER(status) = 'pending' AND type = 'survey'")
fun observePendingSurveys(userId: String?): Flow<List<Submission>>
```

In SQL, `= NULL` is never true, so a null `userId` makes this Flow emit an empty list forever — no error, no crash, just a signal that is permanently dead.

**Null is reachable.** The chain is three hops, all in the tree:

1. `ui/dashboard/DashboardActivity.kt:592` — `collectWhenStarted(dashboardViewModel.dashboardDataFlow(user?.id))`, where `user` is nullable.
2. `ui/dashboard/DashboardViewModel.kt:242-248` — `dashboardDataFlow(userId: String?)` passes it straight through:
   ```kotlin
   return merge(
       resourcesRepository.getRecentResources(userId ?: "").map {},
       resourcesRepository.getPendingDownloads(userId ?: "").map {},
       submissionsRepository.getPendingSurveysFlow(userId).map {},   // <-- no ?: ""
       teamsRepository.getTasksFlow(userId).map {}
   )
   ```
3. `repository/SubmissionsRepositoryImpl.kt:80-82` — `getPendingSurveysFlow(userId: String?)` hands it to the DAO unchanged.

Note the shape of the bug: the two lines **immediately above** already defend with `userId ?: ""`, and this one does not. The dashboard's pending-survey refresh is the branch that goes quiet.

`SubmissionDao` itself shows the house style — `countByUserParentAndType(userId: String?)` at `:23` and `getByParentUserAndStatus(...)` at `:25` use `IS :userId`, and `getByParentUserAndStatus` handles its nullable `status` as `(:status IS NULL OR status = :status)`. `CLAUDE.md` states the rule outright: use `IS` (not `=`) when a null argument should match NULL rows.

### Do

Push the decision to the boundary that already makes it, rather than inventing a new meaning for null in SQL:

1. `SubmissionDao.kt:19` — change the parameter to non-null: `fun observePendingSurveys(userId: String): Flow<List<Submission>>`. Leave the query text exactly as it is (`userId = :userId`, `LOWER(status) = 'pending'`, `type = 'survey'`).
2. `repository/SubmissionsRepository.kt:18` and `SubmissionsRepositoryImpl.kt:80-82` — keep `getPendingSurveysFlow(userId: String?)`'s signature (other callers may rely on it) and resolve the null there: `submissionDao.observePendingSurveys(userId ?: "")`.
3. Nothing changes in `DashboardViewModel` — after step 2 all four `merge` branches treat a null user identically.

### Constraints

- **Do not switch the query to `IS :userId`.** That would make a null user match rows whose `userId` is NULL, which is a new behaviour nobody asked for; `?: ""` keeps "no user ⇒ no pending surveys" while making it explicit.
- Behaviour for every non-null user id must be **bit-identical** — `userId = :userId` with a real id is unchanged.
- Do not touch the sibling queries in this DAO: `getPendingSurveys` (`:17`, already non-null), `countPendingSurveys` (`:18`), `getUniquePendingSurveyCandidates` (`:20`), `countByUserParentAndType` (`:23`), `countCompletedByUserAndExamId` (`:24`), `getByParentUserAndStatus` (`:25`) — the `IS` usages there are correct.
- Do not add `?: ""` inside `DashboardViewModel`; the point is that the repository resolves it once.

### Tests

Add to `app/src/test/java/org/ole/planet/myplanet/repository/SubmissionsRepositoryImplTest.kt` (MockK), keeping existing cases passing: `getPendingSurveysFlow(null)` calls `observePendingSurveys("")` — assert on the argument, since that is the whole fix — and `getPendingSurveysFlow("user1")` still calls `observePendingSurveys("user1")`. If a `SubmissionDao` Room test exists, a case showing a pending survey for `"user1"` is returned for `"user1"` and not for `""` is worth having.

### Out of scope — checked and deliberately excluded

A sweep of all 30 DAO files found 17 nullable-parameter-with-`=` candidates; **15 are correct** and must not be "fixed":

- `UPDATE … SET x = :param` assignments, not comparisons: `HealthExaminationDao.markUploaded` (`:29`), `SubmissionDao.markUploaded` (`:44`), `TeamTaskDao.markUploaded` (`:50`), `RetryDao.recordFailedAttempt` (`:55`).
- Explicit null handling already present: `MyLifeDao` `:12`/`:15`/`:18` (`(:userId IS NULL AND …) OR (:userId IS NOT NULL AND userId = :userId)`), `PersonalDao.countByTitle` (`:15`), `TagDao.getParentTags` (`:12`), `TeamDao.getEligibleNextLeaderCandidates` (`:21`), `TeamDao.teamNameExists` (`:49`), `SubmissionDao.getByParentUserAndStatus` (`:25`).

`AnswerDao.getBySubmissionAndQuestion` (`:12`) also takes a nullable `questionId` compared with `=`, but its only caller (`SubmissionsRepositoryImpl.kt:511`) is guarded by `questionId.isNotBlank()`, so null is unreachable — tightening that parameter is cosmetic and is **not** part of this task.

Do not touch `AppDatabase` or any Room schema — no entity, column or index changes, so no version bump. Do not add or change any log line.

dogi:
Updates SubmissionDao.observePendingSurveys to accept a non-null userId: String parameter and resolves null to empty string at the SubmissionsRepositoryImpl boundary.

Fixes #17145

---
*PR created automatically by Jules for task [14217919261731330172](https://jules.google.com/task/14217919261731330172) started by @dogi*

## Summary by CodeRabbit

* **New Features**
  * Added consistent confirmation dialogs across course, resource, health, storage, team, and voice workflows.
  * Added reliable library-item resolution and improved resource reload behavior after additions.
  * Added member visit statistics for member details.
  * Added localized dictionary parsing error messages in Arabic, Spanish, French, Nepali, and Somali.

* **Improvements**
  * Improved course filtering and question-count loading.
  * Prevented duplicate data requests and refreshed course progress more reliably.
  * Improved PDF rendering responsiveness and theme-aware arrow colors.

* **Release**
  * Updated the app version to 0.70.77.


dogi:
**Review — rating: 78/100**

I can't reproduce the bug in the title, and I'd like to resolve that before this lands — because if there really is a dead flow on the dashboard, this doesn't fix it.

The change swaps a `NULL` bind for an empty-string bind against `WHERE userId = :userId`. In SQL, `userId = NULL` evaluates to NULL (never true) and `userId = ''` is false for every real user id — **both match zero rows**. Room emits the query result either way and re-emits on `submissions` invalidation, so the `Flow` isn't dead in either version; it emits an empty list. Your own test agrees: it still asserts `result.isEmpty()` and only adds `verify { observePendingSurveys("") }` to pin the new argument. Nothing about the observable behaviour moved.

So one of two things is true, and they need different fixes:

- **There is no bug**, and this is a type-safety tightening — a genuinely reasonable one, since `= :userId` can never match NULL and accepting `String?` advertised a capability the SQL doesn't have. In that case please retitle; "fix … dead flow" will send the next person looking for a regression that was never there.
- **There is a real symptom** — the dashboard showing no pending surveys when it should — in which case the cause is upstream of this DAO (what `DashboardViewModel:246` passes as `userId`, or how that id is resolved), and this change won't move it.

One thing to check before either path: are there `submissions` rows with `userId = ''` rather than NULL? If so this isn't behaviour-neutral — those rows would newly match when no user is resolved, surfacing another account's pending surveys on a logged-out dashboard. I couldn't find evidence of empty-string user ids being written, but `ProgressRepositoryImpl` explicitly filters `userId.isNotEmpty()`, which suggests they're at least considered possible somewhere in the data.

The non-null signature itself is right, and consistent with the sibling `getPendingSurveys(userId: String)` right above it.

**Verdict:** Needs another round — safe change, but the description claims a fix I can't substantiate, and the underlying symptom (if real) is elsewhere.

---

dogi:

First, a correction to how this PR reads from the outside. `git diff master...pr17150` shows 63 files and 735 insertions, because commit `de70fee` replays master's own work onto the branch under a commit message duplicated from `b18cb5c`. The actual authored change, measured against master's tip, is:

```
 SubmissionDao.kt                  | 2 +-
 SubmissionsRepositoryImpl.kt      | 2 +-
 SubmissionsRepositoryImplTest.kt  | 3 ++-
 3 files changed, 4 insertions(+), 3 deletions(-)
```

Four lines. The `enormous` label is an artifact of that replayed commit, not of the change. Worth knowing before anyone reads the diff and reacts to its apparent size. On the upside, this is the only PR in the current review queue that is actually current with master — the other eleven are all behind.

**On the change itself**

```kotlin
- fun observePendingSurveys(userId: String?): Flow<List<Submission>>
+ fun observePendingSurveys(userId: String): Flow<List<Submission>>
```
```kotlin
- return submissionDao.observePendingSurveys(userId)
+ return submissionDao.observePendingSurveys(userId ?: "")
```

The stated justification doesn't hold. Both commit messages describe the old behaviour as an "unreactive null SQL comparison flow" and a "dead flow" — but it was neither. Room binds the null, `WHERE userId = NULL` evaluates to NULL so no row matches, and the returned Flow is still registered with Room's invalidation tracker: it emits an empty list and re-emits on every `submissions` table change, exactly like any other Flow. Nothing about it is dead or unreactive. The observable behaviour before and after is the same empty list.

So the change isn't a bug fix. It's a type-signature tightening, which is a real if modest improvement — making the parameter non-null forces the caller to make the null decision explicitly instead of quietly handing SQL a value that can never match.

**The `""` fallback is the part I'd change**

`userId ?: ""` swaps "matches nothing, ever" for "matches rows whose `userId` is the empty string". Those are not the same predicate. An empty-string `userId` is plausible in this codebase — sync data arrives from CouchDB and `JsonUtils.getString` returns `""` for absent fields — so a malformed submission row would now surface in the pending-surveys flow for a dashboard that has no user at all.

If the intent is "no user means no pending surveys", say that directly:

```kotlin
override fun getPendingSurveysFlow(userId: String?): Flow<List<Submission>> {
    if (userId.isNullOrEmpty()) return flowOf(emptyList())
    return submissionDao.observePendingSurveys(userId)
}
```

That keeps the non-null DAO signature you're adding, states the intent in the repository where it belongs, and can't accidentally match anything. It also costs one fewer Room query in the no-user case.

**Also worth fixing:** `de70fee` carries the same commit message as `b18cb5c` while containing entirely different content. If that lands as-is, the history reads as though a 61-file change was a four-line survey fix. A squash-merge flattens it, but it's worth cleaning up so the branch is honest about what it contains.

**Verdict:** Needs a change — prefer `flowOf(emptyList())` over the empty-string sentinel, and correct the PR description, which currently claims a fix that wasn't needed.

---

dogi:

Both points addressed.

```kotlin
override fun getPendingSurveysFlow(userId: String?): Flow<List<Submission>> {
    if (userId.isNullOrEmpty()) return flowOf(emptyList())
    return submissionDao.observePendingSurveys(userId)
}
```

The empty-string sentinel is gone. "No user means no pending surveys" is now stated in the repository rather than encoded as a SQL value that happens not to match, so a row with `userId = ""` — plausible given `JsonUtils.getString` returns `""` for absent CouchDB fields — can no longer slip into the flow. `isNullOrEmpty` covers both the null and the empty case, and it also smart-casts `userId` to non-null for the DAO call, so the tightened signature is satisfied without a cast or an elvis.

The test is better than what I asked for:

```kotlin
verify(exactly = 0) { submissionDao.observePendingSurveys(any()) }
```

Asserting the DAO is *never reached* is what pins the short-circuit. A test that only checked the result was empty would still pass if someone reverted to passing `""` through to Room, since that also returns empty in the fixture. This one wouldn't.

The commit message was also rewritten — "Tighten observePendingSurveys signature and avoid null query" describes what the change actually does, rather than claiming a dead flow that wasn't dead. And the earlier `de70fee` replay is no longer the head, so the diff now reads as the small change it is.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17153, opened 2026-09-10
all: smoother user repository searching (fixes #17151)

### LINKED ISSUE #17151: Escape LIKE wildcards in the user search — typing % in the health patient search matches every user

`data/room/dao/UserDao.kt:17` interpolates a free-text search string straight into three `LIKE` patterns with **no `ESCAPE` clause and no wildcard escaping**:

```kotlin
@Query("SELECT * FROM users WHERE name LIKE '%' || :query || '%' OR firstName LIKE '%' || :query || '%' OR lastName LIKE '%' || :query || '%'")
suspend fun search(query: String): List<UserEntity>
```

The string is typed by a person. Full chain, every hop read:

1. `ui/health/MyHealthFragment.kt:323` — the patient search box: `.onEach { query -> viewModel.searchPatients(query?.toString() ?: "", "joinDate", true) }`
2. `ui/health/HealthViewModel.kt:61-73` — `searchPatients` → `healthRepository.searchPatients(query, sortBy, descending)`
3. `repository/HealthRepositoryImpl.kt:154-160` — blank goes to `getUsersSortedBy`, non-blank to `userRepository.get().searchUsers(query, sortField, descending)`
4. `repository/UserRepositoryImpl.kt:169-172` — `val users = userDao.search(query)`, verbatim, no escaping
5. `data/room/dao/UserDao.kt:17` — the query above

So in SQLite `LIKE`, `%` in the typed text matches any run of characters and `_` matches any single character. Typing `%` into the health patient search returns **every user in the table**; typing `a_b` matches `axb`. A screen whose whole job is narrowing to one patient silently widens instead. There is no crash — without an `ESCAPE` clause a lone `\` is a literal backslash in SQLite — so this is a wrong-results bug, not an exception.

**This codebase already has the fix as a settled convention, in four places.** Each escapes `\` first, then `%` and `_`, and pairs the pattern with `ESCAPE '\'`:

```kotlin
// repository/ResourcesRepositoryImpl.kt:62-68  (userIdPattern)
// repository/CoursesRepositoryImpl.kt:94-100   (userIdPattern)
// repository/VoicesRepositoryImpl.kt:99-105    (teamIdPattern)
val escaped = userId
    .replace("\\", "\\\\")
    .replace("%", "\\%")
    .replace("_", "\\_")
```

```kotlin
// repository/ResourcesRepositoryImpl.kt:107-112 — the closest analogue: free-text title search
val escapedToken = token
    .replace("\\", "\\\\")
    .replace("%", "\\%")
    .replace("_", "\\_")
queryBuilder.append(" AND titleNormal LIKE ? ESCAPE '\\'")
bindArgs.add("%${escapedToken}%")
```

and the DAO side carries `ESCAPE '\'` throughout — `data/room/dao/CourseDao.kt:17,20`, `data/room/dao/MyLibraryDao.kt:103,106,109,114,121,128,133,139,146,165`, `data/room/dao/NewsDao.kt:33,36,80`. `MyLibraryDao.kt:16-18` states the convention in a file comment. `UserDao.search` is the one free-text `LIKE` in the codebase that does not follow it.

### Do

Two files.

**1. `data/room/dao/UserDao.kt:17`** — take a pre-built pattern and add the escape clause, exactly as `CourseDao.kt:17` and `MyLibraryDao.kt:103` do:

```kotlin
@Query("SELECT * FROM users WHERE name LIKE :namePattern ESCAPE '\\' OR firstName LIKE :namePattern ESCAPE '\\' OR lastName LIKE :namePattern ESCAPE '\\'")
suspend fun search(namePattern: String): List<UserEntity>
```

(`ESCAPE '\\'` in Kotlin source is `ESCAPE '\'` in SQL — that is how the existing DAOs spell it.)

**2. `repository/UserRepositoryImpl.kt:169-172`** — build the pattern in the repository, where every other pattern in this codebase is built:

```kotlin
private fun searchPattern(query: String): String {
    val escaped = query
        .replace("\\", "\\\\")
        .replace("%", "\\%")
        .replace("_", "\\_")
    return "%$escaped%"
}

override suspend fun searchUsers(query: String, sortField: String, descending: Boolean): List<UserEntity> {
    val users = userDao.search(searchPattern(query))
    return sortUsers(users, sortField, descending)
}
```

### Constraints

- **Escape the backslash FIRST**, then `%`, then `_`. Any other order double-escapes the backslashes the later replacements insert and corrupts the pattern. All four existing sites do it in this order — match them.
- Keep the four-line escape block **local to `UserRepositoryImpl`**. Do **not** extract a shared helper and do **not** touch `ResourcesRepositoryImpl`, `CoursesRepositoryImpl` or `VoicesRepositoryImpl`. Deduplicating the block across four files is a separate change with its own review; those files have open PRs against them.
- Do not change any other query in `UserDao.kt`, and do not change the `UserRepository.searchUsers` interface signature (`repository/UserRepository.kt:32`) — only the DAO parameter name and the `Impl` body move.
- Do not add a blank-query guard. `HealthRepositoryImpl.kt:155` already routes `query.isBlank()` to `getUsersSortedBy`, so `searchUsers` is never called with a blank string; adding a second guard is dead code.
- A query containing no `%`, `_` or `\` must produce exactly the results it produces today. `"john"` → `"%john%"`, unchanged behaviour.
- Any existing test that stubs or verifies `userDao.search(...)` must be updated to the pattern form (`"%john%"`), not left matching the bare query.

### Tests

In `app/src/test/java/org/ole/planet/myplanet/repository/UserRepositoryImplTest.kt`:

1. `searchUsers escapes LIKE wildcards before querying` — call `searchUsers("100%_a\\b", "joinDate", true)`, capture the argument handed to `userDao.search` with a MockK `slot`, and assert it is exactly `"%100\\%\\_a\\\\b%"` (i.e. the SQL pattern `%100\%\_a\\b%`). This is the test that fails on today's code.
2. `searchUsers leaves a plain query unchanged apart from the surrounding wildcards` — `"john"` → `"%john%"`.

And a Robolectric DAO test proving the SQL half, in `app/src/test/java/org/ole/planet/myplanet/data/room/dao/UserDaoTest.kt` (create it if absent; copy the in-memory-database setup from an existing DAO test in that package):

3. Insert users named `50%off`, `a_b` and `john`. `search("%\\%%")` must return only `50%off`; `search("%a\\_b%")` must return only `a_b`. Without the `ESCAPE` clause both return all three, so this test also fails on today's code.

### Out of scope — the other `LIKE` sites are correct, leave every one of them alone

- `data/room/dao/CertificationDao.kt:12` (`courseIds LIKE '%' || :courseId || '%'`) and `data/room/dao/SubmissionDao.kt:24,31` (`parentId LIKE '%' || :examId || '%'`, `parentId LIKE '%' || :parentIdFragment || '%'`) interpolate **CouchDB document ids**, not typed text. Those ids are server-generated and contain no `%` or `_`. Changing them means changing every caller for no behaviour gain.
- `data/room/dao/AchievementDao.kt:19`, `data/room/dao/CourseProgressDao.kt:28`, `data/room/dao/RatingDao.kt:36` use the literal prefix `'guest%'` — the `%` is meant as a wildcard and there is no parameter to escape.
- `data/room/dao/NewsDao.kt:63,70` match literal JSON fragments (`'%"section":"community"%'`) — same, no parameter.

No open PR touches `data/room/dao/UserDao.kt` or `repository/UserRepositoryImpl.kt`; #17063 restructures `UserInformationFragment` and adds a ViewModel but leaves both of those files untouched, and #17044 is confined to `MyHealthFragment` and `HealthExaminationAdapter`. Keep the PR to the two source files plus the two test files.

dogi:
Escapes backslash, percent, and underscore characters in free-text user search queries in UserRepositoryImpl before querying UserDao, and adds ESCAPE '\' clauses to UserDao.search @Query. Also includes UserRepositoryImplTest and UserDaoTest tests verifying search wildcard escaping.

Fixes #17151

---
*PR created automatically by Jules for task [18157886742544623293](https://jules.google.com/task/18157886742544623293) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/data/room/dao/UserDao.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/data/room/dao/UserDao.kt
@@ -14,7 +14,8 @@ interface UserDao {
     @Query("SELECT * FROM users") suspend fun getAll(): List<UserEntity>
     @Query("SELECT * FROM users WHERE name = :name LIMIT 1") suspend fun getByName(name: String): UserEntity?
     @Query("SELECT * FROM users WHERE name = :name COLLATE NOCASE LIMIT 1") suspend fun getByNameIgnoreCase(name: String): UserEntity?
-    @Query("SELECT * FROM users WHERE name LIKE '%' || :query || '%' OR firstName LIKE '%' || :query || '%' OR lastName LIKE '%' || :query || '%'") suspend fun search(query: String): List<UserEntity>
+    @Query("SELECT * FROM users WHERE name LIKE :namePattern ESCAPE '\\' OR firstName LIKE :namePattern ESCAPE '\\' OR lastName LIKE :namePattern ESCAPE '\\'")
+    suspend fun search(namePattern: String): List<UserEntity>
     @Query("SELECT COUNT(*) FROM users") suspend fun count(): Int
     @Query("DELETE FROM users WHERE id = :id") suspend fun deleteById(id: String): Int
     @Query("DELETE FROM users WHERE id IN (:ids)") suspend fun deleteByIds(ids: List<String>): Int
--- a/app/src/main/java/org/ole/planet/myplanet/repository/UserRepositoryImpl.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/UserRepositoryImpl.kt
@@ -166,8 +166,16 @@ class UserRepositoryImpl @Inject constructor(
         return sortUsers(getAllUsers(), fieldName, descending)
     }
 
+    private fun searchPattern(query: String): String {
+        val escaped = query
+            .replace("\\", "\\\\")
+            .replace("%", "\\%")
+            .replace("_", "\\_")
+        return "%$escaped%"
+    }
+
     override suspend fun searchUsers(query: String, sortField: String, descending: Boolean): List<UserEntity> {
-        val users = userDao.search(query)
+        val users = userDao.search(searchPattern(query))
         return sortUsers(users, sortField, descending)
     }
 
--- a/app/src/test/java/org/ole/planet/myplanet/data/room/dao/UserDaoTest.kt
+++ b/app/src/test/java/org/ole/planet/myplanet/data/room/dao/UserDaoTest.kt
@@ -115,6 +115,21 @@ class UserDaoTest {
     }
 
     @Test
+    fun search_escapesWildcardsCorrectly() = runBlocking {
+        userDao.upsert(createUser("u1", "r1", "50%off"))
+        userDao.upsert(createUser("u2", "r2", "a_b"))
+        userDao.upsert(createUser("u3", "r3", "john"))
+
+        val percentResults = userDao.search("%\\%%")
+        assertEquals(1, percentResults.size)
+        assertEquals("50%off", percentResults[0].name)
+
+        val underscoreResults = userDao.search("%a\\_b%")
+        assertEquals(1, underscoreResults.size)
+        assertEquals("a_b", underscoreResults[0].name)
+    }
+
+    @Test
     fun getDuplicateUsers_groupsByNullNamesCorrectly() = runBlocking {
         userDao.upsert(createUser("id1", "r1", "John"))
         userDao.upsert(createUser("id2", "r2", "John")) // Duplicate of John
--- a/app/src/test/java/org/ole/planet/myplanet/repository/UserRepositoryImplTest.kt
+++ b/app/src/test/java/org/ole/planet/myplanet/repository/UserRepositoryImplTest.kt
@@ -371,6 +371,28 @@ class UserRepositoryImplTest {
     }
 
     @Test
+    fun `searchUsers escapes LIKE wildcards before querying`() = runTest(testDispatcher) {
+        coEvery { userDao.search(any()) } returns emptyList()
+
+        repository.searchUsers("100%_a\\b", "joinDate", true)
+
+        val slot = slot<String>()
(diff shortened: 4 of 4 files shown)


dogi:
**Review — rating: 89/100**

Real bug fix. On master the search term was concatenated straight into a `LIKE` pattern (`name LIKE '%' || :query || '%'`), so any `%` or `_` the user typed was interpreted as a wildcard rather than a literal: searching for `_` matched every user with at least one character in the field, and `%` matched everyone. Members lists with usernames containing underscores — which this codebase clearly has, given fixtures like `john_doe` and `alice_username` — would return nonsense.

The escaping is done in the right order, which is the easy thing to get wrong: `\` first, then `%` and `_`. Reversing that would double-escape the backslashes introduced by the later replacements. Paired with `ESCAPE '\'` on all three `LIKE` clauses, literals now match literally.

Moving pattern construction out of SQL and into `searchPattern()` is a good secondary change — the `'%' || :query || '%'` concatenation hid the fact that the caller was supplying a *pattern* rather than a *term*, and the renamed `namePattern` parameter now says so. The `ESCAPE '\'` idiom also matches what `NewsDao` already does for its `viewIn LIKE :teamPattern` queries, so this isn't introducing a new convention.

To be clear on the risk level: this was never an injection vector — the value was always bound, never interpolated — so it's a correctness and UX bug, not a security one. 37 lines of new tests across the DAO (real database) and the repository (pattern construction), which is the right split.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17158, opened 2026-09-10
courses: smoother dao querying (fixes #17152)

### LINKED ISSUE #17152: CourseDao.getByCourseIds binds its id list three times, so a courses sync page of 334+ crashes

`data/room/dao/CourseDao.kt:13` binds the **same list three times** in one statement:

```kotlin
@Query("SELECT * FROM courses WHERE courseId IN (:courseIds) OR id IN (:courseIds) OR _id IN (:courseIds)")
suspend fun getByCourseIds(courseIds: List<String>): List<MyCourse>
```

Room expands each `IN (:courseIds)` into one host parameter per element, so a call with *n* ids binds **3n** parameters. SQLite's `SQLITE_MAX_VARIABLE_NUMBER` is 999, so this query throws `SQLiteException: too many SQL variables` at **334 ids**, not 1000.

**One caller already knows this** — `repository/CoursesRepositoryImpl.kt:214-216`, `markCoursesAdded`:

```kotlin
val courses = validCourseIds.chunked(300).flatMap { chunk ->
    courseDao.getByCourseIds(chunk)
}.distinctBy { it.id }
```

300 × 3 = 900 < 999, and `distinctBy { it.id }` because one row can match on `courseId` in one chunk and on `_id` in another. (Compare `repository/TeamsRepositoryImpl.kt:412-414` and `:430-432`, which chunk at 500 for the single-binding `TeamDao.getByIds` and dedupe the same way.)

**Three other callers pass the list straight through, unchunked:**

1. `repository/CoursesRepositoryImpl.kt:623-625`, inside `upsertRoomCoursesFromSync`:
   ```kotlin
   val existingCourses = courseDao.getByCourseIds(
       documentList.mapNotNull { JsonUtils.getString("_id", it).takeIf(String::isNotBlank) }
   ).associateBy { it.courseId ?: it.id }
   ```
   `documentList` is one CouchDB sync page. `services/sync/TransactionSyncManager.kt:198-203` sets `pageSize` per table and `courses` falls into the `else -> 1000` branch, so this is called with **up to 1000 ids ⇒ 3000 bound parameters**. It is reached from both `bulkInsertFromSync` (`:612`) and `batchInsertMyCourses` (`:506-508`). Any server whose `courses` database holds 334 or more documents crashes the courses pull.
2. `repository/CoursesRepositoryImpl.kt:175-178`, `getCoursesByIds` — raw list, called from `ui/teams/courses/TeamCoursesViewModel.kt:37` with a team's full course-id list.
3. `repository/TeamsRepositoryImpl.kt:1278-1281`, `getCoursesForSerialization` — raw list, reached from `:91`.

So this is a crash on the sync path, at a threshold four times lower than the list length suggests, in a query one of whose four callers already carries the workaround.

### Do

**One source file: `data/room/dao/CourseDao.kt`.** Move the chunking into the DAO so no caller can forget it, exactly the shape `AnswerDao` uses for the same problem in #17138/PR #17147:

```kotlin
@Query("SELECT * FROM courses WHERE courseId IN (:courseIds) OR id IN (:courseIds) OR _id IN (:courseIds)")
suspend fun getByCourseIdsInternal(courseIds: List<String>): List<MyCourse>

suspend fun getByCourseIds(courseIds: List<String>): List<MyCourse> {
    if (courseIds.isEmpty()) return emptyList()
    return courseIds.chunked(300)
        .flatMap { getByCourseIdsInternal(it) }
        .distinctBy { it.id }
}
```

Keep the `@Query` text byte-for-byte as it is today — only the function it is attached to gets renamed, and the public name `getByCourseIds` keeps its exact signature so all four call sites compile untouched.

### Constraints

- **Chunk at 300, not 900.** The list is bound three times; 900 ids would be 2700 parameters. Write the reason as a comment on the chunk size so nobody "optimizes" it upward: three `IN` clauses over one list, 300 × 3 = 900 < 999.
- **`distinctBy { it.id }` is required, not cosmetic.** A course whose `courseId` is in chunk 1 and whose `_id` is in chunk 2 is returned by both `getByCourseIdsInternal` calls. Today's single-statement query returns it once, so without the dedupe this change would introduce duplicates that the current code does not have. `CoursesRepositoryImpl.kt:623-625` immediately calls `.associateBy { … }` (last-wins, so duplicates are invisible there), but `:177`'s `mapCourses` and `TeamsRepositoryImpl.kt:1281`'s `.map { … }` both propagate them.
- **Do not edit any caller.** Not `CoursesRepositoryImpl`, not `TeamsRepositoryImpl`, not `TeamCoursesViewModel`. In particular **leave `markCoursesAdded`'s existing `chunked(300)` and `distinctBy` in place** — after this change they are redundant but harmless (a 300-id chunk simply becomes a single inner chunk), and removing them would widen the PR into two heavily-contended files. Say in the PR description that the redundancy is deliberate and a follow-up can drop it.
- Do not touch `AppDatabase`, any entity, column or index — no schema change, so no `version` bump.
- Do not change the ordering guarantee: today the query returns rows in SQLite's natural order and no caller sorts; chunk-then-flatMap must not add a sort.
- Do not add or change any log line.

### Tests

Add `app/src/test/java/org/ole/planet/myplanet/data/room/dao/CourseDaoTest.kt` (Robolectric + `Room.inMemoryDatabaseBuilder`; copy the setup from an existing DAO test in that package):

1. `getByCourseIds with 1200 ids returns every match without throwing` — insert 1200 courses, ask for all 1200 ids. On today's code this throws `too many SQL variables`; that is the point of the test.
2. `getByCourseIds does not duplicate a course matched by different columns in different chunks` — the regression guard for the dedupe. Insert one course whose `courseId` and `_id` differ, then build an id list of 400+ entries positioned so that `courseId` lands in the first chunk and `_id` in the second (pad with ids that match nothing). Assert the result has size 1.
3. `getByCourseIds with an empty list returns an empty list and runs no query` — the short-circuit.
4. `getByCourseIds still matches on all three columns` — one course found by `courseId`, one by `id`, one by `_id`, in a single call.

### Out of scope

- The **single-binding** unbounded `IN (:ids)` queries elsewhere in `data/room/dao/` (`ExamDao.kt:11,15,19,29,30`, `SubmissionDao.kt:12,27,32,33,38`, `MyLibraryDao.kt:37,40,43,61`, `NewsDao.kt:18,24,89`, `NotificationDao.kt:39,42,48,60,102`, `TagDao.kt:24,27,30,33`, `TeamTaskDao.kt:17,38,44`, and the rest) are a **different, already-open** class — that is #17138, under review as PR #17147, scoped there to one DAO on purpose so the pattern gets reviewed once. Do not touch any of them here.
- `data/room/dao/UserDao.kt:12` (`id IN (:userIds) OR _id IN (:userIds)`) is the other multi-binding query, but all three of its callers already chunk safely for a 2× multiplier — `repository/EventsRepositoryImpl.kt:77` and `repository/UserRepositoryImpl.kt:109`/`:1080` all use `chunked(400)` (800 < 999). Leave it alone.
- `data/room/dao/CourseProgressDao.kt:25` binds three *different* lists, so its limit is the sum rather than a multiple; its call path is not on a sync page. Not part of this task.

No open PR touches `data/room/dao/CourseDao.kt`. Keep the PR to that one source file plus the one new test file.

dogi:
Chunk course ID lists in CourseDao.getByCourseIds into chunks of 300 to prevent SQLite variable limit crashes (300 * 3 = 900 < 999). Add unit tests in CourseDaoTest.kt covering large lists, cross-chunk deduplication, empty lists, and column matching.

Fixes #17152

---
*PR created automatically by Jules for task [3201346614717213747](https://jules.google.com/task/3201346614717213747) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/data/room/dao/CourseDao.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/data/room/dao/CourseDao.kt
@@ -10,7 +10,16 @@ import org.ole.planet.myplanet.model.MyCourse
 interface CourseDao {
     @Query("SELECT * FROM courses") suspend fun getAll(): List<MyCourse>
     @Query("SELECT * FROM courses WHERE courseId = :courseId OR id = :courseId LIMIT 1") suspend fun getByCourseId(courseId: String): MyCourse?
-    @Query("SELECT * FROM courses WHERE courseId IN (:courseIds) OR id IN (:courseIds) OR _id IN (:courseIds)") suspend fun getByCourseIds(courseIds: List<String>): List<MyCourse>
+    @Query("SELECT * FROM courses WHERE courseId IN (:courseIds) OR id IN (:courseIds) OR _id IN (:courseIds)")
+    suspend fun getByCourseIdsInternal(courseIds: List<String>): List<MyCourse>
+
+    suspend fun getByCourseIds(courseIds: List<String>): List<MyCourse> {
+        if (courseIds.isEmpty()) return emptyList()
+        // Chunk size 300: query uses 3 IN clauses over courseIds, so 300 * 3 = 900 parameters < SQLITE_MAX_VARIABLE_NUMBER (999).
+        return courseIds.chunked(300)
+            .flatMap { getByCourseIdsInternal(it) }
+            .distinctBy { it.id }
+    }
     @Query("SELECT * FROM courses") fun observeAll(): Flow<List<MyCourse>>
     @Query("SELECT * FROM courses WHERE courseId = :courseId OR id = :courseId LIMIT 1") fun observeByCourseId(courseId: String): Flow<MyCourse?>
 
--- a/app/src/test/java/org/ole/planet/myplanet/data/room/dao/CourseDaoTest.kt
+++ b/app/src/test/java/org/ole/planet/myplanet/data/room/dao/CourseDaoTest.kt
@@ -95,4 +95,59 @@ class CourseDaoTest {
         assertEquals(1, resultUser1Percent.size)
         assertTrue(resultUser1Percent.any { it.id == course5.id })
     }
+
+    @Test
+    fun getByCourseIds_withLargeList_returnsAllWithoutThrowing() = runBlocking {
+        val courses = (1..1200).map { i ->
+            MyCourse(
+                id = "id_$i",
+                courseId = "courseId_$i",
+                _id = "_id_$i"
+            )
+        }
+        courseDao.upsertAll(courses)
+
+        val searchIds = (1..1200).map { "id_$it" }
+        val result = courseDao.getByCourseIds(searchIds)
+        assertEquals(1200, result.size)
+    }
+
+    @Test
+    fun getByCourseIds_doesNotDuplicateWhenMatchedByDifferentColumnsInDifferentChunks() = runBlocking {
+        val course = MyCourse(
+            id = "course_pk",
+            courseId = "course_cid",
+            _id = "course_doc_id"
+        )
+        courseDao.upsertAll(listOf(course))
+
+        // Build list of >300 IDs such that "course_cid" is in chunk 1 and "course_doc_id" is in chunk 2
+        val idsChunk1 = listOf("course_cid") + (1..299).map { "dummy_chunk1_$it" }
+        val idsChunk2 = listOf("course_doc_id") + (1..100).map { "dummy_chunk2_$it" }
+        val allIds = idsChunk1 + idsChunk2
+
+        val result = courseDao.getByCourseIds(allIds)
+        assertEquals(1, result.size)
+        assertEquals("course_pk", result[0].id)
+    }
+
+    @Test
+    fun getByCourseIds_withEmptyList_returnsEmptyList() = runBlocking {
+        val result = courseDao.getByCourseIds(emptyList())
+        assertTrue(result.isEmpty())
+    }
+
+    @Test
+    fun getByCourseIds_matchesAcrossAllThreeColumns() = runBlocking {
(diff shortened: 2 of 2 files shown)


dogi:
**Review — rating: 90/100**

The chunk size is the thing that makes this right, and it's the thing a routine version of this patch would get wrong.

`getByCourseIds` binds `courseIds` into **three** `IN` clauses — `courseId IN (…) OR id IN (…) OR _id IN (…)` — so Room emits 3 × N host parameters, not N. Chunking at the 900 used elsewhere in this codebase (`AnswerDao`, `ApkLogDao`, `ProgressRepositoryImpl`) would produce 2,700 parameters and crash exactly as before. 300 × 3 = 900 is correct, and the inline comment spelling out that arithmetic is worth as much as the fix — it's the first thing someone will "helpfully" bump later.

`distinctBy { it.id }` is load-bearing rather than defensive here, which is easy to miss: within a single query a row appears once no matter which of the three columns matched, but across chunks the *same* course row can match chunk A on `courseId` and chunk B on `_id` when those identifiers differ and land in different chunks. Without the dedup, the chunked version would return duplicates the single query never did. So this preserves master's result set exactly, and the empty-list guard avoids emitting `IN ()`.

55 lines of new `CourseDaoTest` against a real database, which is the only way to exercise a chunk boundary meaningfully.

Worth noting this is the third instance of this pattern in flight (#17147 for `AnswerDao`, #17160 for `TeamsRepositoryImpl`) and the author is clearly tracking the per-query parameter multiplier rather than copying a constant — 300 here, 500 there, 900 elsewhere, each matching the clause count.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17160, opened 2026-09-10
teams: smoother repository querying (fixes #17148)

### LINKED ISSUE #17148: Stop loading the whole teams table to look up ids that TeamDao.getByIds already indexes

Three places in `repository/TeamsRepositoryImpl.kt` load **every row of the `teams` table** and then keep the handful whose `_id` is in a known set:

```kotlin
// :133-136, markTeamsUploaded
val teamsToUpdate = teamDao.getAll()
    .filter { it._id in uploadedTeams.keys }

// :1199-1202, inside the sync insert path
val existingTeams = teamDao.getAll()
    .filter { it._id in ids }
    .associateBy { it._id }
    .toMutableMap()

// :1251-1254, bulkInsertFromSync
val existingTeams = teamDao.getAll()
    .filter { it._id in ids }
    .associateBy { it._id }
    .toMutableMap()
```

Everything needed to avoid that already exists:

- **The query exists.** `data/room/dao/TeamDao.kt:11` — `@Query("SELECT * FROM teams WHERE _id IN (:ids)") suspend fun getByIds(ids: List<String>): List<MyTeam>`.
- **It is index-served.** `model/MyTeam.kt:14` declares `indices = [Index("_id"), …]` and `_id` is the `@PrimaryKey`.
- **The chunked call pattern is already used twice in this very file**, at `:412-414` (`getJoinRequestsInfo`) and `:430-432` (`getTeamNamesByIds`):
  ```kotlin
  return ids.chunked(500)
      .flatMap { chunk -> teamDao.getByIds(chunk) }
      .distinctBy { it._id }
  ```

The `teams` table is not small — it holds team documents, membership rows, join requests, reports and resource links, discriminated by `docType` — so `getAll()` pulls thousands of rows to keep at most a page's worth. Two of the three sites are on the **sync** path, and `:1251`'s own comment explains that the surrounding transaction exists precisely because heavy-table sync pages are expensive; loading the full table inside it works against that.

### Do

Replace each `getAll().filter { it._id in <ids> }` with a chunked `getByIds`, matching the file's existing pattern at `:412-414`:

1. **`:133-136` (`markTeamsUploaded`)** — the id set is `uploadedTeams.keys`:
   ```kotlin
   val teamsToUpdate = uploadedTeams.keys.toList().chunked(500)
       .flatMap { chunk -> teamDao.getByIds(chunk) }
       .distinctBy { it._id }
       .map { entity -> entity.apply { _rev = uploadedTeams[_id]; updated = false } }
   ```
   Keep the existing `if (uploadedTeams.isEmpty()) return` guard above it, and keep the `.map { entity.apply { … } }` body exactly as it is.
2. **`:1199-1202`** — `ids` is already in hand from `validDocuments`; chunk it the same way, keep `.associateBy { it._id }.toMutableMap()`, and add `.distinctBy { it._id }` before `associateBy` for parity with the existing pattern.
3. **`:1251-1254` (`bulkInsertFromSync`)** — same shape, `ids` from `syncDocs.map { it.first }`.

Use **500**, the number the two existing call sites use — not 900. Consistency inside this file matters more here, and 500 is comfortably under SQLite's 999-parameter ceiling.

### Constraints

- **Chunking is not optional.** A heavy-table sync page carries on the order of a thousand documents, so an unchunked `IN (:ids)` would exceed SQLite's 999-variable limit and throw `too many SQL variables` — the failure would only show up on real data.
- Behaviour must be identical: the same `MyTeam` rows, keyed the same way. `getByIds` returns only rows that exist, which is exactly what the `filter` produced.
- Add the empty-input guard at each site (`if (ids.isEmpty())` / the existing `uploadedTeams.isEmpty()` return) — `IN ()` is invalid SQL, and the two existing call sites both guard.
- Do **not** change `insertMyTeam`, the surrounding `withTransaction` in `bulkInsertFromSync`, the `toMutableMap()` mutability (`insertMyTeam` writes into that map), or the per-doc `try/catch` at `:1205-1207`.
- Do not change `TeamDao` at all — `getByIds` and its query already exist.
- Leave `teamDao.getAll()` in place for any other caller; this issue only replaces these three id-filtered uses.

### Tests

Extend `app/src/test/java/org/ole/planet/myplanet/repository/TeamsRepositoryImplTest.kt` (MockK), keeping existing cases passing:

- `markTeamsUploaded` with two ids calls `teamDao.getByIds` with exactly those ids and **never** calls `getAll()` — assert `coVerify(exactly = 0) { teamDao.getAll() }`, since not loading the table is the whole point;
- the same assertion for `bulkInsertFromSync`;
- an id list of **1200** is split into three `getByIds` calls of 500/500/200 and every returned row survives;
- an empty id set performs no DAO read at all.

### Out of scope

`CoursesRepositoryImpl.kt:104` and `:303` do `getAll().filter { !it.courseTitle.isNullOrEmpty() }` — a predicate on a column rather than an id lookup, so it needs a new DAO query rather than an existing one. Not part of this task.

Do not touch `AppDatabase` or any Room schema — no entity, column or index changes, so no version bump. Do not add or change any log line.

dogi:
Replaced teamDao.getAll().filter { it._id in <ids> } in markTeamsUploaded, batchInsertMyTeams, and bulkInsertFromSync with chunked teamDao.getByIds(chunk) (chunk size 500) and added empty input guards. Added unit tests to verify getByIds usage, chunking, and empty input handling.

Fixes #17148

---
*PR created automatically by Jules for task [13631849370143478451](https://jules.google.com/task/13631849370143478451) started by @dogi*

## Summary by CodeRabbit

* **Performance**
  * Improved team synchronization and upload operations by processing records in batches, reducing unnecessary data loading.
  * Added handling for large team lists through chunked lookups.

* **Bug Fixes**
  * Empty team batches now complete without unnecessary database operations.
  * Team bulk updates continue to run atomically within a transaction.

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/repository/TeamsRepositoryImpl.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/TeamsRepositoryImpl.kt
@@ -130,8 +130,9 @@ class TeamsRepositoryImpl @Inject constructor(
 
     override suspend fun markTeamsUploaded(uploadedTeams: Map<String, String>) {
         if (uploadedTeams.isEmpty()) return
-        val teamsToUpdate = teamDao.getAll()
-            .filter { it._id in uploadedTeams.keys }
+        val teamsToUpdate = uploadedTeams.keys.toList().chunked(500)
+            .flatMap { chunk -> teamDao.getByIds(chunk) }
+            .distinctBy { it._id }
             .map { entity ->
                 entity.apply {
                     _rev = uploadedTeams[_id]
@@ -1195,9 +1196,11 @@ class TeamsRepositoryImpl @Inject constructor(
                 val id = JsonUtils.getString("_id", doc)
                 id.isNotEmpty() && !id.startsWith("_design")
             }
+            if (validDocuments.isEmpty()) return 0
             val ids = validDocuments.map { JsonUtils.getString("_id", it) }
-            val existingTeams = teamDao.getAll()
-                .filter { it._id in ids }
+            val existingTeams = ids.chunked(500)
+                .flatMap { chunk -> teamDao.getByIds(chunk) }
+                .distinctBy { it._id }
                 .associateBy { it._id }
                 .toMutableMap()
 
@@ -1247,9 +1250,11 @@ class TeamsRepositoryImpl @Inject constructor(
 
     override suspend fun bulkInsertFromSync(jsonArray: JsonArray) {
         val syncDocs = jsonArray.toSyncDocuments()
+        if (syncDocs.isEmpty()) return
         val ids = syncDocs.map { it.first }
-        val existingTeams = teamDao.getAll()
-            .filter { it._id in ids }
+        val existingTeams = ids.chunked(500)
+            .flatMap { chunk -> teamDao.getByIds(chunk) }
+            .distinctBy { it._id }
             .associateBy { it._id }
             .toMutableMap()
         // Wrap the whole batch in a single Room transaction. insertMyTeam upserts one row at a
--- a/app/src/test/java/org/ole/planet/myplanet/repository/TeamsRepositoryImplTest.kt
+++ b/app/src/test/java/org/ole/planet/myplanet/repository/TeamsRepositoryImplTest.kt
@@ -1,6 +1,7 @@
 package org.ole.planet.myplanet.repository
 
 import android.content.SharedPreferences
+import androidx.room.withTransaction
 import com.google.gson.Gson
 import io.mockk.coEvery
 import io.mockk.coVerify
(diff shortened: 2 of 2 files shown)


dogi:
**Review — rating: 89/100**

The title undersells this. The three sites weren't just missing chunking — they were doing `teamDao.getAll().filter { it._id in ids }`, i.e. **loading the entire `teams` table into memory and filtering in Kotlin**. Replacing that with an indexed `_id IN (…)` query is the real win; chunking is what makes the replacement safe.

That matters because all three sites are on sync paths (`markTeamsUploaded`, the bulk-insert reconciliation, `bulkInsertFromSync`), so on a device with a large team set master was materialising every `MyTeam` row three times per sync to keep a handful.

Equivalence checks out: `TeamDao.getByIds` is `SELECT * FROM teams WHERE _id IN (:ids)` (line 11), matching `filter { it._id in ids }` exactly — same column, so no silent switch between `_id` and `id`. The downstream `.associateBy { it._id }.toMutableMap()` and the `markTeamsUploaded` `.map { entity.apply { _rev = uploadedTeams[_id] … } }` are untouched.

Chunking at 500 is right for this query: one `IN` clause, so 500 parameters against the 999 ceiling. Good to see it differ deliberately from #17158's 300 in the same batch — that query has three `IN` clauses, and the author is sizing per clause count rather than reusing a constant.

The two new empty guards (`validDocuments.isEmpty()`, `syncDocs.isEmpty()`) are worth having — they short-circuit before an `IN ()` can be emitted, and `getAll()`-then-filter would previously have done a full table load just to match nothing.

**Nit:** `distinctBy { it._id }` is unnecessary here. With a single `IN` on the primary identifier, each row's `_id` falls in exactly one chunk, so cross-chunk duplicates can't occur — unlike #17158, where the three-column `OR` makes the same dedup load-bearing. Harmless, but it invites the reader to assume duplicates are possible.

79 lines of new tests.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17166, opened 2026-09-11
teams: smoother voices actions date formatting (fixes #17164)

### LINKED ISSUE #17164: Replace the ThreadLocal SimpleDateFormat in VoicesActions with a java.time DateTimeFormatter
### Verified against `origin/master` (`9939519`)

`ui/voices/VoicesActions.kt:33` keeps a formatter alive in a `ThreadLocal` purely because `SimpleDateFormat` is not thread-safe:

```kotlin
object VoicesActions {
    private val dateFormatter = ThreadLocal.withInitial { SimpleDateFormat("MMMM dd, yyyy hh:mm a", Locale.getDefault()) }
```

It has exactly **one** use, at `:225`, inside `showMemberDetails`:

```kotlin
(visitStats.lastVisit?.let { dateFormatter.get()?.format(Date(it)) } ?: "No logout record found"),
```

`java.time.format.DateTimeFormatter` is immutable and thread-safe, so the whole `ThreadLocal` wrapper exists only to work around a class the platform replaced. `minSdk` is **26** (`app/build.gradle:11`), which is exactly the API level where `java.time` became available natively — no desugaring needed, no `@RequiresApi`.

This task was blocked until a few minutes ago: PR #17046 (issue #17023) rewrote this region and merged as `9939519`. Its hunk is gone from master, the formatter and its single call site survived the merge unchanged, and the call site moved from `:214-228` to `:225`. The premise above is re-derived from the post-merge file, not from the old backlog note.

### Do

One file, `ui/voices/VoicesActions.kt`.

1. Replace the field at `:33`:
   ```kotlin
   private val dateFormatter: DateTimeFormatter =
       DateTimeFormatter.ofPattern("MMMM dd, yyyy hh:mm a", Locale.getDefault())
           .withZone(ZoneId.systemDefault())
   ```
2. Replace the call at `:225`:
   ```kotlin
   (visitStats.lastVisit?.let { dateFormatter.format(Instant.ofEpochMilli(it)) } ?: "No logout record found"),
   ```
3. Fix the imports: drop `java.text.SimpleDateFormat` (`:18`) and `java.util.Date` (`:19`); add `java.time.Instant`, `java.time.ZoneId` and `java.time.format.DateTimeFormatter`. **Keep `java.util.Locale` (`:20`)** — the new formatter still takes it.

### Constraints

- **`.withZone(ZoneId.systemDefault())` is mandatory, not optional.** A `DateTimeFormatter` built by `ofPattern` alone has no zone, and formatting an `Instant` with it throws `UnsupportedTemporalTypeException: Unsupported field: YearOfEra` at runtime — the single most common way this migration breaks. `SimpleDateFormat` used the default zone implicitly; `withZone` is what preserves that.
- **Keep passing `Locale.getDefault()` explicitly.** `DateTimeFormatter.ofPattern(pattern)` without a locale resolves `Locale.getDefault(Locale.Category.FORMAT)`, which is not always the same locale `SimpleDateFormat` was given here. Passing it explicitly keeps the output identical.
- **Do not change the pattern string.** `"MMMM dd, yyyy hh:mm a"` stays exactly as it is, including the 12-hour `hh` and the `a` marker.
- **The `?:` fallback must stay bound to `lastVisit` being null, and to nothing else.** Today's `dateFormatter.get()?.format(...)` has a second, accidental null path — `ThreadLocal.get()` is a platform type, so a null formatter would also fall through to `"No logout record found"`. After the change the formatter is a non-null `val` and that path disappears, which is correct. Write it as `visitStats.lastVisit?.let { dateFormatter.format(...) } ?: "No logout record found"` — do not introduce a `?.` on `dateFormatter`, and do not restructure the `let`.
- Do not touch anything else in `VoicesActions.kt`. In particular leave `showMemberDetails`'s signature, its `MembersDetailFragment.newInstance(...)` argument order, and every other argument alone.
- Do not add a `@RequiresApi` annotation and do not enable core library desugaring — `minSdk = 26` already covers `java.time`.
- Do not introduce a shared date-formatting helper or touch `utils/TimeUtils.kt`. Unifying date formatting across the app is a separate, larger task.
- No schema change, no `AppDatabase` version bump, no new or changed log line.

### Tests

`app/src/test/java/org/ole/planet/myplanet/ui/voices/VoicesActionsTest.kt` already exists (PR #17046 added cases to it). Add to it, keeping every existing case passing:

1. `showMemberDetails formats lastVisit in the default zone` — pin the environment so the assertion is deterministic: set a fixed default `TimeZone` and `Locale` (for example `TimeZone.setDefault(TimeZone.getTimeZone("UTC"))` and `Locale.setDefault(Locale.US)`) in `@Before`, restore the originals in `@After`, and assert the formatted string for a known epoch-milli value is exactly what the old `SimpleDateFormat` produced for the same input. Compute the expected string in the test by constructing `SimpleDateFormat("MMMM dd, yyyy hh:mm a", Locale.US)` over `Date(millis)` **or** hard-code the literal — either is fine, but the test must fail if the pattern, locale handling or zone handling changes.
2. `showMemberDetails falls back when lastVisit is null` — assert the argument is exactly `"No logout record found"`.

If asserting through `showMemberDetails` requires more Android scaffolding than the existing tests in that file already set up, it is acceptable to make the formatting a small private-to-internal function and test that directly — but only if the existing test style genuinely cannot reach it. Prefer testing the public path.

### Out of scope

- `repository/SubmissionsRepositoryExporter.kt:34` holds the codebase's other `ThreadLocal` `SimpleDateFormat` (`"yyyy-MM-dd HH:mm"`). Same pattern, different file, different call sites, CSV/PDF output that needs its own equivalence check. Leave it; it is a separate task.
- `utils/TimeUtils.kt` and every other date helper.
- The rest of `ui/voices/`.

No open PR touches `ui/voices/VoicesActions.kt` — the only open PR mentioning it is the draft Flutter port #15226, whose changes live under `flutter/`.

dogi:
Replaced ThreadLocal SimpleDateFormat with java.time.DateTimeFormatter in VoicesActions.kt, updated imports, and added test cases in VoicesActionsTest.kt.

Fixes #17164

---
*PR created automatically by Jules for task [10436094924926631532](https://jules.google.com/task/10436094924926631532) started by @dogi*

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/ui/voices/VoicesActions.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/voices/VoicesActions.kt
@@ -15,8 +15,9 @@ import com.bumptech.glide.Glide
 import com.google.android.material.textfield.TextInputLayout
 import com.google.gson.JsonObject
 import java.io.File
-import java.text.SimpleDateFormat
-import java.util.Date
+import java.time.Instant
+import java.time.ZoneId
+import java.time.format.DateTimeFormatter
 import java.util.Locale
 import org.ole.planet.myplanet.R
 import org.ole.planet.myplanet.callback.OnNewsItemClickListener
@@ -30,7 +31,9 @@ import org.ole.planet.myplanet.utils.JsonUtils
 import org.ole.planet.myplanet.utils.Utilities
 
 object VoicesActions {
-    private val dateFormatter = ThreadLocal.withInitial { SimpleDateFormat("MMMM dd, yyyy hh:mm a", Locale.getDefault()) }
+    private val dateFormatter: DateTimeFormatter =
+        DateTimeFormatter.ofPattern("MMMM dd, yyyy hh:mm a", Locale.getDefault())
+            .withZone(ZoneId.systemDefault())
 
     data class EditDialogComponents(
         val binding: AlertInputBinding,
@@ -222,7 +225,7 @@ object VoicesActions {
             userModel.language.toString(),
             userModel.phoneNumber.toString(),
             visitStats.offlineVisitCount.toString(),
-            (visitStats.lastVisit?.let { dateFormatter.get()?.format(Date(it)) } ?: "No logout record found"),
+            (visitStats.lastVisit?.let { dateFormatter.format(Instant.ofEpochMilli(it)) } ?: "No logout record found"),
             "${userModel.firstName} ${userModel.lastName}",
             userModel.level.toString(),
             userModel.userImage
--- a/app/src/test/java/org/ole/planet/myplanet/ui/voices/VoicesActionsTest.kt
+++ b/app/src/test/java/org/ole/planet/myplanet/ui/voices/VoicesActionsTest.kt
@@ -7,6 +7,9 @@ import io.mockk.coEvery
 import io.mockk.coVerify
 import io.mockk.mockk
 import kotlinx.coroutines.test.runTest
+import java.util.Locale
+import java.util.TimeZone
+import org.junit.After
 import org.junit.Assert.assertEquals
 import org.junit.Assert.assertNotNull
 import org.junit.Assert.assertSame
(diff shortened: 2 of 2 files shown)


dogi:
**Review — rating: 88/100**

The same fix as #17108 (already `ready`), applied to `VoicesActions`. The reasoning carries over unchanged: the `ThreadLocal` existed only because `SimpleDateFormat` is mutable and not thread-safe, and `DateTimeFormatter` is immutable and explicitly thread-safe, so the wrapper stops being necessary rather than being worked around.

Output is preserved. `"MMMM dd, yyyy hh:mm a"` carries the same meaning in both APIs — full month name, day, year, 12-hour clock, minute, am/pm marker — and `Locale.getDefault()` is passed through. The `.withZone(ZoneId.systemDefault())` is required here, not optional: formatting an `Instant` without a zone throws on the zone-dependent fields, so its presence is what makes `Instant.ofEpochMilli(it)` work where `Date(it)` did before.

Worth noting these two PRs together finish the job — grepping `app/src/main` on this branch, the only remaining `SimpleDateFormat` is the one in `SubmissionsRepositoryExporter` that #17108 removes. Once both land there is no `ThreadLocal<SimpleDateFormat>` left in the codebase, which is a better outcome than either alone.

The 64 lines of new `VoicesActionsTest` coverage are welcome given this formats the "last visit" string a user actually reads, and the `?: "No logout record found"` fallback for a null `lastVisit` is unchanged.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17167, opened 2026-09-11
teams: smoother requests members view modelling (fixes #17165)

### LINKED ISSUE #17165: Hoist MembersFragment's team-member load and mutations into the RequestsViewModel it already owns
### Verified against `origin/master` (`8913120`)

`ui/teams/members/MembersFragment.kt` already owns a `RequestsViewModel` — `:29`, `private val requestsViewModel: RequestsViewModel by viewModels()` — and already routes part of its work through it (`respondToRequest` at `:66`, `fetchMembers` at `:79`, and it collects `uiState` at `:81` and `successAction` at `:89`).

But four private functions still reach past it and drive `teamsRepository` (inherited from `BaseTeamFragment:28-29`) directly from `viewLifecycleOwner.lifecycleScope`:

| Function | Lines | Repository calls |
|---|---|---|
| `loadMembers()` | `:95-103` | `getJoinedMembersWithVisitInfo(teamId)` |
| `handleLeaveTeam()` | `:106-123` | `getNextLeaderCandidate`, `updateTeamLeader`, `removeMember` |
| `handleRemoveMember(member)` | `:125-147` | `getNextLeaderCandidate`, `updateTeamLeader`, `removeMember` |
| `handleMakeLeader(userId)` | `:149-160` | `updateTeamLeader` |

Each one opens its own coroutine, calls the repository, catches `Exception` and toasts `"Error …: ${e.message}"` — so database work, its error handling and its result state all live in a view that is destroyed and recreated on every rotation, while the ViewModel built for exactly this screen sits beside it.

**Nothing needs to be widened to fix it.** Every method those four functions call is already on the interface the ViewModel injects, `repository/TeamsMembersRepository`:

- `removeMember(teamId, userId)` — `:12`
- `getJoinedMembersWithVisitInfo(teamId): List<JoinedMemberData>` — `:16`
- `updateTeamLeader(teamId, newLeaderId): Boolean` — `:19`
- `getNextLeaderCandidate(teamId, excludeUserId): UserEntity?` — `:20`

and `RequestsViewModel` already injects both `TeamsMembersRepository` and `UserRepository` (`:26-29`), so it can resolve the current user itself.

This task was blocked on PR #17002 until it merged as `eb718a1`. The line numbers above are re-derived from the post-merge file — #17002 moved `handleLeaveTeam` to the `requireContext().confirmDialog(message = …, onPositive = …)` form imported at `:23`, so the old note's `~104-120` no longer applies.

### Do

Two files: `ui/teams/members/RequestsViewModel.kt` and `ui/teams/members/MembersFragment.kt`.

**1. Add joined-member state to `RequestsViewModel`, alongside the existing request state — not merged into it.**

```kotlin
data class MembersUiState(
    val members: List<JoinedMemberData> = emptyList(),
    val currentUserId: String? = null,
    val isLeader: Boolean = false
)

private val _membersState = MutableStateFlow(MembersUiState())
val membersState: StateFlow<MembersUiState> = _membersState.asStateFlow()

fun loadJoinedMembers(teamId: String) {
    viewModelScope.launch {
        val members = teamsRepository.getJoinedMembersWithVisitInfo(teamId)
        val currentUserId = userRepository.getUserModel()?.id
        _membersState.value = MembersUiState(
            members = members,
            currentUserId = currentUserId,
            isLeader = members.any { it.user.id == currentUserId && it.isLeader }
        )
    }
}
```

Note the `isLeader` rule: it is computed **from the loaded list** (`members.any { it.user.id == currentUserId && it.isLeader }`), exactly as `:99` does today. Do **not** replace it with a `teamsRepository.isTeamLeader(...)` call — that would add a query where the answer is already in hand, and it is a different predicate.

**2. Move the three mutations in, each reporting its outcome as an event.** Add a sealed result type and a `SharedFlow`:

```kotlin
sealed interface MemberActionResult {
    data object LeftTeam : MemberActionResult
    data object MemberRemoved : MemberActionResult
    data object CannotRemoveLastLeader : MemberActionResult
    data object LeaderChanged : MemberActionResult
    data class Failed(val action: MemberAction, val message: String?) : MemberActionResult
}
```

with `leaveTeam(teamId)`, `removeMember(teamId, memberId)` and `makeLeader(teamId, userId)` doing the repository work in `viewModelScope`, emitting the matching result, and calling `loadJoinedMembers(teamId)` on success so the list refreshes from one place.

**3. Reduce the fragment to view work.** `loadMembers()` becomes `requestsViewModel.loadJoinedMembers(teamId)`; the four private functions keep only the dialog, the toasts, `popBackStack()`, `onMemberChangeListener?.onChanged()` and the adapter updates, driven from `collectWhenStarted(requestsViewModel.membersState)` and `collectWhenStarted(requestsViewModel.actionResults)`.

### Constraints

- **`handleRemoveMember`'s three-way branch must survive intact** (`:128-141`). Removing **yourself** when a next-leader candidate exists promotes that candidate, then removes you. Removing **yourself** when `getNextLeaderCandidate` returns `null` shows `R.string.cannot_remove_user` and **removes nobody** — that early `return@launch` at `:136` is the behaviour, not a detail. Removing **someone else** skips the leader logic entirely. Getting this wrong silently deletes the last leader of a team.
- **Do not merge `MembersUiState` into `RequestsUiState`.** They describe different collections — `RequestsUiState.members` is `List<UserEntity>` of *requested* members (`:19-23`), the new one is `List<JoinedMemberData>` of *joined* members. `fetchMembers` (`:36-51`) and `respondToRequest` (`:53-73`) must keep working exactly as they do, including the optimistic update and its rollback.
- **Every error path must stay visible.** Today each function catches `Exception` and toasts `"Error leaving team: …"`, `"Error removing member: …"`, `"Error making leader: …"`. Those strings must still reach the user via the result event. Do not swallow an exception, and do not replace a toast with a silent log.
- **Keep the post-mutation side effects and their order.** After a successful remove the fragment currently refreshes the list, calls `onMemberChangeListener?.onChanged()`, then `requestsViewModel.fetchMembers(teamId)` (`:140-142`) — all three must still happen. After leaving the team it toasts `R.string.left_team` then pops the back stack (`:116-117`). After making a leader it toasts `R.string.leader_selected` then notifies the listener (`:154-156`).
- **Do not touch `BaseTeamFragment`.** Its `teamsRepository`, `user` field and `ensureUserResolved()` stay exactly as they are; other fragments depend on them. The fragment may keep inheriting them even if it no longer calls `teamsRepository` itself.
- **Do not change `TeamsMembersRepository`, `TeamsRepository`, or any repository implementation.** Every method needed is already on the interface — if you find yourself adding one, stop and re-read.
- Do not change `MembersAdapter`, `RequestsAdapter`, `JoinedMemberData`, the layout, or `FragmentCombinedMembersBinding`.
- Keep `BaseRecyclerFragment.showNoData(binding.tvNodata, members.size, "")` (`:101`) in the fragment — it is a view call.
- Use `collectWhenStarted` (already imported at `:24`) for the new flows, matching `:81` and `:89`. Do not hand-roll `repeatOnLifecycle`.

### Tests

Add to `app/src/test/java/org/ole/planet/myplanet/ui/teams/members/RequestsViewModelTest.kt` if it exists, otherwise create it, using MockK + `MainDispatcherRule` + `TestDispatcherProvider` in the style of the other ViewModel tests. Keep any existing cases passing.

1. `loadJoinedMembers publishes members and derives isLeader from the loaded list` — stub three members with the current user flagged leader; assert `membersState` carries all three, the right `currentUserId`, and `isLeader == true`. Then a case where the current user is present but not a leader → `false`. Assert `teamsRepository.isTeamLeader` is **never** called: `coVerify(exactly = 0) { teamsRepository.isTeamLeader(any(), any()) }`.
2. `removeMember promotes the next leader when a user removes themselves` — assert `updateTeamLeader` is called with the candidate's id **before** `removeMember`, then `MemberRemoved` is emitted.
3. `removeMember refuses to remove the last leader` — `getNextLeaderCandidate` returns null and the target is the current user; assert `CannotRemoveLastLeader` is emitted and `coVerify(exactly = 0) { teamsRepository.removeMember(any(), any()) }`. This is the regression guard for the branch above.
4. `removeMember skips the leader logic for another user` — assert `getNextLeaderCandidate` is never called and `removeMember` is.
5. `a repository failure surfaces as Failed with the message` — make one repository call throw and assert the emitted `Failed` carries the exception message rather than being swallowed.
6. `respondToRequest still rolls back its optimistic update on failure` — the existing behaviour, re-asserted so the new state object cannot have broken it.

### Out of scope

- `ui/settings/StorageBreakdownFragment` — the sibling hoist is still blocked on open PR #17000.
- Everything else under `ui/teams/`, including `MembersDetailFragment` and the adapters.
- `BaseTeamFragment` and every other fragment that inherits from it.
- The `printStackTrace` / `!!` clean-ups elsewhere in the tree.

No open PR touches `ui/teams/members/MembersFragment.kt` or `ui/teams/members/RequestsViewModel.kt`.

dogi:
Hoisted MembersFragment team-member loading and mutation logic into RequestsViewModel. Added state flows and action result events, updated MembersFragment, and added comprehensive unit tests.

Fixes #17165

---
*PR created automatically by Jules for task [6383845198477946207](https://jules.google.com/task/6383845198477946207) started by @dogi*

## Summary by CodeRabbit

* **New Features**
  * Team member management now supports leaving a team, removing members, and promoting members to leaders.
  * Leader protections prevent removing the last team leader and support leader reassignment when needed.
  * Member lists and leadership status refresh automatically after actions, with clear success or failure feedback.

* **Bug Fixes**
  * Improved handling of member-loading and team-action failures.

* **Tests**
  * Added coverage for member actions, leader transitions, protections, and error scenarios.

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/ui/teams/members/MembersFragment.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/teams/members/MembersFragment.kt
@@ -90,73 +90,60 @@ class MembersFragment : BaseTeamFragment() {
             onMemberChangeListener?.onChanged()
             loadMembers()
         }
+        collectWhenStarted(requestsViewModel.membersState) { state ->
+            membersAdapter?.setUserId(state.currentUserId)
+            membersAdapter?.updateData(state.members, state.isLeader)
+            BaseRecyclerFragment.showNoData(binding.tvNodata, state.members.size, "")
+        }
+        collectWhenStarted(requestsViewModel.actionResults) { result ->
+            when (result) {
+                MemberActionResult.LeftTeam -> {
+                    Toast.makeText(requireContext(), getString(R.string.left_team), Toast.LENGTH_SHORT).show()
+                    requireActivity().supportFragmentManager.popBackStack()
+                }
+                MemberActionResult.MemberRemoved -> {
+                    onMemberChangeListener?.onChanged()
+                    requestsViewModel.fetchMembers(teamId)
+                }
+                MemberActionResult.CannotRemoveLastLeader -> {
+                    Toast.makeText(requireContext(), R.string.cannot_remove_user, Toast.LENGTH_SHORT).show()
+                }
+                MemberActionResult.LeaderChanged -> {
+                    Toast.makeText(requireContext(), getString(R.string.leader_selected), Toast.LENGTH_SHORT).show()
+                    onMemberChangeListener?.onChanged()
+                }
+                is MemberActionResult.Failed -> {
+                    val actionStr = when (result.action) {
+                        MemberAction.LEAVE_TEAM -> "leaving team"
+                        MemberAction.REMOVE_MEMBER -> "removing member"
+                        MemberAction.MAKE_LEADER -> "making leader"
+                    }
+                    Toast.makeText(requireContext(), "Error $actionStr: ${result.message}", Toast.LENGTH_SHORT).show()
+                }
+            }
+        }
     }
 
     private fun loadMembers() {
-        viewLifecycleOwner.lifecycleScope.launch {
-            val members = teamsRepository.getJoinedMembersWithVisitInfo(teamId)
-            val currentUserId = ensureUserResolved()?.id
-            val isLeader = members.any { it.user.id == currentUserId && it.isLeader }
-            membersAdapter?.setUserId(currentUserId)
-            membersAdapter?.updateData(members, isLeader)
-            BaseRecyclerFragment.showNoData(binding.tvNodata, members.size, "")
-        }
+        requestsViewModel.loadJoinedMembers(teamId)
     }
 
     private fun handleLeaveTeam() {
         requireContext().confirmDialog(
             message = getString(R.string.confirm_exit),
             onPositive = {
-                viewLifecycleOwner.lifecycleScope.launch {
-                    try {
-                        val nextLeader = teamsRepository.getNextLeaderCandidate(teamId, user?.id)
-                        nextLeader?.id?.let { teamsRepository.updateTeamLeader(teamId, it) }
-                        user?.id?.let { teamsRepository.removeMember(teamId, it) }
-                        loadMembers()
-                        Toast.makeText(requireContext(), getString(R.string.left_team), Toast.LENGTH_SHORT).show()
-                        requireActivity().supportFragmentManager.popBackStack()
-                    } catch (e: Exception) {
-                        Toast.makeText(requireContext(), "Error leaving team: ${e.message}", Toast.LENGTH_SHORT).show()
-                    }
-                }
+                requestsViewModel.leaveTeam(teamId)
             }
         )
     }
 
     private fun handleRemoveMember(member: JoinedMemberData) {
         val memberId = member.user.id ?: return
-        viewLifecycleOwner.lifecycleScope.launch {
-            try {
-                if (user?.id == memberId) {
-                    val nextLeader = teamsRepository.getNextLeaderCandidate(teamId, memberId)
-                    if (nextLeader != null) {
-                        nextLeader.id?.let { teamsRepository.updateTeamLeader(teamId, it) }
-                    } else {
-                        Toast.makeText(requireContext(), R.string.cannot_remove_user, Toast.LENGTH_SHORT).show()
-                        return@launch
-                    }
-                }
-                teamsRepository.removeMember(teamId, memberId)
-                loadMembers()
-                onMemberChangeListener?.onChanged()
-                requestsViewModel.fetchMembers(teamId)
-            } catch (e: Exception) {
-                Toast.makeText(requireContext(), "Error removing member: ${e.message}", Toast.LENGTH_SHORT).show()
-            }
-        }
+        requestsViewModel.removeMember(teamId, memberId)
     }
 
     private fun handleMakeLeader(userId: String) {
-        viewLifecycleOwner.lifecycleScope.launch {
-            try {
-                teamsRepository.updateTeamLeader(teamId, userId)
-                loadMembers()
-                Toast.makeText(requireContext(), getString(R.string.leader_selected), Toast.LENGTH_SHORT).show()
-                onMemberChangeListener?.onChanged()
-            } catch (e: Exception) {
-                Toast.makeText(requireContext(), "Error making leader: ${e.message}", Toast.LENGTH_SHORT).show()
-            }
-        }
+        requestsViewModel.makeLeader(teamId, userId)
     }
 
     override fun onNewsItemClick(news: News?) {}
--- a/app/src/main/java/org/ole/planet/myplanet/ui/teams/members/RequestsViewModel.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/teams/members/RequestsViewModel.kt
@@ -8,10 +8,12 @@ import kotlinx.coroutines.async
 import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.flow.MutableSharedFlow
 import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.SharedFlow
 import kotlinx.coroutines.flow.StateFlow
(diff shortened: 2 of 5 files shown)


dogi:
**Review — rating: 86/100**

Hoists `MembersFragment`'s data loading and three mutating actions into `RequestsViewModel`, behind a `MembersUiState` StateFlow and a `MemberActionResult` one-shot flow. 272+/56− across two source files and the ViewModel test.

**Behavioural equivalence — checked, not assumed**

The move that needed verifying is the identity source. The fragment used `ensureUserResolved()?.id` in `loadMembers` and the cached `user?.id` in the two mutators; the ViewModel uses `userRepository.getUserModel()?.id`. `BaseTeamFragment.ensureUserResolved()` is `user ?: userRepository.getUserModel()?.also { user = it }` — the fragment field is a memo over exactly that call, so the two resolve to the same user. Equivalent. ✓

The rest lines up too:
- `removeMember`'s self-removal branch (promote next leader, or refuse if there is none) is preserved including the early return. ✓
- The `MemberRemoved` handler fires `onMemberChangeListener?.onChanged()` and `requestsViewModel.fetchMembers(teamId)` — matching the original, and `fetchMembers` is the join-*requests* list, not a duplicate of the `loadJoinedMembers` the ViewModel already did. ✓
- `when (result)` is exhaustive over the sealed interface, so a future `MemberActionResult` member can't be silently dropped. ✓

**This fixes a real bug in passing**

Worth stating because the PR description doesn't claim it: the old mutators ran in `viewLifecycleOwner.lifecycleScope`. If the view was destroyed mid-flight, `handleRemoveMember` could be cancelled *between* `updateTeamLeader` and `removeMember`, leaving a team with a newly promoted leader and the original member still on it. Running the mutation in `viewModelScope` closes that window. That's the strongest argument for this refactor and it's larger than the tidiness argument.

**One design note on the one-shot flow**

`_actionResults` is `MutableSharedFlow()` — replay 0, no buffer — collected via `collectWhenStarted`. With no subscriber, `emit` drops. So if the fragment is stopped between the action completing and the result arriving, `LeftTeam` is lost and `popBackStack()` never runs: the user ends up sitting on the members screen for a team they've left. `membersState` is a StateFlow so the *list* recovers, but the navigation doesn't.

I'm not treating this as blocking, for two reasons: the old code had a strictly worse version of the same failure (cancellation could lose the mutation itself, not just the notification), and `_successAction` in this same ViewModel already has exactly this shape, so the PR follows local convention rather than inventing a weaker one. If you want to tighten it later, `MutableSharedFlow(replay = 1)` or a `Channel(Channel.BUFFERED)` makes these survive a stop — and it'd be worth doing for `_successAction` at the same time.

**Two nits**

`catch (e: Exception)` in each `viewModelScope` block also catches `CancellationException`, so a ViewModel teardown mid-action turns into a `Failed` emission on a dying flow. Carried over from the original, so not new. `if (e is CancellationException) throw e` at the top of each catch is the usual guard.

The `Failed` branch builds `"Error $actionStr: ${result.message}"` from hardcoded English fragments. Also carried over verbatim — but this repo ships five translations, so these are the strings that quietly stay English.

**Tests**

Five new cases, and they cover the branches that matter: `isLeader` derived from the loaded list rather than a second repository call (asserted via `coVerify(exactly = 0) { isTeamLeader(...) }` — a good negative assertion), self-removal promoting a candidate with `coVerifySequence` pinning the order, last-leader refusal with `removeMember` verified never called, other-user removal skipping the leader path entirely, and failure surfacing with the right `MemberAction`. `makeLeader` and the `leaveTeam` success path are uncovered; minor.

Small unrelated cleanups rode along — dropping redundant `!!` on `user1.id` (`UserEntity.id` is non-null `String`, so these were warnings) and deleting two explanatory comments about the optimistic-update timing assumption. The `!!` removal is fine; I'd have kept the comments, since they documented a non-obvious reason those assertions run before `advanceUntilIdle`.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17169, opened 2026-09-11
chat: smoother history voices share view modelling (fixes #16314)

### LINKED ISSUE #16314: delay showing already shared check
delay showing already shared check after sharing to community, team/enterprise

https://github.com/user-attachments/assets/c27e86a8-aa51-4f6f-9afa-fb351377abbe

J-S-webskas:
https://github.com/user-attachments/assets/8b67cdb9-b71c-4bd4-bf1a-35541d659fbf

## Summary by CodeRabbit

- **Bug Fixes**
  - Shared chat messages now refresh their view-in mappings after successful sharing.
  - Voice message rows update correctly when view-in information changes.
  - Share controls and message details now reflect updated sharing information without requiring a full refresh.

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/ui/chat/ChatHistoryFragment.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/chat/ChatHistoryFragment.kt
@@ -226,6 +226,7 @@ class ChatHistoryFragment : Fragment() {
                     if (isAdded && _binding != null) {
                         if (user?.planetCode != null) {
                             sharedNewsMessages = sharedNewsMessages + result.news
+                            sharedViewInIds = sharedViewModel.extractSharedViewInIds(sharedNewsMessages)
                         }
                         (binding.recyclerView.adapter as? ChatHistoryAdapter)?.let { adapter ->
                             adapter.updateCachedData(user, sharedNewsMessages, sharedViewInIds)
--- a/app/src/main/java/org/ole/planet/myplanet/ui/chat/ChatViewModel.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/ui/chat/ChatViewModel.kt
@@ -316,4 +316,7 @@ class ChatViewModel @Inject constructor(
     suspend fun getUserById(userId: String): UserEntity? {
         return userRepository.getUserById(userId)
     }
+    fun extractSharedViewInIds(sharedNewsMessages: List<News>): Map<String, Set<String>> {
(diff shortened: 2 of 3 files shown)


dogi:
**Review — rating: 88/100**

Targeted fix for #16314 (share checkmark not updating until refresh). 16 insertions across three files, and each one is load-bearing.

**Root cause is correctly identified**

In `ChatHistoryFragment.setupRealtimeSync`, the `ShareChatResult.Shared` branch appended the new item to `sharedNewsMessages` but handed the adapter a `sharedViewInIds` map that was still whatever the last `screenData` emission produced. Since the adapter decides "already shared to this destination" from that map, not from the news list, the list grew but the checkmark didn't. Recomputing the map from the updated list is the right fix, and placing it inside the same `if (user?.planetCode != null)` guard keeps the two in sync — when the list isn't appended, the map isn't recomputed either, so they can't diverge.

**`VoicesAdapter` change is the part that's easy to get wrong, and it's right**

Adding `viewIn` to `areContentsTheSame` without a matching payload bucket would have tripped the invariant the file documents in-line ("If payloads is empty here… we MUST return null to trigger a full rebind"). The PR adds both halves.

I checked the new `PAYLOAD_VIEW_IN_CHANGED` handler against the full-bind path at the top of `onBindViewHolder`: full bind computes `sharedTeamName`, then calls `showShareButton` and `setMessageAndDate`. The payload handler does exactly those three, which are precisely the `viewIn`-dependent operations — nothing more, nothing missed.

I also confirmed the cached `parsedSharedTeamName` can't go stale under this new payload. Every `submitList` overload routes through `prepareSubmitList` → `preParseNews`, which recomputes `parsedSharedTeamName` unconditionally, so the `news` object reaching the payload branch already carries a value derived from the new `viewIn`. The `?: JsonUtils.extractSharedTeamName(news)` fallback is belt-and-braces rather than load-bearing.

**Two nits, neither blocking**

`extractSharedViewInIds` runs Gson parsing over the whole shared-news list on the main thread, once per share. It's a pure function so there's no I/O, and the list is shared news only, but it does grow with every share in a session. If that list ever gets long, moving the recompute into the ViewModel's flow rather than calling it synchronously from the collector would be the cleaner shape.

The new `ChatViewModel.extractSharedViewInIds` is a bare pass-through to the repository, which is a mild abstraction leak — though it matches the existing `getUserById` pass-through a few lines up, so it's consistent with the file rather than novel.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

dogi (approved):

PR #17171, opened 2026-09-11
teams: smoother repository shareable root filtering (fixes #16976)

### LINKED ISSUE #16976: Team and enterprise sharing queries repeat the same resolve-then-filter shape
getShareableTeams, getShareableEnterpriseSummaries, and getMemberTeamIds each resolve membership ids then query root teams by type collapse into one parameterized helper.

Okuro3499:
fixes #16976
Extracted duplicated root-team filtering into `getShareableRootTeams(type, userId)` so both team and enterprise summary paths use the same membership-aware logic. This removes redundant branching and keeps behavior consistent for anonymous users and users with no team memberships.

## Summary by CodeRabbit

* **Refactor**
  * Consolidated internal team and enterprise filtering logic without changing the behavior or user-visible functionality.

### DIFF
--- a/app/src/main/java/org/ole/planet/myplanet/repository/TeamsRepositoryImpl.kt
+++ b/app/src/main/java/org/ole/planet/myplanet/repository/TeamsRepositoryImpl.kt
@@ -218,25 +218,19 @@ class TeamsRepositoryImpl @Inject constructor(
             .toHashSet()
     }
 
-    private suspend fun getShareableTeams(userId: String?): List<MyTeam> {
-        return if (userId.isNullOrBlank()) {
-            teamDao.getRootTeamsByType("team")
-        } else {
-            val memberIds = getMemberTeamIds(userId)
-            if (memberIds.isEmpty()) {
-                emptyList()
-            } else {
-                teamDao.getRootTeamsByTypeAndIds("team", memberIds)
-            }
+    private suspend fun getShareableRootTeams(type: String, userId: String?): List<MyTeam> {
+        if (userId.isNullOrBlank()) {
+            return teamDao.getRootTeamsByType(type)
         }
+        val memberIds = getMemberTeamIds(userId)
+        if (memberIds.isEmpty()) {
+            return emptyList()
+        }
+        return teamDao.getRootTeamsByTypeAndIds(type, memberIds)
     }
 
     override suspend fun getTeamSummaries(userId: String?): List<TeamSummary> {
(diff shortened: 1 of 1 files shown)


dogi:
**Review — rating: 89/100**

Collapses two private helpers with an identical shape into one parameterised method. Net 10 insertions against 26 deletions in a single file, and the deleted code is genuine duplication rather than something being inlined elsewhere.

**Equivalence, checked branch by branch**

The two originals had the same three-branch structure, differing only in the literal `"team"` / `"enterprise"`:

| | `userId` null/blank | `memberIds` empty | otherwise |
|---|---|---|---|
| old `getShareableTeams` | `getRootTeamsByType("team")` | `emptyList()` | `getRootTeamsByTypeAndIds("team", memberIds)` |
| old enterprise path | `getRootTeamsByType("enterprise")` | `emptyList()` | `getRootTeamsByTypeAndIds("enterprise", memberIds)` |
| new `getShareableRootTeams(type, userId)` | `getRootTeamsByType(type)` | `emptyList()` | `getRootTeamsByTypeAndIds(type, memberIds)` |

Every branch maps one-to-one. The DAO already took `type: String` on both queries, so nothing had to change below the repository — this is purely removing a copy. ✓

The rewrite from nested `if/else` expression to early `return`s reads better at this nesting depth, and both call sites collapse to a single `.map { it.toSummary() }` line, which makes the two entry points visibly symmetric where before you had to read both bodies to confirm they agreed.

I also checked that `getShareableTeams` and `getShareableEnterprises` have no remaining references anywhere in the tree — both are fully removed, not orphaned.

**One pre-existing thing worth knowing, not introduced here**

`getRootTeamsByTypeAndIds` is `… AND _id IN (:teamIds)` with no chunking, so it inherits SQLite's 999-parameter ceiling. Several PRs in the current queue are fixing exactly this pattern elsewhere (#17147 in `AnswerDao`, #17158 in `CourseDao`, #17160 in `TeamsRepositoryImpl`). It's not this PR's to fix and the practical risk is negligible — it would take a user belonging to 999+ teams — but now that both callers funnel through one method, this is the natural place to add chunking if anyone wants to close the pattern out repo-wide.

**Nit:** `getShareableRootTeams` is the only consumer of `getMemberTeamIds` in this path, and the "no memberships means nothing shareable" rule is now stated once instead of twice. Worth a line in the KDoc, since "empty memberships → empty list" reads as a bug to someone who hasn't seen the original.

**Verdict:** Mergeable.

---

dogi (approved):

PR #17176, opened 2026-09-14
courses: smoother item utils cover binding (fixes #16926)

### LINKED ISSUE #16926: update team courses UI
update team courses to use same layout as the courses page
Image

Okuro3499:
fixes #16926
Extract shared course row cover, subject, and meta binding into `CourseRowBinder`, then update `CoursesAdapter` to use it. `TeamCoursesAdapter` now reuses the same course list layout and binder so team course rows match the main courses UI while preserving open/remove behavior.
Screenshot_20260914_212603

## Summary by CodeRabbit

* **UI Improvements**
  * Standardized course cards and list items with consistent covers, subject icons, labels, step counts, and grade information.
  * Improved course cover handling, including local covers and fallback subject icons when no cover is available.
  * Updated team course items to show ownership and course status more clearly.
  * Added clearer removal controls for removable team courses while preserving navigation for non-removable courses.


dogi:
**Review — rating: 87/100**

Lifts the cover-binding, subject-mapping and meta-line logic out of `CoursesAdapter` into a shared `CourseRowBinder`, so `TeamCoursesAdapter` stops carrying its own copy. 155 insertions against 125 deletions across three files, and the new file is a stateless `internal object` — no lifecycle, no context held, nothing to leak.

**Why this is the right extraction**

The three `subject*Res` mappings are exhaustive `when` over `CourseSubject`, so adding an enum member now produces one compile error in one place instead of silently falling through in whichever adapter didn't get updated. That alone justifies the move — duplicated exhaustive mappings are where the two adapters were most likely to drift apart.

`bindCover` takes the views it needs as parameters rather than a binding type, which is what lets one implementation serve two different layouts (`item_course_list` / `item_course_grid` and the team variant). Keeping `viewMode` as a parameter rather than state keeps the object stateless.

The cover-resolution order is preserved exactly: local file if it exists, else an authenticated `GlideUrl`, else hide the cover and show the subject icon. The `ObjectKey(course.courseRev.orEmpty())` signature comes across intact, which matters — that's what invalidates a cached cover when the course revision changes.

**One thing carried forward that's worth fixing here**

The sizing ladder arrived with #17036, and it brings along the pessimisation I flagged there:

```kotlin
val fallbackHeight = if (viewMode == ListViewMode.GRID) {
    context.resources.getDimensionPixelSize(R.dimen.course_grid_cover_height)
} else {
    context.resources.getDimensionPixelSize(R.dimen.course_list_cover_size)
}
```

`fallbackHeight` is computed unconditionally on every bind, but it's only *used* when `coverContainer` hasn't been measured — which, after the first layout pass, is never. So every row in both adapters pays a resource lookup whose result is discarded.

I'm raising it again rather than letting it go because this PR changes the economics: previously fixing it meant editing two adapters, now there's exactly one place. Moving the lookup into the `else` branches of the two `when` expressions is a two-line change, and caching the resolved target on the ViewHolder after first measurement would collapse the common case to a field read.

**Nit:** `buildMetaLine` builds a `mutableListOf` and joins it for at most two elements. `listOfNotNull(stepsString, gradeLevel.takeIf { it.isNotBlank() }).joinToString(" · ")` says the same thing without the mutable builder.

**No conflicts** — this touches only the two course adapters and the new file; nothing else in the current review queue goes near them.

**Verdict:** Mergeable.

---

dogi (approved):

PR #17177, opened 2026-09-14
resources: smoother viewer audio error handling (fixes #16738)

### LINKED ISSUE #16738: resource type error
when playing an audio app mistakes it for a video resulting to this error
https://github.com/user-attachments/assets/1c146474-2c73-4f0f-bc42-ad0477499287

Okuro3499:
fixes #16738
This change updates the resource viewer to treat audio and video playback failures separately, using the correct player view and alert text for each media type. It also adds missing localized audio strings in all supported languages so the app presents accurate playback and availability messages when audio content fails to load or play.
[Screen_recording_20260914_221620.webm](https://github.com/user-attachments/assets/8d731223-dfb7-40a4-81a0-b0fea1bdc4fe)


dogi:
**Review — rating: 85/100**

Fixes #16738 by giving audio its own error strings and its own streaming path, rather than borrowing the video ones. Small, coherent, and the localization discipline is better than most PRs in this queue.

**What's done right**

The three new strings — `audio_unavailable`, `audio_playback_error`, `unable_to_play_audio` — are added to `values/` **and all five translations** (ar, es, fr, ne, so) in the same PR. That's the rule `CLAUDE.md` states and the one most commonly skipped; a lot of PRs here leave new strings English-only. The French entry even escapes the apostrophe correctly (`l\'audio`), which is the detail that otherwise breaks the build or silently truncates.

`streamVideoFromUrl` → `streamMediaFromUrl` with the parameter renamed to match: the function was never video-specific, and now the name doesn't lie when audio calls it.

`setupDragToPipGesture` is gated to `ResourceType.VIDEO`. Drag-to-picture-in-picture on an audio player was meaningless at best, and it was previously being attached to whatever view `R.id.video_player` resolved to.

I confirmed the two things a green build alone doesn't make obvious: `R.id.audio_player_view` exists (`viewer_audio.xml:69`), and `setupAudioViewer` can become `suspend` because its only caller, `setupViewer()`, is already `suspend`.

**The substantive change is the online path**

```kotlin
if (isOnline) {
    viewModel.ensureServerUrlUpdated()
    if (::library.isInitialized) { filePath = UrlUtils.getUrl(library) }
    authSessionUpdater = viewModel.getAuthSessionUpdater(this)
} else {
    initializeAudioPlayer(playerView)
}
```

Online audio now goes through the auth-session handshake and lands in `streamMediaFromUrl` via the callback, mirroring video. That's the actual fix — previously audio always went down `initializeAudioPlayer`, which is why a streamed audio resource produced a video-flavoured error. Offline still takes the local-file path unchanged.

**Two nits**

`playerView` is resolved at the top of `setupAudioViewer` but is only used in the `else` branch now — a dead local on the online path. Moving the lookup into the branch that uses it would make the split clearer.

The `type == ResourceType.AUDIO` ternary appears four times (title, playback error, unavailable message, player view id). If a third media type ever arrives, that's four places to find. A small `when (type)` helper returning the string set, or a couple of properties on `ResourceType`, would collapse it — worth doing if audio/video keep diverging, not worth a round-trip now.

**Ordering note:** #17178 in this batch bumps media3 to 1.11.1. These don't conflict (different files), but this PR adds new exposure to `DefaultHttpDataSource` and the player, so landing the bump first means this gets exercised against the newer library. Neither is covered by unit tests — there's no `androidTest` source set — so audio and video playback both want a manual smoke test before release.

**Verdict:** Mergeable.

---

dogi (approved):

PR #17184, opened 2026-09-15
teams: smoother voices notifications repositories dao batch counting (fixes #17179)

### LINKED ISSUE #17179: N+1 DB queries in notification aggregation
NotificationsRepositoryImpl.kt:320-324 fires one query per team via async{}.awaitAll() instead of one grouped query. Add a GROUP BY DAO method.

Okuro3499:
fixes #17179
Replace per-team voice count queries in team notifications with a single batched repository call. The voices/news layer now exposes top-level team membership data so counts can be tallied for multiple teams at once, reducing repeated DAO work while preserving team-specific filtering. Tests were updated to cover the new DAO query, batched notification flow, and empty-input behavior.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved team notification unread-count accuracy for posts associated directly or indirectly with teams.
  - Notifications now consistently include top-level team conversations while excluding replies.
  - Teams without matching conversations correctly show zero unread items.
  - Notifications for multiple teams are processed together, improving consistency and responsiveness when viewing team updates.


dogi:
**Review — rating: 83/100**

Replaces N per-team `COUNT(*)` queries with one projected fetch plus in-memory counting. The N+1 is genuinely gone and the counting logic is equivalent. My reservation is about where the work moved to, not whether it's correct.

**Equivalence**

The old per-team query counted rows matching `viewableBy='teams' AND viewableId=:teamId` **OR** `viewIn LIKE :teamPattern`, so a row satisfying both was counted once. The new Kotlin does `if (matchesViewable || matchesViewIn) count++` — also once per row per team. ✓

The `viewIn` match changes from a LIKE pattern with `ESCAPE '\'` to `contains("\"_id\":\"$teamId\"", ignoreCase = true)`. That's equivalent or safer: a `contains` is literal, so a team id containing `%` or `_` needs no escaping at all, where the old path depended on `teamIdPattern` getting it right.

Projecting to `TeamNewsMembership(viewableBy, viewableId, viewIn)` instead of hydrating full `News` rows is the right instinct — no `@TypeConverters` work, no unused columns.

**Where it gets loose: the second disjunct is unbounded**

```sql
OR viewIn LIKE '%"_id":"%'
```

That matches *any* top-level news row with any `viewIn` entry, for any team. So the query returns every such row and the per-team narrowing happens entirely in Kotlin, across `rows × teamIds` case-insensitive `contains` calls.

In practice this is fine — `notificationsById.keys` is teams with pending notifications, so single digits, and the resulting work is milliseconds. But it's the opposite direction from where the rest of this queue has been going (push the filter into SQL), and it grows with total news volume rather than with the teams being asked about.

A shape that keeps the batching *and* the filtering in SQL:

```sql
SELECT viewableId, COUNT(*) FROM news
WHERE (replyTo IS NULL OR replyTo = '')
  AND viewableBy = 'teams' COLLATE NOCASE AND viewableId COLLATE NOCASE IN (:teamIds)
GROUP BY viewableId
```

for the exact-match half, plus a narrower per-team-pattern query for the `viewIn` half. That returns counts rather than rows and never loads a `viewIn` blob. Not a blocker, but it's the version I'd want if news volume grows.

**Two smaller things**

`IN (:teamIds)` has no chunking — same SQLite 999-parameter ceiling that #17147 and #17158 exist to fix. Realistically unreachable here (it's teams with notifications, not all teams), but this is now the third DAO in the codebase with an unchunked `IN`.

The old code ran its N queries concurrently via `async`/`awaitAll`, so the "N+1" wasn't N sequential round-trips — removing that is still a win through Room's executor, but the before-state was less bad than the framing implies.

**Verdict:** Mergeable.

---

dogi (approved):

PR #17185, opened 2026-09-15
courses: smoother repository dao querying (fixes #17180)

### LINKED ISSUE #17180: full-table scan for course search
CoursesRepositoryImpl.kt:270-296 loads all courses then filters in Kotlin on every keystroke. Push to a SQL LIKE query

Okuro3499:
fixes #17180
Switch course search to a Room raw query that filters on normalized title tokens in SQL, escaping wildcard characters and preserving the existing starts-with-first ordering. Update DAO and repository tests to cover the new query path and matching behavior.


dogi:
**Review — rating: 84/100**

Moves course-title search from "load every course, map it, filter in Kotlin" to "filter in SQL, map only the matches". The escaping is careful and the tests are good. Two things about the framing and one loose end.

**The escaping is right, and tested**

```kotlin
val escapedToken = token
    .replace("\\", "\\\\")
    .replace("%", "\\%")
    .replace("_", "\\_")
queryBuilder.append(" AND courseTitleNormal LIKE ? ESCAPE '\\'")
```

Backslash first, then the two wildcards — the order matters and it's correct. Tokens go in as bind arguments rather than string-concatenated, so the `@RawQuery` is still parameterised. The dedicated test decodes the actual bind values via `SupportSQLiteProgram` and asserts `%100\%\_test%`, which is a real assertion about what reaches SQLite rather than a proxy for it. Same family as #17153's fix.

**I checked the removed null fallback — it's safe**

Old code fell back to `course.courseTitle?.let { normalizeText(it) }` when `courseTitleNormal` was null; the new path queries that column directly and does `item.courseTitleNormal ?: continue`, so a null would make a course invisible to search. That's only safe if the column is always populated, so I traced the writes: `CoursesRepositoryImpl:722` sets `courseTitleNormal = Utilities.normalizeText(title)` at construction, `title` comes from `JsonUtils.getString` (which returns `""` rather than null for a missing field), and the other three `courseDao.upsert*` sites all operate on `copy()` of already-loaded rows. So every production row carries a non-null value. Fallback removal is fine — worth knowing it was checked rather than assumed.

**The performance framing overstates what changed**

Issue #17180 is titled "full table scan for course search", and this doesn't eliminate the scan. `LIKE '%token%'` has a leading wildcard, so SQLite cannot use the `Index("courseTitleNormal")` that `MyCourse` declares — it still walks every row. What actually improves, and it's a genuine improvement, is that the walk now happens in SQLite's C code over one column instead of materialising every `MyCourse` into Kotlin and running `mapCourses` across the whole table. Moving `mapCourses` from "all courses" to "matching courses" is where the win lives.

Worth saying plainly in the PR description, because "uses the index now" is the natural reading and it isn't what happens. If eliminating the scan is the actual goal, that's an FTS table, not a `LIKE`.

**Loose end: `matchesAllParts` is now dead**

The contains branch changed from `else if (matchesAllParts(title, normalizedQueryParts))` to a bare `else` — correct, since the SQL `AND` chain already guarantees every token matched. But `matchesAllParts` is still declared at `CoursesRepositoryImpl:266` and is now referenced only by its own tests. That's exactly the category #17186 is removing in this same batch. Either delete it here or expect it back in the next dead-code sweep.

**Conflict with #17186:** both edit `CourseDao.kt` and `merge-tree` reports a content conflict — this adds the `@RawQuery`, that removes `deleteByCourseId`. Independent in intent, trivial to resolve, but whichever lands second needs a rebase.

**Verdict:** Mergeable.

---

dogi (approved):

PR #17186, opened 2026-09-15
all: less base dao model utils is more (fixes #17182)

### LINKED ISSUE #17182: Remove dead code with zero call sites
Found via a repo-wide cross-reference of declarations against usages (main + test sources, layouts, manifest, DI modules), filtering out false positives like Hilt-bound methods, Room codegen, and Android/library interface overrides dispatched polymorphically.

Pure dead-code removal — no behavior change. Reduces surface area and removes a half-wired feature stub that looked like it could be mistaken for working code.

Okuro3499:
fixes #17182


dogi:
**Review — rating: 90/100**

114 deletions, zero insertions, across 20 files. For a dead-code PR the only question that matters is whether "zero call sites" is actually true, so that's what I checked rather than reading the deletions for style.

**Verification**

Every removed symbol — `handleBackPressed`, `forceDownloadNewsImages`, `showResourceDownloadDialog`, `deleteByCourseId`, `observeByType`, `getMembersByMeetupId`, `getMemberUserIdsByMeetupId`, `getWithResourceId`, `getOfflineByResourceIds`, `getTopLevel`, `getTopLevelFlow`, `getDirectReplies`, `getByUserAndType`, `getByTitle` — has no remaining call site anywhere on the branch. I checked with word-boundary matching specifically because several have live near-namesakes that a substring grep would false-positive on: `getTopLevelByTeam` and `getTopLevelFlow` vs `getTopLevel`, `getByTitles` (plural, still used) vs `getByTitle`. None of the removed names survive.

**The non-obvious risk, ruled out**

Compilation catches unused Kotlin, but not everything is resolved at compile time. `forceDownloadNewsImages` and `showResourceDownloadDialog` are public fragment methods with exactly the signature `android:onClick` uses, and that binding is resolved reflectively at runtime — a layout reference would survive a green build and crash on tap. There is no `android:onClick` anywhere under `res/layout/`, so that class of breakage isn't possible here.

`handleBackPressed` was `open`, so the other reflective-ish risk is an override in a subclass outside the diff. Removing a base declaration that something still overrides is a compile error, and the build is green on both flavors — so nothing overrode it either.

**Why the DAO deletions are worth having beyond tidiness**

`getTopLevelFlow` and `observeByType` were `Flow`-returning queries. An unused `Flow` DAO method isn't just dead text — it's a standing invitation to wire up a second observer on a table that already has one, which is how duplicate-emission bugs start. Same for `getOfflineByResourceIds` and `getWithResourceId`, which overlap with queries that are actually in use; having both present makes it a coin flip which one the next person picks.

**One note:** #17185 in this batch also edits `CourseDao.kt` and the two do **not** merge cleanly (`merge-tree` reports a content conflict). Both changes are small and independent in intent — this one removes `deleteByCourseId`, that one adds a `@RawQuery` — so whichever lands second needs a trivial rebase keeping both edits. Flagged on #17185 as well.

**Verdict:** Mergeable.

---

dogi (approved):

PR #17187, opened 2026-09-15
resources: refactored filter logic (fixes #16282)

### LINKED ISSUE #16282: resources: not filtering all the available resources
https://github.com/user-attachments/assets/32fe7873-8cbc-45be-b1ea-e7b171bc427d

J-S-webskas:
## Summary by CodeRabbit

- **New Features**
  - Resource type labels now display clearer, localized names for PDFs, videos, audio, images, web pages, and books.
  - Resource filters recognize common media type aliases, including e-books, and detect types when metadata is incomplete.

- **Bug Fixes**
  - Filtering by subject, level, and language now handles capitalization differences more reliably.
  - Book filters consistently display the localized book label.
  - Resources without an explicit media type are included in medium facets using their detected type.


dogi:
**Review — rating: 73/100**

The two substantive fixes here are right and address #16282 directly. The medium matching that carries them is a heuristic pile that duplicates logic the codebase already has, and there are no tests despite a test file sitting next to the changed class.

**What's genuinely fixed**

Case-insensitive facet matching is almost certainly the bug. `containsAll` / `contains` are exact-match, so a resource tagged `"Math"` never matched a selected subject of `"math"` — and CouchDB-sourced data is exactly where inconsistent casing comes from. The rewrite to `all { any { equalsIgnoreCase } }` for subject and level, and `any { it.equals(library.language, ignoreCase = true) }` for language, fixes that without changing the all-must-match semantics.

The `ResourcesViewModel` fallback is the other half: a resource with a blank `mediaType` contributed nothing to the medium facet, so it was unreachable through that filter entirely. Falling back to `LibraryTypeClassifier.classify(library).name.lowercase()` surfaces it. I checked `LibraryType` — it has exactly four members (BOOK, VIDEO, AUDIO, PDF), and `matchesMedium` has a specific branch for each, so every value this fallback can produce is matchable. No gap there.

**`matchesMedium` is doing the same work three times**

```kotlin
selLower == "audio" || selLower.contains("audio") || selLower == "mp3" -> …
```

The first test is subsumed by the second — any string equal to `"audio"` also contains it. Same shape in the video, pdf and book branches.

The `else` branch then re-tests all four classifier types with `contains`:

```kotlin
(classifiedType == LibraryType.AUDIO && (selLower.contains("audio") || selLower.contains("mp3"))) ||
(classifiedType == LibraryType.VIDEO && …) || …
```

but any selection containing "audio", "video", "pdf" or "book" has already been caught by the four branches above, so those four clauses are unreachable. And `mediaTypeLower.equals(selLower, ignoreCase = true)` applies `ignoreCase` to two strings that were both already lowercased.

Underneath that: `LibraryTypeClassifier` exists to answer "what kind of resource is this", and this function consults it and then second-guesses it with substring tests on the raw `mediaType`. That's two sources of truth for one question. The version I'd want is a single normalisation — map both the selection and the resource to a `LibraryType` (or a sentinel for "other"), then compare — which collapses all thirty lines and makes the audiobook question ("should selecting Books return audiobooks?") an explicit decision rather than an emergent property of `contains`.

**The label mapping is now duplicated, and already divergent**

`ResourcesAdapter.buildMetaLine` gains a medium→label `when` that `ResourcesFilterFragment.mediumLabel` already has. They don't agree:

| case | adapter | filter sheet |
|---|---|---|
| `graphic/pictures` | → Images | falls through to raw string |
| `other` | falls through to raw string | → "Other" |

So a `graphic/pictures` resource reads "Images" in the list row and `graphic/pictures` in the filter sheet. Two copies of a mapping that disagree on their first day is the divergence hazard #17176 and #17215 were sent back for; one shared function used by both would avoid it.

**No tests**

`app/src/test/.../ui/resources/ResourcesListFilterTest.kt` already exists — #17213 extended it last week. This PR changes matching semantics in three places and adds thirty lines of new matching logic with nothing pinning any of it. Two cases would be enough to make the fix durable: a resource tagged `"Math"` matching a selected `"math"`, and a resource with blank `mediaType` both appearing in the facet list and matching when selected. Without them, the next refactor of `matchesMedium` has nothing to check itself against.

**Nit:** `org.ole.planet.myplanet.utils.LibraryTypeClassifier` is written fully qualified inline in `ResourcesViewModel` where the file imports its other utils normally.

**Verdict:** Needs a change — simplify `matchesMedium`, share the label mapping, and add the two filter tests.

---

dogi:
**Review — rating: 68/100**

Extracts `MediumUtils` as a single source of truth for medium display names and makes subject/level/language facet matching case-insensitive. Both are real improvements, and dropping `Locale.getDefault()` for the root-locale `lowercase()` is a quiet bug fix — `"IMAGE".lowercase(Locale("tr"))` yields a dotless `ı` and would have missed the `"image"` branch. CI is green on all five checks (build default + lite, test shards 1/2 and 2/2, labels). The two new tests do cover what they claim.

The problem is `matchesMedium`, which is the part the PR exists to fix. `LibraryTypeClassifier.classify` has no catch-all branch — its `else` is `LibraryType.BOOK`:

```kotlin
extension == "pdf" -> LibraryType.PDF
extension in videoExtensions || mediaType.startsWith("video") -> LibraryType.VIDEO
extension in audioExtensions || mediaType.startsWith("audio") -> LibraryType.AUDIO
else -> LibraryType.BOOK
```

So when the selected medium resolves to `targetType`, the match becomes purely `classifiedType == targetType` and the declared `mediaType` is discarded. Selecting **book** (or `epub` / `textbook`) now returns *every* resource that isn't a PDF/video/audio by file extension — images, `html`, `text/html`, everything that falls through. Previously `criteria.mediums.contains(library.mediaType)` matched only a resource whose `mediaType` literally was `"book"`. That is a straight over-match on the medium filter.

`ResourcesViewModel.extractFilterOptions` compounds it: blank `mediaType` now contributes `classify(library).name.lowercase()`, which is `"book"` for anything unclassifiable, so the over-matching **book** chip gets synthesized onto image and HTML resources rather than staying absent. Neither added test touches this path — they cover case-insensitive subjects and the PDF fallback only.

Smaller, from the same function: the `mediaTypeLower.contains(selLower)` fallback stops the **html** and **text/html** chips being disjoint, since selecting `html` now also matches `text/html`.

**Nit:** `import java.util.Locale` in `ResourcesFilterFragment` becomes unused once that `when` body is replaced — it was the file's only usage. Also, `buildMetaLine` now renders `MediumUtils.getMediumDisplayName(...)`, which returns the raw `medium` verbatim for unmapped values, so an unrecognized `mediaType` reaches the list UI untranslated where a localized type label used to be.

**Verdict:** needs another round — give `classify` an explicit fallback (or keep `mediaType` in the comparison) so `book` stops swallowing everything unclassified, and add a test that an image or `text/html` resource is *not* returned by the `book` filter.

---

dogi:

**Verdict:** needs one more round. The filtering bugs it fixes are real, but it introduces a second source of truth for the same question in three places — and one of the two copies is the buggy one.

## What's genuinely fixed

**Case-insensitive facet matching.** `library.subject?.containsAll(criteria.subjects)` was exact-match, so a resource tagged `"Math"` didn't match a `"math"` filter. The new `all { … any { equals(…, ignoreCase = true) } }` fixes subjects, levels and language together. Against CouchDB data with inconsistent casing that's a real, user-visible fix.

**Facet de-duplication that keeps display casing.** This one is nicer than it looks:

```kotlin
facets.putIfAbsent(value.lowercase(Locale.ROOT), value)
```

Keying on lowercase while storing the original collapses `"Math"`/`"math"` into one filter chip without lowercasing what the user sees. `linkedMapOf` keeps first-seen order stable. Previously the filter sheet would show both variants as separate options, and picking one would miss the other's resources.

**Three new tests, all real controls** — case-insensitive subject, blank-`mediaType`-falls-back-to-classification, and book-excludes-image-and-html. Each fails if you revert its change.

## Three duplications, introduced by this PR

**1. `MediumUtils.getCanonicalMedium` exists but `matchesMedium` doesn't use it.**

```kotlin
// MediumUtils.getCanonicalMedium
lower.contains("audio") || lower == "mp3" -> "audio"
lower.contains("video") || lower == "mp4" -> "video"
lower.contains("image") || lower.contains("graphic") -> "image"
lower.contains("html") -> "html"

// ResourcesListFilter.matchesMedium — separate when, same job
selLower.contains("audio") || selLower == "mp3" -> LibraryType.AUDIO
selLower.contains("video") || selLower == "mp4" -> LibraryType.VIDEO
selLower.contains("pdf") -> LibraryType.PDF
selLower.contains("book") || selLower == "epub" || selLower == "textbook" -> LibraryType.BOOK
```

The PR adds the shared canonicaliser and then re-implements the mapping inline in the one place where matching actually happens. They already disagree — `getCanonicalMedium` handles `image`, `graphic` and `html`; `matchesMedium` has no branch for any of them and falls through to substring matching. So the label a chip displays and the rule that decides what it selects are derived from two different tables. Routing `matchesMedium` through `getCanonicalMedium` would make the chip and its behaviour provably consistent.

**2. Two extension parsers, and the older one still has the bug.** `LibraryTypeClassifier.classify` stops calling `FileUtils.getFileExtension` and inlines its own:

```kotlin
?.substringBefore('?')?.substringBefore('#')?.substringAfterLast('.', "")?.lowercase()
```

That's a genuine improvement — `FileUtils.getFileExtension` is `substringAfterLast('/')` then `substringAfterLast('.', "")`, so `file.pdf?v=2` comes back as `"pdf?v=2"`. But `matchesMedium`'s BOOK branch still calls `FileUtils.getFileExtension`, so the fixed parser and the broken one now sit in the same feature, and the broken one is what decides book-vs-not. The query/fragment stripping belongs in `FileUtils.getFileExtension` itself — one line, fixes both call sites and every other caller.

**3. `bookExtensions` vs. the inline blocklist.** The classifier gains `bookExtensions = setOf("epub", "mobi", …)`, while `matchesMedium` separately hardcodes `setOf("png", "jpg", "jpeg", "gif", "bmp", "webp", "html", "htm", "txt")` to decide what *isn't* a book. Same question, two tables, neither aware of the other.

## `classify` is nullable but can't return null

```kotlin
fun classify(library: MyLibrary): LibraryType? {
    …
    return when {
        …
        else -> LibraryType.BOOK   // ← exhaustive
    }
}
```

The `else` guarantees a value, so the `?` is dead — and it costs two `?: LibraryType.BOOK` fallbacks at the `ResourcesAdapter` call sites that read as though they're handling a real case. Either drop the `?`, or make the `else` genuinely return `null` if "unclassifiable" is a state you want to model. Right now it's the worst of both: callers defend against something that can't happen.

## One user-visible change worth confirming is intended

`buildMetaLine` now prefers the raw `mediaType` over the classified label:

```kotlin
val typeLabel = if (mediaType != null) MediumUtils.getMediumDisplayName(context, mediaType)
                else context.getString(typeLabelRes(type))
```

`getMediumDisplayName` falls through to `canonical.replaceFirstChar { it.titlecase() }` for anything it doesn't recognise. So a resource whose `mediaType` is a MIME string rather than one of the known words now shows that string, titlecased, in the meta line — `"Application/octet-stream"` where it previously said `"Books"`. For the recognised values this is an improvement (more specific than the classified bucket); for anything else it leaks raw data into the UI. Falling back to `typeLabelRes(type)` when `getCanonicalMedium` doesn't recognise the value would keep the gain without the tail.

CI green 5/5.

**Smaller:** `matchesMedium` is a nested `when`/`if`/`when` with three matching strategies (`classifiedType ==`, exact string, `contains`) layered on each other — the `contains` fallback in particular will match any selected medium that happens to be a substring of another. Collapsing to canonical form first would let most of that go away.

---

dogi:
@J-S-webskas do not send a PR with failing builds to review
also solve the merge conflict

dogi:
**Review — rating: 64/100**
The direction is right. Moving the medium display mapping into `MediumUtils` removes the fragment-only copy. Case-insensitive subject, level and language matching fixes a real mismatch. Stripping `?query`/`#fragment` in `FileUtils.getFileExtension` fixes classification of remote URLs. The three new `ResourcesListFilterTest` cases pass. CI on the head is green (build default + lite, test default). But this can't merge as it stands:

1. **Conflicts with master.** The branch is 216 commits behind, and a trial merge conflicts in `ResourcesFilterFragment.kt`, `FileUtils.kt` and `ResourcesListFilterTest.kt`. The green CI is for the stale base, not for what would land.
2. **The medium facet isn't canonicalized.** `getFilterFacets` dedupes `mediaType` values case-insensitively only. So `video`, `video/mp4` and `mp4` stay as three separate entries, and `getMediumDisplayName` labels all three "Videos". The filter list then shows duplicate rows, and ticking one hides the others' selection state. Key the `mediums` map on `MediumUtils.getCanonicalMedium(value)` instead.
3. **`text/html` changed silently.** `getCanonicalMedium("text/html")` hits `contains("html")` first and returns `"html"`. That makes the `"text/html" -> medium_text_html` branch dead, so those resources now show the generic HTML label. The same applies to `"text/html"` in `KNOWN_CANONICALS`.
4. **Dead code.** The new `extension in bookExtensions || … -> LibraryType.BOOK` branch in `classify` returns the same value as the `else` below it. `MediumUtils.isKnownMedium` has no callers.

**Verdict:** needs another round. Merge master, canonicalize the medium facet, then fix or drop the `text/html` path.

---

PR #17193, opened 2026-09-15
resources: smoother download repository resuming (fixes #17183)

### LINKED ISSUE #17183: Downloads don't resume
DownloadService.kt:375-409 deletes the .tmp file on any interruption and never sends a Range header. For unreliable-connectivity learners, a 90%-done course package re-downloads from zero.

Okuro3499:
fixes #17183
Support HTTP Range requests in the download API and repository so interrupted downloads can resume from a saved offset. Update DownloadService to reuse `.tmp` files when the server returns 206, discard stale partials on full responses or 416 errors, and keep partial data after write failures. Add repository and service tests covering the new resume flow.

## Summary by CodeRabbit

- **New Features**
  - Added resumable file downloads that continue from previously saved partial files.
  - Uses saved validation information to ensure resumed content matches the original download.
  - Preserves partial downloads after interrupted transfers for later continuation.
  - Handles unsupported or invalid resume requests by restarting the download safely.

- **Bug Fixes**
  - Improved handling of download response statuses and file validation metadata.

- **Tests**
  - Added coverage for resumed downloads, interrupted transfers, validation checks, and recovery scenarios.


dogi:
**Review — rating: 74/100**

The resume mechanism is well built. The 206-vs-200 handling is the part that's easy to get wrong and it's exactly right, and I checked the alternative-URL path carefully expecting a bug that turned out not to be there. But there's no validator on the Range request, and that can corrupt files silently.

**What's right**

The `isPartial` gate is the crux:

```kotlin
val resumeOffset = if (isPartial) tempFile.length() else 0L
if (!isPartial) { tempFile.delete() }
...
FileOutputStream(tempFile, isPartial)
```

A server that ignores `Range` answers 200 with the whole body. Appending that to existing bytes would double the file. Tying both the offset and the `append` flag to `code == 206` means a 200 always truncates and rewrites. The inline comment states the reasoning, which is worth keeping.

The supporting details hold up: `fileSize = resumeOffset + remoteRemaining` and `total = resumeOffset` keep progress reporting honest across a resume; `contentLength == 0L && !isPartial` stops a legitimate zero-remaining 206 being read as an empty file; 416 deletes the temp so the next attempt starts clean; and keeping the partial on failure (rather than deleting) is what makes any of this work.

**The alternative-URL path is fine — but not obviously**

`resumeOffset` is computed from the primary URL's temp file and then passed to `downloadFileResponse(altUrl, …)`. That looked like a corruption bug: ask the alt server to skip N bytes, then append to a *different* temp file that doesn't have them. It isn't, because `FileUtils.getSDPathFromUrl` derives the path from `url.toUri().pathSegments` only, and `resolveAlternativeUrl` builds the alternative as `resolvedAltBase + path + query` — same path, same segments, same temp file. Worth a comment at the call site, since the correctness of passing a primary-derived offset to an alternative host depends entirely on that.

**What blocks it: no `If-Range`**

The request sends `Range: bytes=N-` with nothing identifying *which* version of the resource those N bytes came from. If the attachment at that URL changes between attempts, the server happily returns 206 starting at byte N of the **new** file, and it gets appended to bytes 0..N of the **old** one. The result is a file that is the right length, passes every check in this code, and is garbage.

Two things make this more than theoretical here. Resources are revisioned (`_rev`, `downloadedRev`) and re-synced, so the bytes behind `/resources/{id}/{filename}` genuinely can change. And this PR removes the cleanup that used to delete the temp file on failure — so a `.tmp` can now sit on disk indefinitely and be resumed against a resource that has moved on since.

The fix is standard and slots into the structure you already have:

```kotlin
val rangeHeader = if (resumeOffset > 0) "bytes=$resumeOffset-" else null
// plus: If-Range: <etag or last-modified captured from the first response>
```

A server whose validator no longer matches answers **200** instead of 206 — which this code already handles correctly by truncating and restarting. So adding `If-Range` costs one header plus somewhere to persist the validator alongside the `.tmp`, and it turns silent corruption into an automatic clean restart.

If persisting an ETag is more than you want here, a cheaper mitigation is to stamp the temp file with an age limit and ignore (delete) partials older than some threshold. That narrows the window rather than closing it, but it's better than unbounded.

**Smaller point:** with failures no longer deleting temps, abandoned `.tmp` files accumulate for downloads the user never retries. Worth a sweep of stale partials somewhere — `FreeSpaceWorker` already walks this storage.

**Verdict:** Needs a change — Range without a validator.

---

dogi:
**Re-review — Rating: 87/100** (was 74)

`If-Range` is in, and it's implemented properly rather than as a token header.

**What was added**

The validator now round-trips through the whole path: `DownloadRepositoryImpl` reads `ETag` (falling back to `Last-Modified`) off the response and carries it on `DownloadResult.Success`; `DownloadService` persists it in a `<file>.tmp.etag` sidecar next to the partial; and `resumeValidatorFor` reads it back to populate `If-Range` on the next attempt.

Three details make this correct rather than decorative:

`ifRangeHeader = if (resumeOffset > 0) ifRange else null` — `If-Range` without `Range` is meaningless, and sending it would be a protocol error rather than a harmless extra. Gated correctly.

The sidecar is deleted on 416 *and* on successful completion, and written on every response. So the `.tmp` and its `.etag` stay paired across all three exit paths. If the `.tmp` disappears independently (storage cleanup, say), `resumeOffsetFor` returns 0 and the stale validator is never consulted, because the whole resume path is gated on `resumeOffset > 0`.

The comment on `resumeValidatorFor` names the mechanism *and* what happens on a mismatch — the server answers 200, and the `isPartial` branch already discards and restarts. That's the load-bearing interaction between the two halves of this PR, and having it written down where the next reader will be is the right call.

So the corruption scenario from the last round — resource changes server-side, server returns 206 from byte N of the *new* file, bytes get appended to the *old* prefix — is now closed for any server that emits a validator.

**The residual case**

If the server sends neither `ETag` nor `Last-Modified`, `writeValidator` deletes the sidecar, and the next resume sends `Range` with no `If-Range` — i.e. back to the unvalidated behaviour. CouchDB does emit ETags on attachments, so the common path is covered, but the fallback silently degrades rather than declining to resume.

Strictly, no validator means no way to know the bytes still match, so the safe response is to start fresh:

```kotlin
val resumeOffset = if (resumeValidatorFor(url) != null) resumeOffsetFor(url) else 0L
```

which costs a restarted download in a case that shouldn't arise, and removes the only remaining path to a silently corrupt file. Worth doing, not worth blocking on — the change as it stands is a large improvement over where this started.

**Still outstanding from the last round:** abandoned partials accumulate, and now at two files each. Whatever sweeps stale `.tmp` files (`FreeSpaceWorker` is the natural home) should pick up `.tmp.etag` alongside.

**Verdict:** Mergeable.

---

dogi (approved):

PR #17197, opened 2026-09-15
courses: smoother base exam markdown utils creating (fixes #17181)

### LINKED ISSUE #17181: Duplicate Markwon instance
BaseExamFragment.kt:68 builds a second, plugin-less Markwon instance instead of reusing the cached singleton in MarkdownUtils.kt.

Okuro3499:
fixes #17181
Replace direct `Markwon.create(...)` usage in `BaseExamFragment` with the shared `MarkdownUtils.create(...)` helper, and update the fragment tests to mock and verify the new markdown creation path.


dogi:
**Review — rating: 88/100**

One line of production code: `Markwon.create(requireActivity())` → `MarkdownUtils.create(requireActivity())`. It does more than the issue title suggests.

**It genuinely dedupes**

`MarkdownUtils.create` is a double-checked-locked singleton:

```kotlin
fun create(context: Context): Markwon {
    return markwonInstance ?: synchronized(this) {
        markwonInstance ?: buildMarkwon(context.applicationContext).also { markwonInstance = it }
    }
}
```

So the second Markwon instance #17181 is about actually goes away, rather than the call merely being routed through a helper that still constructs one.

**The context handling is the part worth checking, and it's safe**

Passing `requireActivity()` into a process-wide singleton is normally how you leak an Activity across rotation — the cached object would pin the first Activity forever. It doesn't here, because `create` normalizes with `context.applicationContext` before building. The fragment can hand it an Activity and the singleton still only ever holds the Application.

That safety lives in `MarkdownUtils`, not in this diff, but it's the thing that makes this change correct rather than a leak, so it's worth having verified.

**There's a behaviour change too, and the PR doesn't mention it**

`Markwon.create(context)` builds a *bare* instance. `MarkdownUtils.buildMarkwon` configures `HtmlPlugin`, `ImagesPlugin`, `GlideImagesPlugin`, `MovementMethodPlugin`, `TablePlugin`, the `AlignTagHandler`, and the `CustomImageSpan` spans factory.

In this fragment the instance feeds `MarkwonEditor.create(markwon)` → `MarkwonEditorTextWatcher` (line 199), which highlights markdown syntax as the learner types a `textarea` answer. So the editor now recognises the same constructs — tables, HTML, alignment — that the rest of the app renders, instead of just the core set. That's a small improvement in its own right, and it also removes a real inconsistency: what the editor highlighted while typing didn't match what would be rendered afterwards.

Worth a line in the PR description, since it's the only user-visible effect and a reviewer skimming "use the util instead" wouldn't expect one.

**Nits**

The fragment's `by lazy(LazyThreadSafetyMode.NONE)` around the call is now redundant — `create` is itself cached and thread-safe, so the wrapper adds a field to memoize a lookup that's already memoized. Harmless, and removing it would touch the test's reflection setup, so not worth churning now.

The test swap is right: `mockkObject(MarkdownUtils)` / `unmockkObject` is the correct pairing for a Kotlin `object`, where `mockkStatic` was correct for `Markwon`'s static factory. The `verify(exactly = 0) { MarkdownUtils.create(activity) }` in the plain-`text` case still pins the important property — that a non-textarea question never builds the editor at all.

**Verdict:** Mergeable.

---

dogi (approved):

PR #17204, opened 2026-09-15
teams: smoother notifications repository count updating (fixes #17191)

### LINKED ISSUE #17191: replace the team-notification read-modify-write with one counting UPDATE
## Verified premise

`NotificationsRepositoryImpl.updateTeamNotification` runs a read-modify-write for a value it already has in hand (verified at master `10f5ef2`, `repository/NotificationsRepositoryImpl.kt:289-304`):

```kotlin
override suspend fun updateTeamNotification(teamId: String, news: List<News>) {
    val count = news.size
    val existing = teamNotificationDao.findByParentAndType(teamId, "chat")
    if (existing != null) {
        existing.lastCount = count
        teamNotificationDao.update(existing)
    } else {
        val notification = TeamNotification().apply {
            id = UUID.randomUUID().toString()
            parentId = teamId
            type = "chat"
            lastCount = count
        }
        teamNotificationDao.insert(notification)
    }
}
```

The common path (the row already exists) costs a SELECT and then an UPDATE-by-primary-key, to write one Int column whose value does not depend on what was read. This runs on every voices load for a team: `ui/teams/voices/TeamsVoicesViewModel.kt:58` calls it from `getFilteredNews`.

The entity has no unique index on the natural key, so an `@Upsert` is not available — `TeamNotification` (`model/TeamNotification.kt`) is `@Entity(tableName = "team_notification", indices = [Index("type")])` with `@PrimaryKey var id: String`, while the row is identified by `(parentId, type)`. The correct single-statement form is a targeted `UPDATE` whose row count tells you whether an insert is still needed.

## Do

1. In `data/room/dao/TeamNotificationDao.kt`, add a counting update:

```kotlin
@Query("UPDATE team_notification SET lastCount = :count WHERE parentId = :parentId AND type = :type")
suspend fun updateCount(parentId: String, type: String, count: Int): Int
```

2. In `NotificationsRepositoryImpl.updateTeamNotification`, call it first and insert only when it matched nothing:

```kotlin
val count = news.size
if (teamNotificationDao.updateCount(teamId, "chat", count) == 0) {
    teamNotificationDao.insert(TeamNotification().apply {
        id = UUID.randomUUID().toString()
        parentId = teamId
        type = "chat"
        lastCount = count
    })
}
```

3. `findByParentAndType` then has no production caller left (verified: its only non-test use was the line you just deleted). Remove it from the DAO.

## Constraints

- **The `UPDATE` must return `Int`, not `Unit`.** The whole point is the row count — Room returns rows-affected for an `@Query` UPDATE, and that value is what decides whether to insert. A `Unit` return makes the insert branch unreachable-or-always-taken and silently breaks the badge.
- Keep the table name `team_notification` and the column `lastCount` exactly; the entity maps them literally and there are no `@ColumnInfo` renames here.
- Do **not** touch the `id` generation in the insert branch — it stays `UUID.randomUUID().toString()`. Nothing keys off its value, but changing it is out of scope.
- Do **not** add a unique index on `(parentId, type)` and do **not** switch to `@Upsert`. Any index change means bumping `AppDatabase.version`, and `RoomModule` builds with `fallbackToDestructiveMigration(true)`, so a version bump wipes unsynced local writes on every device that has not synced.
- No behaviour change: the existing row keeps its `id` (an `UPDATE ... WHERE parentId AND type` does not rewrite the primary key), a missing row is still created with `lastCount = news.size`, and `news.size` is still the only value written.

## Tests

- Rewrite the two tests in `app/src/test/java/org/ole/planet/myplanet/repository/NotificationsRepositoryImplTest.kt:989-1030` against the new DAO call:
  - `updateCount` returns `1` → assert `coVerify(exactly = 0) { teamNotificationDao.insert(any()) }`.
  - `updateCount` returns `0` → capture the inserted `TeamNotification` with a `slot` and assert `parentId`, `type == "chat"` and `lastCount == news.size`, exactly as the current creation test does.
- `app/src/test/java/org/ole/planet/myplanet/repository/TeamChatBadgeIntegrationTest.kt` drives this path three times against a real database (`:106`, `:119`, `:134`). **It must keep passing without being edited** — that is the regression guard that the badge count still behaves identically across create-then-update.

## Out of scope

- `getTeamNotifications` and its `getByTypeAndParentIds` unchunked `IN (:parentIds)` — a separate concern, and `teamIds` is a user's own team list.
- The `coroutineScope`/`async` fan-out over `countTopLevelByTeam` further down the same function.
- Any other DAO in `data/room/dao/`.
- `BaseDashboardFragment.updateTeamNotifications` — a same-named UI function, unrelated.

dogi:
Optimized team notification updates by replacing the read-modify-write query pattern with a single counting UPDATE query in `TeamNotificationDao`, inserting only when no rows are updated. Updated `NotificationsRepositoryImpl` and corresponding unit tests.

Fixes #17191

---
*PR created automatically by Jules for task [8853385334585597093](https://jules.google.com/task/8853385334585597093) started by @dogi*


dogi:
**Review — rating: 85/100**

Replaces a SELECT-then-UPDATE with a counting `UPDATE` whose affected-row count drives the insert fallback. 8 insertions against 23 deletions, and the direction is right.

**What improves**

The old shape had a genuine TOCTOU window: read the row, mutate the Kotlin object, write it back. Two concurrent `updateTeamNotification` calls for the same team could both read the same row and the second write would clobber the first. The new `UPDATE … WHERE parentId = ? AND type = ?` is a single statement — SQLite applies it atomically, so concurrent updates serialize instead of racing, and the `lastCount` that lands is one of the two values rather than a lost write.

Using the affected-row count as the existence test is also better than a separate `SELECT`: it's one round trip instead of two, and it can't go stale between the check and the write.

**The race is narrowed, not closed**

```kotlin
if (teamNotificationDao.updateCount(teamId, "chat", count) == 0) {
    teamNotificationDao.insert(TeamNotification().apply { id = UUID.randomUUID().toString() … })
}
```

Two callers arriving with no existing row both get `0` back and both insert. Nothing in the schema stops that: `TeamNotification`'s `@PrimaryKey` is `id`, a fresh UUID per call, and the only declared index is `Index("type")` — there's no uniqueness on `(parentId, type)`. The `OnConflictStrategy.REPLACE` on `insert` can't help either, since the conflict it would resolve is on `id`, which never collides.

So the table can end up with two `chat` rows for one team, and `getByTypeAndParentIds` would then return both.

To be clear, this is **not a regression** — the old code had the same hole with a wider window (SELECT→INSERT rather than UPDATE→INSERT). It's just that the PR frames this as fixing the read-modify-write problem, and the insert half is still exposed.

Two ways to close it, both cheap and both already precedented in this repo:

- Add a unique index on `(parentId, type)` to `TeamNotification` and let `insert`'s `REPLACE` do its job. Costs an `AppDatabase` version bump, which under drop-and-resync means a rebuild — worth weighing, since this table is local-only and cheap to lose.
- Or wrap the update-then-insert in a `@Transaction` default method on the DAO. That's exactly what #17045 does for `markUploadedBatch` and what #17147 just added for `deleteBySubmissionIds`, so it's the established pattern here and needs no schema change.

**Leftover: `update` is now dead**

The PR removes `findByParentAndType` but leaves `@Update suspend fun update(item: TeamNotification)` in the DAO, and nothing on the branch calls it any more — it existed solely for the read-modify-write this PR deletes. #17186 (already `merge`) just swept this exact category out of the DAOs; worth removing here rather than letting it come back in the next pass.

**Tests**

The two cases are the right ones, and the strict-argument stubbing does more work than it looks: `coEvery { updateCount(teamId, "chat", 3) } returns 1` only matches if the repository passes exactly `3`, so a wrong count fails the test even though nothing asserts it directly. An explicit `coVerify(exactly = 1) { updateCount(teamId, "chat", 3) }` would state that intent rather than relying on the stub's strictness. The `coVerify(exactly = 0) { insert(any()) }` on the update path is a good negative assertion.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17206, opened 2026-09-15
all: smoother notifications repository task utils view modelling (fixes #17194)

### LINKED ISSUE #17194: collapse the notification enrichment fan-out into NotificationsRepository
## Verified premise

`NotificationsViewModel.loadNotifications` (verified at master `10f5ef2`, `ui/notifications/NotificationsViewModel.kt:70-144`) is doing data-layer orchestration in the UI layer. In one function it:

1. fetches payloads, then classifies them into task/join_request buckets with a hand-rolled `for` loop (`:74-83`),
2. derives `taskIds`, `parsedTaskDates`, `taskTitles`, `joinRequestIds` and `joinRequestsWithoutRelatedId` (`:85-99`),
3. opens a `coroutineScope` and fires **four** repository calls with `async`, including a conditional one and a fallback call (`:101-138`),
4. merges two different team-name maps into one (`:129-131`),
5. and only then maps to display models.

The giveaway is the repository interface: `getTaskTeamNamesByTaskIds`, `getTaskTeamNamesByTaskTitles`, `getJoinRequestDetailsBatch` and `getJoinRequestDetails` exist **only** to feed this one function. Verified — each has exactly one production caller, the lines above. They are the internals of one operation, exposed as four public methods so the ViewModel can sequence them.

## Do

1. Add a result model (`model/NotificationsEnrichment.kt`):

```kotlin
data class NotificationsEnrichment(
    val payloads: List<NotificationPayload>,
    val taskTeamNames: Map<String, String>,
    val joinRequestDetails: Map<String, Pair<String, String>>,
    val parsedTaskDates: Map<String, Pair<String, String>?>,
    val unreadCount: Int
)
```

2. Add one method to `NotificationsRepository`:

```kotlin
suspend fun getEnrichedNotifications(userId: String, filter: String, isAdmin: Boolean = false): NotificationsEnrichment
```

Implement it in `NotificationsRepositoryImpl` by moving `loadNotifications`' steps 1-4 across verbatim — same classification, same `distinct()` calls, same `coroutineScope`/`async` fan-out, same conditional on `taskTitles.isNotEmpty()`, same `""`-keyed fallback entry.

3. Remove `getTaskTeamNamesByTaskIds`, `getTaskTeamNamesByTaskTitles`, `getJoinRequestDetailsBatch` and `getJoinRequestDetails` from the **interface** and make them `private` in the impl.

4. Move `parseTaskDate` and its `TASK_DATE_PATTERN` out of the ViewModel companion into a top-level `object` in `utils/` so the repository and the ViewModel share one definition.

5. `loadNotifications` shrinks to: call `getEnrichedNotifications`, then

```kotlin
_notifications.value = enrichment.payloads.map {
    formatNotification(it, enrichment.taskTeamNames, enrichment.joinRequestDetails, enrichment.parsedTaskDates)
}
_unreadCount.value = enrichment.unreadCount
```

## Constraints

- **`formatNotification` stays in the ViewModel and does not move.** It reads `context.getString(R.string.resource_notification)`, `R.string.storage_running_low`, `R.string.storage_available`, `R.string.user_requested_to_join_team` and `R.string.join_request_prefix` (`:365-392`). Pushing any of it into the repository drags `@ApplicationContext` into the data layer, which is the opposite of the point.
- **Keep the two team-name maps merged in the same precedence order**: `getTaskTeamNamesByTaskTitles(...)` first, then `.putAll(getTaskTeamNamesByTaskIds(...))` so ids win over titles, exactly as `:129-131` does today. Reversing it silently changes which team name a task notification shows.
- Keep the `""` key for the no-`relatedId` join-request fallback, populated only when `joinRequestsWithoutRelatedId.isNotEmpty()`. `formatNotification:380` looks that key up literally.
- Keep `getTaskTeamNamesByTaskTitles` guarded by `if (taskTitles.isNotEmpty())` returning `emptyMap()` — do not make it unconditional.
- `parseTaskDate` must stay behaviourally identical, including the exact regex and the `substring`/`trim` split. Keep it `internal` so tests can still reach it.
- Leave the ViewModel's `parseTaskDate` fallback at `:353-357` in place — `formatNotification` has defaulted `emptyMap()` parameters, so that branch is still reachable.
- Do not touch `resolveType`, `getNotifications`, `getUnreadCount`, or any of the mark-read/delete paths.

## Tests

- `app/src/test/java/org/ole/planet/myplanet/ui/notifications/NotificationsViewModelTest.kt` asserts the fan-out directly: `coVerify { repository.getTaskTeamNamesByTaskIds(listOf("rel1", "rel3")) }` (`:181`), `getJoinRequestDetailsBatch(listOf("rel2"))` (`:182`), and the id/title precedence case at `:225-239`. Move those assertions down to a new `NotificationsRepositoryImplTest` case for `getEnrichedNotifications` — **keep the same expected arguments and the same precedence assertion**; that precedence case is the one that catches a wrong merge order.
- The ViewModel test at `:203-211` statically mocks `NotificationsViewModel.parseTaskDate` and asserts `verify(exactly = 1)` per task message. After the move, retarget it at the new `utils` object, keeping the exactly-once-per-message assertion.
- The remaining ViewModel tests should now stub a single `getEnrichedNotifications` and assert only the formatting and `_unreadCount`.

## Out of scope

- `NotificationsViewModel`'s repeated collection passes and its dead extension — a separate follow-up that touches the same function; it must land **after** this one.
- `buildNotificationGroups` / `buildGroupedList` and the `combine` chain.
- `updateTeamNotification`'s read-modify-write.
- `getTeamNotifications` and its unchunked `IN (:parentIds)`.

dogi:
Collapsed notification enrichment fan-out from NotificationsViewModel into NotificationsRepositoryImpl, moving date parsing into a shared TaskDateParser utility and consolidating repository calls under getEnrichedNotifications. Updated corresponding unit tests.

Fixes #17194

---
*PR created automatically by Jules for task [14040307417922813692](https://jules.google.com/task/14040307417922813692) started by @dogi*

## Summary by CodeRabbit

* **New Features**
  * Notifications now include richer task and join-request details, including team names, requester information, and parsed task dates.
  * Notification loading provides more complete unread counts and organized notification data.
  * Task dates are recognized consistently when displayed in notification messages.

* **Bug Fixes**
  * Improved handling of join requests without a related identifier.
  * Ensured task identifiers take precedence when matching task details, improving notification accuracy.


dogi:
**Review — rating: 87/100**

Moves the notification enrichment fan-out out of `NotificationsViewModel` and into `NotificationsRepositoryImpl`, behind one `getEnrichedNotifications` call returning a `NotificationsEnrichment`. The interface gets smaller, not larger, which is the right shape for this kind of consolidation.

**Interface surface**

Four methods leave the public interface and become private — `getJoinRequestDetails`, `getTaskTeamNamesByTaskIds`, `getJoinRequestDetailsBatch`, `getTaskTeamNamesByTaskTitles` — replaced by one. That's the actual win: those four existed only so the ViewModel could orchestrate them, which meant the ViewModel had to know that task team names come from two different lookups that need merging in a specific order, and that join requests need a separate fallback for missing `relatedId`. None of that is presentation logic, and none of it was testable without a ViewModel.

**The merge precedence is preserved, which is what I checked hardest**

```kotlin
val combinedTaskTeamNames = taskTeamNamesByTitlesDeferred.await().toMutableMap().apply {
    putAll(taskTeamNamesByIdsDeferred.await())
}
```

Byte-identical to what the ViewModel did at `NotificationsViewModel.kt:129-131` on master — titles first, then ids overwriting. That ordering is load-bearing (id lookups are the more specific source) and it would be easy to invert while moving code between files. It didn't move. ✓

The rest of the transplant holds up too: the same type partitioning, the same `distinct()` on ids and titles, the same empty-title short-circuit, the same `details[""] = fallbackDetail` for join requests with no `relatedId`, and the same four `async` blocks so the fan-out is still concurrent rather than serialized by the move.

Pulling `parseTaskDate` into `TaskDateParser` is a good side effect — it takes `java.util.regex.Pattern` out of a ViewModel and makes the parsing testable on its own.

**One loose end**

`getNotifications` is still public on `NotificationsRepository`, but after this change its only caller is `getEnrichedNotifications` inside the same class — I checked the whole branch and there's no production caller outside the repository. Since the PR's purpose is narrowing this interface, that one should come along: make it private, or say why it stays. Leaving it exposed invites a future ViewModel to take the un-enriched path and rebuild the fan-out that this PR just removed.

**Smaller point:** `unreadCount` sits a little oddly inside a type called `NotificationsEnrichment` — it isn't an enrichment of the payload list, it's an independent count that happened to be fetched in the same `coroutineScope`. Not worth restructuring for, but a name like `NotificationsScreenData` would describe the bag more honestly.

**Tests** were migrated on both sides rather than deleted — the repository tests now exercise `getEnrichedNotifications` directly (including the dateless-task and no-`relatedId` cases), and the ViewModel tests stub the single new call. That's the right split: the merge-order logic is now covered where it lives.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17207, opened 2026-09-15
teams: smoother voices view modelling (fixes #17195)

### LINKED ISSUE #17195: reuse the memoized shared-team name and skip the label scan in voices filtering
## Verified premise

`News` carries a memo field for exactly this value (verified at master `10f5ef2`, `model/News.kt:77-78`):

```kotlin
@Ignore
var parsedSharedTeamName: String? = null
```

`VoicesAdapter` populates it once per list at `:639` (`it.parsedSharedTeamName = JsonUtils.extractSharedTeamName(it)`) and then reads it through the memo at `:309` and `:328`:

```kotlin
val sharedTeamName = news.parsedSharedTeamName ?: JsonUtils.extractSharedTeamName(news)
```

`VoicesViewModel` ignores the memo at both of its call sites and goes straight to the parser:

- `filterNews:129` — `JsonUtils.extractSharedTeamName(news) == selectedLabel`, evaluated for **every item on every filter change and every keystroke**.
- `collectLabels:203` — `val sharedTeamName = JsonUtils.extractSharedTeamName(news)`, once per item per rebuild.

`JsonUtils.extractSharedTeamName` (`utils/JsonUtils.kt:37-56`) walks a `JsonArray`, takes `ar[0].asJsonObject`, checks `has("name")` and reads `asString` — and falls back to a full `gson.fromJson(news.viewIn, ...)` whenever `parsedViewIn` is null. None of that is free, and the answer is already sitting on the object.

Separately, `filterNews:107-115` builds `dynamicLabelDisplayToValue` by scanning **every** item's `labels` and calling `Constants.LABEL_VALUE_TO_NAME` / `VoicesLabelManager.formatLabelValue` per label — unconditionally, before `:117` even checks whether the static `labelDisplayToValue` map (`:56`, built from `Constants.LABELS`) already resolves `selectedLabel`. For every built-in label the entire scan is discarded.

## Do

1. In `filterNews:129` and `collectLabels:203`, read through the memo using the adapter's existing form:

```kotlin
news?.parsedSharedTeamName ?: JsonUtils.extractSharedTeamName(news)
```

2. In `filterNews`, resolve against the static map first and build the dynamic map only on a miss:

```kotlin
val staticLabelValue = labelDisplayToValue[selectedLabel]
val resolvedLabelValue = staticLabelValue ?: buildDynamicLabelMap(list)[selectedLabel]
```

Extract the `:107-115` loop into a `private fun buildDynamicLabelMap(list: List<News?>): Map<String, String>` so it is only invoked when `staticLabelValue == null`.

## Constraints

- **Keep the `?:` fallback — do not replace the call with a bare `parsedSharedTeamName`.** `collectLabels` and `filterNews` can both run before `VoicesAdapter.submitList` has populated the memo, so a null memo must still compute the value. This is a read-through cache, not a precondition.
- The memo is `String?` and the adapter stores `""` for "not shared". `?:` therefore recomputes only when the value was never computed, never when it is a legitimate empty string. Do not "fix" this with `isNullOrEmpty()` — that would reintroduce the parse for every non-shared item, which is most of them.
- Do **not** write to `parsedSharedTeamName` from the ViewModel. `VoicesAdapter:639` owns populating it inside the try/catch that keeps one bad row from breaking `submitList`; a second writer on a different thread is a data race on an `@Ignore` field with no `@Volatile`.
- `"Shared Chat"` keeps its own branch and must still be checked **before** `resolvedLabelValue` — do not reorder the `when`.
- The `else ->` branch must keep comparing against `selectedLabel` with `==` on the raw team name. No trimming, no case folding.
- `buildDynamicLabelMap` must preserve `putIfAbsent` semantics and the `if (!labelDisplayToValue.containsValue(label))` guard exactly — first display name wins, and labels already covered by the static map are skipped.
- Do not touch `collectLabels`' output ordering: it still ends with `allLabels.sorted()`, and still seeds `"All"`, every `Constants.LABELS` name, and `"Shared Chat"`.

## Tests

- Add a `VoicesViewModelTest` case proving the memo is used: build a `News` with `parsedSharedTeamName = "Team X"` and a `viewIn` whose parse would yield a **different** name, then assert filtering by `"Team X"` matches it. That fails if the code re-parses.
- Add the converse: `parsedSharedTeamName = null` with a valid `viewIn` still filters correctly — this is the guard for the pre-`submitList` ordering.
- Add a case where `selectedLabel` is a built-in `Constants.LABELS` display name and the list contains items with unknown labels, asserting the result is unchanged from today. Bonus if you can assert `VoicesLabelManager.formatLabelValue` is never called on that path (`verify(exactly = 0)`), which is the actual point of the skip.
- `app/src/test/java/org/ole/planet/myplanet/utils/JsonUtilsTest.kt:182` asserts `extractSharedTeamName` returns `""` — leave it alone.

## Out of scope

- `JsonUtils.extractSharedTeamName` itself — its body does not change.
- `VoicesAdapter`, including the `e.printStackTrace()` in the `:640` catch.
- The `imagesArray` / `parsedViewIn` / `parsedConversations` memos and their cache classes in `News`.
- The `query` substring filter at `:137-141`, including the unused `lowerQuery` lowercasing that `contains(..., ignoreCase = true)` makes redundant.

dogi:
This change reuses the memoized parsedSharedTeamName field on News objects in VoicesViewModel's filterNews and collectLabels methods, with a fallback to JsonUtils.extractSharedTeamName when uninitialized. Additionally, it extracts dynamic label map building into buildDynamicLabelMap and defers its invocation so that filtering by static built-in labels skips scanning and formatting unknown labels.

Fixes #17195

---
*PR created automatically by Jules for task [8324538975365134219](https://jules.google.com/task/8324538975365134219) started by @dogi*


dogi:
**Review — rating: 84/100**

Two independent changes here. The label-map hoist is unambiguously good; the memoization reuse works but rests on an implicit cross-layer contract worth naming.

**The dynamic-label hoist is the bigger win**

Previously `filterNews` built `dynamicLabelDisplayToValue` — a full scan of every news item and every label on it — on *every* filter call where `selectedLabel != "All"`, then usually discarded it, because `labelDisplayToValue[selectedLabel]` resolved first for any static label.

Now:

```kotlin
val staticLabelValue = labelDisplayToValue[selectedLabel]
val resolvedLabelValue = staticLabelValue ?: buildDynamicLabelMap(list)[selectedLabel]
```

Kotlin's `?:` is short-circuiting, so the scan only runs when a static lookup misses — i.e. for a dynamically-discovered label. For the common case of a built-in label, the scan is skipped entirely. Same resolution order as before (static preferred, dynamic as fallback), so no behaviour change. ✓

**The memoization reuse: correct, and it does help — but not on the first pass**

`news?.parsedSharedTeamName ?: JsonUtils.extractSharedTeamName(news)` is safe regardless, because the fallback covers a null field. The question is whether it ever actually hits, and the answer is more nuanced than the PR implies.

`parsedSharedTeamName` is written in exactly one place: `VoicesAdapter.kt:639`, inside `preParseNews` on the `prepareSubmitList` path. `VoicesViewModel.filterNews` runs *before* the list reaches the adapter, so on a freshly loaded list the field is null and both call sites fall through to `extractSharedTeamName`.

It does pay off afterwards. The ViewModel and the adapter hold the same `News` object references, and the adapter mutates them in place — so once a list has been submitted once, every subsequent filter (typing in search, switching labels) reads the memoized value. Since re-filtering is the hot path and first load happens once, the win is real. It's just worth describing as "reused on re-filter" rather than "reused", because a reader would otherwise expect it to help the initial render too.

**The thing I'd flag**

This makes the ViewModel depend on the adapter having populated a field on shared mutable entities. Nothing declares that contract and nothing enforces it — it holds because `preParseNews` mutates rather than copies. If that prep path ever switches to producing copies (a `DiffUtil`-friendly refactor would be the obvious trigger), the ViewModel silently stops hitting the cache and nobody notices, because the fallback keeps the results correct.

Cheapest way to make it robust: populate `parsedSharedTeamName` when the ViewModel loads the list, so the field has one owner and both layers read it. The adapter's `?:` then becomes belt-and-braces instead of the actual source.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17208, opened 2026-09-15
chat: smoother search utils caching (fixes #17202)

### LINKED ISSUE #17202: normalize only the chat field the search mode actually reads
## Verified premise

`ChatSearch.buildPrecomputedChats` normalizes **three** things for every chat, unconditionally (verified at master `10f5ef2`, `utils/ChatSearch.kt:32-46`):

```kotlin
chats.map { chat ->
    val title = if (chat.conversations != null && chat.conversations?.isNotEmpty() == true) {
        chat.conversations?.get(0)?.query?.let { Utilities.normalizeText(it) }
    } else {
        chat.title?.let { Utilities.normalizeText(it) }
    }
    val queries = chat.conversations?.map { it.query?.let { q -> Utilities.normalizeText(q) } } ?: emptyList()
    val responses = chat.conversations?.map { it.response?.let { r -> Utilities.normalizeText(r) } } ?: emptyList()
    PrecomputedChat(chat, title, queries, responses)
}
```

Each search mode then reads **exactly one** of those three:

- `searchByTitle` (`:85-106`) touches only `pChat.normalizedTitle`. It never reads `normalizedQueries` or `normalizedResponses`.
- `fullConvoSearch` (`:48-83`) picks `pChat.normalizedQueries[i]` when `isQuestion` and `pChat.normalizedResponses[i]` otherwise (`:66-70`), and never reads `normalizedTitle`.

So in `TITLE` mode every query and every response in the entire chat history is normalized and discarded; in `QUESTION` mode every response is; in the remaining mode every query is. `Utilities.normalizeText` (`utils/Utilities.kt:80-86`) allocates a `lowercase()` copy, scans it for non-ASCII, and for non-ASCII input runs `Normalizer.normalize(..., NFD)` plus a regex replace — per conversation, per search, on every keystroke.

The two `map` calls at `:42-43` are also two separate traversals of the same `conversations` list where one would do.

## Do

Make the precompute mode-aware, so each search normalizes one field in one pass.

1. Replace `PrecomputedChat` with two narrow carriers:

```kotlin
private data class TitleChat(val chat: ChatHistory, val normalizedTitle: String?)
private data class ConvoChat(val chat: ChatHistory, val normalized: List<String?>)
```

2. Build them from `search` according to `mode`:

```kotlin
suspend fun search(query: String, mode: ChatSearchMode, chats: List<ChatHistory>, dispatcher: CoroutineDispatcher = Dispatchers.Default): List<ChatHistory> =
    if (mode == ChatSearchMode.TITLE) {
        searchByTitle(query, chats, dispatcher)
    } else {
        fullConvoSearch(query, isQuestion = (mode == ChatSearchMode.QUESTION), chats, dispatcher)
    }
```

with each search function doing its own single-pass precompute inside its existing `withContext(dispatcher)` block.

3. In `fullConvoSearch`, iterate the precomputed `normalized` list directly instead of `for (i in 0 until conversations.size)` re-reading `pChat.chat.conversations`. The index is only used for the `i == 0` bucket test, so `forEachIndexed` over `normalized` covers it.

## Constraints

- **The result ordering must not change.** `fullConvoSearch` returns `inTitleStartQuery + inTitleContainsQuery + startsWithQuery + containsQuery` and `searchByTitle` returns `startsWithQuery + containsQuery`. Those four/two buckets, and the concatenation order, are the ranking — preserve them exactly.
- Keep the `break` after a chat is bucketed. Each chat is added **at most once**, on its first matching conversation; turning the loop into a `filter`/`any` that keeps scanning would duplicate entries.
- Keep the `i == 0` distinction that routes a match on the first conversation into the `inTitle*` buckets. That is why `forEachIndexed` and not `forEach`.
- Keep `if (conversation == null) continue` — a null normalized entry skips to the next conversation, it does not bucket the chat and does not break.
- Keep the title fallback exactly: first conversation's **query** when `conversations` is non-null and non-empty, otherwise `chat.title`. Note it is the query, not the response, and not `chat.title` when conversations exist.
- `startsWith`/`contains` keep `ignoreCase = true` even though both sides are already normalized. Removing it is a behaviour change for locale-specific casing and is not part of this task.
- `normalizedQueryParts` stays `s.split(" ").filterNot { it.isEmpty() }` mapped through `normalizeText`, computed once per search, outside the per-chat loop.
- Do not change `Utilities.normalizeText`, `ChatSearchMode`, or the `search` signature's default `dispatcher = Dispatchers.Default`.

## Tests

- The existing `ChatSearch` tests must pass unedited — they pin the bucket ordering, which is the risk here.
- Add a case proving the work is skipped: a `ChatHistory` whose `conversations` contain a response that would match the query, searched in `TITLE` mode, must **not** be returned. That fails if responses are still being normalized and consulted.
- Add a case with a chat whose first conversation matches and another whose third matches, asserting the first sorts ahead of the second in `fullConvoSearch` output — the `i == 0` bucketing guard.
- Add a case with a null `query` on one conversation and a match on the next, asserting the chat is still found — the `continue` guard.

## Out of scope

- `Utilities.normalizeText` and `DIACRITICS_REGEX`.
- Caching precomputed results across successive searches — a different design question, and the current code recomputes per call by design.
- `ChatHistory`, `Conversation`, and the Room converters behind `conversations`.
- `ChatRepository` / `ChatSearchMode` call sites.

dogi:
Refactor ChatSearch to precompute text normalizations in a mode-aware single-pass fashion using TitleChat and ConvoChat carrier data classes. Ensures only relevant fields (title, query, or response) are normalized per search mode and preserves search result ordering and edge case guards.

Fixes #17202

---
*PR created automatically by Jules for task [15429150340356656525](https://jules.google.com/task/15429150340356656525) started by @dogi*


dogi:
**Review — rating: 88/100**

Splits one do-everything precompute into per-mode precomputes. The saving is larger than the diff suggests.

**What it actually saves**

`buildPrecomputedChats` normalized the title *and* every conversation query *and* every conversation response for every chat, then the mode selected one of the three. So `TITLE` mode normalized every query and response it would never look at; `QUESTION` normalized every title and every response; `RESPONSE` normalized every title and every query. Roughly three times the `normalizeText` calls needed, and `normalizeText` over full conversation bodies is the dominant cost in this file.

Now each path normalizes exactly its own field. For a long chat history that's a real reduction, not a micro-optimisation.

Folding the precompute inside the existing `withContext(dispatcher)` also removes a dispatcher hop — it used to be two (`buildPrecomputedChats`, then the search) and is now one.

**The control-flow rewrite is equivalent, which is the part worth checking**

The old inner loop used `break` to stop scanning a chat's conversations once it matched and move to the next chat. `forEachIndexed` has no `break`, and the PR handles that with a labelled `run { … return@run }` rather than reaching for a flag:

```kotlin
run {
    pChat.normalized.forEachIndexed { i, norm ->
        conversation = norm ?: return@forEachIndexed   // was: if (conversation == null) continue
        …
        return@run                                     // was: break
    }
}
```

`return@forEachIndexed` is `continue`, `return@run` exits the whole traversal. Both map one-for-one onto what they replaced. The `i == 0` distinction that separates `inTitleStartQuery` from `startsWithQuery` is preserved, and the null-conversations case still yields zero iterations (`emptyList()` rather than the old `isNullOrEmpty()` guard). Same results, same bucket ordering.

**One nit**

`var conversation: String?` is still declared outside the loop and assigned inside the lambda — a leftover from the indexed-loop shape. Since the value is now produced and consumed in the same iteration, a local `val` inside the lambda would be simpler and would avoid a captured mutable var being read across a lambda boundary:

```kotlin
pChat.normalized.forEachIndexed { i, norm ->
    val conversation = norm ?: return@forEachIndexed
    …
```

Compiles either way, but the `val` version is what the new structure wants.

**Small structural point:** `TitleChat` and `ConvoChat` are each one wrapper field plus the chat. Now that each mode computes its own projection inline, the wrappers buy little over a `Pair` or just carrying the chat alongside its normalized value — not worth churning, just noting that the type split exists to serve a shape that's mostly gone.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17209, opened 2026-09-15
all: smoother user information toast handling (fixes #17201)

### LINKED ISSUE #17201: use the fragment context for UserInformationFragment's result toasts instead of MainApplication.context
## Verified premise

After the ViewModel hoist, every result toast in `UserInformationFragment` lives in one lifecycle-bound collector (verified at master `10f5ef2`, `ui/exam/UserInformationFragment.kt:80-112`):

```kotlin
collectWhenStarted(viewModel.resultEvent) { result ->
    when (result) {
        is UserInformationResult.UpdateProfileSuccess -> {
            Utilities.toast(MainApplication.context, getString(R.string.user_profile_updated))
            if (isAdded) dialog?.dismiss()
        }
        ...
    }
}
```

Six calls — `:83`, `:87`, `:92-95`, `:103-106`, `:108` — reach for the process-wide `MainApplication.context` singleton. The inconsistency is visible inside a single expression: `getString(R.string.user_profile_updated)` is resolved against the **fragment**, and then handed to a toast built on the **application** context. If the fragment context were unavailable, the `getString` would already have failed.

The same file shows the intended form at `:297`: `Utilities.toast(activity, getString(R.string.thank_you_for_taking_this_survey))`.

This is only safe to change now because the toasts moved inside `collectWhenStarted`. That helper is `repeatOnLifecycle(STARTED)` on `viewLifecycleOwner`, so the block cannot run while the fragment is detached — which is precisely the condition that made `MainApplication.context` look necessary before the hoist.

## Do

Replace `MainApplication.context` with `requireContext()` at all six call sites inside the `collectWhenStarted` block, and drop the `MainApplication` import if nothing else in the file uses it.

## Constraints

- **Do not reorder the toast and the `if (isAdded) dialog?.dismiss()` line, and do not wrap the toasts in new `isAdded` / `context != null` guards.** `requireContext()` is safe here *because* the collector only runs in the STARTED state; adding a second guard implies it is not, and would silently drop toasts if anyone later moved this code. The lifecycle scope is the guarantee — leave it as the only one.
- Use `requireContext()`, not `activity` and not `context`. `:297` uses `activity` (nullable, outside this collector) but inside a lifecycle-bound block `requireContext()` is the non-null form and matches the `getString` call sitting next to it.
- Keep every `when` branch, every `Log.d("UserInformationFragment", ...)` call, and every `if (isAdded) dialog?.dismiss()` exactly as they are.
- Keep the message arguments unchanged, including the interpolation `"Error saving submission: ${result.message}"` at `:108`.
- Do **not** touch `:297`, which is outside the collector and correctly uses `activity`.
- Do not change `collectWhenStarted`, `viewModel.resultEvent`, or the `UserInformationResult` sealed type.

## Tests

- If `UserInformationFragmentTest` exists, keep it passing. If there is no test for this fragment, do **not** create a Robolectric harness just for a context swap — the change is mechanical and the compiler plus the existing suite cover it.
- Do add a `verify` that no production source under `ui/exam/` still references `MainApplication.context` if a lint-style test of that kind already exists in the repo; otherwise skip.

## Out of scope

- The two hardcoded English strings at `:104` (`"Error: Unable to save submission - no ID provided"`) and `:108` (`"Error saving submission: ..."`). They belong in `strings.xml` with translations in all five `values-*` locales — a separate change, and bundling it here would put this PR into six resource files.
- `Utilities.toast` itself.
- The other `MainApplication.context` sites elsewhere in the codebase.
- `UserInformationViewModel` and the `resultEvent` flow.

dogi:
Refactored result toasts in UserInformationFragment to use `requireContext()` instead of `MainApplication.context` within the lifecycle-bound `collectWhenStarted` collector block, and removed the unused `MainApplication` import.

Fixes #17201

---
*PR created automatically by Jules for task [815894791254042200](https://jules.google.com/task/815894791254042200) started by @dogi*


dogi:
**Review — rating: 89/100**

Five `MainApplication.context` references in `UserInformationFragment`'s result toasts become `requireContext()`. Same family as #17101 and #17129, and the smallest possible version of it.

**Why `requireContext()` is safe here**

The usual objection is that `requireContext()` throws `IllegalStateException` on a detached fragment, which is a real hazard when a toast fires from a background result. It can't happen in this block: all five call sites are inside `collectWhenStarted(viewModel.resultEvent) { … }`, which only delivers between `onStart` and `onStop`. A fragment in that window is attached by definition, so the context is always available.

That's also why this is a strict improvement rather than a lateral move — the toast now comes from the fragment's own themed context instead of the application's, so it picks up the right configuration (locale, night mode, density) rather than whatever the Application object was created with. On a device where the user has switched app language, an application-context toast can render with the wrong locale.

The `if (isAdded) dialog?.dismiss()` guards after each toast are untouched and remain correct.

**Nit, pre-existing:** two of these toasts are hardcoded English — `"Error: Unable to save submission - no ID provided"` and `"Error saving submission: ${result.message}"` — while their neighbours use `getString(R.string.…)`. Not introduced here, and pulling them into `strings.xml` would mean five translation files, so I wouldn't bundle it. But this file is now the odd one out in an area that's otherwise fully localized, so it's a reasonable follow-up.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17213, opened 2026-09-15
resources: smoother list filter match counting (fixes #17199)

### LINKED ISSUE #17199: count matching resources without materialising the filtered list
## Verified premise

`ResourcesListFilter.countMatching` builds a whole filtered `List` and then throws it away (verified at master `10f5ef2`, `ui/resources/ResourcesListFilter.kt:47-48`):

```kotlin
fun countMatching(models: List<ResourceListModel>, criteria: ResourcesFilterCriteria, locallyOfflineIds: Set<String>): Int =
    filter(models, criteria, locallyOfflineIds).size
```

and `filter` itself allocates up to three lists per call (`:50-64`):

```kotlin
private fun filter(models, criteria, locallyOfflineIds): List<ResourceListModel> {
    val bySearchAndTags = filterBySearchAndTags(models, criteria.searchQuery, criteria.searchTags)
    return filterByFacetsAndDownloadStatus(bySearchAndTags, criteria, locallyOfflineIds)
}
```

— one from `ResourcesSearchUtils.searchLocalModels`, a second from the tag `filter` at `:59`, a third from the facet/download `filter` at `:67`.

The caller is the filter dialog: `ResourcesFragment.getFilteredCount` (`:676-685`) recomputes this for **speculative** facet selections, purely to show a count next to a chip. Every one of those calls allocates two intermediate lists and a result list whose contents are never read.

## Do

Express the post-search stages as a `Sequence` so the two entry points share one predicate chain and only the list-returning one materialises:

```kotlin
private fun matching(
    models: List<ResourceListModel>,
    criteria: ResourcesFilterCriteria,
    locallyOfflineIds: Set<String>
): Sequence<ResourceListModel> {
    val searched = ResourcesSearchUtils.searchLocalModels(models, criteria.searchQuery)
    val searchTagIds = if (criteria.searchTags.isNotEmpty()) {
        criteria.searchTags.mapTo(HashSet()) { it.id }
    } else {
        null
    }
    return searched.asSequence()
        .filter { model -> searchTagIds == null || model.tags.any { it.id in searchTagIds } }
        .filter { model -> matchesFacets(model, criteria) }
        .filter { model -> matchesDownloadFilter(model, criteria.downloadFilterIndex, locallyOfflineIds) }
}

private fun filter(...): List<ResourceListModel> = matching(models, criteria, locallyOfflineIds).toList()

fun countMatching(...): Int = matching(models, criteria, locallyOfflineIds).count()
```

`filterBySearchAndTags` and `filterByFacetsAndDownloadStatus` then have no callers — remove them.

## Constraints

- **`countMatching` must not touch `lastSignature`.** There is already a test for this — `countMatching does not suppress the next filterIfChanged call` (`ResourcesListFilterTest.kt:214-222`). Do **not** implement `countMatching` in terms of `apply` or `filterIfChanged`; it goes straight to `matching`.
- `ResourcesSearchUtils.searchLocalModels(models, searchQuery)` stays eager and stays **first**. It is not a per-element predicate and must not be folded into the sequence.
- Keep the tag-id `HashSet` built **once** outside the sequence, as `:58` does today. Building it inside a `filter` lambda would rebuild it per element and make this change a pessimisation.
- Keep the empty-tags short-circuit: with `searchTags` empty, no tag predicate is applied at all. A model whose own `tags` list is empty must still pass — do not collapse this into `model.tags.any { ... }` over an empty id set, which would reject everything.
- `matchesFacets` and `matchesDownloadFilter` keep their exact current logic, including `containsAll` for subjects/levels versus `contains` for language/medium, and `isDownloaded`'s three-way `isOffline || locallyOfflineIds.contains(...) || isLocallyOffline`.
- `filter`, `apply` and `filterIfChanged` keep their current signatures and return types. `filterIfChanged` still returns `null` on an unchanged signature.

## Tests

- All existing cases in `app/src/test/java/org/ole/planet/myplanet/ui/resources/ResourcesListFilterTest.kt` must pass unedited. That file already covers the facets, the download filter, the tag filter and the signature caching — it is the specification here.
- Add one case asserting `countMatching(models, criteria, ids) == apply(models, criteria, ids).size` for a criteria set that exercises search + tags + facets + download filter together. That is the property the shared predicate chain is supposed to guarantee, and it is what would break if the two paths ever drift again.

## Out of scope

- `ResourcesSearchUtils.searchLocalModels` itself.
- `ResourcesFragment.getFilteredCount`'s call pattern and `currentFilterCriteria()`.
- The `Signature` data class and the `filterIfChanged` caching strategy.
- `getFilterFacets` and the facet-options path.

dogi:
Refactored ResourcesListFilter to express post-search filtering stages as a Sequence. `countMatching` delegates to `matching().count()`, avoiding intermediate list allocations without touching signature caching. Removed unused filterBySearchAndTags and filterByFacetsAndDownloadStatus helpers. Added unit test verifying `countMatching` matches `apply().size` under combined filtering criteria.

Fixes #17199

---
*PR created automatically by Jules for task [10393891762053697113](https://jules.google.com/task/10393891762053697113) started by @dogi*


dogi:
**Review — rating: 85/100**

Turns the filter chain into a `Sequence` so `countMatching` can count without building the list it was previously throwing away.

**Equivalence**

The predicate order is preserved — search, then tags, then facets, then download status — and the results are the same set:

- `filter(...)` is now `matching(...).toList()`, which is what it returned before.
- `countMatching(...)` is `matching(...).count()`, where it was `filter(...).size`.
- The empty-tags case is handled by `searchTagIds == null ||` short-circuiting the predicate, matching the old code's "skip the tag filter entirely when no tags are selected". ✓

Consolidating `filterBySearchAndTags` and `filterByFacetsAndDownloadStatus` into one `matching()` also means the count path and the list path can't drift apart, since they're now literally the same pipeline.

**Where the win is, and where it isn't**

`ResourcesSearchUtils.searchLocalModels(models, criteria.searchQuery)` still runs eagerly and returns a `List` before `.asSequence()` is applied, so the search pass itself is unchanged. What the sequence removes is the intermediate list after tag filtering, the intermediate after facet filtering, and — for `countMatching` — the final result list entirely.

That matters more than it sounds, because `countMatching` is the facet-badge path: it runs once per facet over the same model list, so each call was allocating and discarding a full filtered list just to read `.size`. Now those allocations are gone across all of them.

Worth being accurate in the description, though: this doesn't make resource filtering lazy end-to-end. If the search pass is the expensive part on large libraries, this doesn't touch it — making `searchLocalModels` sequence-aware would be the follow-up that does.

**Nit:** the three chained `.filter {}` calls could be one predicate. Sequence filters don't allocate per stage, so it's purely readability — and three named conditions arguably read better than one compound `&&`, so I'd leave it. Mentioning only because "single-pass" in the framing might suggest fewer stages than there are; the passes are fused either way.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17214, opened 2026-09-15
sync: smoother submissions repository photo uploading (fixes #17196)

### LINKED ISSUE #17196: stop leaking the Room DAO's UploadedPhoto type up into PhotoUploader
## Verified premise

`UploadedPhoto` is declared **inside a Room DAO** (verified at master `10f5ef2`, `data/room/dao/SubmitPhotosDao.kt:35`):

```kotlin
@Dao
interface SubmitPhotosDao {
    ...
    data class UploadedPhoto(val photoId: String, val rev: String, val remoteId: String)
}
```

and it travels all the way up to the service layer. `services/upload/PhotoUploader.kt:14` imports it:

```kotlin
import org.ole.planet.myplanet.data.room.dao.SubmitPhotosDao.UploadedPhoto
```

so an uploader — which has nothing to do with persistence — depends on a type nested in a `@Dao`. The repository interface passes it through unchanged (`repository/SubmissionsRepository.kt:7` and `:63`).

The duplication is already visible in the file. `PhotoUploader:41` declares its own identical triple:

```kotlin
data class UploadedPhotoInfo(val photoId: String, val rev: String, val id: String)
```

and then converts one into the other at `:76`:

```kotlin
val batchedMarks = successfulUploads.map { UploadedPhoto(it.photoId, it.rev, it.id) }
```

Two structurally identical three-string records, one of them owned by the database layer, mapped positionally between each other.

## Do

1. Add a repository-layer transport type, `repository/PhotoUploadResult.kt`:

```kotlin
data class PhotoUploadResult(val photoId: String, val rev: String, val remoteId: String)
```

2. Change `SubmissionsRepository.markPhotosUploadedBatch` to take `List<PhotoUploadResult>` and drop the `SubmitPhotosDao.UploadedPhoto` import from the interface.
3. In `SubmissionsRepositoryImpl.markPhotosUploadedBatch` (`:642`), map to `SubmitPhotosDao.UploadedPhoto` there — that file already imports the DAO type and is the correct boundary for the conversion.
4. In `PhotoUploader`, delete the local `UploadedPhotoInfo` (`:41`) and the `SubmitPhotosDao.UploadedPhoto` import (`:14`), and use `PhotoUploadResult` for both the in-flight value and the batch marks.

## Constraints

- **The third field is named `id` on one side and `remoteId` on the other, and `:76` maps them positionally.** `UploadedPhotoInfo.id` is the CouchDB document id and must land in `UploadedPhoto.remoteId`, which the DAO writes to the `_id` column (`SubmitPhotosDao.UploadUpdate:38`). Standardise on `remoteId` and make sure `getString("id", object)` still feeds it. Getting this backwards swaps `_id` and `_rev` on every uploaded photo and is silent — nothing type-checks it, both are `String`.
- Do **not** change `SubmitPhotosDao`: keep `UploadedPhoto`, `UploadUpdate`, the `@Transaction markUploadedBatch` wrapper and every `@ColumnInfo` name exactly as they are. The point is that the DAO type stops travelling upward, not that it disappears.
- Keep `PhotoUploader`'s upload semantics untouched: `Semaphore(MAX_CONCURRENT_UPLOADS)` with `withPermit`, `chunked(BATCH_SIZE)`, the `catch (e: CancellationException) { throw e }` before the general catch, and the `filterNotNull()` on `awaitAll()`.
- Keep the `markPhotosUploadedBatch` call **before** the listener/attachment block, and keep both guarded by `successfulUploads.isNotEmpty()`.
- The early `return@async null` for a null `photoId` stays.

## Tests

- `PhotoUploader`'s existing tests must keep passing. Add one that captures the `List<PhotoUploadResult>` handed to `markPhotosUploadedBatch` with a `slot` and asserts, for a response body of `{"id": "remote-1", "rev": "2-abc"}`, that the result is `PhotoUploadResult(photoId = <local id>, rev = "2-abc", remoteId = "remote-1")`. **That named-field assertion is the regression guard for the id/remoteId swap** — a positional assertion would pass either way, so assert the fields by name.
- Add a `SubmissionsRepositoryImplTest` case asserting the impl maps `PhotoUploadResult` to `SubmitPhotosDao.UploadedPhoto` with the same field correspondence.

## Out of scope

- `SubmitPhotosDao.getByIds`' unchunked `IN (:ids)` — a real but separate issue, and `ids` here is one `BATCH_SIZE` chunk.
- `getUnuploadedPhotos`, `getPhotosByIds` and the rest of the `SubmissionsRepository` surface.
- `FileUploader`, `uploadAttachment`, and the `"%s/submissions/%s/%s"` format string.
- Any other DAO-nested type elsewhere in the codebase.

dogi:
Refactored SubmissionsRepository and PhotoUploader to use a dedicated repository-layer transport type `PhotoUploadResult`, preventing `SubmitPhotosDao.UploadedPhoto` from leaking into service layer code.

Fixes #17196

---
*PR created automatically by Jules for task [191931006544890572](https://jules.google.com/task/191931006544890572) started by @dogi*


dogi:
**Review — rating: 88/100**

Introduces `PhotoUploadResult` in the repository package so `PhotoUploader` no longer imports `SubmitPhotosDao.UploadedPhoto`. That's the `CLAUDE.md` boundary rule applied literally — a service reaching into a DAO's nested type is the kind of leak that makes the DAO impossible to change without touching the upload layer.

**The mapping is correct**

This is where a transposition would be easy and expensive, so I checked it against the declaration:

```kotlin
// SubmitPhotosDao.kt
data class UploadedPhoto(val photoId: String, val rev: String, val remoteId: String)

// new
data class PhotoUploadResult(val photoId: String, val rev: String, val remoteId: String)

// SubmissionsRepositoryImpl
UploadedPhoto(it.photoId, it.rev, it.remoteId)
```

Identical field names in identical order, so the positional construction lines up. Worth confirming because `rev` and `remoteId` are both opaque strings — swapping them would compile, pass tests that only check the call happened, and produce attachment uploads against the wrong revision.

The rename from `id` to `remoteId` at the call site is a genuine readability gain. `uploadAttachment(photo.photoLocation, "%s/submissions/%s/%s", uploadResult.remoteId, uploadResult.rev, listener)` says which id it is; the old `uploadInfo.id` sat next to `photoId` and `photo.id` with nothing distinguishing them.

Deleting the function-local `UploadedPhotoInfo` is the other half of the win — that was a third shape for the same three fields, declared inside `withContext`.

**The trade-off, stated plainly**

`PhotoUploadResult` is a structural clone of `UploadedPhoto`. The codebase now carries two identical three-field records, and a future field has to be added to both plus the mapping between them. That's the normal cost of a layer boundary and I think it's the right call here — the alternative is the service layer being coupled to Room's DAO surface — but it's worth naming rather than pretending the duplication is free. If a third such pair appears, that's the signal the boundary needs a shared model rather than per-layer twins.

**Nit:** `PhotoUploadResult.kt` is a three-line file for a single data class. The repository package already has `DownloadResult.kt` following the same one-type-per-file pattern, so it's consistent — just noting it could equally live alongside the `SubmissionsRepository` interface that declares it in the signature.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17215, opened 2026-09-15
enterprises: smoother repository attachment resolving (fixes #17198)

### LINKED ISSUE #17198: drop the @ApplicationContext from EnterprisesRepositoryImpl by resolving team attachments through StoragePathResolver
## Verified premise

`EnterprisesRepositoryImpl` injects an Android `Context` into the data layer (verified at master `10f5ef2`, `repository/EnterprisesRepositoryImpl.kt:19-24`):

```kotlin
class EnterprisesRepositoryImpl @Inject constructor(
    @param:ApplicationContext private val context: Context,
    private val teamDao: TeamDao,
    private val timeProvider: TimeProvider,
    private val dispatcherProvider: DispatcherProvider
) : EnterprisesRepository {
```

There is exactly **one** use of it in the whole file — `:130`, inside `attachTeamImage`:

```kotlin
val destFile = MyTeam.getAttachmentFile(context, teamId, imageName) ?: return
```

Every other collaborator in this class is already an injected abstraction, and the `Context` is the only thing standing between this repository and a plain JVM unit test.

The abstraction to use already exists — `utils/StoragePathResolver.kt`:

```kotlin
class StoragePathResolver @Inject constructor(
    @ApplicationContext private val context: Context
) {
    fun resolveFileFromUrl(url: String?): File = FileUtils.getSDPathFromUrl(context, url)
    fun resolveOleDirectory(): File = File(FileUtils.getOlePath(context))
}
```

## Do

1. Add one method to `StoragePathResolver`:

```kotlin
fun resolveTeamAttachment(teamId: String?, imageName: String?): File?
```

with the same null/blank guard and the same path as `MyTeam.getAttachmentFile` (`model/MyTeam.kt:67-72`).

2. In `EnterprisesRepositoryImpl`, replace the `@param:ApplicationContext private val context: Context` constructor parameter with `private val storagePathResolver: StoragePathResolver`, and call `storagePathResolver.resolveTeamAttachment(teamId, imageName)` at `:130`.
3. Drop the now-unused `android.content.Context` and `dagger.hilt.android.qualifiers.ApplicationContext` imports.

## Constraints

- **The resolved path must stay byte-identical, including the degenerate case.** `FileUtils.getOlePath` (`utils/FileUtils.kt:43-45`) returns `""` when `getExternalFilesDir` is null, so today `getAttachmentFile` yields the **relative** path `team_attachments/<teamId>/<imageName>`. Building it as `File(resolveOleDirectory(), "team_attachments/...")` turns that into the **absolute** `/team_attachments/...`, because `File("")` is an empty-path parent. Keep the string-concatenation form — `File("${FileUtils.getOlePath(context)}team_attachments/$teamId/$imageName")` — or handle the empty base explicitly.
- Keep the guard exactly: `if (teamId.isNullOrBlank() || imageName.isNullOrBlank()) return null`, and keep the parameters nullable so the call site's `?: return` still fires.
- Do **not** delete `MyTeam.getAttachmentFile`. It has five other callers (`TeamsRepositoryImpl:534`, `TransactionSyncManager:355`, `TeamsUploader:105`, `EnterprisesFinancesAdapter:73`, `EnterprisesReportsAdapter:109`, `EnterprisesReportsFragment:218`), and moving them all is a much larger change. This issue removes **one** `Context` injection.
- `StoragePathResolver` is constructor-injected with `@Inject` and no module binding — do not add a `@Module`/`@Binds` for it or turn it into an interface. Match what is there.
- Keep `attachTeamImage`'s shape: the `if (teamId.isBlank()) return` early exit, the `withContext(dispatcherProvider.io)` around `parentFile?.mkdirs()` + `writeBytes`, and the `updateTeamEntityById` call after the write.

## Tests

- Add an `EnterprisesRepositoryImplTest` case for `addReport` with a non-null `imageName`/`imageData` that mocks `StoragePathResolver` to return a `File` under a JUnit `TemporaryFolder`, then asserts the bytes landed and that `teamDao.upsert` was called with `imageName` set and `updated == true`.
- Add a case where `resolveTeamAttachment` returns `null`, asserting `coVerify(exactly = 0) { teamDao.upsert(any()) }` for the image branch — the `?: return` must still short-circuit before the entity update.
- Add a `StoragePathResolver` test pinning the path for a non-empty ole directory **and** one for the empty-base case, asserting the result is still relative. That second assertion is the guard for the constraint above.

## Out of scope

- `TeamsRepositoryImpl:534`, which reaches `MainApplication.context` for the same call — a worse variant, but its own issue.
- The other five `getAttachmentFile` callers, and the adapters/fragment that legitimately have a `Context`.
- `exportReportsAsCsv` and the CSV builder (that was #17094).
- `resolveFileFromUrl` / `resolveOleDirectory`, which keep their current behaviour.

dogi:
This change replaces Android Context injection in EnterprisesRepositoryImpl with StoragePathResolver:
- Added `resolveTeamAttachment(teamId, imageName)` to `StoragePathResolver` with exact null/blank guards and path string concatenation.
- Refactored `EnterprisesRepositoryImpl` constructor to accept `StoragePathResolver` instead of `@param:ApplicationContext context: Context`.
- Added unit tests for `StoragePathResolver` covering null/blank inputs, non-empty base directory pathing, and relative path preservation for empty base pathing.
- Updated `EnterprisesRepositoryImplTest` to mock `StoragePathResolver` and verify attachment writing as well as null attachment short-circuiting.

Fixes #17198

---
*PR created automatically by Jules for task [14471724947841371375](https://jules.google.com/task/14471724947841371375) started by @dogi*


dogi:
**Review — rating: 77/100**

The goal is right and the path is preserved exactly — but the implementation clones a function instead of calling it, and the original still has six other callers.

**The path is byte-identical, which I checked**

`StoragePathResolver.resolveTeamAttachment` is a character-for-character copy of `MyTeam.getAttachmentFile`'s body:

```kotlin
if (teamId.isNullOrBlank() || imageName.isNullOrBlank()) return null
return File("${FileUtils.getOlePath(context)}team_attachments/$teamId/$imageName")
```

Same guards, same expression. `getOlePath` returns `"$externalFilesDir/ole/"` with a trailing slash, so the concatenation produces the same absolute path the old call produced. Existing attachments stay reachable — that was the thing most worth verifying, and it holds.

Dropping `@ApplicationContext` from `EnterprisesRepositoryImpl` in favour of an injected resolver is also the right direction, and consistent with #17116, #17106 and #17101.

**What I'd change: delegate rather than duplicate**

`MyTeam.getAttachmentFile` is still there and still used from six places — `EnterprisesFinancesAdapter`, `EnterprisesReportsFragment`, `EnterprisesReportsAdapter`, `TeamsRepositoryImpl`, `TeamsUploader`, `TransactionSyncManager`. This PR converts one of seven call sites and copies the path expression to do it.

So the attachment layout now lives in two places. If anyone later changes where attachments are stored — a subdirectory rename, a per-planet prefix, anything — they'll change one and not the other, and the failure is silent: files written through one path, looked for through the other, and the UI just shows a missing image. That's exactly the divergence hazard #17176 removed for course-row binding and #17186 swept out of the DAOs.

One line fixes it:

```kotlin
fun resolveTeamAttachment(teamId: String?, imageName: String?): File? =
    MyTeam.getAttachmentFile(context, teamId, imageName)
```

`EnterprisesRepositoryImpl` still loses its `@ApplicationContext` — the stated goal — and the path expression stays in exactly one place. If the longer-term intent is to move the definition into `StoragePathResolver` and retire the `MyTeam` companion function, that's a fine direction too, but then it should convert all seven call sites in one go rather than leaving both live.

**Note for the follow-up:** `TeamsRepositoryImpl:534` calls `MyTeam.getAttachmentFile(MainApplication.context, …)` — still reaching for the ambient global. That's the same class of cleanup as this PR and the obvious next one, since `TeamsRepositoryImpl` already has DI wired.

**Verdict:** Needs a change — delegate instead of duplicating the path.

---

dogi:
**Re-review — Rating: 88/100** (was 77)

The duplication is gone, replaced by delegation — the whole change since the last round is three lines becoming one:

```kotlin
fun resolveTeamAttachment(teamId: String?, imageName: String?): File? =
    MyTeam.getAttachmentFile(context, teamId, imageName)
```

That's what I was after. `EnterprisesRepositoryImpl` still gets what the PR set out to give it — no `@ApplicationContext`, an injected `StoragePathResolver` instead — and the attachment path expression now exists in exactly one place. The null/blank guards come along for free rather than being restated, so there's no second copy to drift.

This also means the earlier equivalence check is now structural rather than textual: `resolveTeamAttachment` *is* `getAttachmentFile`, so there's no longer a question of whether the two produce the same path under some future edit.

**The follow-up is still open**, and worth doing while this is fresh: six call sites still reach `MyTeam.getAttachmentFile` directly, and `TeamsRepositoryImpl:534` does it through `MainApplication.context`. That one is the same cleanup as this PR, in a class that already has DI wired — the obvious next step, and now a small one since the resolver exists.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17216, opened 2026-09-15
login: smoother user repository information achievements view modelling (fixes #17203)

### LINKED ISSUE #17203: de-JSON the profile-fields update between the ViewModels and UserRepository
## Verified premise

`UserRepository.updateProfileFields` takes a `JsonObject` and immediately takes it apart with a string switch over eleven known keys (verified at master `10f5ef2`, `repository/UserRepositoryImpl.kt:459-484`):

```kotlin
override suspend fun updateProfileFields(userId: String?, payload: JsonObject) {
    ...
    payload.entrySet().forEach { (key, value) ->
        if (value != null && !value.isJsonNull && value.isJsonPrimitive) {
            val strValue = value.asString
            when (key) {
                "firstName" -> model.firstName = strValue
                "lastName" -> model.lastName = strValue
                "middleName" -> model.middleName = strValue
                "email" -> model.email = strValue
                "language" -> model.language = strValue
                "phoneNumber" -> model.phoneNumber = strValue
                "birthDate" -> model.dob = strValue
                "birthPlace" -> model.birthPlace = strValue
                "level" -> model.level = strValue
                "gender" -> model.gender = strValue
                "age" -> model.age = strValue
            }
        }
    }
    ...
}
```

The field set is closed and known at compile time, so the `JsonObject` is a typed struct in disguise — and a silent one: a typo in a key name is simply ignored.

Both callers are now ViewModels (the fragment caller moved into `UserInformationViewModel`), and **one of them already has the typed value**. `ui/exam/UserInformationFragment.kt:182-183`:

```kotlin
val profile = createUserProfile() ?: return
val user = profile.toJson()
```

`UserSurveyProfile` (`model/UserSurveyProfile.kt`) is a ten-field data class that gets serialised to JSON purely so the repository can parse it back out. The other caller, `ui/user/EditAchievementFragment.kt:180-185`, hand-builds a five-key `JsonObject` for the same trip.

## Do

1. Add `repository/ProfileFieldsUpdate.kt` covering exactly the keys the `when` handles, all nullable, null meaning "leave unchanged":

```kotlin
data class ProfileFieldsUpdate(
    val firstName: String? = null,
    val lastName: String? = null,
    val middleName: String? = null,
    val email: String? = null,
    val language: String? = null,
    val phoneNumber: String? = null,
    val birthDate: String? = null,
    val birthPlace: String? = null,
    val level: String? = null,
    val gender: String? = null,
    val age: String? = null
)
```

2. Change `UserRepository.updateProfileFields(userId: String?, update: ProfileFieldsUpdate)` and rewrite the impl to assign each non-null field directly, dropping the `entrySet()` walk.
3. Give `UserSurveyProfile` a `toProfileFieldsUpdate(): ProfileFieldsUpdate` alongside its existing `toJson()`, and have `UserInformationViewModel.updateProfile` take a `ProfileFieldsUpdate`.
4. Change `AchievementSaveRequest.profileFields` to `ProfileFieldsUpdate` and build it directly in `EditAchievementFragment`.

## Constraints

- **Two of the fields are derived, not copied, and the derivations live in `toJson()`.** `toProfileFieldsUpdate` must reproduce both exactly:
  - `birthDate = TimeUtils.convertToISO8601(dob)`, only when `dob.isNotEmpty()` (`UserSurveyProfile.kt:31-34`).
  - `age = (Calendar.getInstance().get(Calendar.YEAR) - yob.toInt()).toString()`, only when `yob.isNotEmpty()` (`:36-41`). Note it is the **age**, computed from the year of birth — not `yob` itself.
  Copying `dob` and `yob` straight through writes an unformatted date and a year-as-age into the user record.
- **Empty string must mean "leave unchanged", exactly as today.** `toJson()` only calls `addProperty` when the source `isNotEmpty()`, and `EditAchievementFragment` guards `middleName`/`birthPlace` the same way. So map empty to `null` in the conversion — do **not** pass `""` through, which would blank the stored field.
- Do **not** add `betaEnabled` to `ProfileFieldsUpdate`. `toJson()` writes it (`:46`) but the repository's `when` has no branch for it, so it has never been applied. Preserve that.
- Keep `UserSurveyProfile.toJson()` and keep `UserInformationFragment` calling it for the **submission** path (`saveSubmission(user)` at `:188`/`:192`). Only the profile-update path stops going through JSON.
- Keep the two early returns in the repository: `if (userId.isNullOrBlank()) return` and `getUserByAnyId(userId) ?: return`, and keep `model.isUpdated = true` + `upsertUser(model)` at the end.
- Keep `updateProfileFields` a single call in `AchievementViewModel.saveAchievement` (`:82`), still after `updateAchievement`, still using `_user.value?.id`.

## Tests

- `app/src/test/java/org/ole/planet/myplanet/repository/UserRepositoryImplTest.kt:350-390` has two cases, including `updateProfileFields handles empty objects and converts primitive values while skipping nulls`. Rewrite them against `ProfileFieldsUpdate`: an all-null update must leave every field untouched and still call `upsertUser` with `isUpdated = true`; a partially-populated update must change only those fields.
- Add a `UserSurveyProfile` test asserting `toProfileFieldsUpdate()` and `toJson()` agree field-for-field for a fully-populated profile — including that `birthDate` is the ISO8601 form of `dob` and `age` is `currentYear - yob`, not `yob`. **That equivalence test is the regression guard for the two derivations.**
- Add a case with `yob = ""` and `dob = ""` asserting both `age` and `birthDate` come out `null`.
- Update `AchievementViewModelTest:98/:139` and `UserInformationViewModelTest:43/:49/:57` to the new type, keeping the `coVerify(exactly = 1)` assertions.

## Out of scope

- Removing `toJson()` entirely — the submission path still needs it, and that is a separate change.
- `updateAchievement` and the rest of `AchievementSaveRequest` (`header`, `goals`, `achievements`, `references`, …), which stay as they are.
- `createUserProfile()` and the form-reading code in `UserInformationFragment`.
- The `JsonObject` payloads used elsewhere in `UserRepositoryImpl`.

dogi:
Replaced loosely typed JsonObject profile field payload with typed ProfileFieldsUpdate data class across UserRepository, ViewModels (UserInformationViewModel, AchievementViewModel), Fragments (UserInformationFragment, EditAchievementFragment), and UserSurveyProfile.

Fixes #17203

---
*PR created automatically by Jules for task [12814348435551612155](https://jules.google.com/task/12814348435551612155) started by @dogi*


dogi:
**Review — rating: 87/100**

Replaces an untyped `JsonObject` payload with a typed `ProfileFieldsUpdate` across the profile-update path. That's the `CLAUDE.md` repository convention applied where it was most clearly violated — `updateProfileFields` was a stringly-typed dispatcher over arbitrary JSON keys, so a typo in a caller's `addProperty("firstname", …)` was a silent no-op with nothing to catch it.

**Equivalence — the part that needed checking**

The old repository applied a key when the value was a present, non-null JSON primitive, which includes the empty string. The new one applies a field when it's non-null, and the callers map empty to null via `takeIf { it.isNotEmpty() }`. Those differ if any caller ever sent an empty string, so I traced both callers:

- **`UserSurveyProfile`** — the existing `toJson()` already guards every single field with `if (x.isNotEmpty())`. The new `toProfileFieldsUpdate()` uses `takeIf { it.isNotEmpty() }` on exactly the same fields, and derives `birthDate` and `age` with the same `TimeUtils.convertToISO8601(dob)` and `currentYear - yob.toInt()` arithmetic. Field-for-field equivalent. ✓
- **`EditAchievementFragment`** — this one *did* call `addProperty("firstName", firstName)` unconditionally, so on paper an empty name used to clear the stored value and now wouldn't. In practice it can't happen: both fields are validated as required earlier in the same handler (`"First name is required"` / `"Last name is required"`), so an empty value never reaches the payload. `middleName` and `birthPlace` were already conditional and map across unchanged. ✓

**One thing the typed class quietly reveals**

`toJson()` ends with `user.addProperty("betaEnabled", false)`. The old `when (key)` block in `updateProfileFields` had no `betaEnabled` branch — so that property was being constructed, passed across the boundary, and silently dropped on every profile update. `ProfileFieldsUpdate` has no such field, which is correct, and the typed signature is why it's now visible instead of invisible. Worth a line in the PR description; it's a small illustration of the argument for the change.

**Two things I'd tidy**

`UserSurveyProfile` now carries two serializers, `toJson()` and `toProfileFieldsUpdate()`, and `toJson()` is still live for the `saveSubmission` path. The ISO-8601 conversion and the age arithmetic are now written out twice, verbatim. If the age calculation ever changes — say to account for birth month — it has to change in both, and a mismatch would be silent because the two paths serve different screens. Hoisting `birthDateIso` and `calculatedAge` into private helpers (or `val`s on the class) and having both serializers read them would close that.

`yob.toInt()` is still unguarded — a non-empty, non-numeric `yob` throws `NumberFormatException` rather than skipping the field. That's carried over from `toJson()` verbatim, not introduced here, but the new function is a natural place to switch to `yob.toIntOrNull()` since you're rewriting the block anyway.

**Nit:** `ProfileFieldsUpdate` declares `birthPlace`, which `toProfileFieldsUpdate()` never populates — correct, since `UserSurveyProfile` has no such field and only the achievement path supplies it. Just noting it's intentional rather than an omission.

**No conflict** with #17209, which also touches `UserInformationFragment` — different regions of the file, and `merge-tree` reports a clean merge.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17218, opened 2026-09-15
teams: smoother events repository calendar view modelling (fixes #17217)

### LINKED ISSUE #17217: Calendar tab is a bare date picker
```                                                                                      
  CalendarFragment.kt is only 34 lines — confirmed no Meetup/data wiring at all, just a                                            
  CalendarView. It's a dashboard nav destination sitting next to fully-populated per-team calendars                     
  (Meetup data via MeetupDao/TeamsRepository) that it never aggregates or displays.                                                                       
  logged-in user's meetups across all their teams as date markers with a tap-to-see-agenda list      
  reuses data the app already has. 
```

Okuro3499:
fixes #17217
Load meetups across all of a user's teams into the calendar, show agenda details for selected days, and display each meetup's team name in the event list. This adds multi-team meetup queries in the DAO/repository layer, replaces the calendar icon with a vector drawable, updates localized team labels, and covers the new behavior with DAO, repository, view model, and adapter tests.
[Screen_recording_20260915_214253.webm](https://github.com/user-attachments/assets/602f7f21-f3f4-44b1-9990-e7e3828c4d58)

## Summary by CodeRabbit

* **New Features**
  * Added a calendar view that highlights meetup dates and displays scheduled meetups for selected days.
  * Meetup listings now show the associated team name when available.
  * Added a calendar agenda dialog with localized team labels and a close action.

* **Bug Fixes**
  * Improved meetup retrieval for events associated with multiple teams, including large team lists.

* **Tests**
  * Added coverage for team-based meetup retrieval, calendar data loading, and guest-user behavior.


dogi:
**Review — rating: 74/100**

The feature is well put together — full translations, tests at four layers, a PNG→vector swap — but the ViewModel starts a new permanent Flow collector every time the view is created, and those accumulate.

**The problem**

```kotlin
// CalendarFragment.onViewCreated
viewModel.loadMeetups()
```
```kotlin
// CalendarViewModel
fun loadMeetups() {
    viewModelScope.launch {
        val userId = userRepository.getUserModel()?.id ?: return@launch
        teamsRepository.getMyTeamsFlow(userId).collect { teams -> … }
    }
}
```

`getMyTeamsFlow` returns a Room-backed `Flow`, so `collect` never completes — it stays subscribed for the life of `viewModelScope`. Nothing cancels a previous collection before starting a new one.

`by viewModels()` scopes the ViewModel to the fragment, so it survives a configuration change while `onViewCreated` runs again. Rotate the device and you get a second collector; rotate again and a third. Each one holds a live Room invalidation subscription and writes to the same `_meetups` and `_teamNames`.

The data stays correct — every collector writes the same values — so this won't show up as a visible bug, which is exactly why it's worth catching now. What it costs is a Room observer per rotation for the ViewModel's lifetime, with the invalidation tracker fanning out to all of them on every `meetup` table write.

**Fix**

`init {}` is the shape #17088 uses in this same queue and it's the smallest change:

```kotlin
init {
    loadMeetups()
}
```
and drop the call from `onViewCreated`. If you want to keep an explicit entry point (for a pull-to-refresh, say), guard it:

```kotlin
private var meetupsJob: Job? = null

fun loadMeetups() {
    meetupsJob?.cancel()
    meetupsJob = viewModelScope.launch { … }
}
```

Either removes the accumulation. The `stateIn(viewModelScope, …)` form would also work and would drop the manual `MutableStateFlow` entirely.

**What's good**

The write ordering inside the collector is correct and easy to get wrong: `_teamNames` is assigned **before** `_meetups`, so by the time the fragment's `collectWhenStarted(viewModel.meetups)` fires and the adapter calls `teamNameProvider`, the names map is already populated. Reversing those two lines would give you a first render with blank team names.

`bindTeamName(binding, meetup)` is called on both the payload path and the full-bind path, which respects the invariant `EventsAdapter` already maintains — a field bound in one path and not the other is how stale rows appear after a partial update.

The `ic_calendar.png` → `ic_calendar.xml` swap is a delete plus an add, not two resources with the same name, so there's no duplicate-resource problem. Vector over PNG for a calendar glyph is the right call.

All three new strings land in `values/` and all five translations.

**Three smaller things**

`MeetupDao.getByTeamIds` is an unchunked `IN (:teamIds)`. The list is the user's own teams so it won't realistically reach SQLite's 999-parameter ceiling — but this is now the fourth DAO in the codebase with that shape, and #17147/#17158/#17160 all established the chunking pattern. Cheap to apply while the query is new.

`showAgendaDialog` reuses `R.layout.meetup_dialog` and hides `btnadd`. It works, but the calendar screen now depends on a layout owned by the team-calendar flow — a change there breaks this silently. A dedicated layout, or at least a comment noting the shared use, would make that survivable.

A team rename won't refresh visible rows: `areContentsTheSame` compares `Meetup` fields only, and the team name arrives from `teamNameProvider` outside DiffUtil's view. Edge case, but worth knowing it's a known gap rather than an oversight.

**Nit:** `androidx.appcompat.app.AlertDialog.Builder` is fully qualified inline where the rest of the file uses imports.

**Verdict:** Needs a change — the repeated collector.

---

dogi:
**Re-review — Rating: 76/100** (was 74)

All three points from the last round are fixed properly. But the new dialog layout ships a hardcoded placeholder that users will see.

**What's fixed**

The collector leak is closed the right way — `loadMeetups()` moved into `init {}` and made `private`, with the call removed from `onViewCreated`. The ViewModel now starts exactly one collection for its lifetime, no matter how many times the view is recreated.

`MeetupDao.getByTeamIds` is chunked at 900, correct for the single `IN` clause.

The dialog no longer borrows `meetup_dialog` and hides its add button — `calendar_agenda_dialog.xml` is its own layout, so the calendar screen no longer breaks if the team-calendar flow changes. `AlertDialog` is imported rather than fully qualified.

**What the new layout introduces**

```xml
<TextView
    android:id="@+id/tvTitle"
    android:text="Event Title"
    … />
```

Two problems, and they compound:

`tvTitle` is never referenced from Kotlin. `showAgendaDialog` resolves `rvMeetups` and `btnClose` and nothing else — I grepped the whole branch and there is no `tvTitle` reference in any source file. So the dialog renders a header reading literally **"Event Title"**, in every locale, for every user who taps a day with meetups. That's not a placeholder in a mockup; it's shipped UI text.

And it's a hardcoded string in a layout, which this repo doesn't do — `btnClose` three elements down correctly uses `@string/close`, and the PR adds its other new strings to `values/` plus all five translations. This one line is the exception.

Either populate it with something meaningful (the clicked date reads naturally for an agenda dialog — you already compute `clickedDate` in the click listener and could pass it through), or drop the `TextView` if the dialog doesn't need a header. If it stays, the text belongs in `strings.xml` with the five translations like everything else in this PR.

**Still open from last round, non-blocking:** a team rename won't refresh visible rows, since `areContentsTheSame` compares `Meetup` fields and the name comes from `teamNameProvider` outside DiffUtil's view.

**CI:** all five checks green — which won't catch this, since a hardcoded layout string compiles fine.

**Verdict:** Needs a change — the unpopulated "Event Title" header.

---

dogi:
**Re-review — Rating: 74/100** (unchanged)

The branch has moved since the last round, but only because master was merged in — the version bump, the Robolectric 4.17 bits, `BaseBindingBottomSheetFragment`. The finding is untouched.

`calendar_agenda_dialog.xml` line 14 still reads:

```xml
android:text="Event Title"
```

and `tvTitle` still has zero references in `ui/calendar/` — I re-grepped the current head. So the agenda dialog still renders a header saying literally "Event Title" to every user in every locale, and it's still a hardcoded string in a layout in a PR that otherwise puts all its new strings through `values/` plus the five translations (the newly added `team_colon` does exactly that).

Nothing else to re-review — the collector fix, the DAO chunking and the dedicated dialog layout from the previous round all still look right.

Two lines to change:
- populate `tvTitle` from the clicked date (`clickedDate` is already computed in the click listener), or delete the `TextView` if the dialog doesn't want a header
- if it stays, move the text into `strings.xml` with the five translations, like `@string/close` two elements below it

**CI:** all five checks green — as before, this isn't something CI can catch.

**Verdict:** Needs a change — same one.

---

dogi:
**Re-review — Rating: 85/100** (was 74)

The blocking finding is fixed. Commit `1d4049f` drops the `android:text="Event Title"` placeholder from `calendar_agenda_dialog.xml` and populates the title for real:

```kotlin
val titleFormatter = DateTimeFormatter.ofLocalizedDate(FormatStyle.MEDIUM).withLocale(Locale.getDefault())
dialogView.findViewById<TextView>(R.id.tvTitle).text = clickedDate.format(titleFormatter)
```

Localized formatter, locale-aware — that's the right shape for a repo with five translations.

**Verified**

- `MeetupDao.getByTeamIds` chunks at 900 for a single `IN` clause — correct against `SQLITE_MAX_VARIABLE_NUMBER = 999`, and it's a default interface method over a private `@Query`, which is the house pattern.
- `EventsAdapter.bindTeamName` appears twice, but in the two *different* `onBindViewHolder` overloads (payload path and full-bind path), not duplicated in one path. The `teamNameProvider` default of `null` keeps `ltTeamName` `GONE` for every existing call site, so `TeamCalendarFragment` is unchanged in behaviour — and it was correctly migrated from the trailing-lambda form to `EventsAdapter(onMeetupClick = { … })`.
- `ltDescription` re-parents its top constraint onto `ltTeamName`, which is `visibility="gone"` by default; ConstraintLayout collapses a GONE view to a point, so the row layout is unaffected when no team name is supplied.
- `team_colon` added to `values/` and all five `values-{ar,es,fr,ne,so}/` — no missing translation.
- `CalendarFragment` is now `@AndroidEntryPoint` (required for `by viewModels()`), and `collectWhenStarted` is the `Fragment` overload in `FlowExtensions.kt:14`, which scopes to `viewLifecycleOwner` + `repeatOnLifecycle(STARTED)` — no leaked collector, no binding access after view destruction.
- `CalendarViewModel` loads from `init`, not from `onViewCreated`, so a rotation doesn't stack collectors.
- Tests cover what matters: `CalendarViewModelTest` (aggregation across teams + the logged-out early return, asserted with `coVerify(exactly = 0)`), `MeetupDaoTest.getByTeamIds_*` against a real Room instance, and `EventsRepositoryImplTest`. All five checks green.

**Notes, not blockers**

1. `_meetups` only refreshes when `getMyTeamsFlow` re-emits. A meetup arriving from a sync that doesn't change team membership won't repaint the calendar until the fragment is recreated. A `Flow`-returning meetup query (`observeByTeamIds`) with `flatMapLatest` would close that, if a later PR wants it.
2. Only `meetup.startDate` is marked and filtered on. A meetup spanning `startDate..endDate` shows on its first day only. Fine for the issue as written; worth knowing.
3. `setCalendarDays` receives one `CalendarDay` per meetup, so two meetups on the same date produce two entries for that day. Harmless for the marker, but `distinctBy { }` on the date would express the intent better.

---

dogi (approved):

PR #17220, opened 2026-09-15
life: smoother personals repository serialized uploading (fixes #17192)

### LINKED ISSUE #17192: move Personal.serialize out of the Room entity into PersonalsRepositoryImpl

dogi:
Moves `Personal.serialize` out of the Room `@Entity` `Personal.kt` and into `PersonalsRepositoryImpl.kt` as a private helper method, sourcing `customDeviceName` from `deviceNameProvider`. Unused imports from `Personal.kt` and `PersonalTest.kt` have been cleaned up and migrated.

Fixes #17192

---
*PR created automatically by Jules for task [682154609963245420](https://jules.google.com/task/682154609963245420) started by @dogi*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved personal document uploads by consistently including key metadata such as titles, authors, filenames, device details, privacy settings, and timestamps.
  * Improved handling of personal records with missing or default values, ensuring uploads remain valid and retain available information.

* **Tests**
  * Added coverage for personal document uploads with both complete and partially populated records.


dogi:
**Review — rating: 86/100**

This is the change #17205's title promised, now actually present in the diff: `Personal.serialize` moves out of the Room entity's companion and into `PersonalsRepositoryImpl` as a private method. Worth saying plainly, since the previous attempt under this title contained an unrelated build change — this one is 98 insertions against 124 deletions across `Personal.kt`, `PersonalsRepositoryImpl.kt` and their tests, and does what it says.

**The entity comes out clean**

`Personal.kt` sheds five imports — `JsonObject`, `Date`, `FileUtils`, `NetworkUtils`, `addDocumentOrigin` — and is left as a plain `@Entity` with fields. A Room entity that knows how to build a CouchDB upload payload is exactly the coupling `CLAUDE.md` warns against, and this is the same move #17146 made for `MyLibrary.serialize`. Consistent direction.

The `customDeviceName: String` parameter disappears with it. `PersonalsRepositoryImpl` already injects `DeviceNameProvider`, so `serialize` reads it directly instead of having the caller fetch and thread it through. One fewer thing for `uploadPersonalDocument` to get right.

Marking it `private` is correct — `uploadPersonalDocument` is the only caller, and there's no reason for the payload shape to be part of any public surface.

The body is otherwise a verbatim transplant: same properties in the same order, same `resourceType = "Activities"`, same `private = true`, same nested `privateFor.users`. No behaviour change to the upload.

**One consistency note**

`` `object`.addProperty("uploadDate", Date().time) `` keeps the raw wall-clock read. `PersonalsRepositoryImpl` injects `personalDao`, `uploadRepository` and `deviceNameProvider` — no `TimeProvider` — so this would mean adding a dependency rather than using an existing one.

I'd still suggest it, because #17146 took exactly this suggestion for `serializeForUpload` and is already labelled `merge`, so the two sibling serializers would otherwise diverge on the one detail that makes them testable. There's a second `Date().time` at line 44 in the same file, so injecting `TimeProvider` once would cover both and make the whole class deterministic under test.

**Nit:** the backtick-escaped `` `object` `` and `object1` names came across unchanged. Faithful to the original and I wouldn't rename them in this PR, but `payload` and `privateFor` would read better if someone touches this again.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17221, opened 2026-09-15
sync: smoother time upload crash logging (fixes #17200)

### LINKED ISSUE #17200: log the swallowed crash-log upload failure in SyncTimeLogger instead of printStackTrace
## Verified premise

`SyncTimeLogger` swallows an exception into `System.err` (verified at master `10f5ef2`, `utils/SyncTimeLogger.kt:122-125`):

```kotlin
try {
    uploadManager?.uploadCrashLog()
} catch (e: Exception) {
    e.printStackTrace()
}
```

Every other diagnostic in this file goes through `Log.d(TAG, ...)` — `:77`, `:92-95`, `:152`, `:173`, `:188`, `:200` — against the tag the class already owns (`:329`, `private const val TAG = "SyncPerf"`). This one failure path is the exception: it writes to stderr with no tag, so it cannot be filtered with the rest of the sync logging, and on release builds it is the only thing here that is neither tagged nor gated.

The irony is specific: this is the failure of the *crash-log upload*, i.e. exactly the diagnostic you most want to know went missing.

## Do

Replace the body of the catch with a tagged error log that carries the throwable:

```kotlin
} catch (e: Exception) {
    Log.e(TAG, "crash log upload failed", e)
}
```

## Constraints

- **Do not interpolate `e.message`, `e.toString()`, or any URL into the log line.** Upload failures in this app routinely carry the CouchDB endpoint in the exception message, and myPlanet server URLs embed `satellite:<pin>@host`; release builds do not strip logs. Pass the throwable as `Log.e`'s third argument and keep the message a fixed string literal — that is the same information `printStackTrace` already emits today, with nothing new interpolated.
- Reuse the existing `TAG` constant at `:329`. Do not add a second tag, and do not rename `"SyncPerf"`.
- Keep the `try`/`catch` exactly where it is, still catching `Exception`, still around only `uploadManager?.uploadCrashLog()`. The failure must stay swallowed — this call is best-effort cleanup at the end of a sync and must not abort it.
- Do not gate this line behind `isVerbose` or `isLogging`. The `Log.d` calls in this file are deliberately verbose-gated; a swallowed error is not, and `:122-125` sits outside those guards today.
- Change nothing else in the surrounding block — `serverUrlMapper.updateUrlPreferences`, the `prefs.edit()` editor, and the `uri`/`alternativeUrl`/`mapping.primaryUrl` arguments all stay as they are.

## Tests

- Add a `SyncTimeLoggerTest` case where `uploadManager.uploadCrashLog()` throws, asserting the surrounding function still completes normally (no exception propagates) — the swallow is the behaviour that matters and is currently untested.
- If the file's existing tests already use Robolectric, assert via `ShadowLog.getLogs()` that an `ERROR` entry with tag `SyncPerf` was written. If they do not, do not introduce Robolectric just for this; the completion assertion is enough.

## Out of scope

- The other ~90 `printStackTrace` sites in the codebase (16 in `UserRepositoryImpl`, 7 each in `TimeUtils` and `ServerReachabilityWorker`, 5 each in `VoicesAdapter` and `HealthRepositoryImpl`). They need per-site judgement about what is safe to log and are deliberately not bundled here.
- `DownloadUtils`' `printStackTrace`, which is a genuine credential-leak question and is parked pending a decision.
- `UploadManager.uploadCrashLog` itself and its error handling.
- The verbose-gating strategy and `Log.isLoggable(TAG, Log.DEBUG)` at `:68`.

dogi:
Replaced printStackTrace() with Log.e(TAG, "crash log upload failed", e) in SyncTimeLogger.kt to ensure swallowed crash-log upload failures are logged with tag "SyncPerf" without leaking sensitive URLs/credentials. Added a unit test in SyncTimeLoggerTest.kt asserting that exceptions thrown by uploadCrashLog are swallowed and logged via Log.e.

Fixes #17200

---
*PR created automatically by Jules for task [8594870096525345357](https://jules.google.com/task/8594870096525345357) started by @dogi*


dogi:
**Review — rating: 88/100**

One line: `e.printStackTrace()` → `Log.e(TAG, "crash log upload failed", e)`.

Small, but the right kind of small. `printStackTrace()` writes to `System.err`, which on Android is routed to logcat at a level and tag the platform picks — so a failure here was effectively invisible in a filtered log and impossible to attribute. `Log.e` with the class's own tag makes it findable and severity-correct.

The irony is worth noting: this is the crash-log *uploader* silently swallowing its own failure. A device that can't upload crash logs would give no signal that anything was wrong — the reports just never arrive, and the one place that knew about it printed to a stream nobody was watching. That's exactly the failure mode you most want logged.

The supporting pieces were already in the file — `import android.util.Log` at line 3 and `private const val TAG = "SyncPerf"` in the companion — so the change is genuinely one line with no new imports. Using the existing `SyncPerf` tag also keeps it grouped with the rest of this class's sync-timing output rather than inventing a second tag for one call site.

Same family as #17128 (`Log swallowed exceptions in CrashLogStore`), which has already merged — so this is the consistent treatment of the same anti-pattern in a neighbouring file rather than a one-off.

The `catch (e: Exception)` remains, which is right: a failed crash-log upload shouldn't take down the sync-timing path around it. Logging is the correct response, not rethrowing.

**Nit, pre-existing:** the bare `catch (e: Exception)` also catches `CancellationException`, so a cancelled upload would be logged as a failure. Not introduced here and arguably harmless for a log line, but `if (e is CancellationException) throw e` is the guard used elsewhere in this codebase.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17222, opened 2026-09-15
login: smoother shared preferences saved users caching (fixes #17219)

### LINKED ISSUE #17219: cache the parsed saved-users list in SharedPrefManager instead of re-parsing per call
## Verified premise

`SharedPrefManager.getSavedUsers` re-reads and re-parses the whole saved-users JSON on every call (verified at master `052eed9`, `services/SharedPrefManager.kt:73-80`):

```kotlin
fun getSavedUsers(): List<User> {
    val usersJson = pref.getString(SAVED_USERS, null)
    return if (usersJson != null) {
        gson.fromJson(usersJson, userListType)
    } else {
        emptyList()
    }
}
```

`gson.fromJson` against a `TypeToken<List<User>>` reflectively rebuilds every `User` on each call. There are five call sites, and three of them are read-modify-write pairs that parse the list purely to append one entry and serialise it straight back:

- `UserRepositoryImpl:1002` → `setSavedUsers` at `:1017`/`:1032`
- `UserRepositoryImpl:1037` → `setSavedUsers` at `:1053`
- `TeamsRepositoryImpl:951` → `setSavedUsers` at `:955`
- `UserRepositoryImpl:999` (`override suspend fun getSavedUsers()`), which `LoginViewModel:62` calls to populate the saved-user list on the login screen

Two facts make a cache safe here, and both are worth stating because they are what the change rests on:

1. `SharedPrefManager` is `@Singleton` (`:17-21`), so a field lives exactly as long as the preference it mirrors.
2. `SAVED_USERS` has **one** writer in the entire codebase — `setSavedUsers` at `:82-84`. Verified: no other production code references the key.

The codebase already has the right idiom for this, in `model/News.kt:80-99` — a cache record holding both the raw string and the parsed value, compared on read so a changed source can never serve stale data.

## Do

Mirror the `News.imagesCache` pattern in `SharedPrefManager`:

```kotlin
private data class SavedUsersCache(val raw: String?, val parsed: List<User>)

@Volatile
private var savedUsersCache: SavedUsersCache? = null

fun getSavedUsers(): List<User> {
    val usersJson = pref.getString(SAVED_USERS, null)
    val cache = savedUsersCache
    if (cache != null && cache.raw == usersJson) {
        return cache.parsed
    }
    val parsed = if (usersJson != null) {
        gson.fromJson(usersJson, userListType) ?: emptyList()
    } else {
        emptyList()
    }
    savedUsersCache = SavedUsersCache(usersJson, parsed)
    return parsed
}
```

and have `setSavedUsers` refresh the cache with the value it just wrote, so the read that follows a write costs nothing:

```kotlin
fun setSavedUsers(users: List<User>) {
    val json = gson.toJson(users)
    pref.edit { putString(SAVED_USERS, json) }
    savedUsersCache = SavedUsersCache(json, users.toList())
}
```

## Constraints

- **Keep the `pref.getString` on every call and compare it to `cache.raw`.** Do not skip the preference read on a cache hit. `rawPreferences` is exposed publicly at `:24`, so the key is reachable from outside this class; the raw-string comparison is what makes an out-of-band write impossible to serve stale, and it costs an in-memory map lookup, not a disk read. A bare "parse once, remember forever" field is the wrong shape here and will be rejected.
- Store `users.toList()` in `setSavedUsers`, not the caller's list. Three callers build the list with `toMutableList()`/`ArrayList(...)` and could keep mutating it after the call; caching their reference would let the cache drift from what was serialised.
- Keep returning `emptyList()` for a null preference, and keep the existing behaviour where a caller is free to copy the result. Do not wrap the return in `Collections.unmodifiableList` or otherwise change the type — `List<User>` stays.
- `setSavedUsers` must still write through `pref.edit { }` exactly as today. Do not switch to `commit()`.
- Mark the cache field `@Volatile`, as `News.kt:83-84` does for the same reason.
- Do not change `getSavedUsers`' signature, and do not touch `UserRepositoryImpl:999`'s `override suspend fun getSavedUsers()` or any of the five call sites — this change is entirely inside `SharedPrefManager`.
- Do not add caching to any other getter in this class. `SharedPrefManager` has dozens; this issue covers the one that parses JSON.

## Tests

- Add a `SharedPrefManagerTest` case asserting two successive `getSavedUsers()` calls with no intervening write return an equal list, and that `Gson.fromJson` is invoked once — inject a mocked/spied `Gson` so the call count is observable. That is the actual claim of this change.
- Add a case proving the raw comparison works: call `getSavedUsers()`, then write a different `savedUsers` JSON **directly through `rawPreferences`**, then call `getSavedUsers()` again and assert the new value is returned. This is the guard for the constraint above and fails for a naive field cache.
- Add a case asserting `setSavedUsers(list)` followed by `getSavedUsers()` returns an equal list without re-parsing.
- Add a case where the stored JSON is absent, asserting `emptyList()`.

## Out of scope

- The read-modify-write shape of the three call sites in `UserRepositoryImpl` and `TeamsRepositoryImpl` — they would each be a separate change.
- Every other getter and setter in `SharedPrefManager`.
- `SharedPrefManager`'s `rawPreferences` escape hatch, which stays.
- Moving saved users out of SharedPreferences into Room.

dogi:
Cache parsed saved users list in SharedPrefManager to avoid re-parsing JSON on every call. Added unit tests verifying caching behavior and raw string comparison.

Fixes #17219

---
*PR created automatically by Jules for task [15464668619814106154](https://jules.google.com/task/15464668619814106154) started by @dogi*

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved saved-user handling to prevent outdated results after stored data changes.
  - Ensured saved-user data remains isolated from unintended changes when lists or entries are modified.
  - Added safer handling when no saved-user data is available, returning an empty list instead of failing.
  - Improved consistency after saving users so newly saved data is reflected immediately.


dogi:
**Review — rating: 89/100**

Memoizes the Gson parse of the saved-users list, keyed on the raw JSON string. The cache-key choice is what makes this safe rather than fragile.

**Why keying on the raw string is the right call**

```kotlin
private data class SavedUsersCache(val raw: String?, val parsed: List<User>)

@Volatile private var savedUsersCache: SavedUsersCache? = null
```

The obvious implementation is a plain `parsed` field invalidated by `setSavedUsers`. That breaks the moment anything writes the preference by another route — and `SharedPrefManager` exposes `rawPreferences`, so that route exists. Comparing the freshly-read string against the cached one means a write from anywhere invalidates automatically; there is no invalidation path to forget.

`pref.getString` is still called on every access, but after the first load `SharedPreferences` serves from its in-memory map, so what's actually being avoided is the Gson reflection-based parse of a `List<User>` — the expensive half.

**Thread safety**

`@Volatile` on a reference to an immutable holder is the correct minimum: a reader either sees the old holder or the new one, never a half-constructed pair of `raw`/`parsed`. Two threads racing a miss can both parse, but they'd produce equal results and the last write wins — wasted work, never a wrong value. No lock needed for that guarantee.

**`setSavedUsers` priming the cache is correct here, and I checked why**

```kotlin
savedUsersCache = SavedUsersCache(json, users.toList())
```

This caches the caller's objects as if they were the round-trip result, which is only sound if `toJson` → `fromJson` is lossless for `User`. It is: `User` is a five-field data class of nullable `String`s with no transients, no custom adapters, and no computed properties. Gson omits nulls on write and leaves them null on read, so the round trip is an identity. If `User` ever gains a `@Transient` field or a default that differs from `null`, this line becomes a divergence — worth a comment saying the priming depends on that.

`users.toList()` is a defensive copy, so a caller mutating its list after the call can't corrupt the cache. Easy to omit, and it matters.

**Small correctness improvement riding along:** `gson.fromJson<List<User>>(usersJson, userListType) ?: emptyList()` now guards the case where the stored string is the literal `"null"`, which previously would have returned null from a non-null-typed function.

I also confirmed all three production writers (`TeamsRepositoryImpl:939`, `UserRepositoryImpl:1017/1032/1053`) go through `setSavedUsers` rather than touching the preference directly — so the fast path stays warm in practice, and the raw-key comparison is the backstop rather than the primary mechanism.

**Verdict:** Mergeable.

---

Okuro3499 (dismissed):
```
File: app/src/main/java/org/ole/planet/myplanet/services/SharedPrefManager.kt, getSavedUsers() (lines 78-91)                                                                     
    Problem: Returns the cached/parsed List<User> by direct reference instead of a defensive copy, unlike the News.imagesCache pattern it's modeled on. Any in-place mutation of the 
    returned list or its elements corrupts the cache silently, with no re-validation since the raw-JSON comparison won't catch it.                                                   
```

dogi:
**Re-review — Rating: 92/100** (was 89)

`96fdf91` closes a hole I should have caught the first time. My earlier review said "`users.toList()` is a defensive copy, so a caller mutating its list after the call can't corrupt the cache" — that protects the *list* and not the *elements*, and `User` is a data class of five `var`s. Handing out `cache.parsed` directly meant any caller could reach through the returned list and mutate the cached `User` objects in place; the next `getSavedUsers()` would serve the mutation as if it had been read from preferences.

The fix is the right one, in the right place:

```kotlin
private fun List<User>.detach(): List<User> = map { it.copy() }
```

applied on all three boundaries — the cache-hit return, the cache-miss return, and the `setSavedUsers` priming. `User` is five nullable `String` properties and nothing else, so `copy()` is a complete deep copy, not just a shallow one. The KDoc states the reason rather than leaving the `map { it.copy() }` looking like an accident, which is what makes it survivable by the next reader.

**Tests**

Three added, one per aliasing path, and each actually exercises the hole rather than restating the happy path: clearing the returned list (`as MutableList` cast — ugly, but it's the only way to prove the point), mutating a returned element, and mutating both the caller's list and its elements after `setSavedUsers`. Together they pin the invariant from every direction a caller can reach it.

**Still holds from the previous review**

Keying the cache on the raw JSON string rather than invalidating from `setSavedUsers` remains the load-bearing design decision — `rawPreferences` is exposed, so a write can arrive from outside this class, and the string comparison catches it with no invalidation path to forget. `@Volatile` on a reference to an immutable holder is still the correct minimum for safe publication. All five checks green.

**Note, not a blocker**

The cache is no longer allocation-free on a hit: every `getSavedUsers()` now builds a fresh `ArrayList` of fresh `User`s. That's the correct trade — N small copies is far cheaper than a reflective Gson parse of `List<User>` — but it does mean the win is "skip the parse", not "skip the work". And `copy()` is shallow by definition: if `User` ever gains a `MutableList` or another mutable-typed field, `detach()` stops detaching and the KDoc's claim quietly becomes false. The comment names the assumption, which is the best available guard short of making `User`'s properties `val`.

---

Okuro3499 (approved):

dogi:
**Review — rating: 90/100**

Solid optimization with the sharp edges actually filed off. The cache is keyed on the raw JSON string and re-validated against `pref.getString` on every read, so a `List<User>` is parsed once per distinct stored value instead of on every call.

Okuro3499's earlier finding — that handing back the cached list by reference lets any in-place mutation silently corrupt it — is properly fixed rather than patched around. `detach()` copies the elements too (`map { it.copy() }`), which matters because `User` is a data class with `var` properties, so a shallow list copy alone would still have leaked the cached entries. It's applied in both directions: on the way out of `getSavedUsers` and on the way in at `setSavedUsers`, so a caller mutating the list it just passed in can't reach the cache either.

**On CodeRabbit's request for a shared lock — I don't think it's needed, and I'd rather you didn't add one.** I walked the interleavings, and the raw-string validation makes the cache self-correcting:

- Writer stores prefs then cache; a concurrent reader that sees the new prefs but the old cache gets `cache.raw != usersJson`, misses, and parses the new value correctly.
- A reader that reads old prefs and then observes the newer cache also misses, and re-caches the older pair — but the *next* read compares against prefs again, sees the mismatch, and reparses.
- Two interleaved writers can likewise leave `savedUsersCache` holding a superseded pair, and the next read corrects it the same way.

In every ordering the worst outcome is a redundant parse; a stale list can't be returned, because the cached entry is only trusted when its `raw` still equals what's in `SharedPreferences`. `@Volatile` gives the visibility needed for that check. A lock would buy nothing here and would put a monitor on a hot path.

The tests are what make this convincing — `verify(exactly = 1) { spyGson.fromJson(...) }` on the repeat-read case and `exactly = 0` after `setSavedUsers` prove the cache actually engages rather than just that the results match, and there's a case each for out-of-band pref writes, list mutation, element mutation, caller-list mutation after `set`, and the absent-key path. The `?: emptyList()` guard on `fromJson` is a good catch too, since Gson returns null for a stored `"null"`.

**Nit:** the new tests use inline fully-qualified names (`io.mockk.spyk`, `any<java.lang.reflect.Type>()`) where imports would do.

**Verdict:** mergeable — Okuro3499 approved at the current head and the PR carries a triple-star tag, so promoting `ready` → `merge`.

---

PR #17224, opened 2026-09-15
life: smoother repository deduping (fixes #17223)

### LINKED ISSUE #17223: drop the redundant Kotlin sort and factor the repeated dedupe chain in LifeRepositoryImpl
## Verified premise

`LifeRepositoryImpl` repeats the same two-step chain on every read (verified at master `be62dfb`):

```kotlin
:72   val items = myLifeDao.getByUserId(effectiveUserId).distinctBy { it.dedupKey() }.sortedBy { it.weight }
:80   return myLifeDao.getByUserId(effectiveUserId).distinctBy { it.dedupKey() }.sortedBy { it.weight }
:85   return myLifeDao.getVisibleByUserId(effectiveUserId).distinctBy { it.dedupKey() }.sortedBy { it.weight }
```

Three verbatim copies of `.distinctBy { it.dedupKey() }.sortedBy { it.weight }`, and `:72`/`:80` are the *same* call on the same DAO method inside one function.

**The `sortedBy` on those three lines is dead work.** Both DAO queries already sort in SQL (`data/room/dao/MyLifeDao.kt`):

```kotlin
@Query("SELECT * FROM my_life WHERE ... ORDER BY weight ASC")
suspend fun getByUserId(userId: String?): List<MyLife>

@Query("SELECT * FROM my_life WHERE ... AND isVisible = 1 ORDER BY weight ASC")
suspend fun getVisibleByUserId(userId: String?): List<MyLife>
```

and `distinctBy` preserves encounter order, so the list reaching `sortedBy` is already ascending by weight. Each of those three lines allocates a second list and runs a full sort to produce the order it was handed.

## Do

1. Add one private helper expressing "dedupe a DAO-ordered result":

```kotlin
private fun List<MyLife>.dedupedByKey(): List<MyLife> = distinctBy { it.dedupKey() }
```

2. Use it at `:72`, `:80` and `:85`, **without** a trailing `sortedBy`:

```kotlin
val items = myLifeDao.getByUserId(effectiveUserId).dedupedByKey()
```

3. `:72` and `:80` are now the identical expression inside `getMyLifeByUserId`. Hoist the repeated read into a local function or reuse, so the DAO is not called twice through two spellings of the same thing — keeping the existing control flow (early return when `items.isNotEmpty() || defaultItems.isEmpty()`, then the seed attempt, then the re-read).

## Constraints

- **The helper must NOT sort, and it must only be applied to results of `getByUserId` / `getVisibleByUserId`.** Dropping the Kotlin sort is safe *only* because those two `@Query` strings carry `ORDER BY weight ASC`. If the helper is later used on an unordered source the ordering breaks silently and the My Life screen renders in arbitrary order. Name that dependency in a comment on the helper.
- Do **not** add `ORDER BY` anywhere it is missing and do **not** modify `MyLifeDao`. The two relevant queries already have it; that is the premise, not the change.
- Keep `dedupKey()` exactly as it is (`:63-68`) — the `imageId → title → _id → listOf(userId, isVisible, weight)` fallback ladder, in that order. It decides which duplicate survives.
- `distinctBy` keeps the **first** occurrence. Do not swap it for `groupBy`/`associateBy`/`toSet`, which would change which row wins.
- **Leave `:137` alone** (`newItems.distinctBy { it.dedupKey() }.sortedBy { it.weight }` in `seedMyLifeIfEmpty`). Its input is built locally by `items.map { weight++ }`, not returned by a DAO, so the ORDER BY argument does not apply to it.
- **Leave `:110` and `:117` alone.** `:110` sorts items reconstructed from `lifeCache.read`, and `:117` sorts `seeded`, neither of which is guaranteed DAO-ordered.
- `getMyLifeByUserId` must keep its signature, its `defaultItems` default, and its three-branch flow. `getVisibleMyLifeByUserId` stays `private`.
- Do not touch `normalizeUserId`, the `seedMutex`, `LifeCache`, or the `"--"` cache-key convention.

## Tests

- `app/src/test/java/org/ole/planet/myplanet/repository/LifeRepositoryImplTest.kt` must pass unedited.
- Add a case feeding `getByUserId` a list that is **already ascending by weight** with a duplicate `dedupKey` in it, and assert the returned order and the surviving row are unchanged from today. This is the regression guard for removing the sort: it fails if `distinctBy` is swapped for something order-losing.
- Add a case asserting the DAO read in `getMyLifeByUserId` happens the expected number of times for the early-return path (`coVerify(exactly = 1) { myLifeDao.getByUserId(any()) }` when `items.isNotEmpty()`), so the hoist in step 3 cannot quietly turn one query into two.

## Out of scope

- `MyLifeDao.getByIds`' unchunked `IN (:ids)`.
- `getMyLifeForDashboard`'s cache path and its `MyLife(...)` reconstruction.
- `updateMyLifeListOrder`'s `idToIndex`/`mapNotNull` reweighting.
- `updateVisibility`'s `_id OR imageId OR title` matching in the DAO.

dogi:
Removes redundant in-memory sortedBy calls on DAO results that already carry ORDER BY weight ASC in SQL. Factors repeated deduplication into private extension helper dedupedByKey() and hoists DAO read in getMyLifeByUserId.

Fixes #17223

---
*PR created automatically by Jules for task [9098944398210386968](https://jules.google.com/task/9098944398210386968) started by @dogi*

## Summary by CodeRabbit

- **Bug Fixes**
  - Removed duplicate learning-life entries from user results.
  - Preserved the original weight-based order and first occurrence when duplicates are found.
  - Improved handling of existing entries to avoid unnecessary repeated data retrieval.

- **Tests**
  - Added coverage for duplicate removal, ordering, and efficient data retrieval behavior.


dogi:
**Review — rating: 88/100**

Drops `.sortedBy { it.weight }` from three call sites on the grounds that the DAO already orders, and factors the repeated dedupe expression into one place. Both halves check out.

**The sort really is redundant**

The claim rests on a contract in another file, so I verified it against `MyLifeDao`:

```sql
… ORDER BY weight ASC   -- getByUserId
… AND isVisible = 1 ORDER BY weight ASC   -- getVisibleByUserId
```

Both carry it. And `distinctBy` returns a subsequence in encounter order, so a dedupe over an ascending list stays ascending — the subsequent `sortedBy` was sorting an already-sorted list.

Worth noting the ordering interaction is preserved in the other direction too. The old code deduped *then* sorted; the new one deduped only. Those agree because the input is ascending: the first occurrence of any duplicate key is the lowest-weight one, both before and after. If the DAO were unordered the two would genuinely differ in which duplicate survives, which is exactly why the precondition matters.

The saving isn't only the list copy — `sortedBy` re-invokes its selector on every comparison, so this removes O(n log n) property reads per call as well. Same observation that came up on #17119.

**The KDoc is the right call**

```kotlin
/**
 * Deduplicates items by key while preserving existing list order.
 * MUST only be applied to results from [MyLifeDao.getByUserId] or
 * [MyLifeDao.getVisibleByUserId] which carry SQL `ORDER BY weight ASC`.
 */
```

Correctness now depends on a query in a different file, and nothing in the type system says so. Writing the precondition down — with links, so it survives a rename — is what makes this safe to merge rather than a latent trap. The helper being a `private` extension confines the blast radius to this one class, so the MUST is enforceable by reading one file.

**The extraction fixes a drift hazard, not just duplication**

`getMyLifeByUserId` had the same `getByUserId(...).distinctBy{...}.sortedBy{...}` expression written twice — once before the seeding attempt and once after. Two copies of a query-plus-post-processing chain that must stay identical is how the pre-seed and post-seed paths quietly diverge. `loadItems()` makes that impossible.

The early return on non-empty `seeded` is untouched, and `seeded` was never sorted by the old code either, so there's no behaviour change on that branch.

**Nit:** `dedupKey()`'s final fallback is `listOf(userId, isVisible, weight)`, which means two items with no `imageId`, `title` or `_id` and identical weight collapse into one. Pre-existing and not this PR's to fix, but it's the one case where the dedupe is lossy, and it's now one line above a comment explaining the ordering contract — a good spot for a second line explaining that one.

**Verdict:** Mergeable.

---

J-S-webskas (approved):

Okuro3499 (approved):

PR #17229, opened 2026-09-15
life: less health alert personal binding is more (fixes #17226)

### LINKED ISSUE #17226: delete the dead onCreate and the never-read alertMyPersonalBinding in MyHealthFragment
## Verified premise

Two pieces of `MyHealthFragment` do nothing (verified at master `db8b0f1`, post-#17000, `ui/health/MyHealthFragment.kt`).

**1. A super-only `onCreate` override** (`:79-81`):

```kotlin
override fun onCreate(savedInstanceState: Bundle?) {
    super.onCreate(savedInstanceState)
}
```

No body beyond the `super` call, so the override changes nothing — removing it leaves the identical behaviour, since `Fragment.onCreate` runs regardless.

**2. A binding that is inflated and never read.** `alertMyPersonalBinding` has exactly **two** references in the entire file:

```kotlin
:67    private lateinit var alertMyPersonalBinding: AlertMyPersonalBinding
:91    alertMyPersonalBinding = AlertMyPersonalBinding.inflate(LayoutInflater.from(context))
```

Declared, assigned, never used. And `:91` sits in `onViewCreated`, so **a whole dialog layout is inflated on every view creation and immediately discarded** — this is not just dead code, it is per-navigation work. Contrast `alertHealthListBinding` on the next line, which has nine references and is genuinely in use.

## Do

1. Delete the `onCreate` override at `:79-81`.
2. Delete the `alertMyPersonalBinding` field (`:67`) and its assignment (`:91`).
3. Drop the now-unused `import org.ole.planet.myplanet.databinding.AlertMyPersonalBinding` (`:36`).
4. Drop the `android.view.LayoutInflater` import **only if** `LayoutInflater` is not used anywhere else in the file after step 2 — check before removing it.

## Constraints

- **Do not touch `alertHealthListBinding`** (`:68`). It is the nullable sibling declared on the very next line and it has nine usages. The two names differ by a few characters; deleting the wrong one breaks the health-list dialog.
- Do not remove or alter `onViewCreated` itself, and do not move any other statement out of it. Only the one `alertMyPersonalBinding` assignment line goes.
- Do not change the ordering of anything remaining in `onViewCreated` — `setupRealtimeSync()` stays where it is, immediately before the deleted line.
- Leave the `res/layout` file behind `AlertMyPersonalBinding` in place. Other code may inflate it; this issue removes a dead *reference*, not a layout.
- `MyHealthFragment` extends `BaseBindingFragment<FragmentVitalSignBinding>` after #17000 — do not alter that declaration or the `FragmentVitalSignBinding::inflate` argument.

## Tests

- The existing `MyHealthFragment` tests, if any, must pass unedited. Do **not** add a Robolectric harness for a dead-code deletion — the compiler is the check here, and an unused `lateinit` removal cannot change behaviour.
- If a test currently references `alertMyPersonalBinding` (it should not, the field is private), update it rather than reintroducing the field.

## Out of scope

- The `allowDateEdit` flag at `:93` and the `if (allowDateEdit)` block it gates. That is a deliberately disabled feature, not dead code, and turning it off for good is a product decision.
- `getDisplayName` and its list allocations.
- `setupRealtimeSync`, `refreshHealthData`, and the `searchJob` cancellation.
- Every other fragment touched by #17000.

dogi:
Removed the dead `onCreate` override and the unused `alertMyPersonalBinding` property, inflation call, and import in `MyHealthFragment`.

Fixes #17226

---
*PR created automatically by Jules for task [16725281695644048489](https://jules.google.com/task/16725281695644048489) started by @dogi*

## Summary by CodeRabbit

- **Bug Fixes**
  - Health information now refreshes automatically when real-time synchronization updates the selected patient.
  - Improved health-list loading behavior helps ensure the latest available information is displayed when viewing patient health details.


dogi:
**Review — rating: 90/100**

Seven deletions, zero insertions. Both removals check out, and one of them was doing real work for no reason.

**The `onCreate` override**

```kotlin
override fun onCreate(savedInstanceState: Bundle?) {
    super.onCreate(savedInstanceState)
}
```

A pure delegation to super — identical behaviour whether it exists or not. Worse than inert, really: an override like this reads as a hook someone intended to fill, so the next person touching the file has to check whether something was lost.

**`alertMyPersonalBinding` was inflating a layout into the void**

This is the part worth having. The field was `lateinit var`, assigned in `onViewCreated`:

```kotlin
alertMyPersonalBinding = AlertMyPersonalBinding.inflate(LayoutInflater.from(context))
```

and never read again. I checked the whole file on this branch — zero remaining references. So every time the health screen's view was created, it inflated the entire `alert_my_personal` view hierarchy and immediately discarded it. View inflation is the expensive kind of waste: XML parsing plus view construction, on the main thread, during `onViewCreated`.

**Imports**

`LayoutInflater` correctly stays — it's still used at line 276 for `AlertHealthListBinding.inflate`. `Bundle` stays too, since `onViewCreated` takes one. Only the now-unused `AlertMyPersonalBinding` import goes. A removal that leaves a dangling import behind is the usual sloppiness here and it's avoided.

**Nit:** the surviving `alertHealthListBinding` is a nullable `var` while the removed one was `lateinit` — worth noting only because the nullable form is the safer pattern for a binding tied to a dialog's lifetime, so the file is now consistently on the better one.

**CI:** all five checks green. The `less` label is correct — zero additions, some deletions.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17234, opened 2026-09-15
all: smoother storage breakdown view modelling (fixes #17227)

### LINKED ISSUE #17227: hoist the storage-scan state out of StorageBreakdownFragment into a ViewModel
## Verified premise

`StorageBreakdownFragment` owns both the filesystem scan and its results (verified at master `db8b0f1`, post-#17000, `ui/settings/StorageBreakdownFragment.kt`).

The scan result lives in **mutable fields on a fragment-held list** (`:51-54`, `:58-60`):

```kotlin
internal data class CategoryData(
    @StringRes val nameRes: Int,
    val extensions: Set<String>,
    var sizeBytes: Long = 0,
    var fileCount: Int = 0
)

internal val categories: List<CategoryData> = StorageCategories.all.map {
    CategoryData(it.nameRes, it.extensions)
}
```

and the scan writes into them in place (`:208-211`):

```kotlin
categories.forEachIndexed { index, category ->
    category.sizeBytes = result.sizes[index]
    category.fileCount = result.counts[index]
}
```

Because that state is a fragment field, it dies with the view. The whole `scanStorage` walk over the ole directory reruns on every view creation — including a rotation, where nothing on disk has changed. `loadJob?.cancel()` at `:164` confirms the design intent is "restart the scan each time".

This is the standard shape the codebase has been moving away from: the fragment holds a `dispatcherProvider`, launches IO, and mutates its own state (`:172-201`), while sixteen other screens already put exactly this behind a ViewModel.

## Do

1. Add `ui/settings/StorageBreakdownViewModel.kt`, `@HiltViewModel`, injecting `DispatcherProvider`.
2. Move into it: the `categories` construction from `StorageCategories.all`, the `scanStorage` functions, the `ScanResult` and `CategoryData` types, and the available-space read.
3. Expose one immutable UI state via `StateFlow` — the available-space string, the total bytes, and a `List<CategoryData>` carrying the scanned values as **`val`s**, not `var`s.
4. Kick the scan from the ViewModel in `viewModelScope`, guarded so it runs **once** per ViewModel rather than once per view creation.
5. `StorageBreakdownFragment` keeps only rendering: collect the state with `collectWhenStarted`, set the three visibility flags, and call `populateCategoryRows` from the collected value. Delete `loadJob`, the `dispatcherProvider` injection, and the mutation loop.

## Constraints

- **`result.sizes[index]` and `result.counts[index]` are positional against `StorageCategories.all`.** The scan fills parallel `LongArray`/`IntArray` by index, so the category list and those arrays must stay in the same order. Do not sort, filter or re-key the categories anywhere between the scan and the render, and do not switch to a map keyed by `nameRes` without proving the mapping first.
- **Make `CategoryData.sizeBytes`/`fileCount` `val` once the state moves.** They are only `var` today so the fragment can mutate them in place; a ViewModel that emits a new immutable list per scan is the point of the change. Keeping them `var` reproduces the bug in a new location.
- Keep `scanStorage(oleDir: File)` `internal` and keep its signature. `app/src/test/java/org/ole/planet/myplanet/ui/settings/StorageBreakdownFragmentTest.kt` calls it directly against a real directory — that test must keep working, retargeted at the ViewModel.
- Keep the render rules exactly: `progressBar` visible until the scan returns; `emptyText` visible and nothing else when `totalBytes == 0L`; `contentLayout` visible otherwise; and `populateCategoryRows` still filtering `fileCount > 0` (`:230`).
- Keep the two string formats byte-identical — `getString(R.string.available_space_colon) + " " + availableSpaceText` (`:175-176`) and `getString(R.string.storage_total_downloaded) + ": " + FileUtils.formatSize(...)` (`:198-199`) — and the singular/plural choice between `R.string.file_count_one` and `R.string.file_count_many` (`:236-239`).
- **Leave the launch site where it is.** The coroutine currently starts in `onCreateDialog` using `viewLifecycleOwner` (`:172`); replace it with a `collectWhenStarted` on the same flow at the same point rather than relocating it to `onViewCreated`. Moving it is a separate lifecycle question and is not part of this task.
- Do not touch `onCreateDialog`'s `BottomSheetBehavior` setup (`STATE_EXPANDED`, `skipCollapsed = true`), and do not change `BaseBindingBottomSheetFragment<FragmentStorageBreakdownBinding>`.
- Do not touch the `WorkManager.enqueue(freeSpaceWork)` call at `:94-98`.

## Tests

- Retarget `StorageBreakdownFragmentTest`'s `scanStorage(File)` cases at the ViewModel and keep every assertion. That test is the specification for the scan.
- Add a ViewModel case asserting the scan runs **once** across two collections of the state flow (`coVerify`/call-count on the scan, or a counter) — that is the actual claim of this change and the thing a rotation would otherwise break.
- Add a case for `totalBytes == 0L` asserting the emitted state is the empty one, and a case asserting the emitted `CategoryData` values line up with `StorageCategories.all` by index.

## Out of scope

- `StorageCategories` and its extension sets.
- `StorageCategoryDetailFragment`.
- The `FreeSpaceWorker` enqueue and the `WorkManager` unique-work question, which is parked pending a decision.
- `FileUtils.availableOverTotalMemoryFormattedString` and `FileUtils.formatSize`.

dogi:
Extracted filesystem scan state, ScanResult, CategoryData, and available space calculations into StorageBreakdownViewModel exposed via StateFlow. Refactored StorageBreakdownFragment to delegate state management to StorageBreakdownViewModel and retain only rendering logic. Retargeted and expanded StorageBreakdownFragmentTest to test StorageBreakdownViewModel.

Fixes #17227

---
*PR created automatically by Jules for task [16730559029145289294](https://jules.google.com/task/16730559029145289294) started by @dogi*


dogi:
**Review — rating: 87/100**

Hoists the storage scan into a `StorageBreakdownViewModel` behind a `StorageBreakdownUiState`. It does more than relocate code — two things genuinely improve in the move.

**`CategoryData` becomes immutable**

Before, `categories` was a fragment-level `List<CategoryData>` with `var sizeBytes` and `var fileCount`, mutated in place after each scan:

```kotlin
categories.forEachIndexed { index, category ->
    category.sizeBytes = result.sizes[index]
    category.fileCount = result.counts[index]
}
```

Now the fields are `val` and the scan produces fresh instances through `.copy()`, carried on the state object. So the rendered list is a snapshot rather than a shared mutable structure the render path reads while a background scan writes to it. That's the part I'd keep even if the ViewModel move were dropped.

**The load is started correctly**

```kotlin
init { loadStorage() }
```

with `onViewCreated` only setting up `collectWhenStarted(viewModel.uiState)` and no load call of its own. Worth calling out because #17218 in this same queue had to be sent back for exactly the opposite — a load kicked off from `onViewCreated` that accumulated a collector per view creation. This one gets it right first time.

The `hasScanned` guard plus `forceRefresh = true` at the three refresh sites (free-space success/failure/cancel, and the detail-screen result) gives one scan per sheet opening with explicit re-scans where the data actually changed.

**`launch(dispatcherProvider.io)` is right here**

Worth distinguishing from cases where it's redundant: `scanStorage` does `walkTopDown()` with a `length()` per file — genuinely blocking filesystem work, not a Room `suspend` call that manages its own threading. An explicit IO dispatcher is correct, and `_uiState.update` from that dispatcher is fine since `StateFlow` is thread-safe.

`@ApplicationContext` on the injected context is the right choice for a ViewModel that outlives the fragment.

**Four smaller things**

`hasScanned = true` is set before the coroutine launches, so a scan that's cancelled partway leaves the flag set and a later non-forced call no-ops with stale state. Every fragment-side call passes `forceRefresh = true`, so it can't bite today — but setting it in the coroutine after a successful scan would make the guard mean what its name says.

`open class` + `internal open fun scanStorage` exists only so the test can override it. It works, but opening production code for a test is the weaker option; taking a scanner interface as a constructor dependency would give the same seam without the modifier.

`renderUiState` writes `availableSpaceText` before the `isLoading` early return, so during the scan the label reads "Available space: " with nothing after it. Previously the TextView stayed empty until the value arrived. Brief and the spinner is up, but a dangling label reads worse than a blank one — moving that assignment below the loading branch, or gating it on a non-empty value, avoids it.

`ScanResult` is a `data class` holding `LongArray`/`IntArray`, so its generated `equals`/`hashCode` compare array identity rather than contents. It's `internal` and never compared, and it came across verbatim, so nothing is broken — just the kind of thing that bites later if someone assertEquals two of them.

**Verdict:** Mergeable.

---

PR #17236, opened 2026-09-16
all: smoother notifications unread count view modelling (fixes #17235)

### LINKED ISSUE #17235: delete the dead markAsRead overload and factor the duplicated count-then-update in NotificationsViewModel
## Verified premise

Two findings in `ui/notifications/NotificationsViewModel.kt`, verified at master `f4fb54a` (post-#17206).

**1. A dead private overload.** Two extensions sit next to each other (`:186-191`):

```kotlin
private fun List<Notification>.markAsRead(id: String): List<Notification> {
    return map { if (it.id == id && !it.isRead) it.copy(isRead = true) else it }
}
private fun List<Notification>.markAsRead(ids: Set<String>): List<Notification> {
    return map { if (it.id in ids && !it.isRead) it.copy(isRead = true) else it }
}
```

The **`String` overload has zero callers.** Verified across all of `app/src`: the only list-receiver calls are `:118` and `:176`, both passing `markedIds`, a `Set<String>`, which bind to the `Set` overload. The public `fun markAsRead(notificationId: String)` at `:141` — the one `NotificationsFragment:58` and `:137` call — does its own inline `mapNotNull` at `:147` and never touches the extension.

This is the dangerous shape of dead code: an unused overload sharing a name with a used one, where a future single-id caller would silently bind to it.

**2. The same count-then-update shape, written twice.** `markSelectedAsRead` (`:113-118`) and `deleteSelected` (`:133-134`) both read `_notifications.value` for a count and then call `_notifications.update` for the transform:

```kotlin
val wasUnreadCount = _notifications.value.count { it.id in markedIds && !it.isRead }
_notifications.update { currentList -> ... }
_unreadCount.update { maxOf(0, it - wasUnreadCount) }
```

Beyond the duplication, that `.value` read and the `update` are two separate operations on the same `MutableStateFlow`, so the count is taken against a snapshot the update does not necessarily see.

## Do

1. Delete the `markAsRead(id: String)` extension at `:186-188`. Leave the `Set<String>` overload exactly as it is.
2. Factor the shared shape in `markSelectedAsRead` and `deleteSelected` into one private helper that takes the affected ids and a transform, performs a **single** atomic read-and-replace, and returns the number of previously-unread rows it affected, so the caller can feed `_unreadCount.update { maxOf(0, it - affected) }`.

Use `getAndUpdate`, which returns the previous value:

```kotlin
private fun applyAndCountUnread(
    ids: Set<String>,
    transform: (List<Notification>) -> List<Notification>
): Int {
    val previous = _notifications.getAndUpdate(transform)
    return previous.count { it.id in ids && !it.isRead }
}
```

`markSelectedAsRead` passes the `currentFilter == "unread"` branch as its transform; `deleteSelected` passes `{ it.filterNot { n -> n.id in deletedIds } }`.

## Constraints

- **Do NOT compute the count inside `_notifications.update { }`.** `MutableStateFlow.update` is a compare-and-set loop and may invoke its lambda **more than once** under contention; a `var` accumulated inside it can be counted twice, which would subtract too much from `_unreadCount` and is exactly the kind of bug that only shows under load. That is why this issue prescribes `getAndUpdate` — the count comes from the returned previous value, outside the lambda.
- The count predicate must stay `it.id in <ids> && !it.isRead`. Dropping the `!it.isRead` half would decrement `_unreadCount` for rows that were already read.
- Keep `_unreadCount.update { maxOf(0, it - wasUnreadCount) }` with the `maxOf(0, ...)` floor in both functions.
- Keep both `if (ids.isEmpty()) return` guards and both `if (<result>Ids.isNotEmpty())` guards, and keep `_selectedIds.value = emptySet()` firing only inside the non-empty branch.
- Keep the `currentFilter == "unread"` branch in `markSelectedAsRead`: filtered-out when the filter is `unread`, marked-in-place otherwise. `deleteSelected` has no such branch — do not add one.
- **Do not touch `markAsRead(notificationId: String)` at `:141-166` or `markAllAsRead` at `:168-184`.** `markAsRead`'s inline `mapNotNull` with the `wasUnread` flag looks like a third instance of this shape but it is not: it returns `null` to drop a row only in the `unread` filter, tracks a single boolean rather than a count, and decrements by one under a different guard (`wasUnread && _unreadCount.value > 0`). Folding it in would change behaviour.
- Do not touch `buildNotificationGroups`, `buildGroupedList`, the `combine` chain at `:60-64`, or `isGroupDefaultExpanded`.

## Tests

- Existing `NotificationsViewModelTest` cases must pass unedited.
- Add a case for `markSelectedAsRead` where some selected rows are already read, asserting `_unreadCount` drops by the count of the **previously unread** ones only.
- Add the same for `deleteSelected`.
- Add a case under the `unread` filter asserting the marked rows are removed from `_notifications` **and** `_unreadCount` still drops by the right amount — that is the branch where the transform and the count disagree about what they are looking at.
- Assert `_unreadCount` never goes below zero when the stored count is smaller than the number affected.

## Out of scope

- `markAsRead(notificationId)` and `markAllAsRead`, per the constraint above.
- `NotificationsRepository.markNotificationsAsRead` / `deleteNotifications` and their DAO queries.
- The grouping and `NotificationListItem` construction.
- `TYPE_ORDER` and `KNOWN_TYPES`.

dogi:
Deletes dead markAsRead(id: String) overload and factors count-then-update logic in NotificationsViewModel using getAndUpdate to ensure atomic read-and-replace state flow updates. Adds comprehensive unit tests.

Fixes #17235

---
*PR created automatically by Jules for task [17791989438383071929](https://jules.google.com/task/17791989438383071929) started by @dogi*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved notification selection actions to keep unread counts accurate when marking notifications as read or deleting them.
  * Prevented unread counts from becoming negative.
  * Ensured notifications marked as read are removed from unread-filtered views.

* **Tests**
  * Added coverage for partial-read selections, filtered notification lists, and unread count safeguards.


dogi:
**Review — rating: 89/100**

Two changes that read as tidying. The second one is actually a race fix, and I'd lead the description with that.

**`getAndUpdate` closes a real window**

The old shape read the flow, then updated it, as two separate operations:

```kotlin
val wasUnreadCount = _notifications.value.count { it.id in markedIds && !it.isRead }
_notifications.update { currentList -> … }
```

Anything mutating `_notifications` between those two lines — another selection action, a refresh landing — leaves `wasUnreadCount` describing a list that is no longer the one the transform ran against. `_unreadCount` then gets decremented by a number that doesn't correspond to what actually changed, and the badge drifts. It's a narrow window, but it's on the path where a user taps "mark selected as read" while a sync is writing.

```kotlin
val previous = _notifications.getAndUpdate(transform)
return previous.count { it.id in ids && !it.isRead }
```

`getAndUpdate` returns the value the successful compare-and-set actually replaced, so the count is computed from precisely the snapshot the transform consumed. Count and transformation can no longer disagree.

**The CAS-retry detail checks out**

`MutableStateFlow.update`/`getAndUpdate` run a compare-and-set loop, so the transform can be invoked more than once under contention. That makes it unsafe for anything with side effects. Both transforms here are pure — `filterNot`, and `markAsRead` which is `map` + `copy` — so re-invocation is harmless. Worth knowing the helper only stays correct as long as callers keep passing pure lambdas; a one-line comment on `applyAndCountUnread` saying so would protect it.

One thing the change doesn't fix, and can't: `_notifications` and `_unreadCount` are separate flows, so they still update in two steps. A collector can observe the list updated and the count not yet. That's inherent to the two-flow shape rather than something this PR introduced — folding both into one state object would be the fix, and is a bigger change than this.

**The dead overload really was dead**

I checked rather than trusting the compiler here, because the name is misleading: `markAsRead` appears eleven times across the branch, and a grep makes the removed extension look live. The surviving uses are `NotificationDao.markAsRead(String)` (a DAO method), the public `NotificationsViewModel.markAsRead(notificationId)` called from the fragment, and `currentList.markAsRead(markedIds)` at two sites — which is the *plural* extension that stays. The `private fun List<Notification>.markAsRead(id: String)` overload had zero callers. That name shadowing is probably why it survived earlier dead-code passes.

**Nit:** `applyAndCountUnread(deletedIds) { it.filterNot { n -> n.id in deletedIds } }` names `deletedIds` twice — once for the count filter, once inside the transform. Unavoidable given the signature, but it does mean a caller could pass one set and filter on another without the compiler noticing.

**CI:** build and both test shards were still `in_progress` when I reviewed; only `myPlanet labels` had reported (green). Nothing failing — just no verdict yet. The removal is the only compile risk and I confirmed it locally, but worth a glance at the checks before merging.

**Verdict:** Mergeable.

---

Okuro3499 (approved):

PR #17244, opened 2026-09-16
dashboard: smoother bell star coloring (fixes #17243)

### LINKED ISSUE #17243: icons and stars not visible in darkmode
Image

Okuro3499:
fixes #17243
Screenshot_20260916_165636


dogi:
**Review — rating: 90/100**

The root cause here is more specific than "poor contrast in dark mode", and the fix is shaped correctly around it.

**Why the star was invisible**

`setColor` tinted the certified star with `R.color.colorPrimary`. In `values/` that resolves to `@color/md_blue_700`, which is fine — but `values-night/colors.xml` line 3 redefines `colorPrimary` as **`#18191a`**, a near-black used for dark-theme surfaces. So in dark mode the star was being painted near-black on a near-black background. Not low contrast — effectively no contrast, by inheriting a token the night theme deliberately repurposes.

Decoupling the star from `colorPrimary` into `dashboard_star_filled` / `dashboard_star_empty` is exactly right: those are semantic tokens for one purpose, so the night theme can give them visible values (`#64B5F6` / `#90A4AE`) without touching what `colorPrimary` means elsewhere.

**The chip icons: the alias is deliberate, not redundant indirection**

My first read was that `dashboard_chip_icon_*` is unnecessary — why not just add night overrides to `dashboard_tile_*`? Because those colours have a second job:

```
layout/home_card_library.xml:26     android:background="@color/dashboard_tile_library"
layout-sw600dp/home_card_courses.xml:25  android:background="@color/dashboard_tile_courses"
```

…and four more. `dashboard_tile_*` are **tile backgrounds** on the home cards, where the saturated day colour is wanted in both themes. Adding a night override there would have lightened those card backgrounds as a side effect nobody asked for. Introducing a separate icon-tint token and overriding only that is the careful choice, and the light-mode values alias straight back (`@color/dashboard_tile_library` etc.) so day appearance is provably unchanged.

The night values are the Material 300/400 range of the same hues — `#EF5350`, `#FFB74D`, `#66BB6A`, `#BA68C8` against `#D32F2F`, `#F57C00`, `#388E3C`, `#7B1FA2` — so the icons stay recognisably red/orange/green/purple rather than being flattened to a single accent.

**It quietly fixes two more**

`dashboard_avatar_ring` and `dashboard_accent_blue` also gain night entries. Neither is referenced by the rest of this diff, so they weren't part of the reported bug — but both were defined only in `values/` (`#0D47A1`, `#1976D2`, both dark navy) and therefore fell through to those dark values on a dark background. Worth a line in the PR description; they're a real fix and currently look like stray additions.

**One thing worth confirming by eye**

`dashboard_star_empty` is `#90A4AE` in *both* themes. That's Material Blue Grey 300, chosen to match the old `md_blue_grey_300`, so light mode is unchanged — but a mid-grey that reads as "empty" against a light card can read as "filled-ish" against a dark one, where `#64B5F6` is the only thing distinguishing it. Worth a look on a dark device to check the filled/empty distinction still lands; a slightly darker empty state in night would widen the gap if it doesn't.

**Verdict:** Mergeable.

---

dogi (approved):

PR #17247, opened 2026-09-16
sync: smoother transaction semaphore managing (fixes #17246)

### LINKED ISSUE #17246: Unbounded concurrent attachment downloads
TransactionSyncManager.kt:328-377 launches one coroutine per doc for CV/team/course-cover attachments with no cap. A single 1000-doc
sync batch can fire ~1000 simultaneous downloads. Every upload path in the codebase caps concurrency at Semaphore(6) (UploadCoordinator.kt:129); the download side should match.

Okuro3499:
fixes #17246
Add a semaphore to `TransactionSyncManager` so CV, team, and course-cover attachment downloads run with a capped level of parallelism. This avoids overwhelming sync with too many simultaneous attachment fetches while still keeping batch downloads concurrent.


dogi:
**Review — rating: 86/100**

Correct, minimal, and it matches an idiom that already exists in this codebase. `SyncManager` gates its shelf checks with `Semaphore(8)` and its shelf processing with `Semaphore(6)`, both as scope-local semaphores inside a `coroutineScope { … withPermit { … } }` — this PR applies exactly that shape to the three attachment fan-outs in `TransactionSyncManager`, and improves on the originals by naming the bound instead of inlining a literal.

**Verified**

- Each `downloadXFromBatch` is a `suspend fun … = coroutineScope`, so the batch already awaited its own downloads before the paging loop advanced. The unbounded part was *within* a batch: up to `pageSize` `launch`es hitting the network at once. `withPermit` caps that at 6 without changing the await semantics.
- Acquiring the permit *inside* `launch` rather than around it is the right order — the coroutines are created immediately (cheap) and suspend on the permit, so no batch is serialized behind permit acquisition at the dispatch site.
- `inProgress.add(resumeFileName)` in the CV path still runs on the loop's thread, before `launch`, so the dedup guard is unaffected by the semaphore.
- All five checks green.

**Notes, not blockers**

1. `MAX_CONCURRENT_ATTACHMENT_DOWNLOADS` reads like a global cap, but each of the three functions allocates its own `Semaphore`, and `SyncManager` runs `achievements`, `teams` and `courses` concurrently in the same `parallelTables` fan-out. The real ceiling during a full sync is therefore 18, not 6. Still bounded, and still a large improvement over the previous behaviour — but a `@Singleton`-scoped field next to the existing `dbWriteMutex` would make the name true and cost one line. At minimum the constant deserves a comment saying it is per-table.
2. The `companion object` sits between `dbWriteMutex` and `tableSyncHandlers`. The rest of the file keeps its properties contiguous; moving it to the bottom of the class (or making it a private top-level `const`) would read better.
3. No test. Concurrency ceilings are awkward to assert, though a fake `ApiInterface` tracking peak in-flight calls would do it. The existing `Semaphore` sites in `SyncManager` aren't covered either, so this isn't a regression in discipline.

---

dogi (approved):

PR #17248, opened 2026-09-16
Phase 158 Lane 1: harvest master into the port

dogi:
Lane 1 of the Phase 158 parallel round.

## Scope

Merge `origin/master` into the migration branch and triage every commit landed since the branch point, carrying each Follow into the Flutter port.

**14 commits, not the 13 in the brief** — `3152c26 sync: smoother time upload crash logging (fixes #17200) (#17221)` landed after the brief was written and is triaged with the rest.

Every subject is `smoother X`, the wording Phase 95 dismissed as refactors and Phases 96/126/156 each found real behaviour hiding under. Every one also carries a `fixes #NNNN`, so the prior here is higher than a usual harvest batch. Two passes per the standing rule: `harvest-triage` first, then a hand re-read of every diff against the Kotlin.

Status: merge pushed, triage in flight. This body will be updated as the round lands.

## Lane boundaries

Lane 2 owns `flutter/lib/providers/**` (except `feedback_provider.dart`). Lane 3 owns `login_screen.dart`, `feedback_create_screen.dart`, `feedback_provider.dart`, `router.dart`, `redirect_reachability_test.dart`. Follows landing in those files are reported precisely rather than edited here.

## Kotlin CI

A harvest push carries real changes under `app/`, so `build.yml` and `test.yml` both run rather than being skipped by `paths-ignore`. Both flavours building and the Kotlin unit tests passing is the check this round exists to run.

https://claude.ai/code/session_012L8Mduai8Qzyspxvx9qsGK

---
_Generated by [Claude Code](https://claude.ai/code/session_012L8Mduai8Qzyspxvx9qsGK)_


dogi:
Closing unmerged. This lane was stopped ~8 minutes in and replaced by #17251, because the mechanism it was built on does not hold up:

- master runs **110–129 commits/week**, so the 13-commit gap this lane was briefed on was **under 20 hours** of drift — a commit-walk can never catch up, and what it merges is stale on landing;
- Follow yield is flat in batch size (Phase 126: 18 commits → 2 Follows; Phase 156: 136 → 2), so harvesting more often costs more per finding, not less;
- and none of those Follows actually needed the history. Phase 156's `LIKE` escaping hole and missing `ORDER BY` were both visible by comparing the two implementations as they stand; Phase 96 found the ranked-search gap by reading `ResourcesSearchUtils` as *the* resource search, not by reading a diff. Phase 95 read that same batch as diffs and found nothing.

The replacement compares current state to current state and gets the "what changed since anyone last looked?" signal from a ledger over Kotlin's `@Query` annotations instead of from commit subjects.

It has already paid: merging master (by then 26 commits) turned the Flutter gate red on exactly one thing — `NewsDao.getTopLevelTeamMembership`, new upstream, where `NotificationsRepositoryImpl:392` now batches the team chat badge and leaves `countTopLevelByTeam` with no caller in `app/src/main`. Triaged as not a Follow and recorded. One query read instead of 26 commit diffs.

The mechanical `master` merge now happens at integration time, which is the part of a harvest that was actually earning its keep.

---

PR #17251, opened 2026-09-16
Phase 158 Lane 1 — query-semantics differential audit (Kotlin @Query vs Drift builders)

dogi:
## Phase 158 Lane 1 — query-semantics differential (Kotlin `@Query` vs Drift builders)

**Not a harvest.** No commit history was read; `master` was not merged. Kotlin's **312 `@Query` annotations across 37 DAO files** were compared against the port's Drift builders as *current state*, prioritised by consequence: `LIKE`/escaping → `ORDER BY` and collation → `IS` vs `=` on nullables → `LIMIT` → predicate shape.

Gate: `dart format` clean, `flutter analyze` clean, **`flutter test` 3248 passed, exit 0** (checked directly, not through a pipe).

### Files touched
`flutter/lib/data/local/app_database.dart` · `flutter/lib/repository/notifications_repository.dart` · `flutter/lib/repository/configurations_repository.dart` · 5 new test files + 5 existing ones updated. Nothing outside `lib/data/local/**`, `lib/repository/**` and tests.

---

## Fixed

**1. The dashboard team-chat badge counted a population its own watermark doesn't describe.** `getTeamNotifications` read `NewsDao.teamChatCounts` — the port of `countTeamChats`, which Kotlin uses for this badge nowhere, and whose only wrapper (`VoicesRepository.countTeamChats`) has **no caller in `app/src/main` at all**. Kotlin's badge is `countTopLevelByTeam`, *the same statement* the team voices feed loads with, which is what makes `watermark < count` mean anything. The two populations are near-disjoint, and **nothing in either app writes `viewableBy`/`viewableId` on a top-level post it authors** — the composer writes `viewIn`. It failed in both directions: never lit for a team whose posts were all composed in-app, permanently lit for one carrying a server-authored post. `NewsDao` gains `countTopLevelByTeam`/`watchTopLevelByTeam` (both arms, `COLLATE NOCASE`, the `replyTo` test, `teamIdPattern`'s escaping); `teamChatCounts` is **removed** rather than left as a second counter to wire back in.

`updateTeamNotification` now **derives** the watermark from the same query instead of taking the caller's count — which is the port of Kotlin, whose `updateTeamNotification(teamId, news)` receives the list `getFilteredNews` just loaded with that statement. My own second audit pass caught that switching only the badge made a reachable case *worse*: with in-app posts **and** one synced `viewableBy: 'teams'` post the dot went from dark to unclearable. Deriving closes it without depending on the feed.

**2. `TeamDao.watchMemberCount` was three guards short.** `TeamDao.kt:29` is `COUNT(DISTINCT userId) … AND userId IS NOT NULL AND EXISTS (SELECT 1 FROM users u WHERE u.id = teams.userId OR u._id = teams.userId)`; the port counted every membership row. It is drawn as "N members" and gates leave-team on `> 1`, so over-counting offered Leave on a team the user is alone in — the case `TeamDetailFragment:180-185` refuses. `isDeletePending` is deliberately **not** reproduced: the port hard-deletes on leave, so it has no such state.

**3. A team resource link Planet stored any way but one was invisible.** `TeamDao.getResourceIdsByTeamId` accepts four `docType` shapes (NULL, blank, `resourceLink`, `link`); the port read one. Kotlin's non-blank `resourceId` guard comes with them and is load-bearing *because* of them.

**4. The My teams card counted enterprises.** `getMyTeamsFlow` ends `(type == "team" || type.isNullOrBlank())`; `teamsByIds` had no `type` predicate, so a user in one team and one enterprise saw 2 tiles — and `teamNotificationsProvider` reads the same list, so an enterprise got a chat badge too.

**5. `MyLibraryDao.getTeamPrivate` — an arm nothing read.** `getTeamResources` is *links ∪ private*, de-duplicated. The port had only links, and nothing in `lib/` read `privateFor`. A resource added from a team screen (where `isPrivate` defaults true) is excluded from the catalog, from My Library **and** from the team tab — **in no view of the app** until an upload links it, and permanently invisible if pulled from a server holding no `resourceLink` document. DAO half landed; the union is a tripwire (see below).

**6. `ExamDao.getByStepId` was neither of Kotlin's two step lookups** — `getSingleOrNull()` where Kotlin has `getFirstByStepId` (`LIMIT 1`) and `getByStepId` (List), so on a step with two exams it *threw*. No production caller; renamed to `getFirstByStepId` and bounded.

**7. A `_design` row could be read as the community configuration.** `fetchConfiguration` takes `rows[0]` of `configurations/_all_docs` unfiltered and so did the port — inherited parity, which CLAUDE.md is explicit is not a reason to ship it. `_all_docs` returns design documents inline, ordered by raw id, and `_` (0x5F) sorts before lowercase, so on the six uuid first-characters in sixteen beginning `a`–`f` a `_design/…` row is first. **The two apps don't even fail the same way**: Kotlin throws on the missing `code` and reports an unreachable server; the port returned a *success* with a blank code and a `_design/…` id, which `server_config_screen` stores as `_configuredId`. A stated divergence, observable only where Kotlin already fails.

Plus the corrections my own second pass found in this lane's work: `COLLATE NOCASE` instead of a `lower()`/Dart-`toLowerCase()` pair that disagreed with Kotlin in **both** directions; `getTeamNotifications` looping the watermarked teams as Kotlin does; two fixtures that couldn't distinguish what they claimed to pin; two exemption assertions satisfiable by a comment; five off-by-one Kotlin citations; a doc claiming Kotlin filters health-sync users in memory when it filters in SQL.

---

## Reported, not fixed

| # | Where | Kotlin | Change |
|---|---|---|---|
| 1 | `lib/providers/teams_provider.dart` `teamResourcesProvider` | `TeamsRepositoryImpl.getTeamResources:318-323` | Union `myLibraryDao.getTeamPrivate(teamId)` with the link rows, dedup by `id`. **Tripwire: `team_private_resources_test.dart`.** |
| 2 | `lib/providers/voices_provider.dart` `teamVoicesProvider` | `NewsDao.getTopLevelByTeamFlow` (`NewsDao.kt:30-31`) | Replace body with `dao.watchTopLevelByTeam(teamId)`; delete the `jsonDecode` filter and sort. It adds a `docType='message'` predicate Kotlin lacks and drops the `viewableBy` arm. **Tripwire: `team_voices_feed_parity_test.dart`.** |
| 3 | `team_members_screen.dart:162-168` → `teams_provider.dart:283-326` | `RequestsViewModel.kt:75-106`, `TeamDao.getEligibleNextLeaderCandidates` | **Leadership succession is missing.** A sole leader leaving via *Members* is removed with no successor promoted and no last-leader refusal → the team is **leaderless and unrecoverable from in-app**. Note the scoping: `teams_screen.dart:251`'s leave is at parity — Kotlin's `TeamDetailFragment:291-308` path has no succession either. |
| 4 | `teams_provider.dart:402`, `:412` | — | `ref.read(sessionProvider).value?.planetCode` / `?.id` — the standing null trap, on a screen that never watches it. `requestToJoin` ten methods up does it right with `await _session()`. |
| 5 | `lib/` (absent) | `DiagnosticsRepositoryImpl`, `UploadConfigs.CrashLog` | **`apk_log` is entirely unported** — no table, no writer, no uploader, while Kotlin writes from the uncaught-exception handler, download failures and sync summaries and POSTs to `apk_logs`. Operator telemetry, hence the ranking. |
| 6 | `TeamDao` | `TeamDao.teamNameExists:51-52` | No counterpart; the port permits duplicate team names. Kotlin gates create and rename on it, with `COLLATE NOCASE` and `_id != :excludeTeamId OR :excludeTeamId IS NULL`. |
| 7 | `UserDao.pendingSyncUsers` | `UserDao.kt:31-32` | Kotlin is `ORDER BY id LIMIT :limit` with 100 from `UploadToShelfService`; the port has neither. *More* complete, but an unrecorded deviation and a non-deterministic order — **do not add a LIMIT without the ORDER BY.** |
| 8 | `teams_repository.dart` `createLocalResourceLink` | `TeamsRepositoryImpl.kt:709-710` | The prefs planet-code fallback still has no counterpart. The existing doc comment's reasoning survived checking and I did **not** change it: adding `PlanetPrefs.planetCode` would be an accessor with no writer and a second source of truth for what the `users` row already holds. |
| 9 | `CLAUDE.md:1964`, `docs/CODE_STYLE_GUIDE.md:439` | — | Both illustrate the `IS`-not-`=` convention with `RatingDao.getByTypeAndItem`, **which no longer exists**. The real statements are `RatingDao.getByType` (`type IS :type`) and `getAggregate` (`type IS :type AND item IS :item`). The convention is still true; only the method name is stale. Left to the integrator because both files are contended at merge. |
| 10 | smaller | — | `TeamLogDao.getRecentTeamVisits` is `>` where the port is `>=` (one ms at the boundary); `SubmissionDao.getLatestPendingByUserAndParent` — the port adds `type = 'survey'` Kotlin lacks; `getUniquePendingSurveyCandidates` / `getByUserIdWithoutTeam` — the port adds `| teamId = ''`; `NewsDao.getByDocId` is `getSingleOrNull()` on a non-unique column (same shape as #6, no production caller). All defensible, all undocumented as deviations. |

---

## Coverage — honestly, and as an assertion rather than prose

**77 of 312 compared.** That number is not in this description only: `test/data/local/kotlin_query_ledger_test.dart` records the **digest each compared statement had when it was read**, and turns the *Flutter* gate red when one changes upstream, when a compared method is renamed, or when any query is added or removed. That is the harvest signal this phase replaced the commit walk with — delivered by CI instead of by reading 136 commits — and all three directions are mutation-tested against the real Kotlin tree.

The ledger is deliberately **not** a parity claim; it records that a query was *looked at*. Inflating it would be "a test that cannot fail reads as coverage" one level up, and the file says so.

**Largest gaps, in the order I'd take them:** `SubmissionDao` (32 queries, ~half compared — `countPendingOfflineSubmissions`, `getByParentIdsAndTeamId`, `getUnuploadedNonSurveyByParentIds`, `getPendingExamResults` and the update/delete family untouched), then `QuestionDao`/`AnswerDao`'s unfolded members, then `HealthExaminationDao` and `ChatDao` beyond what the audits spot-checked.

### Method notes worth keeping

- **Two independent extractors** (Dart and Python) agree on all 312 keys and digests — that cross-check is what makes the ledger trustworthy.
- **A mutation harness needs its own assertion that its pattern matched.** One run here reported two conjuncts unpinned; the script had matched a string `dart format` reflowed and edited nothing, so the suite passed on *unmutated* code. "Still green" means "nothing was mutated" as readily as "nothing is pinned". Noted at the test.
- **`firstWhere(..., orElse: () => null)` on a statically-`List<dynamic>` value fails the covariance check at the call** when the runtime list's element type is non-nullable — it broke eight existing tests before an explicit loop replaced it.
- Both mandatory `parity-auditor` passes ran at `effort: max`; `git status` was diffed against the expected commit afterwards (Phase 135's lesson) and both were genuinely read-only. **The second pass — against my own green code — found the regression in item 1, which I had reasoned away.** It remains the pass that pays.

https://claude.ai/code/session_01EdoXsnP4pJRr5buFBeqanm


dogi:
## Files actually touched (for collision-spotting, ahead of merge)

Commit `58d9148`:

| File | Change |
|---|---|
| `flutter/lib/data/local/app_database.dart` | `NewsDao`: + `teamIdPattern`, `_viewableByTeam`, `countTopLevelByTeam`, `watchTopLevelByTeam`; **− `teamChatCounts`**. `TeamDao`: `watchMemberCount` rewritten, `@DriftAccessor` gains `Users`. Top level: `likeEscapeMetacharacters` split out of `likeEscapedLiteral`/`likeEscapedUserPattern`. |
| `flutter/lib/repository/notifications_repository.dart` | `getTeamNotifications` counts with `countTopLevelByTeam`, not `teamChatCounts`. |
| `flutter/test/data/local/team_voices_feed_parity_test.dart` | new |
| `flutter/test/data/local/team_member_count_parity_test.dart` | new |
| `flutter/test/repository/notifications_repository_test.dart` | +2 tests, fixture doc comment |
| `flutter/test/repository/teams_repository_test.dart` | seeded a `users` row a fixture was missing |

Nothing outside `lib/data/local/**`, `lib/repository/**` and tests. Gate green: analyze clean, `flutter test` **3230 passed, exit 0**.

---

## @ Lane 2 — one coupled hand-off in your file

`lib/providers/voices_provider.dart`'s **`teamVoicesProvider`** builds the team voices feed from `dao.watchTopLevelMessages()` plus a Dart `viewIn` filter. Against `NewsDao.getTopLevelByTeamFlow` (`NewsDao.kt:30-31`) that is wrong twice: it **adds** a `docType = 'message'` predicate Kotlin's statement does not have, and it **drops** the `viewableBy = 'teams' COLLATE NOCASE AND viewableId = :teamId COLLATE NOCASE` arm entirely.

`NewsDao.watchTopLevelByTeam(teamId)` is the faithful replacement, same argument, already ordered `time DESC`. The change is: call it, delete the `jsonDecode` filter and the `sort`.

**I did not touch your file.** Instead `team_voices_feed_parity_test.dart` carries an exemption group that **passes while the gap is open and fails in both directions once it moves** (Phase 157's shape) — it fails if the provider adopts `watchTopLevelByTeam` (telling the integrator to delete the group) *and* fails if the provider stops using `watchTopLevelMessages` at all, since the exemption would then no longer describe reality. Retiring it is mechanical.

@ integrator: this is the one pairing in my diff that needs landing in the same round, per CLAUDE.md's *"anything a lane reports because of a file boundary is the integrator's to land"*.

---

## Why these two, briefly

Both are the **writer/reader-disagree** class, found by comparing current state, not diffs.

- The team-chat badge was reading `NewsDao.teamChatCounts` — the port of `countTeamChats`, which Kotlin uses for this badge **nowhere**, and whose only wrapper (`VoicesRepository.countTeamChats`) has no caller in `app/src/main` at all. Its predicate is `viewableBy = 'teams'`, and **nothing in either app writes that column locally** — the composer writes `viewIn`. So the badge counted one population while the watermark it is compared against was written from the feed's count of *another*. It failed in both directions: never lights for in-app posts, lights for ever for server-authored ones.
- `TeamDao.watchMemberCount` was a plain `COUNT(id)` against `COUNT(DISTINCT userId) … AND userId IS NOT NULL AND EXISTS (users…)` (`TeamDao.kt:29`).

The existing badge fixture built posts only in the `viewableBy` shape, so it stayed green under both counters — the Phase 156 *"a fixture that cannot distinguish"* class. Every claim is mutation-tested red.

Two `parity-auditor` passes at `effort: max` are running (ground truth + my own finished code). Coverage numbers and the *Reported, not fixed* list land in the PR description.

---

dogi:
## Correction to my comment above, and the final hand-off list

The comment above says the `teamVoicesProvider` gap is *"the one pairing in my diff that needs landing"*. Two things changed after it:

**1. My "never worse" reasoning about half-landing the badge fix was wrong.** The second `parity-auditor` pass caught it: with in-app posts **and** one synced top-level `viewableBy: 'teams'` post, switching only the badge to the union predicate took the dot from *dark* to *unclearable*, because the watermark was still written from the narrower feed. That is now closed inside my own files — `updateTeamNotification` **derives** the watermark from `NewsDao.countTopLevelByTeam` rather than taking the caller's count, which is what Kotlin does (`updateTeamNotification(teamId, news)` receives the list `getFilteredNews` just loaded with that same statement).

So **the badge is correct and self-consistent as it stands**; the provider gap is now a *display* bug, not a badge bug. Landing it is still right, but it is no longer load-bearing for correctness elsewhere.

**2. There are now two coupled hand-offs, not one.** Both are one-line changes in `lib/providers/teams_provider.dart` / `voices_provider.dart`, and both carry a test exemption that **fails in both directions** the moment the provider moves — so retiring them is mechanical and the failure message names the edit:

| Provider | Change | Tripwire |
|---|---|---|
| `teamVoicesProvider` | → `dao.watchTopLevelByTeam(teamId)`; drop the `jsonDecode` filter and the sort | `team_voices_feed_parity_test.dart` |
| `teamResourcesProvider` | union `myLibraryDao.getTeamPrivate(teamId)`, dedup by `id` | `team_private_resources_test.dart` |

The second is the higher-consequence one: a resource added from a team screen is currently in **no view of the app**.

@ Lane 2 — neither touches your `ref.read(p).value` guard work, and I have not edited any file under `lib/providers/`. Item 4 in the PR's *Reported, not fixed* table (`teams_provider.dart:402,412`) is squarely inside your guard's class if you want it.

Full report, coverage (77 of 312, asserted in `kotlin_query_ledger_test.dart` rather than only stated here) and the rest of the *Reported, not fixed* list are in the PR description.

---

PR #17253, opened 2026-09-16
dashboard: smoother profile badge handling (fixes #17252)

### LINKED ISSUE #17252: username cut off
Image

Okuro3499:
fixes #17252
Prevent the profile header from stretching unpredictably by giving the user section a minimum width and capping the badge area width. The badge scroller now shows horizontal fading/scroll indicators so overflowed badges remain accessible.
[Screen_recording_20260916_215848.webm](https://github.com/user-attachments/assets/37ec4573-fa41-4523-871b-73e7b2067761)


dogi:
**Review — rating: 82/100**

The fix works and the mechanism is the right one, but one of the three attribute changes does nothing.

**Verified**

- `ll_user` and `linearLayout` form a two-element horizontal chain (`ll_user.end ↔ linearLayout.start` is bidirectional). Previously `linearLayout` was `wrap_content`, so a long community name plus a badge strip took as much of the row as it wanted and `txtFullName` — `maxLines="1"`, `ellipsize="end"` — absorbed the loss. Capping `linearLayout` with `layout_width="0dp"` + `constraintWidth_default="wrap"` + `constraintWidth_max` gives the chain's spare width back to `ll_user`. That is the correct ConstraintLayout tool for this, and the cap is a named dimen rather than a magic number.
- `layout-land/card_profile_bell.xml` is a different, vertically stacked layout (avatar on top, `txtFullName` at `match_parent` with `maxLines="2"`), so the name never competes with the badges there. Leaving the landscape variant untouched is correct, not an omission.
- `requiresFadingEdge` + `scrollbars="horizontal"` on the badge `HorizontalScrollView` is a reasonable companion change: with the container now capped, badges *will* be clipped, and an affordance saying so is better than silent truncation.
- All five checks green.

**Finding**

`android:minWidth="90dp"` on `ll_user` has no effect. `ll_user` is `layout_width="0dp"` (MATCH_CONSTRAINT), so ConstraintLayout's solver computes its width and measures it with an `EXACTLY` spec; `View.getSuggestedMinimumWidth()` is only consulted for `AT_MOST`/`UNSPECIFIED`. The attribute ConstraintLayout reads for a floor on a match-constraint child is `app:layout_constraintWidth_min` — which is exactly the family the sibling already uses correctly via `layout_constraintWidth_max`. Either change it to:

```xml
app:layout_constraintWidth_min="90dp"
```

or drop the line. Dropping it is defensible: the 140dp cap on `linearLayout` already guarantees `ll_user` everything left over, so the floor is redundant in practice. What isn't good is shipping an attribute that reads as a guarantee and isn't one.

**Note, not a blocker**

`dashboard_profile_badges_max_width` lands only in `values/dimens.xml`. This project carries `values-sw600dp/` and `values-sw720dp/` variants, and unspecified dimens fall through to the base, so a 10" tablet gets the same 140dp badge column it gives a 320dp phone. An `sw600dp` override would be worth adding while the dimen is new.

---

dogi (approved):

dogi (approved):

PR #17254, opened 2026-09-17
teams: refactored resources to match library (fixes #16927)

### LINKED ISSUE #16927: update team resources UI
update team resources UI to match the library UI
Image

J-S-webskas:
https://github.com/user-attachments/assets/67e3bb5d-f6a3-4a1f-a051-111b0ee2816a


dogi:
**Review — rating: 55/100**

The visual goal is right and the layout work is genuinely good — it reuses the library card's existing `bg_cover_top_rounded`, `bg_badge_circle`, `ic_type_*`, `type_*` colors and `ShapeAppearance.MyPlanet.CoverTopRounded`, all of which I confirmed already exist on master, so nothing new was invented to match the library look. CI is green on all five checks.

The problem is *how* it matches. I diffed the new helpers against `ui/resources/ResourcesAdapter.kt`, and every one of them already exists there: `setCoverColor`, `bindCover`, `showTypeIconOnly`, `showImagePreview`, `showVideoPreview`, `showPdfPreview`, `showHtmlPreview`, `buildMetaLine`, `typeColorRes`, `typeIconRes`, `typeLabelRes`, plus the `externalFilesDir` lazy and the `adapterScope` attach/detach lifecycle. This PR copies roughly 120 lines of it into `TeamResourcesAdapter` rather than extracting the shared piece — and the copy is a *degraded* one:

| `ResourcesAdapter` (master) | this PR's copy |
|---|---|
| `CoroutineScope(SupervisorJob() + dispatcherProvider.main)` | `Dispatchers.Main` |
| `withContext(dispatcherProvider.io)` | `withContext(Dispatchers.IO)` (×4) |
| `PdfThumbnailLoader.firstPageBitmap(file, dispatcherProvider, …)` | `firstPageBitmap(file, org.ole.planet.myplanet.utils.DefaultDispatcherProvider(), …)` |
| `htmlCoverCache`, `fileLengthCache` | dropped |

That hits two rules `CODE_STYLE_GUIDE.md` lists explicitly under *Things to Avoid*: "Don't hardcode `Dispatchers.IO`/`Dispatchers.Main`" and "Don't reference types by their fully-qualified name inline." The guide names the consequence of the first one directly — without an injected `DispatcherProvider`, tests can't substitute `TestDispatcherProvider`, so this adapter is untestable in the way the library one is. The PDF path also allocates a fresh `DefaultDispatcherProvider` on every bind, and dropping `htmlCoverCache` means HTML cover lookups hit disk on each bind instead of once.

`ResourcesAdapter` already takes `dispatcherProvider: DispatcherProvider` as a constructor parameter; `TeamResourcesAdapter` should do the same, and the cover-binding block belongs in one shared helper used by both.

Two more real issues:

- **Recycling artifact.** `onBindViewHolder` never resets `ivCoverPreview` (visibility or drawable) before launching the async `bindCover`. Since `bindCover` suspends on `withContext(…) { file.exists() }` before deciding, a recycled row keeps rendering the *previous* item's cover until that resolves — and if the new item resolves to `showTypeIconOnly`, the stale image is visible for the whole window.
- **Accessibility.** `iv_remove` is the only interactive control in the card, and it carries both `contentDescription="@null"` and `importantForAccessibility="no"`, which hides it from TalkBack entirely. The click listener is also on the 16dp `iv_remove` while the visibility toggle is on the 28dp `fl_remove_container` — so the touch target is 16dp, well under the 48dp minimum, and taps on the badge's padding do nothing.

**Nit:** `areContentsTheSame` still compares `description`, which is no longer rendered anywhere after `tv_description` was removed. `getItem(adapterPosition) ?: return` is an elvis on the non-null `MyLibrary` and will warn. The inner `LinearLayout`/`tv_meta` indentation in the XML doesn't match its nesting.

**Verdict:** needs another round — take `DispatcherProvider` as a constructor parameter and extract the shared cover-binding code instead of copying it, then fix the cover reset on rebind and let the remove button be reachable by TalkBack.

---

dogi:
@J-S-webskas do not send a PR with failing tests to review

PR #17255, opened 2026-09-17
dashboard: smoother account header logo landscape handling (fixes #16573)

### LINKED ISSUE #16573: navigation: logo cropped in landscape
Image

J-S-webskas:
https://github.com/user-attachments/assets/3fccc9f1-f59b-441e-bed8-9c62f46b5de9

## Summary by CodeRabbit

- **UI Improvements**
  - Improved account header sizing and spacing for portrait and landscape orientations.
  - Updated app logos to scale and center more consistently across toolbar and account header layouts.
  - Improved logo visibility by applying appropriate color treatment in light and dark modes.


dogi:
**Review — rating: 84/100**

Focused fix for the landscape logo, and it quietly repairs a real unit bug on the way. The old code computed `paddingVerticalDp = (paddingVerticalPx / density).toInt()` and then handed those values straight to `headerBackground.setPadding(...)`, which takes **pixels** — so the padding came out at 1/density of what was intended, and the expression mixed them with a raw px `statusBarHeight` in the same argument. The new code multiplies back by density and is unit-consistent throughout, including `withHeightDp(headerHeightDp + (statusBarHeight / density).toInt())`. `FIT_CENTER` is the right answer for the reported stretching.

I also checked the night-mode rewrite rather than assuming it was a wash. Dropping `AppCompatDelegate.getDefaultNightMode()` for `resources.configuration.uiMode` is fine and slightly better: on an `AppCompatActivity` the configuration already reflects the applied delegate mode, and the old condition only tested `MODE_NIGHT_NO` and `FOLLOW_SYSTEM`, so under `MODE_NIGHT_AUTO_BATTERY` in light appearance it applied no white filter at all. The new form handles that, and the added `clearColorFilter()` on the dark branch is correct given `accountHeader` is a `get()` that rebuilds the header each access. Renaming the old `if (screenWidth > screenHeight) { //sizing for tablets }` to `isLandscape` is right too — that test was never about tablets.

**Nit:** three cosmetic things, none blocking:
- `import androidx.appcompat.app.AppCompatDelegate` becomes unused — lines 869–870 were its only usages in the file and this PR deletes both.
- `.withHeaderBackgroundScaleType(ImageView.ScaleType.FIT_CENTER)` and `headerBackground.scaleType = ImageView.ScaleType.FIT_CENTER` set the same property twice; one is enough.
- `android:adjustViewBounds="true"` is a no-op on an `ImageView` with both dimensions fixed at `@dimen/_40dp` — it only takes effect when a dimension is `wrap_content`.

The 120/160/16/8 values are inline magic numbers where the codebase generally reaches for `@dimen`; a `values-land/dimens.xml` would express the landscape/portrait split more idiomatically. Worth considering, not worth another round over.

**Verdict:** mergeable.

---

Okuro3499:
fixes #16573

Okuro3499 (approved):

dogi (approved):

PR #17262, opened 2026-09-17
all: smoother api selective kotlinx converter providing (fixes #17261)

### LINKED ISSUE #17261: Migrate networking layer off Gson to kotlinx.serialization
## Why                                                                                                                                                                             
                                                                                                                                                                                     
  We're planning to target iOS and desktop with Kotlin Multiplatform. An audit of                                                                                                    
  the current stack found Gson is one of the two hardest blockers (the other is                                                                                                      
  Retrofit itself, which has no non-JVM target and is out of scope here): Gson is                                                                                                    
  reflection-based and JVM-only, and it isn't confined to the network boundary —                                                                                                     
  116 files import `com.google.gson` directly, not just the Retrofit converter.                                                                                                      
  kotlinx.serialization is the multiplatform-safe replacement and was already a                                                                                                      
  declared dependency, but essentially unused (1 file, and only for a built-in                                                                                                       
  `List<String>` type — the compiler plugin was never even applied).                                                                                                                 
                                                                                                                                                                                     
  ## Scope                                                                                                                                                                           
                                                                                                                                                                                     
  `ApiInterface` has 9 endpoints. 3 already return typed models:                                                                                                                     
  - `checkVersion` → `MyPlanet`                                                                                                                                                      
  - `chatGpt` → `ChatResponse`                                                                                                                                                       
  - `getDocuments` → `DocumentResponse`                                                                                          No changes this session                             
                                                                                                                                                                                     
  The other 6 (`getJsonObject`, `postDoc`, `postDocArray`, `uploadResource`,                                                                                                         
  `putDoc`, `getConfiguration`) return raw `JsonObject`/`JsonArray`, consumed by                                                                                                     
  ~25 files across the sync/upload core (`SyncManager`, `TransactionSyncManager`,                                                                                                    
  `UploadManager`, `UploadRepositoryImpl`, etc.) using Gson's `JsonObject` API                                                                                                       
  directly. Migrating those in the same pass would be a large, high-risk diff                                                                                                        
  through the app's most critical offline-sync paths.                                                                                                                                
                                                                                                                                                                                     
  This issue covers only the 3 typed endpoints — proving the converter pattern                                                                                                       
  works before touching the raw-JSON endpoints in a follow-up.

Okuro3499:
fixes #17261
Why this approach: The obvious move — add Json.asConverterFactory(...) alongside                                                                                                   
  GsonConverterFactory on the same Retrofit instance — doesn't work. Per the library's own docs,                                                                                     
  kotlinx's converter factory assumes it can handle every type; it doesn't return null for a type it                                                                                 
  can't serialize, it throws. Placed before Gson, it would break all 6 raw-JsonObject endpoints at                                                                                   
  runtime; placed after Gson, it would never run at all, since GsonConverterFactory also returns                                                                                     
  non-null unconditionally for anything (reflection has no failure mode).                                                        No changes this session                             
                                                                                                                                                                                     
  How it was solved: Added SelectiveKotlinxConverterFactory — a thin Converter.Factory that only                                                                                     
  delegates to kotlinx.serialization for an explicit allow-list of types (MyPlanet, ChatResponse,                                                                                    
  DocumentResponse), returning null for everything else so Retrofit falls through to Gson next in                                                                                    
  the chain. This keeps the single ApiInterface and single Hilt-provided Retrofit instance unchanged                                                                                 
  — no interface splitting, no call-site changes for the 25 files touching raw                                                                                                       
  JsonObject/JsonArray.                                                                                                                                                              
                                                                                                                                                                                     
  Verification: Added SelectiveKotlinxConverterFactoryTest, which builds the real converter chain                                                                                    
  (not mocked) and proves both paths on one Retrofit instance: kotlinx deserializes the 3                                                                                            
  allow-listed types correctly, and Gson still handles JsonObject untouched. Ran the full unit suite                                                                                 
  (2163/2165 passing) and independently confirmed the 2 failures (FreeSpaceWorkerTest) are                                                                                           
  pre-existing on unmodified master, unrelated to this change.

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->

## Summary by CodeRabbit

- **Improvements**
  - Improved network response handling with Kotlin serialization for supported response types.
  - Preserved compatibility with existing JSON responses through automatic fallback handling.
  - Added tolerant parsing for unknown fields, lenient input, and type coercion.
  - Updated key response models to support the new serialization format.
- **Tests**
  - Added coverage for typed responses, nested data, and fallback handling for unsupported JSON payloads.

<!-- end of auto-generated comment: release notes by coderabbit.ai -->


dogi:
**Review — rating: 86/100**

Good piece of migration infrastructure. The core insight in the KDoc is correct and is the reason this needed writing at all: kotlinx's own `asConverterFactory` claims every requested type, so appending it alongside `GsonConverterFactory` would either shadow Gson entirely or blow up on the raw `JsonObject`/`JsonArray` CouchDB endpoints. Returning `null` for anything outside `handledTypes` is exactly the right escape hatch, since Retrofit then walks on to the next factory.

I checked the things that could make this leak beyond its intended blast radius:

- **One Retrofit instance.** `Retrofit.Builder()` appears exactly once in the whole app, so there's no second client still on plain Gson for the same endpoints.
- **`MyPlanet` is safe to hand to kotlinx.** Its properties are `planetVersion`/`apkpath`/`appname`/`localapkpath` (`String?`) and `minapkcode`/`latestapkcode` (`Int`), all with defaults — no `Date`, `Calendar` or `JsonObject` backing fields that would need a custom serializer, and the defaults are why the test's partial JSON deserializes rather than throwing `MissingFieldException`.
- **`DocumentResponse` doesn't regress.** Its `totalRows` and `offset` are `private` with no accessors and the JSON keys are `total_rows`/`offset`; `provideGson()` returns a bare `Gson()` with no `FieldNamingPolicy`, so those two were never populated under Gson either. Only `rows` carries data, and `Rows` correctly picked up `@Serializable` alongside it.

Dependency hygiene is right too — the version lands in `gradle/libs.versions.toml` with a plugin alias rather than an inline version, which is the rule the style guide calls out. And `import kotlinx.serialization.Serializable as KSerializable` is the documented remedy for the `java.io.Serializable` collision, used properly instead of an inline fully-qualified name.

The tests cover the actual contract rather than just the happy path: a handled type, a nested handled type, Gson fall-through for a raw `JsonObject`, and the factory returning `null` directly.

**Two limitations worth writing down in the KDoc rather than changing:**

- The factory only overrides `responseBodyConverter`, so request bodies for these types still serialize through Gson. That's harmless today — all three are response models, and none of them carry a `@SerialName` that differs from its property name, so both libraries would emit identical JSON — but a future `@Body MyPlanet` would silently split across two libraries.
- `type !in handledTypes` matches exact `Class` identity, so `List<ChatResponse>` or any parameterized wrapper falls through to Gson. That's the safe direction to fail, but it's non-obvious.

**Nit:** every `@SerialName` added to `ChatResponse`/`CouchDBResponse` is identical to its property name, so they're all no-ops (as were the pre-existing `@SerializedName`s). Harmless, and arguably useful as migration documentation, but they aren't doing anything. `DocumentResponse.totalRows`/`offset` are unreadable dead fields that could simply go.

**Verdict:** mergeable.

---

PR #17276, opened 2026-09-17
Fix SyncTimeLogger data races and remove detailedLogs

### LINKED ISSUE #17264: SyncTimeLogger races on its per-key lists and retains detail strings nothing ever reads
## Verified premise

`utils/SyncTimeLogger.kt` on master, two findings in one file.

**(a) A write-only accumulator.** `detailedLogs` appears exactly three times in the file — declared at L38, cleared at L73, appended at L195:

```kotlin
private val detailedLogs = ConcurrentHashMap<String, MutableList<String>>()   // L38
...
detailedLogs.clear()                                                          // L73
...
fun logDetail(context: String, message: String) {                             // L192
    if (!isLogging) return
    detailedLogs.getOrPut(context) { mutableListOf() }.add(message)           // L195
```

`generateSummary()` never reads it. Every `logDetail` call site retains a string that nothing consumes.

**(b) A data race at all three append sites.** L167, L184 and L195 are all of the form:

```kotlin
apiCallTimes.getOrPut(processName) { mutableListOf() }.add(log)   // L167
dbOperationTimes.getOrPut(model) { mutableListOf() }.add(log)     // L184
```

`getOrPut` on a `ConcurrentHashMap` is not atomic (read, then conditional put — two coroutines can both miss and one's list is dropped), and the value it stores is a plain `ArrayList`, appended from the parallel sync coroutines that drive `logApiCall` / `logDbOperation`.

## Do

- Delete `detailedLogs`: the field (L38), its `clear()` (L73) and the map append (L195). **Keep `logDetail` itself, its signature and its `if (isVerbose) Log.d(...)` branch** — it has 12 call sites.
- Replace the remaining two `getOrPut { mutableListOf() }` sites with `computeIfAbsent(key) { Collections.synchronizedList(mutableListOf()) }`.
- Wrap every iteration of those lists in `generateSummary()` in `synchronized(list) { ... }` — a synchronized list's iterator is not itself synchronized.

## Constraints

- **`extractProcessName`'s `endpoint.split("/")` (L332) must stay as-is. `substringAfterLast('/')` is NOT equivalent: the current code walks the segments backwards skipping empty ones and ones starting with `?`, so `"api/v1/courses/"` and `"api//v1//courses//"` both yield `Courses` where `substringAfterLast` yields `""` → `Unknown`.** Leave that function untouched.
- Keep `logDetail`'s parameter list and its early `if (!isLogging) return` — callers are not being edited.
- Do not change what `generateSummary()` prints; this is a thread-safety and dead-field change, not an output change.

## Tests

- `app/src/test/java/org/ole/planet/myplanet/utils/SyncTimeLoggerTest.kt` must pass **unedited** — `testExtractProcessName` pins the constraint above, and `testGenerateSummaryMultipleKeysAndLogs` pins the summary output.

## Out of scope

- `extractProcessName`, `shortenEndpoint`, `formatElapsed`, `formatTime`.
- The crash-log upload path in the same file.
- Changing any of the 12 `logDetail` call sites.

dogi:
Fix data race issues on per-key lists and remove write-only detailedLogs in SyncTimeLogger.kt:
1. Deleted unused `detailedLogs` map, its `clear()` call, and append logic while retaining `logDetail()` function signature and `Log.d` behavior.
2. Replaced `getOrPut` on `ConcurrentHashMap` with `computeIfAbsent(key) { Collections.synchronizedList(mutableListOf()) }` in `logApiCall` and `logDbOperation`.
3. Synchronized all iterations over synchronized lists in `generateSummary()` to ensure thread safety.
4. Preserved `extractProcessName` and existing output formatting as constrained.

Fixes #17264

---
*PR created automatically by Jules for task [1131932158457880882](https://jules.google.com/task/1131932158457880882) started by @dogi*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the reliability of synchronization performance summaries when multiple operations are processed concurrently.
  * Prevented potential inconsistencies in API and database operation timing statistics.
  * Improved stability of diagnostic logging during overlapping synchronization activity.


dogi:
**Review — rating: 85/100**

Both halves are correct, and the concurrency half is more careful than most fixes of this shape.

`getOrPut` on a `ConcurrentHashMap` is a Kotlin `MutableMap` extension implemented as get-then-put, so it is not atomic — two threads racing on a new process name each build a list and one set of writes is silently dropped. `computeIfAbsent` is atomic on CHM and is the right replacement. Wrapping the value in `Collections.synchronizedList` then fixes the element-level race on the `ArrayList` itself.

The part I specifically checked is the part that usually goes wrong: `Collections.synchronizedList` synchronizes individual operations but explicitly does **not** cover iteration, so every `sumOf` / `count` / traversal has to be manually guarded — and the single-argument factory sets `mutex = this`, meaning the correct monitor is the returned wrapper. `synchronized(logs)` locks exactly that object, so the guards line up with the internal monitor. Capturing `logCount` inside the same block as `totalTime` and `totalItems` is also right: it makes the printed call count consistent with the summed durations rather than a second, possibly-newer read.

On `detailedLogs`, I confirmed it against the file rather than taking the title's word for it — it appeared at exactly three sites (the field, the `clear()` in reset, and the one `add` in `logDetail`), and nothing ever read it. It was accumulating every detail message for the lifetime of a sync with no consumer, so removing it is a straight memory win, and `logDetail` still does its real job via the `isVerbose` branch where `context` is genuinely used.

Two things I'd like to see, neither blocking:

- **No test, though the harness exists.** `SyncTimeLoggerTest.kt` is already there with six tests and already calls `logDetail`. The `getOrPut` bug is one of the rare races that tests deterministically enough to be worth it — N threads × M `logApiCall` calls, then assert the total is `N*M`; that fails on the old code and passes on the new.
- **The primitive state is still unguarded.** For a PR titled "fix data races", `isLogging`, `startTime` and `endTime` remain plain non-`@Volatile` `var`s, read from the logging threads and written from whoever starts/stops the sync. There is no visibility guarantee that a worker sees `isLogging = false`. Cheap to fix with `@Volatile`.

**Nit:** `sortedByDescending` invokes its selector on every comparison, so each entry's list is re-summed and re-locked O(n log n) times. Taking one snapshot per entry up front — `val snapshot = synchronized(logs) { logs.toList() }` — and aggregating off the immutable copy would drop the repeated locking and also let the four `val totalTime: Long` / assign-inside-`synchronized` blocks collapse into ordinary initializers.

**Verdict:** mergeable.

---

Okuro3499 (dismissed):
```
SyncTimeLogger.kt:39-46 (isLogging, isVerbose, startTime, endTime): mark these @Volatile (or otherwise document/justify why visibility across the workers that call              
    logApiCall/logDbOperation/endProcess doesn't need it). As written, a worker coroutine can observe a stale isLogging=true after stopLogging() has already flipped it and started  
    generateSummary(), letting it append to a list mid-summary or race the apiCallTimes.clear() in the next startLogging(). 
```

dogi:
**Re-review — Rating: 85/100** (was 78)

**Verdict:** mergeable — the races are real ones and the fixes are the correct mechanisms.

**`getOrPut` → `computeIfAbsent` is the substantive fix.** `ConcurrentHashMap.getOrPut` is a Kotlin extension that does `get` then `put` — not atomic, so two threads logging the first call for the same process could each build a list, and whichever lost the `put` race silently dropped its entry. `computeIfAbsent` is atomic on `ConcurrentHashMap`. This is the kind of bug that never shows up in testing and quietly under-reports sync timings.

**`Collections.synchronizedList` plus `synchronized(logs)` at every read is the right pairing.** The `synchronizedList` contract requires manual synchronization on the list for iteration and compound operations — wrapping the list without also guarding the `sumOf`/`count` traversals would have left `ConcurrentModificationException` on the table while looking fixed. Guarding every traversal, including the ones inside `sortedByDescending`, is correct.

**`@Volatile` on `startTime`, `endTime`, `isLogging` and `isVerbose`** closes the visibility gap for the flags read on one thread and written on another.

**Removing `detailedLogs` is a genuine win beyond tidiness.** It was write-only — accumulated on every `logDetail` call and never read by `generateSummary` — so it was an unbounded `ConcurrentHashMap<String, MutableList<String>>` growing for the duration of every sync and discarded at the end. Deleting it removes both the allocation pressure and a fourth structure that would otherwise have needed the same synchronization treatment.

**One inefficiency the fix introduces:**

```kotlin
apiCallTimes.entries.sortedByDescending { entry -> synchronized(entry.value) { entry.value.sumOf { log -> log.duration } } }
```

`sortedByDescending(selector)` delegates to `compareByDescending`, which invokes the selector **on every comparison** — so this locks the list and re-sums it O(n log n) times rather than once per entry. Previously the selector was a bare `sumOf` and the waste was invisible; now each invocation also takes a monitor. This runs once at end of sync over a handful of endpoints, so it is not a practical problem — but computing the totals into a `List<Pair<String, Long>>` first and sorting that would be both cheaper and easier to read, and it would let each list be locked exactly once.

No tests added. Reasonable — these races are not practically reproducible in a unit test — but the `detailedLogs` removal is a behavioural change to a public-ish method (`logDetail` now only logs when verbose), so it is worth confirming nothing downstream expected the accumulation. CI green 5/5 suggests not.

---

Okuro3499 (changes requested):
```
Not resolved. Only SyncTimeLogger.kt changed (+37/-22, single file); SyncTimeLoggerTest.kt is untouched, so there's still no test         
    exercising concurrent logApiCall/logDbOperation calls to guard the fix.
```

dogi:

**Verdict:** `ready`. The concurrency work is correct and the new stress test is a genuine positive control for the main bug. One thing the fix doesn't cover, worth knowing about rather than blocking on.

**The `getOrPut` → `computeIfAbsent` swap is the real fix.** Kotlin's `getOrPut` on a `ConcurrentHashMap` is get-then-put, not atomic — two threads reaching a new process name together each build a list, each adds its entry, and one of those lists is discarded along with its log. `computeIfAbsent` makes creation atomic, and `Collections.synchronizedList` makes the subsequent `add` safe. Both halves were needed.

**The manual `synchronized(logs)` blocks are also necessary, not belt-and-braces.** `Collections.synchronizedList` guards individual calls but explicitly does *not* cover iteration — and `sumOf`, `count` and `size`-plus-read sequences are iteration. Wrapping them, and capturing `logCount` inside the block rather than re-reading `logs.size` at format time, is exactly right. Locking on `logs` is correct too, since the wrapper is its own monitor.

**And the new test would actually catch the original bug.** Ten threads × 100 iterations asserting `Total API calls: 1000` fails if any list is lost to the `getOrPut` race — that's a real regression test, not a smoke test. Interleaving `generateSummary()` calls into the same loop exercises the read path against concurrent writes, which is the other half.

## The sort key still moves under the sort

```kotlin
apiCallTimes.entries
    .sortedByDescending { entry -> synchronized(entry.value) { entry.value.sumOf { log -> log.duration } } }
```

Each comparison is now individually safe, but `sortedByDescending` re-invokes the selector on *every* comparison, and between two comparisons another thread can append to that list and change its total. So the sort operates on a key that isn't stable for the duration of the sort — the same entry can compare as both greater and less than a peer.

Two consequences, both mild:

- The ordering in the report can come out slightly wrong. For a diagnostics dump, that's cosmetic.
- If the entry count ever exceeds 32, Java's sort takes the TimSort merge path, which checks comparator consistency and can throw `IllegalArgumentException: Comparison method violates its general contract!`. Below 32 it uses binary insertion sort and never checks, so with the handful of process names a sync actually produces this is very unlikely to fire. Worth stating precisely rather than overstating it.

Snapshotting fixes both and removes lock churn at the same time — the selector currently re-locks and re-sums the whole list on every comparison:

```kotlin
apiCallTimes.entries
    .map { (endpoint, logs) -> synchronized(logs) { Triple(endpoint, logs.toList(), logs.sumOf { it.duration }) } }
    .sortedByDescending { it.third }
```

One pass, one lock per entry, stable keys, and the `forEach` body then reads an immutable copy instead of re-entering the monitor.

## Smaller

- `mockkStatic(Log::class)` in the new test has no matching `unmockkStatic` in an `@After`, so the static mock outlives the test in that JVM. Same note I left on [#17472](https://github.com/open-learning-exchange/myplanet/pull/17472) this round.
- `executor.awaitTermination(10, TimeUnit.SECONDS)` makes this a wall-clock assertion: 1 000 log calls plus 100 full `generateSummary()` passes on a loaded CI runner is probably well inside 10s, but it's the kind of bound that produces an occasional red build rather than a real failure. Given [#17471](https://github.com/open-learning-exchange/myplanet/pull/17471) in this same batch exists to de-flake tests, a more generous timeout costs nothing.

Removing the write-only `detailedLogs` while keeping `logDetail`'s signature and verbose output is the right call — no caller changes, no behaviour change, one less map.

CI green 5/5.

---

Okuro3499 (approved):

PR #17277, opened 2026-09-17
all: smoother notifications repository team task mapping (fixes #17263)

### LINKED ISSUE #17263: every team shows the task badge whenever any one team has a task due
## Verified premise

`repository/NotificationsRepositoryImpl.kt`, `getTeamNotifications` (L378-406 on master):

```kotlin
val tasks = teamTaskDao.getTasksForUserBetween(userId, current, tomorrow.timeInMillis)
val hasTask = tasks.isNotEmpty()

for (teamId in teamIds) {
    val notification = notificationsById[teamId]
    val chatCount = chatCountsById[teamId] ?: 0L
    notificationMap[teamId] = TeamNotificationInfo(hasTask, hasChat)
}
```

`hasTask` is computed **once, outside the loop**, from the user's tasks across all teams, and the same boolean is then handed to every `TeamNotificationInfo`. `hasChat` is correctly per-team on the line above it, which makes the omission plain. A user with one task due in team A sees the task badge on every team they belong to.

`TeamTask.teamId` exists and is indexed — `model/TeamTask.kt:22`, `@Index("teamId")` — so the per-team grouping needs no DAO or schema change.

## Do

Group the already-fetched tasks by team and test membership inside the loop:

```kotlin
val taskTeamIds = tasks.mapNotNull { it.teamId }.toSet()

for (teamId in teamIds) {
    ...
    notificationMap[teamId] = TeamNotificationInfo(teamId in taskTeamIds, hasChat)
}
```

## Constraints

- **Do not change the `teamTaskDao.getTasksForUserBetween(userId, current, tomorrow.timeInMillis)` call — the window is deliberate (now → this time tomorrow) and the fix is purely about how the already-fetched rows are attributed.** Adding a `teamId` parameter or a second DAO query turns a one-line grouping into a query-per-team fan-out inside a loop that already runs once per team.
- Build the set once, before the loop; do not call `tasks.any { ... }` per iteration.
- No `AppDatabase` version bump: nothing about the schema changes.
- `getTasksForUserBetween` may return tasks whose `teamId` is null — `mapNotNull` drops those, which is correct: a task with no team belongs to no team's badge.

## Tests

- `app/src/test/java/org/ole/planet/myplanet/repository/NotificationsRepositoryImplTest.kt` and `app/src/test/java/org/ole/planet/myplanet/repository/TeamChatBadgeIntegrationTest.kt` must pass **unedited** — in particular `badge tracks the top-level feed, not replies` and `count used for chatCount equals the displayed top-level feed size`, which pin the `hasChat` half of the same return value.
- Add one test to `NotificationsRepositoryImplTest`: two team ids, one task carrying the first team's id, assert `hasTask` is true for that team and false for the other.

## Out of scope

- The `hasChat` computation, `countTopLevelByTeams`, and the `TeamNotification` read path.
- `Calendar.getInstance()` on the `tomorrow` line (a separate wall-clock-through-TimeProvider concern).
- Any other method in `NotificationsRepositoryImpl`.

dogi:
Fixed issue where any due task caused every team to show a task badge by grouping tasks by team ID prior to assigning TeamNotificationInfo.

Fixes #17263

---
*PR created automatically by Jules for task [10581017041443946021](https://jules.google.com/task/10581017041443946021) started by @dogi*


dogi:
**Review — rating: 90/100**

Real bug, minimal fix, with a test that would actually have caught it. `getTasksForUserBetween(userId, …)` is scoped to the user, not to a team, so `val hasTask = tasks.isNotEmpty()` computed one global boolean and then wrote it into every entry of the loop — one task due in any team lit the task badge on *all* of that user's teams. Grouping by `teamId` and testing membership per team is the correct shape, and hoisting `taskTeamIds` out of the loop keeps it a single pass.

The thing worth verifying here is the ID space, since matching against the wrong identifier would have turned a badge that was always on into one that was never on. It lines up: the caller in `BaseDashboardFragment` builds `teamIds` as `teams.mapNotNull { it._id }`, `TeamTask.teamId` is populated from `JsonUtils.getString("teams", link)` — the team's CouchDB `_id` — and `TeamTaskDao.getTasksByTeamId` already queries `WHERE teamId = :teamId` against that same space. So `teamId in taskTeamIds` compares like with like.

`mapNotNull` also quietly does the right thing with a null-`teamId` task: it can't be attributed to any team, so it now badges none, where before it badged every one.

The added test is well-targeted — two teams, a task belonging only to the first — so it distinguishes the fixed behavior from the old one rather than just exercising the path.

**Nit:** two small things in the test. `org.ole.planet.myplanet.model.TeamTask()` is written as an inline fully-qualified name, which `CODE_STYLE_GUIDE.md` lists under *Things to Avoid* — it should be an import like the rest of the file's models. And `assertFalse(result[team2]?.hasTask == true)` also passes if `result[team2]` is null; `assertEquals(false, result[team2]?.hasTask)` asserts what you mean. The `assertEquals(2, result.size)` above it already pins the null case, so this is style, not a hole.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #17278, opened 2026-09-17
all: less activities repository resource open overloads is more (fixes #17267)

### LINKED ISSUE #17267: delete the four uncalled resource-open overloads from the ActivitiesRepository contract
## Verified premise

`repository/ActivitiesRepository.kt:35-38` declares four methods:

```kotlin
suspend fun getResourceOpenCount(userName: String): Long
suspend fun getResourceOpenCount(userName: String, type: String): Long
suspend fun getMostOpenedResource(userName: String): Pair<String, Int>?
suspend fun getMostOpenedResource(userName: String, type: String): Pair<String, Int>?
```

A tree-wide search for those two names across `app/src/main` resolves them to exactly three places: the interface, their `override`s in `ActivitiesRepositoryImpl.kt:174-193`, and two internal uses inside `getProfileActivityStats` in that same impl:

```kotlin
val mostOpenedDeferred = async { getMostOpenedResource(userName) }   // Impl:196
val countDeferred      = async { getResourceOpenCount(userName) }    // Impl:198
```

No caller outside the impl. The declarations also drag `UserSessionManager.KEY_RESOURCE_OPEN` — the default `type` the no-arg overloads pass — into the public contract.

## Do

- Delete the four declarations from `ActivitiesRepository.kt`.
- In `ActivitiesRepositoryImpl.kt`, drop the `override` modifier from the four functions so they remain ordinary members and `getProfileActivityStats` keeps compiling unchanged.

## Constraints

- **Do not delete the functions from the impl. `getProfileActivityStats` calls two of them, and the one-argument forms delegate to the two-argument forms — removing bodies instead of just the `override` keyword breaks that chain.**
- Keep `getProfileActivityStats` on the interface; it is the caller the UI actually uses.
- Do not change visibility beyond removing `override` (leave them public members of the class).
- Do not touch `ResourceActivityDao.getMostOpenedResource(userName, type)` — the DAO stays as-is.

## Tests

- `app/src/test/java/org/ole/planet/myplanet/repository/ActivitiesRepositoryImplTest.kt` must pass **unedited**: it declares `private lateinit var repository: ActivitiesRepositoryImpl` (L72), i.e. the concrete class, so `getResourceOpenCount returns correct count`, `getResourceOpenCount without type defaults to visit` and `getMostOpenedResource returns null when no activities` still resolve after the interface declarations go.

## Out of scope

- Any other declaration in `ActivitiesRepository`.
- The body of `getProfileActivityStats` and its `async` structure.
- `UserSessionManager.KEY_RESOURCE_OPEN` itself.

dogi:
Removes four uncalled resource-open overload declarations (`getResourceOpenCount` and `getMostOpenedResource`) from the `ActivitiesRepository` interface contract, and drops the `override` modifier from their implementations in `ActivitiesRepositoryImpl` so internal calls remain working without polluting the interface contract.

Fixes #17267

---
*PR created automatically by Jules for task [8937367505413377890](https://jules.google.com/task/8937367505413377890) started by @dogi*


dogi:
**Review — rating: 88/100**

Correct interface narrowing, and the "uncalled" claim in the title holds up. I grepped every reference to the four overloads across `app/src`: outside the declaring class the only production hits are `ActivitiesRepositoryImpl.kt:196` and `:198`, both inside `getProfileActivityStats`, which `async`s them and is itself still on the interface. Nothing injects `ActivitiesRepository` and calls them — which makes sense, since Hilt binds the interface, so removing them from the contract makes them genuinely unreachable from consumers rather than just unused today.

Keeping the bodies on the impl rather than deleting them is the right call given `getProfileActivityStats` needs them, and dropping `override` while leaving them `suspend` keeps `ActivitiesRepositoryImplTest`'s four direct tests compiling against the concrete type — which is why CI stays green without touching the test file.

**Nit:** the two methods are now public on a concrete class with nothing declaring them, which is a slightly undefined visibility state — they're reachable only from inside the class and from tests. `internal` would say that out loud; `private` would too, at the cost of routing those four tests through `getProfileActivityStats`. Not worth another round.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #17279, opened 2026-09-17
courses: less ratings repository null guard is more (fixes #17270)

### LINKED ISSUE #17270: delete the getRatingsById null-guard wrapper from RatingsRepository
## Verified premise

`repository/RatingsRepository.kt:7` and `:8` declare two entry points for one query:

```kotlin
suspend fun getRatingsById(type: String, resourceId: String?, userId: String?): RatingSummary?
suspend fun getRatingSummary(type: String, itemId: String, userId: String?): RatingSummary
```

and the implementation of the first is nothing but a null guard over the second — `RatingsRepositoryImpl.kt:20-23`:

```kotlin
override suspend fun getRatingsById(type: String, resourceId: String?, userId: String?): RatingSummary? {
    if (resourceId == null) return null
    return getRatingSummary(type, resourceId, userId)
}
```

It has exactly one caller in `app/src/main` — `ui/resources/ResourceDetailFragment.kt:268`:

```kotlin
val rating = ratingsRepository.getRatingsById("resource", library.resourceId, userModel?.id)
```

The nullable return also pushes the null back onto the caller, which then has to handle it again.

## Do

- Delete `getRatingsById` from both the interface and the impl.
- At `ResourceDetailFragment.kt:268`, guard the id locally and call `getRatingSummary` directly:
  ```kotlin
  val resourceId = library.resourceId ?: return@launch
  val rating = ratingsRepository.getRatingSummary("resource", resourceId, userModel?.id)
  ```

## Constraints

- **`getRatingSummary` returns a non-null `RatingSummary`, not `RatingSummary?` — the call site's existing null handling for the old return type must be removed, not left in place where it is now dead and will read as a live branch.** Check what the fragment does with `rating` after L268 and simplify it to match the non-null type.
- Do not change `getRatingSummary`'s signature or behaviour; four other call sites already use it.
- The `return@launch` form above assumes the surrounding builder — use whatever early-exit matches the actual enclosing scope at that line.

## Tests

- `app/src/test/java/org/ole/planet/myplanet/repository/RatingsRepositoryImplTest.kt:50` — `getRatingsById returns specific aggregated rating summary` targets the deleted method. Retarget that one test to `getRatingSummary` (rename and drop the null-id assertion if present); **every other test in the file must pass unedited**.

## Out of scope

- `submitRating`, `isRatingPrompted`, `setRatingPrompted` and the sync methods on the same interface.
- The rest of `ResourceDetailFragment`.
- The `RatingSummary` / `RatingEntry` data classes.

dogi:
Deletes the getRatingsById null-guard wrapper from RatingsRepository interface and implementation. Updates ResourceDetailFragment to guard library.resourceId locally before invoking getRatingSummary directly. Retargets the corresponding unit test in RatingsRepositoryImplTest.

Fixes #17270

---
*PR created automatically by Jules for task [8848093611109918038](https://jules.google.com/task/8848093611109918038) started by @dogi*


dogi:
**Review — rating: 89/100**

Good removal — `getRatingsById` was a wrapper whose only job was to turn a null `resourceId` into a null result, which forced an artificial nullable onto a repository method that otherwise always returns a `RatingSummary`. Pushing the guard to the one call site is the right direction, and it lets the test drop its `assertNotNull` + `!!` pair, which the style guide asks for anyway.

The thing worth checking is that swapping a null return for `?: return@launch` changes control flow, so I traced what the old null path actually did. `BaseContainerFragment.setRatings(RatingSummary?)` is a no-op on null:

```kotlin
fun setRatings(ratingSummary: RatingSummary?) {
    if (ratingSummary != null) { CourseRatingUtils.showRating(...) }
}
```

So with a null `resourceId` the old code set `lastKnownRating = null` and then called `setRatings(null)`, which did nothing. The new code skips both. The only observable difference is that `lastKnownRating` no longer gets reset to null — and that would need `library.resourceId` to go from non-null to null on an already-loaded fragment, which can't happen since `library` is fixed for the fragment's lifetime. If `resourceId` is null from the start, `lastKnownRating` is null in both versions. Equivalent everywhere reachable.

Removing it from the interface also proves the call-site count: anything else calling it would have failed to compile, and both builds are green.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #17280, opened 2026-09-17
chat: smoother search query text filtering (fixes #17269)

### LINKED ISSUE #17269: chat search case-folds on every keystroke against text it already lowercased
## Verified premise

`utils/ChatSearch.kt` normalizes both sides before comparing, then compares case-insensitively anyway.

`fullConvoSearch` (L33-70):

```kotlin
text?.let { Utilities.normalizeText(it) }                                   // L42  candidates
val queryParts = s.split(" ").filterNot { it.isEmpty() }                    // L48
val normalizedQueryParts = queryParts.map { Utilities.normalizeText(it) }   // L49
val normalizedQuery = Utilities.normalizeText(s)                            // L50
...
if (conversation.startsWith(normalizedQuery, ignoreCase = true)) {          // L60
} else if (normalizedQueryParts.all { conversation.contains(it, ignoreCase = true) }) {   // L63
```

`searchByTitle` (L73-101) repeats the same shape at L81, L83, L90, L91, L97, L99.

`Utilities.normalizeText` lowercases (`str.lowercase(Locale.getDefault())`) before stripping diacritics, and **both** operands go through it. The four `ignoreCase = true` flags therefore change no result — they only force `String.startsWith`/`contains` off the fast path onto case-folding `regionMatches`, on every keystroke, for every conversation.

`queryParts` (L48, L89) is also built only to be mapped to `normalizedQueryParts` on the next line and never read again.

## Do

- Drop `ignoreCase = true` from all four comparisons (L60, L63, L97, L99).
- Collapse `queryParts` + `normalizedQueryParts` into one expression at both sites: `val normalizedQueryParts = s.split(" ").filterNot { it.isEmpty() }.map { Utilities.normalizeText(it) }`.

## Constraints

- **Result ordering must stay byte-identical: title-start → title-contains → body-start → body-contains. If you also collapse the ranked `+` concatenations into a single accumulation, the buckets must still be emitted in that order and duplicates handled exactly as the `+` chain handles them today.** Reordering the buckets silently changes which result lands at the top of the search list.
- Do not change `Utilities.normalizeText` — the lowercasing there is what makes this safe.
- Do not switch `lowercase(Locale.getDefault())` to `Locale.ROOT` as part of this; that is a separate behavioural question for Turkish-locale users.

## Tests

- Existing `app/src/test/java/org/ole/planet/myplanet/utils/` chat-search tests must pass **unedited**.
- Add regression cases that would catch a normalization mistake: a mixed-case query (`"HeLLo"`) matching a lowercase conversation, and an accented query (`"café"`) matching `"cafe"`, asserted through the public search entry point.

## Out of scope

- The DAO/repository layer feeding the candidates.
- `Utilities.normalizeText` itself.
- Any change to what counts as a match.

dogi:
Removed `ignoreCase = true` from string comparisons in `ChatSearch.kt` since inputs are already lowercased and diacritics stripped by `Utilities.normalizeText`. Collapsed `queryParts` and `normalizedQueryParts` single-use chains, preserving result ordering and adding regression tests in `ChatSearchTest.kt`.

Fixes #17269

---
*PR created automatically by Jules for task [6897320726730526322](https://jules.google.com/task/6897320726730526322) started by @dogi*


dogi:
**Review — rating: 90/100**

Removing `ignoreCase` is only safe if both operands are already case-folded, so I checked both sides rather than trusting the title. `Utilities.normalizeText` opens with `str.lowercase(Locale.getDefault())` before the NFD/diacritic pass, and every operand in these two comparisons goes through it — the query at lines 49–50 and 90–91, the conversation text at line 42, the title at line 83. So `ignoreCase = true` was genuinely dead, and dropping it also takes `contains`/`startsWith` off the `regionMatches` path onto the fast one.

What makes this better than a bare cleanup is that the two added tests pin the assumption the removal now depends on, one per half of `normalizeText`: `"HeLLo"` against `"hello world"` covers the lowercasing, `"café"` against `"cafe"` covers the diacritic stripping. If anyone later changes `normalizeText` to stop doing either, these fail instead of chat search quietly going case-sensitive. That's the right test to write when you delete a safety net.

I also checked the early return in `normalizeText` — `if (lower.all { it < '' }) return lower` — since it means ASCII and non-ASCII inputs take different paths. It's consistent either way: stripping diacritics from pure ASCII is a no-op, so `"café"` → `"cafe"` matches content `"cafe"` and vice versa. And because both sides use the same locale-sensitive `lowercase`, the Turkish dotted-I case stays symmetric rather than becoming a mismatch.

Inlining `queryParts` into the `normalizedQueryParts` chain is a wash but reads fine.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #17281, opened 2026-09-17
teams: smoother voices repository shared news querying (fixes #17265)

### LINKED ISSUE #17265: answer "already shared?" with a SQL EXISTS instead of hydrating every matching news row
## Verified premise

`repository/VoicesRepositoryImpl.kt:76-80` produces one Boolean by loading whole rows:

```kotlin
override suspend fun isAlreadyShared(chatId: String, viewInId: String): Boolean {
    return newsDao.getByNewsId(chatId).any { news ->
        news.viewIn?.contains("\"_id\":\"$viewInId\"", ignoreCase = true) == true
    }
}
```

and `data/room/dao/NewsDao.kt:71-72` is the widest possible query for that purpose:

```kotlin
@Query("SELECT * FROM news WHERE newsId = :chatId")
suspend fun getByNewsId(chatId: String): List<News>
```

Every row comes back fully hydrated — `message`, `images`, `viewIn` all through `Converters` — and is then discarded. `getByNewsId` has exactly one caller in the whole tree (verified: `VoicesRepositoryImpl.kt:77` and the declaration itself), so it can be replaced outright rather than added alongside.

## Do

Replace the DAO method with an existence check that pushes the substring match into SQLite:

```kotlin
@Query("SELECT EXISTS(SELECT 1 FROM news WHERE newsId = :chatId AND viewIn LIKE :viewInPattern ESCAPE '\\')")
suspend fun isSharedWith(chatId: String, viewInPattern: String): Boolean
```

and build the pattern in `VoicesRepositoryImpl`, escaping the LIKE metacharacters in `viewInId` before wrapping it in `%`:

```kotlin
val escaped = viewInId.replace("\\", "\\\\").replace("%", "\\%").replace("_", "\\_")
val pattern = "%\"_id\":\"$escaped\"%"
```

## Constraints

- **The literal being searched for is `"_id":"<id>"` — it contains an underscore, and an unescaped `_` in a LIKE pattern matches any single character. Escape `\`, `%` and `_` in the interpolated id *and* leave the `_` of the literal `"_id"` key unescaped in the surrounding pattern text, or write that part of the pattern so it cannot be confused with the escaped segment.** Getting this backwards silently widens or narrows the match.
- Escape order matters: replace `\` first, then `%` and `_`.
- SQLite's `LIKE` is ASCII case-insensitive by default, which is what the current `ignoreCase = true` gives — do not add `COLLATE NOCASE` or a `lower()` wrapper.
- No schema or index change, therefore **no `AppDatabase` version bump**.
- Delete `getByNewsId`; do not leave it behind unused.

## Tests

- Add a DAO test under `app/src/test/java/org/ole/planet/myplanet/data/room/dao/` covering: a match, a non-match, and an id containing `%` and `_` that must not match a row holding a different id.
- Existing `app/src/test/java/org/ole/planet/myplanet/data/room/AppDatabaseRoundTripTest.kt` must pass **unedited**.

## Out of scope

- Any other `NewsDao` query.
- The `viewIn` column's JSON shape or its `Converters` handling.
- `getNewsWithReplies` and the rest of `VoicesRepositoryImpl`.

dogi:
Replaces full entity hydration in VoicesRepositoryImpl.isAlreadyShared with a Room SQL EXISTS query (isSharedWith) in NewsDao. Escapes LIKE metacharacters (\, %, _) in order in VoicesRepositoryImpl and deletes the unused getByNewsId DAO method. Adds test coverage in NewsDaoTest.kt.

Fixes #17265

---
*PR created automatically by Jules for task [12970799614827310100](https://jules.google.com/task/12970799614827310100) started by @dogi*


dogi:
**Review — rating: 85/100**

Good swap. The old path loaded every `News` row matching `newsId` — full entities, with the `Converters` Gson pass for the list-valued columns — and then scanned them in Kotlin; the new one returns a single boolean and short-circuits on the first hit. `getByNewsId` had exactly one caller (`VoicesRepositoryImpl:77`), so removing it from the DAO is clean.

The escaping is right, including the part that's easy to get wrong: `\` is replaced first, then `%`, then `_`. Doing the backslash last would have double-escaped the backslashes introduced by the other two. `ESCAPE '\\'` in the Kotlin literal reaches SQLite as `ESCAPE '\'`, which is what the pattern needs.

Null `viewIn` behaves the same across the rewrite — `news.viewIn?.contains(...) == true` was false before, and `NULL LIKE …` is `NULL` so `EXISTS` excludes it now.

Two things worth knowing, neither blocking:

- **The escaping isn't actually covered by a test.** The metacharacter assertions in `NewsDaoTest` are the best part of this PR — proving `team_100%` doesn't match `teamX100Y` is exactly the bug unescaped `LIKE` would introduce. But the test re-implements the three `replace` calls locally and hands the finished pattern to the DAO, so it exercises SQLite's `ESCAPE` handling, not `isAlreadyShared`'s escaping. `ChatViewModelTest` only mocks `isAlreadyShared`. A repository-level test calling `isAlreadyShared("chat_wildcard", "team_100%")` would close that gap and is the one that would catch a future reordering of those replaces.
- **`newsId` is not indexed.** `News` declares `indices = [Index("userId"), Index("replyTo"), Index("_id")]`, so this is still a table scan — the win here is avoiding row materialization and deserialization, not index seek. Adding `Index("newsId")` would genuinely help since it's a plain equality filter (not one of the JSON-list `LIKE '%…%'` columns the style guide warns against indexing), but it costs an `AppDatabase.version` bump and a destructive resync, so it's a real tradeoff rather than a free win. Not adding it here is the right call for this PR.

**Nit:** `ignoreCase = true` in Kotlin is full Unicode case folding, while SQLite's `LIKE` is case-insensitive for ASCII only. For CouchDB `_id`s that's the same thing in practice, so this is a note rather than a finding.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #17286, opened 2026-09-17
resources: smoother viewer user repository view modelling (fixes #17272)

### LINKED ISSUE #17272: take UserRepository out of the two viewer activities and resolve the active user in the ViewModel
## Verified premise

Both viewer activities field-inject `UserRepository` for one purpose — handing it to a coordinator:

```kotlin
// ui/viewer/ResourceViewerActivity.kt:25,28
lateinit var userRepository: UserRepository
...
ResourcesExitCoordinator(this, userRepository, viewModel)

// ui/viewer/WebViewActivity.kt:48,51
lateinit var userRepository: UserRepository
...
ResourcesExitCoordinator(this, userRepository, viewModel)
```

and `ui/viewer/ResourcesExitCoordinator.kt` uses it in exactly one place (L25):

```kotlin
class ResourcesExitCoordinator(
    private val activity: AppCompatActivity,
    private val viewModel: ResourceViewerViewModel,
) {
    ...
        val userId = userRepository.getUserModel()?.id?.takeIf { it.isNotBlank() }
        if (userId == null) { activity.finish(); return@launch }
        ...
        val showDialog = viewModel.shouldShowResourceRatingDialog(userId, resourceId)
        ...
        viewModel.setRatingPrompted(userId, resourceId)
```

`ResourceViewerViewModel` (L25-33) already injects `resourcesRepository`, `ratingsRepository`, `configurationsRepository` and `sharedPrefManager`. The userId is threaded through an Activity and a plain (non-injected) coordinator only to arrive back at the ViewModel that will use it.

## Do

- Inject `UserRepository` into `ResourceViewerViewModel`.
- Drop `userId` from `shouldShowResourceRatingDialog` and `setRatingPrompted` so they take only `resourceId`, resolving the active user inside the ViewModel.
- Drop the `userRepository` parameter from `ResourcesExitCoordinator` and both `lateinit var userRepository` field injections (and their imports).

## Constraints

- **The "no active user" case must still finish the activity without showing the dialog and without marking the resource as prompted. Today that decision is made in the coordinator *before* `handled = true` is set; if resolving the user moves into the ViewModel, `shouldShowResourceRatingDialog` must return `false` for a null/blank user id and `setRatingPrompted` must be a no-op — otherwise a signed-out viewer either sees a rating dialog or silently burns the prompt.**
- Keep the `handled` guard and its ordering in the coordinator: `handled` is still set only once a user-gated dialog decision is actually being made.
- Keep `getUserModel()?.id?.takeIf { it.isNotBlank() }` semantics — blank counts as absent.
- `ResourceViewerViewModel.isRatingPrompted(userId, resourceId)` (L52) stays as it is; it is called internally by `shouldShowResourceRatingDialog` and is not part of this change's surface.

## Tests

- `app/src/test/java/org/ole/planet/myplanet/ui/viewer/ResourceViewerViewModelTest.kt` mocks `ratingsRepository.isRatingPrompted(userId, resourceId)` at L119/L138/L157 — those mocks stay, since the repository signature does not change; only the ViewModel call sites need the new user lookup stubbed.
- Add a ViewModel test: `getUserModel()` returns null → `shouldShowResourceRatingDialog(resourceId)` is false and `setRatingPrompted(resourceId)` touches nothing on `ratingsRepository`.

## Out of scope

- `RatingsRepository`'s own signatures.
- `RatingsFragment` and the dialog itself.
- Playback-progress, PDF and download methods on the same ViewModel.

dogi:
Injected UserRepository into ResourceViewerViewModel to resolve active user ID internally in shouldShowResourceRatingDialog and setRatingPrompted. Removed UserRepository field injections and parameters from ResourceViewerActivity, WebViewActivity, and ResourcesExitCoordinator.

Fixes #17272

---
*PR created automatically by Jules for task [2376874078797535109](https://jules.google.com/task/2376874078797535109) started by @dogi*


dogi:
**Review — rating: 86/100**

Right direction — two Activities were field-injecting `UserRepository` purely to resolve an ID and thread it through `ResourcesExitCoordinator` into the ViewModel. Moving the lookup into `ResourceViewerViewModel` lets both Activities and the coordinator drop the dependency entirely, and the coordinator's constructor gets narrower as a result.

The part I checked closely is the reordering in `ResourcesExitCoordinator`, since moving the user lookup behind the `handled` guard changes control flow. It comes out ahead:

- **Null user still finishes.** Old code returned early with `activity.finish()`; new code lets `shouldShowResourceRatingDialog` return false and falls into the `else` branch, which also calls `activity.finish()`. Same outcome.
- **The double-dialog race is actually closed.** Previously `handled = true` was set *after* the suspending `getUserModel()`, so two invocations could both clear the `if (handled)` check while suspended. Now the check-and-set happens before any suspension point, and since `lifecycleScope` dispatches with `Main.immediate` the body runs synchronously up to that point. That's a real improvement, not just a move.
- The old null-user path also left `handled` false, so a second invocation would call `finish()` again; the new one is idempotent.

The tests are the right ones: null user, blank user id, `setRatingPrompted` no-op on null user, and the happy path pinned with `coVerify(exactly = 1)`. That covers each new branch rather than just re-greening the existing three.

**Nit:** on the path where the dialog does show, `getUserModel()` is now called twice — once inside `shouldShowResourceRatingDialog` and again inside `setRatingPrompted` — where the coordinator used to resolve it once and pass the ID down. Not hot code, but if you wanted it back to one lookup, having `shouldShowResourceRatingDialog` return the resolved id (or caching the user in the ViewModel) would do it.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #17287, opened 2026-09-17
all: smoother configurations repository check version providing (fixes #17266)

### LINKED ISSUE #17266: drop the redundant SharedPrefManager parameter from ConfigurationsRepository.checkVersion
## Verified premise

`repository/ConfigurationsRepository.kt:15`:

```kotlin
fun checkVersion(callback: CheckVersionCallback, spm: SharedPrefManager)
```

The implementation already injects its own:

```kotlin
class ConfigurationsRepositoryImpl @Inject constructor(
    ...
    private val sharedPrefManager: SharedPrefManager,
    ...
) : ConfigurationsRepository {
```

and uses **both** inside the same function — the parameter at `ConfigurationsRepositoryImpl.kt:99` (`UrlUtils.baseUrl(spm)`) and the injected field a few lines later (`sharedPrefManager.rawPreferences.getLong("last_version_check_timestamp", 0)`). Two `SharedPrefManager` instances can therefore reach one call. The parameter also pins a `services` import into a repository *interface*.

All three call sites pass exactly what the impl already holds:

- `ui/sync/SyncActivity.kt:233` — `configurationsRepository.checkVersion(this@SyncActivity, prefData)`
- `ui/sync/LoginActivity.kt:183` — `configurationsRepository.checkVersion(this, prefData)`
- `services/AutoSyncWorker.kt:69` — `configurationsRepository.checkVersion(this@AutoSyncWorker, sharedPrefManager)`

## Do

- Drop the parameter from the interface declaration and the impl override.
- Inside the impl, use the injected `sharedPrefManager` everywhere the parameter was used.
- Update the three call sites to the one-argument form.
- Remove the now-unused `SharedPrefManager` import from `ConfigurationsRepository.kt`.

## Constraints

- **The private helpers `fetchVersionInfo(spm: SharedPrefManager)` and `fetchApkVersionString(spm: SharedPrefManager)` (around L484 and L495) still take the manager as a parameter — pass the injected field into them rather than deleting their parameters too.** Widening this into a helper-signature sweep is what turns a mechanical change into a conflict with anything else touching the file.
- `UrlUtils.baseUrl(...)` is called *before* the `serviceScope.launch`, on the caller's thread — keep that ordering; only the argument changes.
- Do not change `CheckVersionCallback` or any of its methods.
- Leave `SharedPrefManager` injected in `SyncActivity`, `LoginActivity` and `AutoSyncWorker` — each uses `prefData`/`sharedPrefManager` for other things.

## Tests

- `app/src/test/java/org/ole/planet/myplanet/repository/ConfigurationsRepositoryImplTest.kt` has three `repository.checkVersion(callback, sharedPrefManager)` call sites (L171, L221, L288) that need the argument dropped. Every assertion in that file — including `coVerify(exactly = 1) { apiInterface.checkVersion(any()) }` at L293 — must still hold **with only the call-site arity edited**, no assertion changes.

## Out of scope

- Any other method on `ConfigurationsRepository`.
- `UrlUtils.baseUrl` / `UrlUtils.getUpdateUrl` themselves.
- The version-check throttling logic and the 24-hour constant.

dogi:
Drop the redundant SharedPrefManager parameter from ConfigurationsRepository.checkVersion and ConfigurationsRepositoryImpl.checkVersion, update usages to pass injected sharedPrefManager, and update call sites in SyncActivity, LoginActivity, AutoSyncWorker, and ConfigurationsRepositoryImplTest.

Fixes #17266

---
*PR created automatically by Jules for task [4022746229409676413](https://jules.google.com/task/4022746229409676413) started by @dogi*


dogi:
**Review — rating: 89/100**

The parameter really was redundant, and I checked that rather than inferring it from the name. `SharedPrefManager` is annotated `@Singleton`, and every caller was handing over that same instance: `AutoSyncWorker` passed its own injected `sharedPrefManager`, while `LoginActivity` and `SyncActivity` passed `prefData`, which is the field-injected `SharedPrefManager` declared on `ProcessUserDataActivity:51` that `SyncActivity` inherits. Same object in all three cases, and the repository already holds it via constructor injection.

What makes this more than cosmetic is that the old body was internally inconsistent about it: `checkVersion` used the `spm` parameter for `UrlUtils.baseUrl(spm)`, `fetchVersionInfo(spm)` and `fetchApkVersionString(spm)`, but reached for the *injected* `sharedPrefManager` for `setLastWifiId()` and `setVersionDetail()`. That only ever worked because the two are the same singleton — had a caller passed a different instance, the method would have read config from one and written to another. Collapsing onto the injected field removes that latent trap, not just an argument.

All three production call sites and all three test call sites are updated, and dropping the `SharedPrefManager` import from the `ConfigurationsRepository` interface is correct since the contract no longer mentions it.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #17288, opened 2026-09-17
Move storage breakdown scanning from ViewModel to ResourcesRepository

### LINKED ISSUE #17273: move the storage-breakdown disk walk out of the ViewModel into ResourcesRepository
## Verified premise

`ui/settings/StorageBreakdownViewModel.kt:86-103` does raw filesystem I/O inside a ViewModel:

```kotlin
internal open fun scanStorage(oleDir: File): ScanResult {
    val sizes = LongArray(categories.size)
    val counts = IntArray(categories.size)

    if (!oleDir.exists() || !oleDir.isDirectory) return ScanResult(0L, sizes, counts)

    var total = 0L

    oleDir.walkTopDown().filter { it.isFile }.forEach { file ->
        val ext = file.extension
        val index = if (ext.isEmpty()) StorageCategories.OTHER_INDEX else StorageCategories.indexOf(ext)
        val size = file.length()
        total += size
        sizes[index] += size
        counts[index]++
    }
    return ScanResult(total, sizes, counts)
}
```

`repository/ResourcesRepositoryImpl.kt:860`, inside `getOfflineResourceItems` (declared at L844), already runs the identical traversal over the same tree:

```kotlin
oleDir.walkTopDown().filter { it.isFile }.forEach { file ->
```

So the settings screen walks the `ole` tree once for the summary and again when a category is opened, and this ViewModel is the only one in `ui/` still doing its own file I/O.

## Do

- Add `getStorageBreakdown(oleDir: File): StorageBreakdown` to `ResourcesRepository` / `ResourcesRepositoryImpl`, lifting the `scanStorage(oleDir)` body **verbatim** into `withContext(dispatcherProvider.io) { ... }`.
- Move `ScanResult` into the repository layer as the returned type (rename to `StorageBreakdown` or keep the name — but it belongs beside the new method, not in the ViewModel).
- Have `StorageBreakdownViewModel` call the repository and delete `scanStorage(oleDir)`, `scanStorage(context)` and the local `ScanResult`.
- Keep the caller passing the `ole` path — `File(FileUtils.getOlePath(context))`, exactly as `getOfflineResourceItems` already receives it.

## Constraints

- **`app/src/test/java/org/ole/planet/myplanet/ui/settings/StorageBreakdownFragmentTest.kt` spies the ViewModel and asserts the cache behaviour through `verify(exactly = 0) { viewModel.scanStorage(any<File>()) }` (L148) and `verify(exactly = 1) { ... }` (L154). Those two verifications are the only thing pinning "an unforced reload must not re-walk the disk" — they must be re-expressed against the repository call, not deleted. Dropping them silently removes the cache guarantee.**
- `scanStorage` is `internal open` specifically so the test can spy it; whatever replaces it on the seam must stay mockable the same way.
- The category indexing (`StorageCategories.indexOf(ext)`, `OTHER_INDEX` for an empty extension) moves unchanged — do not "improve" it here.
- The `!oleDir.exists() || !oleDir.isDirectory` early return must stay, returning zeroed arrays of `categories.size`.
- The new repository method must not hold `Context`; take the `File`.

## Tests

- Move the two fixture-tree cases — `scanStorage produces identical counts and total sizes for fixture tree` (L57) and `scanStorage handles empty or nonexistent directory` (L107) — to a repository test, keeping their assertions byte-identical.
- `app/src/test/java/org/ole/planet/myplanet/ui/settings/StorageCategoriesTest.kt` and `StorageCategoryViewModelTest.kt` must pass **unedited**.

## Out of scope

- `getOfflineResourceItems` itself — do not try to merge the two walks into one pass in this change.
- `StorageCategories` and its index constants.
- `StorageCategoryDetailFragment` and its ViewModel.

dogi:
Refactored storage breakdown scanning:
- Added StorageBreakdown data class and getStorageBreakdown(oleDir: File) to ResourcesRepository and ResourcesRepositoryImpl.
- Updated StorageBreakdownViewModel to inject ResourcesRepository and call getStorageBreakdown, removing scanStorage and local ScanResult.
- Moved fixture-tree unit tests to ResourcesRepositoryImplTest.
- Updated StorageBreakdownFragmentTest to verify cache behavior on the repository seam using MockK coVerify.

Fixes #17273

---
*PR created automatically by Jules for task [4135283504306658428](https://jules.google.com/task/4135283504306658428) started by @dogi*


dogi:
**Review — rating: 72/100**

The core move is right: `walkTopDown()` and `file.length()` don't belong in a ViewModel, `ResourcesRepository` already owns comparable filesystem work (`getOfflineResourceItems`, `deleteOfflineResources`), and wrapping the scan in `withContext(dispatcherProvider.io)` means the repository guarantees its own threading rather than trusting the caller. Moving the two fixture tests into `ResourcesRepositoryImplTest` alongside the code is correct, and replacing `spyk(StorageBreakdownViewModel(...))` with a mocked repository is a genuine test-quality win — verifying against a spy of the class under test was the fragile part of the old test.

**The layering goes the wrong way, though.** `ResourcesRepositoryImpl` now does:

```kotlin
import org.ole.planet.myplanet.ui.settings.StorageCategories
```

`StorageCategories` lives in `ui/settings/` and its `StorageCategory` carries `@StringRes val nameRes: Int` off `R` — it is a UI type. So a PR whose purpose is pushing work down into the data layer ends up making the data layer depend on the UI layer, which is a worse coupling than the one it removed.

The same interface already shows the intended pattern: `getOfflineResourceItems(oleDirPath, extensions, allKnownExtensions)` takes the extension sets as *parameters* precisely so the repository never has to know about `StorageCategories`, and the caller passes `StorageCategories.allKnownExtensions` in. `getStorageBreakdown` should follow that precedent, or `StorageCategories` should move to a neutral package (`utils/` or `model/`) with the `@StringRes` mapping left behind in `ui/settings`.

Two smaller things:

- **`StorageBreakdown` is a `data class` holding `LongArray`/`IntArray`,** so its generated `equals`/`hashCode` compare array *identity*, not contents — two structurally identical results are unequal. Nothing compares them today, so it isn't a live bug, but `ScanResult` was `internal` to the ViewModel and this is now public API on the repository interface, so the trap is considerably more exposed. `List<Long>`/`List<Int>` would remove it outright.
- **An unrelated test got weaker.** In `getOfflineResourceItems calculates size and path order in single pass`, the assertion became `assertEquals(listOf(...).sorted(), res1Item.filePaths.sorted())`. That test's stated subject is *path order*, and sorting both sides is exactly the assertion that can no longer detect a wrong order. `getOfflineResourceItems` isn't touched by this PR, so either it was already order-nondeterministic — in which case the production code should sort and the test should say so — or this is scope creep that should come out of the diff. Either way it needs an explanation rather than a silent loosening.

**Nit:** the ViewModel already launches on `dispatcherProvider.io`, so the repository's `withContext(io)` is now nested. That's harmless and I'd keep it — the repository shouldn't depend on callers picking the right dispatcher.

**Verdict:** needs another round — get the `ui.settings` import out of the repository (parameterize it the way `getOfflineResourceItems` does), and either restore or explain the `filePaths` assertion.

---

dogi:
**Re-review — Rating: 80/100** (was 68)

**Verdict:** mergeable — the array-equality bug is fixed and the tests moved to the right layer. One architectural point worth acting on before it sets.

**The custom `equals`/`hashCode` is the right fix and was necessary.** A `data class` holding `LongArray`/`IntArray` generates `equals` over array *references*, so two structurally identical breakdowns compared unequal — which matters here because `StorageBreakdownUiState` flows through a `StateFlow`, and `MutableStateFlow` skips emission on `==`. With reference equality, every scan looked like a change; the override with `contentEquals`/`contentHashCode` makes the comparison mean what the call site assumes.

**The test migration is genuinely better, not just relocated.** The fixture tests moved to `ResourcesRepositoryImplTest` where the logic now lives, with the same tree and assertions preserved. And the caching test improved in the process: it used to `spyk` the ViewModel and verify a call on the subject under test; it now mocks `ResourcesRepository` and does `coVerify(exactly = 0/1) { getStorageBreakdown(any()) }`. Verifying a collaborator instead of spying on the class you are testing is the right shape, and it survives the subject being refactored.

Also noticed: `accumulator.filePaths.sort()` in `getOfflineResourceItems`. That is the `filePaths` ordering that several PRs in this series have had competing assertions about — making it deterministic at the source is the right way to settle it rather than each test asserting its own expectation.

## The thing to fix

**`ResourcesRepositoryImpl` now imports from the UI layer:**

```kotlin
import org.ole.planet.myplanet.ui.settings.StorageCategories
```

The repository sizes its arrays from `StorageCategories.all.size` and maps extensions through `StorageCategories.indexOf(ext)` — so a data-layer class now depends on a constant object that lives in `ui.settings`. The PR moved the logic down a layer but left its data dependency pointing up, which inverts the dependency direction the rest of the package structure follows.

It compiles and works, which is exactly why it is worth doing now: a `ui` import inside a repository is the kind of thing that never gets cleaned up once it is merged and other code starts relying on the arrangement. The fix is a file move — `StorageCategories` to `utils` or alongside the model classes — and nothing about the logic changes.

Related, and a bigger question than this PR: the return value is two parallel arrays whose indices are defined by a UI-layer ordering. `sizes[0]` meaning "videos" is an unwritten contract between the repository and the fragment. A `Map<StorageCategory, CategoryStats>` would carry its own meaning, and the tests would stop needing comments like `// Videos: 2 files, 300 bytes` to be readable.

**One smaller note:** `StorageBreakdownViewModel` now takes the whole `ResourcesRepository` for a single method. Worth contrasting with #17300 in this same queue, which creates a narrow `UserAchievementsRepository` precisely to avoid that. Not worth a dedicated interface for one call today, but the two PRs are pulling in opposite directions on the same question.

CI green 5/5.

---

Okuro3499 (changes requested):
``` 
File: app/src/main/java/org/ole/planet/myplanet/repository/ResourcesRepositoryImpl.kt:37 and getStorageBreakdown() (line 896)                                                    
    Problem: The repository imports StorageCategories from org.ole.planet.myplanet.ui.settings. Repository-layer code should never depend on the ui/ package — this is backwards per 
    the app's layered architecture (UI depends on repositories, not the reverse), and it means any future UI-only change to StorageCategories (e.g. adding a Context-aware helper, a 
    Compose-specific field) risks breaking compilation of the data layer, or forces UI code to stay artificially "pure" just to keep the repository buildable.    
```

dogi:
**Re-review — Rating: 90/100** (was 80)

**Verdict:** `ready`. The layering inversion is properly resolved, and the way it was resolved makes an implicit contract explicit.

**The dependency now points the right way.** Last round `ResourcesRepositoryImpl` imported `org.ole.planet.myplanet.ui.settings.StorageCategories` — the data layer reaching into the UI layer for its classification rules. The split puts the taxonomy in `model/StorageCategoryType` (extensions, ordering, `indexOf`, `allKnownExtensions`) and leaves `ui/settings/StorageCategories` as the display side that pairs each category with its `@StringRes`. The repository now imports from `model/`, and `StorageCategoryDetailFragment` reads `StorageCategoryType.allKnownExtensions`.

Keeping the labels out of the model via a private extension property is the right touch — no `R` reference crosses into `model/`:

```kotlin
@get:StringRes
private val StorageCategoryType.nameRes: Int
    get() = when (this) { … }
```

**The array-index contract went from a comment to a test.** This is the part I'd have worried about: the whole design rests on `sizes[]` / `counts[]` positions matching category order, and that used to be a hardcoded `const val OTHER_INDEX: Int = 4` next to a hand-ordered list. Now it's `OTHER.ordinal`, and `StorageCategoryTypeTest` asserts the canonical order and that `OTHER_INDEX == entries.lastIndex` outright. Reordering the enum now fails a test instead of silently mislabelling every storage category in the UI.

**I checked the one behaviour change.** The old scan special-cased empty extensions:

```kotlin
val index = if (ext.isEmpty()) StorageCategories.OTHER_INDEX else StorageCategories.indexOf(ext)
```

The new code just calls `StorageCategoryType.indexOf(file.extension)`. That's equivalent — `""` isn't in the extension map, so it falls through to `OTHER_INDEX` — and `indexOf falls back to the other category for unknown extensions` asserts `indexOf("")` explicitly, so the removed branch is covered rather than merely assumed. The `README` fixture in the repository test exercises it end to end.

**Tests moved with the code, not alongside it.** Classification cases are now in `model/StorageCategoryTypeTest` against the real enum; the scan fixture tree moved into `ResourcesRepositoryImplTest` where it tests the actual implementation rather than an `internal open` seam on a ViewModel; `StorageCategoriesTest` keeps only labels and ordering, plus a new check that `all` mirrors the model one entry at a time. The `spyk(viewModel)` / `verify { viewModel.scanStorage(any()) }` pattern is gone in favour of `coVerify` on the mocked repository — testing the collaboration instead of the object spying on itself. And `clearMocks(resourcesRepository, answers = false)` correctly keeps the `@Before` stub alive across the clear.

**Also still good from earlier rounds:** the custom `equals`/`hashCode` on `StorageBreakdown` — a `data class` with `LongArray`/`IntArray` members would otherwise compare by reference and break every `assertEquals` and every `StateFlow` de-duplication — and `accumulator.filePaths.sort()`, which settles the `filePaths` ordering that several PRs in this series had competing test assertions about, in production code rather than by loosening assertions.

CI green 5/5.

**Nit:** `getStorageBreakdown(oleDir: File)` takes a `File` while its neighbour `getOfflineResourceItems(oleDirPath: String, …)` takes a path string, so `ResourcesRepository` now has both conventions side by side. Not worth churn on its own, but worth picking one the next time either signature is touched.

---

Okuro3499 (approved):

PR #17293, opened 2026-09-17
resources: smoother library attachments inserting (fixes #17268)

### LINKED ISSUE #17268: re-syncing a resource grows its attachment array without bound
## Verified premise

`model/MyLibrary.kt`, `insertMyLibrary`, the `_attachments` block on master:

```kotlin
if (params.doc.has("_attachments")) {
    val attachmentsObj = params.doc["_attachments"].asJsonObject
    val attachmentList = this.attachments?.toMutableList() ?: mutableListOf()   // seeded from what is already stored

    attachmentsObj.entrySet().forEach { (key, attachmentValue) ->
        val attachmentObj = attachmentValue.asJsonObject
        val realmAttachment = Attachment().apply {
            id = UUID.randomUUID().toString()
            name = key
            ...
        }
        attachmentList.add(realmAttachment)                                     // unconditional

        if (key.indexOf("/") < 0) {
            resourceRemoteAddress = "${params.spm.getCouchdbUrl().ifEmpty { "http://" }}/resources/$resourceId/$key"
            ...
        }
    }
    this.attachments = attachmentList                                           // written back
}
```

The list is seeded from the **existing** attachments, appended to unconditionally, and written back. Re-syncing the same CouchDB document therefore appends a fresh `Attachment` for every key it already has — the array grows on every sync, forever, and each entry is persisted through `Converters` as JSON.

Separately, `params.spm.getCouchdbUrl()` is a SharedPreferences read executed inside the loop, once per attachment whose key has no `/`.

## Do

- Before the `forEach`, resolve the base URL once: `val couchdbUrl = params.spm.getCouchdbUrl().ifEmpty { "http://" }` and use it inside the loop.
- Before the `forEach`, capture the names already present: `val existingNames = attachmentList.mapTo(mutableSetOf()) { it.name }`, and skip the `add` when `key` is already in that set (still running the `key.indexOf("/") < 0` branch, which sets fields rather than appending).

## Constraints

- **Attachment identity is the CouchDB `name` (the `_attachments` map key), never `id` — `id` is a fresh `UUID.randomUUID()` per construction, so deduping on it would never match anything and the array would keep growing.**
- The `if (key.indexOf("/") < 0)` branch must still run for every key, including skipped ones: it sets `resourceRemoteAddress`, `resourceLocalAddress`, `resourceOffline` and `downloadedRev` on the entity, not on the attachment.
- Do not clear `attachmentList` and rebuild it from the document — an attachment stored locally but absent from this particular payload should survive, which is the existing behaviour.
- `Attachment.name` may be null in already-stored rows; a null in the set must not make a real key look present.
- No schema change, therefore **no `AppDatabase` version bump**.

## Tests

- Add a test that calls `insertMyLibrary` twice with the same document and asserts `attachments` has the same size both times, and one that asserts a second document adding a new key appends exactly one entry.
- Existing `app/src/test/java/org/ole/planet/myplanet/data/room/AppDatabaseRoundTripTest.kt` must pass **unedited**.

## Out of scope

- The rest of `insertMyLibrary` (title/description/rating field mapping).
- `Attachment`'s own shape and its converter.
- `FileUtils.checkFileExist` and the offline-resource bookkeeping.

dogi:
Fixes an issue where re-syncing a resource document grows its attachments array without bound.

Changes:
1. MyLibrary.insertMyLibrary:
   - Pre-resolves couchdbUrl once before the attachments loop.
   - Captures existing attachment names into a set (existingNames).
   - Skips adding duplicate Attachment objects when key is already present in existingNames.
   - Continues executing resource address and offline status updates for all attachment keys.
2. MyLibraryTest:
   - Added unit tests verifying that re-syncing the same document maintains attachment array size.
   - Added test verifying that a document with a new key appends exactly one entry.
   - Added test verifying resource address fields are updated even when attachment addition is skipped.

Fixes #17268

---
*PR created automatically by Jules for task [8854574573907811961](https://jules.google.com/task/8854574573907811961) started by @dogi*


dogi:
**Review — rating: 85/100**

Real bug, and the fix is placed precisely. On every re-sync the old loop appended every key in `_attachments` to the existing list unconditionally, so a resource's `attachments` grew by its full attachment count each time it synced. Seeding `existingNames` from the current list and skipping known keys stops that.

The part that's easy to get wrong here is scope, and this PR gets it right: the `resourceRemoteAddress` / `resourceLocalAddress` / `resourceOffline` block stays **outside** the dedup guard. Wrapping the whole loop body would have frozen a resource's URL at whatever CouchDB host it first synced from — and the third test pins exactly that, re-running the insert after changing `getCouchdbUrl()` and asserting the address moves to `http://newserver:5984/...` while the attachment count stays at 1. That's the right test to have written. Hoisting `couchdbUrl` out of the loop is a sensible bonus.

The three tests cover the change well — idempotent re-insert, additive new key, and the address refresh — and `MyLibraryTest` is a new file, so this model had no coverage before.

One gap worth knowing about, though **not** a regression: matching on name alone means an attachment whose *content* changed upstream but kept its filename is now skipped, leaving the stale `length`/`digest`/`revpos` on the existing row. The old code didn't handle this either — it appended a second entry with the same `name`, so any consumer doing `find { it.name == key }` still got the stale one, plus a duplicate. So this is strictly better; it just leaves the adjacent problem unfixed. Since `digest` and `revpos` are captured precisely to detect a changed attachment, a follow-up could replace the existing entry when either differs instead of skipping.

**Nit:** `existingNames.add(key)` inside the branch is defensive only — Gson's `JsonObject` is backed by a `LinkedHashMap`, so `entrySet()` can't yield the same key twice. Harmless, and I'd leave it.

**Verdict:** mergeable.

---

Okuro3499 (approved):

Okuro3499 (approved):

PR #17294, opened 2026-09-17
sync: smoother resource id list filtering (fixes #17271)

### LINKED ISSUE #17271: resource sync copies every id a second time to re-check a condition that is always true
## Verified premise

`services/sync/SyncManager.kt`, `resourceTransactionSync`.

The accumulator is declared nullable (L294):

```kotlin
val newIds: MutableList<String?> = ArrayList()
```

but the only thing ever put into it is already non-blank (L375-377):

```kotlin
if (savedIds.isNotEmpty()) {
    val validIds = savedIds.filter { it.isNotBlank() }
    newIds.addAll(validIds)
    processedItems += validIds.size
}
```

so the cleanup block re-walks the whole list to re-establish what is already guaranteed (L411-420):

```kotlin
val validNewIds = newIds.filter { !it.isNullOrBlank() }
if (hadBatchFailure) {
    syncTimeLogger.logDetail("resource_sync", "Skipping delete-cleanup: one or more batches failed, id list is incomplete")
} else if (validNewIds.isNotEmpty() && validNewIds.size == newIds.size) {
    resourcesRepository.removeDeletedResources(validNewIds)
}
...
    syncTimeLogger.logDbOperation("delete_cleanup", "resources", cleanupDuration, newIds.size - validNewIds.size)
```

`validNewIds.size == newIds.size` is therefore always true, and the logged item count `newIds.size - validNewIds.size` is always 0. On a server holding tens of thousands of resources this allocates a second full copy of the id list per sync.

## Do

- Declare `newIds` as `MutableList<String>` (L294).
- Delete the `validNewIds` line and use `newIds` directly in the guard and in `removeDeletedResources(newIds)`.
- Reduce the guard to `else if (newIds.isNotEmpty())`.
- Log `newIds.size` as the item count on the `logDbOperation` line, since that is what the cleanup actually covered.

## Constraints

- **The `hadBatchFailure` branch must stay exactly as it is and must stay first. It is the one real guard here — it suppresses the delete-cleanup when the id list is incomplete, and deleting "resources not in `newIds`" against a partial list would wipe live rows.**
- `savedIds.filter { it.isNotBlank() }` at L376 stays — that is where the invariant is established.
- Do not change `resourcesRepository.removeDeletedResources`'s signature; it already takes non-null ids.
- Leave the surrounding `try`/`catch` and the `syncTimeLogger.startProcess`/`endProcess` pairing untouched.

## Tests

- Existing tests under `app/src/test/java/org/ole/planet/myplanet/services/sync/` must pass **unedited** — in particular any covering `resourceTransactionSync` and the batch-failure path.
- Add a test asserting `removeDeletedResources` is **not** called when `hadBatchFailure` is true, and is called with the full id list when it is false.

## Out of scope

- The batching and pagination above L376.
- `AdaptiveBatchProcessor` and the batch-size logic.
- Any other table's sync in the same file.

dogi:
Declared newIds as MutableList<String> in SyncManager.resourceTransactionSync and eliminated the redundant validNewIds filter during resource delete-cleanup, preventing unnecessary list allocation and iteration.

Fixes #17271

---
*PR created automatically by Jules for task [8518032741941076718](https://jules.google.com/task/8518032741941076718) started by @dogi*


dogi:
**Review — rating: 86/100**

Removing a guard in front of a delete path deserves proof rather than a plausible argument, so I followed the ids the whole way. `batchInsertResources` only collects ids that pass `id.isNotBlank() && !id.startsWith("_design")`, and its return type is `List<String>`. `SyncManager:375` then filters *again* — `val validIds = savedIds.filter { it.isNotBlank() }` — before `newIds.addAll(validIds)`. So by the time the cleanup block runs, `newIds` cannot contain a null or a blank, `validNewIds` was always equal to `newIds`, and `validNewIds.size == newIds.size` was always true. The old condition therefore reduced to exactly `newIds.isNotEmpty()`, which is what replaces it. Equivalent, and the `MutableList<String?>` → `MutableList<String>` change makes that guarantee hold at compile time instead of by convention.

Worth noting what that implies about the original intent: the `size ==` comparison looks like it was meant to mean "some id was dropped, so the list is incomplete — don't delete." It could never fire, because the drop happens upstream at line 375 and never reaches `newIds`. This PR removes an illusion of protection, not protection. The genuine incompleteness risk is the `hadBatchFailure` path, which is preserved and now has a test.

The two new tests are the right pair — one asserting `removeDeletedResources` is never called after a batch throws, one asserting it's called exactly once with the full id list — and they exercise the real branch rather than mocking around it.

**Nit:** two small things.
- `logDbOperation("delete_cleanup", "resources", cleanupDuration, newIds.size)` — the old fourth argument was `newIds.size - validNewIds.size`, i.e. always `0`, so it was meaningless; but `newIds.size` is the count of ids *kept*, not the count of rows the cleanup deleted, which is what an itemCount on `delete_cleanup` reads as. Having `removeDeletedResources` return the deleted count would make this number mean what it says.
- The new `com.google.gson.JsonArray`, `com.google.gson.JsonObject` and `retrofit2.Response` imports land in the middle of the `org.ole.planet.*` block; the style guide asks for the import block to stay alphabetically sorted.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #17295, opened 2026-09-17
Extract challenge progress arithmetic from DashboardViewModel

### LINKED ISSUE #17274: ChallengePrompter reaches into DashboardViewModel for two pure arithmetic helpers
## Verified premise

`services/ChallengePrompter.kt` is a service that imports and holds a UI ViewModel:

```kotlin
import org.ole.planet.myplanet.ui.dashboard.DashboardViewModel      // L8

class ChallengePrompter(
    private val activity: DashboardActivity,
    private val sharedPrefManager: SharedPrefManager,
    private val viewModel: DashboardViewModel                        // L13
)
```

and uses it for exactly two things, twice each — L41-42 in the completed branch and L55-56 in the in-progress branch:

```kotlin
${activity.getString(R.string.community_earnings, viewModel.calculateCommunityProgress(allVoiceCount, hasUnfinishedSurvey))}
${activity.getString(R.string.your_earnings, viewModel.calculateIndividualProgress(voiceCount, hasUnfinishedSurvey))}
```

Both targets are deterministic arithmetic with no ViewModel state — `ui/dashboard/DashboardViewModel.kt:155-167`:

```kotlin
fun calculateIndividualProgress(voiceCount: Int, hasUnfinishedSurvey: Boolean): Int {
    val earnedDollarsVoice = minOf(voiceCount, 5) * 2
    val earnedDollarsSurvey = if (!hasUnfinishedSurvey) 1 else 0
    val total = earnedDollarsVoice + earnedDollarsSurvey
    return total.coerceAtMost(500)
}

fun calculateCommunityProgress(allVoiceCount: Int, hasUnfinishedSurvey: Boolean): Int {
    val earnedDollarsVoice = minOf(allVoiceCount, 5) * 2
    val earnedDollarsSurvey = if (!hasUnfinishedSurvey) 1 else 0
    val total = earnedDollarsVoice + earnedDollarsSurvey
    return total.coerceAtMost(11)
}
```

`ChallengePrompter` and `DashboardViewModel` are their only callers in `app/src/main`. The dependency runs services → ui, inverting the layer direction.

## Do

- Extract both calculations into a platform-free top-level object or file-level functions (no Android imports, no ViewModel), e.g. in `utils/` or beside the challenge model.
- Have `DashboardViewModel.calculateIndividualProgress` / `calculateCommunityProgress` delegate to it — **keep both methods on the ViewModel, delegating; do not copy the bodies and do not delete the methods.**
- Have `ChallengePrompter` call the extracted functions directly, and drop the `DashboardViewModel` constructor parameter and its import if nothing else in the class uses it.

## Constraints

- **The two caps are different and deliberately so — individual coerces to 500, community to 11 — and the Spanish completion check `courseStatus.contains("terminado", ignoreCase = true)` at L27 and L30 drives the ✅/`[ ]` markers. Preserve both caps exactly and leave the "terminado" check where it is; it is not part of the arithmetic being extracted.**
- The markdown copy, the `✅` / `[ ]` glyphs, the `<br/>` placement and the `minOf(voiceCount, 5)` cap on `cappedVoiceCount` all stay byte-identical.
- Check whether `viewModel` is used anywhere else in `ChallengePrompter` before removing the parameter; if it is, leave the parameter and only change these four call sites.
- `activity.getString(...)` stays in the prompter — string resources are a UI concern and are not moving.

## Tests

- `app/src/test/java/org/ole/planet/myplanet/ui/dashboard/DashboardViewModelTest.kt` must pass **unedited**: `calculateIndividualProgress handles boundary cases properly` (L222) and `calculateCommunityProgress handles boundary cases properly` (L238) call through the ViewModel and pin every boundary, including the two different caps.
- Add a small test for the extracted functions directly, asserting the same boundaries without a ViewModel.

## Out of scope

- `MarkdownDialogFragment` and the dialog itself.
- `ChallengeDialogData` and where its fields come from.
- The `hasShownCongrats` SharedPreferences flag and its lifecycle.

dogi:
Extracts calculateIndividualProgress and calculateCommunityProgress into top-level functions in org.ole.planet.myplanet.utils.ChallengeUtils. Refactors ChallengePrompter to call the extracted functions directly and removes its unused DashboardViewModel dependency. Updates DashboardViewModel to delegate to ChallengeUtils and adds unit tests in ChallengeUtilsTest.

Fixes #17274

---
*PR created automatically by Jules for task [1307930654563115555](https://jules.google.com/task/1307930654563115555) started by @dogi*


dogi:
**Review — rating: 70/100**

The direction is right and the best part is real: `ChallengePrompter` was holding a whole `DashboardViewModel` purely to do two lines of arithmetic, and dropping that constructor parameter is a genuine decoupling win. Pure functions in `utils/` is where the style guide points, and the boundary tests are well chosen.

But the extraction is only half finished, and the unfinished half is what forces a style-guide violation:

```kotlin
return org.ole.planet.myplanet.utils.calculateIndividualProgress(voiceCount, hasUnfinishedSurvey)
```

`CODE_STYLE_GUIDE.md` lists inline fully-qualified names under *Things to Avoid*, and names the remedy for exactly this situation — "On a name collision, use an import alias instead of falling back to the fully-qualified name." The FQN is here only because `DashboardViewModel` still declares members with the same names, which would shadow a plain import and recurse.

That points at the real fix. After this PR, `ChallengePrompter` calls the utils directly, so I checked who still calls the ViewModel versions: nothing in production does. Their only remaining callers are the 18 assertions in `DashboardViewModelTest:222-250` — and the new `ChallengeUtilsTest` asserts the *same eight cases with the same expected values* against the same arithmetic. So the PR leaves behind two dead delegating methods, a duplicated test suite, and an FQN workaround that exists solely to work around the dead methods. Deleting `DashboardViewModel.calculateIndividualProgress`/`calculateCommunityProgress` and the now-redundant test block removes all three at once, and the import collision disappears with them.

One thing the extraction makes newly visible, worth deciding on while you're here: the two functions are behaviorally identical. Both compute `minOf(count, 5) * 2 + (0 or 1)`, whose maximum is 11, so `coerceAtMost(500)` in `calculateIndividualProgress` can never bind. Your own tests demonstrate it — every expected value matches between the two suites, including `assertEquals(11, calculateIndividualProgress(10, false))`, which is capped by the arithmetic rather than by 500. Either the individual cap is meant to be reachable and the formula is wrong, or these should collapse into one function. Preserving behavior verbatim in an extraction is defensible, so this isn't what I'd block on — but side by side in one file it's now obvious enough to resolve.

**Verdict:** needs another round — delete the dead ViewModel wrappers and their duplicated tests, which also lets the FQNs become ordinary imports.

---

dogi:
**Re-review — Rating: 82/100** (was 72)

**Verdict:** mergeable — the extraction is right and the real win is `ChallengePrompter` no longer needing a ViewModel. Two loose ends and one genuinely odd thing the extraction makes visible.

Dropping `DashboardViewModel` from `ChallengePrompter`'s constructor is the payoff here: a class that formats a markdown dialog was holding a ViewModel reference solely to do two lines of arithmetic. Pure top-level functions in `ChallengeUtils.kt` are the right home, and they are now trivially testable without a ViewModel harness.

## The odd thing

**The two functions compute identical results for every input, and both caps are unreachable.**

```kotlin
minOf(voiceCount, 5) * 2        // maxes at 10
+ (if (!hasUnfinishedSurvey) 1 else 0)   // maxes at 11
```

So `calculateIndividualProgress` ends with `.coerceAtMost(500)` on a value that cannot exceed 11, and `calculateCommunityProgress` ends with `.coerceAtMost(11)` on a value that cannot exceed 11. Neither cap can ever fire, and the two functions are the same function with different dead ceilings.

The new test suite demonstrates this without remarking on it — `ChallengeUtilsTest` asserts the same eight values for both:

```
(0,false)→1   (0,true)→0   (1,false)→3   (1,true)→2
(5,false)→11  (5,true)→10  (10,false)→11 (10,true)→10
```

That is worth resolving rather than preserving. Either the caps encode an intent the arithmetic doesn't implement — a 500 ceiling strongly suggests individual earnings were meant to scale past five voices, in which case `minOf(voiceCount, 5)` is the bug — or both caps are vestigial and the two functions should collapse into one. Extracting them verbatim was the correct move for a refactor PR, but now that they sit side by side in one file the duplication is impossible to miss, and somebody should decide which reading is right.

## Two loose ends

**The `DashboardViewModel` wrappers now have no production callers.** `ChallengePrompter` was the only one, and it calls the utils directly after this PR. What remains is two thin delegating methods and a `DashboardViewModelTest` block asserting exactly the same eight values as `ChallengeUtilsTest`. Deleting the wrappers and that test block finishes the extraction; leaving them means two entry points to the same arithmetic and a duplicated suite that will drift.

**Import placement:** in `DashboardViewModel`, the two aliased imports landed above `org.ole.planet.myplanet.utils.DispatcherProvider`. Within the same package prefix, `DispatcherProvider` sorts before `calculateCommunityProgress`, so they belong below it. The aliases themselves are necessary — without them the member functions would recurse — and naming them `…Util` is clear enough, though the need for aliases disappears entirely if the wrappers go.

CI green 5/5.

---

Okuro3499 (changes requested):
```
1. app/src/main/java/org/ole/planet/myplanet/ui/dashboard/DashboardViewModel.kt:157-163 — The calculateIndividualProgress/calculateCommunityProgress wrapper methods have no       
     remaining production caller (only a test). Fix: delete both methods (and the now-unused aliased imports at lines 46-47) to actually complete the extraction.                    
  2. app/src/test/java/org/ole/planet/myplanet/ui/dashboard/DashboardViewModelTest.kt:220-251 — Remove these two test functions (they duplicate ChallengeUtilsTest and would         
     otherwise reference the now-deleted ViewModel methods once fix #1 lands).
```

dogi:

**Verdict:** `ready`. The dead-wrapper finding is closed, and closed more thoroughly than I asked. One question from the last round is still open, but it needs a product decision rather than a code change.

**The wrappers are gone and the dependency went with them.** I'd flagged that `DashboardViewModel.calculateIndividualProgress` / `calculateCommunityProgress` had no production callers left once `ChallengePrompter` could reach `ChallengeUtils` directly. Both are deleted — and `ChallengePrompter` no longer takes a `DashboardViewModel` at all:

```kotlin
class ChallengePrompter(
    private val activity: DashboardActivity,
    private val sharedPrefManager: SharedPrefManager
)
```

That's the better outcome. The prompter was holding a ViewModel solely to reach two pure functions; now it depends on neither. `DashboardActivity` drops the `dashboardViewModel` argument to match.

**No coverage was lost in the move.** The two `DashboardViewModelTest` cases are deleted, but `ChallengeUtilsTest` asserts the same eight boundary values per function against the new top-level functions. Same inputs, same expectations, now testing the code that actually runs.

## Still open from last round

The two functions remain byte-identical apart from the cap:

```kotlin
fun calculateIndividualProgress(voiceCount: Int, hasUnfinishedSurvey: Boolean): Int {
    …
    return total.coerceAtMost(500)   // ← unreachable: max possible total is 11
}
fun calculateCommunityProgress(allVoiceCount: Int, hasUnfinishedSurvey: Boolean): Int {
    …
    return total.coerceAtMost(11)    // ← exactly the ceiling, so also a no-op
}
```

`minOf(count, 5) * 2 + (0 or 1)` tops out at 11, so neither `coerceAtMost` can ever fire, and the two functions return the same value for every input. `ChallengeUtilsTest` now states this outright — both tests assert the identical eight pairs — without remarking on it.

That's fine to merge as-is, because it's a faithful extraction of what the code already did, and the extraction is what makes the oddity visible. But it's worth an issue: a 500 cap only makes sense if individual earnings were meant to scale past five voices, in which case `minOf(voiceCount, 5)` is the real bug and the community function is the one that's correct. Someone who knows the intended economics should say which.

CI green 5/5.

---

Okuro3499 (approved):

PR #17296, opened 2026-09-17
dashboard: smoother bell server reachability view modelling (fixes #17291)

### LINKED ISSUE #17291: BellDashboardViewModel statically imports MainApplication's reachability check instead of injecting the provider
## Verified premise

`ui/dashboard/BellDashboardViewModel.kt:15`:

```kotlin
import org.ole.planet.myplanet.MainApplication.Companion.isServerReachable
```

called bare at L137:

```kotlin
val reachable = isServerReachable(serverUrl)
```

That companion function is a pure delegate — `MainApplication.kt:189-190`:

```kotlin
suspend fun isServerReachable(urlString: String): Boolean =
    coreDependenciesEntryPoint.serverReachabilityProvider().isServerReachable(urlString)
```

So a ViewModel depends on the Application class to reach `utils/ServerReachabilityProvider`, a `@Singleton @Inject` class it could take in its constructor. This makes the reachability branch untestable without standing up `MainApplication`.

## Do

- Add `private val serverReachabilityProvider: ServerReachabilityProvider` to the `@HiltViewModel` constructor.
- Replace the L137 call with `serverReachabilityProvider.isServerReachable(serverUrl)`.
- Delete the static import at L15.

## Constraints

- **The dependency is a *static import of a companion member*, so the call at L137 reads as a bare `isServerReachable(serverUrl)` with no receiver. A grep for `MainApplication.` does not find it, and neither does a grep for `MainApplication` anywhere near line 137 — the import on line 15 is the only trace. Check that no other bare call to it survives in the file before deleting the import.**
- `isServerReachable` is `suspend`; the provider's method is too, so the call site needs no coroutine change.
- The provider is a `@Singleton` and caches internally — injecting it reaches the same instance the static reached, so behaviour is identical.
- Do not touch `MainApplication.isServerReachable`; other callers use it.

## Tests

- Existing `app/src/test/java/org/ole/planet/myplanet/ui/dashboard/BellDashboardViewModelTest.kt` must pass with **only the constructor call updated** to pass a mocked provider — no assertion changes.
- Add a test stubbing the provider to return `false` and asserting the offline branch at L137, which was not reachable in a unit test before.

## Out of scope

- `ChatRepositoryImpl`'s separate use of the same static (tracked separately).
- The rest of `BellDashboardViewModel`'s state handling.
- `ServerReachabilityProvider` itself.

dogi:
Injected `ServerReachabilityProvider` into `BellDashboardViewModel` constructor and replaced the static import call `isServerReachable(serverUrl)` with `serverReachabilityProvider.isServerReachable(serverUrl)`. Updated `BellDashboardViewModelTest` to inject the mock provider and added test cases covering both reachable (`true`) and unreachable (`false`) server scenarios.

Fixes #17291

---
*PR created automatically by Jules for task [12168925073697937189](https://jules.google.com/task/12168925073697937189) started by @dogi*


dogi:
**Review — rating: 90/100**

Behaviour-preserving and a clear testability win. `MainApplication.Companion.isServerReachable` is a one-line delegate:

```kotlin
suspend fun isServerReachable(urlString: String): Boolean =
    coreDependenciesEntryPoint.serverReachabilityProvider().isServerReachable(urlString)
```

so injecting `ServerReachabilityProvider` directly reaches the identical `@Singleton` instance with the same 30s reachability cache, alternative-URL fallback and HEAD→GET retry — it just drops the static hop through the Application object. Nothing about the runtime behaviour changes; the ViewModel simply stops depending on a framework global.

The payoff is visible in the tests: both new cases stub an ordinary injected mock instead of needing `mockkObject(MainApplication)`, which is why they could be written at all. They also assert the *side effect* — that `networkStatus` lands on `Connected` / `Disconnected` — not just the return value, which is the half that's easy to leave uncovered.

Leaving the remaining static callers (`ChatDetailFragment`, `SyncActivity`, `MainApplication` itself) alone is the right scope for a small PR; they can follow the same path individually.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #17300, opened 2026-09-17
Segregate UserAchievementsRepository interface from UserRepository

### LINKED ISSUE #17285: give AchievementUploader a narrow achievement interface instead of the whole 45-member UserRepository
## Verified premise

`repository/UserRepository.kt` declares 45 functions plus properties. `services/upload/AchievementUploader.kt` injects the whole thing (L20) and uses exactly two members:

```kotlin
class AchievementUploader @Inject constructor(
    ...
) {
    ...
    val list = userRepository.getAchievementsForUpload()      // L26
    ...
    userRepository.markAchievementUploaded(id, rev)           // L36
```

The achievement surface on `UserRepository` is a contiguous, self-contained block: `achievementUpdates: Flow<Unit>` (L28) and `initializeAchievement` / `updateAchievement` / `getAchievementData` / `getAchievementsForUpload` / `markAchievementUploaded` (L98-117).

The interface-segregation pattern already exists in this codebase — `repository/TeamsRepository.kt:43`:

```kotlin
interface TeamsRepository : TeamsFinancesRepository, TeamsMembersRepository, TeamsNotificationsRepository {
```

bound in `di/RepositoryModule.kt:192`:

```kotlin
@Binds
abstract fun bindTeamsFinancesRepository(impl: TeamsRepositoryImpl): TeamsFinancesRepository
```

## Do

- Create `repository/UserAchievementsRepository.kt` holding the achievement members currently on `UserRepository` (the `Flow` property and the five functions at L98-117).
- Change `UserRepository` to extend it and delete those declarations from its own body.
- Add one `@Binds` to `di/RepositoryModule.kt`, mirroring L192 exactly: `abstract fun bindUserAchievementsRepository(impl: UserRepositoryImpl): UserAchievementsRepository`.
- Narrow `AchievementUploader`'s constructor parameter type to `UserAchievementsRepository` and fix its import.

## Constraints

- **`UserRepositoryImpl` needs no edit at all: it already implements every one of those members as `override`, and they stay `override` because `UserRepository` still extends the new parent. If the build asks you to touch `UserRepositoryImpl`, the split is wrong — re-check that the moved declarations are byte-identical (default parameter values included) rather than changing the impl to match.**
- Move the declarations verbatim; do not rename, reorder parameters, or drop a default value.
- Move the `Achievement` / `AchievementData` / `JsonArray` / `JsonObject` imports the moved members need into the new file, and remove any that `UserRepository.kt` no longer uses.
- One `@Binds` in the existing `RepositoryModule` — do not create a new Hilt module for this; the `TeamsFinancesRepository` precedent lives in `RepositoryModule` and this should match it.
- Every other injection site of `UserRepository` stays on `UserRepository` — only `AchievementUploader` narrows.

## Tests

- Existing `app/src/test/java/org/ole/planet/myplanet/repository/UserRepositoryImplTest.kt` must pass **unedited** — it exercises the impl, whose members do not change.
- Any `AchievementUploader` test needs only its mock's declared type narrowed; its stubs and verifications stay as they are.
- `./gradlew testDefaultDebugUnitTest` and `./gradlew assembleDefaultDebug` (the second is what proves Hilt resolves the new binding — a missing `@Binds` compiles fine at the Kotlin level and fails at KSP/Dagger).

## Out of scope

- `UserRepositoryImpl`'s bodies and its other 40 members.
- Splitting any other slice off `UserRepository`.
- The achievement upload's network behaviour.

dogi:
Segregates achievement-related members from UserRepository into a narrower UserAchievementsRepository interface and updates AchievementUploader and Hilt dependency binding accordingly.

Fixes #17285

---
*PR created automatically by Jules for task [7391773019480342324](https://jules.google.com/task/7391773019480342324) started by @dogi*


dogi:
**Review — rating: 86/100**

Clean interface segregation. `AchievementUploader` needs exactly two methods and was injecting a ~40-method `UserRepository` to get them; the new interface is scoped to what it actually uses. Having `UserRepository` extend `UserAchievementsRepository` is the pragmatic choice — every existing consumer keeps compiling, which is why this lands as a small diff, and consumers can be migrated to the narrower type one at a time.

The thing I wanted to rule out is whether the new `@Binds` produces a *second* `UserRepositoryImpl`. It does: the impl carries no `@Singleton` of its own (the scope lives on the `@Binds` methods), so Dagger caches one instance per binding key rather than one overall. That's currently harmless, and I checked why rather than assuming:

- `achievementUpdates` is `realtimeSyncManager.dataUpdateFlow`, and `RealtimeSyncManager` *is* `@Singleton`, so both instances hand out the same flow — an observer in `AchievementViewModel` still sees updates triggered through the other binding.
- `UserRepositoryImpl` holds no mutable instance state (no `private var`, no `MutableStateFlow`/`MutableSharedFlow`, no backing `_` fields), so there's nothing to diverge.

Worth saying clearly: this PR didn't introduce that. `UserRepositoryImpl` was already bound twice on master (`UserRepository` and `UserSyncRepository`), so this makes it three instances of an existing pattern rather than starting a new one. But the fix is one line — annotate `UserRepositoryImpl` with `@Singleton` and all three bindings collapse onto a single cached instance — and it's worth doing before someone adds a cache or a `MutableSharedFlow` to that class, because the resulting bug (updates published on one instance, observed on another) is a genuinely confusing one to track down. Here or as a follow-up, either is fine.

No tests, which I think is right: this is interface plumbing with no behaviour change, and the achievement methods keep their existing coverage through `UserRepositoryImpl`.

**Verdict:** mergeable.

---

Okuro3499 (approved):

Okuro3499 (dismissed):
```
File: app/src/main/java/org/ole/planet/myplanet/di/RepositoryModule.kt (lines ~211-218) and repository/UserRepositoryImpl.kt (class declaration, line 64)                        
    Problem: UserRepositoryImpl has an @Inject constructor with no @Singleton on the class itself. The module now has two separate @Binds @Singleton methods pointing at the same    
    impl class (bindUserAchievementsRepository and bindUserRepository). Because the scope lives on the @Binds edge rather than on the impl type, Dagger/Hilt will memoize each       
    interface binding independently — this produces two separate singleton instances of UserRepositoryImpl in the graph: one served to everything that injects UserRepository, and a 
    distinct one served to AchievementUploader (which now injects UserAchievementsRepository). Today this is harmless only because UserRepositoryImpl carries no mutable instance    
    state (everything delegates to already-singleton deps like RealtimeSyncManager, DAOs, ApiInterface) — achievementUpdates in particular is safe because it's derived from         
    realtimeSyncManager.dataUpdateFlow, a genuinely shared singleton. But nothing enforces that invariant going forward; the next person who adds a cache field, an in-flight-request
    guard, or any per-instance state to UserRepositoryImpl will silently get divergent state between the "UserRepository view" and the "UserAchievementsRepository view" of what's   
    supposed to be the same object. 
```

dogi:

**Verdict:** mergeable — and this fixes a latent bug that predates the PR, which is worth knowing because the description doesn't claim it.

**The `@Singleton` on `UserRepositoryImpl` is the load-bearing change.** With `@Binds`, the scope annotation applies to the *binding*, not to the implementation — so an unscoped `@Inject`-constructed class bound twice with `@Singleton` yields **two separate instances**, one per binding. That is the classic Dagger trap, and this module was already sitting in it: on master, `RepositoryModule` binds `UserRepositoryImpl` twice —

```kotlin
@Binds @Singleton abstract fun bindUserRepository(impl: UserRepositoryImpl): UserRepository
@Binds @Singleton abstract fun bindUserSyncRepository(impl: UserRepositoryImpl): UserSyncRepository
```

— and `UserRepositoryImpl` carries no scope of its own. So `UserRepository` and `UserSyncRepository` have been resolving to **different objects** all along. Adding `@Singleton` to the class closes that, and it had to be added here because this PR introduces a third binding that would otherwise have made it a three-way split.

That matters more than a wasted allocation: `UserRepositoryImpl` exposes `achievementUpdates: Flow<Unit>`, so duplicate instances mean a writer emitting on one flow while a collector listens to another. `AchievementUploader` is now on the new binding, so without the scope fix it would have been watching a flow nothing published to.

**The segregation itself is done the pragmatic way.** `UserRepository : UserAchievementsRepository` means existing `UserRepository` consumers are untouched and still see the achievement methods — so this is segregation for *new* consumers rather than a breaking narrowing. `AchievementUploader` is the first beneficiary and now declares only what it uses. That's the right increment; a hard split would have churned every call site for no behavioural gain.

CI green 5/5.

**One thing worth a follow-up, not this PR:** now that the scope trap is visible, `RepositoryModule` is worth a scan for other implementations bound more than once without a class-level scope. The same pattern would be silently duplicating any of them, and the symptom — shared state that mysteriously doesn't propagate — is miserable to debug from the outside.

---

Okuro3499 (approved):

PR #17303, opened 2026-09-17
feedback: smoother repository owner manager querying (fixes #17283)

### LINKED ISSUE #17283: tighten FeedbackRepository: hide the two internal-only builders and stop passing a whole UserEntity into the list query
## Verified premise

`repository/FeedbackRepository.kt` exposes three members that do not need to be on the contract.

**Two builders whose only callers are each other.** `createFeedback` (L8) and `saveFeedback` (L29) are public interface methods, and the only call sites in `app/src/main` are inside `createAndSaveFeedback` in the impl — `FeedbackRepositoryImpl.kt:32-33`:

```kotlin
val feedback = createFeedback(user, urgent, type, message, item, state)
saveFeedback(feedback)
```

**A suspend function that never suspends, taking more than it reads.** L24:

```kotlin
suspend fun getFeedback(userModel: UserEntity?): Flow<List<Feedback>>
```

The impl body (L73) only picks a DAO `Flow` based on two things it reads off the entity, and the single app caller is `ui/feedback/FeedbackListViewModel.kt:36`.

## Do

- Change the signature to `fun getFeedback(ownerName: String?, isManager: Boolean): Flow<List<Feedback>>` — drop `suspend`, take the two values instead of the entity.
- Update `FeedbackListViewModel.loadFeedback` to pass the two fields off the user it already loads.
- Reduce the visibility of `createFeedback` and `saveFeedback`: remove them from the interface.

## Constraints

- **`app/src/test/java/org/ole/planet/myplanet/repository/FeedbackRepositoryImplTest.kt:42` calls `repository.createFeedback("user1", "High", "Bug", "Test message", null, null)` on a `FeedbackRepositoryImpl` (declared L29), so making it `private` breaks `createFeedback_usesTimeProviderNow` — which is the only thing pinning `timeProvider.now()` into `openTime` and the message timestamp. Drop it from the *interface* but leave the impl member callable (no `private`), or retarget that test through `createAndSaveFeedback` while keeping both assertions. Do not delete the test.**
- Read the impl body at L73 to get the two field names right — do not assume `name` / `isManager` without checking what it actually reads off `UserEntity`.
- `getFeedback` keeps returning the same DAO `Flow` for the same inputs; this is a signature change, not a query change.
- Leave `createAndSaveFeedback` public and its behaviour identical.

## Tests

- `app/src/test/java/org/ole/planet/myplanet/ui/feedback/FeedbackListViewModelTest.kt` stubs `getFeedback` with `coEvery` at L42/64/87/98 and verifies call counts at L75 (`coVerify(exactly = 1)`) and L106 (`coVerify(exactly = 2)`). Dropping `suspend` means those move to `every` / `verify` — **the exact counts must be preserved**, they pin that a refresh restarts the collector exactly once.
- `FeedbackRepositoryImplTest` L138 and L154 collect `repository.getFeedback(user)`; update the arguments only.

## Out of scope

- `getPendingFeedback`, `getFeedbackById`, `closeFeedback`, `addReply`, `markFeedbackUploaded`.
- The `RealtimeSyncMixin` wiring in `FeedbackListFragment` (tracked separately).
- The feedback DAO's queries.

dogi:
Refactor FeedbackRepository to remove internal builder methods createFeedback and saveFeedback from interface contract. Simplify getFeedback signature to take ownerName and isManager parameters instead of taking a UserEntity and remove unnecessary suspend modifier.

Fixes #17283

---
*PR created automatically by Jules for task [3914116844870642668](https://jules.google.com/task/3914116844870642668) started by @dogi*


dogi:
**Review — rating: 83/100**

Three separate changes, all sound. I checked the two interface removals rather than trusting them:

- `saveFeedback` has exactly one caller, `FeedbackRepositoryImpl:33` inside `createAndSaveFeedback`, and no test touches it — so `private` is exactly right.
- `createFeedback` is called from `createAndSaveFeedback` plus one test (`FeedbackRepositoryImplTest:42`), which is why it stays public on the impl rather than private. Same shape as #17278, so at least it's consistent.

Dropping `suspend` from `getFeedback` is correct — it returns a `Flow` and the body only selects a DAO flow and wraps it in `distinctByContent`, with nothing to suspend on. That matches the convention the style guide states for `Flow`-returning methods.

**The signature change is the one real trade-off.** Taking `(ownerName, isManager)` decouples the repository from `UserEntity`, which is good. But it also moves the *policy* — "managers see everything, everyone else sees their own" — out of the repository and onto the caller, who now has to remember `user?.isManager() == true`. With `FeedbackListViewModel` as the only caller today the risk is nil, but a second caller that gets that argument wrong fails silently by showing owner-scoped feedback to a manager, where previously the rule lived in exactly one place. Worth a conscious decision rather than a side effect; if you keep it, the naming is clear enough that I wouldn't block on it.

**Nit:** the decoupling isn't actually cashed in by the tests. `FeedbackRepositoryImplTest` still builds a `UserEntity` mock and then calls `repository.getFeedback(user.name, user.isManager())` — stubbing two accessors purely to feed primitives back in. `repository.getFeedback("managerName", true)` drops the mock entirely and is the whole point of the new signature.

Also, the added `clearMocks(feedbackRepository, answers = false)` in two ViewModel tests is there because `setUp` already constructs a `FeedbackListViewModel` whose init consumes the stub, so the `verify(exactly = …)` counts would otherwise be off by one. It works, but it's a sign `setUp` is doing too much — not creating the ViewModel there would remove the need.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #17304, opened 2026-09-17
life: smoother health repository examination uploading (fixes #17284)

### LINKED ISSUE #17284: move the health-examination upload sequence out of UploadToShelfService into HealthRepository
## Verified premise

`services/UploadToShelfService.kt` hand-rolls the same three-step sequence twice, differing only in the fetch:

```kotlin
fun uploadHealth() {
    appScope.launch(dispatcherProvider.io) {
        val myHealths = healthRepository.getUpdatedHealthExaminations()
        val uploadedHealths = healthRepository.uploadHealthData(myHealths)
        healthRepository.markHealthExaminationsUploaded(uploadedHealths)
    }
}

fun uploadSingleUserHealth(userId: String?, listener: OnSuccessListener?) {
    appScope.launch(dispatcherProvider.io) {
        try {
            if (userId.isNullOrEmpty()) return@launch
            val myHealths = healthRepository.getUpdatedHealthForUser(userId)
            val uploadedHealths = healthRepository.uploadHealthData(myHealths)
            healthRepository.markHealthExaminationsUploaded(uploadedHealths)
            withContext(dispatcherProvider.main) { listener?.onSuccess("Health data for user $userId uploaded successfully") }
        } catch (e: Throwable) {
            withContext(dispatcherProvider.main) { listener?.onSuccess("Error uploading health data for user $userId: ${e.localizedMessage}") }
        }
    }
}
```

Three members stay public on `HealthRepository` purely so a service can orchestrate them — `getUpdatedHealthExaminations` (L15), `getUpdatedHealthForUser` (L16), `markHealthExaminationsUploaded` (L17), plus `uploadHealthData` (L20).

## Do

- Add `suspend fun syncPendingHealthExaminations()` and `suspend fun syncPendingHealthExaminationsForUser(userId: String)` to `HealthRepository` / `HealthRepositoryImpl`, each running fetch → `uploadHealthData` → `markHealthExaminationsUploaded`.
- Reduce each service body to one repository call, keeping its `appScope.launch(dispatcherProvider.io)`, its try/catch and its listener callbacks exactly as they are.
- Remove `getUpdatedHealthExaminations`, `getUpdatedHealthForUser` and `markHealthExaminationsUploaded` from the interface once nothing outside the impl calls them — **check first**, this interface has 30+ members and other screens may use them.

## Constraints

- **`uploadHealth()` and `uploadSingleUserHealth(userId, listener)` must keep their exact public signatures: `services/AutoSyncWorker.kt:113` and `services/UserDataWorker.kt:54` call the first, and `ui/sync/ProcessUserDataActivity.kt:177` calls the second with an anonymous `OnSuccessListener`. Changing either signature breaks three callers outside this file.**
- The `if (userId.isNullOrEmpty()) return@launch` guard stays in the *service*, before the repository call — it short-circuits without invoking the listener, and that is observable behaviour.
- Both listener messages, including the error one that also calls `onSuccess`, stay byte-identical (`"Health data for user $userId uploaded successfully"` / `"Error uploading health data for user $userId: ${e.localizedMessage}"`).
- The `withContext(dispatcherProvider.main)` around each listener call stays in the service.
- Do not remove a method from the interface without grepping `app/src/main` and `app/src/test` for it first.

## Tests

- Existing `app/src/test/java/org/ole/planet/myplanet/repository/HealthRepositoryImplTest.kt` and any `UploadToShelfService` test must pass **unedited**.
- Add repository tests for both new methods: the fetch feeds `uploadHealthData`, and its result feeds `markHealthExaminationsUploaded`.

## Out of scope

- `uploadHealthData`'s own network behaviour.
- The other upload paths in `UploadToShelfService` (user data, shelf).
- `HealthRepository`'s patient-search and profile members.

dogi:
Encapsulated the health examination upload sequence inside HealthRepository and HealthRepositoryImpl, added syncPendingHealthExaminations and syncPendingHealthExaminationsForUser, removed unneeded members from the HealthRepository contract, simplified UploadToShelfService calls, and updated tests.

Fixes #17284

---
*PR created automatically by Jules for task [10524848042675678497](https://jules.google.com/task/10524848042675678497) started by @dogi*

## Summary by CodeRabbit

* **Improvements**
  * Health examination uploads now use a consolidated synchronization flow for all users or a specific user.
  * Successful uploads continue to be recorded, while existing handling for invalid user IDs and synchronization errors is preserved.

* **Tests**
  * Added coverage for health examination synchronization and updated upload service tests for the streamlined workflow.


dogi:
**Review — rating: 80/100**

Good "tell, don't ask" refactor. The fetch → upload → mark sequence was spelled out twice in `UploadToShelfService`, which meant the service knew the repository's internal protocol and could get the ordering wrong; now each variant is one call. The composed methods are a faithful transcription of the old sequence, and they still run inside the service's existing `appScope.launch(dispatcherProvider.io)`, so the threading is unchanged. The service tests get materially better as a result — no more mocking intermediate `healthRecords`/`uploadedRecords` values just to thread them between three stubs.

The new repository tests are the right kind too: they drive the whole sequence through the mocked DAO and API and assert `markUploaded(mapOf("exam1" to "rev1"))`, so they verify the id/rev actually flows from the API response into the DAO rather than just that three methods were called.

**The segregation is left half-done.** This PR takes `getUpdatedHealthExaminations`, `getUpdatedHealthForUser` and `markHealthExaminationsUploaded` off the interface, but leaves `uploadHealthData` on it — and after this change its only production callers (`UploadToShelfService:73` and `:84`) are gone. It's now in exactly the same category as the three that were demoted: called internally by the two new sync methods, plus directly from `HealthRepositoryImplTest`. It should come off the interface with them, or the argument for demoting the other three doesn't hold either.

**The static mocks can leak.** Both new tests end with:

```kotlin
unmockkObject(org.ole.planet.myplanet.utils.UrlUtils)
io.mockk.unmockkStatic(android.text.TextUtils::class)
```

in the test body rather than in `@After` or a `finally`. If an assertion fails or the code under test throws, those never run, and `UrlUtils` and `TextUtils` stay mocked for every subsequent test in the class — so one real failure turns into a cascade of unrelated ones that are painful to diagnose. Moving both into `@After` (the class already has the lifecycle hooks) makes the cleanup unconditional.

**Nit:** the new tests carry about eight inline fully-qualified names — `org.ole.planet.myplanet.utils.UrlUtils`, `android.text.TextUtils`, `retrofit2.Response`, `io.mockk.unmockkStatic` — which the style guide lists under *Things to Avoid*; these all import cleanly. And `syncPendingHealthExaminations`/`syncPendingHealthExaminationsForUser` differ only in which DAO call they make, though I think the explicit names read better than one nullable-`userId` method, so I'd leave that as is.

**Verdict:** mergeable — please move the unmock calls into `@After` before merging.

---

Okuro3499 (approved):

PR #17308, opened 2026-09-17
all: smoother notifications repository time providing (fixes #17289)

### LINKED ISSUE #17289: hide the internal-only getNotifications and route the five raw Date() constructions through TimeProvider
## Verified premise

Two small things in the notifications repository.

**A contract member with no external caller.** `repository/NotificationsRepository.kt:22`:

```kotlin
suspend fun getNotifications(userId: String, filter: String, isAdmin: Boolean = false): List<NotificationPayload>
```

Its only call site in `app/src/main` is `NotificationsRepositoryImpl.kt:166`, inside `getEnrichedNotifications`:

```kotlin
val payloadNotifications = getNotifications(userId, filter, isAdmin)
```

**Five wall-clock reads bypassing the injected clock.** The impl injects `timeProvider` and still constructs `java.util.Date()` directly at L107, L115, L128, L136 and L461:

```kotlin
this.createdAt = Date()                                              // L107
this.createdAt = Date()                                              // L115
notificationDao.markAsRead(existingIds, Date())                      // L128
notificationDao.markAllUnreadAsRead(actualUserId, Date())            // L136
createdAt = doc.get("time")?.let { Date(it.asLong) } ?: Date()       // L461
```

so five code paths are untestable against a fixed clock in a class that already has one injected.

## Do

- Delete the `getNotifications` declaration from `NotificationsRepository.kt` and drop `override` from the impl member (it stays a callable member of the class, exactly as `getEnrichedNotifications` needs).
- Replace each of the five `Date()` calls with `Date(timeProvider.now())`. At L461, only the `?: Date()` fallback changes — `Date(it.asLong)` is parsing a document timestamp and stays.

## Constraints

- **`NotificationDao` has its own method with the same name — `data/room/dao/NotificationDao.kt:37`, `suspend fun getNotifications(userId: String, filter: String, isAdmin: Boolean)` — and `NotificationsRepositoryImpl.kt:145` calls it (`notificationDao.getNotifications(...)`). Do not touch the DAO method, and do not let a search-and-replace on the name reach it.**
- Once the impl member is no longer an `override`, it may carry the `isAdmin: Boolean = false` default that the override could not. Adding it is optional; if you do, the L166 call site still passes all three arguments explicitly and must keep doing so.
- `NotificationsRepositoryImpl.kt` L394-406 is being changed by an open PR (the per-team task badge fix). Stay out of `getTeamNotifications`.
- Do not change what any of the five timestamps mean — this is the same instant, read through the injected provider.

## Tests

- `app/src/test/java/org/ole/planet/myplanet/repository/NotificationsRepositoryImplTest.kt` stubs the **DAO** method at L528, L577 and L613 (`coEvery { notificationDao.getNotifications(...) }`) — those stubs must stay exactly as they are, since the DAO is untouched.
- Add a test pinning one of the `markAsRead` paths to a fixed `timeProvider.now()`.

## Out of scope

- `getTeamNotifications` (open PR).
- `getEnrichedNotifications`'s enrichment logic.
- The `Calendar.getInstance()` usage elsewhere in the file.

dogi:
Removed internal-only `getNotifications` function declaration from `NotificationsRepository` interface, removed the `override` keyword on `NotificationsRepositoryImpl.getNotifications` while keeping default parameter values, replaced 5 raw `Date()` constructions with `Date(timeProvider.now())`, and updated unit test assertions.

Fixes #17289

---
*PR created automatically by Jules for task [18208120231517890317](https://jules.google.com/task/18208120231517890317) started by @dogi*


dogi:
**Review — rating: 82/100**

The `TimeProvider` half is complete and worth having. `NotificationsRepositoryImpl` already took `timeProvider` as a constructor dependency and already used it at line 394, so the class was half-deterministic and half-wall-clock; this makes it uniform. I checked coverage rather than assuming — master has exactly five raw `Date()` call sites (107, 115, 128, 136, 461) and the diff converts all five, so nothing is left behind.

Replacing `assertTrue(saved.createdAt.after(initialDate))` with an exact equality is the right direction too — the old form is the kind of assertion that passes for the wrong reason.

**Two things about how that assertion is written, though.** `TestTimeProvider`'s constructor defaults to `currentTime = 0L`, and the repository under test is built with a bare `TestTimeProvider()` at line 52. So:

- `assertEquals(TestTimeProvider().now(), saved.createdAt.time)` compares against a *throwaway* provider rather than the one the repository holds. It only passes because both default to `0L`; the moment someone sets a realistic time at line 52 for any reason, all three assertions fail in a way that looks unrelated to the change that broke them. Keeping the provider in a field and asserting `timeProvider.now()` makes the coupling explicit and survives that edit.
- Because that value is `0L`, `Date(timeProvider.now())` is `Date(0)` — so these tests can't distinguish "routed through `TimeProvider`" from "hardcoded to the epoch", which is precisely the property the PR is establishing. Giving the provider a non-zero fixed time would turn them into real assertions.

**On the interface change:** I traced the repository's `getNotifications` and it has no caller outside the class at all — `NotificationsRepositoryImpl:166` inside `getEnrichedNotifications` is the only one, and the test hits at 528/577/613 are `notificationDao.getNotifications(...)`, the DAO method, not this one. So unlike the recent demotions that stayed public to keep direct tests compiling, this one can just be `private`, the way `saveFeedback` was handled in #17303. Leaving it public keeps it reachable for no reason.

**Nit:** `slot<java.util.Date>()` uses an inline fully-qualified name where an import would do.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #17309, opened 2026-09-17
chat: smoother repository server reachability providing (fixes #17290)

### LINKED ISSUE #17290: ChatRepositoryImpl reaches through MainApplication for a reachability check it could inject
## Verified premise

`repository/ChatRepositoryImpl.kt:38-42` calls an Android Application static from inside a repository:

```kotlin
@VisibleForTesting
internal var reachabilityCheck: suspend (String) -> Boolean = { url ->
    org.ole.planet.myplanet.MainApplication.isServerReachable(url)
}
```

That static is a pure delegate — `MainApplication.kt:189-190`:

```kotlin
suspend fun isServerReachable(urlString: String): Boolean =
    coreDependenciesEntryPoint.serverReachabilityProvider().isServerReachable(urlString)
```

so the repository routes through the Application class and a Hilt entry point to reach `utils/ServerReachabilityProvider`, an ordinary `@Singleton @Inject` class it could simply take in its constructor. The class already injects six dependencies (`chatDao`, `chatApiService`, `serverUrlMapper`, `sharedPrefManager`, `dispatcherProvider`, `gson`).

## Do

- Add `private val serverReachabilityProvider: ServerReachabilityProvider` to the constructor.
- Change the lambda's default body to `serverReachabilityProvider.isServerReachable(url)`.
- Remove the fully-qualified `MainApplication` reference.

## Constraints

- **Keep the `@VisibleForTesting internal var reachabilityCheck` seam exactly as it is — same name, same `suspend (String) -> Boolean` type, same mutability. Tests assign to it to simulate an unreachable server; replacing the lambda with a direct call to the provider removes that seam and breaks them.** Only the lambda's *default body* changes.
- `ServerReachabilityProvider.isServerReachable` is `suspend` and so is the lambda — no dispatcher change is needed, and the provider already does its own `withContext`.
- The provider caches results internally (`reachabilityCache`); calling it directly rather than through `MainApplication` reaches the same singleton instance, so caching behaviour is unchanged.
- Do not touch `MainApplication.isServerReachable` itself — other callers use it.

## Tests

- Existing `app/src/test/java/org/ole/planet/myplanet/repository/ChatRepositoryImplTest.kt` must pass with **only the constructor call updated** to pass a mocked `ServerReachabilityProvider`; every `reachabilityCheck = { ... }` assignment and every assertion stays untouched.

## Out of scope

- `BellDashboardViewModel`'s separate use of the same static (tracked separately).
- `ServerReachabilityProvider`'s own caching and HTTP behaviour.
- The chat request/response paths in the same file.

dogi:
Refactored ChatRepositoryImpl to inject ServerReachabilityProvider directly instead of accessing it statically through MainApplication.isServerReachable. Updated ChatRepositoryImplTest and ChatRepositoryTest.

Fixes #17290

---
*PR created automatically by Jules for task [6434712015997011323](https://jules.google.com/task/6434712015997011323) started by @dogi*


dogi:
**Review — rating: 88/100**

Same substitution as #17296 and correct for the same reason: `MainApplication.isServerReachable` is a one-line delegate to `coreDependenciesEntryPoint.serverReachabilityProvider().isServerReachable(...)`, so the injected provider is the identical `@Singleton` with the same cache and fallback behaviour — no runtime change, one fewer static dependency. This one also deletes an inline fully-qualified `org.ole.planet.myplanet.MainApplication.isServerReachable(url)`, which the style guide lists under *Things to Avoid*, so it's a style fix as well as a DI one.

I checked the property-initialiser ordering, since `reachabilityCheck` closes over a constructor parameter: `serverReachabilityProvider` is a constructor property and the lambda only dereferences it when invoked at line 116, so there's no initialisation-order hazard.

**Nit:** the injection makes the test seam vestigial. `@VisibleForTesting internal var reachabilityCheck` existed because a static call couldn't be mocked; now that the provider is a constructor dependency, `ChatRepositoryImplTest:65`'s `chatRepository.reachabilityCheck = { true }` could become `coEvery { serverReachabilityProvider.isServerReachable(any()) } returns true`, and the mutable var could come out of production code entirely. That's the payoff this change unlocks, and it's a small follow-up.

Worth noting the relaxed mock now returns `false` by default for any test that reaches line 116 without stubbing, which is a safer default than the previous static would have given in a unit test.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #17310, opened 2026-09-17
teams: smoother requests current user view modelling (fixes #17292)

### LINKED ISSUE #17292: RequestsFragment fetches the current user a second time for data its ViewModel already loaded
## Verified premise

`ui/teams/members/RequestsViewModel.kt:125-140` already loads the current user and throws the entity away, keeping only the id:

```kotlin
fun fetchMembers(teamId: String) {
    viewModelScope.launch {
        coroutineScope {
            val membersDeferred = async { teamsRepository.getRequestedMembers(teamId) }
            val memberCountDeferred = async { teamsRepository.getJoinedMemberCount(teamId) }
            val userDeferred = async { userRepository.getUserModel() }
            val user = userDeferred.await()
            val isLeader = teamsRepository.isTeamLeader(teamId, user?.id)
            _uiState.value = RequestsUiState(
                members = membersDeferred.await(),
                isLeader = isLeader,
                memberCount = memberCountDeferred.await()
            )
        }
    }
}
```

while `ui/teams/members/RequestsFragment.kt:40-48` injects a second source and fetches the same user again, in a coroutine that races the state collector:

```kotlin
viewModel.fetchMembers(teamId)
    currentUser = userSessionManager.getUserModel() ?: UserEntity()
    (adapter as? RequestsAdapter)?.setUser(currentUser)
}
collectWhenStarted(viewModel.uiState) { uiState ->
    (adapter as? RequestsAdapter)?.setData(uiState.members, uiState.isLeader, uiState.memberCount)
    ...
}
```

`UserSessionManager` is injected at L26 for this one use.

## Do

- Add `currentUser` to `RequestsUiState` **as the last property, with a default**, and populate it in `fetchMembers` from the `user` it already awaits.
- In the fragment, set the adapter's user from `uiState` inside the existing `collectWhenStarted(viewModel.uiState)` collector.
- Delete the standalone `lifecycleScope.launch` fetch and the `UserSessionManager` injection + import.

## Constraints

- **Call `setUser(...)` before `setData(...)` inside the collector. Today they run in two separate coroutines and `setData` can win the race, so the adapter can already render rows with no user set; folding both into one collector must fix that ordering, not preserve it.**
- `RequestsUiState` is constructed positionally at L132 and `.copy(...)`-ed in the optimistic update inside `respondToRequest` — appending the new property last with a default keeps both valid. Do not insert it in the middle.
- Preserve the `?: UserEntity()` fallback: the adapter's `setUser` takes a non-null `UserEntity`, while `userRepository.getUserModel()` returns `UserEntity?`. Decide where the empty-entity default lives (state default or call site) and keep the adapter receiving non-null.
- `respondToRequest`'s optimistic `copy` must keep carrying `currentUser` — `copy` does this for free, so do not rebuild the state with a positional constructor there.

## Tests

- Existing `app/src/test/java/org/ole/planet/myplanet/ui/teams/` requests tests must pass **unedited** — in particular anything asserting the optimistic member-count update in `respondToRequest`.
- Add a ViewModel test: `fetchMembers` puts the entity from `userRepository.getUserModel()` into `uiState.currentUser`, and a null user yields the empty-entity default.

## Out of scope

- `MembersUiState` and `loadJoinedMembers` (a separate state object in the same file).
- `RequestsAdapter`'s rendering.
- The other `userRepository.getUserModel()` calls in this ViewModel (L79, L94).

dogi:
This change optimizes user data loading in `RequestsFragment` by eliminating duplicate fetch calls:
1. Appends `currentUser: UserEntity = UserEntity()` to `RequestsUiState`.
2. Populates `currentUser` in `RequestsViewModel.fetchMembers` from the user entity already awaited.
3. Updates `RequestsFragment` to update the adapter's user state from `uiState.currentUser` before calling `setData` in `collectWhenStarted(viewModel.uiState)`.
4. Removes standalone coroutine user fetch, `UserSessionManager` injection, and unused imports from `RequestsFragment`.
5. Adds unit tests in `RequestsViewModelTest` verifying `uiState.currentUser` population and null fallback.

Fixes #17292

---
*PR created automatically by Jules for task [6645113955242414085](https://jules.google.com/task/6645113955242414085) started by @dogi*


dogi:
**Review — rating: 86/100**

This does more than remove a duplicate load — it fixes a real ordering race, which I don't think the title gives it credit for.

Before, `setUser` was called from a standalone `viewLifecycleOwner.lifecycleScope.launch` while `setData` came from a separate `collectWhenStarted(viewModel.uiState)` collector. Nothing ordered those two, so the adapter could be handed its member list before it had the current user, and which one won depended on whether `userSessionManager.getUserModel()` resolved faster than `fetchMembers`. Folding the user into `RequestsUiState` means both now arrive in one emission with `setUser` before `setData`, deterministically.

It's also genuinely deduplicating: `RequestsViewModel.fetchMembers` already resolved the user via `userRepository.getUserModel()` for its `isTeamLeader(teamId, user?.id)` call, so the fragment's separate `userSessionManager.getUserModel()` was a second read of the same thing — and from a *different* source, which meant the leader check and the displayed user could in principle disagree. Now they're the same object by construction.

Since removing the only assignment in `onViewCreated` could have left a `lateinit` unset, I checked: `onAttach` still does `currentUser = UserEntity()`, and the `by lazy` adapter reads it at construction time, so there's no `UninitializedPropertyAccessException` and the adapter still gets the same empty seed it got before. No regression.

**Nit:** that makes `currentUser` vestigial. Its only remaining job is seeding the adapter constructor with an empty `UserEntity()` — the real user arrives through `setUser` — so the field, the `lateinit`, and the `onAttach` override could all go away in favour of passing `UserEntity()` directly at the constructor call. Worth clearing out while the file is open.

The null-fallback test is a reasonable addition, though its name promises both branches and it only asserts the fallback; the "sets currentUser" half is covered by the assertion added to the test above it.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #17315, opened 2026-09-17
all: smoother update dialog utils checksum checking (fixes #17298)

### LINKED ISSUE #17298: stop threading ConfigurationsRepository into two DialogUtils helpers for a single checksum call
## Verified premise

`utils/DialogUtils.kt` — a UI utility — imports a repository at L21 and takes one as a parameter in two functions:

```kotlin
import org.ole.planet.myplanet.repository.ConfigurationsRepository    // L21
...
fun getUpdateDialog(                                                   // L188
    ...
    configurationsRepository: ConfigurationsRepository                 // L193
)
...
fun startDownloadUpdate(                                               // L208
    ...
    configurationsRepository: ConfigurationsRepository                 // L213
) {
    ...
    val checksumMatch = configurationsRepository.checkCheckSum(path)   // L216
```

`checkCheckSum(path)` is the only thing either function does with it. Both call sites already hold the repository:

- `ui/sync/SyncActivity.kt:763` — `getUpdateDialog(this@SyncActivity, info, customProgressDialog, lifecycleScope, configurationsRepository)`
- `services/AutoSyncWorker.kt:93` — `startDownloadUpdate(context, UrlUtils.getApkUpdateUrl(info?.localapkpath), null, workerScope, configurationsRepository)`

## Do

- Change both parameters to `checkCheckSum: suspend (String) -> Boolean`.
- Pass `configurationsRepository::checkCheckSum` at both call sites.
- Remove the `ConfigurationsRepository` import from `DialogUtils.kt`.

## Constraints

- **`getUpdateDialog` forwards its repository into `startDownloadUpdate` — check whether it does before changing either signature, and thread the function reference through in the same shape. Changing one signature and leaving the other taking a repository leaves the import in place and defeats the point.**
- Keep the parameter in the same position in both signatures so the positional call sites need only the argument swapped.
- `checkCheckSum` is `suspend`; the lambda type must be `suspend (String) -> Boolean`, and both call sites already run inside a coroutine scope (`lifecycleScope` / `workerScope`).
- Do not change `ConfigurationsRepository.checkCheckSum` itself.
- Leave `configurationsRepository` injected in `SyncActivity` and `AutoSyncWorker` — both use it for other things.

## Tests

- Existing tests covering the update dialog and `AutoSyncWorker` must pass with **only the argument form changed** — a test that stubbed the repository now passes a lambda, but its assertions stay.
- Add a `DialogUtils` test passing a lambda that returns `false` and asserting the checksum-mismatch branch, which previously needed a whole repository mock.

## Out of scope

- The other dialogs in `DialogUtils.kt`.
- The APK download and install flow.
- `UrlUtils.getApkUpdateUrl`.

dogi:
Refactors getUpdateDialog and startDownloadUpdate in DialogUtils.kt to accept checkCheckSum: suspend (String) -> Boolean instead of taking the entire ConfigurationsRepository. Updates call sites in SyncActivity.kt and AutoSyncWorker.kt, removes the unused ConfigurationsRepository import from DialogUtils.kt, and adds unit tests in DialogUtilsUpdateTest.kt.

Fixes #17298

---
*PR created automatically by Jules for task [10239278091358518674](https://jules.google.com/task/10239278091358518674) started by @dogi*


dogi:
**Review — rating: 89/100**

Nicely scoped. `DialogUtils` is a UI utility object that was taking an entire `ConfigurationsRepository` to call exactly one method on it; narrowing the parameter to `suspend (String) -> Boolean` removes the repository import from the utils layer altogether and depends on precisely the capability it uses. `configurationsRepository::checkCheckSum` is a bound method reference, so the call sites read cleanly and the invocation is identical — no behavioural change.

The payoff is the new test file, and it's the good kind: the refactor is what makes it writable. Both branches are covered, and rather than verifying a mock it asserts real behaviour through Robolectric's `nextStartedService` — that `DownloadService` is started when the checksum misses and that nothing is started when it matches — plus that the lambda received the right path. Testing this before would have meant standing up a whole `ConfigurationsRepository` mock for a one-boolean decision.

Worth noting this is the same layering problem #17288 has, solved the right way round: there a repository reached up into `ui.settings` for a type it needed, where this one hands the dependency down as a function. Passing the capability in is the pattern to keep.

**Nit:** two trivia — `assertTrue(nextStartedService != null)` reads better as `assertNotNull`, and a `typealias ChecksumCheck = suspend (String) -> Boolean` would carry the intent that the old parameter type conveyed for free. The parameter name already does most of that work, so neither is worth a round.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #17319, opened 2026-09-17
all: smoother room list converters mapping (fixes #17302)

### LINKED ISSUE #17302: Migrate Room type converters (Converters.kt) from Gson to kotlinx.serialization
Unlike the network-payload conversions, this changes the JSON format Room
stores in **already-installed users' local databases**. A wrong decode here
doesn't just fail a request, it could silently corrupt or drop local data.

- Verified no Room schema version bump is required: this is a converter
  logic change, not a table/column definition change, so
  fallbackToDestructiveMigration never triggers and existing rows are read
  by the new converter as-is.
- The Json instance is configured with ignoreUnknownKeys/isLenient/
  coerceInputValues specifically to tolerate whatever a prior Gson-written
  row looks like.
- Added explicit backward-compatibility tests that hand-construct JSON in
  Gson's exact prior output shape (null reference fields omitted entirely,
  primitive fields always written even at their zero/false default) and
  confirm the new converter decodes it correctly - covering an all-defaults
  row, a fully-populated row, and a row with only some nulls, for both
  Conversation and Attachment lists.

Okuro3499:
fixes #17302
Replace the Room list converters for strings, conversations, and attachments with kotlinx.serialization, and enable the Kotlin serialization plugin for the app module. Mark the embedded Conversation and Attachment models as serializable, and add converter tests to lock in compatibility with JSON already written by the previous Gson-based implementation.


dogi:
**Review — rating: 84/100**

The risk that matters in this PR is rows already sitting in installed databases, and it's taken seriously rather than hand-waved — the `ignoreUnknownKeys` / `isLenient` / `coerceInputValues` trio is the right defensive set for decoding legacy data, and the four new tests simulate the actual shapes Gson produces (nulls omitted, primitives always written) instead of round-tripping kotlinx against itself.

I checked the assumption those tests rest on: `JsonUtils.gson` is a bare `Gson()` — no `serializeNulls`, no `FieldNamingPolicy`, no registered adapters — so stored JSON really is default Gson output with verbatim field names, exactly as the fixtures assume. Both embedded models are clean for this too: `Conversation` is two nullable strings, and `Attachment` is four nullable strings plus `length: Long = 0`, `isStub: Boolean = false`, `revpos: Int = 0`, all with defaults, so missing keys decode fine.

Worth noting one thing that happens to be safe rather than by design: `isStub` is the classic place these two libraries diverge, since a Kotlin `is`-prefixed boolean has JVM field `isStub` but getter-derived name `stub`. Gson is field-based by default, so it wrote `isStub` and kotlinx reads `isStub` — they line up. Had Gson been configured against getters, this field would silently have stopped decoding.

The "no schema version bump needed" claim is also right: Room's schema hash is derived from the entity declarations, not from converter internals, and the column stays TEXT.

**The KDoc overstates one thing.** It says a plain field-matching class "round-trips identically either way", but kotlinx's `Json` defaults `encodeDefaults = false`, so on the *write* side an `Attachment` at its defaults now serializes to `{}` where Gson wrote `{"length":0,"isStub":false,"revpos":0}`. The decoded *values* are identical — defaults are restored — so this is safe in both directions, including if an older build reads the column back through Gson. But the JSON text is no longer byte-identical, and that's the one behavioural change this migration actually introduces. I'd reword the comment and add a round-trip test that pins it: encode a default-valued `Attachment`, assert it decodes back equal. Right now every new test exercises the read path only.

**Nit / merge-order note:** this PR and #17262 both add the `kotlin-serialization` alias to `libs.versions.toml` and the `alias(libs.plugins.kotlin.serialization)` line to `app/build.gradle`. They're not duplicates — different targets — but whichever lands second will conflict on those two hunks, and #17320 is behind them in the same chain.

**Verdict:** mergeable.

---

dogi (approved):

PR #17320, opened 2026-09-17
Migrate model JSON builders to kotlinx (fixes #17275)

### LINKED ISSUE #17275: Migrate model write-path JSON building from Gson to kotlinx.serialization
## Why                                                                                                                                                                             
                                                                                                                                                                                  
Continues the Gson -> kotlinx.serialization migration (KMP prep, tracked in #17261 / PR #17262, which handles the Retrofit boundary). An audit of the 116 files importing com.google.gson found the vast majority (102 of 116) hand-build or parse raw JsonObject/JsonArray trees rather than relying on Gson's reflection-based (de)serialization of @Serializable-style classes - so this isn't an annotation swap, it's a rewrite of the mapping code itself, file by file.   

## Scope                                                                                                                                                                                                                                                                                                                                                         
Converts the **write direction** only: every model-file function that builds and returns a JsonObject/JsonArray (serialize()/toJson()/getXArray()-style payload builders), across 23 files - Achievement, Answer, ApkLog, CourseActivity, CourseProgress, ExamQuestion, Feedback, HealthExamination, Meetup, MyCourse, MyLibrary, MyPlanet, MyTeam, News, NewsLog, Rating, SearchActivity, StepExam, SubmitPhotos, TagEntity, TeamTask, UserEntity, UserSurveyProfile.                                                                                                                                                                 
                                                                                                                                                                                     
These now build via kotlinx.serialization's buildJsonObject/buildJsonArray DSL internally, then bridge back to Gson's JsonObject/JsonArray at each function's return (via two new JsonUtils extensions, toGson()/toKotlinx()) - so every caller (repositories, upload/sync code) is unchanged, and the diff is contained entirely to how each function's own body is built.                                                                                                                       
                                                                                                                                                        
The **read direction** (fromJson(obj: JsonObject): Entity-style parsing) is deliberately out of scope: it still consumes JsonObject sourced from the un-migrated raw Retrofit endpoints (getJsonObject/postDoc/putDoc/etc., ~25 call sites in the sync/upload core), so converting the parsing logic now would only add a Gson<->kotlinx bridge with no actual reduction in Gson usage at the real source. That's a separate, larger, higher-risk follow-up once the network boundary itself moves off Gson. 

## Testing                                                                                                                                                                         
                                                                                                                                                                                     
  - Existing regression tests for these functions (MyTeamTest, RatingTest,TeamTaskTest, StepExamTest, ExamQuestionTest, AchievementTest,                                                                                                                   
    UploadConfigsTest, MyPlanetTest) pass unchanged, since return types never changed.                                                                                                                                                                         
  - Added golden-output tests for the functions that had zero prior coverage and the most branching logic: UserEntity.serialize (new-user vs   existing-user branches, invalid-iterations fallback), HealthExamination.serialize (every zero/null-omission rule, plus the pre-existing creatorId/profileId quirk - preserved as-is, now explicitly documented rather than silently relied on), MyCourse.serialize (nested steps/resources embedding), and MyPlanet.getNormalMyPlanetActivities.                                                                                                                                            
  - Full unit suite green against current master (2182 tests; the only 2 failures, in FreeSpaceWorkerTest, are pre-existing and unrelated - confirmed present on unmodified master too).

Okuro3499:
fixes #17275
Add Gson↔kotlinx JSON bridge helpers and refactor many model serialization methods to build payloads with kotlinx.serialization instead of manual Gson mutation. The new tests lock in payload shape for nested data, null handling, device metadata, and preserved quirks so sync/upload JSON stays compatible during the migration.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved consistency and null handling when preparing data for synchronization and storage.
  * Ensured unset health examination values and invalid optional data are omitted appropriately.
  * Preserved existing JSON fields and conditional behavior across courses, users, activities, news, and other records.

* **Tests**
  * Added coverage for serialization, nested course resources, health examinations, user data, activity payloads, news logs, and JSON conversion.


dogi:
**Review — rating: 68/100**

Reviewed at the hunk level given the size (30 files, +942/−518): I read the `JsonUtils` bridge in full, sampled `NewsLog` as a representative model conversion, read the new bridge test, and checked the pattern's consistency by counting call sites rather than reading all 23 model diffs.

The migration itself is careful. The conversion is faithful — `put("user", log.userId)` with a null `String?` produces `"user": null` exactly as Gson's `addProperty` did, and numeric `put` matches `addProperty` — the payload shape is locked in by five updated model tests, and the pattern is applied uniformly across all 23 models. The reasoning behind a private `bridgeGson`, so a test doing `mockkStatic(JsonParser::class)` for its own business logic can't break an unrelated internal round-trip, is genuinely thoughtful.

**The problem is how the bridge converts.** Both directions go through serialized text:

```kotlin
fun kotlinx.serialization.json.JsonObject.toGson(): JsonObject =
    bridgeGson.fromJson(toString(), JsonObject::class.java)
```

So the typical model now reads `buildJsonObject { … }.toGson()` — build a kotlinx tree, flatten it to a string, re-parse it into a Gson tree, then keep mutating it with Gson (`ob.addDocumentOrigin()`). The diff has 36 `toGson()`, 22 `toKotlinx()` and one `toGsonElement()`, so 59 conversion sites, each a full serialize-plus-parse of an entire document.

These are the per-document sync paths — `MyLibrary`, `MyCourse`, `News`, `UserEntity`, `TeamTask`, `HealthExamination` — called once per CouchDB document during upload, which for a resync is thousands of documents. The old code did zero serialization to build the same payload. So this lands as a measurable slowdown on a hot path in exchange for an intermediate state whose payoff only arrives if Gson later leaves entirely and the bridges are deleted.

The fix doesn't require rethinking the PR: implement the bridges as direct tree walks (recurse the kotlinx `JsonObject`/`JsonArray`/primitives into the Gson equivalents and back) instead of `toString()` + reparse. Same API, same call sites, no string round-trip — and `bridgeGson` stops being needed at all, which also removes the test-mocking workaround it exists for.

**Two more, both small:**

- **Inline fully-qualified names.** The four bridge declarations spell out `kotlinx.serialization.json.JsonObject`, `…JsonArray`, `…JsonElement` and `…Json` inline. That's the collision case the style guide addresses explicitly, and #17262 in this same chain already does it the prescribed way with `import kotlinx.serialization.Serializable as KSerializable`. An alias such as `import kotlinx.serialization.json.JsonObject as KJsonObject` fits here.
- **Bridge coverage is thin.** `JsonUtilsKotlinxBridgeTest` has one test, and it exercises only `JsonObject.toGson()`. `toKotlinx()` — 22 call sites — plus `JsonArray.toGson()` and `toGsonElement()` have no test at all. For code every model's serialization now routes through, each direction deserves a round-trip case.

**On CI:** the check runs do belong to the current head (`56db3f4`, committed 13:19Z, runs started 13:32Z), so they aren't stale. But both builds and both test shards finished in 45–60 seconds, where the same workflows take 4–8 minutes on sibling PRs like #17262 and #17319. That may just be a warm Gradle cache; given this is a 30-file diff, it's worth confirming the jobs actually executed before leaning on the green.

**Verdict:** needs another round — convert the bridges to direct tree walks so the model serializers stop paying a serialize/parse per document, then alias the kotlinx imports and cover the other three bridge functions.

---

dogi:
**Re-review — Rating: 85/100** (was 76)

**Verdict:** mergeable. Reviewed at hunk level again given the size, this time reading the rewritten bridge in full since that is where the change landed.

**Replacing the string round-trips with direct tree conversion is the right call, and the implementation holds up.** Serialising to a string and reparsing through the other library is exactly where type fidelity goes missing — an integer comes back as a double, a string that looks like a boolean comes back as a boolean — and it costs two full parses per conversion on a path used by every model's upload payload. The tree walk avoids both.

**The primitive conversion ordering is the part that has to be right, and it is:**

```kotlin
if (isString) return JsonPrimitive(content)
booleanOrNull?.let { return JsonPrimitive(it) }
longOrNull?.let { return JsonPrimitive(it) }
doubleOrNull?.let { return JsonPrimitive(it) }
return JsonPrimitive(content)
```

Checking `isString` **first** is what keeps a quoted `"true"` or `"42"` a string rather than silently promoting it — the single most likely way this migration could have corrupted sync payloads. And `longOrNull` before `doubleOrNull` preserves integer-ness, so a timestamp does not come out the far side as `1.7e12`. The reverse direction is symmetric: `isBoolean` / `isNumber` / else string, and passing Gson's `LazilyParsedNumber` straight into `KJsonPrimitive(Number)` keeps the original literal rather than forcing a widening.

I checked that `JsonUtilsKotlinxBridgeTest.kt` exists and covers the arrays/primitives/round-trip cases the commit message describes.

**One narrow lossy case worth a comment in the code:** an integer larger than `Long.MAX_VALUE` fails `longOrNull`, falls through to `doubleOrNull`, and comes back as a double — so `12345678901234567890` round-trips to `1.2345678901234567E19`. Gson's `LazilyParsedNumber` would have preserved it. I do not think this app hits it — CouchDB `_rev` values are strings and the numeric fields here are timestamps and counts, all comfortably inside `Long` — but it is the one input class where the bridge is not faithful, and a one-line note above `toGsonPrimitive` would save the next person the same analysis.

Related and even narrower: a primitive whose content is `NaN` or `Infinity` parses via `doubleOrNull` into a Gson primitive that throws when Gson serialises it. Not reachable from server data, but it is the other edge of the same fall-through.

CI green 5/5.

---

dogi (approved):

PR #17321, opened 2026-09-17
login: smoother activities repository device name providing (fixes #17299)

### LINKED ISSUE #17299: serializeLoginActivities threads a Context to a NetworkUtils call that ignores it
## Verified premise

`repository/ActivitiesRepositoryImpl.kt:307` takes a `Context` parameter:

```kotlin
private fun serializeLoginActivities(activity: OfflineActivity, context: Context): JsonObject {
```

passed in at L215 (`serializeLoginActivities(activity, context)`), and used for exactly one thing at L317:

```kotlin
ob.addProperty("customDeviceName", NetworkUtils.getCustomDeviceName(context))
```

That callee ignores the argument entirely — `utils/NetworkUtils.kt:243-245`:

```kotlin
    return sharedPrefManager.getCustomDeviceName()
}
```

So a `Context` is threaded through a repository method purely to be discarded one frame later.

The replacement already exists and is already used by two sibling repositories — `utils/DeviceNameProvider.kt:11` declares the interface, `SharedPrefDeviceNameProvider` (L16) implements it, it is `@Binds`-bound at L27, and `ResourcesRepositoryImpl.kt:64` and `PersonalsRepositoryImpl.kt:23` both inject it.

## Do

- Inject `private val deviceNameProvider: DeviceNameProvider` into `ActivitiesRepositoryImpl`, matching how `ResourcesRepositoryImpl` and `PersonalsRepositoryImpl` declare it.
- Replace L317 with the provider call.
- Drop the `context` parameter from `serializeLoginActivities` and from the L215 call site.

## Constraints

- **Do not remove `@ApplicationContext private val context: Context` from the constructor (L46). It has other live uses in this file — `MyPlanet.getNormalMyPlanetActivities(context, ...)` at L435, `VersionUtils.getAndroidId(context)` at L440, `MyPlanet.getTabletUsages(context, ...)` at L448 and `MyPlanet.getMyPlanetActivities(context, ...)` at L454. Only the `serializeLoginActivities` parameter goes.**
- Do not change `NetworkUtils.getCustomDeviceName`'s signature. Its unused parameter has further call sites elsewhere in the tree; those are not this task's to fix.
- Do not invent a logging pattern in this file — leave existing `Log` calls exactly as they are.
- `DeviceNameProvider` returns the same `sharedPrefManager.getCustomDeviceName()` value, so the serialized `customDeviceName` property is unchanged.

## Tests

- `app/src/test/java/org/ole/planet/myplanet/repository/ActivitiesRepositoryImplTest.kt` must pass with **only the constructor call updated** to pass a mocked `DeviceNameProvider`; all assertions stay.
- Add a test asserting the serialized activity carries the provider's device name.

## Out of scope

- The other six call sites of the dead `Context` parameter elsewhere in the tree.
- `NetworkUtils` itself.
- The upload path that consumes the serialized activities.

dogi:
Injected DeviceNameProvider into ActivitiesRepositoryImpl, replaced NetworkUtils.getCustomDeviceName(context) in serializeLoginActivities with deviceNameProvider.getCustomDeviceName(), dropped the unused Context parameter from serializeLoginActivities, and updated unit tests.

Fixes #17299

---
*PR created automatically by Jules for task [6072464069884721651](https://jules.google.com/task/6072464069884721651) started by @dogi*


dogi:
**Review — rating: 90/100**

Verified equivalent, and the reason is better than the diff lets on. Following both paths:

```kotlin
// NetworkUtils
    return sharedPrefManager.getCustomDeviceName()   // context is never touched
}

// SharedPrefDeviceNameProvider
override fun getCustomDeviceName(): String = sharedPrefManager.getCustomDeviceName()
```

Both land on `pref.getString(CUSTOM_DEVICE_NAME, "")`, so the serialized `customDeviceName` is byte-identical — which matters here, since this value goes up to CouchDB.

The interesting part is that `NetworkUtils.getCustomDeviceName` declares a `Context` it completely ignores. So `serializeLoginActivities(activity, context)` was threading a parameter through solely to feed an argument that was discarded on arrival. Dropping it removes dead plumbing at two levels, not just one, and the injected provider replaces a static that was reaching for a `sharedPrefManager` of its own.

I also checked that `context` is still earned in the constructor — lines 435, 440, 448 and 454 use it for `MyPlanet.getNormalMyPlanetActivities`, `VersionUtils.getAndroidId`, `getTabletUsages` and `getMyPlanetActivities` — so keeping it is right and this isn't a half-removal.

The new test is the right shape too: it captures the actual posted `JsonObject` and asserts `customDeviceName` in the serialized payload, rather than just verifying the provider was called.

**Nit:** `NetworkUtils.getDeviceName()` on the line above stays static, but it's context-free `Build` data and `DeviceNameProvider` doesn't expose an equivalent, so there's nothing to do there. Separately, since `NetworkUtils.getCustomDeviceName(context)` ignores its argument, its other callers are easy follow-up candidates for the same treatment.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #17322, opened 2026-09-17
resources: smoother add resource user view modelling (fixes #17306)

### LINKED ISSUE #17306: resolve the current user inside AddResourceViewModel instead of injecting UserSessionManager into the fragment
## Verified premise

`ui/resources/AddResourceFragment.kt` injects a session manager at L65:

```kotlin
import org.ole.planet.myplanet.services.UserSessionManager   // L42
...
lateinit var userSessionManager: UserSessionManager          // L65
```

used in exactly one place, L264:

```kotlin
val userModel = userSessionManager.getUserModel() ?: return@launch
showAlert(
    ...
```

Meanwhile `ui/resources/AddResourceViewModel.kt:14-16` injects only one dependency:

```kotlin
class AddResourceViewModel @Inject constructor(
    private val personalsRepository: PersonalsRepository
) : ViewModel() {
```

so the fragment reaches for the current user itself rather than asking the ViewModel it already owns.

## Do

- Add `private val userRepository: UserRepository` to `AddResourceViewModel` and expose a `suspend fun currentUser(): UserEntity?` accessor.
- Replace the L264 lookup with a call to that accessor, keeping the `?: return@launch` early exit.
- Delete the `UserSessionManager` field injection and its import from the fragment.

## Constraints

- **Check what `showAlert(...)` at L265 reads off `userModel` before changing anything — if it needs more than one field, the accessor must return the entity, not an id. Narrowing it to a `userId` here would break the call and is not what this task asks for.**
- Keep the `?: return@launch` guard exactly where it is: the alert must not be shown when there is no active user.
- `UserSessionManager.getUserModel()` and `UserRepository.getUserModel()` must return the same active user — confirm which one `AddResourceViewModel`'s siblings use and match them, rather than assuming.
- Do not touch `AddResourceActivity.kt` — it is owned by an open PR.
- Leave `checkTitleExists` and the `AddResourceState` flow alone.

## Tests

- Existing `app/src/test/java/org/ole/planet/myplanet/ui/resources/` tests must pass with **only the ViewModel's constructor call updated** to pass a mocked `UserRepository`.
- Add a ViewModel test asserting `currentUser()` returns what `userRepository.getUserModel()` returns, including null.

## Out of scope

- `AddResourceActivity.kt` (PR-owned).
- `PersonalsRepository` and the resource-save path.
- `UserSessionManager` itself and its other consumers.

dogi:
Refactored AddResourceFragment to resolve current user via AddResourceViewModel rather than directly injecting UserSessionManager into the fragment. Added UserRepository dependency to AddResourceViewModel and exposed currentUser() accessor.

Fixes #17306

---
*PR created automatically by Jules for task [13430437186528070545](https://jules.google.com/task/13430437186528070545) started by @dogi*


dogi:
**Review — rating: 84/100**

Behaviour is unchanged, which I checked rather than assumed: `UserSessionManager.getUserModel()` is itself a one-line delegate to `userRepository.getUserModel()`, so the fragment ends up at exactly the same `sharedPrefManager.getUserId()` → `userDao.getById(...)` lookup. The gain is structural — the fragment drops an injected data-access collaborator and talks to its own ViewModel instead, which is the same direction as #17286 and #17310 and matches how this codebase uses field injection in fragments (dispatchers and the like, not repositories).

Two things keep this from scoring higher, both minor:

- **`currentUser()` is a pure passthrough**, so the two new tests mostly assert that a delegation delegates. Compare #17286, where the equivalent ViewModel method actually absorbed the blank-id check and early return, and its tests covered real branches. Not wrong, just low yield.
- **Declaration placement.** `suspend fun currentUser()` sits between the constructor and `private val _state`, splitting the backing-StateFlow pair off from the top of the class. The convention in this codebase is the `_state`/`state` pair first, methods after.

Worth considering, though I wouldn't block on it: ViewModels here generally expose state as `StateFlow` rather than suspend getters. A suspend getter is pragmatic given the fragment needs the user imperatively inside a click handler, so this is a reasonable exception rather than a problem.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #17323, opened 2026-09-17
sync: smoother user data upload working (fixes #17314)

### LINKED ISSUE #17314: collapse the six repeated upload-await blocks in UserDataWorker
## Verified premise

`services/UserDataWorker.kt`, the `UPLOAD_TYPE_BULK` branch (L38 onward), repeats one shape six times — at L52/57, L61/67, L71/77, L83/89, L94/100 and L104/110:

```kotlin
val d = CompletableDeferred<Unit>()
... start an upload, completing d from an OnSuccessListener ...
withTimeoutOrNull(30000L) { d.await() }
```

once each for user-activities, exam-result, resource, submit-photos, activities and shelf-user-data. Only the call that kicks off the upload differs; the deferred, the listener plumbing and the 30-second timeout are identical every time.

## Do

- Add one private helper, e.g.:
  ```kotlin
  private suspend fun awaitUploadCompletion(timeoutMs: Long = 30000L, start: (onComplete: () -> Unit) -> Unit) {
      val d = CompletableDeferred<Unit>()
      start { d.complete(Unit) }
      withTimeoutOrNull(timeoutMs) { d.await() }
  }
  ```
- Rewrite the six blocks as calls to it, each passing only its own upload kick-off.

## Constraints

- **Leave the `UPLOAD_TYPE_LOGIN` branch alone (L28-36). It uses `CompletableDeferred<String?>` and reads the awaited value — `val successMsg = withTimeoutOrNull(30000L) { deferred.await() }` at L35 — so it is a different result type with a different consumer. Folding it into a `Unit`-returning helper drops the message.**
- The 30-second timeout applies per upload, not to the whole branch — six sequential calls still mean up to six timeouts, exactly as today. Do not wrap the whole branch in one `withTimeoutOrNull`.
- `withTimeoutOrNull` returning null (a timed-out upload) must keep being ignored and the next upload must still run, as it does now.
- Keep the six uploads in their current order.
- Some of the six may complete via a `() -> Unit` callback rather than an `OnSuccessListener` — the helper's `start` parameter must accommodate both without changing either call's semantics.

## Tests

- Existing `app/src/test/java/org/ole/planet/myplanet/services/UserDataWorkerTest.kt` must pass **unedited**, including anything covering the login branch.
- Add a test that a timed-out upload does not prevent the following one from starting.

## Out of scope

- The upload implementations themselves.
- `UploadToShelfService` and `UploadManager`.
- The worker's scheduling and constraints.

dogi:
Collapse six repeated upload-await blocks in UserDataWorker into a single private helper function awaitUploadCompletion and add UserDataWorkerTest unit test covering timeout handling and execution order.

Fixes #17314

---
*PR created automatically by Jules for task [11731695035946063201](https://jules.google.com/task/11731695035946063201) started by @dogi*


dogi:
**Review — rating: 88/100**

Six near-identical `CompletableDeferred` + `withTimeoutOrNull(30000L)` blocks collapse into one helper, and the extraction is faithful — each call still gets its own deferred and its own independent 30s window, so a stuck upload still can't consume another's budget. Putting the lambda last in the signature keeps the call sites in trailing-lambda form, which is why they read as one line each.

Replacing the `object : OnSuccessListener { override fun onSuccess(success: String?) { … } }` boilerplate with `{ onComplete() }` is a large readability win, and it's safe for the same reason it compiles — `OnSuccessListener` is SAM-convertible — while the discarded `success: String?` was already ignored in every one of the six originals.

**The new test file is the best part of this PR.** `UserDataWorker` had no coverage at all, and the bulk-upload test pins the property that actually matters about this structure: `uploadUserActivities` is stubbed to never invoke its listener, and `coVerifyOrder` then asserts that all six subsequent uploads still run, in order. That's a direct test of the `runCatching` + timeout resilience design rather than of the refactor's mechanics. It also works out cheaply — `runTest`'s virtual clock means the 30s timeout elapses instantly rather than stalling CI.

**Nit:** the `30000L` default is the only value any caller uses — all six take the default — and this class already has a `companion object` holding its constants, so a named `UPLOAD_TIMEOUT_MS` there would fit the file's own conventions better than an inline literal in a parameter default.

One thing worth being aware of rather than changing: `withTimeoutOrNull` wraps only `d.await()`, not the `start` call, so a `start` that suspends on its own network work isn't covered by the timeout. That was equally true of all six originals, so it's preserved behaviour, not a regression — but the helper's name makes it read like a bound on the whole operation.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #17326, opened 2026-09-17
sync: less feedback list realtime mixin is more (fixes #17305)

### LINKED ISSUE #17305: FeedbackListFragment restarts a collector that its ViewModel already keeps running
## Verified premise

`ui/feedback/FeedbackListViewModel.kt` already collects the repository flow continuously, cancelling any previous job first:

```kotlin
private var fetchJob: Job? = null

init { loadFeedback() }

private fun loadFeedback() {
    fetchJob?.cancel()
    fetchJob = viewModelScope.launch {
        val user = userRepository.getUserModel()
        feedbackRepository.getFeedback(user).collectLatest { feedback ->
            _feedbackList.value = feedback
        }
    }
}
fun refreshFeedback() { loadFeedback() }
```

A `Flow` from the feedback DAO already re-emits when the table changes. Yet `ui/feedback/FeedbackListFragment.kt` wires up a second refresh path on top of it:

```kotlin
import org.ole.planet.myplanet.services.sync.RealtimeSyncManager      // L18
import org.ole.planet.myplanet.ui.sync.RealtimeSyncHelper             // L19
import org.ole.planet.myplanet.ui.sync.RealtimeSyncMixin              // L20

class FeedbackListFragment : ..., OnChangedListener, RealtimeSyncMixin {   // L24
    lateinit var realtimeSyncManager: RealtimeSyncManager              // L29
    private lateinit var realtimeSyncHelper: RealtimeSyncHelper        // L32
    ...
    private fun setupRealtimeSync() {                                  // L51
        realtimeSyncHelper = RealtimeSyncHelper(this, this, realtimeSyncManager)
        realtimeSyncHelper.setupRealtimeSync()
    }
    private fun refreshFeedbackListData() { viewModel.refreshFeedback() }   // L56
    ...
    override fun getWatchedTables(): List<String> = listOf("feedback")      // L97
    override fun onSyncTableChanged(...) {                                  // L100
        if (table == "feedback" && update.shouldRefreshUI) refreshFeedbackListData()
```

So every `"feedback"` sync event tears down a live collector and starts a new one, re-running `getUserModel()` and re-subscribing — to observe a change the existing flow was already going to deliver. It also couples a fragment to sync internals.

## Do

- Remove `RealtimeSyncMixin` from the class declaration, the `realtimeSyncManager` injection, the `realtimeSyncHelper` field, `setupRealtimeSync()` and its call site, and the `getWatchedTables` / `onSyncTableChanged` overrides, plus the three imports.
- Keep the ViewModel's single collection alive for its lifetime.

## Constraints

- **Keep `refreshFeedback()` on the ViewModel and keep the fragment's explicit call after a feedback submission — that path is not the sync mixin and is the one case where the list must refresh from a user action rather than a table event. Find every caller of `refreshFeedbackListData()` before deleting it; only the `onSyncTableChanged` caller goes.**
- `loadFeedback()` already does `fetchJob?.cancel()` before relaunching, so `refreshFeedback` cannot leave parallel collectors — do not add extra guarding.
- `OnChangedListener` stays on the class; only `RealtimeSyncMixin` is removed.
- Do not change the ViewModel's collection or `_feedbackList`.

## Tests

- `app/src/test/java/org/ole/planet/myplanet/ui/feedback/FeedbackListViewModelTest.kt` must pass **unedited** — its `verify(exactly = 1)` / `verify(exactly = 2)` assertions pin that one refresh restarts the collector exactly once, which this change must not alter.

## Out of scope

- `FeedbackRepository`'s signatures (changed by a separate open PR — rebase on it if it has landed).
- `RealtimeSyncManager`, `RealtimeSyncHelper` and `RealtimeSyncMixin` themselves.
- The four other fragments doing the same thing; this issue covers only the feedback list.

dogi:
Removed `RealtimeSyncMixin` implementation and related sync helper/manager dependencies from `FeedbackListFragment`. The `FeedbackListViewModel` already maintains an active collection flow from the database flow, rendering the table change listener in the fragment redundant. `OnChangedListener` and `refreshFeedback()` remain for handling manual user feedback submission.

Fixes #17305

---
*PR created automatically by Jules for task [14386634549694548963](https://jules.google.com/task/14386634549694548963) started by @dogi*


dogi:
**Review — rating: 89/100**

"Redundant" checks out, and the mixin was doing more than nothing — it was doing harm.

`FeedbackDao.getAllSortedFlow()` and `getByOwnerFlow()` are Room `@Query` methods returning `Flow<List<Feedback>>`, so Room's `InvalidationTracker` re-emits on any write to the `feedback` table through the shared `AppDatabase` — including sync writes, since those go through `feedbackDao.upsert`. `FeedbackListViewModel.loadFeedback()` collects that flow into `_feedbackList`, and the fragment already renders it via `collectWhenStarted`. The data was always going to arrive on its own.

What the mixin added on top was `onDataUpdated` → `refreshFeedbackListData()` → `viewModel.refreshFeedback()` → `loadFeedback()`, which does `fetchJob?.cancel()` and starts over — tearing down a live subscription, re-running `userRepository.getUserModel()`, and re-subscribing to a flow that had already delivered the same update. So every sync event was cancelling and rebuilding the collection that had just done the work, with a `collectLatest` cancellation racing in the middle. Removing it is a behaviour improvement, not just dead-code cleanup.

Supporting evidence that the flow is the intended mechanism: `FeedbackRepositoryImpl.getFeedback` wraps the DAO flow in `distinctByContent` specifically to swallow byte-identical re-emissions, which is only worth doing for a continuously observed flow.

Also worth noting `shouldAutoRefresh` already returned `false`, so that manual refresh was the mixin's only actual effect here — there's nothing else being given up.

No test, which is right for a deletion of this shape: `FeedbackListViewModelTest` already covers the flow path that remains.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #17327, opened 2026-09-17
Optimize file stat calls in InlineResourceAdapter

### LINKED ISSUE #17307: InlineResourceAdapter stats the same file three times on every row bind
## Verified premise

`ui/courses/InlineResourceAdapter.kt`, `updateStatusAndPreview` (L162-203), touches the filesystem for one row at least three times.

First a resolve-and-stat through a URL:

```kotlin
val isDownloaded = resource.isResourceOffline() || withContext(dispatcherProvider.io) {
    FileUtils.checkFileExist(context, UrlUtils.getUrl(resource))
}
```

then the same file is constructed again, by hand:

```kotlin
val resourceFile = File(
    externalFilesDir,
    "ole/${resource.id}/${resource.resourceLocalAddress}"
)

when {
    mimeType?.startsWith("image") == true -> showImagePreview(binding, context, resourceFile)
    ...
}
```

and the preview helpers stat it a third time — `getFileCacheKeyIfExist` (L312-318) plus `getCacheKey` (L320):

```kotlin
private suspend fun getFileCacheKeyIfExist(file: File): String? = withContext(dispatcherProvider.io) {
    if (file.exists()) getCacheKey(file) else null
}

private fun getCacheKey(file: File): String = "${file.absolutePath}_${file.lastModified()}_${file.length()}"
```

`exists()`, `lastModified()` and `length()` are three more syscalls, run per bind, and RecyclerView rebinds every row on every scroll pass.

## Do

- Resolve the `File` once at the top of the coroutine and thread it through: use it for the existence check, for the `when` dispatch, and for the cache key.
- Capture `exists()` / `lastModified()` / `length()` once into locals and build the cache key from those, rather than re-statting inside `getCacheKey`.

## Constraints

- **`FileUtils.checkFileExist(context, UrlUtils.getUrl(resource))` and `File(externalFilesDir, "ole/${resource.id}/${resource.resourceLocalAddress}")` are two different derivations of a path. Read `checkFileExist` and `UrlUtils.getUrl` and prove they land on the same file before collapsing them into one resolve — if they differ for any resource shape, the existence check and the preview would start disagreeing and rows would render blank.**
- `resource.isResourceOffline()` short-circuits the check with `||`; keep that ordering so an offline-flagged resource still skips the disk entirely.
- Do not change the cache-key **format** (`absolutePath_lastModified_length`) — a different string invalidates every cached preview.
- Do not restructure the `holder.setPreviewJob(adapterScope.launch { ... })` coroutine or the `when` branches; this is about how many times the file is resolved and statted, not about control flow.
- The `html` branch passes a *directory* (`File(externalFilesDir, "ole/${resource.id}")`), not `resourceFile` — leave it as it is.

## Tests

- Existing `app/src/test/java/org/ole/planet/myplanet/ui/courses/` adapter tests must pass **unedited**.
- Add a test pinning the cache-key format for a fixture file, so a later refactor cannot change it silently.

## Out of scope

- Adding a cross-bind `FileExistenceCache` — that is a separate change and needs profiling to justify.
- `FileUtils.checkFileExist` and `UrlUtils.getUrl` themselves.
- The individual preview renderers' contents.

dogi:
Optimized InlineResourceAdapter to resolve resourceFile once and capture file existence and metadata (lastModified, length) on the IO dispatcher in a single pass, eliminating duplicate stat calls per row bind while preserving offline short-circuit logic and cache key formatting. Added unit test pinning cache key format.

Fixes #17307

---
*PR created automatically by Jules for task [14130583675469500218](https://jules.google.com/task/14130583675469500218) started by @dogi*

## Summary by CodeRabbit

* **Improvements**
  * Improved inline resource preview loading for faster, more reliable handling of external files.
  * HTML resources are now identified before other preview types.
  * Image, video, PDF, audio, CSV, and text previews now use more accurate file availability and freshness checks.
  * Preview caching now updates when a file’s location, modification time, or size changes.

* **Tests**
  * Added coverage to verify preview cache keys accurately reflect file details.


dogi:
**Review — rating: 68/100**

Passing the stat results down instead of having each preview function re-fetch them is a reasonable shape, and threading `exists`/`cacheKey` as parameters reads fine. But I counted the actual work per branch before and after, and the optimization in the title doesn't land — for most resource types this costs more, not less.

The old code only ever entered one `when` branch, so it only ever paid for one `withContext(dispatcherProvider.io)`:

| branch | before | after |
|---|---|---|
| image / video | 1 switch, 1 stat (`exists`) | 1 switch, **3 stats** |
| pdf | 1 switch, 1 stat (`exists`) | 1 switch, **3 stats** |
| html | **no stat on `resourceFile`** | 1 switch, **3 stats** |
| audio / csv / text | 1 switch, 3 stats | 1 switch, 3 stats |

The audio/csv/text path was already a single switch doing all three stats, because `getFileCacheKeyIfExist` wrapped `exists()` *and* `getCacheKey(file)` — which called `lastModified()` and `length()` — inside one `withContext`. So those three branches are unchanged, and the four that only needed `exists()` (or nothing) now pay for `lastModified()` and `length()` as well, because the stat block was hoisted above the type dispatch.

These are cheap `stat()` calls, so the real cost is small — but this is a RecyclerView bind path and the PR is framed as an optimization, so it should at least not go backwards. The fix is mechanical: `val mimeType = Utilities.getMimeType(...)` is computed immediately *after* the block and doesn't depend on it, so hoisting it above and only collecting `lastModified`/`length` for the branches that need a cache key gives you the intended consolidation with no branch regressing.

**`getFileCacheKeyIfExist` is now completely dead.** All three of its call sites take `cacheKey` as a parameter after this change, and nothing else references it — not the new test, not any existing one. It's marked `@VisibleForTesting internal`, but that annotation means "widened for tests", not "kept alive because nothing uses it"; the single-argument `getCacheKey(file)` overload is in the same position. Both should be deleted rather than annotated.

**The new test is a change detector.** It rebuilds the format string `"${file.absolutePath}_${file.lastModified()}_${file.length()}"` and asserts the implementation produces it, then asserts the one-arg overload agrees with the two-arg one it delegates to. That can only fail if someone edits the format — which, to be fair, is worth pinning since a format change silently invalidates `textCache` — but it duplicates the implementation rather than testing a property of it.

Credit where due: I checked `externalFilesDir ?: FileUtils.getExternalFilesDir(context)` expecting it to be a pointless retry, and it isn't — `externalFilesDir` is a plain `var` assigned in `onCreateViewHolder`, not a `by lazy`, so a genuine second attempt can succeed. That guard is a real improvement.

**Verdict:** needs another round — hoist `mimeType` above the stat block so only the cache-key branches collect `lastModified`/`length`, and delete `getFileCacheKeyIfExist` and the one-arg `getCacheKey` now that nothing calls them.

---

dogi:
**Re-review — Rating: 84/100** (was 68)

**Verdict:** mergeable — the optimization now actually optimizes, and the directory inconsistency is gone.

**The stat reduction is real this time.** Before, each `show*Preview` opened its own `withContext(dispatcherProvider.io) { file.exists() }`, and the three text-ish branches then called `getFileCacheKeyIfExist`, which opened a *second* IO dispatch to call `exists()` again plus `lastModified()` and `length()`. That was the inversion I objected to: a change framed as reducing filesystem work that added a dispatch per bind. Now there is exactly one `withContext(io)` block per bind, and `lastModified`/`length` are only read when `needsCacheKey` says a cache key will actually be built — so image, video and PDF binds do a single `exists()` and nothing more.

**Hoisting the HTML branch above the stat block is the neat part.** HTML previews resolve a *directory*, not `resourceFile`, so statting that path was pure waste. The early `return@launch` skips it entirely.

**And the directory inconsistency is fixed properly.** `val dir = externalFilesDir ?: FileUtils.getExternalFilesDir(context)` is now computed once and used for both the HTML directory and `resourceFile`, where previously the HTML branch reached for `externalFilesDir` directly and could diverge from the path the other branches used when it was null.

Removing `getFileCacheKeyIfExist` and the single-arg `getCacheKey` is good housekeeping — they were unreachable once the metadata was threaded through, and leaving them would have left two ways to build the same key.

**Two notes, neither blocking:**

**The `var exists/lastModified/length` assigned from inside `withContext` is an awkward shape.** It is correct — `withContext` suspends until the block completes, so there is a happens-before edge and no visibility problem — but it reads like shared mutable state when it is really a single return value. Returning a small data class (or `Triple`) from the `withContext` and destructuring it says the same thing without the `var`s:

```kotlin
val (exists, lastModified, length) = withContext(dispatcherProvider.io) { … }
```

**The new test checks the cache-key format, not the stat reduction.** `assertEquals("${path}_${lastModified}_${length}", adapter.getCacheKey(file, file.lastModified(), file.length()))` is close to true by construction — it passes the same values it compares against. It does have modest value as a guard: reordering the interpolation would silently invalidate every cached preview, and this would catch that. But the property this PR actually changed is "how many times do we touch the filesystem per bind", and nothing pins it. Counting IO-dispatcher dispatches, or asserting that an image-type bind never reads `lastModified`, would be the on-point test.

CI green 5/5.

---

Okuro3499 (approved):

PR #17328, opened 2026-09-17
life: smoother health examination examiner view modelling (fixes #17312)

### LINKED ISSUE #17312: move the examiner lookup into HealthExaminationState instead of injecting UserSessionManager into the activity
## Verified premise

`ui/health/HealthExaminationActivity.kt:43` injects a session manager:

```kotlin
import org.ole.planet.myplanet.services.UserSessionManager   // L27
...
lateinit var userSessionManager: UserSessionManager          // L43
```

read once at L78:

```kotlin
currentUser = userSessionManager.getUserModel()
```

and used for the examiner identity at L251 and L256:

```kotlin
examination?.isSelfExamination = currentUser?._id == pojo?._id
...
sign.createdBy = currentUser?._id
```

Meanwhile `ui/health/HealthExaminationViewModel.kt` already injects `userRepository`, and its state object (L22-29) carries the *patient*:

```kotlin
data class HealthExaminationState(
    val isLoading: Boolean = true,
    val user: UserEntity? = null,
    val pojo: HealthExamination? = null,
    val health: MyHealth? = null,
    val examination: HealthExamination? = null,
    val conditionsMap: Map<String, Boolean> = emptyMap()
)
```

So the activity resolves the examiner through a second source while the ViewModel already has a user repository in hand.

## Do

- Add `currentUser: UserEntity? = null` to `HealthExaminationState` **as the last property with a default**, and populate it from `userRepository.getUserModel()` in the same load path that fills the rest of the state.
- Read the examiner from the collected state at L251 and L256.
- Delete the `UserSessionManager` injection and its import.

## Constraints

- **`state.user` is the PATIENT and `currentUser` is the EXAMINER — the two are different people and the whole point of L251 is comparing them (`currentUser?._id == pojo?._id`). Do not touch a single `state.user` reference, and do not let the new field shadow or replace it. Conflating them makes every examination self-reported.**
- `state.currentUser` must be populated before the activity reaches the save path; if the load is asynchronous, make sure L251/L256 read the state after it has been emitted rather than an initial default.
- Keep the nullability: `currentUser?._id` must still handle a signed-out examiner exactly as today.
- `HealthExaminationState` is likely constructed positionally somewhere — appending last with a default keeps those call sites valid.

## Tests

- Existing `app/src/test/java/org/ole/planet/myplanet/ui/health/` tests must pass **unedited**.
- Add a ViewModel test asserting `state.currentUser` is the examiner from `userRepository.getUserModel()` while `state.user` remains the loaded patient — one test that would fail if the two were swapped.

## Out of scope

- The patient load path and `state.user`'s semantics.
- The examination save/sign flow beyond the two lines named.
- `AddExaminationActivity` and the other health screens.

dogi:
Moved examiner identity lookup into HealthExaminationState via UserRepository in HealthExaminationViewModel, and removed UserSessionManager field injection from HealthExaminationActivity.

Fixes #17312

---
*PR created automatically by Jules for task [11119591376844144547](https://jules.google.com/task/11119591376844144547) started by @dogi*


dogi:
**Review — rating: 89/100**

Like #17310, this is doing more than relocating a lookup — it removes a real ordering hazard. The examiner was resolved in its own `lifecycleScope.launch` while everything else (`user`, `pojo`, `health`, `examination`) came from a second launch awaiting `viewModel.state.first { !it.isLoading }`, with nothing ordering the two. Since `currentUser` is the *examiner* and `user` is the *patient*, a save that landed while the first coroutine hadn't finished would attribute the examination with a null examiner. Now all five fields are assigned together from one state emission, so they're consistent by construction.

The swap itself is behaviour-preserving: `UserSessionManager.getUserModel()` is a one-line delegate to `userRepository.getUserModel()`, and `userRepository` was already a ViewModel dependency (it's used for `ensureUserSecurityKeys`), so this adds no new wiring. Putting the call inside the existing `withContext(dispatcherProvider.io)` block alongside the other reads is the right placement.

The new test is the part I'd single out. Using two *distinct* mocks — `patientUser` and `examinerUser` — and asserting `finalState.user == patientUser` alongside `finalState.currentUser == examinerUser` is precisely the assertion this change needs, because the plausible failure mode here isn't "the field is empty", it's "the two users get crossed". A single shared mock would have passed either way. Adding the `getUserModel()` default stub in `setup()` is also necessary given `userRepository` is a strict `mockk()`.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #17331, opened 2026-09-17
life: smoother adapter content descriptions binding (fixes #17318)

### LINKED ISSUE #17318: LifeAdapter sets accessibility labels inside listeners, so TalkBack has no drag label until first touch
## Verified premise

`ui/life/LifeAdapter.kt`, `onBindViewHolder`, sets one `contentDescription` correctly at bind time:

```kotlin
holder.imageView.contentDescription = context.getString(R.string.icon, myLife.title)
```

but sets the other two *inside listeners*:

```kotlin
holder.dragImageButton.setOnTouchListener { _: View?, event: MotionEvent ->
    holder.dragImageButton.contentDescription = context.getString(R.string.drag, myLife.title)   // L67
    if (event.actionMasked == MotionEvent.ACTION_DOWN) {
        mDragStartListener.onStartDrag(holder)
    }
    false
}
holder.visibility.setOnClickListener {
    holder.visibility.contentDescription = context.getString(R.string.visibility_of, myLife.title)   // L74
    updateVisibility(holder)
}
```

Both strings depend only on `myLife.title`, which is fixed at bind. Two consequences:

1. **An accessibility defect.** The drag handle and the visibility toggle have no label until the user has already touched or clicked them — which is exactly the information a TalkBack user needs *before* deciding to touch.
2. `setOnTouchListener` fires for every `MotionEvent`, so the drag label is rebuilt through `getString` on every move event of a drag, not just once.

## Do

- Assign both `contentDescription`s once in `onBindViewHolder`, next to the existing `holder.imageView.contentDescription` line.
- Delete the two assignments from inside the listeners.

## Constraints

- **The touch listener must keep returning `false`.** It returns false so the event keeps propagating; an early return or a restructure that changes that value breaks dragging entirely.
- Keep the `event.actionMasked == MotionEvent.ACTION_DOWN` guard around `onStartDrag` exactly as it is.
- Use the same three string resources with the same arguments — `R.string.drag` and `R.string.visibility_of`, both formatted with `myLife.title`.
- `updateVisibility(holder)` toggles state; the label is bound to `myLife.title`, not to the visibility state, so it does not need updating when the toggle fires. Confirm that by reading `R.string.visibility_of` before assuming it.
- The `@SuppressLint("ClickableViewAccessibility")` annotation stays.

## Tests

- Existing `app/src/test/java/org/ole/planet/myplanet/ui/life/` tests must pass **unedited**.
- Add a bind-time test asserting all three `contentDescription`s are set after `onBindViewHolder` alone, with no touch or click — the test that pins the accessibility fix.

## Out of scope

- The drag-and-drop reordering behaviour.
- `updateVisibility` / `changeVisibility` and the icon swap.
- The `drawableCache` lookup.

dogi:
Fixes accessibility defect where dragImageButton and visibility toggle had no TalkBack accessibility labels until first touch/click. Content descriptions for all views in LifeAdapter are now set during onBindViewHolder, and redundant assignments inside listeners are removed. Added unit test verifying contentDescription setup at bind time.

Fixes #17318

---
*PR created automatically by Jules for task [13688892370921170791](https://jules.google.com/task/13688892370921170791) started by @dogi*


dogi:
**Review — rating: 91/100**

Genuine accessibility fix, not a cosmetic move. Setting `contentDescription` inside the listeners meant it didn't exist until the user had already interacted with the control — TalkBack announces a description when focus *lands* on a view, so the drag handle was unlabelled until touched, and the visibility toggle was unlabelled until clicked. A screen-reader user had to activate a control to find out what it was, which defeats the purpose. Setting both at bind time, next to the `imageView.contentDescription` line that was already doing it correctly, is the right shape.

There's a second benefit worth noting: `setOnTouchListener` fires for every `MotionEvent`, not just `ACTION_DOWN`, and that listener returns `false`, so `context.getString(R.string.drag, myLife.title)` was being formatted on every move event for the duration of a drag. Hoisting it to bind removes that entirely.

Recycling is handled correctly by construction — all three descriptions are now assigned unconditionally on each `onBindViewHolder`, so a reused holder can't keep the previous item's label.

The test is well targeted: it asserts all three descriptions immediately after `onBindViewHolder` with no interaction, which is exactly the assertion that fails on the old code (drag and visibility would both be null) and passes on the new one.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #17332, opened 2026-09-17
courses: smoother submissions repository pending querying (fixes #17313)

### LINKED ISSUE #17313: fetch one pending submission with a bounded query instead of materialising the whole list
## Verified premise

`repository/SubmissionsRepositoryImpl.kt` loads and hydrates a full list twice, to keep one row.

L456:

```kotlin
val submission = getSubmissionsByParentId(parentId, userId, "pending").firstOrNull()
```

L541:

```kotlin
getSubmissionsByParentId(parentId, userId ?: submission?.userId, "pending").firstOrNull()
```

`getSubmissionsByParentId` (L367) returns `List<Submission>` — every matching pending row, fully hydrated, so that `.firstOrNull()` can discard all but one.

The bounded query already exists — `data/room/dao/SubmissionDao.kt:28`:

```kotlin
@Query("SELECT * FROM submissions WHERE userId IS :userId AND parentId = :parentId AND status = 'pending' ORDER BY lastUpdateTime DESC LIMIT 1")
suspend fun getLatestPendingByUserAndParent(userId: String?, parentId: String): Submission?
```

## Do

- Compare `getSubmissionsByParentId`'s ordering and predicates against `getLatestPendingByUserAndParent` at both call sites.
- Where they match, call the bounded DAO query and hydrate only the row it returns.
- Where they do not, add one precisely ordered `LIMIT 1` query rather than bending the existing one.

## Constraints

- **The two call sites do not have the same parameters. L456 passes `userId`; L541 passes `userId ?: submission?.userId`, and the DAO query uses `userId IS :userId`, which *matches NULL rows* rather than matching nothing. Check what each site's `parentId` nullability is too — `getLatestPendingByUserAndParent` takes a non-null `parentId` while `getSubmissionsByParentId` takes `parentId: String?`. Collapsing both onto one query without checking these changes which row comes back, or crashes on a null parentId.**
- `.firstOrNull()` on the current list takes whatever order `getSubmissionsByParentId` produces — confirm that is `lastUpdateTime DESC` before assuming the bounded query returns the same row. If the existing query has no `ORDER BY`, the current behaviour is unspecified and the new one must match what the callers actually rely on; say so in the PR rather than guessing.
- Hydrate only the returned submission; do not keep the list-hydration path alive alongside.
- Do not change `getSubmissionsByParentId` itself — it has other callers that want the full list.

## Tests

- Existing `app/src/test/java/org/ole/planet/myplanet/repository/SubmissionsRepositoryImplTest.kt` must pass **unedited**.
- Add a DAO test with several pending rows at different `lastUpdateTime` values, asserting which one comes back — the test that would catch an ordering change.
- Add a case with a null `userId` to pin the `IS` (NULL-matching) semantics.

## Out of scope

- The other `SubmissionDao` queries.
- The submission upload path.
- `SubmissionsRepositoryExporter`.

dogi:
Replaced unbounded getSubmissionsByParentId(...).firstOrNull() calls in startExamSession and saveExamAnswer in SubmissionsRepositoryImpl with a bounded query getPendingByUserAndParent in SubmissionDao. The bounded query uses ORDER BY startTime DESC LIMIT 1 and IS NULL semantics for nullable parameters, matching caller ordering expectations while only hydrating the single returned submission.

Fixes #17313

---
*PR created automatically by Jules for task [13074045673223207876](https://jules.google.com/task/13074045673223207876) started by @dogi*


dogi:
**Review — rating: 90/100**

Correct optimization with the equivalence properly preserved. Both replaced sites were calling `getSubmissionsByParentId(..., "pending").firstOrNull()`, which materialised *and hydrated* every pending submission for that parent/user just to keep one. The old predicate was `(:status IS NULL OR status = :status)` with `status = "pending"` ordered by `startTime DESC`, so `status = 'pending' … ORDER BY startTime DESC LIMIT 1` is exactly the same row — same filter, same ordering, and `LIMIT 1` is `firstOrNull` of that ordering.

The detail that's easy to lose in this refactor is hydration, and it's kept: `hydrateSubmission(fetchPendingByUserAndParent(...))` preserves what `getSubmissionsByParentId` was doing via `hydrateSubmissions`, so callers still get a fully-populated `Submission` rather than a bare DAO row. The new query also gets the `IS` vs `=` discipline right — `IS :parentId` / `IS :userId` for the nullable parameters, `=` only against the `'pending'` literal.

Scoping is right too: `getByParentUserAndStatus` still has legitimate list-returning callers at `SubmissionsRepositoryImpl:376` and `:392` and through `ExamTakingFragment:106`, and those are left alone. Only the two "load all, take first" sites changed.

**The new `SubmissionDaoTest` is the best part.** It runs against a real in-memory Room database rather than a mocked DAO, so it actually validates the SQL, and the ordering test is deliberately constructed: three pending rows where `startTime` and `lastUpdateTime` rank differently, with `sub1` holding the highest `lastUpdateTime` (500) and `sub2` the highest `startTime` (300), asserting `sub2`. That's guarding a live hazard — the very next query in the same DAO, `getLatestPendingByUserAndParent`, orders by `lastUpdateTime DESC`, so a future edit conflating the two would be caught immediately. The null-semantics test earns its place as well, proving `IS` really does match NULL rows rather than trusting the convention.

**Nit:** `fetchPendingByUserAndParent` is a one-line private passthrough to the DAO; calling `submissionDao.getPendingByUserAndParent(...)` directly at both sites would read the same and carry one less indirection.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #17333, opened 2026-09-17
all: smoother user repository member device name providing (fixes #17324)

### LINKED ISSUE #17324: UserRepositoryImpl reaches for MainApplication.context to build a member document
## Verified premise

`repository/UserRepositoryImpl.kt:527-528`, inside the member-document builder:

```kotlin
addProperty("uniqueAndroidId", VersionUtils.getAndroidId(MainApplication.context))
addProperty("customDeviceName", NetworkUtils.getCustomDeviceName(MainApplication.context))
```

Two static reads of the Application's context from inside a repository — the layer furthest from Android. Both are on the path that builds the JSON for `becomeMember(obj)`, so the whole member-creation flow needs a live `MainApplication` to run, which is why it cannot be unit-tested without one.

## Do

- Add `@ApplicationContext private val context: Context` to `UserRepositoryImpl`'s constructor and use it for `VersionUtils.getAndroidId(context)`.
- Inject `DeviceNameProvider` and replace the second call with `deviceNameProvider.getCustomDeviceName()`.
- Remove the `MainApplication` import if nothing else in the file uses it — **grep first**, this file is ~1265 lines.

## Constraints

- **`NetworkUtils.getCustomDeviceName(context)` ignores its argument entirely — `NetworkUtils.kt:243-245` is `return sharedPrefManager.getCustomDeviceName()`. So passing the newly injected `context` to it would "fix" the static read while leaving a parameter that does nothing. Route the device name through `DeviceNameProvider` (precedent: `ResourcesRepositoryImpl`, `PersonalsRepositoryImpl`), not through the injected context.**
- `VersionUtils.getAndroidId(context)` **does** use its context — that one needs the real injected `@ApplicationContext`.
- Do not change `NetworkUtils.getCustomDeviceName` or `VersionUtils.getAndroidId`.
- The two JSON property names and their values must not change: the document goes to CouchDB and the server reads both keys.
- `UserRepositoryImpl` may already have a long constructor; add the two parameters, do not reorder existing ones.

## Tests

- Existing `app/src/test/java/org/ole/planet/myplanet/repository/UserRepositoryImplTest.kt` must pass with **only the constructor call updated** to pass a mocked `Context` and `DeviceNameProvider`.
- Add a test asserting the built member document carries the provider's device name and the injected context's android id — the test that would previously have required `MainApplication`.

## Out of scope

- The rest of the member-creation flow and `becomeMember`.
- The other ~40 methods on `UserRepositoryImpl`.
- Splitting this file (a separate, larger job).

dogi:
Refactored member creation in UserRepositoryImpl to remove static reads of MainApplication.context. Injected DeviceNameProvider and passed the constructor-injected @ApplicationContext Context into VersionUtils.getAndroidId, updated unit test suites, and verified all unit tests pass.

Fixes #17324

---
*PR created automatically by Jules for task [2298347696255482606](https://jules.google.com/task/2298347696255482606) started by @dogi*


dogi:
**Review — rating: 87/100**

Both substitutions are behaviour-preserving, and together they get `MainApplication.context` out of this repository entirely — the removed import is itself the proof that no other usage remained in the file, since the build would have failed otherwise.

- `NetworkUtils.getCustomDeviceName(context)` ignores its `Context` argument and delegates straight to `sharedPrefManager.getCustomDeviceName()`, which is exactly what `SharedPrefDeviceNameProvider` does, so the value is identical (same finding as #17321).
- `VersionUtils.getAndroidId(MainApplication.context)` → `getAndroidId(context)` resolves to the same object: the injected `@ApplicationContext` *is* the application context that `MainApplication.context` statically holds, and `ANDROID_ID` is read from `Settings.Secure` regardless of which context is passed.

The test is better than a typical DI-swap test. Capturing the `JsonObject` handed to `becomeMember` and asserting the two fields in the *built payload* checks the thing that actually ships to CouchDB, and `verify { VersionUtils.getAndroidId(context) }` pins that the **injected** context was used rather than merely that some context was — which is the specific claim this PR makes.

**Nit:** `mockkObject(VersionUtils)` is undone by an `unmockkObject` at the end of the test body rather than in `@After` or a `finally`. If an assertion throws, `VersionUtils` stays mocked for every later test in the class and one real failure turns into a cascade. Same pattern I flagged on #17304 — worth fixing consistently.

Worth noting in passing, not for this PR: `UserRepositoryImpl` is now at roughly eighteen constructor parameters, and `UserRepositoryBulkInsertTest` threads that many positional `mockk(relaxed = true)` arguments through three call sites. The types differ enough that a mis-ordered argument wouldn't compile, so it's safe — but the class is getting large, which is the same pressure #17300 started relieving by segregating `UserAchievementsRepository` out of it.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #17334, opened 2026-09-17
Share ConnectionPool across OkHttp clients in NetworkModule

### LINKED ISSUE #17317: every reachability probe opens a fresh TCP+TLS handshake because the two OkHttp clients do not share a connection pool
## Verified premise

`di/NetworkModule.kt`, `buildOkHttpClient` constructs a new pool on every call (L81):

```kotlin
private fun buildOkHttpClient(connect: Long, read: Long, write: Long, retryInterceptor: RetryInterceptor? = null): OkHttpClient {
    val dispatcher = Dispatcher().apply { maxRequestsPerHost = MAX_REQUESTS_PER_HOST }
    val connectionPool = ConnectionPool(MAX_REQUESTS_PER_HOST, 5, TimeUnit.MINUTES)
    val builder = OkHttpClient.Builder()
        .dispatcher(dispatcher)
        .connectionPool(connectionPool)
        ...
}
```

and it is called twice, for two `@Singleton` clients that talk to the same Planet host:

```kotlin
@Provides @Singleton @StandardHttpClient
fun provideStandardOkHttpClient(retryInterceptor: RetryInterceptor): OkHttpClient =
    buildOkHttpClient(CONNECT_TIMEOUT_SECONDS, READ_TIMEOUT_SECONDS, WRITE_TIMEOUT_SECONDS, retryInterceptor)

@Provides @Singleton @ReachabilityHttpClient
fun provideReachabilityOkHttpClient(): OkHttpClient =
    buildOkHttpClient(REACHABILITY_TIMEOUT_SECONDS, REACHABILITY_TIMEOUT_SECONDS, REACHABILITY_TIMEOUT_SECONDS)
```

So the reachability client can never reuse a connection the sync client already has open to the same server — every probe pays a full TCP plus TLS handshake.

## Do

- Hoist one `ConnectionPool(MAX_REQUESTS_PER_HOST, 5, TimeUnit.MINUTES)` to a single instance and pass it to both clients.

## Constraints

- **Each client must keep its OWN `Dispatcher`. The pool is safe to share; the dispatcher is not — a shared one would queue short reachability probes behind long-running sync traffic against `maxRequestsPerHost = 20`, which is the opposite of what a probe needs.**
- Keep the pool's parameters exactly as they are (`MAX_REQUESTS_PER_HOST`, 5, `TimeUnit.MINUTES`).
- The two clients keep their different timeouts and the standard client keeps its `RetryInterceptor`.
- A shared pool must be created once, not per `@Provides` call — make it a module-level value or its own `@Provides @Singleton`, not something `buildOkHttpClient` constructs.

## Tests

- Existing `app/src/test/java/org/ole/planet/myplanet/di/NetworkModuleTest.kt` (or equivalent) must pass **unedited**.
- Add an assertion that both provided clients return the same `connectionPool` instance and different `dispatcher` instances — one test that pins both halves of the constraint above.

## Out of scope

- The timeout constants and `MAX_REQUESTS_PER_HOST`'s value.
- `RetryInterceptor` and `TaggedSocketFactory`.
- Retrofit wiring and the API interfaces.

dogi:
This change hoists the ConnectionPool in NetworkModule to a shared singleton instance provided via @Provides @Singleton fun provideConnectionPool(): ConnectionPool. Both standard and reachability OkHttpClient instances now share this connection pool while retaining their distinct Dispatcher instances, timeouts, and interceptors.

Fixes #17317

---
*PR created automatically by Jules for task [7756069137385122762](https://jules.google.com/task/7756069137385122762) started by @dogi*


dogi:
**Review — rating: 76/100**

The goal is right and worth doing. Both clients talk to the same CouchDB host, and OkHttp's own guidance is to share a `ConnectionPool` so sockets opened by one client can be reused by the other — previously each `buildOkHttpClient` call constructed its own pool, so the reachability probe could never reuse a connection the standard client had already established. `ConnectionPool` is thread-safe and explicitly designed to be shared, so nothing here is unsafe.

**The DI shape is the problem.** There are two mechanisms doing one job, and the one that's visible is the one Dagger ignores:

```kotlin
private val sharedConnectionPool = ConnectionPool(MAX_REQUESTS_PER_HOST, 5, TimeUnit.MINUTES)

@Provides @Singleton
fun provideConnectionPool(): ConnectionPool = sharedConnectionPool

@Provides @Singleton @StandardHttpClient
fun provideStandardOkHttpClient(
    retryInterceptor: RetryInterceptor,
    connectionPool: ConnectionPool = provideConnectionPool()
): OkHttpClient
```

Dagger does not honour Kotlin default parameter values — the generated factory satisfies *every* parameter from the graph, so `connectionPool` is always injected from the `@Provides` binding and that default is never evaluated in production. It happens to be harmless because both routes return the same object, but it tells a future reader that the parameter is optional when it isn't, and if anyone removed the `ConnectionPool` binding on the strength of the default, the build would break. The default exists only so the test can call the provider with one argument — production API shaped around test convenience.

The `@Provides @Singleton` binding already guarantees a single instance for the whole graph, so the `private val` static isn't buying anything either.

**That combination also weakens the test.** `assertSame(standardClient.connectionPool, reachabilityClient.connectionPool)` calls both providers directly as plain functions, so it takes the default-parameter path rather than the one Dagger uses — and since `sharedConnectionPool` is a static `val`, the assertion would hold even if the `@Singleton` scoping were wrong. It proves a static field is a static field, not that the graph shares the pool.

Dropping the static, keeping `@Provides @Singleton fun provideConnectionPool() = ConnectionPool(...)`, and making `connectionPool` a required parameter fixes all three at once: one mechanism, no misleading default, and a test that has to obtain the pool explicitly and therefore actually means something.

**One thing worth deciding while you're here:** `assertNotSame(standardClient.dispatcher, reachabilityClient.dispatcher)` locks in that each client keeps its own `Dispatcher`, which means `MAX_REQUESTS_PER_HOST = 20` is enforced twice independently — an effective ceiling of 40 concurrent requests to the same host. That's pre-existing rather than introduced here, but the test now enshrines it as intended. If the cap is meant to be global, the dispatcher should be shared too; if it isn't, the constant name is misleading.

**Nit:** `ConnectionPool(MAX_REQUESTS_PER_HOST, 5, TimeUnit.MINUTES)` reuses the per-host request cap as `maxIdleConnections`, which are unrelated quantities. Also pre-existing, but it's now a shared, longer-lived pool so the sizing matters slightly more.

**Verdict:** needs another round — remove the static and the `@Provides` default parameter so the binding is the single source of the shared pool, then let the test assert sharing through the real wiring.

---

dogi:
**Re-review — Rating: 87/100** (was 72)

**Verdict:** mergeable — the pool is now genuinely shared, and the tests prove it rather than asserting around it.

The pool moved out of `buildOkHttpClient` into a `@Provides @Singleton fun provideConnectionPool()`, injected into both client providers. That is the right mechanism: with the pool constructed inside the builder, each `@Singleton` client got its own, so "shared" was only true in the comment. Now the sharing is enforced by the graph, and a third client added later gets it for free by declaring the parameter.

**The tests are the part that makes this stick.** `assertSame(connectionPool, okHttpClient.connectionPool)` is an identity assertion on the thing that was previously wrong — the old `assertNotNull(okHttpClient.connectionPool)` would have passed against the broken version. And the new third test pins both halves of the intended design at once:

```kotlin
assertSame(standardClient.connectionPool, reachabilityClient.connectionPool)
assertNotSame(standardClient.dispatcher, reachabilityClient.dispatcher)
```

Keeping the dispatchers separate is correct and worth having asserted — a reachability probe queued behind sync traffic on a shared dispatcher would defeat its own 5s timeout.

I checked whether sharing a pool across clients with different timeouts is safe: it is. OkHttp timeouts are per-call, not properties of a pooled connection, so the reachability client reusing a connection the standard client established carries none of its timeout state. The saved TLS handshake on a probe is the whole point.

**One nit:** `ConnectionPool(MAX_REQUESTS_PER_HOST, 5, TimeUnit.MINUTES)` reuses the dispatcher's constant for the pool's `maxIdleConnections`. They are unrelated quantities — one caps concurrent in-flight requests per host, the other caps idle connections retained across all hosts. 20 is a fine value for both, so nothing is broken, but a separate `MAX_IDLE_CONNECTIONS` constant would stop a future tuning change to one from silently moving the other.

CI green 5/5.

---

Okuro3499 (approved):

PR #17335, opened 2026-09-17
courses: smoother submissions exporter context providing (fixes #17311)

### LINKED ISSUE #17311: stop threading Context through SubmissionsRepositoryImpl into the PDF exporter
## Verified premise

`repository/SubmissionsRepositoryImpl.kt:50` injects an application Context:

```kotlin
@ApplicationContext private val context: Context,
```

and uses it for exactly two things. First, forwarding it into the exporter (L62, L66):

```kotlin
return exporter.generateSubmissionPdf(context, submissionId)
...
return exporter.generateMultipleSubmissionsPdf(context, submissionIds, examTitle)
```

Second, at L814 and L885:

```kotlin
`object`.addProperty("customDeviceName", NetworkUtils.getCustomDeviceName(context))
jsonObject.addProperty("customDeviceName", NetworkUtils.getCustomDeviceName(context))
```

where `NetworkUtils.getCustomDeviceName` ignores its argument entirely (`NetworkUtils.kt:243-245` returns `sharedPrefManager.getCustomDeviceName()`).

`repository/SubmissionsRepositoryExporter.kt:26` is itself an `@Inject` class that takes the Context as a *method* parameter (L46, L132) and uses it only to resolve an output directory (L112, L229):

```kotlin
val directory = File(context.getExternalFilesDir(Environment.DIRECTORY_DOCUMENTS), "Submissions")
```

## Do

- Add `@ApplicationContext private val context: Context` to `SubmissionsRepositoryExporter`'s constructor and drop the `context` parameter from `generateSubmissionPdf` and `generateMultipleSubmissionsPdf`.
- Inject `DeviceNameProvider` into `SubmissionsRepositoryImpl` and use it at L814 and L885, following the precedent in `ResourcesRepositoryImpl` and `PersonalsRepositoryImpl`.
- Remove `@ApplicationContext private val context: Context` from `SubmissionsRepositoryImpl` **only once nothing else in the file uses it** — grep the whole file first; it is ~900 lines.

## Constraints

- **`SubmissionsRepository.generateSubmissionPdf` / `generateMultipleSubmissionsPdf` are public interface methods with callers outside this file — their signatures must not change. Only the *exporter's* internal method signatures lose the parameter.**
- `SubmissionsRepositoryExporter` is `internal class ... @Inject constructor` and is injected into `SubmissionsRepositoryImpl` rather than bound to an interface; keep it that way.
- Do not change `NetworkUtils.getCustomDeviceName`; its dead parameter has other call sites that are not this task's to fix.
- The output directory (`getExternalFilesDir(DIRECTORY_DOCUMENTS)/Submissions`) and the PDF layout must not change.

## Tests

- Existing `app/src/test/java/org/ole/planet/myplanet/repository/SubmissionsRepositoryImplTest.kt` must pass with **only constructor calls updated**; its assertions stay.
- Add a test asserting the serialized submission carries the `DeviceNameProvider`'s device name.

## Out of scope

- Splitting PDF rendering from the file write into a separate helper — a larger change, not this one.
- The PDF's visual layout, fonts and pagination.
- The other call sites of the dead `Context` parameter elsewhere in the tree.

dogi:
This refactoring removes ApplicationContext from SubmissionsRepositoryImpl by injecting ApplicationContext into SubmissionsRepositoryExporter's constructor, removing the context parameter from generateSubmissionPdf and generateMultipleSubmissionsPdf, and using DeviceNameProvider in SubmissionsRepositoryImpl to retrieve customDeviceName.

Fixes #17311

---
*PR created automatically by Jules for task [2898768491963275934](https://jules.google.com/task/2898768491963275934) started by @dogi*


dogi:
**Review — rating: 89/100**

The nice thing about this one is that it finishes the job rather than relocating it. `SubmissionsRepositoryImpl` held a `Context` for exactly two reasons — passing it to the exporter, and `NetworkUtils.getCustomDeviceName(context)` — and both are removed, so the `@ApplicationContext` parameter and the `android.content.Context` import disappear from the repository entirely. A data-layer repository that no longer needs an Android `Context` is a real improvement, and the fact that the build still passes is itself the proof there was no third usage.

The `Context` lands where it belongs: the exporter genuinely needs one for `PdfDocument`, `Environment` and file paths, so injecting it there instead of threading it through every call signature is correct. Both substitutions are value-identical — `NetworkUtils.getCustomDeviceName(context)` ignores its argument and delegates to `sharedPrefManager.getCustomDeviceName()`, which is exactly what `DeviceNameProvider` does, and the exporter's injected `@ApplicationContext` is the same instance the repository was handing it.

The two new tests are the right ones: rather than verifying a call, they assert `customDeviceName` in the payload returned by `serializeSubmission` and `getExamUploadPayload` — the two production sites this PR touches, and the values that actually reach CouchDB. The thirteen mechanical stub updates from `NetworkUtils.getCustomDeviceName(any())` to `deviceNameProvider.getCustomDeviceName()` are all consistent.

Also worth saying, since I've raised the opposite on two sibling PRs today: this test class already does `@After fun tearDown() { unmockkAll() }`, so its `mockkObject(NetworkUtils)` calls are unwound unconditionally even when an assertion throws. That's the pattern #17304 and #17333 should copy.

**Nit:** the exporter uses `@param:ApplicationContext` while the repository it was extracted from used plain `@ApplicationContext`; both resolve the same for constructor injection, but the file's neighbours use the shorter form. And `deviceNameProvider` is appended after the `@PlainGson` qualified parameter — grouping it with the other `utils` dependencies would read more naturally.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #17336, opened 2026-09-17
life: smoother health examination display name resolving (fixes #17330)

### LINKED ISSUE #17330: health examination list runs a regex and two array copies to take one substring
## Verified premise

`ui/health/HealthExaminationAdapter.kt:69` derives a fallback display name the expensive way:

```kotlin
model?.getFullName() ?: createdBy.split(colonRegex).dropLastWhile { it.isEmpty() }.toTypedArray().getOrNull(1) ?: createdBy
```

That is a `Regex` split, a `dropLastWhile` copy and a `toTypedArray()` copy — three allocations plus regex machinery — to read the text after the first `:`. `colonRegex` exists solely for this one call site (L176):

```kotlin
private val colonRegex by lazy { ":".toRegex() }
```

Separately, L104:

```kotlin
binding.txtDate.text = context.getString(R.string.two_strings, item.formattedDate, item.resolvedName).trimIndent()
```

and `values/strings.xml:1028` is:

```xml
<string name="two_strings">%1$s %2$s</string>
```

a single-line format, so `trimIndent()` splits the string into lines, computes a common indent, and rejoins it into exactly the same string — on every bind.

## Do

- Replace the split chain with `createdBy.substringAfter(':', "").takeIf { it.isNotBlank() } ?: createdBy`.
- Delete `colonRegex` from the companion object.
- Drop `.trimIndent()` from L104.

## Constraints

- **Leave the `.trimIndent()` at L140 alone.** It formats `R.string.vitals_format`, which is genuinely multi-line, and removing it would change the rendered vitals block.
- `substringAfter(':', "")` returns the default `""` when there is no colon, where the old chain returned null from `getOrNull(1)` — both then fall through to `createdBy` via the `?:`, but only if you keep the `takeIf { it.isNotBlank() }`. Do not drop it.
- A `createdBy` of the form `"org.couchdb.user:name"` must still yield `name`. Check whether any value contains **more than one** colon before assuming `substringAfter` (first colon) matches `getOrNull(1)` (second segment) — they agree only for a single-colon string.
- `displayNameCache.getOrPut(createdBy) { ... }` stays; this is about what runs on a cache miss.
- The `model?.getFullName()` preference stays first.

## Tests

- Existing `app/src/test/java/org/ole/planet/myplanet/ui/health/` tests must pass **unedited**.
- Add cases for the fallback: `"org.couchdb.user:alice"` → `alice`; a `createdBy` with no colon → the whole string; a trailing-colon value → the whole string, not an empty label.

## Out of scope

- `getFullName()` and the user map lookup.
- `vitals_format` and the examination detail dialog.
- The DiffUtil callback.

dogi:
Optimized fallback display name calculation in HealthExaminationAdapter by replacing regex split and array copies with substringAfter(':', ""), removing the unused colonRegex property, and dropping redundant .trimIndent() call. Added unit tests verifying fallback name resolution.

Fixes #17330

---
*PR created automatically by Jules for task [8355497442901869440](https://jules.google.com/task/8355497442901869440) started by @dogi*


dogi:
**Review — rating: 85/100**

`split(regex).dropLastWhile { it.isEmpty() }.toTypedArray().getOrNull(1)` really was three intermediate collections to extract one substring, and `substringAfter(':', "")` replaces it cleanly — dropping the `colonRegex` companion val along with it. The `.trimIndent()` removal is right too: on a single-line runtime-formatted string it never had anything to trim, and it was allocating on every `onBindViewHolder`.

I worked through the equivalence case by case, and the two forms agree on everything your tests cover:

| `createdBy` | old | new |
|---|---|---|
| `org.couchdb.user:alice` | `alice` | `alice` |
| `alice` | `alice` | `alice` |
| `org.couchdb.user:` | whole string | whole string |

**But they diverge on two inputs, and neither is tested:**

- **More than one colon.** `a:b:c` — the old code took index 1 and yielded `b`; `substringAfter` takes everything past the *first* colon and yields `b:c`. Since a CouchDB `_users` `_id` is `org.couchdb.user:` + name and the name may itself contain a colon, the new behaviour is the more correct one — but it is a behaviour change.
- **Whitespace-only remainder.** `org.couchdb.user: ` — the old `dropLastWhile { it.isEmpty() }` doesn't drop `" "` (it's not empty), so index 1 returned `" "` and the row displayed a blank name. The new `takeIf { it.isNotBlank() }` rejects it and falls back to the whole string. Also an improvement.

Both changes are in the right direction, which is why I'm not blocking — but the two cases where old and new actually differ are exactly the ones missing from the four new tests. Adding them locks in the intended behaviour instead of leaving it incidental.

The test file itself is good: building the payload through a real `AndroidDecrypter.encrypt` round-trip exercises the genuine decrypt path rather than stubbing it, and the `userInMap_prefersModelFullNameFirst` case pins the precedence of `model?.getFullName()` over the colon fallback.

**Nit on the framing:** the name resolution sits inside `displayNameCache.getOrPut`, so it only runs once per distinct `createdBy`, not per row. The allocation saving is real but bounded by the number of distinct authors, not by list size — the `trimIndent()` removal is the part that's actually per-bind.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #17339, opened 2026-09-17
teams: smoother calendar teams distinct view modelling (fixes #17329)

### LINKED ISSUE #17329: CalendarViewModel re-queries meetups on every teams emission, even when the team list is unchanged
## Verified premise

`ui/calendar/CalendarViewModel.kt`, `loadMeetups`:

```kotlin
private fun loadMeetups() {
    viewModelScope.launch {
        val userId = userRepository.getUserModel()?.id ?: return@launch
        teamsRepository.getMyTeamsFlow(userId).collect { teams ->
            _teamNames.value = teams.associate { it._id to it.name.orEmpty() }
            _meetups.value = eventsRepository.getMeetupsForTeams(teams.map { it._id })
        }
    }
}
```

`getMyTeamsFlow` is a Room `Flow`, so it re-emits on **any** write to the teams table — including writes that leave this user's team list identical. Each emission rebuilds the name map and fires a fresh `getMeetupsForTeams` database query. During a sync that touches teams, that is a burst of redundant reads behind the calendar screen.

The list is also walked twice per emission: once for `associate`, once for `map { it._id }`.

## Do

- Apply `distinctUntilChanged()` to the collected flow so an unchanged team list does not re-trigger the body.
- Derive the ids once into a local and feed both outputs from it:
  ```kotlin
  teamsRepository.getMyTeamsFlow(userId).distinctUntilChanged().collect { teams ->
      _teamNames.value = teams.associate { it._id to it.name.orEmpty() }
      _meetups.value = eventsRepository.getMeetupsForTeams(teams.map { it._id })
  }
  ```
  (hoist `teams.map { it._id }` into a `val` if you also want the single walk).

## Constraints

- **`distinctUntilChanged()` on a `List<Team>` compares with `equals`, so it only suppresses an emission when the list elements themselves are equal. Check whether `Team` is a `data class` or an unequal-by-default Room `@Entity` — if it has no meaningful `equals`, comparing whole entities will never dedupe and you must key on what this screen actually uses (`distinctUntilChangedBy { it.map { t -> t._id to t.name } }`) instead. Getting this wrong makes the change a silent no-op.**
- Keep the public `meetups` and `teamNames` `StateFlow` API exactly as it is.
- Keep the `?: return@launch` early exit for a null user id.
- Do not extract a loader class or restructure the three-repository fan-out — that is a separate, larger change.

## Tests

- Existing `app/src/test/java/org/ole/planet/myplanet/ui/calendar/` tests must pass **unedited**.
- Add a test emitting the same team list twice and asserting `eventsRepository.getMeetupsForTeams` is called **once** — the test that proves the dedupe actually works rather than compiling.

## Out of scope

- `getMyTeamsFlow` and `getMeetupsForTeams` themselves.
- The calendar rendering and `CalendarFragment`.
- Any other ViewModel collecting the same flow.

dogi:
Updated `CalendarViewModel.kt` to deduplicate team emissions via `distinctUntilChangedBy` and single-walk team ID mapping. Added unit test in `CalendarViewModelTest.kt` verifying deduplication behavior when duplicate team list emissions occur.

Fixes #17329

---
*PR created automatically by Jules for task [4414662268214604001](https://jules.google.com/task/4414662268214604001) started by @dogi*


dogi:
**Review — rating: 86/100**

Right fix for a real problem. `getMyTeamsFlow` is a Room flow, so it re-emits on any write that touches its tables regardless of whether the query result changed — and every one of those re-emissions was firing a fresh `getMeetupsForTeams` database round trip.

The selector is chosen precisely, which is the part that matters here. `teams.map { t -> t._id to t.name }` covers exactly the two fields the collector consumes — `_id` and `name` for `_teamNames`, `_id` for the meetups query — so it can't suppress an emission that would have changed the output, and it doesn't include fields that would cause needless wake-ups. `List` and `Pair` both compare structurally, so the key works by content rather than identity.

Where it's imprecise, it errs safely: the selector compares raw `name` while the body applies `.orEmpty()`, so a null → `""` transition counts as a change and re-emits redundantly rather than being swallowed. Same for list ordering.

The test is well targeted — two structurally equal but distinct list instances, then `coVerify(exactly = 1)`. That fails at 2 without the operator, and using separate instances proves the comparison is structural rather than referential.

**One consequence worth being deliberate about:** meetups are fetched imperatively inside the collector rather than observed as their own flow, so previously a new meetup would show up incidentally whenever *any* team-table write happened to re-trigger the collect. With the dedup in place that accidental refresh path is gone — the calendar will now only re-query meetups when the team set or a team name actually changes. The old behaviour was unreliable rather than correct, so I don't think this blocks, but if the calendar is expected to reflect newly added meetups live, that wants its own flow from `EventsRepository` rather than riding on team emissions.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #17340, opened 2026-09-17
Cache course cover art file existence check in CoursesItemUtils

### LINKED ISSUE #17316: course cover art stats the disk on the main thread once per row bind
## Verified premise

`utils/CoursesItemUtils.kt`, `bindCover` (L49 onward), runs a synchronous filesystem stat while deciding what to load:

```kotlin
val coverFile = MyCourse.getCoverImageFile(context, course.courseId, course.coverFileName)
val model: Any? = if (coverFile?.exists() == true) {
    coverFile
} else {
    UrlUtils.getCourseImageUrl(course.courseId, course.coverFileName)?.let { url -> ... }
}
```

`bindCover` is called from three bind paths, all on the main thread:

- `ui/courses/CoursesAdapter.kt:330`
- `ui/courses/CoursesAdapter.kt:378`
- `ui/teams/courses/TeamCoursesAdapter.kt:45`

so scrolling a course list issues one `File.exists()` syscall per visible row per rebind.

The cache for exactly this already exists — `utils/FileUtils.kt:381`:

```kotlin
class FileExistenceCache(private val ttlMs: Long = DEFAULT_TTL_MS) {
    private val cache = HashMap<String, Pair<Boolean, Long>>()
    fun clear() { cache.clear() }
    fun exists(file: File?, now: Long): Boolean { ... }
}
```

already used by `EnterprisesReportsAdapter`.

## Do

- Give `CoursesItemUtils` a private `FileExistenceCache` and a `TimeProvider`, and replace `coverFile?.exists() == true` with `coverExistenceCache.exists(coverFile, timeProvider.now())`.

## Constraints

- **Keep `bindCover`'s signature exactly as it is — three call sites pass the same seven positional arguments and must compile untouched.** If `CoursesItemUtils` is an `object` and cannot be injected, get the time from whatever clock abstraction the file can already reach rather than adding a parameter; do not hard-code `System.currentTimeMillis()`.
- `FileExistenceCache.exists(null, now)` already returns false, so the `coverFile?` null case needs no extra guard.
- The 5-second default TTL is the intended behaviour: a cover that finishes downloading appears within 5 s or on the next reload. Do not shorten it to "fix" that.
- The cache is a plain `HashMap` with no synchronisation — only call it from the main thread, which is where all three bind paths already run. Do not move the check onto a background dispatcher.
- Do not change the Glide model fallback or the `UrlUtils.getCourseImageUrl` branch.

## Tests

- Existing `app/src/test/java/org/ole/planet/myplanet/utils/` and courses-adapter tests must pass **unedited**.
- Add a test asserting two `bindCover`-style lookups for the same path within the TTL produce one `exists()` call, and that a lookup after the TTL re-stats.

## Out of scope

- `ResourcesAdapter.bindCover` — a different, private method with the same name (L312).
- `MyCourse.getCoverImageFile` and the cover download path.
- `setCoverColor` and the subject-icon binding.

dogi:
This change updates CoursesItemUtils to use FileExistenceCache and TimeProvider for checking cover image file existence in bindCover. By caching the existence result within the default 5s TTL, redundant File.exists() disk stat calls on the main thread during row binding and list scrolling are avoided. Signature of bindCover remains unchanged.

Fixes #17316

---
*PR created automatically by Jules for task [9191557838494914448](https://jules.google.com/task/9191557838494914448) started by @dogi*

## Summary by CodeRabbit

* **Performance**
  * Improved course cover loading by caching local cover-file availability checks for a short period.
  * Reduced repeated disk checks while preserving fallback behavior when a local cover is unavailable.

* **Tests**
  * Added coverage verifying cached checks refresh after the cache period expires.


dogi:
**Review — rating: 76/100**

The caching itself is correct and reuses existing infrastructure rather than inventing any — `FileExistenceCache` already lives in `FileUtils` with a 5s default TTL, and crucially its `exists(file: File?, now: Long)` returns `false` for a null file, which is exactly what `coverFile?.exists() == true` did. So the substitution preserves semantics including the null case, and `bindCover` stops re-`stat`ing the same cover on every row bind.

The test is also well constructed *in substance*: spying the `File` and stepping a controllable clock across the TTL boundary (1000 → 2000 → 6001) to assert one `exists()` call inside the window and two after it is precisely the behaviour a TTL cache should be pinned on. Using `mockkObject(MyCourse)` with a matching `unmockkObject` in `@After` is the right hygiene too.

**The seam it needs to do that is the problem.**

```kotlin
internal object CoursesItemUtils {
    private val coverExistenceCache = FileExistenceCache()
    var timeProvider: TimeProvider = SystemTimeProvider()
```

That's a publicly mutable clock on a process-wide singleton, existing only so a test can swap it — and the test then sets `CoursesItemUtils.timeProvider = testTimeProvider` in `@Before` and never restores it. `tearDown` only unmocks `MyCourse`. So once this class runs, the global clock stays frozen at 6001L for every later test in the same JVM that touches `bindCover`, and `coverExistenceCache` keeps its entry for the temp-file path with no `clear()` — even though `FileExistenceCache` exposes one for exactly this purpose. Today only this test calls `bindCover`, so nothing breaks yet; it's a landmine rather than a live failure, and the failure it eventually causes will look unrelated to this file.

Two ways out, either fine:

- **Minimal:** restore the previous `timeProvider` and call `coverExistenceCache.clear()` in `@After`. That needs the cache exposed or a reset hook, which is itself a hint the design is inside-out.
- **Better:** have `bindCover` take `now: Long` as a parameter — `FileExistenceCache.exists` already takes the timestamp, so the object needs no clock of its own, callers pass `System.currentTimeMillis()`, and the test passes whatever it likes with no global state at all. That also removes the `var` from production code entirely.

**One thing worth confirming:** `FileExistenceCache` is backed by a plain `HashMap`, and `coverExistenceCache` is a single instance on an `object`, so it's shared process-wide. That's fine if `bindCover` is only ever called from the main thread during bind — which it appears to be, since it isn't `suspend` — but it's worth being sure, because other adapters in this codebase do their cover binding from coroutines.

**Verdict:** needs another round — take the timestamp as a parameter instead of a mutable global clock (or at minimum restore it and clear the cache in `@After`).

---

dogi:
**Re-review — Rating: 87/100** (was 71)

**Verdict:** mergeable — the test-hygiene gap I raised is closed, and the new test is a genuinely good one.

`resetForTesting()` clears both the cache and the injected `timeProvider`, and is called from **both** `@Before` and `@After` — which is the right belt-and-braces for state living on an `object`, since a test that fails partway still leaves the next one clean.

**The new test discriminates properly, which is the part that matters.** Using a `spyk(testFile)` and counting `exists()` calls across three binds — 1 call, then still 1 within the TTL, then 2 after crossing it — it verifies both halves of the cache contract: that a hit avoids the stat, and that expiry actually re-stats. Driving that with `TestTimeProvider` rather than real sleeps makes it deterministic. This is the shape the other caching PRs in this series should follow; compare #17396, where both fixture users shared a `joinDate` so the assertion couldn't tell a keyed cache from an unkeyed one.

Two small things, neither blocking:

**`timeProvider` is a public `var` on a singleton with no annotation.** `resetForTesting()` carries `@VisibleForTesting(otherwise = PRIVATE)`, but the seam it resets does not — so production code can reassign the clock of a process-global object and nothing flags it. Marking it `@VisibleForTesting` too, or making it private with a test-only setter, closes that.

**Misplaced import:** `androidx.annotation.VisibleForTesting` landed after the `com.bumptech.glide.*` block; `androidx` sorts before `com`. Same slip as #17412, #17414, #17433 and #17351 in this series — worth one optimise-imports pass across the branch rather than fixing them individually.

The cache itself is sound: 5s TTL bounds staleness, `FileExistenceCache` is the existing shared helper rather than another hand-rolled map, and the `object` lifetime means the cache survives adapter churn during scrolling, which is the point.

CI green 5/5.

---

Okuro3499 (approved):

PR #17349, opened 2026-09-17
courses: smoother adapter course removal filtering (fixes #17325)

### LINKED ISSUE #17325: CoursesAdapter.removeCourses does a linear scan of the removal list per row
## Verified premise

`ui/courses/CoursesAdapter.kt`, `removeCourses` (L119-124):

```kotlin
fun removeCourses(courseIds: List<String>, onComplete: (() -> Unit)? = null) {
    val updated = currentList.filter { it.courseId !in courseIds }
    submitList(updated) {
        onComplete?.invoke()
    }
}
```

`courseIds` is a `List<String>`, so `it.courseId !in courseIds` compiles to `List.contains` — a linear scan. Run once per item in `currentList`, that is O(n·m): with a few hundred courses and a large removal selection it is tens of thousands of string comparisons on the main thread during a list update.

## Do

- Convert `courseIds` to a `Set` once before filtering, and filter against the set.

## Constraints

- **`onComplete` must keep firing from `submitList`'s commit callback and nowhere else. It signals that the list has actually been committed, so an empty-`courseIds` short-circuit that calls `onComplete()` directly changes the timing that callers depend on — route the empty case through the same `submitList` path.**
- `currentList` order must be preserved; `filter` already does this, so keep using it rather than set arithmetic that would reorder.
- Do not change `removeCourses`'s signature — it keeps taking `List<String>` so callers are untouched; the `Set` conversion is internal.
- Build the set once, outside the `filter` lambda.

## Tests

- Existing `app/src/test/java/org/ole/planet/myplanet/ui/courses/CoursesAdapterTest.kt` must pass **unedited**.
- Add a test asserting `onComplete` fires exactly once for an empty `courseIds` list, and one asserting the surviving items keep their original order after a partial removal.

## Out of scope

- `setProgressMap`, `updateIdentity` and the other adapter methods.
- The DiffUtil callback.
- `TeamCoursesAdapter`.

dogi:
Optimized CoursesAdapter.removeCourses to convert courseIds into a Set before filtering, reducing complexity from O(n*m) to O(n). Added unit tests in CoursesAdapterTest asserting callback timing for empty courseIds and order preservation after partial removal.

Fixes #17325

---
*PR created automatically by Jules for task [8471166785920384451](https://jules.google.com/task/8471166785920384451) started by @dogi*


dogi:
**Review — rating: 82/100**

Correct and exactly equivalent. `List.contains` and `Set.contains` both dispatch through `equals`, `toSet()` collapsing duplicate ids doesn't change any membership result, and a null `courseId` is handled the same by both, so this is a pure O(n·m) → O(n+m) change with no behavioural difference. Hoisting the conversion out of the `filter` predicate is the right way to do it — building the set inside the lambda would have been worse than the original.

Worth being honest about the scale, though: `removeCourses` runs on a user-initiated deletion with a course list in the tens-to-low-hundreds, so the old form was maybe a few thousand string comparisons. This is the right way to write it and costs nothing, but it isn't fixing a felt problem.

The two new tests don't target the change — both pass identically before and after — but they're useful regression coverage in their own right: that `onComplete` fires exactly once (rather than being dropped or double-invoked by `AsyncListDiffer`), and that surviving items keep their relative order after a partial removal.

**The thing I'd fix before merging is the spin loop:**

```kotlin
while (!removeCommitted) {
    org.robolectric.shadows.ShadowLooper.idleMainLooper()
}
```

It appears four times, and if the commit callback ever stops firing — a real regression, or a future `AsyncListDiffer` behaviour change — this doesn't fail the test, it **hangs the build**. A test that never terminates is considerably more painful to diagnose than one that reports a failed assertion, and on CI it burns the job timeout. Idling the looper once and then asserting, or bounding the loop with an iteration cap before failing, gives you the same coverage with a diagnosable failure mode.

**Nit:** `org.robolectric.shadows.ShadowLooper` is written as an inline fully-qualified name at all four sites where an import would do.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #17351, opened 2026-09-17
Use projection query for achievement resource picker

### LINKED ISSUE #17297: load only id and title for the achievement resource picker instead of every 74-property library row
## Verified premise

`ui/user/EditAchievementFragment.kt:413` populates a checkbox dialog by materialising every library row:

```kotlin
val list = viewModel.getAllLibraries()
```

which goes to `ui/user/AchievementViewModel.kt:85-87`:

```kotlin
suspend fun getAllLibraries(): List<MyLibrary> {
    return resourcesRepository.getAllLibraries()
}
```

an unbounded `SELECT *` over `MyLibrary` — ~74 properties per row, several going through `Converters`. The dialog reads exactly one field per row (L493, `names.add(list[i].title ?: "")`); the full entity is needed only for the rows the user confirms, at L427 (`resourceArray?.add(list[ii].serializeResource())`).

`getAllLibraries` has this one consumer — verified: `ResourcesRepository.kt:39`, `ResourcesRepositoryImpl.kt:77`, `AchievementViewModel.kt:86`, `EditAchievementFragment.kt:413` and nothing else. It is distinct from `getAllLibrariesToSync`, which has its own callers.

The projection pattern already exists in this DAO — `data/room/dao/MyLibraryDao.kt:162-163`:

```kotlin
@Query("SELECT resourceId, title FROM my_library WHERE resourceId IS NOT NULL")
suspend fun getResourceTitles(): List<ResourceTitleProjection>
```

## Do

- Add a projection query beside it selecting `id, title` from `my_library` (no `WHERE`, no `ORDER BY`), with its own small data class.
- Replace `getAllLibraries` on `ResourcesRepository` / `ResourcesRepositoryImpl` with an accessor returning that projection, and update `AchievementViewModel` to match.
- In `EditAchievementFragment`, build the checkbox list from the projection, and refetch the full entities for the confirmed ids through the existing `resourcesRepository.getLibraryItemsByIds(ids)` before calling `serializeResource()`.

## Constraints

- **The projection must carry `id`, not `resourceId`. `getLibraryItemsByIds(ids: Collection<String>)` (`ResourcesRepositoryImpl.kt:185`) keys on the `id` primary key, while the neighbouring `ResourceTitleProjection` carries `resourceId` — reusing that existing class would hand the refetch the wrong key and silently return nothing.**
- Preserve the current unspecified row order: `getAllLibraries` has no `ORDER BY`, so add none — the list must look identical.
- `serializeResource()` stays on `MyLibrary` and is called only on refetched entities.
- Delete `getAllLibraries`; do not leave it behind unused.
- No schema or index change, therefore **no `AppDatabase` version bump**.

## Tests

- Existing `app/src/test/java/org/ole/planet/myplanet/repository/ResourcesRepositoryImplTest.kt` must pass **unedited** apart from the `getAllLibraries` case, which retargets to the new accessor.
- Add a DAO test asserting the projection returns one entry per row with the right `id`/`title`, and a test that `getLibraryItemsByIds` round-trips the projection's ids back to full entities.

## Out of scope

- `getAllLibrariesToSync` and its callers (`MainApplication`, `SyncActivity`).
- `getResourceTitles` / `ResourceTitleProjection`.
- The achievement save path and `serializeResource()`'s output.

dogi:
- Added `LibraryTitleProjection` data class and `getLibraryTitles()` projection query in `MyLibraryDao`.
- Removed `getAllLibraries()` from `ResourcesRepository`, `ResourcesRepositoryImpl`, and `AchievementViewModel`, replacing it with `getLibraryTitles()`.
- Updated `EditAchievementFragment` to build the resource selection dialog using `LibraryTitleProjection` and refetch full entities by ID via `getLibraryItemsByIds` upon user confirmation.
- Updated repository/ViewModel unit tests and added `MyLibraryDaoTest` verifying projection loading and full entity refetching.

Fixes #17297

---
*PR created automatically by Jules for task [11497048563194361341](https://jules.google.com/task/11497048563194361341) started by @dogi*


dogi:
**Review — rating: 62/100**

The idea is right — the picker dialog only renders titles, so loading full `MyLibrary` entities (and running the `Converters` Gson pass over `subject`, `level` and `attachments` for every row) just to read `.title` is genuinely wasteful, and a projection is the correct fix. But the round trip back to full entities is wired to the wrong column.

**The projection selects `id`; the lookup queries `_id`.**

```kotlin
// new projection
@Query("SELECT id, title FROM my_library")          // LibraryTitleProjection.id  == primary key `id`

// what getLibraryItemsByIds resolves to
ResourcesRepositoryImpl:187  →  myLibraryDao.getByUnderscoreIds(ids)
@Query("SELECT * FROM my_library WHERE _id IN (:ids)")
```

`id` and `_id` are separate columns on `MyLibrary` — the DAO carries `getByIds` (`WHERE id IN`) and `getByUnderscoreIds` (`WHERE _id IN`) precisely because they differ, and `_id` is nullable while `id` is the primary key. So `selectedIds` comes from one column and is matched against another. Wherever a row has `id != _id`, or a null `_id`, the lookup returns nothing and the user's selected resource is **silently dropped** from the achievement — no error, just a shorter `resourceArray`. `getByIds` is the method that matches this projection.

**The new DAO test hides exactly this.** `getLibraryItemsByIds_roundTripsProjectionIdsBackToFullEntities` sets `id = "pk1"` *and* `_id = "pk1"` on both fixtures, so the two columns are indistinguishable, and it calls `myLibraryDao.getByUnderscoreIds(...)` directly rather than going through `repository.getLibraryItemsByIds`. A fixture where `id != _id`, exercised through the repository, would have caught this immediately.

**Second problem: the OK handler became asynchronous.**

```kotlin
builder.setPositiveButton("Ok") { _, _ ->
    val selectedIds = items.map { list[it].id }
        val fullLibraries = viewModel.getLibraryItemsByIds(selectedIds)
        resourceArray = JsonArray()
        ...
    }
}
```

Previously `resourceArray` was populated synchronously before the handler returned. Now the dialog dismisses immediately while both the reset *and* the fill happen later on a coroutine, so anything reading `resourceArray` between dismissal and completion sees the stale value — and if the user taps save quickly, the attachments are lost. If the fetch has to be async, the save path needs to await it (or the dialog should stay up / the control disable until it lands).

**Smaller, related:** `getByUnderscoreIds` returns rows in database order, not in `selectedIds` order, so `resourceArray` no longer follows the user's selection order the way the old index loop did.

**Nit / scope:** the change to `assertEquals(setOf(...), res1Item.filePaths.toSet())` in `getOfflineResourceItems calculates size and path order in single pass` is unrelated to this PR — and notably #17288 weakens the *same* assertion in the same test a different way (`.sorted()` on both sides). Two independent PRs loosening one order assertion suggests `getOfflineResourceItems` really is order-nondeterministic and the production code should sort, rather than each PR relaxing the test. Worth resolving once, in its own change.

**Verdict:** needs another round — switch the lookup to `getByIds` so it matches the projected column, make the DAO test use `id != _id` and go through the repository, and keep `resourceArray` populated before the dialog's result is observable.

---

dogi:

**Verdict:** the bug is fixed and the test is exactly right — but the fix changes a shared method, and one of its other two callers needs a decision before this merges.

**The original finding is properly closed.** `getLibraryTitles()` projects `id`; `getLibraryItemsByIds` now routes to `myLibraryDao.getByIds` instead of `getByUnderscoreIds`, so the round trip queries the same column the projection produced. No more silent drop of every selected resource.

**And the regression test is the one I'd have written.** `getLibraryItemsByIds_roundTripsProjectionIdsBackToFullEntitiesWhenIdDiffersFromUnderscoreId` deliberately sets `id = "pk1"; _id = "doc1"` so the two columns disagree, then round-trips projection ids through `getByIds`. That test fails against the old routing and passes against the new one — it pins the actual defect rather than the happy path. The in-memory Room setup is the right level for it too.

Two things beyond what I asked for, both good: `fetchResourcesJob?.join()` before `saveAchievement` closes the race where the dialog could be submitted before the refetch completed, and iterating `selectedIds` against `libMap` rather than the fetched list preserves selection order.

## The one thing to resolve

`getLibraryItemsByIds` is not private to the achievement picker — it has **two other production callers**, and both just silently changed which column they match on:

**`UploadManager:170` — this PR fixes it too.** It does `libraries.associateBy { it.id }` then `libMap[item.localId]`. Under the old `_id` query, an entity whose `id != _id` was fetched by `_id` but keyed by `id`, so the lookup missed and the attachment was skipped. Now the query and the map agree. Worth knowing you fixed a second instance of the same bug.

**`VoicesViewModel:235` — this one needs checking.** Its ids come from the news images JSON:

```kotlin
val resourceId = JsonUtils.getString("resourceId", ob?.asJsonObject)
if (!resourceId.isNullOrBlank()) { resourceIds.add(resourceId) }
…
val libraries = resourcesRepository.getLibraryItemsByIds(resourceIds)
```

Those are `resourceId` values, and `MyLibrary` carries `id`, `_id` and `resourceId` as three distinct fields — there is even a dedicated `getLibraryItemsByResourceIds` sitting directly below the method this PR edited. So this call looks mismatched both before and after; what changes is *which* wrong column it misses against. If `resourceId` values happen to coincide with `_id` in practice, this PR turns a working path into a broken one and voice image attachments stop downloading — silently, because the result just comes back short and `downloadResources` gets a partial list.

I can't settle that from the code alone, and it is a one-lookup question for you: confirm what `resourceId` holds relative to `id`/`_id`. If it matches neither, switching that call to `getLibraryItemsByResourceIds` is the right fix and belongs here, since this PR is what moved the semantics. If the current call is correct, a one-line comment saying so would save the next reader the same investigation.

CI green 5/5.

**Nits:** `private var fetchResourcesJob: kotlinx.coroutines.Job? = null` uses an inline fully-qualified name — `docs/CODE_STYLE_GUIDE.md` asks for an import. And `data.room.dao.LibraryTitleProjection` landed in the middle of the `databinding` import block (`data` sorts before `databinding`, so it belongs above them) — the same import-placement slip showing up across this whole series.

---

dogi:

**Verdict:** `ready`. My open question is answered, both nits are fixed, and the unrelated test change is gone. One low-severity gap in the new `VoicesViewModel` lookup that I'd fix but wouldn't block on.

## What's closed

**The `id`/`_id` fix holds, and for a structural reason.** `getLibraryTitles()` projects `id`, `getLibraryItemsByIds` queries `getByIds` (`WHERE id IN`) — projection and lookup name the same column, so the round trip is correct by construction regardless of whether the columns ever diverge in practice. Worth saying because `insertMyLibrary` actually sets `id`, `_id` and `resourceId` all to the CouchDB `_id`, so the test fixture's `id = "pk1"; _id = "doc1"` isn't a shape production produces today. The fix doesn't depend on that — but the test is still the right test, since it pins the routing rather than the data.

**Verified the picker itself didn't change behaviour:**
- `getAll()` was `SELECT * FROM my_library` with no `WHERE`, so `SELECT id, title FROM my_library` covers the same row set — no items appear or disappear from the dialog.
- `createResourceList` preselects with `prevList.contains(list[i].title)` — title only, which the projection still carries. The type change from `MyLibrary` to `LibraryTitleProjection` can't break preselection.
- `LibraryTitleProjection.id: String` is non-null and `MyLibrary.id` is `@PrimaryKey var id: String = ""`, so the projection can't take a null.

**Both nits fixed** — `Job` is imported rather than inline-qualified, and `LibraryTitleProjection` now sorts above the `databinding` block.

**The unrelated `filePaths` assertion change is gone from the diff.** Right call — #17288 sorts the production output (`accumulator.filePaths.sort()`), which is where that belonged.

## The one thing left

`VoicesViewModel:235` now uses `getLibraryItemsByResourceIds`, and the commit message frames that as the answer. It's the better of the two options, but it's still only half of what the codebase's own lookup for this exact value does:

```kotlin
// ResourcesRepositoryImpl:180 — the single-item path for news image resourceIds
override suspend fun getLibraryItemByResourceId(resourceId: String): MyLibrary? {
    return myLibraryDao.getByResourceId(resourceId)
        ?: myLibraryDao.getByUnderscoreId(resourceId)   // ← the fallback
}
```

That's what `VoicesAdapter.loadLibraryImage` resolves through, on the same `resourceId` field from the same news `images` JSON. It tries `resourceId` **and falls back to `_id`** — the codebase does not trust either column alone here. The new batch call is `WHERE resourceId IN` with no fallback, so it covers the first half only.

Tracing where the value comes from explains the caution: `UploadManager:304` writes it as `getString("id", imageResponse)`, i.e. the doc id CouchDB returns from `POST /resources`. That lands in `_id` for certain and in `resourceId` only because `insertMyLibrary` happens to assign all three from the same string.

Consequence is mild, which is why this isn't blocking: `downloadReferencedResources` is a prefetch, so a miss means the image isn't pre-downloaded, not that it fails to render — the display path still has the fallback. But a batch method mirroring `getLibraryItemByResourceId` (match `resourceId`, then `_id` for the ids that came back empty) would make the two paths agree, and this PR is what put them out of step.

CI green 5/5.

**Nit:** `fetchResourcesJob` is reassigned without cancelling the previous one, so opening the picker twice in quick succession leaves two coroutines racing on `resourceArray` — each does `resourceArray = JsonArray()` then fills, so the slower one wins regardless of which selection was last. `fetchResourcesJob?.cancel()` before the `launch` closes it. Unlikely to bite given the query is local, but it's one line.

---

Okuro3499 (approved):

PR #17353, opened 2026-09-17
resources: smoother collections switch flag handling (fixes #17344)

### LINKED ISSUE #17344: move the collection-switch flag off MainApplication into CollectionsFragment
## Verified premise

`MainApplication.kt:114` holds a piece of one screen's UI state as a process-wide mutable static:

```kotlin
var isCollectionSwitchOn = false
```

It has exactly one reader and one writer, both in the same fragment:

```kotlin
// ui/resources/CollectionsFragment.kt:162
val isSelectMultiple = MainApplication.isCollectionSwitchOn
// ui/resources/CollectionsFragment.kt:233
MainApplication.isCollectionSwitchOn = b
```

So a checkbox mode lives on the Application object, survives the fragment it belongs to, and is shared across every instance of it.

## Do

- Add a private field to `CollectionsFragment` holding the flag.
- Point L162 and L233 at it.
- Delete `isCollectionSwitchOn` from `MainApplication`.

## Constraints

- **Grep the whole tree for `isCollectionSwitchOn` before deleting the static — the two call sites above are what I found, but confirm there is no third reader (including in the `lite` source set, which is easy to miss) before removing the declaration.**
- No behaviour is lost by scoping it to the fragment: the static already resets to `false` on process death, so nothing persisted it across launches. Do **not** add `SharedPreferences` persistence or `onSaveInstanceState` handling as part of this — that would be a new feature, not this cleanup.
- If `CollectionsFragment` already owns a ViewModel, the field may live there instead; if it does not, a plain private field is correct and you should not introduce a ViewModel for one boolean.
- Leave the other statics on `MainApplication` (`syncFailedCount`, `showDownload`, `isSyncRunning`) alone.

## Tests

- Existing `app/src/test/java/org/ole/planet/myplanet/ui/resources/` tests must pass **unedited**.
- Add a test asserting two separate `CollectionsFragment` instances no longer share the flag, if the existing test setup makes that reachable; otherwise assert the default is `false` on a fresh instance.

## Out of scope

- The other `MainApplication` statics.
- The collection selection and tag-filtering behaviour itself.
- `ResourcesFragment` and the shared resource list.

dogi:
Refactored isCollectionSwitchOn off MainApplication into a private field in CollectionsFragment, updated its reader and writer, and added test assertions verifying independent flag state across fragment instances.

Fixes #17344

---
*PR created automatically by Jules for task [11074301927704462491](https://jules.google.com/task/11074301927704462491) started by @dogi*


dogi:
**Review — rating: 88/100**

Correct move, and it fixes a real bug rather than just tidying a static away.

As a `var` on `MainApplication`'s companion, `isCollectionSwitchOn` lived for the whole process and was never reset. So toggling multi-select on, dismissing the dialog, and reopening it left the flag `true` while the freshly inflated switch widget rendered in its default off state — `buildTagDataList` would build the list in multi-select mode with the visible control disagreeing. Making it a per-instance field means the flag and the widget now start from the same state every time the dialog opens.

I checked the removal is complete: `isCollectionSwitchOn` had exactly one production consumer (`CollectionsFragment`, lines 162 and 233) plus the test, so deleting the declaration from `MainApplication` leaves nothing dangling — and the build passing is itself confirmation, since any other reader would have failed to compile.

The new test asserts the property the change establishes: two fragments holding `true` and `false` simultaneously, which is impossible with a shared static. It's on the weak side in form — `newFragment` injects the value by reflection and the test reads it back the same way — but the *simultaneity* is what matters and that part is genuine.

**Nit:** `import org.ole.planet.myplanet.MainApplication` is now unused in both files. Lines 162 and 233 were its only uses in `CollectionsFragment`, and line 30 was its only use in `CollectionsFragmentTest`, so both imports can go.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #17354, opened 2026-09-17
Refactor DashboardElementActivity data calls to DashboardElementViewModel

### LINKED ISSUE #17337: DashboardElementActivity makes session and repository calls with no ViewModel
## Verified premise

`ui/dashboard/DashboardElementActivity.kt` — an abstract Activity — injects a repository at L35:

```kotlin
import org.ole.planet.myplanet.repository.ActivitiesRepository   // L19
...
lateinit var activitiesRepository: ActivitiesRepository          // L35
```

and runs data and session calls directly in `lifecycleScope` blocks, L107-120:

```kotlin
lifecycleScope.launch {
    val userModel = profileDbHandler.getUserModel()
    ...
    activitiesRepository.recordSyncUserChallengeAction("${userModel?.id}")
}
...
lifecycleScope.launch {
    profileDbHandler.logoutAsync()
}
```

Three data/session operations — current-user lookup, challenge-action record, logout — with no ViewModel between the Activity and the repository, in a class every dashboard screen inherits from.

## Do

- Add `ui/dashboard/DashboardElementViewModel.kt`, a `@HiltViewModel` exposing `suspend fun currentUser()`, `suspend fun recordUserChallengeAction(userId: String)` and `suspend fun logout()`.
- Reroute the three call sites through it.
- Delete the `activitiesRepository` field injection and its import.

## Constraints

- **`profileDbHandler` is inherited from `SyncActivity`, which still uses it for other things. Do not rename it, do not remove it, and do not touch `SyncActivity` at all.** Only this subclass's use of it for `getUserModel()` / `logoutAsync()` moves behind the ViewModel; the field itself stays where it is.
- `recordSyncUserChallengeAction("${userModel?.id}")` currently string-interpolates a nullable id, so a null user produces the literal `"null"`. Preserve that exact behaviour unless you can show no caller depends on it — if you tighten it, say so explicitly in the PR rather than changing it silently.
- This is an `abstract` class with subclasses; adding a ViewModel to it means every subclass gets it. Verify the subclasses still compile and none of them already declare a conflicting `viewModel` property.
- Keep both operations inside `lifecycleScope` at the call sites — the ViewModel exposes suspend functions; it does not take over the Activity's scoping.

## Tests

- Existing `app/src/test/java/org/ole/planet/myplanet/ui/dashboard/` tests must pass **unedited**.
- Add a ViewModel test for each of the three methods, asserting it delegates to the repository / session handler.
- `./gradlew testDefaultDebugUnitTest` and `./gradlew assembleDefaultDebug` (the second proves Hilt resolves the new ViewModel across every subclass).

## Out of scope

- `SyncActivity` and `profileDbHandler`'s definition.
- `DashboardActivity`'s own ViewModel and the challenge dialog.
- The logout flow beyond routing the call.

dogi:
Refactored DashboardElementActivity to delegate current user lookup, challenge action recording, and logout operations through a new @HiltViewModel DashboardElementViewModel. Removed field injection of ActivitiesRepository from DashboardElementActivity and added unit tests for DashboardElementViewModel.

Fixes #17337

---
*PR created automatically by Jules for task [11563855221360146515](https://jules.google.com/task/11563855221360146515) started by @dogi*


dogi:
**Review — rating: 84/100**

Right direction, and the two things that could have made it unsafe both check out:

- **`by viewModels()` on an abstract Activity.** Hilt needs `@AndroidEntryPoint` on the concrete class for the ViewModel factory to resolve, and `DashboardActivity` is the only subclass of `DashboardElementActivity` and is annotated. So there's no subclass that would blow up at runtime while still compiling.
- **The calls are equivalent.** `profileDbHandler` is a `UserSessionManager` (injected on `SyncActivity`), and `UserSessionManager.getUserModel()` is a one-line delegate to `userRepository.getUserModel()` — so `dashboardElementViewModel.currentUser()` reaches the same lookup, just without the intermediate hop. `logoutAsync()` is the same method on the same class.

Dropping `@Inject lateinit var activitiesRepository` from the Activity is the real win: an abstract Activity holding a repository purely to record one action is exactly the coupling this batch has been unwinding.

**Two observations, neither blocking:**

- **It isn't really a ViewModel yet.** Three stateless passthroughs with no `StateFlow` and nothing to survive a configuration change — it's a dependency holder that gives the Activity a DI seam. That's a reasonable incremental step (#17322 does the same), but the three tests inherit the consequence: each asserts that a one-line delegate delegates, which will pass regardless of anything interesting changing.
- **The layering is split down the middle.** `currentUser()` goes straight to `UserRepository` while `logout()` goes through `UserSessionManager`, and the Activity still holds `profileDbHandler` — the same `UserSessionManager` — for its other uses in `SyncActivity`. So there are now two routes to that class in one hierarchy. Not wrong, but worth settling on one as the remaining `profileDbHandler` calls migrate, otherwise the next reader has to work out which path is canonical.

**Verdict:** mergeable.

---

Okuro3499 (approved):

PR #17355, opened 2026-09-17
Optimize checkedCount computation in StorageCategoryUiState

### LINKED ISSUE #17341: storage category state scans its whole item list twice on every checkbox tap
## Verified premise

`ui/settings/StorageCategoryViewModel.kt:19-27`:

```kotlin
data class StorageCategoryUiState(
    val items: List<OfflineResourceItem> = emptyList(),
    val isLoading: Boolean = false,
    val isDeleting: Boolean = false,
    val isEmpty: Boolean = false
) {
    val checkedCount: Int get() = items.count { it.isChecked }
    val allChecked: Boolean get() = checkedCount == items.size && items.isNotEmpty()
}
```

Both are computed **getters**. `checkedCount` walks the whole list; `allChecked` calls `checkedCount`, so reading both walks it twice. `StorageCategoryDetailFragment` reads both on every state emission, and `toggleItemChecked` emits a new state on every single checkbox tap — so a user ticking through a long offline-resource list pays two full scans per tap.

The ViewModel already `map`s the list when building each new state, so the count can be produced during that pass for free.

## Do

- Make `checkedCount` a real constructor property of `StorageCategoryUiState`, computed during the `map` the ViewModel already performs.
- Redefine `allChecked` as a single comparison against the stored count.

## Constraints

- **Keep both names and both types exactly (`checkedCount: Int`, `allChecked: Boolean`) so `StorageCategoryDetailFragment` compiles unchanged. This issue must not touch the fragment.**
- `allChecked` must keep the `&& items.isNotEmpty()` guard — an empty list has `checkedCount == items.size == 0`, and without the guard an empty category would render as "all selected".
- Give `checkedCount` a default (`= 0`) and place it so existing positional constructions of `StorageCategoryUiState` stay valid — check every construction site in the ViewModel before choosing its position.
- Every path that builds a new state must set `checkedCount` consistently with `items`; a state where they disagree is worse than the double scan. Look at all the update sites around L42-73.

## Tests

- Existing `app/src/test/java/org/ole/planet/myplanet/ui/settings/StorageCategoryViewModelTest.kt` must pass **unedited**.
- Add a test that toggling one item updates `checkedCount` and `allChecked` consistently, and one asserting an empty list yields `allChecked == false`.

## Out of scope

- `StorageCategoryDetailFragment`.
- `OfflineResourceItem` and the deletion flow.
- `StorageBreakdownViewModel` (a separate ViewModel).

dogi:
Made checkedCount a stored constructor property on StorageCategoryUiState and computed it during list mapping in StorageCategoryViewModel to avoid double scanning lists on checkbox taps. Redefined allChecked in terms of checkedCount.

Fixes #17341

---
*PR created automatically by Jules for task [18103626160631340865](https://jules.google.com/task/18103626160631340865) started by @dogi*


dogi:
**Review — rating: 84/100**

Denormalizing a derived value is the kind of change that goes wrong when a mutation site gets missed, so I enumerated them. `items` is reassigned in exactly three places — `loadItems` (line 51), `toggleItemChecked` (line 64) and `toggleAllChecked` (line 72) — and all three set `checkedCount`. `deleteItems` only flips `isDeleting` and sends `deleteCompleteEvent`; it never rewrites `items`, so the list is refreshed through `loadItems` afterwards and the count is recomputed there. No drift today.

I also checked the `toggleAllChecked` rewrite, since `items.all { it.isChecked }` and `state.allChecked` disagree on an empty list — `all {}` is vacuously **true** where `allChecked` is explicitly **false**, so `targetChecked` flips the opposite way. It doesn't matter, because mapping over an empty list produces an empty list either way, and the new `empty list yields allChecked false` test pins the observable behaviour. Good that it's covered rather than left implicit.

The toggling test is the right one: it walks `checkedCount` and `allChecked` together across four transitions, including crossing into all-checked and back out, which is exactly where a stale count would show up.

**The trade worth naming:** the old `val checkedCount: Int get() = items.count { it.isChecked }` was impossible to get wrong — it couldn't disagree with `items` by construction. The stored field is correct now but requires every future mutation site to remember it, and `allChecked` depends on it, so a miss would corrupt the select-all control rather than just a label. Against that, the saving is two `count {}` passes per state read over a list of at most a few hundred items. The tests you've added are the thing that makes this safe, so it's fine — but it's discipline bought with test coverage rather than a free win.

**Nit:** the `var checkedCount = 0` incremented inside `map` makes the transform impure and relies on `List.map` being sequential and eager (it is). `updatedItems.count { it.isChecked }` after the map reads more plainly at the cost of one pass — which, given the scale involved, is probably the better trade.

**Verdict:** mergeable.

---

Okuro3499 (approved):

